PR #4, opened 2017-02-18
using non blocking call if possible to generate randomness

devnexen:


mbasaglia:
Linking old request: https://github.com/Unvanquished/Unvanquished/pull/963

mbasaglia:
We already said it was good in the old request.
CI fails but looks like it's just misconfigured.

PR #7, opened 2017-03-11
cmake: Prefer to find the raw lua library instead of a versioned one

DolceTriade:
The FindLua.cmake function prefers to find the lua library that is
formatted like lua<version>.<os suffix>. This leads to it picking
incompatible headers and libraries on certain distros (notably
Arch). Therefore, if we prefer to find lua.<os suffix> it is more
likely to match with the includes.

### DIFF
--- a/cmake/FindLua.cmake
+++ b/cmake/FindLua.cmake
@@ -111,7 +111,7 @@ find_path(LUA_INCLUDE_DIR lua.h
 unset(_lua_include_subdirs)
 
 find_library(LUA_LIBRARY
-  NAMES ${_lua_library_names} lua
+  NAMES lua ${_lua_library_names}
   HINTS
     ENV LUA_DIR
   PATH_SUFFIXES lib


DolceTriade:
Yes, that's my point Viech. It finds the includes for 5.3.4 but it links the library for lua 5.2, which is why you get the linker errors. My change makes it that it finds the library for 5.3.4 as well.

Viech:
Oh well, I didn't test the correct branch because cmake resets the branch in the daemon folder… And it also does this when I checkout the `Melanosuchus/cmake` branch of Unvanquished. Now I'm mildly confused, I thought the latter pull request was supposed to fix that very issue.

DolceTriade:
So, this is kinda confusing, but we have the same "submodule" checkout in two places. On in the Unvanquished level CMakeLists.txt and in the daemon CMakeLists, so we do it twice. Mela's PR deletes the submodule checkout in daemon because it is redundant. We should keep the Unvanquished level one (Which is fixed in https://github.com/Unvanquished/Unvanquished/pull/991) and delete the daemon command that does the same.

DolceTriade:
How about this one?

Viech:
I'm unable to test while Unvanquished/Unvanquished#991 does not work for me. If you tested this with a clean cmake cache and it locates and links the proper library on Arch Linux while without your changes it does not, I'd say go ahaed and merge it or have someone *not* on Arch Linux test this if you fear regressions.

DolceTriade:
I'm going to go ahead and merge this. If this casues problems, I'll revert.

PR #8, opened 2017-03-11
Clean up the demo console interface

mbasaglia:
This is part of the work to improve demos (see Unvanquished/Unvanquished#988).
It does the following:
* Renames cvars to `demo.*`
* Renames commands to `demo_*`
* Removes demo-related functionality that is completely broken (eg: audio recording)
* Fixes CMake to avoid overwriting submodules during builds

### DIFF
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -99,6 +99,13 @@ Cvar::Cvar<std::string> cvar_cl_demo_status_filename(
     ""
 );
 
+Cvar::Cvar<std::string> cvar_cl_demo_next(
+    "cl_demo_next",
+    "Name of the demo to play after the current one",
+    Cvar::NONE,
+    ""
+);
+
 cvar_t *cl_aviFrameRate;
 
 cvar_t *cl_freelook;
@@ -877,19 +884,16 @@ If the "nextdemo" cvar is set, that command will be issued
 */
 void CL_NextDemo()
 {
-	char v[ MAX_STRING_CHARS ];
+	std::string v = cvar_cl_demo_next.Get();
 
-	Q_strncpyz( v, Cvar_VariableString( "nextdemo" ), sizeof( v ) );
-	v[ MAX_STRING_CHARS - 1 ] = 0;
-	Log::Debug( "CL_NextDemo: %s", v );
-
-	if ( !v[ 0 ] )
+	if ( v.empty() )
 	{
 		return;
 	}
 
-	Cvar_Set( "nextdemo", "" );
-	Cmd::BufferCommandTextAfter(v, true);
+	Log::Debug( "CL_NextDemo: %s", v );
+	cvar_cl_demo_next.Set("");
+	Cmd::BufferCommandTextAfter("demo_play " + v, false);
 	Cmd::ExecuteCommandBuffer();
 }
 
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -729,10 +729,6 @@ void CL_WriteWaveOpen()
 		return;
 	}
 
-	// yes ... no ? leave it up to them imo
-	//if (cl_avidemo.integer)
-	//  return;
-
 	if ( Cmd_Argc() == 2 )
 	{
 		s = Cmd_Argv( 1 );
@@ -1419,8 +1415,6 @@ void CL_Connect_f()
 		cls.state = connstate_t::CA_CONNECTING;
 	}
 
-	Cvar_Set( "cl_avidemo", "0" );
-
 	// we need to setup a correct default for this, otherwise the first val we set might reappear
 	Cvar_Set( "com_errorMessage", "" );
 
@@ -1794,7 +1788,6 @@ void CL_Vid_Restart_f()
 	// settings may have changed so stop recording now
 	if ( CL_VideoRecording() )
 	{
-		Cvar_Set( "cl_avidemo", "0" );
 		CL_CloseAVI();
 	}
 
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -738,7 +738,7 @@ void CL_ShutdownAll()
 	cls.soundRegistered = false;
 
 	// Gordon: stop recording on map change etc, demos aren't valid over map changes anyway
-    CL_StopRecord();
+	CL_StopRecord();
 }
 
 /*
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -329,10 +329,10 @@ void CL_StopRecord()
     Log::Notice("%s", "Stopped demo.\n" );
 }
 
-class DemoStopRecordCmd: public Cmd::StaticCmd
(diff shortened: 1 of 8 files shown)


src/engine/client/cl_main.cpp
```diff
@@ -302,213 +307,204 @@ void CL_WriteDemoMessage( msg_t *msg, int headerBytes )
...
+    int len = -1;
+    FS_Write( &len, 4, clc.demofile );
+    FS_Write( &len, 4, clc.demofile );
+    FS_FCloseFile( clc.demofile );
+    clc.demofile = 0;
+
+    clc.demorecording = false;
+    Cvar::SetValueForce(cvar_demo_status_isrecording.Name(), "0");
+    Cvar::SetValueForce(cvar_demo_status_filename.Name(), "");
+    Log::Notice("%s", "Stopped demo.\n" );
```
DolceTriade: Is the extra newline intentional? Log:: functions should add a newline by themselves. (If it is, comment so)

src/engine/client/cl_main.cpp
```diff
@@ -302,213 +307,204 @@ void CL_WriteDemoMessage( msg_t *msg, int headerBytes )
...
-
-record <demoname>
-
-Begins recording a demo from the current position
-====================
-*/
-
-static char demoName[ MAX_QPATH ]; // compiler bug workaround
-void CL_Record_f()
+class DemoStopRecordCmd: public Cmd::StaticCmd
```
DolceTriade: nit: The command is demo_record_stop but the function is DemoStopRecord. Can the two be consistent?

src/engine/client/cl_main.cpp
```diff
@@ -302,213 +307,204 @@ void CL_WriteDemoMessage( msg_t *msg, int headerBytes )
...
-	if ( clc.demorecording )
-	{
-		Log::Warn("Already recording." );
-		return;
-	}
+    void Run(const Cmd::Args&) const OVERRIDE
+    {
+        if ( !clc.demorecording )
+        {
+            Log::Notice("%s", "Not recording a demo.\n" );
```
DolceTriade: ditto comment about newlines.

DolceTriade: Also, how come sometimes you do Log::Notice ("%s", "Static Text");, but other times you just do Log::Notice("Static text"); ?

src/engine/client/cl_main.cpp
```diff
@@ -302,213 +307,204 @@ void CL_WriteDemoMessage( msg_t *msg, int headerBytes )
...
 
-	if ( Cmd_Argc() == 2 )
-	{
-		s = Cmd_Argv( 1 );
-		Q_strncpyz( demoName, s, sizeof( demoName ) );
-		Com_sprintf( name, sizeof( name ), "demos/%s.dm_%d", demoName, PROTOCOL_VERSION );
-	}
+    void Run(const Cmd::Args& args) const OVERRIDE
+    {
+        if ( args.size() > 2 )
```
DolceTriade: Will this prevent people from just calling demo_record?

src/engine/client/cl_main.cpp
```diff
@@ -302,213 +307,204 @@ void CL_WriteDemoMessage( msg_t *msg, int headerBytes )
...
-		{
-			if ( *period == '/' )
-			{
-				break;
-			}
-		}
+        if ( args.size() == 2 )
+        {
+            demo_name = args[1];
+        }
```
DolceTriade: imo, you should generate a name here for readability. It's unclear to the reader that you're going to generate a name if the demo name is empty.

src/engine/client/cl_main.cpp
```diff
@@ -863,31 +678,28 @@ class DemoCmd: public Cmd::StaticCmd {
             return {};
```
DolceTriade: There should already be helper functions for command completion that just lists files in a directory and filters on a pattern.

src/engine/client/cl_main.cpp
```diff
@@ -926,15 +738,7 @@ void CL_ShutdownAll()
...
-	if ( clc.demorecording )
-	{
-		CL_StopRecord_f();
-	}
-
-	if ( clc.waverecording )
-	{
-		CL_WavStopRecord_f();
-	}
+    CL_StopRecord();
```
DolceTriade: indentation looks wrong... I know we are fighting competing styles, but we should at least match the code that we are around.

src/engine/client/cl_parse.cpp
```diff
@@ -234,43 +234,7 @@ void CL_ParseSnapshot( msg_t *msg )
 		{
 			if ( cl_autorecord->integer /*&& Cvar_VariableValue( "g_synchronousClients") */ )
```
DolceTriade: Let's delete the commented out code as well

src/engine/client/cl_main.cpp
```diff
@@ -302,213 +307,204 @@ void CL_WriteDemoMessage( msg_t *msg, int headerBytes )
...
-	Log::Notice("%s", "Stopped demo.\n" );
+    // finish up
+    int len = -1;
+    FS_Write( &len, 4, clc.demofile );
+    FS_Write( &len, 4, clc.demofile );
+    FS_FCloseFile( clc.demofile );
+    clc.demofile = 0;
+
+    clc.demorecording = false;
+    Cvar::SetValueForce(cvar_demo_status_isrecording.Name(), "0");
```
DolceTriade: Why not just use Cvar::Set here?

mbasaglia: Bacause it's a read-only cvar and Set doesn't work on those.

src/engine/client/cl_main.cpp
mbasaglia: CL_Record is called elsewhere with an automatically-generated name, it used to have the generation code duplicated so I moved it in CL_Record.

src/engine/client/cl_main.cpp
mbasaglia: Nope

src/engine/client/cl_main.cpp
mbasaglia: Isn't it already using it?

src/engine/client/cl_main.cpp
DolceTriade: Weird. Didn' t know that.

src/engine/client/cl_main.cpp
DolceTriade: I guess. It's just unclear from reading this that it does that. I don't feel strongly here.

src/engine/client/cl_main.cpp
DolceTriade: Read it wrong. :(

src/engine/client/cl_main.cpp
mbasaglia: It was already like that so I don't know

src/engine/client/cl_main.cpp
DolceTriade: It is. My bad.

Viech:
Note that stuff like 9c3071a should ideally go in a PR of their own. 😉

PR #10, opened 2017-03-14
few memory leak fixes

devnexen:


src/engine/framework/System.cpp
```diff
@@ -282,6 +282,7 @@ static void Shutdown(bool error, Str::StringRef message)
     Application::Shutdown(error, message);
 
 	// Always run CON_Shutdown, because it restores the terminal to a usable state.
+	Cvar_ClearCvarMap();
```
Kangz: This should be above the comment, in an if(!error) block. When an error occurs we want to just quit as fast as possible (but need to reset the console to avoid garbling it).

src/engine/framework/CvarSystem.h
```diff
@@ -88,6 +88,7 @@ namespace Cvar {
     void SetValueCProxy(const std::string& cvarName, const std::string& value);
 
     void SetCheatsAllowed(bool allowed);
+    void ClearCvarMap();
```
Kangz: nit: void Shutdown()

src/engine/framework/CvarSystem.cpp
```diff
@@ -168,6 +168,34 @@ namespace Cvar {
         return *cvars;
     }
 
+	void ClearCvarMap() {
+		CvarMap _cvars = GetCvarMap();
+
+		for (auto it: _cvars) {
```
mbasaglia: maybe it could be a const reference.

src/engine/framework/CvarSystem.cpp
```diff
@@ -168,6 +168,34 @@ namespace Cvar {
...
+			if (cvar.latchedString)
+			{
+				Z_Free(cvar.latchedString);
+			}
+
+			if (cvar.string)
+			{
+				Z_Free(cvar.string);
+			}
+		}
```
Kangz: Just for good measure, can you call cvars.clear()?

src/engine/framework/CvarSystem.cpp
```diff
@@ -168,6 +168,34 @@ namespace Cvar {
         return *cvars;
     }
 
+	void ClearCvarMap() {
+		CvarMap _cvars = GetCvarMap();
```
Kangz: nit: _cvars -> cvars

src/engine/framework/CvarSystem.cpp
mbasaglia: I believe at the end here, you should also clear the map itself (ie: removing its elements)

src/engine/qcommon/cvar.cpp
```diff
@@ -169,3 +169,7 @@ Cvar_InfoString
 char* Cvar_InfoString(int flag, bool big) {
     return Cvar::InfoString(flag, big);
 }
+
```
Kangz: This and the change in cvar.h are unnecessary

src/engine/framework/CvarSystem.cpp
mbasaglia: _cvars should be a reference

src/engine/client/cl_keys.cpp
```diff
@@ -1676,6 +1676,22 @@ void CL_InitKeyCommands()
...
+{
+	int team, keynum;
+
+	for ( team = 0; team < MAX_TEAMS; team++ )
+	{
+		for ( keynum = 0; keynum < MAX_KEYS; keynum++ )
+		{
+			if ( keys[ keynum ].binding[ team ] )
+			{
+				Z_Free( keys[ keynum ].binding[ team ] );
```
mbasaglia: Other places that free `keys[ keynum ].binding[ team ]`, also assign it to **nullptr**, perhaps this should do the same.

src/engine/qcommon/net_ip.cpp
```diff
@@ -2055,6 +2063,31 @@ void NET_Shutdown()
...
+		geoip_data_6 = nullptr;
+	}
+
+	if ( geoip_data_4 )
+	{
+		GeoIP_delete(geoip_data_4);
+		geoip_data_4 = nullptr;
+	}
+
+	if (GeoIPDBFileName)
```
Kangz: call GeoIP_cleanup instead

src/engine/client/cl_input.cpp
```diff
@@ -1342,3 +1342,16 @@ void CL_ClearKeys()
...
+
+void CL_ClearInput()
+{
+	int i;
+
+	for ( i = 0; i < USERCMD_BUTTONS; ++i )
+	{
+		if ( registeredButtonCommands[ i ] )
+		{
+			Z_Free( registeredButtonCommands[i] );
```
mbasaglia: Other places that free `registeredButtonCommands[ i ]`, also assign it to **nullptr**, perhaps this should do the same.

devnexen:
good points :-)

Kangz:
Thanks for the fixes, LGTM!

mbasaglia:
Looks good to me too, merging

PR #11, opened 2017-03-15
Generate tr.charsetImage at runtime from a .ttf file

DefaultUser:
This eliminates the need for the bitmap file at gfx/2d/consolechars. AFAIS this is only needed for debug (compile with REFBONE_NAMES), so there should not be much difference


mbasaglia:
From Travis:
```
CMake Error at /usr/share/cmake-3.2/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find SDL2TTF (missing: SDL2TTF_LIBRARY SDL2TTF_INCLUDE_DIR)
```

Can you make sure `.travis.yml` installs `libsdl2-ttf-dev`?

mbasaglia:
BTW, not strictly related but, as part of one of my other projects, I wrote a CMake find script that generalizes SDL2 modules (see https://github.com/mbasaglia/CMake-Lib/blob/master/modules/FindSDL2.cmake) if needed we can use it in Daemon.
Usage would be `find_package(SDL2 REQUIRED ttf)`.

src/engine/renderer/tr_image.cpp
```diff
@@ -2790,15 +2794,79 @@ void R_CreateBuiltinImages()
...
+			c[0] = i*16+j;
+			charSurf = TTF_RenderText_Solid ( font, c, {255, 255, 255} );
+			SDL_BlitSurface ( charSurf, NULL, charsetSurf, &dest );
+			SDL_FreeSurface ( charSurf );
+		}
+	}
+	SDL_LockSurface ( charsetSurf );
+	byte  data[ CHARSETIMAGE_CHARSIZE*16 ][ CHARSETIMAGE_CHARSIZE*16 ][ 4 ];
+	byte  *dataPtr = &data[0][0][0];
+	byte  *out = &data[ 0 ][ 0 ][ 0 ];
```
mbasaglia: Any reason data is declared as a 3D array but used as a flat array?
You could just declare byte  `data[ CHARSETIMAGE_CHARSIZE*16 * CHARSETIMAGE_CHARSIZE*16 * 4 ];`

DefaultUser: This code was part of R_CreateBuiltinImages so I reused these 3 arrays

src/engine/renderer/tr_image.cpp
```diff
@@ -2790,15 +2794,79 @@ void R_CreateBuiltinImages()
...
+			c[0] = i*16+j;
+			charSurf = TTF_RenderText_Solid ( font, c, {255, 255, 255} );
+			SDL_BlitSurface ( charSurf, NULL, charsetSurf, &dest );
+			SDL_FreeSurface ( charSurf );
+		}
+	}
+	SDL_LockSurface ( charsetSurf );
+	byte  data [ CHARSETIMAGE_CHARSIZE*16 * CHARSETIMAGE_CHARSIZE*16 * 4 ];
+	byte  *dataPtr = &data [ 0 ];
+	byte  *out = &data [ 0 ];
```
mbasaglia: Now you should be able to just use `data` instead of `&data [ 0 ]`

mbasaglia:
Changes look good, is anyone opposed to the new dependency to sdl2-ttf or worried about overhead?

Viech:
If the dependency is optional for users of `REFBONE_NAMES` (whatever that even does) I don't see a problem.

Kangz:
LGTM too.

DolceTriade:
Could you also note why you didn't opt to use the existing ttf bitmap generation code in your commit message?

DefaultUser:
I missed that there was already ttf bitmap generation code in the engine. If you want me to, I can adapt this branch to use that and make some performance measurements

DolceTriade:
Using existing code would be preferable to duplicating code and adding dependencies.

Viech:
I don't know the code, but ttf bitmap generation *sounds* like something that is better done by an external, specialized library than maintaining it yourself.

DolceTriade:
Except when that code already exists and is well tested :)

mbasaglia:
@DolceTriade are you referring to the functions in https://github.com/DaemonEngine/Daemon/blob/master/src/engine/renderer/tr_font.cpp?

DolceTriade:
Yes those functions.

mbasaglia:
The only documentation for them is "font stuff", how is someone supposed to know how to use them?
https://github.com/DaemonEngine/Daemon/blob/b6e41f7baae75adbd7121e70d7b73b7e19ec4b6a/src/engine/renderer/tr_local.h#L3960

DefaultUser:
I've tried using these font functions for the last few days and haven't been successfull. Most of the time I just get segfaults. I'm probably not using it correctly but there are nearly no comments. To me that system seems overly complicated.

The closest I got is using the console font bitmap from r_imageHashTable, but that's not really an elegant solution. Also, the rest of the code to display the bone names does only expect a very certain image size (there is some hardcoded magic values like 0.0625 in that code with again no comment) so I'm currently trying to fix that.

illwieckz:
Is this one obsoleted by #12? Can we close this one?

DefaultUser:
After #12 this PR is no longer needed, I'm closing it

PR #13, opened 2017-03-23
do not prefer sound file path with same extension, always use newer one

illwieckz:
- before 4ae82bf daemon was not able to load extension independent sound file path
- before this commit, daemon was preferring sound file path with given extension
  it was not a good idea, if a legacy pk3 provided sound.wav and an updated pk3
  provides sound.ogg and the game code looks for sound.wav, the newer sound.ogg
  will never be read, this commit fixes that issue

With this fix, daemon loading behavior for sound file paths becomes truly extension independent.

This is needed since we plan to update sound assets, we must be sure the latest one is loaded.

### DIFF
--- a/src/engine/audio/SoundCodec.cpp
+++ b/src/engine/audio/SoundCodec.cpp
@@ -55,21 +55,6 @@ AudioData LoadSoundCodec(std::string filename)
 
 	std::string ext = FS::Path::Extension(filename);
 
-	// if filename has extension, try to load it
-	if (ext != "") {
-		// look for the correct loader and use it
-		for (int i = 0; i < numSoundLoaders; i++) {
-			if (ext == soundLoaders[i].ext) {
-				// if file exists, load it
-				if (FS::PakPath::FileExists(filename)) {
-					return soundLoaders[i].SoundLoader(filename);
-				}
-			}
-		}
-	}
-
-	// if filename does not have extension or there is no file with such extension
-	// or if there is no codec available for this file format,
 	// try and find a suitable match using all the sound file formats supported
 	// prioritize with the pak priority
 	int bestLoader = -1;


illwieckz:
That was an old commit that was sleeping on my local repo since ages. I use it when I have to deal with some sound pk3dir.

DolceTriade:
I don't think this is the right solution to commit to master. You should be able to override our selection logic by specifying the extension. If we hardcode an extension anywhere, that should be removed instead, and all assets should be modified to not hard code an extension.

illwieckz:
Oh, I already have an old branch for that on Unvanquished side. :smile:

illwieckz:
It looks like texture finding also looks for extension firts:
[src/engine/renderer/tr_image.cpp:1679](https://github.com/DaemonEngine/Daemon/blob/master/src/engine/renderer/tr_image.cpp#L1679)

I was believing it was just looking for the file whatever the extension in any case. Well, it's better to have an unified behavior across all formats, even if I don't see any reason to enforce a given format.

Kangz:
+1 on what @DolceTriade said.

illwieckz:
The game code checks in some place the hardcoded sound path and abort even if the sound system is able to find an alternative with another extension.

See:
https://github.com/Unvanquished/Unvanquished/issues/992
https://github.com/Unvanquished/Unvanquished/issues/993

The fun fact was: with that commit that was making the system not preferring hardcoded extension, it was still needed to supply a `sound.wav` to load a `sound.ogg` alternative if `sound.wav` path were hardcoded in voice or buildable sound configuration.

illwieckz:
So, I close this PR, but it must be noticed that until Unvanquished/Unvanquished#992 and Unvanquished/Unvanquished#993 are fixed, the game code is not sound file extension independent yet even if the engine is.

PR #14, opened 2017-03-27
introduce dpk format

illwieckz:
See [this forum post](http://forums.unvanquished.net/viewtopic.php?f=58&t=2122&p=17813) for the reasons and explanations behind this change.
See [this merge request](https://gitlab.com/xonotic/netradiant/merge_requests/39) for related changes in NetRadiant/q3map2.

What this PR does:

- dpk/dpkdir format for versioned packages

- pk3/pk3dir are for legacy unversioned packages

- legacy pk3/pk3dir now uses internal version "" so
  * it's always smaller than a dpk/dpkdir version since dpk/dpkdir forbid this version
  * code can rely on that version to know if it's a legacy pak, and it now does

- the engine now loads dpk as versioned ones and pk3 as legacy unversioned one
  when fs_legacypaks is set to on,  previously the code was loading versioned pk3
  as unversioned one when fs_legacypaks was set to on

- the engine now prefers the versioned pak over the unversioned one if both exist

- this code fixes the autodownload mechanism for legacy pk3s
  for sure, it was not able to work before since the code was adding version string
  to legacy pk3s that do not have version string, now the code relies on that
  special internal "" version to compute legacy pk3 names

- modify FS::MakePakeName() to include + extension
  * engine/qcommon/files.cpp uses it now instead of calling FS::MakePakName()
  plus adding extension on every FS::MakePakName() call
  * src/engine/server/sv_client.cpp uses it instead of duplicating Fs:MakePakName()
  plus adding extension

- factorise pak name parsing
  FS:PakParseName is used to parse both legacy and non-legacy pak names

- always look for pak with version, this way, if server host a legacy pak
  but the client owns both legacy and non-legacy paks sharing the same name,
  the client is ensured to load the legacy one.

- do not compare pk3 checksum
  legacy paks can't have checksum, so do not check it

- ensure dpk paks with bad version are not loaded when fs_legacypaks is enabled

- do not read DEPS from legacy paks

- enable pk3 downloading

Previously, the server was sending pak list to client that way:

[
	name1_version_checksum,
	name2,
	n_a_m_e_3,
	name4_version_checksum,
]

So, the client was not able to parse this names with dpk/pk3
specificities, and to know the file format.

Now, the server sends pak list to client that way:

[
	name1_version_checksum.dpk,
	name2.pk3,
	n_a_m_e_3.pk3,
	name4_version_checksum.dpk,
]

Extension is used by client to know how to parse these names (pk3 have
empty version and name can't contain checksum) and reconstruct final
file name in local file system. On local file system the package can
be a dpkdir or pk3dir, that information is not told in that list since
there is no need to tell it.

So, this PR not only introduces `dpk` format, it is also meant to fix autodownloading legacy `pk3` that was broken before, and factorize some code.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -122,16 +122,12 @@ template<> struct SerializeTraits<FS::LoadedPakInfo> {
 
 namespace FS {
 
-static Cvar::Cvar<bool> fs_legacypaks("fs_legacypaks", "Don't use versioned pk3 files", Cvar::NONE, false);
+static Cvar::Cvar<bool> fs_legacypaks("fs_legacypaks", "Also load pk3s, ignoring version.", Cvar::NONE, false);
 
 bool UseLegacyPaks()
 {
 	return *fs_legacypaks;
 }
-
-// Pak zip and directory extensions
-#define PAK_ZIP_EXT ".pk3"
-#define PAK_DIR_EXT ".pk3dir/"
 
 // Error variable used by throws(). This is never written to and always
 // represents a success value.
@@ -1050,6 +1046,7 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 	bool hasDeps = false;
 	offset_t depsOffset;
 	ZipArchive zipFile;
+	bool isLegacy = pak.version.empty();
 
 	// Check if this pak has already been loaded to avoid recursive dependencies
 	for (auto& x: loadedPaks) {
@@ -1059,8 +1056,20 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 			return;
 	}
 
-	// Add the pak to the list of loaded paks
-	fsLogs.Notice("Loading pak '%s'...", pak.path.c_str());
+	if (pak.type == pakType_t::PAK_ZIP) {
+		if (!isLegacy) {
+			fsLogs.Notice("Loading pak '%s'...", pak.path.c_str());
+		} else {
+			fsLogs.Notice("Loading legacy pak '%s'...", pak.path.c_str());
+		}
+	} else if (pak.type == pakType_t::PAK_DIR) {
+		if (!isLegacy) {
+			fsLogs.Notice("Loading pakdir '%s'...", pak.path.c_str());
+		} else {
+			fsLogs.Notice("Loading legacy pakdir '%s'...", pak.path.c_str());
+		}
+	}
+
 	loadedPaks.emplace_back();
 	loadedPaks.back().name = pak.name;
 	loadedPaks.back().version = pak.version;
@@ -1103,7 +1112,7 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 
 		// Get the file list and calculate the checksum of the package (checksum of all file checksums)
 		realChecksum = crc32(0, Z_NULL, 0);
-		zipFile.ForEachFile([&pak, &realChecksum, &pathPrefix, &hasDeps, &depsOffset](Str::StringRef filename, offset_t offset, uint32_t crc) {
+		zipFile.ForEachFile([&pak, &realChecksum, &pathPrefix, &hasDeps, &depsOffset, &isLegacy](Str::StringRef filename, offset_t offset, uint32_t crc) {
 			// Note that 'return' is effectively 'continue' since we are in a lambda
 			if (!Str::IsPrefix(pathPrefix, filename) && filename != PAK_DEPS_FILE)
 				return;
@@ -1113,8 +1122,13 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 				fsLogs.Warn("Invalid filename '%s' in pak '%s'", filename, pak.path);
 				return;
 			}
-			realChecksum = crc32(*realChecksum, reinterpret_cast<const Bytef*>(&crc), sizeof(crc));
-			if (filename == PAK_DEPS_FILE) {
+
+			// Legacy paks don't have version neither checksum
+			if (!isLegacy) {
+				realChecksum = crc32(*realChecksum, reinterpret_cast<const Bytef*>(&crc), sizeof(crc));
+			}
+
+			if (!isLegacy && (filename == PAK_DEPS_FILE)) {
 				hasDeps = true;
 				depsOffset = offset;
 				return;
@@ -1127,29 +1141,34 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 		}, err);
 		if (err)
 			return;
-
-		// Get the timestamp of the pak
-		loadedPaks.back().timestamp = FS::RawPath::FileTimestamp(pak.path, err);
-		if (err)
-			return;
 	}
 
-	// Save the real checksum in the list of loaded paks (empty for directories)
+	// Save the real checksum in the list of loaded paks (empty for directories, not used for legacy paks)
 	loadedPaks.back().realChecksum = realChecksum;
+
+	// Get the timestamp of the pak
+	loadedPaks.back().timestamp = FS::RawPath::FileTimestamp(pak.path, err);
+	if (err)
+		return;
+
 	loadedPaks.back().pathPrefix = pathPrefix;
 
-	// If an explicit checksum was requested, verify that the pak we loaded is the one we are expecting
-	if (expectedChecksum && realChecksum != *expectedChecksum) {
-		SetErrorCodeFilesystem(err, filesystem_error::wrong_pak_checksum);
-		return;
+	// Legacy paks don't have version neither checksum
+	if (!isLegacy) {
+		// If an explicit checksum was requested, verify that the pak we loaded is the one we are expecting
+		if (expectedChecksum && realChecksum != *expectedChecksum) {
+			SetErrorCodeFilesystem(err, filesystem_error::wrong_pak_checksum);
+			return;
+		}
+
+		// Print a warning if the checksum doesn't match the one in the filename
+		if (pak.checksum && *pak.checksum != realChecksum)
+			fsLogs.Warn("Pak checksum doesn't match filename: %s", pak.path);
 	}
 
-	// Print a warning if the checksum doesn't match the one in the filename
-	if (pak.checksum && *pak.checksum != realChecksum)
-		fsLogs.Warn("Pak checksum doesn't match filename: %s", pak.path);
-
 	// Load dependencies, but not if a checksum was specified
-	if (hasDeps && !expectedChecksum) {
+	// Do not look for dependencies if it's a legacy pak (pk3)
+	if (!isLegacy && hasDeps && !expectedChecksum) {
 		std::string depsData;
 		if (pak.type == pakType_t::PAK_DIR) {
 			File depsFile = RawPath::OpenRead(Path::Build(pak.path, PAK_DEPS_FILE), err);
@@ -2155,19 +2174,28 @@ bool IsInitialized()
 // Add a pak to the list of available paks
 static void AddPak(pakType_t type, Str::StringRef filename, Str::StringRef basePath)
 {
+	bool isPakDir = (type == pakType_t::PAK_DIR);
 	std::string fullPath = Path::Build(basePath, filename);
 
-	size_t suffixLen = type == pakType_t::PAK_DIR ? strlen(PAK_DIR_EXT) : strlen(PAK_ZIP_EXT);
 	std::string name, version;
 	Util::optional<uint32_t> checksum;
-	if (!ParsePakName(filename.begin(), filename.end() - suffixLen, name, version, checksum) || (type == pakType_t::PAK_DIR && checksum)) {
-		if (!UseLegacyPaks()) {
-			fsLogs.Warn("Invalid pak name: %s", fullPath);
-			return;
+
+	if (!ParsePakName(filename.begin(), filename.end(), name, version, checksum) || (isPakDir && checksum)) {
+		fsLogs.Warn("Invalid pak name: '%s'", fullPath);
+		return;
+	}
+
+	if (isPakDir) {
+		if (version.empty()) {
+			fsLogs.Verbose("Found legacy pakdir: '%s'...", fullPath);
 		} else {
-			fsLogs.Notice("Loading legacy pak: %s", fullPath);
-			name = filename.substr(0, filename.size() - suffixLen);
-			version = "1";
+			fsLogs.Verbose("Found pakdir: '%s'...", fullPath);
+		}
+	} else {
+		if (version.empty()) {
+			fsLogs.Verbose("Found legacy pak: '%s'...", fullPath);
+		} else {
+			fsLogs.Verbose("Found pak: '%s'...", fullPath);
 		}
 	}
 
@@ -2178,14 +2206,20 @@ static void AddPak(pakType_t type, Str::StringRef filename, Str::StringRef baseP
 static void FindPaksInPath(Str::StringRef basePath, Str::StringRef subPath)
 {
 	std::string fullPath = Path::Build(basePath, subPath);
+	bool useLegacyPaks = UseLegacyPaks();
 	try {
 		for (auto& filename: RawPath::ListFiles(fullPath)) {
 			if (Str::IsSuffix(PAK_ZIP_EXT, filename)) {
 				AddPak(pakType_t::PAK_ZIP, Path::Build(subPath, filename), basePath);
 			} else if (Str::IsSuffix(PAK_DIR_EXT, filename)) {
 				AddPak(pakType_t::PAK_DIR, Path::Build(subPath, filename), basePath);
-			} else if (Str::IsSuffix("/", filename))
+			} else if (useLegacyPaks && Str::IsSuffix(LEGACY_PAK_ZIP_EXT, filename)) {
+				AddPak(pakType_t::PAK_ZIP, Path::Build(subPath, filename), basePath);
+			} else if (useLegacyPaks && Str::IsSuffix(LEGACY_PAK_DIR_EXT, filename)) {
+				AddPak(pakType_t::PAK_DIR, Path::Build(subPath, filename), basePath);
+			} else if (Str::IsSuffix("/", filename)) {
 				FindPaksInPath(basePath, Path::Build(subPath, filename));
+			}
 		}
 	} catch (std::system_error&) {
 		// If there was an error reading a directory, just ignore it and go to
@@ -2351,16 +2385,42 @@ const PakInfo* FindPak(Str::StringRef name, Str::StringRef version, uint32_t che
 	return &*(iter - 1);
 }
 
+// This function is only used to parse dpk/dpkdir package names (pk3 name can't contain version neither checksum)
 bool ParsePakName(const char* begin, const char* end, std::string& name, std::string& version, Util::optional<uint32_t>& checksum)
 {
-	const char* nameStart = std::find(std::reverse_iterator<const char*>(end), std::reverse_iterator<const char*>(begin), '/').base();
-	if (nameStart != begin)
+	const char* nameStart;
+
+	if (Str::IsSuffix(LEGACY_PAK_ZIP_EXT, begin)) {
+		name.assign(begin, end - strlen(FS::LEGACY_PAK_ZIP_EXT));
+		// prefer versioned dpk over legacy pk3 if pak name collides, uses the smallest version string available
+		// this version can't be found in standard dpk, so we can also uses this value to test later if it's a legacy pk3
+		version = "";
+		// legacy paks can't have checksum
+		checksum = Util::nullopt;
+		return true;
+	} else if (Str::IsSuffix(LEGACY_PAK_DIR_EXT, begin)) {
+		name.assign(begin, end - strlen(FS::LEGACY_PAK_DIR_EXT));
+		// empty version string for legacy pak, see above for explanations
+		version = "";
+		// dir can't have checksum
+		checksum = Util::nullopt;
+		return true;
+	} else if (Str::IsSuffix(PAK_ZIP_EXT, begin)) {
+		end -= strlen(PAK_ZIP_EXT);
+	} else if (Str::IsSuffix(PAK_DIR_EXT, begin)) {
+		end -= strlen(PAK_DIR_EXT);
+	}
+
+	nameStart = std::find(std::reverse_iterator<const char*>(end), std::reverse_iterator<const char*>(begin), '/').base();
+	if (nameStart != begin) {
 		nameStart++;
+	}
 
 	// Get the name of the package
 	const char* underscore1 = std::find(nameStart, end, '_');
-	if (underscore1 == end || (UseLegacyPaks() && underscore1 == begin))
+	if (underscore1 == end) {
 		return false;
+	}
 	name.assign(begin, underscore1);
 
 	// Get the version of the package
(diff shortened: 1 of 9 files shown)


illwieckz:
The only thing I haven't tested yet is autodownloading.

Here I describe my test environment, on engine path:

```
pkg/map-antares_a04+1.dpk
pkg/map-antares_a04.dpk
pkg/map-chasm_b2.dpk
pkg/map-forlorn_a15.dpk
pkg/map-parpax_d05.dpk
pkg/map-perseus_b5a+1.dpk
pkg/map-plat23_b13+1.dpk
pkg/map-spacetracks_1.0+1.dpk
pkg/map-station15_1.0+1.dpk
pkg/map-station15_1.0+2.dpk
pkg/map-station15_1.0.dpk
pkg/map-thunder_b3+2.dpk
pkg/map-vega_b4.dpk
pkg/map-yocto_b2c+1.dpk
pkg/tex-all_2.0.dpk
pkg/tex-common_2.0.dpk
pkg/tex-ej01-clean_1.0.dpk
pkg/tex-ej01-common_1.0.dpk
pkg/tex-ej01-ice_1.0.dpk
pkg/tex-ex_1.0.dpk
pkg/tex-exm_1.0.dpk
pkg/tex-pk01_1.0.dpk
pkg/tex-pk02_1.0.dpk
pkg/tex-space_1.0.dpk
pkg/tex-tech_1.0.dpk
pkg/tex-trak5_1.0.dpk
pkg/tex-vega_b4.dpk
pkg/unvanquished_0.25.0.dpk
pkg/unvanquished_0.26.0.dpk
pkg/unvanquished_0.27.0.dpk
pkg/unvanquished_0.28.0.dpk
pkg/unvanquished_0.29.0.dpk
pkg/unvanquished_0.30.0.dpk
pkg/unvanquished_0.31.0.dpk
pkg/unvanquished_0.32.0.dpk
pkg/unvanquished_0.33.0.dpk
pkg/unvanquished_0.34.0.dpk
pkg/unvanquished_0.35.0.dpk
pkg/unvanquished_0.36.0.dpk
pkg/unvanquished_0.37.0.dpk
pkg/unvanquished_0.38.0.dpk
pkg/unvanquished_0.39.0.dpk
pkg/unvanquished_0.40.0.dpk
pkg/unvanquished_0.41.0.dpk
pkg/unvanquished_0.42.0.dpk
pkg/unvanquished_0.43.0.dpk
pkg/unvanquished_0.44.0.dpk
pkg/unvanquished_0.45.0.dpk
pkg/unvanquished_0.46.0.dpk
pkg/unvanquished_0.47.0.dpk
pkg/unvanquished_0.48.0.dpk
pkg/unvanquished_0.49.0.dpk
pkg/unvanquished_0.50.0.dpk
```

on homepath:

```
pkg/map-1984b4.pk3dir
pkg/map-flap.pk3
pkg/map-parpax.pk3
pkg/map-rsmse_test.dpkdir
pkg/map-thermal.pk3dir
```

I was able to load _chasm_ map from `map-chasm_b2.dpk`, _rsmse_ map from `map-rsmse_test.dpkdir`, _flap_ from `map-flap.pk3`, and both _1984b4_ map from `map-1984b4.pk3dir` and `map-thermal.pk3dir`.

Of course loading badly named pk3/bsp does not work, it's out of scope there, and it's not my own problem.

The `map-parpax.pk3` in homepath was the vega one renamed, and the engine successfully loads `map-parpax_d05.dpk` instead, loading the _parpax_ map instead of the _vega_ one.

illwieckz:
```
$ mkdir etmain
$ cp ~/.etlegacy/etmain/pak0.pk3 etmain/pak0.pk3
$ mkdir etmain/map-railgun_2.60b.dpkdir/
$ ./daemon -set fs_legacypaks on -pakpath etmain +devmap railgun
```

Works.

illwieckz:
legacy pk3dir relying on versioned dpk using DEPS (!)

```
$ cp …/xonotic/data/xonotic-20150825-data.pk3 xonotic/data_20150825.dpk
$ cp …/xonotic/data/xonotic-20150825-maps.pk3 xonotic/maps_20150825.dpk
$ mkdir xonotic/map-drain.pk3dir
$ vim xonotic/map-drain.pk3dir/DEPS
$ cat xonotic/map-drain.pk3dir/DEPS
data
maps
$ ./daemon -set fs_legacypaks on -pakpath xonotic +devmap drain
```

Works

illwieckz:
This change also meant a game like _Xonotic on Daemon_ can switch to `dpk` versioned format for their official packages while allowing third-party legacy unversioned `pk3` being loaded. It adds safety since now the engine set priority for `dpk`over `pk3`.

illwieckz:
Beware, do not merge this PR too fast, we must first ensure every mapper is able to switch to dpk vfs, so it's better to wait for myself uploading the assets repositories I'm working on (it's now a matter of days).

src/common/FileSystem.cpp
```diff
@@ -2386,12 +2395,36 @@ bool ParsePakName(const char* begin, const char* end, std::string& name, std::st
 	return true;
 }
 
+std::string MakePakFileName(Str::StringRef name, Str::StringRef version, Util::optional<uint32_t> checksum)
+{
+	if ( version.compare("") != 0 ) {
+		// versioned dpk
+		return Str::Format("%s.%s", MakePakName(name, version, *checksum), "pk3");
```
DefaultUser: Should "versioned dpk" paks not have the suffix "dpk" and "legacy unversioned pk3" have the suffix "pk3"?

src/common/FileSystem.cpp
```diff
@@ -2386,12 +2395,36 @@ bool ParsePakName(const char* begin, const char* end, std::string& name, std::st
 	return true;
 }
 
+std::string MakePakFileName(Str::StringRef name, Str::StringRef version, Util::optional<uint32_t> checksum)
+{
+	if ( version.compare("") != 0 ) {
```
DefaultUser: For readability I would prefer:
`if ( !version.empty() )`

src/common/FileSystem.cpp
illwieckz: oh yes, that's the only part I haven't fully tested yet since it's used in autodownload stuff (have to setup a server etc.), nice catch!

src/common/FileSystem.cpp
illwieckz: I agree!

illwieckz: fixed

src/common/FileSystem.cpp
illwieckz: fixed

src/common/FileSystem.cpp
```diff
@@ -2155,9 +2157,14 @@ bool IsInitialized()
...
 static void AddPak(pakType_t type, Str::StringRef filename, Str::StringRef basePath)
 {
+	size_t suffixLen;
 	std::string fullPath = Path::Build(basePath, filename);
 
-	size_t suffixLen = type == pakType_t::PAK_DIR ? strlen(PAK_DIR_EXT) : strlen(PAK_ZIP_EXT);
+	if (!UseLegacyPaks()) {
+		suffixLen = type == pakType_t::PAK_DIR ? strlen(PAK_DIR_EXT) : strlen(PAK_ZIP_EXT);
+	} else {
+		suffixLen = type == pakType_t::PAK_DIR ? strlen(LEGACY_PAK_DIR_EXT) : strlen(LEGACY_PAK_ZIP_EXT);
```
illwieckz: that part makes no sense at all and only works because `strlen(pk3) == strlen(dpk)`, I'm fixing this.

illwieckz: fixed

illwieckz:
This new code adds a `LoadLegacyPak()` function, so the code checks against the right extension length. It was not a real issue since the two known extensions had same length and these extensions were hardcoded, so the code would always have been running well, but the new code is cleaner.

My first code was likely to loads dpk without version when `fs_legacypaks` was set to `on`, the new code ensures this to not occurs.

This new code also forbid `dpk` zip version to start with a letter, this way people can't make the mistakes to name their `map-castle_b5.dpk` with a version number greater than `map-castle_1.0.dpk`, the right way to specify a alpha/beta version would be `map-castle_0.5b.dpk` for example, if you really want to write this information in version string. The `dpkdir` loading code allows for version to start with a letter, this way, you can ensure your source or testing `dpkdir` is always greater than your release `dpk`, using `src` or `test` version number for example. The `dpkdir` code checks if version starts with alphanumeric character (will not load weird versions like `map-castle_~.pk3dir` but will load `map-castle_0~me.pk3dir`).

illwieckz:
After @Viech remarks I modified the version check to only check on dpk. I agree it's not meant to change the debian sorting, it is meant to rely on it.

src/common/FileSystem.cpp
```diff
@@ -122,16 +122,18 @@ template<> struct SerializeTraits<FS::LoadedPakInfo> {
 
 namespace FS {
 
-static Cvar::Cvar<bool> fs_legacypaks("fs_legacypaks", "Don't use versioned pk3 files", Cvar::NONE, false);
+static Cvar::Cvar<bool> fs_legacypaks("fs_legacypaks", "Loads unversioned legacy pk3 files too", Cvar::NONE, false);
```
DolceTriade: Also load pk3s, ignoring version.

src/common/FileSystem.cpp
```diff
@@ -1114,7 +1129,8 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 				return;
 			}
 			realChecksum = crc32(*realChecksum, reinterpret_cast<const Bytef*>(&crc), sizeof(crc));
-			if (filename == PAK_DEPS_FILE) {
+			// can't reuse isLegacy in lambda
```
DolceTriade: You can if you capture it in the capture stanza like so:

		zipFile.ForEachFile([&pak, &realChecksum, &pathPrefix, &hasDeps, &depsOffset, &isLegacy](Str::StringRef filename, offset_t offset, uint32_t crc) {

src/common/FileSystem.cpp
```diff
@@ -2155,37 +2171,77 @@ bool IsInitialized()
...
-		if (!UseLegacyPaks()) {
-			fsLogs.Warn("Invalid pak name: %s", fullPath);
+
+	if (!ParsePakName(filename.begin(), filename.end() - suffixLen, name, version, checksum) || (isPakDir && checksum)) {
+       fsLogs.Warn("Invalid pak name: '%s'", fullPath);
+       return;
+	}
+
+	if (type == pakType_t::PAK_ZIP) {
+		if (!isdigit(version.at(0))) {
```
DolceTriade: This shouldn't be required. Our version sorting algorithm accounts for letters in the version.

src/common/FileSystem.cpp
```diff
@@ -2155,37 +2171,77 @@ bool IsInitialized()
...
+	size_t suffixLen;
+	bool isPakDir = (type == pakType_t::PAK_DIR);
+	std::string fullPath = Path::Build(basePath, filename);
+
+	suffixLen = type == pakType_t::PAK_DIR ? strlen(LEGACY_PAK_DIR_EXT) : strlen(LEGACY_PAK_ZIP_EXT);
+
+	std::string name, version;
+	Util::optional<uint32_t> checksum;
+
+	ParsePakName(filename.begin(), filename.end() - suffixLen, name, version, checksum) || (isPakDir && checksum);
```
DolceTriade: It looks like you don' use this, so why not delete this?

src/common/FileSystem.cpp
```diff
@@ -2155,37 +2171,77 @@ bool IsInitialized()
...
+	if (type == pakType_t::PAK_ZIP) {
+		if (!isdigit(version.at(0))) {
+			fsLogs.Warn("Invalid pak version, must start with a digit: '%s'", fullPath);
 			return;
-		} else {
-			fsLogs.Notice("Loading legacy pak: %s", fullPath);
-			name = filename.substr(0, filename.size() - suffixLen);
-			version = "1";
 		}
+		fsLogs.Verbose("Adding pak: '%s'...", fullPath);
```
DolceTriade: This log seems redundant compared to the one in InternalLoadPak

src/common/FileSystem.cpp
```diff
@@ -2155,37 +2171,77 @@ bool IsInitialized()
...
+	Util::optional<uint32_t> checksum;
+
+	ParsePakName(filename.begin(), filename.end() - suffixLen, name, version, checksum) || (isPakDir && checksum);
 
+	name = filename.substr(0, filename.size() - suffixLen);
+	// prefer versioned dpk over legacy pk3 if pak name collides, uses the smallest version string available
+	// this version can't be found in standard dpk, so we can also uses this value to test later if it's a legacy pk3
+	version = "";
+
+	if (type == pakType_t::PAK_ZIP) {
```
DolceTriade: Log seems redundant as well.

src/common/FileSystem.cpp
```diff
@@ -2386,12 +2442,37 @@ bool ParsePakName(const char* begin, const char* end, std::string& name, std::st
 	return true;
 }
 
+std::string MakePakFileName(Str::StringRef name, Str::StringRef version, Util::optional<uint32_t> checksum)
+{
+	if ( !version.empty() ) {
+		// versioned dpk
+		return Str::Format("%s.%s", MakePakName(name, version, *checksum), "dpk");
```
DolceTriade: PAK_ZIP_EXT, not dpk

src/common/FileSystem.cpp
```diff
@@ -2386,12 +2442,37 @@ bool ParsePakName(const char* begin, const char* end, std::string& name, std::st
...
+	} else {
+		// legacy unversioned pk3
+		return Str::Format("%s.%s", MakePakName(name, version, *checksum), "pk3");
```
DolceTriade: LEGACY_PAK_ZIP_EXT, not pk3

src/common/FileSystem.cpp
illwieckz: They do not have the same purpose and are not printed for the same events. This one is printed when a pak is added to the known list of paks. It will be printed on all pak successfully found. The other one _("loading pak: ")_ is printed only when the pak is effectively loaded. The idea of this other log line is to allow user to check if his pak is found even if it's not loaded, using `+set logs.logLevel.fs debug`.

illwieckz:
Just a side note that comes to my mind, currently it's not possible to support some other legacy formats (like pk4) because the code does not store the extension internally. For example the http download code guesses the final path like that (pseudo-code):

```
if version.empty():
  get http://<path>/<name>_<version>.dpk
else:
  get http://<path>/<name>.pk3
```

Making Dæmon being able to handle more than one legacy file extension would require to change many more files and it's outside the scope of this PR.

src/common/FileSystem.cpp
DolceTriade: Ah, I see. Sounds good.

src/common/FileSystem.cpp
illwieckz: nice catch, removed

src/common/FileSystem.cpp
illwieckz: I modified that part like this:

```diff
        if (type == pakType_t::PAK_ZIP) {
+               if (version.at(0) > '9') {
+                       fsLogs.Warn("Pak version is very high: '%s'", fullPath);
+               }
                fsLogs.Verbose("Found pak: '%s'...", fullPath);
```

It enforces nothing, user is free to make mistakes, but he gets a warning saying a01 (for example) is a very high version number, something that he probably not expect from an alpha version number. Can be helpful.

illwieckz:
@DolceTriade I addressed all the issues you pointed out.

src/common/FileSystem.cpp
illwieckz: I modified these lines like this:

```diff
-		fsLogs.Verbose("Adding pak: '%s'...", fullPath);
+		fsLogs.Verbose("Found pak: '%s'...", fullPath);
```

This way the meaning of this log line looks more obvious.

Viech:
> enforce dpk version to start with a digit, dpkdir is allowed to start with an alphanumeric character, this way a mapper can be sure his source or test dpkdir always have greater version string than any released dpk using special version string like "src" or "test"

As I already pointed out on the [Unvanquished forums](http://forums.unvanquished.net/viewtopic.php?f=58&t=2122):
- Please don't distinguish `.pk3` and `.pk3dir` at all, despite for the obvious difference of one being a folder and the other being a packed folder. Apply the KISS principle.
- Please don't enfore a subset of version strings used by debian, instead allow the full range so we can just point people at the debian docs and don't need to explain our little differences.

Mappers should have the full freedom to make errors when they have no idea what a version string is, while people who do know should not have the extra complications.

Viech:
Also, is this PR implying that you simply discard the version of pk3s? That doesn't sound like sufficiently good legacy support, then, since people may want to have multiple pk3s and load only the newest. Ideally pk3s would even use the old ioquake3 version order.

illwieckz:
> is this PR implying that you simply discard the version of pk3s?

This PR fixes version sorting and dependency loading for non-legacy paks when legacy pak support is enabled. 

> Ideally pk3s would even use the old ioquake3 version order.

This PR tries to fix non-legacy paks, legacy pak support improvement is welcome but out of scope. By the way the changes introduced in this PR can be helpful for someone wanting to improve legacy pak support.

illwieckz:
> Please don't enforce a subset of version

There is no enforcement anymore. Are you reading the last iteration of this PR? There was many talks about that and this was rewritten. The last iteration was just a warning “the number is very high”, but since this causes too much problem I ditched that too. I thought it was a good time to suggest that other change since I believed the opinion was shared, which was not.

This PR is about giving to Dæmon the ability to discriminate legacy pak and non legacy pak when legacy paks support is enabled.

> (On a sidenote, I feel a01, b05, r23 are very good official release version strings for our .dkp.)

Well, that's another topic, but alpha/beta semantic and version ordering are two different things. Version ordering is about production iteration and sorting, alpha/beta semantic is about valuation.

So, that's my last comment about sorting since it's not the topic of this PR. I just removed that commit (even if it was enforcing nothing) because that's out of topic. We can reopen that “iteration _vs._ valuation” talk later, this is not about package management.

illwieckz:
dpk autodownloading works now (both udp and http one), but I noticed pk3 autodownloading does not work yet due to the way pak names are passed internally. Well this PR is about dpk, not pk3 so that's not really my concern but since I have some idea to also fix pk3 I will try this idea.

illwieckz:
So, last commit fixed legacy pak (pk3) download, since pk3 can't have checksum, the fix was simply to only validate checksum for dpk packages.

So, current code was tested in all these scenarii:

- pk3 udp download
- pk3 http download
- dpk udp download
- dpk http download
- dpk download with DEPS
- pk3 download with extrapaks
- dpk download with DEPS with extrapaks
- pk3dir loading
- dpkdir loading

If local dpk has wrong checksum, client re-download it again and uses the one with right checksum.
If local pk3 is different, engine loads it notwithstanding (that's expected legacy behavior).

src/common/FileSystem.cpp
```diff
@@ -2155,37 +2173,68 @@ bool IsInitialized()
...
 				AddPak(pakType_t::PAK_ZIP, Path::Build(subPath, filename), basePath);
 			} else if (Str::IsSuffix(PAK_DIR_EXT, filename)) {
 				AddPak(pakType_t::PAK_DIR, Path::Build(subPath, filename), basePath);
-			} else if (Str::IsSuffix("/", filename))
+			} else if (useLegacyPaks && Str::IsSuffix(LEGACY_PAK_ZIP_EXT, filename)) {
+				AddLegacyPak(pakType_t::PAK_ZIP, Path::Build(subPath, filename), basePath);
+			} else if (useLegacyPaks && Str::IsSuffix(LEGACY_PAK_DIR_EXT, filename)) {
+				AddLegacyPak(pakType_t::PAK_DIR, Path::Build(subPath, filename), basePath);
+			} else if (Str::IsSuffix("/", filename)) {
 				FindPaksInPath(basePath, Path::Build(subPath, filename));
```
Kangz: Here: https://github.com/DaemonEngine/Daemon/blob/master/src/common/Assert.h#L156 can you add a 
```
#define UNREACHABLE DAEMON_ASSERT_CALLSITE(false, , false, "Unreachable code hit.")
```
and add an else clause containing UNREACHABLE();

src/common/FileSystem.cpp
```diff
@@ -2351,16 +2400,27 @@ const PakInfo* FindPak(Str::StringRef name, Str::StringRef version, uint32_t che
...
 {
-	const char* nameStart = std::find(std::reverse_iterator<const char*>(end), std::reverse_iterator<const char*>(begin), '/').base();
-	if (nameStart != begin)
+	const char* nameStart;
+
+	if (Str::IsSuffix(PAK_ZIP_EXT, begin)) {
+		end -= strlen(PAK_ZIP_EXT);
+	} else if (Str::IsSuffix(PAK_DIR_EXT, begin)) {
+		end -= strlen(PAK_DIR_EXT);
+	}
```
Kangz: Same here, add an UNREACHABLE in an else clause.

src/common/FileSystem.cpp
illwieckz: I see, I thought about something like that but I was wondering how to implement it, thanks.

src/engine/qcommon/files.cpp
```diff
@@ -627,12 +627,18 @@ bool FS_LoadServerPaks(const char* paks, bool isDemo)
 	for (auto& x: args) {
 		std::string name, version;
 		Util::optional<uint32_t> checksum;
-		if (!FS::ParsePakName(x.data(), x.data() + x.size(), name, version, checksum)) {
-			Com_Error(errorParm_t::ERR_DROP, "Invalid pak reference from server: %s", x.c_str());
-		} else if (!checksum) {
-			if (isDemo || allowRemotePakDir.Get())
-				continue;
-			Com_Error(errorParm_t::ERR_DROP, "The server is configured to load game data from a directory which makes it incompatible with remote clients.");
+		if (Str::IsSuffix(FS::LEGACY_PAK_ZIP_EXT, x.data())) {
```
illwieckz: well, in fact not, but I can using it twice. I have some idea to improve things on that part.

illwieckz: For explanations, currently `FS::ParsePakName` is a function only used with non-legacy pak, that's why it does not make sense to move this legacy-pak related code inside `FS::ParsePakName`, but I'm looking for a way to use `FS::ParsePakName` for every pak.

illwieckz: I refactored that part, I was also able to remove the `AddLegacyPak` function I added before… :smiley:

src/common/FileSystem.cpp
illwieckz: done

src/common/FileSystem.cpp
illwieckz: done

src/common/FileSystem.cpp
```diff
@@ -1059,8 +1056,20 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
...
+		} else {
+			fsLogs.Notice("Loading legacy pak '%s'...", pak.path.c_str());
+		}
+	} else if (pak.type == pakType_t::PAK_DIR) {
+		if (!isLegacy) {
+			fsLogs.Notice("Loading pakdir '%s'...", pak.path.c_str());
+		} else {
+			fsLogs.Notice("Loading legacy pakdir '%s'...", pak.path.c_str());
+		}
+	}
```
illwieckz: @Kangz do you think I have to add an `else UNREACHABLE()` there?
In fact everywhere else the code is just doing `if PAK_DIR then else `(and does not test for `PAK_ZIP`).
I can also modify the other occurrence with explicit test for `PAK_ZIP` when not `PAK_DIR`, plus an `else UNREACHABLE`. What do you think about it?

src/engine/qcommon/files.cpp
DolceTriade: Smaller the footprint the better :D

src/common/FileSystem.cpp
Kangz: There are two choices, either test all cases in else-ifs and add an else clause with UNREACHABLE, or instead of finishing with else if (some condition) {code ...} do else {ASSERT(some condition), code...}. This way if a new case arises, we'll get a warning and know that things should be fixed.

illwieckz:
@Kangz I also modified some existing lines that were testing if a pak was a zip one by checking if it wasn't a dir one. Since you recommended to assert when a pak is not a zip and not a dir, I assume it's not good to trust a pak to be a zip just because it's not a dir. So, these lines now check if a pak is a zip by checking if it's a zip (bonjour Lapalisse).

illwieckz:
I squashed all the commits related to dpk support (because some were adding code the later removed). I left the one with UNREACHABLE asserts unsquashed since it acts as a very good standalone implementation example.

I verified this code works with both udp and http download with both dpk and pk3 archives.
I was able to load maps from both dpkdir and pk3dir.

Kangz:
Thanks @illwieckz, LGTM

illwieckz:
So, since this PR is now approved, is there an agenda to merge this PR?

Note: once this PR is merged, git Dæmon client will only be able to connect properly do git Dæmon server.
Probably not a problem since there is not so many people hacking Unvanquished code, and all of them can handle that. Perhaps it's time to raise protocol number, so we can host testing servers without having to care about stable clients mistakenly connecting to.

DolceTriade:
Yes, I would strongly prefer this to be merged into another branch. As soon as we prepare our next release, we will merge this in including the new paks based on your asset repo.

illwieckz:
So I just merged it in [for-0.51.0](https://github.com/DaemonEngine/Daemon/tree/for-0.51.0).

PR #15, opened 2017-04-04
Remove unused errorSample

DefaultUser:
AFAIS errorSample is unused, but still "sound/feedback/hit.wav" needs to be present. So removing it also removes the need for an external file.

### DIFF
--- a/src/engine/audio/Audio.cpp
+++ b/src/engine/audio/Audio.cpp
@@ -232,9 +232,16 @@ namespace Audio {
         BeginSampleRegistration();
     }
 
+    // return the handle to a Sample, -1 if not found
     sfxHandle_t RegisterSFX(Str::StringRef filename) {
         // TODO: what should we do if we aren't initialized?
-        return RegisterSample(filename)->GetHandle();
+        auto sample = RegisterSample(filename);
+        if (sample == nullptr)
+        {
+            Log::Warn ( "RegisterSFX: Sample %s not found", filename );
+            return -1;
+        }
+        return sample->GetHandle();
     }
 
     void EndRegistration() {
--- a/src/engine/audio/Sample.cpp
+++ b/src/engine/audio/Sample.cpp
@@ -70,8 +70,6 @@ namespace Audio {
 
     // Implementation of the sample storage
 
-    static const char errorSampleName[] = "sound/feedback/hit.wav";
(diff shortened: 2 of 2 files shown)


mbasaglia:
With this change, `sampleManager->defaultValue` will be `nullptr`, which will cause `GetResource` to return `nullptr` if the resource has not been found.
Does anyone know if this is going to cause issues?

Kangz:
It is going to cause issues indeed. If you want to get rid of the file you should instead generate it in code, for example with a 440Hz sinusoidal wave.

DefaultUser:
Generating this sample at runtime was my first idea, so I will do that soon

illwieckz:
Beware, that's the sound played when a sound is missing (like the "shader not found" for textures), the difference between fallback sound and fallback texture is you just have to look away to not see the fallback texture, but you can't avoid a fallback sound, especially when it is played in a loop, so the fallback sound must be hearable to detect the fault, but not too unbearable. As an example, you can delete the `sound/ui/heartbeat*.ogg` to hear the fallback sound looping in it's all glory. I'm not sure an infinite continuous 440hz sinusoidal wave is a good idea, it will be played infinitely if a looping ambient sound is missing for example, or some background music.

illwieckz:
By the way I agree with the idea of generating that sample at runtime.

mbasaglia:
A sine wave is that's unpleasant, anyway do we really need that? How about generating a silent sample and print an error on the console?

illwieckz:
That _sounds_ better to me.

DolceTriade:
I can record myself saying "error" and make that a required asset.

illwieckz:
what's the status on this?

illwieckz:
See #165 for an alternate implementation.

illwieckz:
The alternate implementation does not remove the need for an external file, but the best way to remove the need for an external file is to embed it in engine like glsl files are. But before doing that we had to make sure the builtin sound file is not unvanquished specific, hence the switch to the `null` sound path instead of `sound/feedback/hit`.

So we can close this PR. The PR to fullfill the need would not have to be a PR about sound file, but a PR about embedding `null.wav` and other files like the `lights/mkintsquarelight1a.png`/`lights/mkintsquarelight1b.png` and others from [`unvanquished_src.dpkdir`](http://github.com/UnvanquishedAssets/unvanquished_src.dpkdir).

PR #16, opened 2017-04-12
WIP: Remove hardcoded Unvanquished references

DefaultUser:
This adds support for other games by unhardcoding several references to
Unvanquished.
This switch automatically changes:
- default home path
- displayed product and version info
- default basepak
- masterserver, URI scheme and www_baseurl
- IRC channel and server

In my opinion, `PRODUCT_VERSION`/`getGameinfo()->version` should be defined by the game's Menu code, but for now I will leave it in.


mbasaglia:
Going to link this here since this seems related: https://github.com/Unvanquished/Unvanquished/issues/817

src/common/Defs.h
```diff
@@ -61,9 +67,38 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+	game(UNV)
+
+#define game(x) GAME_##x
+
+typedef enum { GAMES } Game_e;
+
+#undef game
+#define game(x) { PRODUCT_NAME_##x, PRODUCT_NAME_LOWER_##x, PRODUCT_VERSION_##x, DEFAULT_BASE_PAK_##x, GAMENAME_STRING_##x, GAMENAME_FOR_MASTER_##x, MASTER_SERVER_NAME_##x, IRC_SERVER_##x, IRC_CHANNEL_##x, WWW_BASEURL_##x, UNNAMED_SERVER_##x }
+
+const gameinfo_t gamesinfo [ ] = { GAMES };
```
mbasaglia: I really don't like the macro-based approach and the overall C-ness of gameinfo_t, I think it could lead to nasty issues and make it harder to extend in the future

src/common/System.h
```diff
@@ -31,6 +31,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef COMMON_SYSTEM_H_
 #define COMMON_SYSTEM_H_
 
+extern const gameinfo_t* current_game;
```
mbasaglia: Why do you use current_game directly if you have a getter? I think this should be an implementation detal.

src/common/Defs.h
```diff
@@ -31,28 +31,34 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+#define MASTER_SERVER_NAME_UNV  "master.unvanquished.net"
 
-#define MASTER_SERVER_NAME  "master.unvanquished.net"
+#define IRC_SERVER_UNV          "irc.freenode.org"
+#define IRC_CHANNEL_UNV         "unv-lobby"
 
-#define IRC_SERVER          "irc.freenode.org"
-#define IRC_CHANNEL         "unv-lobby"
+#define WWW_BASEURL_UNV         "dl.unvanquished.net/pkg"
```
mbasaglia: Some of these could even become cvars.

src/engine/client/cl_irc.cpp
```diff
@@ -1929,7 +1929,7 @@ static ircCmd_t CTCP_Version( bool is_channel, const char * )
 		return ircCmd_t::SUCCESS;
 	}
 
-	return IRC_Send( "NOTICE %s :\001VERSION Daemon IRC client — v\n" Q3_VERSION "\001", IRC_String( pfx_nickOrServer ) );
+	return IRC_Send( "NOTICE %s :\001VERSION Daemon IRC client — v\n" Q3_ENGINE "\001", IRC_String( pfx_nickOrServer ) );
```
mbasaglia: Does anyone know if the IRC code is supposed to work at all?

src/engine/qcommon/files.cpp
```diff
@@ -31,9 +31,19 @@ along with Daemon Source Code.  If not, see <http://www.gnu.org/licenses/>.
 const char TEMP_SUFFIX[] = ".tmp";
 
 // Cvars to select the base and extra packages to use
-static Cvar::Cvar<std::string> fs_basepak("fs_basepak", "base pak to load", 0, DEFAULT_BASE_PAK);
+static Cvar::Cvar<std::string>* fs_basepak;
```
mbasaglia: Why does this need to be a pointer? Can't you just change its value instead of allocating a new one?

src/engine/qcommon/qcommon.h
```diff
@@ -285,7 +285,7 @@ You or the server may be running older versions of the game."
 
 #define PROTOCOL_VERSION       86
 
-#define URI_SCHEME             GAMENAME_STRING "://"
+#define URI_SCHEME             va("%s://", current_game->gamename_string)
```
mbasaglia: I think this is one of the strings that should be engine-specific, not game-specific.

src/engine/server/sv_main.cpp
```diff
@@ -399,8 +399,8 @@ but not on every player enter or exit.
 ================
 */
 static const int HEARTBEAT_MSEC = (300 * 1000);
-#define HEARTBEAT_GAME PRODUCT_NAME
-#define HEARTBEAT_DEAD PRODUCT_NAME "-dead"
+#define HEARTBEAT_GAME getGameinfo()->name
+#define HEARTBEAT_DEAD va("%s-dead", HEARTBEAT_GAME)
```
mbasaglia: I think we can do without macro declarations for these.

src/engine/qcommon/files.cpp
mbasaglia: Looks like this is also what breaks the build.

DolceTriade:
I don't think we need to have this be a runtime thing. I don't think using the same game engine binary for different games should really be a goal. Each game should set overriding defines and have these compiled into their binary. I think Unvanquished made a mistake when we released our game with a "daemon" binary instead of an "unvanquished" binary.

mbasaglia:
I think we should ensure the client engine is able to work regardless of the game it came from.

src/common/Defs.h
illwieckz: cvars for this kind of stuff sounds good

src/engine/qcommon/qcommon.h
illwieckz: I agree, that's a protocol thing, so that's related to the engine. Perhaps unv:// is not enough reusable for a generic Dæmon thing, any suggest?

mbasaglia: dmn:// ?

illwieckz:
That's not a problem to just rename `daemon` to `unvanquished` for convenience at build time, even if the engine is generic enough to run another game.

src/engine/qcommon/qcommon.h
illwieckz: Does the protocol have to optionally announce the game?
like¹ `dgp://gg.illwieckz.net:27960/unvanquished`

That `/unvanquished` would be optional, but if you have multiple games running dæmon engine, your game manager can use the binary (xonotic for `/xonotic`, unvanquished for `/unvanquished`, would fill the `-game` switch), if game not installed but daemon from another game installed, run it with that, if that optional `/game` is not there, same. That would help Dæmon based games with some outdated engine to not break. That would be very useful for games that enables `fs_legacypaks` by default (like xonotic), ensuring a game client enabling that legacy stuff is started before joining server.

¹ `dgp://` for _Daemon Game Protocol_ for example or another variation with _Connect Protocol_ or words like that, meaning it's a protocol to make game clients connecting a server, following _Dæmon guidelines_, the same way other engine can reuse the _DPK_ guidelines because they are tired of PK3 design flaws if they want. If defined well, other games would be able to reuse the url format even if they do not rely on Dæmon engine. Well, that's a new topic we can have in forum so I will probably not elaborate more about that topic on that thread. :grin:

Kangz:
It is a worthy goal to have a single Daemon binary be able to run all games seamlessly and not have any hardcoded references to Unv. In most cases replacing "Unvanquished" with "Daemon" should be enough.

However I don't think having a single "daemon.exe" for all games will be convenient in the short-term. On example is that all of Unv's packages are versioned and that we'll want to force the usage of .dpk over .pk3 (see #14). On the contrary Xonotic will want support for legacy packages to stay compatible with their maps and mods. I don't think it makes sense to be able to switch between the two modes at runtime as it would lower security if somehow servers can expose game mods that dynamically switch on support for legacy packages.

src/engine/qcommon/qcommon.h
mbasaglia: Servers would use different port numbers for different instances, and the game logic should be discovered during connection, so I don't think we need a path component to the URI.

mbasaglia: Also, different game servers would be listed separately by the master.

mbasaglia:
That can be enforced by cvars on the client side.

illwieckz:
not related to that current topic but:

illwieckz:
not related to that current topic but:

> On example is that all of Unv's packages are versioned and that we'll want to force the usage of .dpk over .pk3 (see #14). On the contrary Xonotic will want support for legacy packages to stay compatible with their maps

All the work done in #14 is thought to allow xonotic being able to use dpk for official assets one day (with version, deps, checksum and all awesome features), loading legacy contrib map as compat ones. I don't think it makes sense for a game using Dæmon engine to rely on broken pk3 vfs for their own official assets they can repackage to get the best from the engine.

To give another example with a slightly different use case, games with open source engine and legacy official closed source assets from company (like wolf:et) can't repackage the `pak*.pk3`, but they can put these `pak0.pk3`, `pak1.pk3` as DEPS in `wolfet_4.3.dpk` (containing the nacl game code setting `fs_legacypaks` to `true`). I think it's not wrong if `dpk` is the default assumed format, it would be silly to package in legacy pk3 container a game code you have right to develop and package, by the way the code #14 does not prevent using pk3 for the game pak, but since Dæmon does not load legacy paks by default, the best way to provide seamless user experience is to put a game code in a dpk that sets `fs_legacypaks` to true then insta reload vfs to load missing stuff from legacy pk3.

src/common/Defs.h
Viech: I feel an `UNV`-suffixed definition or Cvar is still a branding that shouldn't be present in a game-agnostic engine.

DefaultUser:
I removed the "-game" switch again, instead the gameinfo will be read from a file called "gameinfo.cfg" located in the libpath at runtime.

This does break compiling the unvanquished game logic since the new "Gameinfo" class depends on file access for "libpath". This can be fixed by redefining "Q3_VERSION" in the game logic.

src/common/Defs.h
mbasaglia: The cvar doesn't need the unv suffix

DolceTriade:
Sorry if I'm being a pain, but I'm not sure this is any better than having the information compiled into the binary. The libpath dependency makes it so that you cannot use the same binary with different games without changing gameino.cfg. It doesn't seem there is a way to change this dynamically, and even if there were, i'm not sure there is a strong benefit to this.

Still strongly in favor of simply compiling all of this information into the binary to decrease complexity.

DefaultUser:
Actually you can use the same binary for different games as "libpath" can be set in the commandline arguments.

Kangz:
I'm still not happy with the design of this and agree with @DolceTriade. There are also a whole lot of coding style / nits I want to make but will wait for the final design before doing so to avoid churn. Please refrain from merging before that "style" review.

illwieckz:
I don't see the point of having a named engine or having the engine knowing how to launch the game. Isn't it the launcher job? The Quake Live engine is named quakezero internally, the Natural Selection 2 engine is named spark, no one cares. The engine is just another component of the game.

Having the engine named like the game and having the engine embedding hardcoded game reference is only needed when then engine is also the launcher. We [already have a launcher](https://github.com/Unvanquished/updater/). If one day Unvanquished is distributed by some game-directory like Steam, users will not care about the engine name and people will not need things to be hardcoded.

To host a GoldenEye:Source server I just do: `srcds_i486 -game ../gesource`, using an unmodified source engine, the `-game` option just need a path, like our `-libpath` can do. @DefaultUser 's suggest make Dæmon having the same behavior, which is very common and expected by people hosting games. People playing games don't care, they use the launcher after all.

src/engine/qcommon/qcommon.h
```diff
@@ -285,7 +288,7 @@ You or the server may be running older versions of the game."
 
 #define PROTOCOL_VERSION       86
 
-#define URI_SCHEME             GAMENAME_STRING "://"
+#define URI_SCHEME             "dgp://"
```
DefaultUser: Looking at this I am no longer sure that the URI_SCHEME can be engine specific. How should the OS know what game to start when you click a link with an engine specific URI_SCHEME?

mbasaglia: External tools would only need to call the engine, the game for the server should be part of the out of band protocol, I think getstatus shows it.

illwieckz: @mbasaglia with currently file layout it works only if you mix assets from different games within same directories, which is ok for small mods but not for complete different games (I don't want to share the same `pkg/` dir for both xonotic and unvanquished games for example) so the engine must known the game _before_ joining the server :confused: or the file layout must be completely redesigned.

illwieckz: By the way, it's not engine's task to redirect url calls to engine, not a problem to have both unvanquished and xonotic reads the url scheme if the desktop knows how to redirect these url to xonotic or unvanquished (I don't have magical solution, though). The problem of this kind of url stuff is it's more the job for a launcher (like the unvanquished updater), or a job for a game library manager like Steam or Lutris. If we want a common url for multiple games, it's something to design outside of Unvanquished : it's designing a standard with protocols, formats and common tools others can reuse, praising for adoption. :neutral_face:

illwieckz:
> In my opinion, PRODUCT_VERSION/getGameinfo()->version should be defined by the game's Menu code, but for now I will leave it in.

I guess there must be two versions: the engine version (built-in), and the game version (defined by game code).

lavacano201014:
If we were going to insist on daemon somehow being able to figure out what game you meant, we could combine it with the idea of naming the binary after the game - if `-libpath` is not set at launch, try the binary's name first, then try `pkg`.

e.g. an `unvanquished` binary with no `-libpath` option would first try the `unvanquished` path, then fall back to `pkg` if that doesn't work.

illwieckz:
_wat?_

MarioSMB:
Any news on this @illwieckz @Kangz?

src/engine/qcommon/qcommon.h
illwieckz: We can keep `unv://` for the moment with a glorious `FIXME:`, the _protocol_ is by definition something about _others_, so we have to take time to think properly about it, and it must not be a blocker for other things.

slipher:
Come over to #108 for more in-depth discussion of these issues.

src/engine/client/cl_irc.cpp
illwieckz: IRC is now nuked: https://github.com/DaemonEngine/Daemon/commit/ab3f6591e3c9c42b0447c100427d28aba32e87c6

illwieckz:
@DefaultUser, can you rebase this on master? I would like to experiment with it. :-)

DefaultUser:
I cherry-picked the relevant commits and force-pushed the result

src/engine/qcommon/qcommon.h
illwieckz: @DefaultUser as it seems too early to think about protocol uri scheme and I don't want this to block the PR, can you revert this line to `unv://` and add a `FIXME:` comment?

src/engine/qcommon/gameinfo.h
```diff
@@ -0,0 +1,63 @@
...
+===========================================================================
+*/
+
+#ifndef GAMEINFO_H_
+#define GAMEINFO_H_
+
+#include <iostream>
+
+
+class Gameinfo
```
illwieckz: I know it would probably need more work, but is it possible to leverage existing cfg file code?
also some of the variables already exists as cvar we can set as command line option, maybe we can make the engine reading those cvars first and, if command line option is set, override.

DefaultUser: I've looked into the existing cfg file parsing code again. I don't think we can leverage this code as it simply treats every line as a console command. The gameinfo.cfg doesn't contain commands and I don't think that NAME, VERSION, DEFAULT_BASEPAK or GAMENAME_STRING should be cvars. Adding commands just for those is way to complicated for what I try to achieve.
I renamed that file to gameinfo.conf to make it obvious that it is not a cfg file that contains commands and cvars.

About the cvars: MASTERSERVER<num> and WWW_BASEURL could probably dropped completely from this config file since they are only default values.

illwieckz: OK, then it looks OK to me as reusing the cfg parsing code looks to be the over-engineered way.
Also, by not using cfg parsing we have less thing to think about security: we are sure non-cvar variable cannot be abused with cvar code.

gameinfo.conf.template
```diff
@@ -0,0 +1,7 @@
+NAME               GameName
+VERSION            0.0
+DEFAULT_BASEPAK    mybasepak
```
illwieckz: why not just BASEPAK? We already know this is the default since it's stored in that file.

src/engine/qcommon/qcommon.h
illwieckz: After deep thoughts (I'm not sure) and long time passed (for sure), it now feels stupid to me to have the URI_SCHEME to be engine specific. What we need in engine is a way to grok the URI_SCHEME the game developer wants it to grok. It's up to the game developer to properly configure player system to redirect given URI_SCHEME to the engine with proper options (libpath containing the config file telling the URI_SCHEME proto).

So it now looks very OK to me to set that value in the config file.

Maybe name that constant `URI_PROTO` to be more explicit than `GAMENAME_STRING`.
Years later I completely forgotten what `GAMENAME_STRING` meant and I had to read the code to understand it was the URI proto string.

gameinfo.conf.template
illwieckz: After deep thoughts, this may be called `BASENAME`, there may be more than one usage for a basename.

We can also use it to compute `~/.local/share/${BASENAME}` while other OS may do something with `NAME` instead if it's more common on such OS to do that (`My Games/${GAME}`).

I don't think it's a bad idea to enforce the basename is also the basepak name.

It may also make easier to reimplement per-directory full-conversion downloadable mod like quake3, by using this BASENAME (`unvanquished` or `quake6`) string instead of a shortcut (`unv` or `baseq6`).

gameinfo.conf.template
```diff
@@ -0,0 +1,7 @@
+NAME               GameName
```
illwieckz: I don't think anything prevent to use a real-life example like Unvanquished default strings in the template.

illwieckz: Same for the default values in the code if undefined.

Note that in the future we would be able to make the engine being able to load some built-in asset and display an error (like Darkplaces does) so the template and the hardcoded version would reference that dummy game that does nothing at all.

illwieckz:
This is an idea I just got, let's imagine those two `.cfg` file (yes, `.cfg`):

```sh
builtin in_name "Unvanquished"
builtin in_baseName "unvanquished"
builtin in_basePak "unvanquished"
builtin in_version "0.51.2"
builtin in_master1 "master.unvanquished.net"
builtin in_master2 "master2.unvanquished.net"
builtin in_wwwBaseURL "dl.unvanquished.net/pkg"
builtin in_protoUrl "unv"
```

```sh
builtin in_name "Xonotuc"
builtin in_baseName "xonotic"
builtin in_basePak "xonotic"
builtin in_version "0.9.0"
builtin in_master1 "master.xonotic.org"
builtin in_master2 "master2.xonotic.org"
builtin in_wwwBaseURL "dl.xonotic.org/pkg"
builtin in_protoUrl "xon"
set fs_legacyPaks on
set fs_maxSymlinkDepth 1
set r_dpMaterial on
set r_dpBlend on
```

Your non-cfg parser would read all lines starting with `^builtin` and only them, this is only written to mimic the `.cfg` file syntax for cosmetic purpose.

Then, we would hack the command line parser to read every line that does not start with a `builtin` or a `//`, and pass them to the command line interpreter as `-line`, before reading the true command line options, so they can be overridden.

Note that some of these builtins may use normal cvars, like the `sv_basePak` one. But by doing a `in_basePak` builtin we configure the one to use as fallback when the one set in cvar is not found.
We may even just not implement the ability to have a default basepak that is not the basename, and use the basename for the basepak in all case. I just kept that `in_basePak` for the example of a builtin that is not a cvar, even if a cvar exists.

I imagined that because Xonotic would win a lot by being able to set compatibility cvars in that file, about those, an user may choose to disable them without breaking the game itself, but they must be set by default.

illwieckz:
Also, I don't think it's a good idea to set the version in that file. Either the engine has a version, whatever the game, and it's better to hardcode it, either the game has a version, end the game can set it without recompiling the engine.

So, basically, I think that is good:

```sh
builtin in_name "Unvanquished"
builtin in_baseName "unvanquished"
builtin in_master1 "master.unvanquished.net"
builtin in_master2 "master2.unvanquished.net"
builtin in_wwwBaseURL "dl.unvanquished.net/pkg"
builtin in_protoUrl "unv"
```

```sh
builtin in_name "Xonotuc"
builtin in_baseName "xonotic"
builtin in_master1 "master.xonotic.org"
builtin in_master2 "master2.xonotic.org"
builtin in_wwwBaseURL "dl.xonotic.org/pkg"
builtin in_protoUrl "xon"
set fs_legacyPaks on
set fs_maxSymlinkDepth 1
set r_dpMaterial on
set r_dpBlend on
```

illwieckz:
Superseded by #588.

PR #17, opened 2017-04-29
FindLua: Also prefer to find ${PREFIX}/include/lua before versioned lua

DolceTriade:
This fixes a bug where the game finds liblua.so, but picks include/lua53
as the include path, causing a mismatch.


Kangz:
LGTM

DolceTriade:
Note that this does not fix the underlying issue that the library search and include search algorithms may select incompatible libraries and includes, but I'm not sure there is a way around that. This minimizes the likelihood of that happening. If someone has a weird setup, they can set paths manually.

PR #19, opened 2017-05-30
[Review only] Redesign key bind handling, to support Unicode and other things

slipher:
In this proposed system, there are 3 types of bindable keys: scancode, keynum, and character. Scancode binds are tied to a particular spot on the keyboard, independent of keyboard layout. Keynum binds are used for keys that don't correspond to characters and pseudo-keys (e.g. ALT or MOUSE1) and work the same as before. Character binds are tied to the character that would be entered by a key (when it is typed without modifiers) and would change locations with a keyboard language change. Character binds work the same as binding a letter or number in the old system, but are unreliable for non-ASCII characters (see below).

One goal of this rewrite is to make it possible to bind Unicode characters (the old code only supports ASCII). This has to rely on SDL reporting the correct character for a key. Unfortunately, the library's behavior is inconsistent across platforms. Sometimes it reports the ASCII character that a QWERTY keyboard has at that location, instead of the correct Unicode character. On one system I tested, with the French layout, digits were reported for any decimal digit keys, while on another the characters actually generated by an unmodified keystroke were reported. But both systems reported the right. non-ASCII Unicode characters for other locations. And with a Russian layout, one system reported all Russian characters, while another reported none of them. Due to these issues, scancodes are the preferred way to record binds in the new design. The interface will make it "difficult" to create character-based binds by accident. 

`/bind hw:*char*` makes a scancode bind ("hardware"-based) to a key whose physical location is specified by QWERTY position.
`/bind 0x*nnn*` makes a scancode bind to a scancode specified by hexadecimal number.
`/bind char:*char*` makes a character-based bind.
`/bind SHIFT` and friends work as usual.
`/bind *char*` tries to figure what key is specified by the character, and then creates a scancode-based bind for it. First it uses an SDL lookup function to get the scancode from a character. If that fails, but the character is a member of the QWERTY layout, then it creates the binding based on QWERTY. If neither of those works then an error is reported.

The GUI binding interface details are not yet worked out, but it should also favor creating scancode-based binds.

### Other goals that this design tries to achieve
- Ensure that a set of bindings is portable between computers.
- Automatically convert existing keybindings.cfg files to the new format the first time they are read.
- Make the generated bindings file somewhat self-explanatory.
- If someone uses the GUI-based bindings for a while and then switches to a text-based method, don't have everything be screwed up.
- Optimize for cases where a large percentage of the keys used are in the same positions as QWERTY.
- Make it possible to bind keys that are not listed in the key enumerations of Daemon or SDL.

### Drawbacks
- Added complexity.
- The confusing possibility of having two actions bound to one key (if both scancode and char binds are made). As I write this, I am realizing that char binds may be fairly pointless; maybe they should be removed. Does anyone think there are good use cases for it?
- When using a bind-related command, the canonical string form of a key can confusingly look different than the way the user entered it. E.g. `/bind a` -> `hw:q is not bound.` Although, command line users are already somewhat "advanced".
- Non-portability of SDL keyboard handling exposed (but this is equally true of the old code).
    
### Testing
There is a lot of code cleanup yet to be done, but the patch is ready to be tested in its current state. Note that if you have any binds you care about, you should make a backup copy before testing (or use a different homepath) as the new format is not backwards-compatible.


src/engine/client/cl_cgame.cpp
```diff
@@ -633,6 +633,8 @@ static int LAN_ServerIsVisible( int source, int n )
  */
 void Key_GetBindingBuf( int keynum, int team, char *buf, int buflen )
 {
+    *buf = 0;
+#if 0
```
DolceTriade: Is this section TODO?

slipher: Right.

slipher:
For reviewing it could be better to look at [this commit](https://github.com/DaemonEngine/Daemon/pull/19/commits/f8c11cc53e8a4756e38a6bf566b0e19b513dc8a5) since there is some noise in the diff, and this is the only commit that does anything.

src/engine/client/cl_input.cpp
```diff
@@ -1159,7 +1164,7 @@ void IN_KeysUp_f()
 		{
 			if ( first )
 			{
-				Cmd::ExecuteCommand(va("setkeydata %d %d %u", check, key + 1, time));
+				Cmd::ExecuteCommand(va("setkeydata %d %s %u", check, Key_KeynumToString(key), time));
```
DolceTriade: Key_KeyToString

src/engine/client/cl_input.cpp
```diff
@@ -1145,12 +1147,15 @@ void IN_BuiltinButtonCommand()
...
 {
 	unsigned int check;
-	int key, time;
+	int time;
 	int i;
 	bool first = true;
 
 	check = atoi( Cmd_Argv( 1 ) );
-	key   = atoi( Cmd_Argv( 2 ) );
+	Keyboard::Key key = Key_StringToKeynum( Cmd_Argv( 2 ) );
```
DolceTriade: Key_StringToKey

src/engine/client/client.h
```diff
@@ -544,15 +544,15 @@ void CL_Record(std::string demo_name);
 //
 // cl_keys (for input usage)
```
DolceTriade: Let's remove the Key_* prefix and put all new keyboard code inside the Keyboard namespace.

src/engine/client/client.h
```diff
@@ -601,8 +601,8 @@ void IN_PrepareKeyUp();
 //----(SA)
 
 float    CL_KeyState( kbutton_t *key );
-int      Key_StringToKeynum( const char *str );
-const char *Key_KeynumToString( int keynum );
+Keyboard::Key Key_StringToKeynum( const char *str );
```
DolceTriade: Str::StrRef instead of const char str.

src/engine/client/client.h
```diff
@@ -601,8 +601,8 @@ void IN_PrepareKeyUp();
...
+const char *Key_KeynumToString( Keyboard::Key key );
```
DolceTriade: Return std::string

src/engine/client/key_identification.cpp
```diff
@@ -0,0 +1,504 @@
...
+        return ch - '1' + Util::ordinal<SDL_Scancode>(SDL_SCANCODE_1);
+    }
+    switch (ch) {
+        case ' ': return SDL_SCANCODE_SPACE;
+        case '\'': return SDL_SCANCODE_APOSTROPHE;
+        case ',': return SDL_SCANCODE_COMMA;
+        case '.': return SDL_SCANCODE_PERIOD;
+        case '/': return SDL_SCANCODE_SLASH;
+        case '-': return SDL_SCANCODE_MINUS;
+        case '0': return SDL_SCANCODE_0;
```
DolceTriade: 0 isn't next to 1?

src/engine/client/key_identification.cpp
```diff
@@ -0,0 +1,504 @@
...
+        case SDL_SCANCODE_RIGHTBRACKET: return ']';
+        case SDL_SCANCODE_GRAVE: return '`';
+
+        default: return '\0';
+    }
+
+}
+
+
+struct keyname_t
```
DolceTriade: Let's abandon the C convention for structs for new structs.

src/engine/client/key_identification.cpp
```diff
@@ -0,0 +1,504 @@
...
+{
+	const char *name;
+	Keyboard::Key keynum;
+};
+
+// names not in this list can either be lowercase ascii, or '0xnn' hex sequences
+static const keyname_t keynames[] =
```
DolceTriade: Make this a std::unordered_map?

src/engine/client/key_identification.cpp
```diff
@@ -0,0 +1,504 @@
...
+		return Key::NONE;
+	}
+
+	if ( int c = ParseCharacter( str ) )
+	{
+		return KeyFromUnprefixedCharacter( c );
+	}
+
+	// check for hex code
+	if ( str[0] == '0' && str[1] == 'x' )
```
DolceTriade: Str::IsPrefix()

src/engine/client/key_identification.cpp
```diff
@@ -0,0 +1,504 @@
...
+	{
+		int n = Com_HexStrToInt( str );
```
DolceTriade: Not really a comment on your commit, but the code for Com_HexStrToInt could be replaced entirely by strtol(char, end, 16)...

src/engine/client/key_identification.cpp
```diff
@@ -0,0 +1,504 @@
...
+		return Key::FromScancode( n );
+	}
+
+    // Physical key by QWERTY location, from ascii char
+    int prefixLen = strlen(SCANCODE_ASCII_BIND_PREFIX);
```
DolceTriade: Save this outside the function so you don't have to run strlen every call.

src/engine/client/key_identification.cpp
```diff
@@ -0,0 +1,504 @@
...
+    if ( !Q_strnicmp( SCANCODE_ASCII_BIND_PREFIX, str, prefixLen ) && strlen( str ) == prefixLen + 1 )
```
DolceTriade: Str::IsPrefix?

src/engine/client/key_identification.cpp
```diff
@@ -0,0 +1,504 @@
...
+Key_KeynumToString
+
+Returns a string (either a single ascii char, a K_* name, or a 0x11 hex string) for the
+given keynum.
+===================
+*/
+const char *Key_KeynumToString( Key key )
+{
+	const keyname_t *kn;
+	static std::string tinystr;
```
DolceTriade: I think making a copy is fine.

src/engine/client/key_identification.cpp
```diff
@@ -0,0 +1,504 @@
...
+			return kn->name;
+		}
+	}
+
+	if ( key.kind() == Key::Kind::UNICODE_CHAR )
+	{
+		tinystr = CHARACTER_BIND_PREFIX;
+        tinystr += Q_UTF8_Encode( key.AsCharacter() );
+        tinystr = Cmd::Escape( tinystr );
+        return tinystr.data();
```
DolceTriade: Any reason you use data() instead of c_str()? Since it is being used as a string, using c_str() makes the intention more clear.

src/engine/client/keys.h
```diff
@@ -32,40 +32,141 @@ Maryland 20850 USA.
 ===========================================================================
 */
 
+#ifndef COMMON_KEYS_H_
+#define COMMON_KEYS_H_
```
DolceTriade: CLIENT_KEYS_H_ or ENGINE_CLIENT_KEYS_H_

src/engine/client/keys.h
```diff
@@ -32,40 +32,141 @@ Maryland 20850 USA.
...
+        size_t operator()(Key k) const {
+            return std::hash<uint64_t>()(
+                Util::ordinal<Kind, uint64_t>(k.kind_) << 32 | uint64_t(k.id_));
+        }
+    };
+
+    // TODO(slipher): remove need for all of the following
+    int AsLegacyInt() const {
+        if (kind_ == Kind::CONSOLE)
+            return (int) K_CONSOLE;
```
DolceTriade: C++ cast

src/engine/sys/sdl_input.cpp
```diff
@@ -94,50 +90,19 @@ static void IN_PrintKey( const SDL_Keysym *keysym, keyNum_t key, bool down )
...
 /*
 ===============
 IN_IsConsoleKey
 
-TODO: If the SDL_Scancode situation improves, use it instead of
-      both of these methods
 ===============
 */
-static bool IN_IsConsoleKey( keyNum_t key, const unsigned char character )
+static bool IN_IsConsoleKey( Keyboard::Key key )
```
DolceTriade: const Keyboard::Key&?

src/engine/sys/sdl_input.cpp
```diff
@@ -765,6 +615,18 @@ static void IN_ShutdownJoystick()
 	SDL_QuitSubSystem( SDL_INIT_JOYSTICK );
 }
 
+static void QueueKeyEvent(Keyboard::Key key, bool down)
```
DolceTriade: const Keyboard::Key&

src/engine/client/cl_keys.cpp
```diff
@@ -599,7 +292,7 @@ void Console_Key( int key )
 	}
 
 	// ctrl-L clears screen
-	if (key == 'l' && keys[ K_CTRL ].down) {
+	if (key == Key::FromCharacter('l') && keys[ K_CTRL ].down) {
```
DolceTriade: Would be nice to cache these between frames...

src/engine/client/cl_keys.cpp
```diff
@@ -588,7 +281,7 @@ Console_Key
 Handles history and console scrollback for the ingame console
 ====================
 */
-void Console_Key( int key )
+void Console_Key( Key key )
```
DolceTriade: const Key&

src/engine/sys/sdl_input.cpp
```diff
@@ -765,6 +615,18 @@ static void IN_ShutdownJoystick()
...
+{
+    const int UNUSED = 0;
+    using Keyboard::Key;
+    if (!key.IsValid()) {
+        return;
+    }
+    Key* keyPtr = static_cast<Key*>(Z_Malloc(sizeof(Key)));
```
DolceTriade: Note where this is freed.

src/engine/client/keys.h
```diff
@@ -32,40 +32,141 @@ Maryland 20850 USA.
...
+    Kind kind_;
+    int id_;
+    constexpr Key(Kind kind, int id): kind_(kind), id_(id) { }
+
+public:
+    static const Key NONE;
+    static const Key CONSOLE;
+
+    Key(): Key(Kind::INVALID, 0) { }
+    Key(keyNum_t key) {
```
DolceTriade: explicit Key(keyNum_t key)

I know it's syntactic sugar to convert a keyNum_t into a Key and you use it heavily for SDL, but it makes the code unclear. You can create the Key in QueueEvent and still keep the same syntactic sugar.

src/engine/client/cl_keys.cpp
slipher: What do you mean by this comment?

DolceTriade: I was thinking that it sucked that we have to call this function many times per frame for a constant value. Just looking at the definition again, it seems Key::FromCharacter is pretty simple. Can it be a constexpr?

slipher: It could be a constexpr, although I'd have to rewrite it in a less readable way to make the function consist of a single return statement. I expect the compiler to optimize it out regardless (see https://godbolt.org/g/yp5sua - it turns into 'mov rax 463856467969').

DolceTriade: Eh, this is fine. probably me prematurely optimizing.

PR #20, opened 2017-07-04
Linux: follow XDG guidelines

illwieckz:
- use ~/.local/share/unvanquished instead of ~/.unvanquished
- move legacy path if exists
- fix Unvanquished/Unvanquished#730

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -2080,33 +2080,41 @@ std::string DefaultHomePath()
 
 	xdgHomePath = Path::Build(xdgDataHome, PRODUCT_NAME_LOWER);
 
-	// move legacy ~/.unvanquished directory to XDG ~/.local/share/unvanquished
-	// if ~/.unvanquished exists but ~/.local/share/unvanquished does not exist
-	// unless ~/.unvanquished is a symlink (symlink can be relative, behavior is unpredictable)
 	if (lstat(legacyHomePath.c_str(), &stl) == 0) {
-		if (S_ISLNK(stl.st_mode) && stat(xdgHomePath.c_str(), &stx) != 0) {
-			Sys::Error("Legacy home path %s is a symlink and symlink can be relative, will not rename to %s"
-						", do it yourself!", legacyHomePath, xdgHomePath);
-		}
-
-		if (S_ISDIR(stl.st_mode)) {
+		if (S_ISDIR(stl.st_mode) || S_ISLNK(stl.st_mode)) {
 			if (stat(xdgHomePath.c_str(), &stx) != 0) {
 				std::error_code err;
 
 				RawPath::CreatePathTo(xdgDataHome, err);
 
 				if (err) {
-					Sys::Error("Could not create %s: %s", xdgDataHome, err.message());
+					Sys::Error("Could not create XDG data directory %s: %s", xdgDataHome, err.message());
 				}
 
-				fsLogs.Warn("Renaming legacy home path %s to %s", legacyHomePath, xdgHomePath);
-				RawPath::MoveFile(xdgHomePath, legacyHomePath, err);
+				if (S_ISLNK(stl.st_mode)) {
+					int ret;
+					int fd = open(xdgDataHome.c_str(), O_DIRECTORY);
+
+					if (fd == -1) {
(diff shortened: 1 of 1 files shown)


illwieckz:
There is just a minor issue, the early warning is not colored:
```
^3Warn: [FS] Renaming legacy home path /home/illwieckz/.unvanquished to /home/illwieckz/.local/share/unvanquished
```

src/common/FileSystem.cpp
```diff
@@ -2067,7 +2067,51 @@ std::string DefaultHomePath()
 #ifdef __APPLE__
 	return std::string(home) + "/Library/Application Support/" PRODUCT_NAME;
 #else
-	return std::string(home) + "/." PRODUCT_NAME_LOWER;
+	struct stat stl, stx;
+
+	std::string legacyHomePath = std::string(home) + "/." PRODUCT_NAME_LOWER;
```
DolceTriade: Use FS::Path::Build() to build paths.

illwieckz:
Note that if someone symlink `~/.unvanquished` to `~/.local/share/unvanquished` (or the opposite), this one can load both current game and `for-0.51.0` game. On 0.51.0 release, player that only have `~/.unvanquished` will get his `~/.unvanquished` renamed to `~/.local/share/unvanquished`.

src/common/FileSystem.cpp
illwieckz: done

illwieckz:
I modified my code to handle the symlink use case (if `~/.unvanquished` was a symlink), in this case, it create a new symlink in `~/.local/share/unvanquished` but do not remove the first one (the safest choice), displaying a warning.

src/common/FileSystem.cpp
```diff
@@ -2067,7 +2067,59 @@ std::string DefaultHomePath()
...
+				if (err) {
+					Sys::Error("Could not create XDG data directory %s: %s", xdgDataHome, err.message());
+				}
+
+				if (S_ISLNK(stl.st_mode)) {
+					int ret;
+					int fd = open(xdgDataHome.c_str(), O_DIRECTORY);
+
+					if (fd == -1) {
+						Sys::Error("Could not open XDG data directory %s", xdgDataHome);
```
DolceTriade: include strerror(errno) at the end so we know why it failed.

src/common/FileSystem.cpp
```diff
@@ -2067,7 +2067,59 @@ std::string DefaultHomePath()
...
+					}
+
+					fsLogs.Warn("Creating legacy home path symlink %s to XDG home path %s", legacyHomePath, xdgHomePath);
+					ret = symlinkat(legacyHomePath.c_str(), fd, xdgHomePath.c_str());
+
+					if (ret == -1) {
+						Sys::Error("Could not create symlink %s", xdgHomePath);
```
DolceTriade: strerror(errno)

src/common/FileSystem.cpp
```diff
@@ -2096,14 +2096,14 @@ std::string DefaultHomePath()
 					int fd = open(xdgDataHome.c_str(), O_DIRECTORY);
 
 					if (fd == -1) {
-						Sys::Error("Could not open XDG data directory %s", xdgDataHome);
+						Sys::Error("Could not open XDG data directory %s: %s", xdgDataHome, strerror(errno));
```
illwieckz: @DolceTriade like that?

DolceTriade: yes

illwieckz:
Do you think it's OK to merge that change in `for-0.51.0` branch?

Not doing this would prevent @DefaultUser to rebase his #16 work over this code. I don't know how much this change things for him and how much he need to rebase on this code.

I guess it's OK if we say people wanting to try out for-0.51.0 are invited to symlink `~/.unvanquished` and `~/.local/share/unvanquished` (or creating a brand new `~/.local/share/unvanquished` if they want to test a clean start). Dæmon will not change anything if both exists, so developers can already work with the new path while keeping the old one to play some games using the last release.

DolceTriade:
yes its fine

DefaultUser:
@illwieckz updating #16 should be easy since all that needs to be changed in your new code are the two instances of `PRODUCT_NAME_LOWER` (change to `Gameinfo::getInstance().name()`)

PR #21, opened 2017-07-09
add some unreachable builtin on UNREACHABLE() assert macro

illwieckz:
On @Kangz [suggest](https://github.com/DaemonEngine/Daemon/pull/14#discussion_r114242327), I added an `UNREACHABLE()` assert macro, but  now the compiler raises some miswarning:

```
warning: control reaches end of non-void function [-Wreturn-type]
```

But we know the code asserts before reaching end of non-void function.
So, instead of adding some useless `return` after the `UNREACHABLE()` assert macro everwhere the compiler raises mistakenly a warning, I added support for some builtin that tells the compiler this code can't be reached.

If the compiler is too old or is'nt supported, it will raise the useless warning, which is not a big issue.

The `UNREACHABLE()` assert macro was renamed to `ASSERT_UNREACHABLE()` like
other assert macros.

### DIFF
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -235,4 +235,31 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #   define CONSTEXPR_FUNCTION_RELAXED
 #endif
 
+/* Compiler can be fooled when calling ASSERT_UNREACHABLE() macro at end of non-void function.
+ * In this case, compiler is complaining because control reaches end of non-void function,
+ * even if the execution flow is expected to be taken down by assert before.
+ *
+ * That's why we use these compiler specific unreachable builtin on modern compilers,
+ * ASSERT_UNREACHABLE() macro makes use of this UNREACHABLE() macro, preventing useless warnings.
+ * Unsupported compilers will raise "control reaches end of non-void function" warnings but
+ * that's not a big issue and that's likely to never happen (these compilers would be too old and
+ * would lack too much features to compile Daemon anyway).
+ *
+ * See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0627r0.pdf
+*/
+#if defined(_MSC_VER) //UNREACHABLE
+	#define UNREACHABLE() __assume(0)
+// both gcc, clang and icc defines __GNUC__, so the order is important
+#elif defined(__INTEL_COMPILER)
+	#define UNREACHABLE() __builtin_unreachable()
+#elif defined(__clang__)
+	#if __has_builtin(__builtin_unreachable)
+		#define UNREACHABLE() __builtin_unreachable()
+	#endif
+#elif (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5))
+	#define UNREACHABLE() __builtin_unreachable()
+#else // UNREACHABLE
+	#define UNREACHABLE()
+#endif // UNREACHABLE
+
 #endif // COMMON_COMPILER_H_
--- a/src/common/Assert.h
+++ b/src/common/Assert.h
@@ -154,7 +154,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     #define ASSERT_GE DAEMON_ASSERT_GE
 #endif
 
-// You can put UNREACHABLE() in places that must be never reached.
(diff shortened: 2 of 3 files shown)


illwieckz:
For information, this is the detailed warning I got:

```
[ 14%] Building CXX object CMakeFiles/engine-lib.dir/src/common/FileSystem.cpp.o
Daemon/src/common/FileSystem.cpp: In function ‘std::__cxx11::string FS::PakPath::ReadFile(Str::StringRef, std::error_code&)’:
Daemon/src/common/FileSystem.cpp:1313:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
Daemon/src/common/FileSystem.cpp: In function ‘std::chrono::_V2::system_clock::time_point FS::PakPath::FileTimestamp(Str::StringRef, std::error_code&)’:
Daemon/src/common/FileSystem.cpp:1408:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

illwieckz:
I tested against both gcc and clang, I can't test on msvc.

src/common/Compiler.h
```diff
@@ -235,4 +235,39 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 
+/* Compiler can be fooled when calling ASSERT_UNREACHABLE() macro at end of non-void function.
+ * In this case, compiler is complaining because control reaches end of non-void function,
+ * even if the execution flow is expected to be taken down by assert before.
+ *
+ * That's why we use these compiler specific unreachable builtin on modern compilers,
+ * ASSERT_UNREACHABLE() macro makes use of this UNREACHABLE() macro, preventing useless warnings.
+ * Unsupported compilers will raise "control reaches end of non-void function" warnings but
+ * that's not a big issue and that's likely to never happen (these compilers would be too old and
+ * would lack too much features to compile Daemon by the way).
```
DolceTriade: "to compile Daemon anyways"

src/common/Compiler.h
```diff
@@ -235,4 +235,39 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ *
+ * See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0627r0.pdf
+*/
+#if defined _MSC_VER
+	#define UNREACHABLE() __assume(0)
+#elif defined __GNUC__
```
DolceTriade: Does this define even do anything?

This might be more cleanly written as:
```
#if defined(_MSC_VER)  // UNREACHABLE
...
#elif defined(__INTEL_COMPILER) 
...

#elif (defined(_clang) && __has_builtin(__builtin_unreachable)) 
...
#elif (__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >=5)
...
#else
#define UNREACHABLE()
#endif // UNRREACHABLE

Comment (illwieckz) [inline src/common/Compiler.h]:
```c++
true
```

except it `__GNUC__ > 4` even if we don't know it's GCC, by the way, there would probably no usable compiler to fit that case.

illwieckz: For information, GCC reads this whole line:

```
#elif (defined(_clang) && __has_builtin(__builtin_unreachable)) 
```

and complain about `missing binary operator before token`, which is a cryptic error for not supporting `__has_builtin`. I have to make a sub test.

src/common/Compiler.h
```diff
@@ -235,4 +235,31 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ * That's why we use these compiler specific unreachable builtin on modern compilers,
+ * ASSERT_UNREACHABLE() macro makes use of this UNREACHABLE() macro, preventing useless warnings.
+ * Unsupported compilers will raise "control reaches end of non-void function" warnings but
+ * that's not a big issue and that's likely to never happen (these compilers would be too old and
+ * would lack too much features to compile Daemon anyway).
+ *
+ * See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0627r0.pdf
+*/
+#if defined(_MSC_VER) //UNREACHABLE
+	#define UNREACHABLE() __assume(0)
```
Kangz: I'm worried this might cause define name conflicts with test frameworks like gtest. We might have to namespace it as DAEMON_UNREACHABLE when that happens.

Kangz:
LGTM too, sorry for the slow review

PR #24, opened 2017-07-11
Fix a bunch of PVS-Studio static analyser warnings №2

illwieckz:
This a PR replaying Unvanquished/Unvanquished#935 by @dimhotepus on Unvanquished source tree after the engine split.

It obsoletes Unvanquished/Unvanquished#935

This is all @dimhotepus' work, except a small commit I just added to rename an iterator to prevent misread.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -1122,7 +1122,7 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips,
 
 				if( image->bits & IF_NORMALMAP ) {
 					c = scaledWidth * scaledHeight;
-					for ( i = 0; i < c; i++ )
+					for ( int i = 0; i < c; i++ )
 					{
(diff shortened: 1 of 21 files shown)


src/engine/renderer/tr_image.cpp
```diff
@@ -1122,7 +1122,7 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips,
 
 				if( image->bits & IF_NORMALMAP ) {
 					c = scaledWidth * scaledHeight;
-					for ( i = 0; i < c; i++ )
+					for ( int j = 0; j < c; j++ )
```
Kangz: This should do s/i/j/ inside the loop body.

illwieckz: oh, I'm stupid ^^

illwieckz: fixed

PR #27, opened 2017-09-06
fix badly rendered skyboxes

illwieckz:
fix badly rendered skyboxes, @gimhael's suggestion.

example of bad rendering:
[badly rendered skybox](http://dl.illwieckz.net/b/unvanquished/bugs/ugly-skybox/unvanquished_2017-08-09_042549_000.jpg)
[badly rendered skybox](http://dl.illwieckz.net/b/unvanquished/bugs/ugly-skybox/unvanquished_2017-08-09_055650_000.jpg)
[badly rendered skybox](http://dl.illwieckz.net/b/unvanquished/bugs/ugly-skybox/unvanquished_2017-08-09_060043_000.jpg)

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -2703,7 +2703,7 @@ static void ParseSkyParms( const char **text )
 	{
 		Q_strncpyz( prefix, token, sizeof( prefix ) );
 
-		shader.sky.outerbox = R_FindCubeImage( prefix, IF_NONE, filterType_t::FT_DEFAULT, wrapTypeEnum_t::WT_EDGE_CLAMP );
+		shader.sky.outerbox = R_FindCubeImage( prefix, IF_NOCOMPRESSION, filterType_t::FT_DEFAULT, wrapTypeEnum_t::WT_EDGE_CLAMP );
 
 		if ( !shader.sky.outerbox )
 		{


illwieckz:
For reference (quoted from irc development channel):
> gimhael> illwieckz: the cause of the skybox badness is that the engine asks the GL driver to use on-the-fly DXT compression for the images
> gimhael> the GL compressors are usually written for speed not for quality, so the image looks bad
> gimhael> this is the default for all images unless the user adds the "nocompress" option in the shader file
>  gimhael> only for skyboxes nocompress was ignored and compression was always used
>  gimhael> now we could disable no-the-fly compression for skyboxes, or we could make "nocompress" work in skybox shaders, but I think we could drop the texture compression for all images and make "nocompress" obsolete
> gimhael> if anybody wants to use compressed textures, they should be compressed with a offline compressor like crunch, because the quality is much better

The skybox texture is a very special texture since it's static and unlike walls and other objects, does not move on screen while the player is moving, so any artefact is very visible and never blurred by move. Also, the skybox is renderer on the whole viewport on open area, letting any artefact very visible. That's why badly compressed skyboxes are doing nothing else than ruining the scene.

Because the game is very fast and because of movement, we can use bad compression everywhere except skyboxes and lightmaps: nothing can hide a skybox artifact (same for lightmaps).

You can have a look at this comparison I did once:

[lightmap compression comparison](https://www.unvanquished.net/wp-content/uploads/2015/08/wasteland.webp-jpg-comparison1-1024x398.png)

It does not included crunch comparison, but crunch is generally worst than jpeg, and on-the-fly opengl compression is far far worst than crunch in any case. So, as you see, compressing a skybox with a bad algorithm is always a very bad idea.

@gimahel also suggested to drop the ugly on-the-fly texture compression since modern rendering techniques (like Vulkan) does not provide it. Instead, we must offline compression (using crunch) what we already do.

For reference, my [Urcheon](https://github.com/illwieckz/Urcheon) packaging tool picks the best compression profile for each use case, so there is really no need to keep on the fly compression.

Whatever we decide about removing on-the-fly compression ability or not, this commit disallow on-the-fly compression on skyboxes because that's always a bad idea to badly compress a skybox.

The `nocompress` shader keyword is hence deprecated: it's never a good idea to badly compress a skybox, and keeping this bad compression as a default behavior makes no sense anymore.

PR #28, opened 2017-09-09
Remove IF_NOCOMPRESSION and compression-related cvars.

gimhael:
This change removes support for on-the-fly texture compression in the Daemon engine. The cvars /r_compress*Maps have been removed and the "nocompress" shader keyword has no function, but is kept for backward compatibility.

Anybody who wants to use compressed textures should use an offline texture compression tool, and provide the textures in .dds or .crn format.


illwieckz:
Works on my end.

PR #36, opened 2017-10-03
Fix -Wimplicit-fallthrough warnings

DolceTriade:
Add a conditional macro "DAEMON_FALLTHROUGH" for compilers that care
about this type of thing (GCC7 and Clang6 for now). Update Tinyformat
because they claimed to have fixed the warnings.

### DIFF
--- a/src/engine/client/cl_cgame.cpp
+++ b/src/engine/client/cl_cgame.cpp
@@ -604,11 +604,7 @@ static int LAN_ServerIsVisible( int source, int n )
 				return cls.localServers[ n ].visible;
 			}
 
-		if ( Cmd_Argc() == 1 )
-		{
-			Log::Warn("Server sent a pubkey_decrypt command, but sent nothing to decrypt!" );
-			return false;
-		}
+			break;
 
 		case AS_GLOBAL:
 			if ( n >= 0 && n < MAX_GLOBAL_SERVERS )
--- a/src/engine/qcommon/cmd.cpp
+++ b/src/engine/qcommon/cmd.cpp
@@ -347,6 +347,7 @@ static const char *EscapeString( const char *in, bool quote )
 			// no need to quote semicolons if in ""
 			quoted = true;
 			if ( quote ) break;
+            DAEMON_FALLTHROUGH;
 		case '"':
 		case '$':
 		case '\\':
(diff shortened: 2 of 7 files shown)


src/engine/client/cl_cgame.cpp
```diff
@@ -604,11 +604,7 @@ static int LAN_ServerIsVisible( int source, int n )
 				return cls.localServers[ n ].visible;
 			}
 
-		if ( Cmd_Argc() == 1 )
-		{
-			Log::Warn("Server sent a pubkey_decrypt command, but sent nothing to decrypt!" );
-			return false;
-		}
+			break;
```
illwieckz: what's that?

Kangz: Yeah it's weird that this cade was there.

src/engine/qcommon/cmd.cpp
```diff
@@ -347,8 +347,11 @@ static const char *EscapeString( const char *in, bool quote )
 			// no need to quote semicolons if in ""
 			quoted = true;
 			if ( quote ) break;
+            DAEMON_FALLTHROUGH;
```
Kangz: Don't you need only the first FALLTHROUGH?

src/engine/qcommon/parse.cpp
```diff
@@ -2377,30 +2379,31 @@ static int Parse_EvaluateTokens( source_t *source, token_t *tokens, signed long
 									break;
 								}
 							}
+							DAEMON_FALLTHROUGH;
 
-						case punctuationSub_t::P_MUL:
-						case punctuationSub_t::P_DIV:
-						case punctuationSub_t::P_MOD:
-						case punctuationSub_t::P_ADD:
+						case punctuationSub_t::P_MUL: DAEMON_FALLTHROUGH;
```
Kangz: It seems that here and in other places in the CL, having multiple case statements together shouldn't require FALLTHROUGH.

src/engine/client/cl_cgame.cpp
DolceTriade: That code was wrong. Probably the result of a bad merge conflict resolution. You can find the exact same code earlier in the file.

src/engine/qcommon/parse.cpp
DolceTriade: Hm, maybe. I was over zealous. I will remove some and see if the warnings come back.

src/engine/client/cl_cgame.cpp
illwieckz: okie!

src/engine/qcommon/cmd.cpp
DolceTriade: Yes, you are correct.

DolceTriade:
Fixed unnecessary DAEMON_FALLTHROUGHs

PR #37, opened 2017-10-08
Add utility class for performing an action only if at least N milliseconds have passed since the last time.

slipher:
I want to use this to log warnings for some conditions which may occur every frame, without causing massive spam.

### DIFF
--- a/src/common/Util.h
+++ b/src/common/Util.h
@@ -181,6 +181,29 @@ private:
 	uninitialized& operator=(uninitialized&&) = delete;
 };
 
+// Use to perform an action only if a minimum time has passed since the last time it was performed.
+// Not thread-safe.
+class MinimumDelay {
+public:
+	// duration: length of delay in milliseconds
+	MinimumDelay(int duration): duration(duration), lastTime(std::numeric_limits<int>::min()) {}
+
+	// Returns true if it is OK to perform the action again.
+	bool Check() {
+		int Sys_Milliseconds();
+		int now = Sys_Milliseconds();
+		if (now >= lastTime && now < lastTime + duration) {
+			return false;
+		} else {
+			lastTime = now;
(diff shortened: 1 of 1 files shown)


DolceTriade:
It would be helpful if you could commit this with how you use it.

src/common/Util.h
```diff
@@ -181,6 +181,29 @@ template<typename T> class uninitialized {
...
+		int now = Sys_Milliseconds();
+		if (now >= lastTime && now < lastTime + duration) {
+			return false;
+		} else {
+			lastTime = now;
+			return true;
+		}
+	}
+private:
+	int duration;
```
DolceTriade: const

src/common/Util.h
```diff
@@ -181,6 +181,29 @@ template<typename T> class uninitialized {
...
+	int lastTime;
```
DolceTriade: const

src/common/Util.h
```diff
@@ -181,6 +181,29 @@ template<typename T> class uninitialized {
...
 
+// Use to perform an action only if a minimum time has passed since the last time it was performed.
+// Not thread-safe.
+class MinimumDelay {
+public:
+	// duration: length of delay in milliseconds
+	MinimumDelay(int duration): duration(duration), lastTime(std::numeric_limits<int>::min()) {}
+
+	// Returns true if it is OK to perform the action again.
+	bool check() {
```
DolceTriade: const function. Also, I think our style has us capitalizing function names.

src/common/Util.h
slipher: Done.

src/common/Util.h
slipher: It is a key part of this method's contract that the internal state is modified.

PR #39, opened 2017-10-09
fix xdgDataHome retrieving

illwieckz:
Uh, was a real bug, thanks Travis & LLVM (GCC you still have things to learn)!
Le'ts wait on Travis to see if it does not have more stuff to tell us.


illwieckz:
OK, [Travis is now happy](https://travis-ci.org/DaemonEngine/Daemon/builds/285668143).

PR #40, opened 2017-10-10
don't substract player count by allowed privateClients number

illwieckz:
## don't substract player count by allowed privateClients number

A private client is a kind of slot only available to someone having the password,
allowed to ensure this one can join even if the server is full.

1. this is a real kind of players so there is no reason to substract the number
of private client to player count.
2. It makes no sense to substract all the time, even when no one private client
is connected…

Fix Unvanquished/Unvanquished#1028

## do not disclose private slot number except for those taken

returns to client sv_maxclients as sv_maxclients - sv_privateClients
except if client count + bot count is greater (which means some private
slots are taken) so returns client count + bot count in this case instead.

i.e. hide private slot number when possible, but never returns a sv_maxclient
count that is smaller than the advertised client number.

### DIFF
--- a/src/engine/server/sv_main.cpp
+++ b/src/engine/server/sv_main.cpp
@@ -627,13 +627,18 @@ void SVC_Info( netadr_t from, const Cmd::Args& args )
 		}
 	}
 
+	// do not disclose private slot number except for those taken
+	int maxclients = sv_maxclients->integer - sv_privateClients->integer;
+	int clients = count + botCount;
+	maxclients = clients > maxclients ? clients : maxclients;
+
 	info_map["protocol"] = std::to_string( PROTOCOL_VERSION );
 	info_map["hostname"] = sv_hostname->string;
 	info_map["serverload"] = std::to_string( svs.serverLoad );
 	info_map["mapname"] = sv_mapname->string;
 	info_map["clients"] = std::to_string( count );
 	info_map["bots"] = std::to_string( botCount );
-	info_map["sv_maxclients"] = std::to_string( sv_maxclients->integer - sv_privateClients->integer );
+	info_map["sv_maxclients"] = std::to_string( maxclients );
 	info_map["pure"] = std::to_string( sv_pure->integer );
 
 	if ( sv_statsURL->string[0] )
--- a/src/engine/server/sv_main.cpp
+++ b/src/engine/server/sv_main.cpp
@@ -571,11 +571,10 @@ void SVC_Info( netadr_t from, const Cmd::Args& args )
 
 	SV_ResolveMasterServers();
 
-	// don't count privateclients
 	int botCount = 0;
 	int count = 0;
 
-	for ( int i = sv_privateClients->integer; i < sv_maxclients->integer; i++ )
+	for ( int i = 0; i < sv_maxclients->integer; i++ )
 	{
 		if ( svs.clients[ i ].state >= clientState_t::CS_CONNECTED )
 		{


illwieckz:
Note that I haven't compiled neither tested the code, but it looks very obvious isn't it?

illwieckz:
I just tested the code and it displays as expected.

illwieckz:
@Kangz can I have your thought now that I also modified the way the server returns maxclients (including private slots connected)?

DolceTriade:
I think the goal should be to accurately display the number of empty slots (accessible without a password), since that's what most people care about.

illwieckz:
> the current code will show the server as full if there is 1 private client connected and the server has 1 empty slot left, which is not good

Which is _expected_ because private slot is to allow people with password to override the limit, not to give anyone the ability to override the limit as soon as there is one private guy on server.

These two scenarii are wrong:

> bouncer: do _you have an invitation card?_
> stuffed guy: _no but this guy who is already in have one for himself_
> bouncer: _ok, come in_

Or

> bouncer: do _you have an invitation card?_
> stuffed guy: _no, but the owner is already in_
> bouncer: _ok, come in_

It's wrong…

@DolceTriade please read these two comments there:
https://github.com/Unvanquished/Unvanquished/issues/1028#issuecomment-335846473

DolceTriade:
Keep in mind that the private clients are subtracted from the max clients, so you're not allowing anyone to override the limit.

Maximum occupancy is 32, but we will unconditionally reserve two seats for VIPs. Regardless of whether there are any VIPs present, we will always allow 30 regular people.

illwieckz:
Well, so it means the VIP always comes with an extra seat for someone-else if the server is not yet full. Why not but we need a way to know how many VIP are currently connected, at this place of this code.

illwieckz:
Well, so it means the VIP always comes with an extra seat (for himself if the server is full, or for someone else if the server is not yet full). My current implementation being "the VIP comes with an extra seat for himself if the server is full”. Why not but we need a way to know how many VIP are currently connected, at this place of this code.

Note that the behavior of these two policy if there is 1 private slot but two VIP wants to join:

- _“the VIP always comes with an extra seat for himself if the server is full, or for someone else if the server is not yet full”_ policy:  
- - 30 players join, 1 VIP can join.
- - 29 players and 1 VIP join, 1 player or 1 VIP can join. If 1 player join, the other VIP can't.

- _“the VIP comes with an extra seat for himself if the server is full”_ policy:
- - 30 players join, 1 VIP can join.
- - 29 players and 1 VIP join, only 1 other VIP can join.

NOTE: the real slot number is always disclosed by _“B”_ and _“P”_ rules in statusResponse packet, the maxclients value in infoResponse is just a number to help the UI to flag a server as full. It's always up to the server to say if the server is really full when the player is trying to connect to even if it's flagged as full in UI.

The VIP knows he can try to connect even if the UI is flagging the server as full. A server owner can allocate a large number of VIP seats (to be sure a VIP will not hit the wall) but if VIP seats are given to non-VIP people anytime a VIP enters the room, it makes allocating large number of VIP seats a dangerous idea. It means a VIP is not someone who has a free pass for himself, but someone who is giving a free pass to the first random guy coming.

slipher:
The way I look at it, this question (which is orthogonal to the issue of how the number of free slots is displayed) is the matter of whether someone with VIP credentials will sit in the VIP seat, or a normal seat, in the case that both are available (I think the analogy to seats really holds since the player sits at fixed location in an array, with the current implementation). Apparently the default is to sit in the VIP seat, though I haven't checked that. It might make more sense to do as @illwieckz suggests and have VIPs use normal seats if they are available, to reserve VIP seats in the case that there exist more VIPs than VIP seats.

illwieckz:
I still don't know if VIP people takes a VIP sit or not when they connect on a non-full server, and we can't know because that code is completely broken (see #41): currently all slots (private or not) are available to anyone.

The purpose of this PR is to makes the number displayed right, which it does. We can do other PR and create other issues lately, and perhaps when the private mechanism will be repaired it will give us better ideas for private seat policy and someone will also rewrite it.

The fact is:
1. The client count is always wrong as soon as at least one private slot is configured
2. Private mechanism is broken after all and act like if it was not existing at all, so we are just losing our time talking about things we can't implement yet.

So I just need to know if I can merge that code to ensures the client count is always correct even when someone activates the not-working private client mechanism, and if I can merge that code that ensures the maxclient is never smaller than the total number of players on the server.

i.e using that PR ensures client/maxclient is never displaying something weird, even if private mechanism is broken. Btw even if private mechanism is broken it at least already works to inadvise a player to join a server, making it looking full in server browser, which is expected.

slipher:
In what way is it completely broken? #41 is just about what happens when the password is the empty string.

illwieckz:
Or perhaps I'm doing it wrong, but it totally ignores my wrong passwords…

slipher:
In the code snippet @illwieckz posted in #41, there is a special case for connections from a local address. Can you still connect without the right password when the client is not on the same local network as the server?

slipher:
Apologies for the earlier confusion: it's true that the password really is ignored in (almost) all cases.

By the way, I thought of a counterexample to the assumption that filling public slots first should always be preferable to filling private slots first. Suppose you have a 4v4 clan match, so you reserve 8 password slots for the combatants. Then you want them to occupy the private slots, while leaving as many public slots for spectators as possible.

Perhaps the really sensible thing to do would be to abandon the concept of private slots, and simply count the number of players who know the password, regardless of their client number. Could be an annoying amount of work though.

DolceTriade:
idk if private clients are even worth keeping.

illwieckz:
obsoleted by #54

PR #43, opened 2017-10-12
Redesign key bind handling, to support Unicode among other things

slipher:
See #19 for more detail (I had to make a new PR since it was merging into the wrong branch).

This change implements the engine side of the proposed bind system. You can bind keys via the console use the binds, but not yet create them in the UI. The next step after this will be to make an ABI change to allow the cgame to fully use the new binds, and modify the bind menu logic.

This can be tested with the current gamelogic. If you do, make sure to back up `keybindings.cfg` if necessary (the format is forward (ish) but not backward-compatible).


src/engine/client/cl_input.cpp
```diff
@@ -1146,21 +1149,24 @@ void IN_BuiltinButtonCommand()
...
-	int key, time;
+	int time;
 	int i;
 	bool first = true;
 
 	check = atoi( Cmd_Argv( 1 ) );
-	key   = atoi( Cmd_Argv( 2 ) );
-	time  = atoi( Cmd_Argv( 3 ) );
+	Keyboard::Key key = Keyboard::StringToKey( Cmd_Argv( 2 ) );
+    if ( !key.IsValid() ) {
```
DolceTriade: indentation seems off

illwieckz:
Not related to the PR but:

> I had to make a new PR since it was merging into the wrong branch

I discovered recently I can change PR's target branch while cliking on "Editing" button, the one to edit the PR title for example. Well, it's not well documented and I discovered that by luck.

Kangz:
This is an extremely hard to review PR. Is it possible to start landing chunks of it that modernize the engine while keeping the old behavior. Then add the changes that introduce new functionality? I understand this might be a hassle :/

slipher:
@Kangz This is doable to some extent. But I'd like to get some review of the actual functionality first, separate from the lines of code. I mean, no point in spending all the time splitting up the commits if people think the design is bad.

Viech:
I'm closing Unvanquished/Unvanquished#813, so please make sure to create a new issue for this repository in case the former issue is not fixed once this PR gets merged or closed.

slipher:
Re #809: I doubt it, sounds more like an SDL level issue.

src/engine/client/cl_input.cpp
TimePath: tabs vs spaces

slipher:
I've updated the PR to point at a topic branch (which is currently a copy of for-0.51.0), since it may not be ready for 0.51. At this point it seems difficult to split off any more small, readable pieces without more work than writing the code in the first place.

src/engine/client/cl_cgame.cpp
```diff
@@ -1485,35 +1484,33 @@ void CGameVM::QVMSyscall(int index, Util::Reader& reader, IPC::Channel& channel)
...
-                        if (bind == buffer) {
-                            result.back().push_back(i);
-                            continue;
-                        }
-                    }
-                }
+				for (const auto& bind : binds) {
+					result.push_back({});
+					for (int i = 1; i < Util::ordinal(keyNum_t::MAX_KEYS); i++) {
+						if (i >= 'A' && i <= 'Z') {
```
DolceTriade: Could you add a comment on why you can skip these? Are we only allowed to return binds for keyboard keys here?

slipher: If it looks up both 'a' and 'A' then those binds are double-counted. The whole thing is a temporary implementation; to really work the API needs to change.

DolceTriade: Sounds like a good comment to include in the source code.

slipher: Done.

PR #46, opened 2017-10-22
Make the use of subject-specific loggers automatic.

slipher:
Instead of doing `Logger fooLogger("foo");` ... `fooLogger.Notice(...);` this allows us to set up the foo logger once, and then simply use `Log::Notice(...)` later, instead of having to remember to use `fooLogger`. 

A motivation for this is to be able to make a macro for logging every N seconds without having to reference the name of a specific logger.

`CommandSystem.cpp` is included as one example of what the code would look like with this method.


slipher:
@Kangz Can you take a look at this?

Kangz:
TBH I'm not a fan of this approach of having an implicit logger. What's the issue with referencing the name of a logger in the macros?

slipher:
It's annoying in general to have to remember to check for a special logger and remember what it is called. I would normally just write `Log::Warn` out of rote memory.

Also, in the existing examples in the codebase, every logger that is declared is used for 100% of the logging calls in the file, and no logger is used in more than one file.

slipher:
Any thoughts @Kangz? Do you think more varied usage patterns of loggers are likely to arise often?

Kangz:
Same comment as before. Having things explicit is better than them being implicit imho (and the static logger in the header is a bit too much magic). It isn't that much overhead to figure out the name of the logger, but it makes it very clear what goes where.

PR #49, opened 2017-10-29
Refactor event loop

slipher:
This is another one of the commits from #43 that is mostly independent of the functional changes.

### DIFF
--- a/libs/tinyformat/tinyformat.h
+++ b/libs/tinyformat/tinyformat.h
@@ -876,7 +876,8 @@ class FormatListN : public FormatList
         template<TINYFORMAT_ARGTYPES(n)>                       \
         FormatListN(TINYFORMAT_VARARGS(n))                     \
             : FormatList(&m_formatterStore[0], n)              \
-        { TINYFORMAT_ERROR(n == N); init(0, TINYFORMAT_PASSARGS(n)); }   \
+        { if(n != N) TINYFORMAT_ERROR("metaprogramming fail"); \
+          init(0, TINYFORMAT_PASSARGS(n)); }                   \
                                                                \
         template<TINYFORMAT_ARGTYPES(n)>                       \
         void init(int i, TINYFORMAT_VARARGS(n))                \
--- a/src/common/Assert.h
+++ b/src/common/Assert.h
@@ -51,6 +51,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *     - Use the assert information to help the compiler optimizer in release builds.
  */
 
+#ifdef DEBUG_BUILD
+#define DAEMON_ASSERTS_ENABLED
+#endif
+
 // MSVC triggers a warning in /W4 for do {} while(0). SDL worked around this by using
 // (0,0) and points out that it looks like an owl face.
 #if defined(_MSC_VER)
@@ -61,7 +65,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // DAEMON_ASSERT_CALLSITE_HELPER generates the actual assert code. In Debug it does what you would
 // expect of an assert and in release it tries to give hints to make the compiler generate better code.
-#if defined(DEBUG_BUILD)
+#ifdef DAEMON_ASSERTS_ENABLED
     #define DAEMON_ASSERT_CALLSITE_HELPER(file, func, line, _0, code, condition, message) \
         do { \
             code; \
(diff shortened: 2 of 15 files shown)


src/engine/qcommon/common.cpp
```diff
@@ -770,62 +771,37 @@ EVENT LOOP
 static const int MAX_QUEUED_EVENTS  = 1024;
 static const int MASK_QUEUED_EVENTS = ( MAX_QUEUED_EVENTS - 1 );
 
-static sysEvent_t eventQueue[ MAX_QUEUED_EVENTS ];
+static std::unique_ptr<Sys::EventBase> eventQueue[ MAX_QUEUED_EVENTS ];
```
DolceTriade: How come you didn't use a vector or list?

src/engine/qcommon/common.cpp
```diff
@@ -912,39 +871,92 @@ void Com_RunAndTimeServerPacket( netadr_t *evFrom, msg_t *buf )
...
 int Com_EventLoop()
 {
-	sysEvent_t ev;
 	netadr_t   evFrom;
 	byte       bufData[ MAX_MSGLEN ];
 	msg_t      buf;
 
-	int        mouseX = 0, mouseY = 0, mouseTime = 0;
+	int        mouseX = 0, mouseY = 0;
 	bool       mouseHaveEvent = false;
```
DolceTriade: hadMouseEvent?

src/engine/qcommon/common.cpp
slipher: I didn't really think about it, but it doesn't necessarily seem like a good thing for an unlimited amount of events to accumulate if the game is frozen or whatever.

src/engine/qcommon/common.cpp
slipher: Done.

Kangz:
What is a "multi-configuration build" mentioned in the thirdcommit message?

slipher:
@Kangz In Visual Studio, you configure Debug/Release/RelWithDebInfo all at once. You can switch between the modes without wiping out the binaries or rerunning CMake.

src/engine/sys/sys_events.h
```diff
@@ -0,0 +1,140 @@
...
+#include "engine/qcommon/qcommon.h"
+
+namespace Sys {
+
+class EventBase {
+public:
+    const sysEventType_t type;
+    EventBase(sysEventType_t type) : type(type) { }
+
+    virtual ~EventBase() = 0;
```
Kangz: Choose if it is = 0 or defined below :P

src/engine/sys/sys_events.h
```diff
@@ -0,0 +1,140 @@
...
+    const int time;
+    KeyEvent(int key, bool down, int time):
+        EventBase(ClassType()), key(key), down(down), time(time) {}
+};
+
+class CharEvent: public EventBase {
+public:
+    static constexpr sysEventType_t ClassType() { return sysEventType_t::SE_CHAR; }
+
+    const int ch; // Unicode code point
```
Kangz: nit: uint32_t for a code point?

Kangz:
Ooooh nice, that multi-configuration sounds great then

libs/tinyformat/tinyformat.h
```diff
@@ -876,7 +876,8 @@ class FormatListN : public FormatList
         template<TINYFORMAT_ARGTYPES(n)>                       \
         FormatListN(TINYFORMAT_VARARGS(n))                     \
             : FormatList(&m_formatterStore[0], n)              \
-        { TINYFORMAT_ERROR(n == N); init(0, TINYFORMAT_PASSARGS(n)); }   \
+        { if(n != N) TINYFORMAT_ERROR("metaprogramming fail"); \
```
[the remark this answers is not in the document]
slipher: Well, it was `assert(n == N)` upstream. Then that was changed in the interest of removing `assert` from the dependencies.

src/common/Assert.h
```diff
@@ -157,4 +157,41 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // You can put ASSERT_UNREACHABLE() in places that must never be reached.
 #define ASSERT_UNREACHABLE() DAEMON_ASSERT_CALLSITE(false, , false, "Unreachable code hit."); UNREACHABLE();
 
+#ifdef DEBUG_BUILD
+// This stuff is so that the ASSERT_cc variants can be used on objects that don't have a
+// function defined for printing to a stream.
+namespace AssertionImpl {
```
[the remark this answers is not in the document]
slipher: I don't think it's normal to have a namespace detail in the global namespace. Should I create a namespace Assert and have Assert::detail?

src/engine/client/cl_console.cpp
```diff
@@ -1270,10 +1270,13 @@ void Con_ScrollToTop()
 	//consoleState.bottomDisplayedLine = consoleState.scrollLineIndex;
 }
 
-void Con_ScrollToBottom()
+void Con_ScrollToBottom(bool instantly)
 {
 	consoleState.scrollLineIndex = consoleState.lines.size() - 1;
-	//consoleState.bottomDisplayedLine = consoleState.currentLine;
+	if (instantly)
```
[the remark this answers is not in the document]
slipher: A crash occurred because `bottomDisplayedLine` is used [here](https://github.com/DaemonEngine/Daemon/blob/master/src/engine/client/cl_console.cpp#L885) to index into the lines array.

src/engine/sys/sys_events.h
slipher: There's some weird rule with virtual destructors where you have to have both.

libs/tinyformat/tinyformat.h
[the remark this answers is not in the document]
Kangz: Thanks for the explanation, sounds good.

src/engine/sys/sys_events.h
Kangz: I didn't know about this, do you have a reference?

slipher: I should say, *pure* virtual destructors, to be precise. https://stackoverflow.com/a/21110040

src/common/Assert.h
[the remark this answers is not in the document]
Kangz: Or AssertDetail?

src/engine/qcommon/common.cpp
slipher: Hmm, it may be irrational but I am loath to replace a perfectly good array with some wacked out vector of linked list nodes thing which is std::queue.

slipher:
I'm gonna merge this unless someone wants me to use a deque really badly :)

src/engine/qcommon/common.cpp
Kangz: std::queue should be implemented with std::dequeue which is a bit smarter than that.

Kangz:
There's a bunch of other comments that haven't been addressed. Please address them first.

src/engine/qcommon/common.cpp
slipher: What do you mean about deque being smarter than that? That's what I was trying to describe. Just the part where the linked list nodes are small arrays?

deque seems like such a pointless data structure to have... I guess it's some kind of workaround for the lack of unique_ptr in older versions? Otherwise you could just have a double-ended vector and put unique_ptrs in it.

slipher:
Sorry, I had fixed those but never uploaded the code.

Kangz:
Thanks LGTM

PR #50, opened 2017-11-01
Revert "bump protocol since dpk/pk3 support introduced pak list incompatibilities"

illwieckz:
See #18 for the initial change that is now to be reverted.

The reverted change was motivated by the fact the VFS change to introduced DPK support and enable legacy PK3 download brokes too many things (which has implication on package downloading) so it sounded like a good idea to raise protocol number.

Well, it raises many other issues. First, protocol number is hardcoded in client, it can be fixed easily but _meh_. Second, unlike some idtech derivative games that has history of multiple protocol, tremulous never changed the number and unvanquished neither, because of that, no one third-party tool implemented a way to ask for server from all numbers, but they all hardcode the protocol number, it means `qstat`, `xqf`, `apelsin`, `osavul` and `obozrenie` will be broken, and most of them would have to be recompiled, which looks to be unlikely to happen for some of them. For example some people who plays both tremulous and unvanquished still use apelsin today. Bumping the protocol means Unvanquished will disappear from their screens.

I've checked, and we introduced enough compat breaking changes to not have to worry. I verified that Unvanquished 0.50.0 client can't connect to Unvanquished 0.51.0 server and Unvanquished 0.51.0 client can't connect to Unvanquished 0.50.0 (it just returns to main menu in both case). The same way you can't connect from early Unvanquished client to current server just thanks to introduced incompatibilities and it was considered OK at this time.

This reverts commit a2177af6e3f709bcbbef07ff0bc43f28daa3b76d.

### DIFF
--- a/src/engine/qcommon/qcommon.h
+++ b/src/engine/qcommon/qcommon.h
@@ -283,7 +283,7 @@ The server you are attempting to join is running an incompatible version of the 
 The server you attempted to join is running an incompatible version of the game.\n\
 You or the server may be running older versions of the game."
 
-#define PROTOCOL_VERSION       87
+#define PROTOCOL_VERSION       86
 
 #define URI_SCHEME             GAMENAME_STRING "://"
 #define URI_SCHEME_LENGTH      ( ARRAY_LEN( URI_SCHEME ) - 1 )


illwieckz:
Basically 0.50.0 client trying to connect to 0.51.0 server disconnects itself and returns to menu printing that warning on console:

```
Warn: CL_ParseServerMessage: Illegible server message 8 
```

On the other side, 0.51.0 client trying to connect to 0.50.0 server disconnects itself and returns to menu printing that warning on console:

```
Warn: CL_ParseServerMessage: Illegible server message 9 
```

So, no need to bump protocol: both versions are already mutually exclusive and client/server mismatch leads to a clean disconnection.

slipher:
But it has been bumped before, e.g. 01fcfd7a5503f2cd867b24cda6951f6059c22f76

illwieckz:
Oh, nice catch!

By the way it was 5 years ago and all softwares out there enforce `86` on master queries…

illwieckz:
See also the 85→86 bump by @DolceTriade 5 years ago:
https://github.com/DaemonEngine/Daemon/commit/a93d08da518bd8d78991cccf75b98c541f0df6e6#diff-302ac15228a45521d6788eceaf981874L292

illwieckz:
So, even if we already did it in the past, it don't change the fact it's probably a very bad idea to break all the third-party software out there when client and server already handles the mismatch nicely.

slipher:
I don't understand this thing about 3rd party code; regardless, I think the current UX is poor and should be improved (which doesn't necessarily mean changing the number).

Also we should remember there are 2 kinds of protocols which can be mismatched, vm<>engine and netcode.

illwieckz:
Here it's all about netcode.

It's (0.50 vm on 0.50 engine) on a side trying to connect to (0.51 vm on 0.51 engine) on another side.

On this scenario the 0.50 vm is never executed on 0.51 engine and 0.51 vm is never executed on 0.50 engine, because disconnect happens before any package are downloaded.

If you try to make a frankenstein stuff mixing 0.50 and 0.51 stuff badly, then trying to connect a 0.51 server, you get nicely disconnected with a warning saying a string about package to download is not parseable, which is expected and correct. And this case will never happen with regular players because even sometime downloading 0.50 asset and 0.51 engine will not be able to load the assets at all (the engine will not find them in file system), same for the opposite. To get that warning about package string not being parseable you have to build some intermediary commits that will not be released as is and use intermediary assets no one will use. And in that worst case, you just get a warning and a clean disconnection.

PR #51, opened 2017-11-03
FileSystem: Replace bad assert with a verbose log

DolceTriade:
This assert will trigger everytime there is an unrecongized file
in the pakpath. This is bad and wrong.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -2285,7 +2285,7 @@ static void FindPaksInPath(Str::StringRef basePath, Str::StringRef subPath)
 			} else if (Str::IsSuffix("/", filename)) {
 				FindPaksInPath(basePath, Path::Build(subPath, filename));
 			} else {
-				ASSERT_UNREACHABLE();
+				fsLogs.Verbose("Ignoring file: %s", filename);
 			}
 		}
 	} catch (std::system_error&) {


illwieckz:
Yeah, perturbed spoted that some days ago, it's ugly. :scream: 
> 2017-10-30 00:06:48 +0100 \<perturbed\> wtf, why is this here https://github.com/DaemonEngine/Daemon/blob/for-0.51.0/src/common/FileSystem.cpp#L2288
> 2017-10-30 00:06:49 +0100 \<granger\>     ¯\\_(ツ)_/¯
> 2017-10-30 00:07:07 +0100 <perturbed\> to crash if any file with a non-approved extension is found?
> 2017-10-30 00:09:10 +0100 \<illwieckz\>   Was a suggest by kangz there: https://github.com/DaemonEngine/Daemon/pull/14#discussion_r114242327
> 2017-10-30 00:10:11 +0100 \<perturbed\> only found it now because there was a bug in the cmake code that caused debug assertions to be disabled
> 2017-10-30 00:10:22 +0100 \<illwieckz\>   but mmmm, for (auto& filename: RawPath::ListFiles(fullPath))
> 2017-10-30 00:10:37 +0100 \<illwieckz\>   is those if really done on every file, pak or not?
> 2017-10-30 00:11:14 +0100 \<perturbed\> RawPath::ListFiles seems pretty self explanatory, it would be every file
2017-10-30 00:13:19 +0100 \<illwieckz\>   it does not seems to assert on the dummyfile I just created
2017-10-30 00:14:41 +0100 \<perturbed\> did you build in debug mode?
2017-10-30 02:39:15 +0100 \<perturbed\> once you actually enable assertions, it's amazing how many there are that always fail
```

+1 for the log instead

PR #54, opened 2017-11-12
Fix some issues regarding private client slots

### LINKED ISSUE #41: sv_privateclients does not require sv_privatePassword to be set
I just spawn a server with `sv_maxclients` set to `4` and `sv_privateclients` set to `2` and 4 non-private clients can join.

slipher:
- Avoid potential crashes with out of bounds value of `sv_privateClients`.
- Implement my suggestion [here](https://github.com/Unvanquished/Unvanquished/issues/1028#issuecomment-335695254) for server stats in the presence of private slot users. Namely, the difference between the number of current clients and the max should be the number of open public slots. For comparison, #40 proposes max clients = max(number of public slots, number of clients).
- Actually prevent users without the right password from occupying the private slots (fixes #41).

### DIFF
--- a/src/engine/server/sv_main.cpp
+++ b/src/engine/server/sv_main.cpp
@@ -32,8 +32,9 @@ Maryland 20850 USA.
 ===========================================================================
 */
 
+#include "common/Common.h"
+
 #include "server.h"
-#include "common/Assert.h"
 #include "CryptoChallenge.h"
 #include "framework/Rcon.h"
 
@@ -53,7 +54,8 @@ cvar_t         *sv_privatePassword; // password for the privateClient slots
 cvar_t         *sv_allowDownload;
 cvar_t         *sv_maxclients;
 
-cvar_t         *sv_privateClients; // number of clients reserved for password
+Cvar::Range<Cvar::Cvar<int>> sv_privateClients("sv_privateClients",
+    "number of password-protected client slots", CVAR_SERVERINFO, 0, 0, MAX_CLIENTS);
 cvar_t         *sv_hostname;
 cvar_t         *sv_statsURL;
 cvar_t         *sv_master[ MAX_MASTER_SERVERS ]; // master server IP addresses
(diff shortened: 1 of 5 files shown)


illwieckz:
If server has 1 private slot (`-set sv_maxclients 1 -set sv_privateclients 1 -set sv_privatePassword "53CR37"`) and one private client is connected, other people see server (client count/max client count) as (0/0), is this expected? It means the server browser displays an empty server if a whole clan using private slots is playing on it.

I notice I still can connect the local server and taking a private slot without password or with wrong password if I connect either to my localhost address and my LAN address, is this expected? If yes, how can I easily test the various scenarii? I have not tested from a non-localhost or non-LAN address, it's just not easy to do so.

slipher:
@illwieckz I checked to make sure, and none of the issues you report occur when I test my patch. With privateclients=1 maxclients=1 and 1 player connected, it displays `1 + (0) / 1`. It is not possible to enter with the wrong password. All of these behaviors sound like you are using the old version.

src/engine/server/sv_main.cpp
```diff
@@ -632,9 +638,11 @@ void SVC_Info( netadr_t from, const Cmd::Args& args )
 	info_map["hostname"] = sv_hostname->string;
 	info_map["serverload"] = std::to_string( svs.serverLoad );
 	info_map["mapname"] = sv_mapname->string;
-	info_map["clients"] = std::to_string( count );
-	info_map["bots"] = std::to_string( botCount );
-	info_map["sv_maxclients"] = std::to_string( sv_maxclients->integer - sv_privateClients->integer );
+	info_map["clients"] = std::to_string( publicSlotHumans + privateSlotHumans );
+	info_map["bots"] = std::to_string( bots );
+    // Satisfies (number of open public slots) = (displayed max clients) - (number of clients).
```
DolceTriade: weird indent

src/engine/server/sv_main.cpp
```diff
@@ -571,21 +573,25 @@ void SVC_Info( netadr_t from, const Cmd::Args& args )
 
 	SV_ResolveMasterServers();
 
-	// don't count privateclients
-	int botCount = 0;
-	int count = 0;
+	int bots = 0; // Bots always use public slots.
+	int publicSlotHumans = 0;
```
DolceTriade: maybe use players instead of humans?

slipher: `Humans` does sound silly, but `Players` doesn't convey the idea of non-bot clients that I'm looking for.

DolceTriade: fair enough

illwieckz:
@slipher yes that was probably a mistake on my side, I rebuilt it to be sure and now it works as expected. that's good.

PR #57, opened 2017-11-19
Remove unused var and improve readme

HumanG33k:


Viech:
Thanks, I reworded your `README.md` change in a7aa4c2 and cherry-picked 61b983b into 8880609.

README.md
```diff
@@ -19,6 +19,7 @@ This repository contains the engine of the game Unvanquished. See below for buil
 ## Build Instructions
 
 Instead of `make`, you can use `make -jN` where `N` is your number of CPU cores to speed up compilation.
+Or on linux you can do `make -j$(getconf _NPROCESSORS_ONLN)`.
```
illwieckz: Why not just doing `make -j$(nproc)`?

Do `getconf _NPROCESSORS_ONLN` also works on Windows and/or MacOS (in this case it would be better than `nproc` despite being longer to type)?

HumanG33k: You are right. I don’t know about Win or Mac. I use it because i use always the same skeleton for this kind of things. I will update it :)

PR #60, opened 2017-12-09
Cmake fixes

slipher:
Pass more flags to the nacl vms build (such as build type and use pch), and fix the bug in pch generation introduced by using a generator expression for DEBUG_BUILD.


CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ if(NOT DAEMON_EXTERNAL_APP)
         set(BUILD_GAME_NACL_NEXE 0)
     endif()
     option(BUILD_TTY_CLIENT "Build Daemon headless client" 1)
-    option(BUILD_DUMMPY_APP "Build the dummy app for daemon" 0)
```
DolceTriade: lol

PR #62, opened 2017-12-18
Some key binding work

slipher:
Some more miscellaneous commits related to key binding. I'm hoping to get some review on some of this stuff before the diff becomes unmanageably large again. If so I recommend viewing the commits one at a time as they aren't closely related to each other.


src/engine/client/cl_keys.cpp
```diff
@@ -43,7 +43,7 @@ key up events are sent even if in console mode
 
 */
 
-bool key_overstrikeMode;
+static bool overstrikeModeOn;
 bool bindingsModified;
```
Kangz: Can this be made static too?

src/engine/client/keys.h
```diff
@@ -43,14 +43,19 @@ Maryland 20850 USA.
 #include "framework/ConsoleField.h"
 #include "qcommon/q_unicode.h"
 
-#define MAX_TEAMS 4
-#define DEFAULT_BINDING 0
+namespace Keyboard {
+enum BindTeam {
+    BIND_TEAM_DEFAULT = 0, // Default bind is used if there is no bind set for the player's current team
+    BIND_TEAM_SPECTATORS = 3,
+    MAX_TEAMS
```
Kangz: nit: NUM_TEAMS

src/engine/client/key_binding.cpp
```diff
@@ -269,6 +269,17 @@ void SetBinding(Key key, int team, std::string binding)
 
 // -ve team no. = don't return the default binding
```
Kangz: That can we push the boolean to the caller? That negative argument thing is pretty ew.

Kangz:
LGTM overall

src/engine/client/cl_keys.cpp
slipher: Well, it is used somewhere, in Com_WriteConfiguration.

slipher:
Thanks, I have fixed the issues other than the bindingsModified one.

Kangz:
LGTM thanks for the iterations

PR #63, opened 2017-12-29
Disable extension GL_ARB_texture_gather on NVidia GeForce.

gimhael:
Some versions of NVidias OpenGL driver seem to have a bug, and it was just a
performance optimization anyway.

This should fix Unvanquished issue #919.


illwieckz:
Will test !

illwieckz:
tested, it works

DolceTriade:
NIce!!

PR #70, opened 2018-02-11
import unity's improvements to crnlib

illwieckz:
See https://forums.unvanquished.net/viewtopic.php?f=4&t=2158
for a large forum post explaining what's happening.

Basically, Unity is shipping an improved crnlib within their own tools
and they litterally broke the format to make it better.

See https://github.com/Unity-Technologies/crunch/tree/unity
for the original unity's crunch code.

See https://github.com/illwieckz/crunch/tree/unity-daemon
for a temporary branch applying our improvements above unity's ones.

Using their tool on a corpus large by 1979 files from our asset repository
I reported that compression time was divided by more than 4 (reducing from
multiple hours to less than half one) and more than 11% file space saved
in comparison with the crnlib version we currently use (which is an older
version of the same source tree they use).

Since next release will use DPK format instead of PK3 one it's the very good time to
also break this kind of stuff since we already have to re-ship everything. Community mappers
gave permission to repackage and recompress their maps so we can do it
without losing anything: neither our stuff, neither third-party stuff.

See also https://forums.unvanquished.net/viewtopic.php?f=33&t=2153
for a forum post about asset repackaging for next release.

### DIFF
--- a/libs/crunch/crn_decomp.h
+++ b/libs/crunch/crn_decomp.h
@@ -1,345 +1,34 @@
 // File: crn_decomp.h - Fast CRN->DXTc texture transcoder header file library
-// Copyright (c) 2010-2012 Rich Geldreich and Tenacious Software LLC
+// Copyright (c) 2010-2016 Richard Geldreich, Jr. and Binomial LLC
 // See Copyright Notice and license at the end of this file.
 //
 // This single header file contains *all* of the code necessary to unpack .CRN files to raw DXTn bits.
 // It does NOT depend on the crn compression library.
 //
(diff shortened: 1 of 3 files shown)


illwieckz:
I rebuilt the whole unvanquished resources, textures sets and map packages using that new crunch, same with community maps I maintain and the ones from Matth and tvezet, everything looks fine once the engine is running the changes from that PR. It ran around an hour to process everything instead of half a day.

illwieckz:
To get your textures rendered fine with these changes you need to compress your files using this `crunch`:

[DaemonEngine/crunch](https://github.com/DaemonEngine/crunch)

To compress normal maps you must use the `-rtopmip` option with the `--renormalize` one.

PR #73, opened 2018-02-21
crypto: Support libnettle 3.4+ without warnings

DolceTriade:
libnettle 3.4 changed the base64 API to make the args
accept char instead of uint8_t. This causes compile time
warnings (and errors if you compile without -fpermissive),
so wrap the functions for newer versions of nettle to
silence those warnings.

### DIFF
--- a/src/engine/framework/Crypto.cpp
+++ b/src/engine/framework/Crypto.cpp
@@ -34,6 +34,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <nettle/base64.h>
 #include <nettle/md5.h>
 #include <nettle/sha2.h>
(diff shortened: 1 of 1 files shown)


slipher:
Can we drop support for versions less than 3? 3.0 was released in June 2014.

slipher:
It doesn't build with gcc on travis.

illwieckz:
is the travis failure meaningful? can we merge that or does it require some further changes?

illwieckz:
> Can we drop support for versions less than 3? 3.0 was released in June 2014.

Even Debian stable is shipping libnettle 3.3. :pencil:

illwieckz:
```
<illwieckz> `Ishq, what is your thought on this?
<Ishq> illwieckz: As for that PR, if its not building on travis, we should fix travis.
```

PR #75, opened 2018-02-26
New keyboard cgame ABI

slipher:
This updates the cgame ABI to use the new keyboard key representations.

See https://github.com/Unvanquished/Unvanquished/pull/1065 for the cgame side.

### DIFF
--- a/src/common/Serialize.h
+++ b/src/common/Serialize.h
@@ -31,7 +31,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef COMMON_SERIALIZATION_H_
 #define COMMON_SERIALIZATION_H_
 
+#include <limits>
+#include <map>
+#include <set>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
 #include "IPC/Common.h"
+#include "Optional.h"
+#include "String.h"
 
 namespace Util {
 
--- a/src/engine/client/cl_cgame.cpp
+++ b/src/engine/client/cl_cgame.cpp
@@ -49,11 +49,6 @@ Maryland 20850 USA.
 #define C__(x, y) Trans_PgettextGame(x, y)
 #define P__(x, y, c) Trans_GettextGamePlural(x, y, c)
 
-// NERVE - SMF
-void                   Key_GetBindingBuf( int keynum, int team, char *buf, int buflen );
-
-// -NERVE - SMF
-
 
 /*
 ====================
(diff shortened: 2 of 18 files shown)


src/engine/client/cl_cgame.cpp
```diff
@@ -1474,74 +1447,65 @@ void CGameVM::QVMSyscall(int index, Util::Reader& reader, IPC::Channel& channel)
...
+					result = Keyboard::ScancodeToAscii(scancode);
+				}
 			});
 			break;
 
 		case CG_KEY_SETBINDING:
-			IPC::HandleMsg<Key::SetBindingMsg>(channel, std::move(reader), [this] (int keyNum, int team, std::string cmd) {
-				Keyboard::SetBinding(Keyboard::Key::FromLegacyInt(keyNum), team, std::move(cmd));
+			IPC::HandleMsg<Keyboard::SetBindingMsg>(channel, std::move(reader), [this] (Keyboard::Key key, int team, std::string cmd) {
+				Keyboard::SetBinding(key, team, std::move(cmd));
```
Kangz: Is any validation necessary for key?

src/engine/client/cl_cgame.cpp
```diff
@@ -1474,74 +1447,65 @@ void CGameVM::QVMSyscall(int index, Util::Reader& reader, IPC::Channel& channel)
...
 				Key_ClearStates();
 			});
 			break;
 
 		case CG_KEY_KEYSDOWN:
-			IPC::HandleMsg<Key::KeysDownMsg>(channel, std::move(reader), [this] (std::vector<int> keys, std::vector<int>& list) {
+			IPC::HandleMsg<Keyboard::KeysDownMsg>(channel, std::move(reader), [this] (std::vector<Keyboard::Key> keys, std::vector<bool>& list) {
 				list.reserve(keys.size());
-				for (unsigned i = 0; i < keys.size(); ++i)
+				for (Keyboard::Key key : keys)
```
Kangz: ditto

src/common/Serialize.h
```diff
@@ -31,7 +31,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef COMMON_SERIALIZATION_H_
 #define COMMON_SERIALIZATION_H_
 
+#include <limits>
```
Kangz: These should already be included in common/Common.h no?

src/engine/client/key_binding.cpp
```diff
@@ -288,12 +288,15 @@ Util::optional<std::string> GetBinding(Key key, BindTeam team, bool useDefault)
...
+std::vector<Key> GetKeysBoundTo(int team, Str::StringRef command)
 {
+	if (!IsValidTeamNumber(team)) {
+		return {};
+	}
 	std::vector<Key> result;
 	for (const auto& key : keys) {
-		for (BindTeam team : {GetTeam(), BIND_TEAM_DEFAULT}) {
-			const auto& binding = key.second.binding[team];
+		for (BindTeam t: {Util::enum_cast<BindTeam>(team), BIND_TEAM_DEFAULT}) {
```
Kangz: Needs validation?

src/engine/client/cl_cgame.cpp
slipher: Good idea, I added some validation here as well as in the deserialization.

src/engine/client/cl_cgame.cpp
slipher: `IsValid` is called in `IsDown` (though maybe `IsNonnull` or something like that would be a more accurate name).

src/common/Serialize.h
slipher: Well, we don't seem to use the precompiled header in cgame or sgame, so I can't assume everything is included.

src/engine/client/key_binding.cpp
slipher: Of what?

slipher:
@Kangz does it look OK now?

PR #80, opened 2018-03-11
do not abort shader on “unknown general shader parameter” warning

illwieckz:
There is two possibilites:

- if it's a warning, it does not have to return;
- if it has to return, it's not a warning but an error.

After some testing it appeared it's a warning
so we just have to ignore the whole line and
let the parser read the next line.

It's very useful to ignore unknown dp* and xon* keywords
from darkplaces and xonotic when loading maps from them
for example.

It will help any project wanting to port existing game to
Dæmon to have a better partial support more quickly: this
kind of project would appreciate to be able to focusing on
implementing the ones required for gameplay while not having
any urge to implement the ones that are only needed for
eye candy.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -3684,7 +3684,8 @@ static bool ParseShader( const char *_text )
 		else
 		{
 			Log::Warn("unknown general shader parameter '%s' in '%s'", token, shader.name );
-			return false;
+			SkipRestOfLine( text );
+			continue;
 		}
 	}
 


illwieckz:
This was discussed at first in https://github.com/Unvanquished/Unvanquished/pull/966#issuecomment-253390514 but this specific change was not part of the PR (the PR merged another similar change in another place of code).

Before:

[Xonotic map on Dæmon](https://dl.illwieckz.net/b/xonotic/screenshots/experiments/daemon/unvanquished_2018-03-11_045429_000.jpg)

After:

[Xonotic map on Dæmon](https://dl.illwieckz.net/b/xonotic/screenshots/experiments/daemon/unvanquished_2018-03-11_045621_000.jpg)

Other screenshots to have a look:

[Xonotic map on Dæmon](https://dl.illwieckz.net/b/xonotic/screenshots/experiments/daemon/unvanquished_2018-03-11_051510_000.jpg)
[Xonotic map on Dæmon](https://dl.illwieckz.net/b/xonotic/screenshots/experiments/daemon/unvanquished_2018-03-11_063117_000.jpg)
[Xonotic map on Dæmon](https://dl.illwieckz.net/b/xonotic/screenshots/experiments/daemon/unvanquished_2018-03-11_063156_000.jpg)

Note that there is probably some other places with a bizarre `return` after a _warning_, but I prefer to wait to encounter real use cases to test against before changing the `return` to something else in case the fix is not to not return but to raises an _error_ instead of the _warning_.

In the precise case discusses in this PR, this _is_ a true _warning_ so the code does not have to return but just have to ignore the unknown stuff.

slipher:
LGTM

PR #81, opened 2018-03-11
support “dds/<path without ext>.dds” texture alternatives like Darkplaces or Doom3

illwieckz:
in this case, the `dds/<path without ext>.dds` alternative found must be
used as the best format even if engine usually prefers another one.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -1653,6 +1653,7 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
 
 	ext = COM_GetExtension( filename );
 
+	// the Daemon's default strategy is to use the hardcoded path if exists
 	if ( *ext )
 	{
 		// look for the correct loader and use it
@@ -1694,6 +1695,20 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
 
 	int bestLoader = -1;
 	const FS::PakInfo* bestPak = nullptr;
+
+	// Darkplaces or Doom3 packages can ship alternative texture path in the form of
+	//   dds/<path without ext>.dds
+	std::string altName = Str::Format("dds/%s.dds", filename);
+	bestPak = FS::PakPath::LocateFile(altName);
+
+	// If this alternative path exists, it's expected to be loaded as the best one
+	// except when it goes against Daemon's rule to load the hardcoded one if exists
+	// because this dds alternative is only supported for compatibility with
+	// third-party content
+	if ( bestPak != nullptr ) {
+		LoadDDS( altName.c_str(), pic, width, height, numLayers, numMips, bits, alphaByte );
+		return;
+	}
 
 	// try and find a suitable match using all the image formats supported
 	// prioritize with the pak priority


illwieckz:
See [darkplaces/gl_rmain.c](https://gitlab.com/xonotic/darkplaces/blob/adf31edb8c49c3f8e9ddd1293ef6b6c7254933bf/gl_rmain.c#L3421) as a reference implementation.

src/engine/renderer/tr_image.cpp
```diff
@@ -1702,6 +1703,22 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
 		std::string altName = Str::Format("%s.%s", filename, imageLoaders[i].ext);
 		const FS::PakInfo* pak = FS::PakPath::LocateFile(altName);
 
+		// Darkplaces or Doom3 packages can ship alternative texture path in the form of
+		//   dds/<path without ext>.dds
+		if ( pak == nullptr && !Q_stricmp("dds", imageLoaders[i].ext) )
```
slipher: No need to use `imageLoaders[i].ext` here since you already know it's dds.

src/engine/renderer/tr_image.cpp
```diff
@@ -1714,7 +1731,7 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
 
 	if ( bestLoader >= 0 )
 	{
-		char *altName = va( "%s.%s", filename, imageLoaders[ bestLoader ].ext );
+		char *altName = va( "%s%s.%s", prefix, filename, imageLoaders[ bestLoader ].ext );
```
slipher: I suggest storing a `bestFilename` instead of rebuilding the name again.

src/engine/renderer/tr_image.cpp
```diff
@@ -1702,6 +1703,22 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
...
 		const FS::PakInfo* pak = FS::PakPath::LocateFile(altName);
 
+		// Darkplaces or Doom3 packages can ship alternative texture path in the form of
+		//   dds/<path without ext>.dds
+		if ( pak == nullptr && !Q_stricmp("dds", imageLoaders[i].ext) )
+		{
+			altName = Str::Format("dds/%s.%s", filename, imageLoaders[i].ext);
+			pak = FS::PakPath::LocateFile(altName);
+
+			// If this alternative path exists, it's expected to be loaded as the best one
```
slipher: The `dds/` path will not be tried if there is a .dds in the regular path, which seems inconsistent with this comment. If the dds/ one should always be highest priority, I think it would be better to move the check before this loop.

src/engine/renderer/tr_image.cpp
illwieckz: we don't have to look for `dds/` prefixes for other formats.

src/engine/renderer/tr_image.cpp
illwieckz: but doing it outside the loop makes sense :-)

src/engine/renderer/tr_image.cpp
slipher: Oops, I meant to write that on the `altName = ...` line.

illwieckz: This whole part is now obsoleted by last commit

src/engine/renderer/tr_image.cpp
```diff
@@ -1653,6 +1653,7 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
 
 	ext = COM_GetExtension( filename );
 
+	// the Daemon's default strategy is to use the hardcoded path if exists
```
illwieckz: Note that I don't like this strategy but I remember people prefer this. Basically I don't like the behaviour of loading the hardcoded path if exists because it means a pak shipping `.crn` files alternatives to `.tga` ones will never seen the `.crn` files used if an older pak has the `.tga` one, defeating the idea of « shipping better formats », but when I asked to remove that part and only rely on a « best strategy », some people had strong arguments at this time. Basically, id Software did the exact opposite thing with `dds` files for Doom3, and Xonotic guys did the exact opposite thing with Darkplaces (hence the current patch).

It was told people can hardcode stuff if they want to force a format, and use extension independent path if they want to never force it. But in case of some uncommon (or now uncommon) formats like md3, it means editing a binary file to make the path ext independent.

PR #84, opened 2018-03-11
add support for DDS DXT3, fix #82

### LINKED ISSUE #82: R_LoadDDSImage: unsupported FOURCC 0x33545844
Dæmon fails to load some Xonotic textures, here is an example:

```
Warn: R_LoadDDSImage: unsupported FOURCC 0x33545844, "dds/models/desertfactory/textures/misc/grass01.dds" 
```

You can find the file there: [grass01.dds](http://dl.illwieckz.net/b/daemon/bugs/dds-invalid-fourcc/grass01.dds)

Note that `crunch` is able to convert this `dds` file back to `tga` without complaining so perhaps we can rely on `crnlib` for `dds` files too?

illwieckz:
DXT5 and DXT3 formats are very close (see that [documentation](https://www.fsdeveloper.com/wiki/index.php?title=DXT_compression_explained)).

The define for DXT3 was already there and code path for DXT5 too hence this code just copypaste the DXT5 switch for DXT3.

DXT3 `dds` files loaded this way look to be rendered the exact same way they would be rendered if they were converted to `tga` by `crunch` at first.

Xonotic has textures using this format.

- Before:
[DDS DXT3](https://dl.illwieckz.net/b/daemon/bugs/dds-invalid-fourcc/unvanquished_2018-03-11_191901_000.jpg)
- After:
[DDS DXT3](https://dl.illwieckz.net/b/daemon/bugs/dds-invalid-fourcc/unvanquished_2018-03-11_191959_000.jpg)

- Before:
[DDS DXT3](https://dl.illwieckz.net/b/daemon/bugs/dds-invalid-fourcc/unvanquished_2018-03-11_192118_000.jpg)
- After:
[DDS DXT3](https://dl.illwieckz.net/b/daemon/bugs/dds-invalid-fourcc/unvanquished_2018-03-11_192317_000.jpg)

We see in this last texture that transparency is handled. Transparency is probably badly rendered (or not rendered at all on first textures) because of unknown darkplaces-specific shader keywords, if I convert those `dds` textures to `tga` ones it does not look better, so it's not a format parsing issue.

### DIFF
--- a/src/engine/renderer/tr_image_dds.cpp
+++ b/src/engine/renderer/tr_image_dds.cpp
@@ -277,6 +277,11 @@ void R_LoadDDSImageData( void *pImageData, const char *name, byte **data,
 			blockSize = 8;
 			break;
 
+		case FOURCC_DXT3:
+			*bits |= IF_BC3;
+			blockSize = 16;
+			break;
+
 		case FOURCC_DXT5:
 			*bits |= IF_BC3;
 			blockSize = 16;


illwieckz:
Note that the current code defines both `FOURCC_DXT2` and `FOURCC_DXT4` but there is no related switches, and `DXT1A` is not defined at all.

For information this is some extra defines the current code does not know at all (see `crunch`'s [`dds_defs.h`](https://github.com/DaemonEngine/crunch/blob/c275f974bc261003a6c0191bf1b3fbdcb485bfe3/inc/dds_defs.h#L23-L38)):

```c++
// Non-standard, crnlib-specific pixel formats (some of these are supported by ATI's Compressonator)
#define FOURCC_DXT5_CCxY        MAKEFOURCC('C', 'C', 'x', 'Y')
#define FOURCC_DXT5_xGxR        MAKEFOURCC('x', 'G', 'x', 'R')
#define FOURCC_DXT5_xGBR        MAKEFOURCC('x', 'G', 'B', 'R')
#define FOURCC_DXT5_AGBR        MAKEFOURCC('A', 'G', 'B', 'R')

#define FOURCC_DXT1A            MAKEFOURCC('D', 'X', '1', 'A')
#define FOURCC_ETC1             MAKEFOURCC('E', 'T', 'C', '1')
#define FOURCC_ETC2             MAKEFOURCC('E', 'T', 'C', '2')
#define FOURCC_ETC2A            MAKEFOURCC('E', 'T', '2', 'A')

#define FOURCC_R8G8B8           MAKEFOURCC('R', 'G', 'B', 'x')
#define FOURCC_L8               MAKEFOURCC('L', 'x', 'x', 'x')
#define FOURCC_A8               MAKEFOURCC('x', 'x', 'x', 'A')
#define FOURCC_A8L8             MAKEFOURCC('L', 'x', 'x', 'A')
#define FOURCC_A8R8G8B8         MAKEFOURCC('R', 'G', 'B', 'A')
```

Viech:
The lighting in your second "after" image looks weird and in the first "after" iamge, apart from transparency not working for the decals as you mentioned, that logo is glowing like it's nuclear. Are you sure this is a shader issue? Maybe both crunch as well as our engine are handling something badly here.

illwieckz:
Xonotic uses sRGB lightmap and Dæmon does not handle this kind of stuff yet so it's expected that lighting looks weird: the lightmap color curve is the wrong one.

On the shader side, all these shaders makes an heavy usage of darkplaces-specific keywords that are unknown by Dæmon hence ignored, so basically diffuse textures are rendered without any processing.

Also Darkplaces offers some keyword we don't have equivalent for yet (for example the the `dpglossintensitymod` keyword allows per-shader specular scale when Dæmon seems to only implement global specular scale option via a cvar.

Then Darkplaces does an heavy implicit mapping process: it loads as normal maps every <basename>_norm.<anyext> file if <basesename>.<anyext> is loaded as diffuse, with similar mechanism for specular, gloss and other things like that. This makes sense in Darkplaces since it's an engine to improve existing assets from existing games: modders just has to add new maps in extra pk3 but does not have to modify the shader (avoiding some legal issues). Xonotic uses that feature a lot, so basically only diffuse map are loaded at this time.

At this early state it's expected Dæmon rendering Xonotic map does not produces much better pictures than netradiant itself produces, Xonotic maps rendered in Xonotic basically looks like netradiant rendering plus a wrong lighting applied.

For reference, this is how looks these scenes on Darkplaces:

- Erbium map on Darkplaces:  
[Xonotic Erbium map on Darkplaces](https://dl.illwieckz.net/b/daemon/bugs/dds-invalid-fourcc/xonotic20180312190456-00.jpg)
- Solarium map on Darkplaces:  
[Xonotic Solarium map on Darkplaces](https://dl.illwieckz.net/b/daemon/bugs/dds-invalid-fourcc/xonotic20180312185055-00.jpg)

You'll notice the shiny water effect on the warp zone (we don't have neither a water effect like this neither warp zone) on the first picture, and you'll notice the heavy parallax mapping on the second one (something we don't have in a usable state yet) plus the destructible models. :wink:

And since I said rendering Xonotic maps on Dæmon would looks like a netradiant render with a wrong lighting, have a look at that interesting bug: :blush: 

- On Dæmon:  
[Xonotic Erbium map](https://dl.illwieckz.net/b/daemon/wip/texturing/unvanquished_2018-03-12_190807_000.jpg)
- On Darkplaces:  
[Xonotic Erbium map](https://dl.illwieckz.net/b/daemon/wip/texturing/xonotic20180312190745-00.jpg)

- On Dæmon:  
[Xonotic Erbium map](https://dl.illwieckz.net/b/daemon/wip/texturing/unvanquished_2018-03-12_190940_000.jpg)
- On Darkplaces:  
[Xonotic Erbium map](https://dl.illwieckz.net/b/daemon/wip/texturing/xonotic20180312190946-00.jpg)

Btw for this one Daemon say the shader format is invalid, so I really don't know why the `qer_editorimage` is loaded instead of the diffuse map one, but it's very possible the engine is not broken, at least the engine says it's not its fault. :stuck_out_tongue:

illwieckz:
I'm closing this since #82 seems to obsolete it.

PR #86, opened 2018-03-15
DXT3/BC2 support for .dds, .crn and .ktx file formats

gimhael:
This should enable DXT3/BC2 support for the daemon engine (see issue #82 ).


slipher:
LGTM

illwieckz:
I tested it and I don't see any issue that can be related to that code.

Issue #76, opened 2018-03-03
pure white line artifact on edges


illwieckz:
This is a weird behavior we can see on some different use cases:

- lines because two edges are rendered very near even if they are on different depth, note that in some condition (and angle) the lines can be red or other colors:

[![white edge artifact](https://dl.illwieckz.net/b/unvanquished/bugs/white-edge-artifact/unvanquished_2017-09-28_002547_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/white-edge-artifact/unvanquished_2017-09-28_002547_000.jpg)
[![white edge artifact](https://dl.illwieckz.net/b/unvanquished/bugs/white-edge-artifact/unvanquished_2017-10-18_233510_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/white-edge-artifact/unvanquished_2017-10-18_233510_000.jpg)

- lines on surfaces junction:

[![white edge artifact](https://dl.illwieckz.net/b/unvanquished/bugs/white-edge-artifact/unvanquished_2018-02-06_231004_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/white-edge-artifact/unvanquished_2018-02-06_231004_000.jpg)
[![white edge artifact](https://dl.illwieckz.net/b/unvanquished/bugs/white-edge-artifact/unvanquished_2018-02-11_233456_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/white-edge-artifact/unvanquished_2018-02-11_233456_000.jpg)

illwieckz:
Disabling deluxeMaps makes that bug disappear.

illwieckz:
A good information to know is that lower resolution lightmaps do not trigger the bug. I discovered this after having rebuild lower-res lightmaps from vega and chasm.

Viech:
Have you maybe compressed the deluxemaps? They should be as sensitive as lightmaps towards compression, if not more, since they are not to be understood as a visual image but as a vector map that stores the direction of incoming light. In particular deluxemaps are used for both normal mapping and specular reflections, and it could be that corrupted deluxe vectors give you a sort of specular super-reflection that isn't quite based on your view angle anymore.

illwieckz:
[...]

Issue #68, opened 2018-02-10
weird patterns on translucent textures


illwieckz:
This bug affects the for-0.51.0 branch, I cannot reproduce with a 0.50.0 build and I haven't tried master.

Look at these screenshots of the `hangar28` maps by _tvezet_ and the `thunder` one by _EmperorJack_:

[![wrong translucency hangar28](https://dl.illwieckz.net/b/unvanquished/bugs/wrong-translucency/unvanquished_2018-02-09_231104_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/wrong-translucency/unvanquished_2018-02-09_231104_000.jpg)
[![wrong translucency thunder](https://dl.illwieckz.net/b/unvanquished/bugs/wrong-translucency/unvanquished_2018-02-10_012544_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/wrong-translucency/unvanquished_2018-02-10_012544_000.jpg)

The translucency looks obviously wrong. That's not a compression fault (I have not recompressed the thunder's textures and it's rendered right on 0.50.0 but rendered wrong on for-0.51.0).

These is the two related shaders:

* The one used in `thunder` map, from `map-thunder` package:

```
textures/thunder/nexusglass
{
	qer_editorimage textures/thunder/nexusglass
	qer_trans .40
	surfaceparm trans
	cull none
	{
		map textures/thunder/nexusglass
		blendfunc add
		rgbGen const ( 0.01 0.01 0.01 )
	}
	{
		map $lightmap 
		blendfunc gl_dst_color gl_src_alpha
		rgbGen identity
		tcGen lightmap 
	}
}
```

* The one used in `hangar28` from `tex-trak5` package:

```
textures/shared_trak5/glass
{
	qer_editorImage     textures/shared_trak5_src/glass_p
	qer_trans           0.30

	cull                none
	surfaceparm         trans

	{
		map       textures/shared_trak5_src/glass_d
		stage     diffuseMap
		blend     blend
	}
}
```

illwieckz:
this is very likely to be a lightmap issue

illwieckz:
it's really awful:
[![translucency bug](https://dl.illwieckz.net/b/unvanquished/bugs/wrong-translucency/unvanquished_2018-02-12_013656_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/wrong-translucency/unvanquished_2018-02-12_013656_000.jpg)

[...]

PR #87, opened 2018-03-19
Try to fix the bug with some lightmaps on translucent textures.

slipher:
I bisected #68 and found https://github.com/Unvanquished/Unvanquished/commit/833ae67bd157c818eda18791ea8d858b1b9db7ac as the cause. This is a commit that fixes the issues with the particular texture on Thunder. It's probably completely wrong though, so @gimhael should feel free to write his own fix.


illwieckz:
I confirm this fixes the #68 issue for the bug in thunder, station15 and hangar28. It does not fix #76 which is now clearly another issue.

PR #88, opened 2018-03-20
Disable the clang errors on 'this' lambda captures.

slipher:


DolceTriade:
I kinda feel like we should remove the unnecessary captures...

slipher:
> I kinda feel like we should remove the unnecessary captures...

Are you volunteering? :stuck_out_tongue:

DolceTriade:
Eventually.

slipher:
Well I am just going to merge it now so we can use the CI.

PR #95, opened 2018-04-15
Clean up cl_consolekeys cvar

slipher:


src/common/Cvar.h
```diff
@@ -169,22 +169,22 @@ namespace Cvar {
...
+     * The modified flag is set to true on construction (before the cvar has been read the first time).
      */
 
     template<typename Base> class Modified : public Base {
         public:
             using value_type = typename Base::value_type;
 
             template<typename ... Args>
-            Modified(std::string name, std::string description, int flags, value_type defaultValue, Args ... args);
+            Modified(std::string name, std::string description, int flags, value_type defaultValue, Args&& ... args);
```
DolceTriade: I wonder why these aren't passed by reference or Str::StringRefs.

Kangz: It was before StringRef and std::moved inside I guess?

slipher: In cases (such as this one) where the string is going to eventually be stored somewhere,, it can be more efficient to pass by value since if you have an rvalue at the beginning, it can be moved all the way down. In this particular case though it's probably always going to be constructed from a string literal, so it wouldn't make much difference either way; only one std::string would ever be constructed.

PR #97, opened 2018-04-21
Fix crash from #94.

slipher:
Also an unhandled exception in the "math" command.


illwieckz:
I confirm it fixes the crash, thank you! :-)

PR #98, opened 2018-04-21
pass DAEMON_DIR back to nacl vm build

illwieckz:
for some reason some nacl vm related build steps are driven by the engine cmake
configuration, hence when the engine is not built as a submodule, the cmake script
has to tell the vm build process where the engine source tree is.


slipher:
It's a bit odd that DAEMON_DIR is referenced here when it is an option defined in the Unvanquished project, but I'd say go ahead with this since it would be way too painful to do something else.

The reason for the nacl build weirdness is that cmake does not let you define more than one toolchain at once, so the nacl build has to be done in a separate instance.

PR #106, opened 2018-05-13
A couple small things

slipher:


src/engine/sys/sdl_input.cpp
```diff
@@ -564,7 +570,7 @@ static void IN_InitJoystick()
 
 	for ( i = 0; i < total; i++ )
 	{
-		Log::Debug( "[%d] %s", i, SDL_JoystickNameForIndex( i ) );
+		Log::Warn( "[%d] %s", i, JoystickNameForIndex( i ) );
```
DolceTriade: Back to debug?

slipher: Right

PR #110, opened 2018-05-20
Memset

slipher:


DolceTriade:
Nice cleanup!

src/engine/renderer/tr_shader.cpp
```diff
@@ -3706,236 +3706,187 @@ CollapseMultitexture
...
 {
-	int           i, j;
-	int           numStages = 0;
-
 	if ( !r_collapseStages->integer )
 	{
 		return;
 	}
 
-	shaderStage_t tmpStages[ MAX_SHADER_STAGES ]{};
```
DolceTriade: hwo come 
```
shaderStage_t tmpStages[ MAX_SHADER_STAGES ] = {};
```

doesn't work?

slipher: I tried that but still got the same looking backtrace.

PR #112, opened 2018-06-20
add a second master server, rework the defines

illwieckz:
- add a second master server
- retrieve motd from first working master server
- extend the `globalservers` command to query every master server at once
- always resolve master server to not have to restart servers when master ip changes
- deduplicate server lists from multiple master servers
- drop “send gamestats to master server” support, see Unvanquished/Unvanquished#1083


slipher:
LGTM

I believe that this will cause *servers* to report to both masters, but that the client lacks any logic for trying alternative servers. The user would still have to type a command to get servers from a master other than the 1st one.

src/common/Defs.h
```diff
@@ -46,7 +46,13 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define GAMENAME_STRING     "unv"
 #define GAMENAME_FOR_MASTER PRODUCT_NAME_UPPER
 
-#define MASTER_SERVER_NAME  "master.unvanquished.net"
+#define MASTER1_SERVER_NAME "master.unvanquished.net"
+#define MASTER2_SERVER_NAME "master2.unvanquished.net"
+#define MASTER3_SERVER_NAME ""
+#define MASTER4_SERVER_NAME ""
+#define MASTER5_SERVER_NAME ""
+#define MOTD_MASTER_SERVER_NAME     MASTER1_SERVER_NAME
```
DolceTriade: Doesn't this mean that if master 1 goes down, these two won't work. I'd rather move back to a cvar scheme...

illwieckz:
I updated the PR to make the server sending gamestat to every master server and retrieving motd from first working master server.

src/common/Defs.h
illwieckz: This was already a case and worst than that, it was using the hardcoded first master server even if the first master server cvar was updated. There was a similar behavior for the gamestat sending (sending to the first hardcoded master server only). See my new commits attempting to fix that.

illwieckz:
A better client implementation would be to merge the server lists (like a server browser like [XQF](http://github.com/XQF/xqf) does). Basically we would have to add to the internal server list any server found in a list from a master server that was not already added to the internal list before.

illwieckz:
Extending the `globalservers` command to allow something like that would be good:

```
globalservers * 86 full empty
```

illwieckz:
I just added a commit to do that btw I don't know yet where we have to deduplicate the received lists.

illwieckz:
If I'm reading things the right way there is already some deduplication for both packets and addresses:
https://github.com/DaemonEngine/Daemon/blob/3176f14586519acdabf626fca9d43b4373efce04/src/engine/client/cl_main.cpp#L2429-L2440

illwieckz:
Also, if I read it right if two master servers reply the same list of servers in the same order a false-positive warning about duplicate packet can occurs. This is probably OK.

slipher:
What is MOTD, is it even used ever? Can we delete it?

slipher:
The global stats thing seems deletable too. It's a fairly pointless idea since it is so easy to vandalize.

illwieckz:
> is it even used ever? Can we delete it?

I like this kind of question! :grin:

illwieckz:
what is globalstats for? what does it do?

slipher:
The game stats thing is mostly just the same information as the server status, plus which team won the game. So I guess gamestats was added in Tremulous for making the statistics on global humans win vs. aliens vs. draw rate.

illwieckz:
I see, a kind of cheap and unreliable http://stats.xonotic.org/ ?
In this case I'm OK to drop it: we have to do something better in either way

slipher:
We used to have something like that for the official server (with stats per player per weapon, etc.). Based on just parsing the game log I guess? Anyway, the information sent to the global stats server is very minimal - you only get a couple graphs like average number of players or team win percentages. Probably not too meaningful at the moment since most games are played by bots.

illwieckz:
I removed the Windows 95 workaround for name resolution, hence the master servers are now resolved when needed. It fixes Unvanquished/Unvanquished#1082.

illwieckz:
I discovered that

- server deduplication only worked on already known servers, not on multiple responses from server
- packet deduplication was just deduplicating “_n_-th packet from any master server” whatever the content, meaning the first packet from first master server will make the engine drops the  first packets from all other masters…

So I removed that useless packet deduplication and added some code to deduplicate servers themselve.

slipher:
You can remove the "send gamestat every master server" commit since we are deleting gamestats now.

slipher:
Why should we care about detecting duplicate master servers? I don't see the point.

DolceTriade:
Master MOTD is not used because it wasn't carried over somehow when we redid our UI with librocket (whoops...). It's still set on the master server. I was going to use it to tell people to upgrade unv lol

illwieckz:
This PR now drops support for gamestats. See Unvanquished/Unvanquished#1083

illwieckz:
@DolceTriade you initially approved this PR but I added a lot of things since. Can you review it again?

Do we drop MOTD or not?

DolceTriade:
ya, we drop it.

illwieckz:
I added a commit to drop motd support.

To announce new versions it would be better to have a code that compares the current running version number with the one from servers in server browser and when connecting to them, something saying _“hey, this server is in the future, do you wanna jump forward???_”

illwieckz:
Oh and btw, the _GetNews_ thing is more a game thing than an engine one.

DolceTriade:
It's an engine thing because with QVMs, the game had no way to fetch external data. However, now that we can compile curl directly into the game/cgame, it can be a game thing.

illwieckz:
I see.

Issue #102, opened 2018-04-29
Proposal: Log spam suppression


slipher:
This is a proposed mechanism to avoid the console being swamped by spammy log messages (for example if one starts printing every frame). The idea is that a log message be suppressed if it uses a format string used more than (approximately) Q times in the last T milliseconds.

Log filtering is done at the module level (i.e. one each for engine, cgame, sgame). A buffer contains records of the N (say 50) most recent unique format strings used for log messages. Besides the format string, the record contains an initial time (for the T ms period), a most recently seen time, number of times seen, and whether the message has been disabled. When a log message is received, the record for it is updated if it exists; otherwise the least recently seen message is removed. When updating a record, if T milliseconds have passed since the initial time, the initial time and count are reset. If the count reaches Q, then the message is disabled, and remains disabled until it drops out of the cache. A message like "Suppressing further messages of this type" is printed.

Naturally, there will be a cvar to disable log suppression (if you really want to see that debug message printed every frame). Log suppression should not used for messages printed by commands as they often print lengthy lists, and are usually triggered by the user's request. For messages that are known to legitimately print a lot of times, such as the list of loaded paks, there can be an API to log while bypassing the filter.

DolceTriade:
Hmm, I would prefer spammy logs police themselves. Maybe something similar to glog's LOG_EVERY_N (https://github.com/google/glog/blob/master/src/glog/logging.h.in#L176). I almost want to switch to glog, but that seems like a lot of work so :-/

slipher:
Using glog would make it slightly easier to write individual rate-limited log statements, but it doesn't do anything to help with the thousands of log statements that already exist. It's infeasible to try to figure out an appropriate limit for all of them. And if we have a limit by default, it will be easier in the future for developers to write log warnings, since they will not have to worry about excessive log spam.

DolceTriade:
Fair enough. Another option would be to allow call site specific filtering. This would require us to make our log functions macros though, but I think it is possible to do this in a way that doesn't require us to rewrite every log statement.

slipher:
I'm going to commit the stuff not directly involving log filtering so the PR does not accumulate too much cruft.

slipher:
Implemented in #115.

PR #115, opened 2018-07-12
Log spam suppression

slipher:
Implements #102, suppression of log messages that are printed an excessive number of times.

Messages printed by commands are exempt from suppression (assuming they used the new command API, though there are many unmigrated ones).

~~TODO: Add cvar for disabling suppression.~~
~~TODO: Add Logger API for printing without suppression and use it in any cases where we expect a lot of messages to be printed intentionally.~~


src/common/Log.cpp
```diff
@@ -85,27 +85,94 @@ namespace Log {
...
+                // Replace oldest entry if there is not already one for this message
+                oldest->intervalStartTime = now;
+                oldest->messageFormat = messageFormat;
+                oldest->numOccurrences = 1;
+                return OK;
+            }
+        };
+    } // namespace
+
+    void DispatchWithSuppression(std::string message, Log::Level level, Str::StringRef format) {
```
DolceTriade: Please add a way to disable this. Sometimes we want to be spammed so we can post process the data.

src/common/Log.cpp
```diff
@@ -85,27 +85,94 @@ namespace Log {
...
-    void CodeSourceDebug(std::string message) {
-        Log::Dispatch({"^5Debug: " + message}, debugTargets);
+    namespace {
+        // Log-spam suppression: if more than MAX_OCCURRENCES log messages with the same format string
+        // are sent in less than INTERVAL_MS milliseconds, they will stop being printed.
+        // (Unless the spammy message desists long enough to be flushed out of the buffer: then it can
+        // be printed again.)
+        class LogSpamSuppressor {
+            static constexpr int INTERVAL_MS = 2000;
+            static constexpr int MAX_OCCURRENCES = 10;
```
DolceTriade: Seem like values meant for cvars?

src/common/Log.cpp
```diff
@@ -85,27 +85,94 @@ namespace Log {
...
 
-    static const int debugTargets = (1 << GRAPHICAL_CONSOLE) | (1 << TTY_CONSOLE);
-    void CodeSourceDebug(std::string message) {
-        Log::Dispatch({"^5Debug: " + message}, debugTargets);
+    namespace {
+        // Log-spam suppression: if more than MAX_OCCURRENCES log messages with the same format string
+        // are sent in less than INTERVAL_MS milliseconds, they will stop being printed.
+        // (Unless the spammy message desists long enough to be flushed out of the buffer: then it can
+        // be printed again.)
+        class LogSpamSuppressor {
```
DolceTriade: So for logs from VMs, There will be no suppression for messages sent within the same frame? (Admittedly, this is pretty useless though.

slipher:
I replaced all the instances of `Log::Notice(%s, str)` and similar with `Log::Notice(str)` to address the issue of having an uninformative format string with those.

src/common/Log.cpp
slipher: Huh? Why wouldn't it work on logs within the same frame?

slipher:
I'm going to commit the stuff not directly involving log filtering so the PR does not accumulate too much cruft.

src/common/Log.cpp
DolceTriade: nvm

slipher:
Well I added cvars controlling the log suppression parameters. It's kind of ugly though.

slipher:
This should be more or less done now.

src/common/FileSystem.cpp
```diff
@@ -1058,15 +1058,15 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 
 	if (pak.type == pakType_t::PAK_ZIP) {
 		if (!isLegacy) {
-			fsLogs.Notice("Loading pak '%s'...", pak.path.c_str());
+			fsLogs.WithoutSuppression().Notice("Loading pak '%s'...", pak.path.c_str());
```
DolceTriade: nit: I prefer "ForcePrint()" or something to denote what it will do and not what it will not do.

src/common/Log.cpp
DolceTriade: Has this been addressed?

src/common/FileSystem.cpp
slipher: I'm not a fan of this suggestion since ForcePrint sounds like it should actually print something, while what the WithoutSuppression function actually does is only returning an altered logger object. Also, the mention of "suppression" makes it clear that it is related to the other functions with "suppression" in the name.

DolceTriade: sg

PR #117, opened 2018-07-28
Require a -uri switch before a command line URI.

slipher:


illwieckz:
I have a mixed feeling about that. On one hand I clearly prefer being able to pass only one argument (the uri) like other engines do (etlegacy etc.), on the other hand the uri thing something that is usually configured by the installer (via desktop file on linux or registry keys on windows), hence we can virtually do everything we want.

And I have a secret hope for having one day a standard common protocol that would be kiss enough to be implemented by multiple games, something like dgp://unvanquished/server/gg.illwieckz.net:27960/team/aliens (the first option being the game name a generic launcher can use to launch the game or any compatible game can use to redirect to the right game thanks to a common libdgp).

slipher:
It would be easy enough to support both ways (with and without `-uri`). At this time though, there seems to be little point in having multiple ways to do the same thing.

illwieckz:
why not just walking the command line and doing `s|unv://|+connect |` before parsing it?

slipher:
It's impossible to escape the URL, so the URL switch requires special treatment to ignore all following parameters.

DolceTriade:
I'm fine with the general concept, but I'd prefer we call it -connect or something to make it clear what the action is.

slipher:
Renamed to -connect.

illwieckz:
This whole thread is surrealist and I feel very weird reading it. There is already a `+connect` command.

If we gonna make an extra option a requirement for url call, we just have to make `+connect` droping the `unv://` prefix if exists.

If we wanna make `unv://…` url callable without extra option, and if we can't properly sanitize it, we can just define that this behavior is only supported when the url is the very last argument.

illwieckz:
Please note that I don't think it's a bad idea by itself to make the `connect` command en engine command, I just find bizarre to reinvent the wheel.

illwieckz:
I was talking about user interaction. I don't mind if code is larger if usage is meaningful for people. After all computers are just machines made to do complex tasks to makes human life easier, same about software. I like the idea that code purity comes after usability because the whole point of a code is to be used.

slipher:
Well, this `-connect` is primarily useful as a machine interface, to be used in protocol handlers. Users can still put `+connect` in the command line if they want to (and it will work as expected anywhere, instead of causing arguments to be ignored if they are at the end).

PR #124, opened 2018-10-01
Minor fixes

maek:


src/engine/botlib/nav.h
```diff
@@ -172,7 +172,7 @@ struct FastLZCompressor : public dtTileCacheCompressor
 							   unsigned char *compressed, const int /*maxCompressedSize*/, int *compressedSize )
 	{
 
-		*compressedSize = fastlz_compress( ( const void *const ) buffer, bufferSize, compressed );
+		*compressedSize = fastlz_compress( ( void * ) buffer, bufferSize, compressed );
```
slipher: `static_cast<const void*>` would be better

src/engine/client/key_binding.cpp
```diff
@@ -75,7 +75,6 @@ static const struct
 	{ "super", 5, 32, K_SUPER },
 	{ "compose", 6, 64, K_COMPOSE },
 	{ "menu", 7, 128, K_MENU },
-	{ "" }
```
slipher: This can't be removed as the zero array element is used in the loop condition `modifierKeys[ i ].bit`. Instead, you could change it from `{ "" }` to `{}`.

src/engine/botlib/nav.h
maek: I think the `const` shouldn't be required there (i.e. do you cast to `const` the buffers you pass as _src_ to `memcpy`?). Anyway, I added it back again as in this specific case we are doing a cast and we don't want to remove the const-ness as that's not our intent.

src/engine/client/key_binding.cpp
maek: My bad, I didn't realize it.

PR #125, opened 2018-10-02
make crunch a submodule

illwieckz:
it requires https://github.com/DaemonEngine/crunch/pull/7 to be merged first


slipher:
Can we call the directory something other than `crunch`? Having a submodule in the same place as a formerly non-submodule directory always screws everything up when you try to check out old commits.

illwieckz:
Yeah, those things are annoying btw it would be weird to not name crunch crunch. :-/

@slipher since you wrote this in https://github.com/DaemonEngine/crunch/pull/8 
> To move the exception throwing hack into Daemon.

is there something more to do on that side?

slipher:
Maybe we could have different directories for submodule and non-submodule libs? (Eventually; it's probably too much work to move everything from `libs` now). Then at least we wouldn't have to think of an arbitrary distinct name when something becomes a submodule.

PR #130, opened 2018-10-24
define dtAssert and rcAssert as DAEMON_ASSERT

illwieckz:
The idea is to define `dtAssert` and `rcAssert` as `DAEMON_ASSERT` and to patch Detour/Recast to reuse those definitions if already defined. If it works and if it's OK, I'll try to upstream the Detour/Recast patch so we can drop our custom _recastnavigation_ fork.

See https://github.com/Unvanquished/recastnavigation/pull/1

### DIFF
--- a/src/common/Assert.h
+++ b/src/common/Assert.h
@@ -148,6 +148,12 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
         Str::Format("\"%s >= %s\" bound: %s, actual: %s", #a, #b, AssertDetail::Printable(a), AssertDetail::Printable(b)) \
     );
 
+/*
+ * Reuse DAEMON_ASSERT in Detour and Recast
+ */
+#define dtAssert DAEMON_ASSERT
+#define rcAssert DAEMON_ASSERT
+
 #if !defined(DAEMON_SKIP_ASSERT_SHORTHANDS)
     #define ASSERT DAEMON_ASSERT
     #define ASSERT_EQ DAEMON_ASSERT_EQ


slipher:
Defining the assert this way won't work. It worked for crunch decompression because it was a header-only library. Recast has its own .cpp files, each one of which must have the assert definition supplied somehow when it is built.

The normal way to add a #define without editing the code is to use compiler flags. But this is also not so simple to apply since DAEMON_ASSERT will not be defined without including Daemon headers.

cmf028:
The latest recast/detour allow the setting of custom failure functions (used internally by dtAssert/rtAssert). We should probably use those.

See `rcAssertFailSetCustom` and `dtAssertFailSetCustom`.

illwieckz:
oh nice !

so do I just have to do something like that:
```c
#define dtAssertFailSetCustom DAEMON_ASSERT
#define rcAssertFailSetCustom DAEMON_ASSERT
```

and rebase `recastnavigation` on upstream?

or is there more to do?

cmf028:
A bit different than that.

Take a look at [DetourAssert.h](https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Include/DetourAssert.h) and [RecastAssert.h](https://github.com/recastnavigation/recastnavigation/blob/master/Recast/Include/RecastAssert.h).

You should:

1. Create a function with signature `void (dtAssertFailFunc)(const char* expression, const char* file, int line);`
2. Have that function use `DAEMON_ASSERT_CALLSITE_HELPER` from [Assert.h](https://github.com/DaemonEngine/Daemon/blob/74262f043e3668970fb8441d1fa1781478729639/src/common/Assert.h) using the function's arguments.
3. Call `dtAssertFailSetCustom` from `BotSetupNav` in [bot_load.cpp](https://github.com/DaemonEngine/Daemon/blob/master/src/engine/botlib/bot_load.cpp) using the function you created as the argument.

illwieckz:
I'm not sure to really understand how to do it ^_^

illwieckz:
closing it as I don't know yet how to do it, no need to keep it open forever.

@cmf028 do you know how to do it?

PR #133, opened 2018-10-28
Specify minimum C++ compiler versions.

### LINKED ISSUE #114: Define supported compiler versions
We should define which minimum compiler versions are supported, so we know which workarounds for old compilers can be removed.

With GCC, the engine still has to build with version 4.8 because it's the one we are using on Travis. If it weren't for that we would surely want to choose something newer.

Regarding MSVC, I can say that the code currently builds in Visual Studio 2015, but doesn't build in VS 2013. VS 2017 is the latest version.

I don't know much about the versions of Clang. Perhaps the version shipped by NaCl would be the constraint here.

slipher:
I picked these based on what we have for CI.

Fixes #114.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,11 @@ The standalone engine that powers the multiplayer first person shooter [Unvanqui
 
 ### Buildtime
 
-`cmake`
+- CMake
+- A C++ compiler. The following are actively supported:
+  - GCC (at least 4.8)
+  - Visual Studio/MSVC (at least Visual Studio 2015)
+  - Clang (at least 3.6)
 
 ### Optional 
 


Kangz:
I was able to get a travis build work using recent clang versions in another project. Let me know if you're interested in taking a look.

slipher:
I actually based the clang version number on the NaCl compiler version, but forgot to check which version is used on travis (apparently it's 3.5).

Kangz:
You can get LLVM 7 by doing stuff like in this https://github.com/google/nxt-standalone/blob/master/.travis.yml

PR #134, opened 2018-10-29
Remove old compiler workarounds

slipher:
According to the minimum supported versions now defined in the README.

### DIFF
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -110,11 +110,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 # define ATTRIBUTE_NO_SANITIZE_ADDRESS
 #endif
 
-#ifdef __cplusplus
 #include <new>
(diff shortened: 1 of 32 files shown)


src/common/Compiler.h
```diff
@@ -177,25 +154,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #define BREAKPOINT() __debugbreak()
-#define OVERRIDE override
-#define FINAL final
-// VS2015 supports this
-#if _MSC_VER >= 1900
 #define NOEXCEPT noexcept
 #define NOEXCEPT_IF(x) noexcept(x)
-#define NOEXCEPT_EXPR(x) false
-#define HAS_EXPLICIT_DEFAULT
+#define NOEXCEPT_EXPR(x) false // TODO: Find out why this is defined to false?
```
t4im: I think that's just a copy-paste oversight in 8d3667272
12c950a93 Introduced this with _IF for the specifier and _EXPR with a boolean return for the operator.

The [release](https://docs.microsoft.com/en-us/visualstudio/releasenotes/vs2015-rtm-vs) notes for VS2015 state

> The noexcept operator can now be used to check whether an expression might throw an exception.
> The noexcept specifier can now be used to specify that a function does not throw exceptions. C++11

So both is supported.

slipher: Thanks, I have fixed the definition.

PR #135, opened 2018-11-06
Daemon 0.51, NOW SUPPORTING STRINGS OVER 100,000 BYTES!

slipher:


t4im:
lgtm

t4im:
Run some asserts on previous and new behavior:

```C
	ASSERT_EQ(-1, Q_stricmp("a Dæmon", "B dæmon"))
	ASSERT_EQ(1, Q_stricmp("b Dæmon", "A dæmon"))
	ASSERT_EQ(-1, Q_stricmp("", "daemon"))
	ASSERT_EQ(1, Q_stricmp("dæmon", ""))
	ASSERT_EQ(0, Q_stricmp("", ""))
	ASSERT_EQ(0, Q_stricmp("unVANQUISHED", "UNvanquished"))
```

Passing with the old this way, it fails with
`"-1 == Q_stricmp("", "daemon")" expected: -1, actual: 1`
in the new (i suspect the same with the following one)

Would be nice if could add unit tests somewhere :/

slipher:
Thanks for checking! It is unfortunate that dark forces conspire to prevent our codebase from having any tests.

Viech:
Do you know the **measured aggression™** approach to code testing? Only write new tests for as long as the number of open issues brought to light is bounded above by the square root of the total number of test cases. By having no test cases, we're already at a sweet spot (don't jinx it)! :+1:

PR #136, opened 2018-11-07
Add CBSE-templates to cmake DEPENDS to force regeneration on template change

t4im:


cmake/DaemonCBSE.cmake
```diff
@@ -49,6 +49,10 @@ function(CBSE target definition output)
         OUTPUT ${GENERATED_CBSE}
         COMMENT "Generating CBSE entities for ${definition}"
         DEPENDS ${definition}
+		${CMAKE_SOURCE_DIR}/src/utils/cbse/templates/Backend.cpp
+		${CMAKE_SOURCE_DIR}/src/utils/cbse/templates/Backend.h
+		${CMAKE_SOURCE_DIR}/src/utils/cbse/templates/Component.cpp
+		${CMAKE_SOURCE_DIR}/src/utils/cbse/templates/Component.h
```
illwieckz: it looks like original file used space-based spacing™ (I don't know why but all the cmake files I've seen use spaces even if tabs work)

PR #137, opened 2018-11-08
improve travis build: +ccache, +osx build, -dependencies

t4im:
Might remove the engine builds from Unvanquished next so that things only get built in the places they are relevant in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,65 @@
+sudo: required
+
 language:
   - cpp
-#  - objective-c
 
-dist: trusty
-sudo: required
+os:
+  - linux
+  - osx
+
+cache: ccache
 
 compiler:
   - gcc
   - clang
 
+#Prepare dependency installation
+before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+      sudo apt-get -qq update
+    fi
+
 install:
-  - sudo add-apt-repository ppa:zoogie/sdl2-snapshots -y
-  - sudo add-apt-repository ppa:george-edison55/cmake-3.x -y
-  - sudo apt-get -qq update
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,33 +1,66 @@
...
+        liblua5.2-dev;
+    fi
+  - |
+    if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+      sudo apt-get -y -q --no-install-recommends install ninja-build
+    elif [ "$TRAVIS_OS_NAME" == "osx" ]; then
+      brew install ninja
+    fi
+  - |
+    # workarounds to make ccache work
```
illwieckz: I'm almost a totally newb there, what means that pipe?

t4im: a literal multiline string; so the new lines from the yaml file are preserved

Here a great overview of this one and the other 8 types:
https://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines/21699210#21699210

illwieckz: that's what my brain inferred, but in this case why is this pipe missing for the first `if` block in `install` target?

illwieckz: ok so  the other way to not use pipe would be to put a `;` at the end of `sudo apt-get -y -q --no-install-recommends install ninja-build` and at the end of `brew install ninja`, right?

t4im: In the apt-get command you don't want any new lines, so no pipe there; it will then be passed to the OS as a single line.

I.e.
```bash
apt-get install
something
```
would fail as not knowing the command `something`

.travis.yml
```diff
@@ -1,33 +1,66 @@
+sudo: required
+
 language:
   - cpp
-#  - objective-c
 
-dist: trusty
```
slipher: Does removing this change anything?

t4im: no, it still uses trusty after

illwieckz: trusty being supported until april 2019 (six months ahead) I guess travis will update to something else (xenial, 16.04, up to april 2021) around this date

PR #140, opened 2018-11-13
Fix buffer over-read via std::string::assign copying from buffers past '\0'.

t4im:
The to-be-assigned strings are already providing null-terminated strings, often by means of `Q_strncpyz`, so `std::string::assign(cstring)` will correctly stop copying on termination (excluding '\0').

The replaced usages of `std::string::assign(cstring, size)` were continuing copying past the message body/past any termination to the full buffer length (often `MAX_STRING_CHARS`=1024 bytes)

This was mostly hidden by additional `Q_strncpyz` calls for c-string api compatibility in trap-functions but would fail in modernized code (e.g. with `empty()`).


slipher:
I'm not a fan of this change as-is. It relies on all of these functions always populating the buffer with a valid null-terminated string.

Maybe set the first element of each buffer to '\0' after allocating it?

t4im:
Yea, good idea. Using the null-renderer might actually have triggered such a situation.

PR #141, opened 2018-11-15
Some c++11 modernizations

t4im:
A few modernization replacements with the help of clang's analyzer.
The libraries (including tinyformat and tinygettext) were skipped.

### DIFF
--- a/src/engine/audio/Emitter.cpp
+++ b/src/engine/audio/Emitter.cpp
@@ -258,11 +258,9 @@ namespace Audio {
 
     // Implementation for Emitter
 
-    Emitter::Emitter() {
-    }
+    Emitter::Emitter() = default;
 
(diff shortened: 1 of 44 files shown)


src/engine/audio/Emitter.cpp
```diff
@@ -275,8 +273,7 @@ namespace Audio {
     EntityEmitter::EntityEmitter(int entityNum): entityNum(entityNum) {
     }
 
-    EntityEmitter::~EntityEmitter(){
-    }
+    EntityEmitter::~EntityEmitter()= default;
```
illwieckz: space before `=` #bikeshedding

t4im:
According to clang the declaration with an explicit `default` creates more optimization opportunities.

PR #146, opened 2018-12-25
tty: Do not use nonblocking stdin 

DolceTriade:
This is buggy because if the engine is launched from a shell,
stdin will be shared and if anything on the shell (like gdb)
changes the stdin status, then unvanquished will also block.
Instead use select to determine if there is input ready to
be read. I didn't not notice any performance impact from this
change.

### DIFF
--- a/src/engine/sys/con_tty.cpp
+++ b/src/engine/sys/con_tty.cpp
@@ -231,6 +231,22 @@ static void CON_Show()
 	}
 }
 
+static bool CON_InputAvailable() {
+	fd_set         fdset;
+	struct timeval timeout;
+
+	FD_ZERO( &fdset );
+	FD_SET( STDIN_FILENO, &fdset );  // stdin
+	timeout.tv_sec = 0;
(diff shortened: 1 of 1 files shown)


DolceTriade:
Pretty sure this is wrong.

illwieckz:
what's the status on this?

DolceTriade:
It's still wrong and there is no good fix.

illwieckz:
is there any hope to be able to run dæmon on gdb again one day (even with dirty solutions)?

DolceTriade:
Run it with ncurses lol

illwieckz:
Oh thank you! That's ugly because gdb messes with the curses output, but it works :+1:

But why it suddenly stopped to work?

illwieckz:
Is there a way to entirely disable input for such case? there is still in game console and `daemon-tty`…

DolceTriade:
Maybe we can disable tty. I'm not sure if there is a flag to disable it.

PR #149, opened 2019-01-01
Miscellaneous patches

slipher:


src/common/Command.cpp
```diff
@@ -377,7 +377,12 @@ namespace Cmd {
     }
 
     const std::string& Args::Argv(int argNum) const {
-        return args[argNum];
+        static std::string empty;
```
DolceTriade: I don't think we should mask programming errors like this. I think out of bounds accesses should crash. There is no reason that a programmer should attempt to access an out of bounds item.

src/engine/renderer/tr_init.cpp
```diff
@@ -661,82 +646,84 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+		cmd->format = format;
 
-			ri.RealTime( &t );
+		return true;
+	}
 
-			// scan for a free filename
-			for ( lastNumber = 0; lastNumber <= 999; lastNumber++ )
+namespace {
+	class ScreenshotCmd : public Cmd::StaticCmd {
```
DolceTriade: personal preference but we shouldn't indent namespaces.

src/engine/renderer/tr_init.cpp
```diff
@@ -661,82 +646,84 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+		const ssFormat_t format;
+		const std::string fileExtension;
+	public:
+		ScreenshotCmd(std::string cmdName, ssFormat_t format, std::string ext)
```
DolceTriade: why are we passing these by value?

src/engine/renderer/tr_init.cpp
slipher: Dedented.

src/common/Command.cpp
slipher: Well there is no guarantee it crashes, you can get garbage data instead which is a lot less nice. I don't want to have the undefined behavior here since it would be easy to miss when porting an old-style command. I added an assert so it will crash in debug mode.

src/engine/renderer/tr_init.cpp
slipher: It is rarely used code with small strings, so I don't find the textual noise added by `std::move` worthwhile. While if you used `const&`, there wouldn't actually be less copies.

src/common/Command.cpp
Kangz: I think @DolceTriade means that the ASSERT should be enough.

DolceTriade: ya

slipher: You want it to silently return garbage in release mode?

Kangz: Yes

PR #150, opened 2019-01-04
Delete unused stuff in cgame ABI

slipher:


src/engine/client/cg_msgdef.h
```diff
@@ -133,7 +133,6 @@ enum cgameImport_t
   CG_SETUSERCMDVALUE,
   CG_GET_ENTITY_TOKEN,
   CG_REGISTER_BUTTON_COMMANDS,
-  CG_GETCLIPBOARDDATA,
```
DolceTriade: I'm surprised librocket doesn't use this..

PR #151, opened 2019-01-05
Fix stack alignment in build script for x86 Windows

slipher:


src/engine/client/cl_irc.cpp
```diff
@@ -2533,7 +2533,7 @@ static HANDLE IRC_ThreadHandle = nullptr;
 IRC_SystemThreadProc
 ==================
 */
-static DWORD WINAPI IRC_SystemThreadProc( LPVOID )
+static DWORD ALIGN_STACK_FOR_MINGW WINAPI IRC_SystemThreadProc( LPVOID )
```
Kangz: Should this attribute be used on every thread entrypoint? There are at least 3 other threads I know of: in . System.cpp, VirtualMachine.cpp and somewhere in the renderer.

slipher: I assume it should be handled correctly in the ones that use the `std::thread` library provided by mingw, so I didn't put it on the singleton socket one. Added the attribute to the DLL entry and the render thread.

PR #152, opened 2019-01-06
Fixing FS::InternalLoadPak to only get timestamp for .dpk files.

JacksonTech:
No longer attempts to get timestamp for directories.

Fixes bug where my_stat (_wstat64i) was failing on Windows when
trailing slash was present.

In the future, trim trailing slash should work on all platforms.

Not sure if this has unintended consequences (are timestamps _really_ useless for dpkdirs?). Please review.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -1142,7 +1142,8 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 	// Save the real checksum in the list of loaded paks (empty for directories, not used for legacy paks)
 	loadedPaks.back().realChecksum = realChecksum;
 
-	// Get the timestamp of the pak, but only for directories. Directories don't need timestamp.
+	// Get the timestamp of the pak, but only for dpk files. 
+	// Directories (aka a dpkdir) don't need timestamp.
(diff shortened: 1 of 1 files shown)


src/common/FileSystem.cpp
```diff
@@ -1142,10 +1142,14 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 	// Save the real checksum in the list of loaded paks (empty for directories, not used for legacy paks)
 	loadedPaks.back().realChecksum = realChecksum;
 
-	// Get the timestamp of the pak
-	loadedPaks.back().timestamp = FS::RawPath::FileTimestamp(pak.path, err);
-	if (err)
-		return;
+	// Get the timestamp of the pak, but only for directories. Directories don't need timestamp.
```
slipher: Comment says "only for directories" when the opposite is intended.

Kangz: What is the timestamp used for?

slipher: The timestamp is not used, except in the implementation of a VM syscall to return the timestamp, which is also not used. Maybe the intention was to eventually avoid reloading paks that are already in memory and not modified, or something.

PR #155, opened 2019-01-26
Remove hunk memory system from server / replace Com_Error

slipher:


slipher:
It was broken but now it isn't.

src/engine/server/sv_init.cpp
```diff
@@ -465,8 +465,9 @@ void SV_SpawnServer(const std::string pakname, const std::string server)
 		}
 	}
 
-	// allocate the snapshot entities on the hunk
-	svs.snapshotEntities = ( entityState_t * ) Hunk_Alloc( sizeof( entityState_t ) * svs.numSnapshotEntities, ha_pref::h_high );
+	// allocate the snapshot entities
+	free(svs.snapshotEntities);
+	svs.snapshotEntities = static_cast<entityState_t*>(calloc(svs.numSnapshotEntities, sizeof(entityState_t)));
```
DolceTriade: Please comment about the lifetime of the allocations.

slipher: The lifetime was... it gets leaked when svs gets memset. I've changed it to use a unique_ptr and destroy svs instead of using memset.

src/engine/client/cl_main.cpp
```diff
@@ -754,14 +754,14 @@ void CL_FlushMemory()
 	if ( !com_sv_running->integer )
 	{
 		// clear the whole hunk
-		Hunk_Clear();
+		Hunk_ShutDownRandomStuffAndClear();
```
DolceTriade: heh, maybe rename it to ShutdownSubsystems() or something and don't have it call Hunk_Clear()... idk, this seems like a weird construct anyways.

slipher: I'm hoping to do some more work on initialization and shutdown soon. In the meantime I like to give it a name that highlights the weirdness.

PR #157, opened 2019-02-04
Don't serialize bools by copying memory

### LINKED ISSUE #148: IPC: Unsound handling of bools
The IPC serialization code assumes that any POD type can be communicated by simply copying the memory. This is a problem if not every possible sequence of bytes represents a valid value.

In particular, the assumption opens the possibility of sending spooky `bool`s which behave as true in some contexts and false in others, or have values other than `0` or `1` after an integer promotion.

Of course it's easy to make a specialization for `bool` itself, but `bool`s in structs pose much more of a problem. There are various large structs which are serialized by copying, some of which contain `bool`s. Maybe we would need some static analysis to prevent any `bool`s from being added to them.

slipher:
Mostly fixes #148 although there could be a few more `bool`s hiding out in structs.

### DIFF
--- a/src/common/Serialize.h
+++ b/src/common/Serialize.h
@@ -220,9 +220,11 @@ namespace Util {
 		size_t handles_pos;
 	};
 
-    // Implementation of the serialization traits for common types and std containers
+	// Implementation of the serialization traits for common types and std containers
 
 	// Simple implementation for POD types
+	// This is only safe to use if every possible byte sequence represents a valid value for the
+	// object. So it cannot be used for bool or a struct containing a bool.
 	template<typename T>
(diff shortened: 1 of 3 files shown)


src/common/Serialize.h
```diff
@@ -237,6 +239,19 @@ namespace Util {
 		}
 	};
 
+	// bool
+	template<>
+	struct SerializeTraits<bool> {
+		static void Write(Writer& stream, bool value)
+		{
+			stream.Write<uint8_t>(+value);
```
DolceTriade: what does the + do

slipher: Convert bool to int

slipher:
Had to add some extra preprocessor magic since `bool` was defined as a macro on gcc.

PR #159, opened 2019-02-09
ParseShader: load extra maps the DarkPlaces way if r_dpMaterial enabled

illwieckz:
If `r_dpMaterial` is set to `on`, load extra textures the [DarkPlaces way](https://gitlab.com/xonotic/darkplaces/blob/7f47f772ef10b488f4ae8012bda40deb30862ed2/gl_rmain.c#L3526) by looking for extra files with special suffixes like `_norm` or `_gloss`.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -49,6 +49,9 @@ static cullType_t    implicitCullType;
 
 static char          whenTokens[ MAX_STRING_CHARS ];
 
+// DarkPlaces material compatibility
+static Cvar::Cvar<bool> r_dpMaterial("r_dpMaterial", "Enable DarkPlaces material compatibility", Cvar::NONE, false);
+
 /*
 ================
 return a hash value for the filename
@@ -1483,6 +1486,86 @@ static bool ParseClampType( char *token, wrapType_t *clamp )
 	if (t) { clamp->t = type; }
 
 	return true;
+}
+
+/*
+===================
+FindMapInStage
+
+returns true if the shader stage contains a map line
+
+for example this code will match:
+
+```
+	{
+		map textures/map_solarium/water4/water4.tga
+		tcmod scale 0.3 0.4
+		tcMod scroll 0.05 0.05
+		blendfunc add
+		alphaGen vertex
+	}
+```
+
+the pointeri to the found map file name is stored in *buffer
+here it would be the pointer to
+
+```
+textures/map_solarium/water4/water4.tga
+```
+
+nothing more is done at this point, but ParseShader
+will be able to rely on it to look for extra maps based the
+basename of this filename and suffixes
+
+===================
+*/
+static bool FindMapInStage( const char *text, char *buffer, int bufferlen )
+{
+	char *token;
+	bool foundMap = false;
+
+	while ( true )
+	{
+		token = COM_ParseExt2( &text, true );
+
+		if ( token[ 0 ] == '\0' )
+		{
+			// the real parser will complain about it
+			// no need to print twice the same warning
+			return false;
+		}
+		if ( token[ 0 ] == '}' )
+		{
+			break;
+		}
+		else if ( !Q_stricmp( token, "stage" ) )
+		{
+			return false;
+		}
+		else if ( !Q_stricmp( token, "map" ) )
+		{
+			if ( !ParseMap( &text, buffer, bufferlen ) )
+			{
+				return false;
+			}
+			foundMap = true;
+		}
+	}
+
+	if ( foundMap )
+	{
+		// texture path starting with one of those chars is known
+		// to be en engine internal texture (usually produced by code)
+		if ( buffer[0] == '*'
+			|| buffer[0] == '_'
+			|| buffer[0] == '$' )
+		{
+			return false;
+		}
+		return true;
+	}
+
+	return false;
 }
 
 /*
@@ -3106,6 +3189,20 @@ static void ParseLightFalloffImage( shaderStage_t *stage, const char **text )
 	}
 }
 
+struct extraMapParser_t
+{
+	const char *suffix;
+	const char *description;
+	void ( *parser ) ( shaderStage_t*, const char** );
+};
+
+static const extraMapParser_t extraMapParsers[] =
+{
+	{ "norm",    "normal map",     ParseNormalMap     },
+	{ "gloss",   "specular map",   ParseSpecularMap   },
+	{ "glow",    "glow map",       ParseGlowMap       },
+};
+
 /*
 =================
 ParseShader
@@ -3150,6 +3247,125 @@ static bool ParseShader( const char *_text )
 		// stage definition
 		else if ( token[ 0 ] == '{' )
 		{
+			if ( *r_dpMaterial )
+			{
+				/* DarkPlaces material compatibility
+
+				note that the DarkPlaces renderer code is very different than Dæmon one
+				so that code is just trying to produce the same results but it shares
+				no one code at all
+
+				note that dp* keywords will be processed elsewhere
+
+				look if the stage to be parsed has a map keyword without a stage keyword,
+				if there is look for extra maps based on their suffixes,
+				if they exist load them,
+				and tell the stage to be parsed is a diffuseMap one just before parsing it
+
+				basically it will turn this:
+
+				```
+					textures/map_solarium/water4
+					{
+						qer_editorimage textures/map_solarium/water4/water4.tga
+						qer_trans 20
+						surfaceparm nomarks
+						surfaceparm trans
+						surfaceparm water
+						surfaceparm nolightmap
+						cull none
+						q3map_globaltexture
+						tessSize 256
+
+						{
+							map textures/map_solarium/water4/water4.tga
+							tcmod scale 0.3 0.4
+							tcMod scroll 0.05 0.05
+							blendfunc add
+							alphaGen vertex
+						}
+						dpreflectcube cubemaps/default/sky
+						{
+							map $lightmap
+							blendfunc add
+							tcGen lightmap
+						}
+						dp_water 0.1 1.2  1.4 0.7  1 1 1  1 1 1  0.1
+					}
+				```
+
+				into this:
+
+				```
+					textures/map_solarium/water4
+					{
+						qer_editorimage textures/map_solarium/water4/water4.tga
+						qer_trans 20
+						surfaceparm nomarks
+						surfaceparm trans
+						surfaceparm water
+						surfaceparm nolightmap
+						cull none
+						q3map_globaltexture
+						tessSize 256
+
+						normalMap textures/map_solarium/water4/water4_norm
+						specularMap textures/map_solarium/water4/water4_gloss
+						{
+							stage diffuseMap
+							map textures/map_solarium/water4/water4.tga
+							tcmod scale 0.3 0.4
+							tcMod scroll 0.05 0.05
+							blendfunc add
+							alphaGen vertex
+						}
+						dpreflectcube cubemaps/default/sky
+						{
+							map $lightmap
+							blendfunc add
+							tcGen lightmap
+						}
+						dp_water 0.1 1.2  1.4 0.7  1 1 1  1 1 1  0.1
+					}
+				```
+
+				*/
+
+				char colorMapName[ MAX_QPATH ] = "";
+				bool foundMap = FindMapInStage( *text, colorMapName, sizeof( colorMapName ) );
+				if (foundMap) {
+					char colorMapBaseName[ MAX_QPATH ];
+					bool foundExtraMap = false;
+
+					Log::Debug( "looking for extra maps for color map: '%s'", colorMapName );
+					COM_StripExtension3( colorMapName, colorMapBaseName, sizeof( colorMapBaseName ) );
+
+					for ( const extraMapParser_t parser: extraMapParsers )
+					{
+						if ( s >= ( MAX_SHADER_STAGES - 1 ) )
+						{
+							Log::Warn( "too many extra stages in shader %s", shader.name );
+							return false;
+						}
+
+						std::string extraMapName = Str::Format( "%s_%s", colorMapBaseName, parser.suffix );
+						if( R_FindImageLoader( extraMapName.c_str() ) >= 0 )
+						{
+							foundExtraMap = true;
+							Log::Debug( "found extra %s '%s'", parser.description, extraMapName.c_str() );
+							const char *name = extraMapName.c_str();
+							parser.parser( &stages[ s ], &name );
+							s++;
+						}
+					}
+
+					if ( foundExtraMap )
+					{
+						stages[ s ].type = stageType_t::ST_DIFFUSEMAP;
+					}
+				}
+			}
+
 			if ( s >= ( MAX_SHADER_STAGES - 1 ) )
 			{
 				Log::Warn("too many stages in shader %s", shader.name );


illwieckz:
By not working I mean the engine reports to find the extra texture files, but unfortunately it only renders diffuse maps.

src/engine/renderer/tr_shader.cpp
```diff
@@ -3678,6 +3701,46 @@ static bool ParseShader( const char *_text )
...
+	shaderStage_t colorMapStage;
+
+	for ( c = 0; c < s; c++ ) {
+		if ( stages[c].type == stageType_t::ST_COLORMAP && stages[c].textureName != nullptr ) {
+			colorMapStage = stages[c];
+			break;
+		}
+	}
+
+	if ( colorMapStage.textureName[0] != '\0' )
```
slipher: If the immediately preceding for loop does not find anything, this reads from an uninitialized value.

src/engine/renderer/tr_shader.cpp
```diff
@@ -3678,6 +3701,46 @@ static bool ParseShader( const char *_text )
...
+			{
+				Log::Warn( "too many extra stages in shader %s", shader.name );
+				return false;
+			}
+
+			mapName = va( "%s%s", baseName, extraMapParsers[i].suffix );
+			if( R_FindAltImage( (char*) mapName, &prefix ) >= 0 ) {
+				Log::Debug( "found extra %s '%s'", extraMapParsers[i].description, mapName );
+				// colorMapStage.type = stageType_t::ST_DIFFUSEMAP;
+				extraMapParsers[i].parser( &stages[ s ], &mapName );
```
slipher: It looks to me like the 2nd parameter to Parse*Map functions is supposed the code of a q3 shader, rather than a filename.

illwieckz: I've put `Log::Notice` everywhere, and when there is `normalMap something` in a shader, the original code passes ` something` to `ParseNormalMap`.

src/engine/renderer/tr_shader.cpp
illwieckz: good catch!

src/engine/renderer/tr_shader.cpp
illwieckz: for reference, this comment is misleading (I verified the keyword is never there):

https://github.com/DaemonEngine/Daemon/blob/caf9df48ec68021e7ebbf66d68972a14f11d6416/src/engine/renderer/tr_shader.cpp#L1330-L1336

this code says the keyword has no parameter, not that the keyword is not passed, which is the observed behavior:

https://github.com/DaemonEngine/Daemon/blob/caf9df48ec68021e7ebbf66d68972a14f11d6416/src/engine/renderer/tr_shader.cpp#L1352-L1356

src/engine/renderer/tr_image.cpp
```diff
@@ -1611,6 +1611,58 @@ static const imageExtToLoaderMap_t imageLoaders[] =
...
 
+/*
+=================
+R_FindAltImage
+
+Finds an alternative image path, set the required prefix if needed,
+then returns the loader index.
+=================
+*/
+int R_FindAltImage( char *filename, const char **prefix ) {
```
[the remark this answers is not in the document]
illwieckz: in fact the function takes filename as argument (in fact a basename, I edited this), then return a prefix string and an index to a given format (for extension/loader), but this is c++ and wen can't return two variable neither return string :grin:

src/engine/renderer/tr_image.cpp
```diff
@@ -1611,6 +1611,58 @@ static const imageExtToLoaderMap_t imageLoaders[] =
 
 static int                   numImageLoaders = ARRAY_LEN( imageLoaders );
 
+/*
+=================
+R_FindAltImage
```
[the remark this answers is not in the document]
illwieckz: maybe, naming is part of "work in progress"

src/engine/renderer/tr_image.cpp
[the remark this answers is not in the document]
illwieckz: also the prefix is only set on rare use case (dds prefix like DarkPlaces/Doom3), it looks a bit costly to compute a string for every texture because of a use case no one unvanquished player will face.

illwieckz:
I think there is a special command to list opened files, does someone remember it?

illwieckz:
For information this is how it looks on console:

```
Debug: looking for extra maps for color map: 'textures/phillipk2x/computer/pk02_switches01c' 
Debug: found extra glow map 'textures/phillipk2x/computer/pk02_switches01c_glow' 
Debug: found extra normal map 'textures/phillipk2x/computer/pk02_switches01c_norm' 
Debug: found extra specular map 'textures/phillipk2x/computer/pk02_switches01c_gloss' 
```

That's why it _seems_ to work: the files are found. But what's painted on screen obviously proves that only diffusemap is rendered.

illwieckz:
At this point the engine code claims to turn this:

```c
textures/phillipk2x/computer-pk02_switches01c
{
    qer_editorimage textures/phillipk2x/computer/pk02_switches01c

    dpoffsetmapping - .68120712951256245813 match8 89.89403100000000000000
    dpglossintensitymod 3
    dpglossexponentmod 4

    q3map_bouncescale 1.25
    {
        map textures/phillipk2x/computer/pk02_switches01c
    }
    {
        map $lightmap
        rgbgen identity
        tcgen lightmap
        blendfunc filter
    }
}
```

into this (`dp*` keywords are ignored with an warning):

```c
textures/phillipk2x/computer-pk02_switches01c
{
    qer_editorimage textures/phillipk2x/computer/pk02_switches01c

    q3map_bouncescale 1.25
    {
        map textures/phillipk2x/computer/pk02_switches01c
    }
    {
        map $lightmap
        rgbgen identity
        tcgen lightmap
        blendfunc filter
    }
    glowMap textures/phillipk2x/computer/pk02_switches01c_glow
    normalMap textures/phillipk2x/computer/pk02_switches01c_norm
    specularMap textures/phillipk2x/computer/pk02_switches01c_gloss
}
```

Note that the code runs after everything else is parsed, hence right before the ending `}`.
It means the code runs after the second stage (`map $lightmap`), perhaps that's the problem ?
I verified that `glowMap`, `normalMap` and `specularMap` are generated as stage `2`, `3` and `4` (original ones being stages `0` and `1`).

I also verified that those files are loaded:

```
dds/textures/phillipk2x/computer/pk02_switches01c.dds
dds/textures/phillipk2x/computer/pk02_switches01c_gloss.dds
dds/textures/phillipk2x/computer/pk02_switches01c_glow.dds
dds/textures/phillipk2x/computer/pk02_switches01c_norm.dds
```

illwieckz:
To get a testbed, people may copy (or symlink) `xonotic-20170401-maps.pk3` into a brand new directory, then run this code within that directory:

```sh
p="$(ls 'xonotic-'*'-maps.pk3' | tail -n 1)"
b='xonotic-maps'

listMaps () {
    unzip -l "${p}" \
    | tail -n +4 \
    | head -n -2 \
    | egrep '\.bsp$' \
    | sed -e 's|^.* maps/||;s|\.bsp$||'
}

filterSystem () {
    egrep -v '^_'
}

filterInvalid () {
    egrep -v '_'
}

ln -sfv "${p}" "${b}_0.dpk"

for m in $(listMaps | filterSystem | filterInvalid)
do
    d="map-${m}_0.dpkdir"
    mkdir -pv "${d}"
    echo "${b}" > "${d}/DEPS"
done
```

then run dæmon this way:

```sh
daemon -pakpath /path/to/xonotic/pkg \
    -set logs.logLevel.default debug \
    -set language en \
    -set developer 1 \
    +set g_neverEnd 1 \
    +devmap solarium
```

src/engine/renderer/tr_image.cpp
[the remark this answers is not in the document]
illwieckz: I renamed it to `R_FindImageLoader` because it's its primary purpose, a loader is found because an alternative image is found for this loader, and the function returns the loader index. It also fills the `prefix` blank if the image loader needs this prefix.

Since this function checks if an image basename has a loader, it may be abused to check if a file exists with a given basename. This is what the DarkPlaces extra texture lookup code does: it computes a brand new image basename and checks if it exists.

src/engine/renderer/tr_image.cpp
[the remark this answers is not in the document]
illwieckz: Also, when this function is only used to tell if a file with a given basename exists (like in DarkPlaces compat code), both the prefix and the final filepath are not used, would be useless to compute that string for nothing. The good thing is that setting a prefix is just setting a pointer to a `const char*`, processing is only done when people really wants to load the picture. And since this is done later in any way, no need to compute two times the same string at two memory places to only use one if them.

illwieckz:
When I print `numActiveStages` the in `CollapseStages()` I get the expected amount of stages, for example 6 for this shader:

```c
textures/exomorphx/light-metal_light03
{
	qer_editorimage textures/exomorphx/light/metal_light03
	dpoffsetmapping - 5.05836666682224888188 match8 160.48228200000000000000
	dpglossintensitymod  2
	dpglossexponentmod  4
	dpreflectcube env/exomorph/exomorph
	surfaceparm metalsteps
	q3map_bouncescale 1.50

	{
		map textures/exomorphx/light/metal_light03
	}

	{
		map $lightmap
		rgbGen identity
		tcGen lightmap
		blendfunc filter
	}
}

```

produces:

```
Warn: unknown general shader parameter 'dpoffsetmapping' in 'textures/exomorphx/light-metal_light03' 
Warn: unknown general shader parameter 'dpglossintensitymod' in 'textures/exomorphx/light-metal_light03' 
Warn: unknown general shader parameter 'dpglossexponentmod' in 'textures/exomorphx/light-metal_light03' 
Warn: unknown general shader parameter 'dpreflectcube' in 'textures/exomorphx/light-metal_light03' 
Debug: looking for extra maps for color map: 'textures/exomorphx/light/metal_light03' 
Debug: found extra glow map 'textures/exomorphx/light/metal_light03_glow' 
Debug: found extra normal map 'textures/exomorphx/light/metal_light03_norm' 
Debug: found extra specular map 'textures/exomorphx/light/metal_light03_gloss' 
Debug: found extra reflection map 'textures/exomorphx/light/metal_light03_reflect' 
Debug: active stage count: 6 
```

Six stages because of the two original stages plus the four computed ones:

```
glowMap textures/exomorphx/light/metal_light03_glow
normalMap textures/exomorphx/light/metal_light03_norm
specularMap textures/exomorphx/light/metal_light03_gloss
reflectionMap textures/exomorphx/light/metal_light03_reflect
```

illwieckz:
If I hack stuff by disabling the color map stage and creating a brand new diffuse map stage instead with the same texture path, it works, but it's a hack since we lose all the other keywords that may be there (like `tcMod` and so on), but it helped me to validate we have to turn the colormap into a diffusemap to get things working.

Note that the lighting is wrong because Xonotic uses sRGB lightmaps and Dæmon does not support them yet.

## without hack

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-09_224937_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-09_224949_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-09_225513_000.jpg)

## with hack

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-09_225100_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-09_225120_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-09_225632_000.jpg)

illwieckz:
Note that since Xonotic ships a heightmap with almost all its normalmap, I enabled temporarily parallax on all shaders just to have a look. It works (and I haven't experienced preformance issue) but I discovered that xonotic heightmaps are upside down, or our engine is.

Perhaps it's a trick to not have to detect the presence of heightmap hidden in alpha channel of normalmap, as a missing alphachannel would mean a flat surface at the good level. Currently our engine displays a flat surface at the wrong level if we enable parallax on a normalmap without heightmap in alpha channel.

[xonotic upside-down parallax](https://dl.illwieckz.net/b/daemon/bugs/upside-down-parallax/unvanquished_2019-02-09_231313_000.jpg)

[xonotic upside-down parallax](https://dl.illwieckz.net/b/daemon/bugs/upside-down-parallax/unvanquished_2019-02-09_231324_000.jpg)

[xonotic upside-down parallax](https://dl.illwieckz.net/b/daemon/bugs/upside-down-parallax/unvanquished_2019-02-09_231419_000.jpg)

[xonotic upside-down parallax](https://dl.illwieckz.net/b/daemon/bugs/upside-down-parallax/unvanquished_2019-02-09_231555_000.jpg)

slipher:
If you actually write the shader in [this comment](https://github.com/DaemonEngine/Daemon/pull/159#issuecomment-462055874) in a file, does it do what you want?

illwieckz:
I have not tried but I'm now almost sure it won't work because this is a colorMap stage:

```c
	{
		map textures/exomorphx/light/metal_light03
	}
```

and a diffuseMap stage looks like this:

```c
	diffuseMap textures/exomorphx/light/metal_light03
```

or
```c
	{
		stage diffuseMap
		map textures/exomorphx/light/metal_light03
	}
```

The engine code is explicit: normalMap, glowMap, specularMap and others only works with diffuseMap stage, not colormap stage.

https://github.com/DaemonEngine/Daemon/blob/be89425dd92889b0f965d6890a804148b7832711/src/engine/renderer/tr_shader.cpp#L3768-L3783

So what we need is to transform the colorMap to be a diffuseMap. I don't know why but setting the diffuseMap bits in a colorMap stage does not do the trick. The hacky way that work is to disable the colorMap and load a brand new diffuseMap that is activated instead. That's how I produced those pictures above, the problem is that sometime those colorMap stages ship more informations that are then lost.

For example the colorMap of the solarium water is this one:

```
	{
		map textures/map_solarium/water4/water4.tga
		tcmod scale 0.3 0.4
		tcMod scroll 0.05 0.05
		blendfunc add
		alphaGen vertex
	}
```

If I create a diffuseMap using the `textures/map_solarium/water4/water4.tga` path and disable the original colorMap, I just lost everything else that came after that: wobbling, transparency…

illwieckz:
I also think that changing the stage type bits later does not work because the image is already loaded, and I guess the bits must be set before the image loading. It's a strong guess but a guess. There is still things I have not read or not understood enough to say it's the issue, but there is a very high probability that the issue is there.

illwieckz:
So, I pushed a new attempt. Instead of waiting the end of the shader parsing (and having to store the diffuse map name somewhere), the code does a quick lookup on the current stage before parsing it, if it finds a map without stage, it looks for extra maps, if it finds extra maps, it load them and set the current stage as a diffusemap one, then it parses the stage, so no one data is deleted and the stage is loaded as a diffuse one from the start.

This is how it looks now, it's not bad. Remind the lighting must be wrong since we don't have sRGB support yet. Also, it's hard to know whats wrong things comes from potential mistakes in the code or from unsupported features.

That's the best result I got _without doing nasty stuff_.

It's important to know we can get things that looks better but doing things that are wrong on purpose. I got things that looks better by disabling entire non-picture stages (i.e. disabling features), that's obviously wrong and can't count as reference.

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_041801_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_041858_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_041916_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_042001_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_042058_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_042336_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_042142_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_042249_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_042449_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_042613_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_042829_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_042854_000.jpg)

[xonotic extramap](https://dl.illwieckz.net/b/daemon/wip/extramap/unvanquished_2019-02-10_044032_000.jpg)

illwieckz:
So, currently this:

```c
textures/phillipk2x/computer-pk02_switches01c
{
    qer_editorimage textures/phillipk2x/computer/pk02_switches01c

    dpoffsetmapping - .68120712951256245813 match8 89.89403100000000000000
    dpglossintensitymod 3
    dpglossexponentmod 4

    q3map_bouncescale 1.25
    {
        map textures/phillipk2x/computer/pk02_switches01c
    }
    {
        map $lightmap
        rgbgen identity
        tcgen lightmap
        blendfunc filter
    }
}
```

is turned into this:

```c
textures/phillipk2x/computer-pk02_switches01c
{
    qer_editorimage textures/phillipk2x/computer/pk02_switches01c

    q3map_bouncescale 1.25
    normalMap textures/phillipk2x/computer/pk02_switches01c_norm
    glowMap textures/phillipk2x/computer/pk02_switches01c_glow
    specularMap textures/phillipk2x/computer/pk02_switches01c_gloss
    {
        stage diffuseMap
        map textures/phillipk2x/computer/pk02_switches01c
    }
    {
        map $lightmap
        rgbgen identity
        tcgen lightmap
        blendfunc filter
    }
}
```

and this:

```c
textures/map_solarium/water4
{
	qer_editorimage textures/map_solarium/water4/water4.tga
	qer_trans 20
	surfaceparm nomarks
	surfaceparm trans 
	surfaceparm water
	surfaceparm nolightmap
	cull none
	q3map_globaltexture
	tessSize 256

	{
		map textures/map_solarium/water4/water4.tga
		tcmod scale 0.3 0.4
		tcMod scroll 0.05 0.05
		blendfunc add
		alphaGen vertex
	}
		dpreflectcube cubemaps/default/sky
	{
		map $lightmap
		blendfunc add
		tcGen lightmap
	}
	dp_water 0.1 1.2  1.4 0.7  1 1 1  1 1 1  0.1
}
```

into this:

```c
textures/map_solarium/water4
{
	qer_editorimage textures/map_solarium/water4/water4.tga
	qer_trans 20
	surfaceparm nomarks
	surfaceparm trans 
	surfaceparm water
	surfaceparm nolightmap
	cull none
	q3map_globaltexture
	tessSize 256

	normalMap textures/map_solarium/water4/water4_norm
	specularMap textures/map_solarium/water4/water4_gloss
	reflectionMap textures/map_solarium/water4/water4_reflect
	{
		stage diffuseMap
		map textures/map_solarium/water4/water4.tga
		tcmod scale 0.3 0.4
		tcMod scroll 0.05 0.05
		blendfunc add
		alphaGen vertex
	}
	{
		map $lightmap
		blendfunc add
		tcGen lightmap
	}
}
```

The `dpreflectcube` keyword being a stage we can't expect this shader to be properly rendered since that stage line is entirely removed because of the unknown keyword.

illwieckz:
I implemented the detection and auto enablement of heightmap in normalmap alpha channel.

I added a hack that reverses `r_parallaxDepthScale`, it looks to be an engine bug.

illwieckz:
The code looks OK to me now, I added a `r_dpMaterial` cvar that is required to be set to `on` to look for the extra files.

This `dpMaterial` cvar may be reused later to parse DarkPlaces special keywords.

I removed from this PR all the related stuff that was useful for testing purpose but must be done in dedicated PR, like attempt to fix parallax mapping or code to detect heightmap in normalmap.

slipher:
It looks like there are still some commits from one of the other PRs, can you remove those?

illwieckz:
@slipher, yeah, those commits were required for that patch to be applied, but once they have been merged on master they just disappeared from there.

src/engine/renderer/tr_shader.cpp
```diff
@@ -1479,6 +1482,58 @@ static bool ParseClampType( char *token, wrapType_t *clamp )
...
+===================
+FindMapInStage
+===================
+*/
+static bool FindMapInStage( const char **text, char *buffer, int bufferlen )
+{
+	char *token;
+	bool foundMap;
+
+	while ( true )
```
slipher: This function could use some comments explaining what it's looking for or what it's trying to do. Is this the exactly algorithm of Darkplaces or is it some kind of approximation? An example like the ones in the PR comments would be good.

src/engine/renderer/tr_shader.cpp
```diff
@@ -1479,6 +1482,58 @@ static bool ParseClampType( char *token, wrapType_t *clamp )
 	return true;
 }
 
+/*
+===================
+FindMapInStage
+===================
+*/
+static bool FindMapInStage( const char **text, char *buffer, int bufferlen )
```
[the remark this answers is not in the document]
illwieckz: right, and it would make useless the dark magic to not lose the pointer.

src/engine/renderer/tr_shader.cpp
illwieckz: I will add a comment, this part of the code has no chance to be the same algorithm of Darkplaces, the engines are too much different on this part, there is only high level ideas like "for basename look for suffixes” to share. The only part that is probably shareable with Darkplaces are the suffixes themselves…

src/engine/renderer/tr_shader.cpp
```diff
@@ -1479,6 +1482,58 @@ static bool ParseClampType( char *token, wrapType_t *clamp )
...
+{
+	char *token;
+	bool foundMap;
```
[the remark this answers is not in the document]
illwieckz: in fact it was the narrowest possible scope, I just forgot to set it to false, now done.

illwieckz:
all remarks are now fixed

src/engine/renderer/tr_shader.cpp
```diff
@@ -49,6 +49,9 @@ static cullType_t    implicitCullType;
 
 static char          whenTokens[ MAX_STRING_CHARS ];
 
+// DarkPlaces material compatibility
+static Cvar::Cvar<bool> r_dpMaterial("r_dpMaterial", "Enable DarkPlaces material compatibility", Cvar::NONE, false);
```
illwieckz: @slipher, do you think this requires the annoying `#ifdef BUILD_ENGINE ` magic too?

illwieckz: seems not

PR #160, opened 2019-02-09
Add list files command

illwieckz:
[old commit](https://github.com/slipher/Unvanquished/commit/6242b10adeebfaf23ceed8bb3cc385a1347bc0a3) by @slipher initially wrote for kharnov's purpose, was intensively used to clean-up the game packs from dust for 0.51 release. This is also very convenient to check if a file is loaded as expected, like for troubleshooting #159.

Note that some files are loaded dynamically, so be sure to have spawned any character model / weapon and have walked any possible menu to be sure every used file is listed.

I don't know why this was never merged before, is there any issue with it?

I modified the patch a bit to make it working on the now-extracted engine directory, reverted to lowercase `override` and did some camelCase, whitespace and blank line bikeshedding.


DolceTriade:
I feel like some combination of "which" and autocomplete somehow already does this.

slipher:
This is not the only interface for opening a pak file, for example there is also FS::PakPath::ReadFile. So it's not exactly correct.

illwieckz:
To be sure we talk about the same thing: the purpose of this is to only list what's loaded because of it being used, not to list what's available.

illwieckz:
can we merge this?

DolceTriade:
This only handles stuff from the legacy code paths I think?

https://github.com/DaemonEngine/Daemon/blob/3a29a12342f80bd6056e1caa5a002e4d09d42271/src/common/FileSystem.cpp#L710

illwieckz:
I don't know what you are talking about 🤔

illwieckz:
note that we can rename the `listFiles` command as `listLoadesFiles` since the purpose of the PR is to provide a way to only lists files that are loaded because needed, not to list all files available (that would be another command).

illwieckz:
I just renamed from `listFiles` to `listUsedFiles` since it basically prints the `usedFiles` list, and `used` makes more sense than loaded in this context.

slipher:
It only counts files that were opened with `FS_FOpenFileRead`. The engine has other file APIs that don't go through this function, so the results are incomplete (therefore it should not be merged).

illwieckz:
Ah ok now I understand.

PR #161, opened 2019-02-16
responsive: compute console font size from screen size

illwieckz:
This is a PR similar to https://github.com/Unvanquished/Unvanquished/pull/1106 and https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/6

This is to resize the console font size according to screen size so low resolutions don't have huge chars and high resolutions don't have tiny chars.

A `cl_consoleFontScaling` is available for people who wants to disable console font scaling. Making this enabled by default ensures people have a readable console on hiDPI screens.

### DIFF
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -139,6 +139,7 @@ cvar_t                 *cl_packetdelay; //bani
 
 cvar_t                 *cl_consoleFont;
 cvar_t                 *cl_consoleFontSize;
+cvar_t                 *cl_consoleFontScaling;
 cvar_t                 *cl_consoleFontKerning;
 cvar_t                 *cl_consoleCommand; //see also com_consoleCommand for terminal consoles
 
@@ -3030,7 +3031,19 @@ bool CL_InitRenderer()
 	{
 		if ( FS_FOpenFileRead( cl_consoleFont->string, &f, false ) >= 0 )
 		{
-			re.RegisterFont( cl_consoleFont->string, nullptr, cl_consoleFontSize->integer, &cls.consoleFont );
+			if ( cl_consoleFontScaling->value == 0 )
+			{
+				re.RegisterFont( cl_consoleFont->string, nullptr, cl_consoleFontSize->integer, &cls.consoleFont );
+			}
+			else
+			{
+				// This gets 12px on 1920×1080 screen, which is libRocket default for 1em
+				int fontScale = std::min(cls.glconfig.vidWidth, cls.glconfig.vidHeight) / 90;
+
+				// fontScale / 12px gets 1px on 1920×1080 screen
+				re.RegisterFont( cl_consoleFont->string, nullptr, cl_consoleFontSize->integer * fontScale / 12, &cls.consoleFont );
+			}
+
 			cls.useLegacyConsoleFont = false;
 		}
 
@@ -3321,6 +3334,7 @@ void CL_Init()
 
 	cl_consoleFont = Cvar_Get( "cl_consoleFont", "fonts/unifont.ttf",  CVAR_LATCH );
 	cl_consoleFontSize = Cvar_Get( "cl_consoleFontSize", "16",  CVAR_LATCH );
+	cl_consoleFontScaling = Cvar_Get( "cl_consoleFontScaling", "1", CVAR_LATCH );
 	cl_consoleFontKerning = Cvar_Get( "cl_consoleFontKerning", "0", 0 );
 
 	cl_consoleCommand = Cvar_Get( "cl_consoleCommand", "say", 0 );
--- a/src/engine/client/client.h
+++ b/src/engine/client/client.h
@@ -456,6 +456,7 @@ extern cvar_t *cl_altTab;
 
 extern cvar_t *cl_consoleFont;
 extern cvar_t *cl_consoleFontSize;
+extern cvar_t *cl_consoleFontScaling;
 extern cvar_t *cl_consoleFontKerning;
 extern cvar_t *cl_consoleCommand;
 


slipher:
It's a bit odd to have the font size change when the window is resized. Not many applications do that (though come to think of it, the old UI system with the 640x480 coordinate system might be one). Maybe we could consider the total monitor dimensions instead of the window dimensions (but not sure this number can be reliably obtained). If we assume everyone just uses full screen all the time, it doesn't really matter I guess.

illwieckz:
The problem is that to get things properly displayed, and to provide consistent ui, and _consistent interaction_, things have to be at the same place whatever the resolution/size.

Look at Unvanquished 0.51.1 on 4K:

[unvanquished responsive](https://dl.illwieckz.net/b/unvanquished/shots/experiments/responsive//unvanquished_2019-02-16_212213_000.jpg)

But in 1080p it looks somewhat like that:

[unvanquished responsive](https://dl.illwieckz.net/b/unvanquished/shots/experiments/responsive//unvanquished_2019-02-16_212611_000.jpg)

See how the utilisability is absolutely not the same, minimap is super-tiny, the buildable indicator on the left is just very small and feels very far, and the build menu is just hardly usable. In fact, the main problem is not that things are tiny, the main problem is that your mouse path to reach things or your eye path to read things are never the same between two resolutions.

On 1080p I have to mouse down to reach the rocket pod in build menu, on 4K I have to mouse up, that's very boring. And in fact on 1080p each buildable in build menu has its own gesture to get it, on 4K it's « almost the same gesture but with an accurate aim to get the one next to the other », and it's requiring more attention, so it slows down player and tire them quicker. Basically 4K players get handicap.

And things being two tiny is a problem btw, I always miss what people are saying while playing on 4K.

Now this is how looks Unvanquished on 4K (I  just notice now that I forgot to scale one text):

[unvanquished responsive](https://dl.illwieckz.net/b/unvanquished/shots/experiments/responsive//unvanquished_2019-02-16_212542_000.jpg)

This is not an app, it's a game, it's all about _interaction_.

I discovered the need for this work when I bought my 4K screen wand I found that I had to relearn all my ui gestures in Unvanquished but I did not had to relearn them in Xonotic (that properly scales).

illwieckz:
This is how console looked before:

[unvanquished responsive](https://dl.illwieckz.net/b/unvanquished/shots/experiments/responsive//unvanquished_2019-02-16_025842_000.jpg)

[unvanquished responsive](https://dl.illwieckz.net/b/unvanquished/shots/experiments/responsive//unvanquished_2019-02-16_030024_000.jpg)

This is how it looks now:

XGA:  
[unvanquished responsive XGA](https://dl.illwieckz.net/b/unvanquished/shots/experiments/responsive//unvanquished_2019-02-16_211652_000.jpg)

HD1080 (FullHD):  
[unvanquished responsive HD1080](https://dl.illwieckz.net/b/unvanquished/shots/experiments/responsive//unvanquished_2019-02-16_211719_000.jpg)

UHD (~4K):  
[unvanquished responsive UHD](https://dl.illwieckz.net/b/unvanquished/shots/experiments/responsive//unvanquished_2019-02-16_211732_000.jpg)

People with very large screens can still leverage their big screen to display more stuff in console by setting a smaller font size by tweaking `cl_consoleFontSize` (using `seta` and entirely restarting the engine).

slipher:
Of course you are right about the librocket UI needing to scale (4K circle menu is hilarious!). And the fonts within will need to scale together with the other elements. But the console is an independent system which does not suffer from the same issues, being merely a rectangular grid of characters. 

I am not enthused about the fact that following the proposed change, it is impossible to configure the console font to a size that is independent of the window size. I want at least one of the following to be true:

- Console font can be resized without restarting the entire program.
- Pixel-based console font size can be specified.

Maybe you could make positive values of the cvar mean an absolute number of pixels and negative ones be relative or something.

illwieckz:
I agree with the fact we must give people the ability to _not_ scale the console, even if I think we also must provide a way to _scale_ it. I personnaly think that scaling must be the default option because it ensures things are readable in an expected way in any way. This way we know that people can't get weird results unless there is a bug and we know what they must have so we can know if there is a bug.

But I agree that having the ability to disable console scaling to not have manually reduce the scale to get the default size is very OK, and it's better since it only requires to set things from on to off without having to think more about it. We can define a cvar named like `cl_consoleFontScaling` people with advanced need can switch off. Once `cl_consoleFontScaling` would be set to off, `cl_consoleFontSize` becomes a pixel-based console font size people can specify this way.

There is already:

- `cl_consoleFont` _<string>_
- `cl_consoleFontSize` _<number>_
- `cl_consoleFontKerning` _<boolean>_

I think it would be good to add this cvar :

- `cl_consoleFontScaling` _<boolean>_

And that it would be better to set it to `true` so the behavior can be expected (i.e. we can expect how things are rendered on someone else's screen).

__________________

About this:

> Console font can be resized without restarting the entire program.

I would like to see that, because for some time I really thoughts that there was a bug in dæmon since I was seeing no changes while tweaking it, even after a `vid_restart`. I've only managed to get that value read by setting it with `seta` and quitting the engine and restarting it.

If it's possible, we must at least make this change applied on `vid_restart`. That may be another PR since it's another issue and since that issue preexists.

illwieckz:
So, I added a new cvar named `cl_consoleFontScaling` that can be set to `0` to disable console font scaling. By default console font is scaled to make sure the console is properly readable on hiDPI screens.

For some unknown reason, `cl_consoleFontScaling` cvar suffers from the same problem `cl_consoleFontSize` does: it requires the engine to be entirely shut down and started again to get this cvar applied, `vid_restart` is not enough. That other issue is out of scope of this PR.

PR #165, opened 2019-02-22
null default sound sample

illwieckz:
Use `null.wav` as default sound sample instead of `sound/feedback/hit/wav`.

When loading a map with an ambient (looping) sound missing, player just hear an horribly annoying endless `tick tick tick tick` sound. Having a warning or notice in console saying the sound is missing must be enough. No need for a sound feedback.

### DIFF
--- a/src/engine/audio/Sample.cpp
+++ b/src/engine/audio/Sample.cpp
@@ -70,7 +70,7 @@ namespace Audio {
 
     // Implementation of the sample storage
 
-    static const char errorSampleName[] = "sound/feedback/hit.wav";
+    static const char errorSampleName[] = "sound/null";
     static std::shared_ptr<Sample> errorSample = nullptr;
     bool initialized = false;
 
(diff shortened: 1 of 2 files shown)


illwieckz:
I'm not sure there is currently a warning when a sound is missing, typing something like:

```
/playMusic bleh 
```

Does not print any error/warning/notice message.

illwieckz:
But there is code to warn about missing sound files:
https://github.com/DaemonEngine/Daemon/blob/be89425dd92889b0f965d6890a804148b7832711/src/engine/audio/SoundCodec.cpp#L105

So, if it does not work it's a bug and it's out of scope of that PR.

slipher:
So change that log to a `Warning` I guess?

illwieckz:
I just added a commit to make the “missing sound” message a warning.

slipher:
LGTM

Issue #30, opened 2017-09-09
parallax mapping bugs


illwieckz:
I noticed yesterday that XreaL had working parallax mapping, I know Darkplaces (used by Xonotic) has it too. What's missing to have that effect in our maps?

- [ ] ship height maps
- [x] automatically discover height map in normal map alpha channel
- [x] fix the reversed height map bug
- [x] support parallax offset

DolceTriade:
We had (have?) it too. But it was SUPER expensive and no one could use it. In order to use it, we'd need to generate an additional height map for each of our normal maps I think...

Viech:
I recall there is a texture shader keyword that needs to be set in order for the effect to appear. Not sure if height maps needed to be included.

illwieckz:
So, after investigation, this is what I discovered:

- height maps are required (no magic computation at run time)
- height maps are expected to be stored in alpha channel of normal maps
- special keyword is required to be set in shader before the normalMap stage, this keyword is named `parallax` and would have been better named `heightMapInNormalMap`
- engines does not know how to load height map that is not stored in a normal map alpha channel

so, to get parallax in game we need a lot of work to do, btw that's possible.

I discovered that 2010 edition of XreaL shipped heightmaps for ex (external file) and trak5 (in normal map alpha channel), but better than that, I discovered that xonotic ships heightmap for almost every texture they ship, including pk01, pk2, trak5, ex… 99.99% of their heightmaps are stored in normal map alpha channel, without requiring shader keywords. Some rare files (less than 10 in their maps pk3) are stored as external files.

But another great news, they have code (GPLv2+) to generate height maps from normal maps, in three stages:
- [`makenorms.sh`](https://gitlab.com/xonotic/xonotic-maps.pk3dir/blob/master/textures/phillipk1x/makenorms.sh)
- [`fft-normalmap-to-heightmap-soft.sh`](https://gitlab.com/xonotic/xonotic/blob/master/misc/tools/fft-normalmap-to-heightmap-soft.sh)
- [`fft-normalmap-to-heightmap.c`](https://gitlab.com/xonotic/xonotic/blob/master/misc/tools/fft-normalmap-to-heightmap.c)

So, what we can do:

- import their height maps, compute the others

What could be good:

- not requiring any shader keyword to guess the availability of an height map in a normal map file
- optional heightmap loading from dedicated file, perhaps by merging the height map to normal map alpha channel in engine so the internal representation is kept.

So, it requires work, but we can do it.

[...]

PR #170, opened 2019-02-27
WIP: fix heightmap orientation, black is low and white is high

illwieckz:
issue found while experimenting with xonotic assets in #159, see also #30, especially https://github.com/DaemonEngine/Daemon/issues/30#issuecomment-466850340

there is only one heightmap standard: black is low and white is high

xreal was doing white is low and black is high for heightmap in alpha channel of normal map but was doing black is low and white is high for heightmap in loose file which does not make sense

in any way daemon seems to have been forked before the addition of shader keywords to support heightmap loose file so only heightmap in alpha channel is supported at this point

### DIFF
--- a/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
+++ b/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
@@ -965,7 +965,7 @@ void	main()
 	Vts = normalize(Vts);
 
 	// size and start position of search in texture space
-	vec2 S = Vts.xy * -u_DepthScale / Vts.z;
+	vec2 S = Vts.xy * u_DepthScale / Vts.z;
 
 	float depth = RayIntersectDisplaceMap(texNormal, S, u_NormalMap);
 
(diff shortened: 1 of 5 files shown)


slipher:
Maybe it's not so simple. You can see [here](https://github.com/DaemonEngine/Daemon/blob/master/src/engine/renderer/glsl_source/reliefMapping_fp.glsl#L227) some sort of binary search on the heightmap values, which means you would not get the same result after reversing the depth scale and inverting the image. I can't tell which convention it is intended to use.

illwieckz:
Yay, maybe, I tried to import the similar code straight from ioquake3's renderergl2 and it renders the good way without to invert stuff. It also fixed a second bug for free. Will push code later.

Btw there is still another bug that is not fixed yet (or maybe two if I'm wrong by assuming two weirdness I see come from the same bug), but there is maybe more parallax code I can pick…

illwieckz:
I'm wrong, even renderergl2's code is wrong if used properly, the problem seems to have something to do with `SWIZ`.

illwieckz:
I'm closing it, parallax needs far more work than a scaling inversion, and perhaps a rework will just fix the issues by doing things right by working on causes instead of reverting bad results.

PR #172, opened 2019-03-03
fix buggy parenthesis in liquid glsl shader

illwieckz:
introduced in 86d4352d6aad71d66360403264ff4a22cd77f72d

I still don't know why the shader compiler never complained before today


illwieckz:
@gimhael, do you have any idea about why the glsl compiler never failed on that mistake before?

It only started to fail today while I was modifying other lines in that glsl file…

slipher:
Looks like all uses are commented out out by `#if !defined(GLSL_COMPILE_STARTUP_ONLY)`.

Issue #120, opened 2018-08-28
broken water translucency on station15 map


illwieckz:
See the waterfall and the pond (two different shaders):

[![incorrect](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_235232_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_235232_000.jpg)

Note there is also a problem with depth:

[![incorrect](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_223241_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_223241_000.jpg)

[![incorrect](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_223259_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_223259_000.jpg)

[![incorrect](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_235106_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_235106_000.jpg)

This is a third and similarly broken translucent water shader:

[![incorrect](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_235113_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/waterbug/unvanquished_2018-08-28_235113_000.jpg)

illwieckz:
Here is one example of buggy material:

```c
// Crappy water with leaves
textures/station15_custom/water2
{
	qer_editorImage textures/station15_custom_src/leaves_d
	qer_trans .3
	q3map_globaltexture
	entityMergable
	surfaceparm nobuild
	surfaceparm trans
	surfaceparm water
	surfaceparm nonsolid
	surfaceparm nolightmap
	cull disable
	tesssize 64
	deformVertexes wave 64 sin .8 1.5 2 .8
	{
		map textures/station15_custom_src/water_d
		rgbGen const ( .8 .8 .3 )
		alphaGen const .2
		blendfunc blend
[...]

PR #173, opened 2019-03-03
glsl: deduplicate RayIntersectDisplaceMap

illwieckz:
`RayIntersectDisplaceMap` is already defined in `reliefMapping_fp.glsl`, doing the exact same thing:

https://github.com/DaemonEngine/Daemon/blob/9322bda72627da2846310f5e4f624449b22122bf/src/engine/renderer/glsl_source/reliefMapping_fp.glsl#L203-L253

### DIFF
--- a/src/engine/renderer/glsl_source/liquid_fp.glsl
+++ b/src/engine/renderer/glsl_source/liquid_fp.glsl
@@ -51,60 +51,6 @@ IN(smooth) vec3		var_Normal;
 
 DECLARE_OUTPUT(vec4)
 
-#if defined(USE_PARALLAX_MAPPING)
-float RayIntersectDisplaceMap(vec2 dp, vec2 ds)
-{
-	const int linearSearchSteps = 16;
-	const int binarySearchSteps = 6;
-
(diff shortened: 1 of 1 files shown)


slipher:
It would be good to make sure that the liquid shader is actually used first (see my comment on #172).

illwieckz:
If the liquid shader is not used, it would explain #120

illwieckz:
Hmm no, commenting out the `#if !defined( GLSL_COMPILE_STARTUP_ONLY )` test does not fix that water issue, this is probably not related.

slipher:
I checked my copy of station15 1.1 and it doesn't actually use the `liquidMap` keyword, which is required to activate these code paths. So no help in showing whether the code is used or works. Do you know some map that actually uses a `liquidMap` shader stage?

illwieckz:
I think no one map use liquidStage yet. Like parallax, it's maybe entirely broken no one knows…
By the way the change is very straightforward, in any way it's better to have a cleaner code even if it's a dead code. This is to be sure any fix done to that parallax code is forwarded everywhere it is used. By deduplicating we make sure a potentially fixed bug will not resurface the day this code will be live again.

PR #174, opened 2019-03-03
server: 64 bytes are not enough to fit a pak name, client already expects 256, fix #167

### LINKED ISSUE #167: autodownload broken if pak name larger than 64 bytes
I noticed that the feature of loading an extrapak in a subdirectory now breaks the autodownload mechanism, basically the pak name does not look complete.

We were able to do before:

```
./daemon -set fs_extrapaks dev/backport
```

We are only able to do now:

```
./daemon -set fs_extrapaks backport
```

It may have been broken in #14 (effort to rewrite the TimePath's original effort to support legacy pk3).

illwieckz:
See #167,  `downloadName` is set to `256` in client but `64` in server, and in anyway, 64 is not large enough.

https://github.com/DaemonEngine/Daemon/blob/9322bda72627da2846310f5e4f624449b22122bf/src/engine/qcommon/q_shared.h#L199-L200
https://github.com/DaemonEngine/Daemon/blob/9322bda72627da2846310f5e4f624449b22122bf/src/engine/client/client.h#L344
https://github.com/DaemonEngine/Daemon/blob/9322bda72627da2846310f5e4f624449b22122bf/src/engine/server/server.h#L171
https://github.com/DaemonEngine/Daemon/blob/9322bda72627da2846310f5e4f624449b22122bf/src/engine/server/sv_ccmds.cpp#L143

### DIFF
--- a/src/engine/server/server.h
+++ b/src/engine/server/server.h
@@ -168,7 +168,7 @@ struct client_t
 	char           name[ MAX_NAME_LENGTH ]; // extracted from userinfo, high bits masked
 
 	// downloading
-	char          downloadName[ MAX_QPATH ]; // if not empty string, we are downloading
+	char          downloadName[ MAX_OSPATH ]; // if not empty string, we are downloading
 	FS::File*     download; // file being downloaded
 	int           downloadSize; // total bytes (can't use EOF because of paks)
 	int           downloadCount; // bytes sent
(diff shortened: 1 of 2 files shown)


illwieckz:
I tested it locally, it works, at least with udp download.

It looks like we are lucky because it seems that an unmodified client is still compatible with a fixed server.

PR #175, opened 2019-03-03
bikeshedding: just a bunch of some annoying things that bore me

illwieckz:
Just a bunch of some annoying things that bore me.

Among them, there is three files that have a useless indent for 95~99% of their length…

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -191,32 +191,32 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 // max. 16 dynamic lights per plane
 #define LIGHT_PLANES ( MAX_REF_LIGHTS / 16 )
 
-	enum class renderSpeeds_t
-	{
-	  RSPEEDS_GENERAL = 1,
-	  RSPEEDS_CULLING,
-	  RSPEEDS_VIEWCLUSTER,
-	  RSPEEDS_LIGHTS,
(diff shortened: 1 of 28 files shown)


src/engine/renderer/tr_local.h
```diff
@@ -191,32 +191,32 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
...
-		GLDEBUG_NONE,
-		GLDEBUG_ERROR,
-		GLDEBUG_DEPRECIATED,
-		GLDEBUG_UNDEFINED,
-		GLDEBUG_PORTABILITY,
-		GLDEBUG_PERFORMANCE,
-		GLDEBUG_OTHER,
-		GLDEBUG_ALL
-	};
+enum class renderSpeeds_t
```
DolceTriade: why is the indentation different here and the enum class below?

slipher:
Well I don't know, the indents are annoying, but so is not being able to use blame...

src/engine/renderer/tr_local.h
illwieckz: I'm not sure to understand what you are saying, btw I just removed one extra level of indentation when it was useless, no more, there is no one effort to check if the remaining indentation is consistent or not. This is just to avoid having thousands of lines having an extra indentation for nothing.

illwieckz:
this is absolutely obsolete, has to be redone from scratch, then closing it for the moment.

PR #176, opened 2019-03-03
FileSystem: implement symlink support in zip archives, fix #83

### LINKED ISSUE #83: symbolic links unsupported in archives
It looks like the engine does not support symbolic links in archives, Xonotic makes use of that, for example:

```
Warn: R_LoadDDSImage: invalid dds header "dds/textures/trak5x/floor/floor_floor2a.dds" 
Warn: R_FindImageFile could not find image 'textures/trak5x/floor/floor_floor2a.tga' in shader 'textures/trak5x/floor-floor2a' 
```

The `dds/textures/trak5x/floor/floor_floor2a.dds` is in fact a symbolic link to a file named `floor_floor2a_gloss.dds` in the same directory.

illwieckz:
- original C++ and BSD zip symlink resolver
- only load zip symlink from the same pak
- optional fs_pakSymlink cvar to enable the feature (disabled by default)

See Darkplaces reference implementation:
https://gitlab.com/xonotic/darkplaces/commit/1c55fd95ff34d9aafb76c927215ea8fbab5210c6
No code from Darkplaces is copypasted to not taint with GPLv2+.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -58,6 +58,15 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <mach-o/dyld.h>
 #endif
 
+#define MAX_SYMLINK_LEVELS 16
+// see https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/stat.h
+// redefine so it works outside of Unices
+#define DAEMON_S_IFMT       00170000
+#define DAEMON_S_IFLNK      0120000
+#define DAEMON_S_ISLNK(m)   (((m) & DAEMON_S_IFMT) == DAEMON_S_IFLNK)
+// see https://trac.edgewall.org/attachment/ticket/8919/ZipDownload.patch
+#define PKZIP_EXTERNAL_ATTR_FILE_TYPE_SHIFT 16
+
 Log::Logger fsLogs(VM_STRING_PREFIX "fs", "[FS]", Log::Level::NOTICE);
 
 // SerializeTraits for PakInfo/LoadedPakInfo
@@ -124,16 +133,27 @@ namespace FS {
 
 #ifdef BUILD_ENGINE
 static Cvar::Cvar<bool> fs_legacypaks("fs_legacypaks", "Also load pk3s, ignoring version.", Cvar::NONE, false);
+static Cvar::Cvar<bool> fs_pakSymlink("fs_pakSymlink", "Resolve symbolic links in paks the DarkPlaces way", Cvar::NONE, false);
 
 bool UseLegacyPaks()
 {
 	return *fs_legacypaks;
+}
+bool UsePakSymlink()
+{
+	return *fs_pakSymlink;
 }
 #else
 bool UseLegacyPaks()
 {
 	bool result = false;
 	Cvar::ParseCvarValue(Cvar::GetValue("fs_legacypaks"), result);
+	return result;
+}
+bool UsePakSymlink()
+{
+	bool result = false;
+	Cvar::ParseCvarValue(Cvar::GetValue("fs_pakSymlink"), result);
 	return result;
 }
 #endif
@@ -918,6 +938,20 @@ public:
 		return fileInfo.uncompressed_size;
 	}
 
+	// Check if currently open file is a symbolic link
+	bool IsSymlink(std::error_code& err) const
+	{
+		unz_file_info64 fileInfo;
+		int result = unzGetCurrentFileInfo64(zipFile, &fileInfo, nullptr, 0, nullptr, 0, nullptr, 0);
+		if (result != UNZ_OK) {
+			SetErrorCodeZlib(err, result);
+			return false;
+		}
+		ClearErrorCode(err);
+		uLong attr = fileInfo.external_fa >> PKZIP_EXTERNAL_ATTR_FILE_TYPE_SHIFT;
+		return ( DAEMON_S_ISLNK(attr) );
+	}
+
 	// Read from the currently open file
 	size_t ReadFile(void* buffer, size_t length, std::error_code& err) const
 	{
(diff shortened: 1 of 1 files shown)


illwieckz:
It produces useful debug log like that:

```
Debug: found symlink: dds/textures/trak5x/floor/floor_floor2a.dds → dds/textures/trak5x/floor/floor_floor2a_gloss.dds 
```

illwieckz:
This trick is meant to be supported even on systems that do not provide `S_ISLNK`:

```c
		uLong attr = fileInfo.external_fa >> 16;
		return ( ( attr & 0120000 ) == 0120000 );
```

illwieckz:
Note that I can also import `FS_CheckNastyPath` from DarkPlaces, it may be useful, and not only for dpk/pk3:
https://gitlab.com/xonotic/darkplaces/blob/cf8dfc912d76d0a26c5d80202fb0b8089df0901f/fs.c#L2439-2494

src/common/FileSystem.cpp
```diff
@@ -1244,7 +1258,71 @@ const std::vector<LoadedPakInfo>& GetLoadedPaks()
...
+// code adapted from DarkPlaces fs.c FS_OpenReadFile
+// commit 1c55fd95ff34d9aafb76c927215ea8fbab5210c6
+// Copyright 2008 divverent
+// GPLv2+
+std::string expandSymlink(Str::StringRef path, Str::StringRef content)
+{
+		const char *filename = path.c_str();
+
+		char linkbuf[MAX_QPATH];
+		strncpy(linkbuf, content.c_str(), MAX_QPATH);
```
DolceTriade: Should we be using Q_strncpyz?

src/common/FileSystem.cpp
```diff
@@ -909,6 +909,20 @@ class ZipArchive {
...
+	{
+		unz_file_info64 fileInfo;
+		int result = unzGetCurrentFileInfo64(zipFile, &fileInfo, nullptr, 0, nullptr, 0, nullptr, 0);
+		if (result != UNZ_OK) {
+			SetErrorCodeZlib(err, result);
+			return false;
+		}
+		ClearErrorCode(err);
+		uLong attr = fileInfo.external_fa >> 16;
+		return ( ( attr & 0120000 ) == 0120000 );
```
DolceTriade: Can we have named constants for these magic numbers?

src/common/FileSystem.cpp
```diff
@@ -909,6 +909,20 @@ class ZipArchive {
 		return fileInfo.uncompressed_size;
 	}
 
+	// Check if currently open file is a symbolic link
+	bool CheckSymlink(std::error_code& err) const
```
DolceTriade: Rename function to IsSymlink() to be more clear. CheckSymlink sounds like its checking for validity.

src/common/FileSystem.cpp
```diff
@@ -1298,17 +1376,46 @@ std::string ReadFile(Str::StringRef path, std::error_code& err)
...
+
 		return out;
 	}
 
 	ASSERT_UNREACHABLE();
 }
 
+std::string ReadFile(Str::StringRef path, std::error_code& err)
+{
+	std::string out = ReadFile(path, 16, err);
```
DolceTriade: named constant for this number. MAX_SYMLINK_LEVELS or something.

src/common/FileSystem.cpp
```diff
@@ -1244,7 +1258,71 @@ const std::vector<LoadedPakInfo>& GetLoadedPaks()
...
+			mergestart += 3;
+			while(mergeslash > filename) {
+				--mergeslash;
+				if(*mergeslash == '/') {
+					break;
+				}
+			}
+		}
+		// Now, mergestart will point to the path to be appended, and mergeslash points to where it should be appended
+		if(mergeslash == filename) {
```
DolceTriade: where is this appending done?

src/common/FileSystem.cpp
```diff
@@ -1244,7 +1258,71 @@ const std::vector<LoadedPakInfo>& GetLoadedPaks()
...
+		const char *filename = path.c_str();
+
+		char linkbuf[MAX_QPATH];
+		strncpy(linkbuf, content.c_str(), MAX_QPATH);
+		int count = strlen(linkbuf);
+
+		const char *mergeslash;
+		char *mergestart;
+
+		// Now combine the paths...
```
DolceTriade: I don't seem to understand this code... Shouldn't it find the last occurrence of '/' and then do the relative directory handling instead of the first occurrence? I'd really prefer if this was broken out in terms of functions like to make this more readable. With all the pointer operations and magic numbers, it's hard to decipher.

illwieckz: I'm not sure to understand all that c magic either (but I'm confident in divverent's work), I would appreciate myself a clean c++ rewrite using c++ strings…

src/common/FileSystem.cpp
illwieckz: probably

src/common/FileSystem.cpp
illwieckz: yep for the idea, I'm still looking for a proper name for these magic numbers (especially 16 which I not found in Darkplaces…)

src/common/FileSystem.cpp
illwieckz: I agree

src/common/FileSystem.cpp
illwieckz: yep

slipher:
This appears to allow symlinks to refer to arbitrary files anywhere in the filesystem which seems really gross to me. If symlinks are really needed, they should only be able to refer to files in the same zip package.

illwieckz:
I updated the code with constants, and made the feature disabled by default.
A cvar named `fs_pakSymlink` can be set to `true` to enable the feature.

@slipher wrote:
> This appears to allow symlinks to refer to arbitrary files anywhere in the filesystem which seems really gross to me. If symlinks are really needed, they should only be able to refer to files in the same zip package.

Basically this feature has only one purpose: provide backward compatibility with Darkplaces, and unless I'm wrong it looks like Darkplaces is doing it the cross-pk3 way, but perhaps I missed something. Xonotic uses this feature to deduplicate files: since they rely on automatic normal/specular/glow map detection based on file name, they _have_ to provide a working _filename_ for every texture, the also deduplicate sounds and other stuff. Basically they have almost 2K symlinks in their pk3. I have not yet checked if they were really using them accross pk3 or not.

slipher:
How would it make sense to have e.g. the normalmap and the main texture in different packages? The Darkplaces implementation may permit that, but it's a misfeature that should be avoided if at at all possible.

illwieckz:
The code now only looks for symlink in the same pak.
As an extra benefit it would load the zip only once.

src/common/FileSystem.cpp
```diff
@@ -122,21 +131,13 @@ template<> struct SerializeTraits<FS::LoadedPakInfo> {
 
 namespace FS {
 
-#ifdef BUILD_ENGINE
```
slipher: Bad merge?

src/common/FileSystem.cpp
```diff
@@ -1253,6 +1268,143 @@ const std::vector<LoadedPakInfo>& GetLoadedPaks()
...
+			return "";
+		}
+		*/
+
+		return mergestart;
+}
+
+std::string ReadZipFile(LoadedPakInfo pak, ZipArchive *zipFile, Str::StringRef originalPath, Str::StringRef path, int symlinkLevels, std::error_code& err)
+{
+	auto it = fileMap.find(path);
```
slipher: `fileMap` is a global variable containing files for all paks so it doesn't make sense to be accessing it for resolving links within one zip archive.

src/common/FileSystem.cpp
illwieckz: Probably, now fixed. Btw that looks overly complicated… :-/

src/common/FileSystem.cpp
illwieckz:
```c
// Map of filenames to pak files. The size_t is an offset into loadedPaks and
// the offset_t is the position within the zip archive (unused for PAK_DIR).
static std::unordered_map<std::string, std::pair<uint32_t, offset_t>> fileMap;
```

`fileMap.find(path)->second.second` is the offset to the content of the file with `path` in zip archive.

I haven't found a better way to load a file content without reparsing the zip again.

This is problematic if findMap returns a file that matches the path in another pak. I don't know if that can happen.

illwieckz:
So, to sum up some talks. My first motivation into this and into any Darkplaces compatibility layer is to use Xonotic maps as test bed for Dæmon, given

1. they were not tested on Dæmon at design time hence not designed to workaround Dæmon bugs;
2. they are very polished ones with a whole community able to contribute to refine every detail (they have public map repositories and on-commit build bot since years);
3. they rely on features we don't use yet and that are likely to be broken (parallax, water…);
4. they rely on features we don't implement yet (sRGB…).

Everything of this make Xonotic maps a very good test bed, and probably the better one available out there, this test bed breaks Dæmon in many ways and sets a minimum feature and quality level for Dæmon to target. I'm intimately convinced that Dæmon must reach graphical feature parity and graphical quality parity with Darkplaces, not because of Xonotic but because of Unvanquished: we need a better engine. In the past Dæmon looked sexy but in 2019 it just looks like Quake3 with normal maps.

On the zip symlink topic, after a deep thought I now think that it's a bad idea to restrain the zip symlink to one pak, in fact it defeat the whole purpose of pk3/dpk: being able to overload an existing vfs. I'm also intimately convinced that putting symlinks in pk3 was a bad idea at first, but that's too late. It's still a very efficient solution since pkzip format is not a solid archive and does not deduplicate, but it's meh. So, the purpose of this PR is not to encourage people to symlink stuff neither to say it's a good idea. The purpose of this PR is to make Dæmon able to load the best test bed we can put our hands on.

That's also why I think that this PR makes no sense if it does not target 1:1 behavior. That's also why, with the cross-pak symlink support, I ported the Darkplaces code to solve those symlinks. The purpose of this PR is not to provide symlink support to Dæmon based games (my opinion is to seriously discourage this), the purpose of this PR is to provide the test bed map the algorithm they expect. So, even if I would like to see a more c++ey symlink solver, I think it's better to keep it at it, because we know the test bed was built to work on that algorithm.

Note that we may display a warning just telling people that pak symlink is enabled, just to tell that's creepy.

illwieckz:
But I'm very concerned by the addition of a GPLv2 code into a BSD one (the symlink resolver). That's enough to me to ask for a rewrite.

illwieckz:
I tried to do the symlink stuff within the `zipFile` class, then after a lot of trouble I just thought it was a bad idea from the start. What makes me think it's a bad idea is that it would be a very bad design choice to implement this whithin the `zipFile` code _if we were supporting more than one archive format_. So, even if we only support one archive format, it looks like doing it in the pak abstraction is the best place.

It looks like `fileMap` is a `std::unordered_map` which has unique keys. Hence testing if the pak providing the file listed in `fileMap` is the same pak containing the symlink does the expected job: it will not bypass `DEPS`.

src/common/FileSystem.cpp
illwieckz: `findMap` references a file that matches the path in another pak if the file is only available in the other pak (or take precedence with DEPS), so accessing `findMap` for that purpose is safe.

It maybe not very efficient, but it's OK for a compatibility layer. Doing this compute in zip internals is likely to add hard to read and maintain code. The need for a compatibility layer is probably not worth messing with zip internals.

src/common/FileSystem.cpp
illwieckz: this was entirely rewritten

src/common/FileSystem.cpp
illwieckz: this was entirely rewritten

illwieckz:
I added extra comments to explain how it works.

slipher:
Alternative implementation: #182

illwieckz:
#182 looks better so I'm closing this one.

PR #177, opened 2019-03-04
export game surfaceflags to CustomSurfaceFlags.h, fix #107

### LINKED ISSUE #107: surfaceflags are defined in engine code, not in game code
See: [src/engine/qcommon/surfaceflags.h](https://github.com/DaemonEngine/Daemon/blob/master/src/engine/qcommon/surfaceflags.h)

They contains Unvanquished-specific code like `CONTENTS_NOALIENBUILD`. Those flags are game-specific, for example the value for Unvanquished's nobuild surfaceflag is known to be the value for the Urban Terror metalstep surface.

illwieckz:
This is an easy way to fix #107

I just noticed that those surfaceflags either add new values, either replace existing ones, like game's `CONTENTS_NOBUILD` sharing the `0x00004000` value with engine's `CONTENTS_MOVER`.

So in any way, games just want to add values or to override one to do ugly tricks like our game does.

If one day games want a more elaborated behavior, it would be still time to elaborate more.

See also Unvanquished/Unvanquished#1107 for related PR on the game side.


slipher:
LGTM

illwieckz:
That header file was entirely rewritten from scratch, allowing the new file to be licensed under a BSD license, which is better for a code that is expected to be shared with game code (reduce the risk of license incompatibilities between engine and game).

PR #178, opened 2019-03-05
R_FindCubeImage: avoid the boring “Failed to open” on skybox format try

illwieckz:
This is to avoid those boring, false-positive debug messages:

```
Debug: Failed to open 'env/shared_vega_src/sky' for reading: No such file or directory
Debug: Failed to open 'env/shared_vega_src/sky' for reading: No such file or directory
```

Note that the `R_FindImageLoader` code (first two commits) is the one I first pushed in #159.
If first worked on `R_FindImageLoader` for the need of the removal of this debug message, but working on Xonotic stuff was more exciting. :stuck_out_tongue: 

This is better to merge this first, as the image commits will be together, and then the dpMaterial PR would just be 1 commit.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -1613,6 +1613,58 @@ static int                   numImageLoaders = ARRAY_LEN( imageLoaders );
 
 /*
 =================
+R_FindImageLoader
+
+Finds and returns an image loader for a given basename,
+tells the extra prefix that may be required to load the image.
+=================
+*/
+int R_FindImageLoader( char *baseName, const char **prefix ) {
+	const FS::PakInfo* bestPak = nullptr;
+	int i;
+
+	// Darkplaces or Doom3 packages can ship alternative texture path in the form of
+	//   dds/<path without ext>.dds
+	std::string altName = Str::Format( "dds/%s.dds", baseName );
+	bestPak = FS::PakPath::LocateFile( altName );
+
+	// If this alternative path exists, it's expected to be loaded as the best one
+	// except when it goes against Daemon's rule to load the hardcoded one if exists
+	// because this dds alternative is only supported for compatibility with
+	// third-party content
+	if ( bestPak != nullptr ) {
+		for ( i = 0; i < numImageLoaders; i++ ) {
+			if ( !Q_stricmp( "dds", imageLoaders[i].ext ) ) {
+				*prefix = "dds/";
+				return i;
+			}
+		}
+	}
+	
+	int bestLoader = -1;
+	*prefix = "";
+	// try and find a suitable match using all the image formats supported
+	// prioritize with the pak priority
+	for ( i = 0; i < numImageLoaders; i++ )
+	{
+		std::string altName = Str::Format( "%s.%s", baseName, imageLoaders[i].ext );
+		const FS::PakInfo* pak = FS::PakPath::LocateFile( altName );
+
+		// We found a file and its pak is better than the best pak we have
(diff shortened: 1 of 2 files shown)


src/engine/renderer/tr_image.cpp
```diff
@@ -1978,22 +2013,25 @@ image_t        *R_FindCubeImage( const char *imageName, int bits, filterType_t f
...
+	char cubeMapBaseName[ MAX_QPATH ];
 
-	// try to load .KTX cubemap
-	LoadKTX( buffer, pic, &width, &height, &numLayers, &numMips, &bits, 0 );
-	if( numLayers == 6 && pic[0] ) {
-		numPicsToFree = 1;
-		goto createCubeImage;
-	} else {
-		R_FreeCubePics( pic, numLayers );
+	COM_StripExtension3( buffer, cubeMapBaseName, MAX_QPATH );
```
slipher: It's better to use `sizeof(cubeMapBaseName)` since that's more resistant to the buffer size being changed later.

src/engine/renderer/tr_image.cpp
```diff
@@ -1978,22 +2013,25 @@ image_t        *R_FindCubeImage( const char *imageName, int bits, filterType_t f
...
+
+	for ( i = 0; i < numCubeMapLoaders; i++ )
+	{
+		cubeMapName = va( "%s.%s", cubeMapBaseName, cubeMapLoaders[ i ].ext );
```
slipher: `std::string cubeMapName = Str::Format(...)`

src/engine/renderer/tr_local.h
```diff
@@ -3150,6 +3150,8 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	void    R_InitImages();
 	void    R_ShutdownImages();
 
+	int R_FindImageLoader( char *filename );
```
slipher: These functions are only used in the same file, so you can remove them from the header and declare them `static`.

src/engine/renderer/tr_image.cpp
```diff
@@ -1611,6 +1611,58 @@ static const imageExtToLoaderMap_t imageLoaders[] =
...
 
+/*
+=================
+R_FindImageLoader
+
+Finds and returns an image loader for a given basename,
+tells the extra prefix if required to load the image.
+=================
+*/
+int R_FindImageLoader( char *baseName, const char **prefix ) {
```
slipher: `baseName` can be a `Str::StringRef` (or `const char*` at the least).

src/engine/renderer/tr_image.cpp
```diff
@@ -1682,57 +1734,24 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
...
-		return;
-	}
-
-	// try and find a suitable match using all the image formats supported
-	// prioritize with the pak priority
-	for ( i = 0; i < numImageLoaders; i++ )
-	{
-		std::string altName = Str::Format("%s.%s", filename, imageLoaders[i].ext);
-		const FS::PakInfo* pak = FS::PakPath::LocateFile(altName);
+	// loader failed, most likely because the file isn't there;
```
slipher: The comment doesn't make sense here since this path can be reached without a loader failing.

src/engine/renderer/tr_local.h
illwieckz: The declaration without `prefix` will be used from somewhere else by code in #159 
I may remove the other one with the `prefix`.

illwieckz: I'll made it exportable later, when needed.

PR #179, opened 2019-03-05
revamp cubemap code and fix quake3 skybox with small bottom face, fix #142

### LINKED ISSUE #142: broken skybox when one of the face have a different size than the others
This a bug I first faced while porting maps from Tremulous for my [InterstellarOasis](https://github.com/InterstellarOasis/InterstellarOasis) initiative. At this time I workarounded the bug by modifying the skyboxes, but it's just a one-shot workaround and not a fix.

So, what's the issue? Sometime some skyboxes have one face with a different size than the others, usually the bottom one is smaller since it's unseen from the player point of view because of the floor of the map itself. Quake3 mods looked to rely on this trick to save file space (especially when pictures where shipped uncompressed and bandwith was low).

Have a look at the _alamo_ skybox by Joe Kari from Smokin' Guns:

[smaller face](https://dl.illwieckz.net/b/smokinguns/screenshots/bugs/mixed-skybox/20181123-212844-000.smokinguns.alamo-sky.png)

The skybox rendering is broken:

[broken alamo skybox](https://dl.illwieckz.net/b/smokinguns/screenshots/bugs/mixed-skybox/daemon_2018-11-23_225340_000.jpg)

To reproduce the issue it's possible to download some preview dpks of Smokin' Guns assets this way:

```sh
wget -O- http://gg.illwieckz.net/dl/smokinguns/dev/SmokinGunsAssets.get | xargs wget
```

and load the `dm-alamo` map this way:

```
/set g_neverEnd 1
/devmap dm-alamo
```

There is no such issue in idtech3 derivated games like Tremulous or Smokin' Guns so this looks to be a regression (it can even predate Unvanquished if introduced by XreaL/ET:XreaL, I don't know).

illwieckz:
## Skybox code revamp

The skybox code was very ugly with 10 gotos and buggy behavior, for example it would not attempt to load any multifile skybox if a single-file skybox was successfully loaded but had invalid content.

Also, the code was inefficient, for example it was doing the image rotation operation for every face even if the rotation has an angle of 0°, hence allocating a new image, copying all the pixels in the new image et. for nothing.

## Irregular skybox fixing

Some quake3 skyboxes were shipped with a smaller bottom face, for example a `16×16` bottom face while other faces were `512×512` that was probably a trick to save file size when textures were stored losslessly in a tga and bandwith was low as voice modems.

To fix this, a new `R_Resize` function is added to `tr_image.cpp` that is just a wrapper to `ResampleTexture` that makes in-place resizing easier.

The faces are resized to the size of a square that has for width and height the greater width and height found in any face. This way it would not only fix the quake3 skyboxes with smaller bottom faces, but also non-square skyboxes.

A warning is displayed so the user is encouraged to fix the skybox because 1. we workaround something that is broken, 2. user may is likely to have access to tools with resize algorithms that produce better looking results.

In any way, those kind of small bottom faces were likely to not be seen, that's also why they were downscaled at first.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -2050,9 +2050,15 @@ static const multifileCubeMapFormat_t multifileCubeMapFormats[] =
 	},
 };
 
+struct face_t
+{
+	int width;
+	int height;
+};
+
 image_t *R_FindCubeImage( const char *imageName, int bits, filterType_t filterType, wrapType_t wrapType )
 {
-	int i;
+	int i, j;
 	image_t *image = nullptr;
 	int width = 0, height = 0, numLayers = 0, numMips = 0;
 	byte *pic[ MAX_TEXTURE_MIPS * MAX_TEXTURE_LAYERS ];
@@ -2105,6 +2111,9 @@ image_t *R_FindCubeImage( const char *imageName, int bits, filterType_t filterTy
 
 	for ( const multifileCubeMapFormat_t format : multifileCubeMapFormats )
 	{
+		int greatestEdge = 0;
+		face_t faces[6];
+
 		for ( i = 0; i < 6; i++ )
 		{
 			Com_sprintf( filename, sizeof( filename ), "%s_%s", buffer, format.suffixes[ i ] );
@@ -2133,32 +2142,68 @@ image_t *R_FindCubeImage( const char *imageName, int bits, filterType_t filterTy
 				break;
 			}
 
-			if ( width != height )
+			if ( width > greatestEdge )
 			{
-				Log::Warn("cubemap face '%s' is not a square with %d×%d dimension", filename, width, height);
-				break;
-			}
-			
-			if ( format.flipX[ i ] )
-			{
-				R_Flip( pic[ i ], width, height );
+				greatestEdge = width;
 			}
 
-			if ( format.flipY[ i ] )
+			if ( height > greatestEdge )
 			{
-				R_Flop( pic[ i ], width, height );
+				greatestEdge = height;
 			}
 
-			if ( format.rot[ i ] != 0 )
-			{
-				R_Rotate( pic[ i ], width, height, format.rot[ i ] );
-			}
+			faces[ i ].width = width;
+			faces[ i ].height = height;
 		}
 
 		if ( i == 6 )
 		{
+
+			for ( j = 0; j < 6; j++ )
+			{
+				width = faces[ j ].width;
+				height = faces[ j ].height;
+
+				bool badSize = false;
+
+				if ( width != height )
+				{
+					Log::Warn("cubemap face '%s_%s' is not a square with %d×%d dimension, resizing to %d×%d",
+						imageName, format.suffixes[ j ], width, height, greatestEdge, greatestEdge );
+					badSize = true;
+				}
+
+				if ( width < greatestEdge || height < greatestEdge )
+				{
+					Log::Warn("cubemap face '%s_%s' is too small with %d×%d dimension, resizing to %d×%d",
+						imageName, format.suffixes[ j ], width, height, greatestEdge, greatestEdge );
+					badSize = true;
+				}
+
+				// make face square before doing other operations like rotation
+				if ( badSize )
+				{
+					pic[ j ] = R_Resize( pic[ j ], width, height, greatestEdge, greatestEdge );
+				}
+
+				if ( format.flipX[ j ] )
+				{
+					R_Flip( pic[ j ], width, height );
+				}
+
+				if ( format.flipY[ j ] )
+				{
+					R_Flop( pic[ j ], width, height );
+				}
+
+				if ( format.rot[ j ] != 0 )
+				{
+					R_Rotate( pic[ j ], width, height, format.rot[ j ] );
+				}
+			}
+
 			Log::Debug( "found %s multifile cube map '%s'", format.name, imageName );
-			image = R_CreateCubeImage( ( char * ) buffer, ( const byte ** ) pic, width, height, bits, filterType, wrapType );
+			image = R_CreateCubeImage( ( char * ) buffer, ( const byte ** ) pic, greatestEdge, greatestEdge, bits, filterType, wrapType );
 			R_FreeCubePics( pic, i );
 			return image;
 		}
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -1959,6 +1959,30 @@ static void R_Rotate( byte *in, int width, int height, int degrees )
 }
 
 /*
+========
+R_Resize
+
+wrapper for ResampleTexture to help to resize a texture in place like this:
+pic = R_Resize( pic, width, height, newWidth, newHeight );
+
+please not resize normalmap with this, use ResampleTexture directly instead
+
+========
+*/
+
+byte *R_Resize( byte *in, int width, int height, int newWidth, int newHeight )
+{
+
+	byte *out;
+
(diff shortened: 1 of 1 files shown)


illwieckz:
Example of such skybox (_alamo_, for the Smokin' Guns game):

[smaller face in alamo skybox](https://dl.illwieckz.net/b/smokinguns/screenshots/bugs/mixed-skybox/20181123-212844-000.smokinguns.alamo-sky.png)

Before:

[broken alamo skybox](https://dl.illwieckz.net/b/smokinguns/screenshots/bugs/mixed-skybox/daemon_2018-11-23_225340_000.jpg)

After:

[fixed alamo skybox](https://dl.illwieckz.net/b/unvanquished/bugs/ugly-skybox/unvanquished_2019-03-05_080831_000.jpg)

[fixed alamo skybox](https://dl.illwieckz.net/b/unvanquished/bugs/ugly-skybox/unvanquished_2019-03-05_080845_000.jpg)

illwieckz:
Note that this issue was first faced with Tremulous maps, so the issue is very likely to occurs because that was a supported feature of Tremulous.

g-bougard:
Hi @illwieckz 
the fix seems fine to me.
Btw you should not mix tab/space indentation in your patch.

megatog615:
Why not examine the sizes of all sides and scale them all to the biggest just in case there exists a skybox with this trick done on another side than the bottom? You could even have the algorithm skip sides that already equal the detected size.

illwieckz:
oh right, there was some boring spaces, fixed!

illwieckz:
@megatog615 I thought about it but it would require some extra code for an occurrence that will probably never happen… probably not worth it.

illwieckz:
I just thought about a way to check on all faces while not being so much intrusive so I added a commit trying that instead.

By the way even if it may look possible to do this for all kind of supported cubemap, I really doubt this kind of feature exists outside of quake3.

src/engine/renderer/tr_image.cpp
```diff
@@ -2076,6 +2107,11 @@ image_t        *R_FindCubeImage( const char *imageName, int bits, filterType_t f
 	}
 	numPicsToFree = 0;
 
+	// no need to store greaterHeight since cubmap face is a square
```
slipher: I don't see anything forcing the image to be a square. `R_CreateCubeImage` takes both width and height arguments.

illwieckz: The code returns earlier if `width ≠ height`:

https://github.com/DaemonEngine/Daemon/blob/7774ff80c78f2083e23027c7a24561f07b64cf9f/src/engine/renderer/tr_image.cpp#L2092-L2096

And in anyway, a cube is a regular hexahedron, which means all faces are equal squares. :wink: 

It's possible to also fix any potential non-cube hexahedron, btw. reading this code with all those goto and stuff just make me thing the whole can be rewritten…

illwieckz:
I was bored by the skybox code so I revamped it entirely. The skybox code was very ugly with 10 gotos and buggy behavior, for example it would not attempt to load any multifile skybox if a single-file skybox was successfully loaded but had invalid content.

Also, the code was inefficient, for example it was doing the image rotation operation for every face even if the rotation has an angle of 0°, hence allocating a new image, copying all the pixels in the new image et. for nothing.

The skybox faces are now resized to the size of a square that has for width and height the greater width and height found in any image. This way it would not only fix the quake3 skyboxes with smaller bottom faces, but also non-square skyboxes.

src/engine/renderer/tr_image.cpp
```diff
@@ -2038,122 +2099,109 @@ image_t        *R_FindCubeImage( const char *imageName, int bits, filterType_t f
 		pic[ i ] = nullptr;
 	}
 
-	for ( i = 0; i < 6; i++ )
+	for ( f = 0; f < numMultifileCubeMapFormats; f++ )
```
slipher: Optional nit: Instead of having an index it's nicer to use a for-each loop like `for (const multifileCubeMapFormat_t format : multifileCubeMapFormats)`

src/engine/renderer/tr_image.cpp
```diff
@@ -2038,122 +2099,109 @@ image_t        *R_FindCubeImage( const char *imageName, int bits, filterType_t f
...
 	{
-		Com_sprintf( filename, sizeof( filename ), "%s_%s", buffer, openglSuffices[ i ] );
+		int greaterEdge = 0;
```
slipher: should be "greatest"

src/engine/renderer/tr_image.cpp
```diff
@@ -1937,23 +1937,53 @@ static void R_Rotate( byte *in, int width, int height, int degrees )
 	ri.Hunk_FreeTempMemory( tmp );
 }
 
+byte *R_Resize( byte *in, int width, int height, int width2, int height2 )
```
slipher: This could use a comment or a different name to indicate that it's specifically for upsampling, not any resize.

src/engine/renderer/tr_image.cpp
illwieckz: right

src/engine/renderer/tr_image.cpp
illwieckz: do you think that algorithm does not work for downscaling?
we only use it to upscale, but isn't it generic enough to downscale too?

src/engine/renderer/tr_image.cpp
illwieckz: oh, interesting!

src/engine/renderer/tr_image.cpp
slipher: Well maybe, if not with the best quality. But there's already a downsampling function `ResampleTexture` in the same file.

src/engine/renderer/tr_image.cpp
```diff
@@ -1944,6 +1944,9 @@ byte *R_Resize( byte *in, int width, int height, int width2, int height2 )
 
 	out = (byte*) ri.Z_Malloc( width2 * height2 * 4 );
 
+	ResampleTexture( (unsigned int*) in, width, height, (unsigned int*) out, width2, height2, false );
```
illwieckz: @slipher I tried to reuse `ResampleTexture` since I see no reason it would not work for upscaling too.
It works but I had to hack some casts, basically the cubemap faces are bytes (`unsigned char`) while `ResampleTexture` expects `unsigned int`).

It looks like that `ResampleTexture` function was never used before, so perhaps we can fix it…

This is with `R_Resize`:

 [skybox](https://dl.illwieckz.net/b/daemon/bugs/noncube-skybox/unvanquished_2019-03-06_200229_000.jpg)

This is with `ResampleTexture`:

[skybox](https://dl.illwieckz.net/b/daemon/bugs/noncube-skybox/unvanquished_2019-03-09_202118_000.jpg)

It looks like we may drop one of them.

illwieckz: There is one place where `ResampleTexture` was used, but it used the cast hack too, so we can rewrite the `ResampleTexture` definition…

illwieckz: @slipher I have not managed to get valid results by redefining `ResampleTexture`, so I had to kept the ugly casts.

In any way I dropped the custom resizing code to rely on `ResampleTexture`. I kept `R_Resize` as a wrapper to make in-place resizing easier.

slipher: OK with me. The difference is that ResampleTexture takes an evenly weighted average of 4 pixels. Maybe the algorithm doesn't make that much sense for scaling up, but of course it doesn't matter since it's only used as a workaround for assets which by rights shouldn't work at all.

illwieckz: Thanks for the review!

PR #180, opened 2019-03-10
refactor and fix normalmap, parallax, reflectionmap, glowmap, GL/DX normal switch

illwieckz:
This PR first refactors some normalmap code to avoid it to be copypasted in 6 files, and to be sure the same features are supported anywhere, and that was not true since r_NormalScale was not used anytime.

Then, the PR adds a keyword shader to allow people to revert a normalmap that is stored the wrong way, without changing the file itself.

Then the PR auto enables the DirectX format for DarkPlaces detected normalmaps.

So, this code not only fixed bugs and is not only more readable, it also enables engine to load a xonotic map after an unvanquished map in a totally transparent manner.

The PR also refactors parallax code and fixes the orientation.

Reflection maps are now loaded as cubemap (that was a bug), a new kind of cubemap naming scheme is added (for compatibility with Darkplaces).

The `parallax` keyword can now receive the same arguments `dpoffsetmapping` expects.

Parallax reflection cubemap is now implemented.


illwieckz:
I also refactored parallax code and fixed orientation the way it seems right to do (transforming `[1-0]` into `[0-1]` at the pixel level).

illwieckz:
I added a commit that implements `dpoffsetmapping`. This is a very useful keyword allowing to scale and offset an heightmap on a per shader basis.

See the texture in foreground, the edge is not correct because the heightmap provides height from the bottom instead of depth from the top, hence the whole texture is displayed lower than the surface itself:

[dpoffsetmapping](https://dl.illwieckz.net/b/daemon/wip/parallax/dpoffsetmapping/unvanquished_2019-03-13_192558_000.jpg)

The same texture once `dpoffsetmapping` is implemented:

[dpoffsetmapping](https://dl.illwieckz.net/b/daemon/wip/parallax/dpoffsetmapping/unvanquished_2019-03-13_193501_000.jpg)

Some more screenshots, light mapping on edges:

[dpoffsetmapping](https://dl.illwieckz.net/b/daemon/wip/parallax/dpoffsetmapping/unvanquished_2019-03-13_194237_000.jpg)

realtime light mapping on edges:

[dpoffsetmapping](https://dl.illwieckz.net/b/daemon/wip/parallax/dpoffsetmapping/unvanquished_2019-03-13_193654_000.jpg)

OK from the dretch point of view floor with `0.02` depth scale.
I recommend to switch from `0.03` to `0.02` since we never used `0.03` yet and `0.02` produces less artifacts when seen from dretch point of view and it is Darkplaces default, hence `0.02` is a better choice for both Unvanquished and Xonotic:

[dpoffsetmapping](https://dl.illwieckz.net/b/daemon/wip/parallax/dpoffsetmapping/unvanquished_2019-03-13_193718_000.jpg)

Note that the bricks on the wall are flats because those textures use loose heightmap (and darkplaces computes normalmap from heightmap), something we don't do yet, otherwise everything else is right, notice there is no dark blotches in grooves anymore:

[dpoffsetmapping](https://dl.illwieckz.net/b/daemon/wip/parallax/dpoffsetmapping/unvanquished_2019-03-13_193850_000.jpg)

Some more coolness:

[dpoffsetmapping](https://dl.illwieckz.net/b/daemon/wip/parallax/dpoffsetmapping/unvanquished_2019-03-13_194002_000.jpg)

Note that I'm adding engine code to GPLv2 files or sometime rewriting code but I prefer to contribute it under BSD , perhaps I will add some comments to identify what is more permissive. The normal map code is brand new, the dpoffsetmapping code does not copypaste DarkPlaces (and the way our engine is designed is not very compatible with the way it is implemented in Darkplaces).

illwieckz:
I discovered that our original code was not applying parallax offset to glowmaps, this is now fixed.

Before:  
[dpoffsetmapping](https://dl.illwieckz.net/b/daemon/wip/parallax/dpoffsetmapping/unvanquished_2019-03-13_205149_000.jpg)

After:  
[dpoffsetmapping](https://dl.illwieckz.net/b/daemon/wip/parallax/dpoffsetmapping/unvanquished_2019-03-13_212527_000.jpg)

Note tat the displayed texture has a buggy parallax on Dæmon but is also buggy in Darkplaces (the artifacts probably just differs because of implementation details).

DolceTriade:
Very cool! @gimhael if you get a chance, could you review?

illwieckz:
So, more awesome stuff:

Reflection maps are now loaded as cubemap (that was a bug), a new kind of cubemap naming scheme is added (for compatibility with Darkplaces).

The `parallax` keyword can now receive the same arguments `dpoffsetmapping` expects.

Parallax reflection cubemap is now implemented.

Shader stage collapse was revamped with a lot more simpler code, and less buggy one.

It allowed to merge the implicit lightmap that comes with diffusemap and the legacy explicit ones, to avoid double shadowing and shadows being painted over glow maps.

illwieckz:
OK, people must know about the fact that ALL Unvanquished maps are broken with this branch because I fixed a bug that hidden another.

Basically, all our normal maps are compressed with `crunch -dxtn` and our engine loads them with reverted blue channel (Z component). This was hidden by the bug in the normal code that was loading the Z component from alpha channel. Hopefully we did not shipped any heightmap in alpha channel and since `crunch -dxtn` drops the alpha channel in any way, it looks like the engine was just using an empty 2 component). The normalmap load correctly if loaded from another format (like webp).

The current tree is meant to be _correct_. I created a test map with assets I created to be sure they are the way they must be, then fixed the engine to render the test map. The _correct_ engine in this tree is able to render correctly xonotic maps and unvanquished maps if their normal maps were not compressed with `crunch -dxtn`. I'll write an issue about this soon, but we need someone to fix the normal map loading when they are compressed with `dxtn` (use this branch as a base as master is not _correct_).

slipher:
It's too hard to review 27 commits. Can you split this up into multiple PRs?

illwieckz:
I reduced the number of commits to 10 by:

- reordering and squashing commits
- merging one that is very trivial: 178bb54
- moving stage collapse commits to another PR: #184
- retaining some bikeshedding on a local branch for later

note that the reflection map parallax does not work until #184 is merged

The dpreflectcube and later commit may be another PR but they require those from this branch to compile, so I left them.

src/engine/renderer/tr_image.cpp
```diff
@@ -2029,20 +2030,31 @@ static const multifileCubeMapFormat_t multifileCubeMapFormats[] =
...
 	{
 		"Quake",
+		"_",
 		{ "rt", "lf", "bk", "ft", "up", "dn" },
 		{ true, true, false, true, true, false },
 		{ false, false, true, false, false, true },
 		{ 90, -90, 0, 0, 90, -90 },
 	},
+	{
+		"DarkPlaces",
```
illwieckz: I had no idea to name this, Xonotic use some OpenGL-like skyboxes without `_` separator. I may name it “OpenGL” too.

illwieckz:
note that second commit, the one named “_glsl/normal: fix code to compute normal in tangent space with simpler code_” is the one that makes dæmon unable to lit our maps due to the normalmap not being properly loaded from dxtn crunch files. But this commit is needed as it fixes rendering bugs and is a required foundation for many other fixes.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -199,6 +199,27 @@ void computeDLights( vec3 P, vec3 N, vec3 I, vec4 diffuse, vec4 specular,
...
+vec3 ComputeNormalInTangentSpace(sampler2D normalMap, vec2 texNormal)
+{
+	vec3 N = texture2D(u_NormalMap, texNormal.st).xyw;
+	N.x *= N.z;
+	N.xy = 2.0 * N.xy - 1.0;
+	N.z = sqrt(1.0 - dot(N.xy, N.xy));
+
+#if defined(r_NormalScale)
+	N.z *= r_NormalScale;
+	normalize(N);
```
slipher: does this line have any effect?

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -201,10 +201,8 @@ void computeDLights( vec3 P, vec3 N, vec3 I, vec4 diffuse, vec4 specular,
 
 vec3 ComputeNormalInTangentSpace(sampler2D normalMap, vec2 texNormal)
 {
-	vec3 N = texture2D(u_NormalMap, texNormal.st).xyw;
-	N.x *= N.z;
-	N.xy = 2.0 * N.xy - 1.0;
-	N.z = sqrt(1.0 - dot(N.xy, N.xy));
+	vec3 N = texture2D(normalMap, texNormal.st).rgb;
```
slipher: I guess this commit should be removed now that Fuma has explained why things are extra magical?

slipher:
If some of the code makes everything totally broken it should be pushed to a branch other than `master`.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: the `normalize` one? I really don't know

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: more code is required to get it working properly in every case, this line:
```
vec3 N = texture2D(u_NormalMap, texNormal.st).xyw; 
```

means that normalmap is sampled from rga for each case, even non DXn normal maps with heightmap in alpha channel. Because of heightmap in alpha channel use case we may use a “_if DXn_” switch (the code only claims to not require such test when the only faced use cases are RGB files or DXn ones abusing Alpha.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
slipher: I don't think it does. If you don't know, where are you getting the code from?

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
slipher: So why is it a good idea to remove support of all the formats other than the "uncompressed RGB" one? Did you find that no assets use any of them?

illwieckz: It's not a good idea at all. All Unvanquished maps use the other format.

I fixed it.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: copy pasted from our code, this is a refactor, some of the normal code in other files had this.

illwieckz:
I restored the Capcom code to abuse DXT1/5 alpha channel to compress normal maps and I passed the `heightMapInNormalMap` boolean to glsl so the engine does not use the trick when it loads a normal map file with an height map in alpha channel.

This `heightMapInNormalMap` will be used in the future by parallax code to know from where computing the parallax when height map from loose file support will be added. Loading of height map from loose is required since `DXn`  can't store the height map in alpha channel because of the alpha channel abuse.

illwieckz:
I also fixed the compilation of the `liquid` related code so it builds if enabled. By the way I left it disabled and produces visual garbage in any way.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
slipher: Let's just delete this `r_NormalScale` stuff. The code does not make sense: (1) Why would you want to multiply just one component of the vector by some value? (2) According to any docs/examples I find, `normalize` is supposed to be used by the return value, not an in-place modification.

Nothing will be lost as it would only have an effect if you are playing with cvars.

src/engine/renderer/tr_shader.cpp
```diff
@@ -3571,7 +3578,51 @@ static bool ParseShader( const char *_text )
...
+				}
+				else if ( !Q_stricmp( token, "1" ) )
+				{
+					// do nothing, that's the default
+				}
+				else
+				{
+					Log::Warn("unknown normalFormat keyword '%s' not supported in '%s'", token, shader.name );
+					SkipRestOfLine( text );
+					continue;
```
slipher: I think you want `break` here.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -199,13 +201,37 @@ void computeDLights( vec3 P, vec3 N, vec3 I, vec4 diffuse, vec4 specular,
 #endif
 }
 
+vec3 NormalFlip(vec3 N)
```
slipher: Nice hack :)

src/engine/renderer/tr_shader.cpp
```diff
@@ -3571,7 +3578,51 @@ static bool ParseShader( const char *_text )
...
+		//  1  1  1  OpenGL format (default)
+		//  1 -1  1  DirectX format with reverted green (Y) channel
+		// -1  1  1  weird other format with reverted red (X) channel
+		else if ( !Q_stricmp( token, "normalFormat" ) )
+		{
+			for ( int i = 0; i < 3; i++ )
+			{
+				token = COM_ParseExt2( text, false );
+
+				if ( token[ 0 ] == 0 )
```
slipher: Since `token[0]` is a char it would be nicer to do `!token[0]` or `token[0] == '\0'` instead of comparing to an int.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -268,10 +279,10 @@ float RayIntersectDisplaceMap(vec2 dp, vec2 ds, sampler2D normalMap)
 	{
 		depth += size;
 
-		vec4 t = texture2D(normalMap, dp + ds * depth);
+		float t = 1.0 - texture2D(normalMap, texCoords + ds * depth).a;
```
slipher: Why is the 'one minus' introduced?

src/engine/renderer/tr_shader.cpp
illwieckz: you're right, and I'm especially the kind of guy that likes explicit correct comparisons, I'll do that :-)

illwieckz: In fact I copy pasted this line, there is 51 of them in that file. I fixed it (will push) but I may fix the 51 others in a bikeshedding commit in the future.

src/engine/renderer/tr_shader.cpp
illwieckz: `break` would stop right there the parsing of the material, hence disabling it entirely. This would not be a warning but an error, and such syntax issue is not problematic enough to entirely invalidate the material (at least diffuse, glow and specular would be properly rendered and it may be rendered with a flipped reflection on y, with a warning).

There was a similar talk in #80 (about `return` and shader keyword not parm, but that's almost the same).

`SkipRestOfLine` is there to be sure next material line is read.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: That's because parallax digs hole into the texture from the top and top is 1.0:

```
 1 ____      1-0 ____
  |    \        /
  |     \      /
.5|      \____/ 1-.5
  |
  |
 0|
```

The height the must be the floor, basically the heightmap parallax is a depth map.

A bias may then be introduced between `1.0` and `texture2D(normalMap, texCoords + ds * depth).a`

Notice the `-` in the `-offsetScale` that is just a `* -1.0` which makes the depth a negative number that may be substracted to `1.0`.

illwieckz: The miss of that “_one minus_” was the cause of the reverted parallax (and maybe wrong offset).

That is the proper fix.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: The `z` component is the vector that goes toward the viewer, that may increase the perceived depth and contrast the shadows. That may be useful as a per-shader keyword, and it is a per-shader keyword. The mistake there is that i's named `r_normalScale` and not `u_normalScale`.

I noticed that numbers under 1 darken the shadows, numbers over 1 soften. I also noticed the `normalize` line seems to not affect the rendering and may be removed. If you want to try it with `r_normalScale`, you need `vid_restart` to see the changes.

With `r_normalScale .1`: 

[normalscale](https://dl.illwieckz.net/b/daemon/features/normalscale/unvanquished_2019-03-18_193327_000.jpg)

[normalscale](https://dl.illwieckz.net/b/daemon/features/normalscale/unvanquished_2019-03-18_193453_000.jpg)

[normalscale](https://dl.illwieckz.net/b/daemon/features/normalscale/unvanquished_2019-03-18_193525_000.jpg)

I removed the useless `normalize` call.

I was already thinking about revamping that part to be sure the shader keyword value is taken in account _and_ in any case, but I'm not sure to do it within that PR. It's planed in any way.

src/engine/renderer/tr_shader.cpp
slipher: No, `continue` or `break` applies to the innermost loop. In this case that's the `for ( int i = 0; i < 3; i++ )` for the normal directions, which are all supposed to be on one line, right? So if you have skipped the rest of the line then you should break out of the loop.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
slipher: It was confusing since the commit was labeled "refactoring" and all other parts moved stuff around without behavioral changes.

src/engine/renderer/tr_shader.cpp
illwieckz: Oh, you're right, I'm dumb. Fixed.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: Oh right, that slipped in with a fixup, I extracted it to its own commit.

illwieckz:
Just to say that I have some other changes in local branches that become really hard to rebase each one I edit this PR and/or #184, so I hope this is getting merged soon. :grin:

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -199,59 +202,133 @@ void computeDLights( vec3 P, vec3 N, vec3 I, vec4 diffuse, vec4 specular,
 #endif
 }
 
+vec3 normalFlip(vec3 normal)
```
gimhael: IMHO this function should not use if()s, just compute normal * u_NormalFormat and make sure that the u_NormalFormat components are either 1.0 or -1.0 when the uniform is written. Even better would be to rename it to vec3 u_NormalScale and premultiply r_NormalScale into this uniform and delete r_NormalScale in the shaders. Then changing the r_NormalScale cVar wouldn't require a shader recompile.

illwieckz: I plan to revamp the `u_NormalScale` thing in a future PR because yes that feature is currently broken. There is both `r_NormalScale` (which is global and may be useful for testing purposes) and `normalScale` which is per-shader but lost in translation. My idea is to remove `r_normalScale` define in glsl and just pass `u_NormalScale` which would be a `r_normalScale × shader.normalScale`.

I'm not very happy with the idea of multiplying `u_normalScale` with `u_normalFormat`, I was thinking about a define when normalFormat's `x × y × z < 0` so the code is even not compiled in glsl at all if there is no flip to do.

In any way, is this blocking? I have a lot of work to push in other PRs that relies on this code, and code that _may even rewrite some part of this_. This is just iterating process, I need something getting done before going the next step, and at one point it's hard to put everything doable (or to reach perfectibility) in one PR. This is a stable state on which I can base other work. Can I really postpone the the `u_normalScale` refactor? for sure I will do it, but there is a lot of clean-up in `tr_shade.cpp` to do first and that will be my next PR.

@gimhael I have a special question for you. I've read it may be better to do operations for nothing than branching in glsl (like multiplying by `1.0` instead of doing an `if`) because of the way GPU works. But what is the performance hit of #ifdef in glsl? People with a C heritage would think nothing, but that's probably wrong as you just said changing `r_normalScale` requires a shader recompile, and I noticed that behaviour.

Basically there is a lot of things that is done every pixel every frame for nothing. We have options to disable normalMapping for example, but it just has no influence on performance since that option is just mapping a flat image and do the computation on it instead (which is wrong but that's out of topic). I have not-yet pushed branches with added `r_deluxeMapping`, `r_specularMapping` and `ifdef` that just disable the related code in glsl not only on global variable state, but on per-shader state. Is it a problem regarding to GPU architecture if a given texture is using `lightMapping_fp.glsl` with all features enabled (normal, parallax, specular, glow), and another texture is using `lightMapping_fp.glsl` with all features disabled but diffuse and lightmap? There is basically way to have many versions of the same shader in memory, one with disabled lightmap and one with disabled specularmap, is that an issue as long as those shaders don't do themselves `if` branching in their code?

gimhael: Passing just one uniform variable that contains the product of all scale factors is definitely the right way. It's just a matter of scale, the multiplication in the pixel shader runs millions of times per frame while the premultiplication runs only once per shader binding. But if you want to use such a scale factor anyway, you can use it for normal flipping too.

#ifdef works like in C and C++, GLSL has a precompiler that removes all conditional code before it is passed to the real compiler, so the removed code has no runtime cost at all. The problem with #ifdef or similar is that you end up with a much larger number of shaders, which means a) long compile times at startup and b) that you have to switch shaders more often within a frame.

The cost of switching shaders is higher than the cost of switching texture bindings and much higher than the cost of switching uniforms (see e.g. page 48 of http://media.steampowered.com/apps/steamdevdays/slides/beyondporting.pdf), so most engines try to minimize the number of shader changes by using very generic "supershaders". In more modern APIs like DX12 and Vulkan the cost of shader switches is much lower, but for GLSL it is usually not advisable to double the number of shader variants just to save a few multiplications in one of them.

The cost of texture fetches in a shader is more significant, so removing some unneeded texture fetches may be profitable, however the flat images we bind are very small, so that they should be completely in the texture cache and the additional cost is hopefully also low.

Regarding regular if statements, you have to know that GPUs are not really good at branching, in fact the ability to jump over unused instructions was only introduced in DX10-era GPUs. The problem is that GPUs are based on SIMD, when a GPU manufacturer claims that his GPU supports e.g. 2048 threads in parallel, then means likely that the GPU has say 32 compute units with 64xSIMD each.

A compute unit is like a CPU core, it can run code independently from all the other compute units, and it runs the vertex or pixel shader for up to 64 vertices or pixels in parallel by using the 64 components of the SIMD registers it has. Ergo there is only one program counter for all 64 pixels and if there is a conditional jump instruction, the compute unit can only jump over the code, if all pixels have the same condition.

For non-uniform jumps the compute unit will still step through the code, but the results of the computation is discarded at the end. And pre-DX10 GPUs regard all jumps as non-uniform, so it usually is not a big gain to introduce ifs for performance reasons.

I don't think that this particular issue should block the PR, but I'm pretty sure that performance-wise the ifs in this function don't help.

illwieckz: Thank you for all those explanations, are you ok to put those words under CC By SA 3.0 so we may one day copy paste them in the wiki? knowledge is precious.

Regarding ifdef, I noticed that our code writing them looks to be broken. I tried to replace the `if (u_HeightMapInNormalMap)` code by some `ifdef USE_HEIGHTMAP_IN_NORMALMAP` and the normalmaps are now flipping on/off while I move. The macro code is really asking for a revamp in any way but it's really not the time to do that revamp. Especially because I have a lot of commits in the pipe that rely on the current code. Also, some of those commits are cleaning-up things, and it's always easier to revamp after a clean-up/refactor.

So I let the current code this way since it works with code we know to not be buggy (uniforms). We will be able to switch to more efficient code (macro) when the buggy code will not be buggy anymore.

PR #182, opened 2019-03-15
Archive-level symlink support

slipher:


slipher:
Also I'm thinking it was silly to support arbitrary-depth symlinks. Probably better to make the cvar a bool and not support depths greater than 1 (unless @illwieckz found any real examples with depth more than 1).

illwieckz:
Xonotic `20170401` has 439 first-level (`../`) symlink and, second-level (`../../`) symlinks and no third-level (`../../../`) symlinks.

illwieckz:
Also, some are absolute symlink (but they don't have leading `/` as expected in q3 vfs), for example:

`dds/wakazachi_reflect.dds` -> `textures/wakazachi_reflect.dds`

Note that his one is in fact a symlink to `dds/textures/wakazachi_reflect.dds`, the implicit `dds/` prefix being already handled by dæmon (see #81) at image loading level, it may work if that's not a recursive symlink and the path is returned to engine, file being seen by zip code or not.

I don't have any idea yet to easily check for recursive symlink and for cross pak symlinks.

illwieckz:
So, there is no absolute symlink at all, all the symlink that have a `/` in their target name without having any leading `../` prefix are relative to `dds/`.

illwieckz:
OK, given each pk3 is extracted as a pk3dir, this code list all cross-pk3 symlinks:

```sh
for p in *.pk3dir
do
    (
        cd "${p}"
        find . -type l \
        | while read l
        do
            t="$(readlink "${l}")"
            d="$(dirname "${l}")"
            (
                cd "${d}"
                [ -f "${t}" ] || echo "${t}"
            )
        done
    )
done
```

hint: there is none :lollipop:

illwieckz:
@slipher, that looks good to me, why not printing a message (at least a debug one) on each resolved symlink (with original path and final one, it's not a problem if the two names are on two debug messages lines if implementation does not allow better, the fs code is sequential after all)?

illwieckz:
I confirm this work.

Thanks to in-pak symlink compatibility layer I was able to simplify and make configurable normal map code, fix parallax one, and fix reflection map leveraging xonotic assets, that's very precious!

illwieckz:
WAIT, the file system error code just make the loading a map fail entirely if a symlink fails to resolve.

src/common/FileSystem.cpp
```diff
@@ -905,6 +910,74 @@ class ZipArchive {
...
+				fsLogs.Warn("Symlink resolved to invalid filename '%s'", name);
+				SetErrorCodeFilesystem(err, filesystem_error::invalid_filename);
+				return 0;
+			}
+			auto maybeOffset = FindOffsetForName(name, err);
+			if (err)
+				return 0;
+			if (!maybeOffset) {
+				fsLogs.Warn("Symlink points to nonexistent file '%s'", name);
+				SetErrorCodeFilesystem(err, filesystem_error::no_such_file);
```
illwieckz: this entirely aborts map loading, that must not

src/common/FileSystem.cpp
```diff
@@ -905,6 +910,74 @@ class ZipArchive {
...
+				return 0;
+			}
+			offset = *maybeOffset;
```
illwieckz: This would please me a lot:

```diff
diff --git a/src/common/FileSystem.cpp b/src/common/FileSystem.cpp
index f88c31ee5..85fd981a6 100644
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -960,20 +960,20 @@ public:
                        resolvedName = ResolveLinkPath(name, link, err);
                        if (err)
                                return 0;
-                       name = resolvedName;
-                       if (!Path::IsValid(name, /*allowDir=*/false)) {
-                               fsLogs.Warn("Symlink resolved to invalid filename '%s'", name);
+                       if (!Path::IsValid(resolvedName, /*allowDir=*/false)) {
+                               fsLogs.Warn("Symlink resolved to invalid filename '%s' → '%s'", name, resolvedName);
                                SetErrorCodeFilesystem(err, filesystem_error::invalid_filename);
                                return 0;
                        }
-                       auto maybeOffset = FindOffsetForName(name, err);
+                       auto maybeOffset = FindOffsetForName(resolvedName, err);
                        if (err)
                                return 0;
                        if (!maybeOffset) {
-                               fsLogs.Warn("Symlink points to nonexistent file '%s'", name);
+                               fsLogs.Warn("Symlink points to nonexistent file '%s' → '%s'", name, resolvedName);
                                SetErrorCodeFilesystem(err, filesystem_error::no_such_file);
                                return 0;
                        }
+                       fsLogs.Debug("Symlink resolved: '%s' → '%s'", name, resolvedName);
                        offset = *maybeOffset;
                }
        }
```

src/common/FileSystem.cpp
```diff
@@ -124,6 +124,7 @@ namespace FS {
 
 #ifdef BUILD_ENGINE
 static Cvar::Cvar<bool> fs_legacypaks("fs_legacypaks", "Also load pk3s, ignoring version.", Cvar::NONE, false);
+static Cvar::Cvar<int> fs_maxSymlinkDepth("fs_maxSymlinkDepth", "Max depth of symlinks in zip paks. 0 means disabled.", Cvar::NONE, 0);
```
illwieckz: The trailing dot in `fs_legacypaks` description string was a mistake of mine, the common usage is to not add trailing dot to cvar descriptions. Can you fix `fs_maxSymlinkDepth` description by removing the dot?

You may add an extra commit to remove the trailing dot in `fs_legacypaks` description and even rename to `fs_legacyPaks` while you're at it (another mistake of mine).

slipher:
I addressed the comments. Also by a depth 2 symlink I mean a symlink that points to a symlink that finally points to an honest file. Are there any depth 2 symlinks like that? (Otherwise arbitrary depth support can be removed.)

slipher:
OTOH it's not really more complex to implement arbitrary depth, so it's probably not worth checking.

illwieckz:
I used this code to look for recursive symlinks.

```sh
resolve() {
	t="$(readlink "${1}")"
	d="$(dirname "${1}")"
	(
		cd "${d}"
		if [ -h "${t}" ]
		then
			printf ' → %s' "${t}"
			resolve "${t}"
		fi
	)
}

for p in *.pk3dir
do
	(
		cd "${p}"
		find . -type l \
		| while read l
		do
			t="$(readlink "${l}")"
			d="$(dirname "${l}")"
			(
				cd "${d}"
				if [ -h "${t}" ]
				then
					printf "%s → %s" "${l}" "${t}"
					resolve "${t}"
					printf '\n'
				fi
			)
		done
	)
done
```

There is 38 depth-2 symlinks and no one depth-3 symlinks. All depth-2 symlinks are for player model files, not maps.

illwieckz:
Great, I tested broken link on shader file, it's correctly handled:

```
Debug: ...loading 'scripts/map_erbium.shader' 
Warn: [FS] Symlink points to nonexistent file 'scripts/map_erbium.shader' → 'scripts/toto.shader' 
Debug: Failed to open 'scripts/map_erbium.shader' for reading: No such file 
Warn: Couldn't load shader file scripts/map_erbium.shader 
```

I also tested some legit (if it can be legit) symlink loading, it works:

```
Debug: looking for extra maps for color map: 'textures/phillipk2x/panels/pk02_panels01b' 
Debug: found extra normal map 'textures/phillipk2x/panels/pk02_panels01b_norm' 
Debug: [FS] Symlink resolved: 'dds/textures/phillipk2x/panels/pk02_panels01b_norm.dds' → 'dds/textures/map_oilrig/pk02_panels01b_dark_norm.dds' 
Debug: found extra specular map 'textures/phillipk2x/panels/pk02_panels01b_gloss' 
Debug: [FS] Symlink resolved: 'dds/textures/phillipk2x/panels/pk02_panels01b_gloss.dds' → 'dds/textures/map_oilrig/pk02_panels01b_dark_gloss.dds' 
```

It's OK to merge.

illwieckz:
@slipher I noticed that you like to rebase and merge the fast forward way, so I let you merge this one.

PR #187, opened 2019-03-24
gl_shader: clean-up some useless macro boilerplate

illwieckz:
clean-up some useless macro boilerplate, because it's easier to debug code that has less code


illwieckz:
basically deletes 209 lines by rewriting 122 lines in an obvious way

illwieckz:
well, this is pretty straightforward and I need it for two PR

PR #188, opened 2019-03-24
implement r_deluxeMapping, r_specularMapping, r_glowMapping etc.

illwieckz:
Some code that implements `r_deluxeMapping`, `r_specularMapping` and `r_glowMapping`.

It's made the way we can disable normal mapping while keeping deluxe mapping.
Of course disabling deluxe mapping disable normal mapping too.

It's made the way we can disable normal mapping while keeping parallax mapping.

Some glsl code is not compiled when some features are disabled. More disablements will be done in future PRs, but this already makes a GMA965 going from `10` fps to `15` fps while rendering the parpax default scene (spectator intermission) which is already 50% more (but this is unnoticeable on R9 390X).

Some specular code is refactored the way it's not copy-pasted across multiple files anymore.


slipher:
If I understand correctly, each one of {r_deluxeMapping, r_specularMapping, r_glowMapping} is implemented as a switch that can be individually turned on or off for each texture? Maybe it would make more sense to make this a global variable (requiring a renderer restart to change) so that it does require 2x as many variants of each GLSL shader to be compiled.

illwieckz:
right

illwieckz:
Note that we can't do `#ifdef tr_deluxeMapping` since that macro would be set on vid restart while the data to set it (`tr.worldDeluxeMapping`) is set at bsp load. This setting is content dependent (the lightmap files) and this kind of macros are not redefined between map loads.

illwieckz:
Uh no! those defines look to be set once for all at startup. The `GenEngineConstants` function is called on each `vid_restart` call but the same shader is used across each `vid_restart`. A complete engine restart is required to see the shader being “optimized”. o.O

illwieckz:
`r_normalMapping`, `r_specularMapping` and `r_glowMapping` are now set for all the shaders, but I'm not sure code is properly disabled because I have hard time to check those macros are taken in account or not and when. I haven't compared for perf on my old gma965 yet.

Note that more code disablement is to be done.

Because of those macros being for all the materials and not per-material, I had to light & specular stuff from `reliefMapping_fp.glsl` into a new lib named `computeLight_fp.glsl`. This is something I wanted to do in anyway since it was stupid to pass all the lighting stuff to the heat haze shader that just requires the normalmap code.

src/engine/renderer/tr_shade.cpp
```diff
@@ -1977,14 +1989,22 @@ static void Render_forwardLighting_DBS_directional( shaderStage_t *diffuseStage,
 	// bind u_NormalFormat
 	gl_forwardLightingShader_directionalSun->SetUniform_NormalFormat( tess.surfaceShader->normalFormat );
 
-	// bind u_SpecularMap
-	if ( r_forceSpecular->integer )
-	{
-		GL_BindToTMU( 2, diffuseStage->bundle[ TB_DIFFUSEMAP ].image[ 0 ] );
-	}
-	else if ( diffuseStage->bundle[ TB_SPECULARMAP ].image[ 0 ] )
+	if ( specularMapping )
```
slipher: This changes the effect of `r_forceSpecular`. Before, it would do something even without a specular map image, but now it doesn't.

src/engine/renderer/tr_shade.cpp
```diff
@@ -2297,6 +2312,8 @@ static void Render_liquid( int stage )
 
 	GLimp_LogComment( "--- Render_liquid ---\n" );
 
+	bool specularMapping = r_specularMapping->integer; // TODO: specular map?
```
slipher: You did it, right?

src/engine/renderer/tr_init.cpp
```diff
@@ -162,9 +162,12 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_specularExponentMin;
 	cvar_t      *r_specularExponentMax;
 	cvar_t      *r_specularScale;
+	cvar_t      *r_specularMapping;
 	cvar_t      *r_deluxeMapping;
 	cvar_t      *r_normalScale;
 	cvar_t      *r_normalMapping;
+	cvar_t      *r_parallaxDepthScale;
```
slipher: These seem to be in the wrong commit but that's OK, it's not going to hurt anyone.

src/engine/renderer/tr_shade.cpp
```diff
@@ -1589,14 +1593,22 @@ static void Render_forwardLighting_DBS_omni( shaderStage_t *diffuseStage,
...
-	if ( r_forceSpecular->integer )
-	{
-		GL_BindToTMU( 2, diffuseStage->bundle[ TB_DIFFUSEMAP ].image[ 0 ] );
-	}
-	else if ( diffuseStage->bundle[ TB_SPECULARMAP ].image[ 0 ] )
+	if ( specularMapping )
 	{
-		GL_BindToTMU( 2, diffuseStage->bundle[ TB_SPECULARMAP ].image[ 0 ] );
+		// bind u_SpecularMap
+		if ( r_forceSpecular->integer )
```
slipher: same comment about forceSpecular

src/engine/renderer/tr_shade.cpp
```diff
@@ -1782,14 +1790,22 @@ static void Render_forwardLighting_DBS_proj( shaderStage_t *diffuseStage,
...
```
slipher: same comment about forceSpecular

src/engine/renderer/tr_init.cpp
```diff
@@ -1195,10 +1195,12 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_specularExponentMin = ri.Cvar_Get( "r_specularExponentMin", "0", CVAR_CHEAT );
 		r_specularExponentMax = ri.Cvar_Get( "r_specularExponentMax", "16", CVAR_CHEAT );
 		r_specularScale = ri.Cvar_Get( "r_specularScale", "1.0", CVAR_CHEAT | CVAR_SHADER );
+		r_specularMapping = ri.Cvar_Get( "r_specularMapping", "1", CVAR_ARCHIVE );
```
slipher: Needs `CVAR_LATCH`.

src/engine/renderer/tr_init.cpp
```diff
@@ -1201,6 +1202,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_normalMapping = ri.Cvar_Get( "r_normalMapping", "1", CVAR_ARCHIVE );
 		r_parallaxDepthScale = ri.Cvar_Get( "r_parallaxDepthScale", "0.03", CVAR_CHEAT );
 		r_parallaxMapping = ri.Cvar_Get( "r_parallaxMapping", "0", 0 );
+		r_glowMapping = ri.Cvar_Get( "r_glowMapping", "1", CVAR_ARCHIVE );
```
slipher: needs `CVAR_LATCH`

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -79,13 +88,16 @@ vec3 NormalInTangentSpace(sampler2D normalMap, vec2 texNormal)
 #endif
 
 	return normal;
+#else // !r_normalMapping
+	return vec3(1.0, 1.0, 1.0);
```
slipher: I believe it should be (0, 0, 1), based on what you said the flat texture was. Presumably this means pointing directly outward. (1, 1, 1) is definitely not right as it is not a normalized vector (i.e., length 1).

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -21,6 +21,19 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 // computeLight_fp.glsl - Light computing helper functions
 
+#if defined(r_specularMapping) || defined(USE_PHYSICAL_SHADING)
+uniform vec2            u_SpecularExponent;
+#define getSpecular() texture2D(u_SpecularMap, var_TexSpecular + texOffset)
```
slipher: It would be better to take the coordinate as an argument instead of hard-coding the name of some local variable.

src/engine/renderer/tr_init.cpp
DolceTriade: OW! Stubbed my toe on this.

illwieckz: this just moves stuff along similar things for better reading

illwieckz: That probably slipped up with some fixup, that rebase switching to `USE_` to `r_` macros is the most crazy one I ever did, with a file split in the middle.

src/engine/renderer/tr_shade.cpp
illwieckz: hmmm, what is “_`r_forceSpecular` without specular map image_” for?

src/engine/renderer/tr_shade.cpp
illwieckz: yes I taken care of this question, I will remove that comment
note that liquid shader does not use specular map from file, the specular one is somewhat computed. That prevents factorisation a lot, especially we have to define all the specular map in the glsl files that relies on `reliefMapping_fp.glsl` instead of declaring it in `reliefMapping_fp.glsl` and doing some compute there…).

illwieckz: (hence the bizarre macro with hardcoded names in `reliefMapping_fp.glsl` to avoid a pile of ifdef in all other files)…

src/engine/renderer/glsl_source/computeLight_fp.glsl
illwieckz: this is to not have to put ifdefs in all files that does specular mapping to pass or not the variables… :-/

The clutter to add the ifdef to pass the variable or not is the same to do this compute or not :-/

src/engine/renderer/tr_shade.cpp
slipher: I don't know, probably none will shed a tear if you delete it.

src/engine/renderer/glsl_source/computeLight_fp.glsl
slipher: It's OK to use a nonexistent variable in the macro argument as long as it does not appear in the final preprocessed output. This is possible since preprocessing is a conceptually separate step from compiling.

illwieckz: The thing is that if the function expects a parameter, you have to set this parameter in the calling code and if this parameter does not exist anytime, you have to put ifdef around every call in every file calling this function.

We know this functions is called by every code doing specular stuff _but_ the liquid shader. If there was not such exception there would be no reason to declare the stuff in the calling code, only in the common library.

There is another alternative I tried and it works but I don't know if it's clean enough: it would be to define an useless `var_TexSpecular` in `liquid_vp.glsl` so `computeLight_fp.glsl` is able to compile while prepended to `liquid_fp.glsl`. The code would not be run in any way, the symbol just needs to be defined so the whole file compile including code that's not used.

slipher: What I'm saying is you can do something like this:
```
#ifdef USE_FOO
#define foo(foo_factor) get_foo(foo_factor, 3)
#else
#define foo(foo_factor) 1.0
#endif

...

#ifdef USE_FOO
float foo_factor = 2;
#endif

...

float x = foo(foo_factor);
```

illwieckz: That can't help because in the liquid case, `foo_factor` does not exist while `USE_FOO` is defined.

The liquid shader use the exact same specular code but don't get specular map from files.
So we would be able to define both `USE_SPECULAR_MAPPING` and `WITHOUT_SPECULAR_MAP` for it but it would defeat the macro count reduction.

illwieckz: I removed this commit refactoring specular stuff since I may have some other PR to merge before this refactoring to make life easier.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: yes, the code was already there, I probably just forgot to push it.

src/engine/renderer/tr_shade.cpp
illwieckz: deleted

src/engine/renderer/tr_init.cpp
illwieckz: I extracted it to its own commit

slipher:
LGTM

illwieckz:
@slipher can you look at the latest commit I added? I just noticed I forgot to include this one in this PR while extracting some commits from my workspace to branches for other PR to come. This is very straightforward, it just use parallaxDepth/Offset thing in reliefMapping_fp only hence no need to pass the variables anywhere parallax is done hence there is only one place to put ifdef. Basically they are defined where they are used. This avoids boiler plate.

slipher:
:+1:

illwieckz:
I just force pushed because I squashed this very minor thing (both are vec2, no need for mask):

```diff
diff --git a/src/engine/renderer/glsl_source/liquid_fp.glsl b/src/engine/renderer/glsl_source/liquid_fp.glsl
index 7154718c8..ac7a19099 100644
--- a/src/engine/renderer/glsl_source/liquid_fp.glsl
+++ b/src/engine/renderer/glsl_source/liquid_fp.glsl
@@ -90,7 +90,7 @@ void  main()
 
        // calculate the screen texcoord in the 0.0 to 1.0 range
        vec2 texScreen = gl_FragCoord.st * r_FBufScale;
-       vec2 texNormal = var_TexNormal.st;
+       vec2 texNormal = var_TexNormal;
 
 #if defined(USE_PARALLAX_MAPPING)
        // ray intersect in view direction
```

PR #189, opened 2019-03-24
WIP: glsl/normal: use ifdef for normal z reconstruction

illwieckz:
**DO NOT MERGE** until the `#ifdef` bug is fixed, see #190.

This is a simple change that makes glsl relying on `#ifdef` instead of an `uniform` to test for the fact the normalmap is embedding heightmap in alpha channel or not. This makes GMA965 able to render parpax default scene with `11` fps instead of `10` (+10%) (unnoticeable on R9 390X).

This is not merged yet because of a nasty bug, it looks like the define is not defined any time as while I move in map, the wrong code is sometime run instead of the expected one:

[glsl ifdef](https://dl.illwieckz.net/b/daemon/bugs/glsl-ifdef/unvanquished_2019-03-23_203425_000.jpg)

[glsl ifdef](https://dl.illwieckz.net/b/daemon/bugs/glsl-ifdef/unvanquished_2019-03-23_203429_000.jpg)

The bug is on the wall texture, not on the blooming reactor.

I reproduced the bug on both `radeonsi` (R9 390X) and `i965` (Haswell) drivers, this is likely to not be a driver bug but a bug in dæmon code, the ifdef code is very cryptic and would benefit a lot from a rewrite.

I'm basically pushing this branch so people have code to debug the `#ifdef` bug (#190).

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -1456,12 +1456,12 @@ GLShader_lightMapping::GLShader_lightMapping( GLShaderManager *manager ) :
 	u_ModelViewProjectionMatrix( this ),
 	u_ParallaxDepthScale( this ),
 	u_ParallaxOffsetBias( this ),
-	u_HeightMapInNormalMap( this ),
 	u_NormalFormat( this ),
 	u_numLights( this ),
 	u_Lights( this ),
 	GLDeformStage( this ),
 	GLCompileMacro_USE_DELUXE_MAPPING( this ),
+	GLCompileMacro_USE_HEIGHTMAP_IN_NORMALMAP( this ),
 	GLCompileMacro_USE_PARALLAX_MAPPING( this ),
 	GLCompileMacro_USE_PHYSICAL_SHADING( this )
 {
@@ -1507,7 +1507,6 @@ GLShader_vertexLighting_DBS_entity::GLShader_vertexLighting_DBS_entity( GLShader
 	u_VertexInterpolation( this ),
 	u_ParallaxDepthScale( this ),
 	u_ParallaxOffsetBias( this ),
-	u_HeightMapInNormalMap( this ),
 	u_NormalFormat( this ),
 	u_EnvironmentInterpolation( this ),
 	u_LightGridOrigin( this ),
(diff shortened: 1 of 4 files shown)


slipher:
A couple suspicious things:
- You only set the height mapping macro if the parallax cvar is enabled, but some of the GLSL that depends on it is outside of ifdef parallax blocks.
- Liquid shader has the macro but it is never enabled.

illwieckz:
>  You only set the height mapping macro if the parallax cvar is enabled, but some of the GLSL that depends on it is outside of ifdef parallax blocks.

good catch!

> Liquid shader has the macro but it is never enabled.

right, this code is not yet used but it's better to not have bugs in dormant code

I'll reread the stuff to be sure this variable is set in any code that use normal map (they may be more), not only those that use parallax.

illwieckz:
I postpone changes to this one for after #188 merge since once #188 is merged I would be able to do things like that:

```c
if (normalMapping || parallaxMapping )
```

illwieckz:
I rebased on master but the #190 bug is still there so merge is blocked

slipher:
It looks like in some shaders, you are doing `SetHeightMapInNormalMap` only if a condition is met. So when the condition is not met, you get a random value left over from the last time the shader was used.

illwieckz:
do someone can explain me why the latest commit is fixing the bug?

slipher:
Do you really need the 2nd commit for it to work? It looks like the first one was changed to avoid the bug that I talked about before.

illwieckz:
Yes, the second commit is required to fix the bug. The changes I added to the first commit to fix the issue you talked about before is not enough.

illwieckz:
well, that only fixes it for now, if I add some commits editing some unrelated stuff the bug comes back again… the unrelated commit I was working on was a commit merging reflection in lightmap stage to avoid double parallax compute… this one added a macro since the reflectCube glsl would be useless in that case, and the bug came back… I know the macro thing is tricky since each macro is a bit of one integer,  may be we overflow that integer?

illwieckz:
@slipher, note that `MAX_SHADER_MACROS` is set to `9` here:

https://github.com/DaemonEngine/Daemon/blob/fb8b561ed9aea150d755388d1f9b309167af8325/src/engine/renderer/gl_shader.h#L33

illwieckz:
But I never get this error in any way:
https://github.com/DaemonEngine/Daemon/blob/fb8b561ed9aea150d755388d1f9b309167af8325/src/engine/renderer/gl_shader.h#L201-L204

illwieckz:
This may be a bug similar to the one fixed in #211

illwieckz:
This will be obsoleted by my PR introducing loose heightmap support.

I'm not sure there is a bug in that ifdef code, similar artifact can happen with zero normalscale (hence undefined normalscale). I noticed some issues with that other feature, so maybe it's the same.

PR #192, opened 2019-03-26
WIP: glsl/parallax: ifless reflection mapping 🤔

illwieckz:
@gimhael or @cmf028, if `if` in glsl is very most costly than operations, why is this giving me no one extra fps _at all_?

What if I unroll the loops?

Note that I have an R9 390X, which is a beast, and maybe that beast is modern enough to handle `if` properly?

### DIFF
--- a/src/engine/renderer/glsl_source/reliefMapping_fp.glsl
+++ b/src/engine/renderer/glsl_source/reliefMapping_fp.glsl
@@ -302,6 +302,7 @@ vec2 ParallaxTexOffset(sampler2D normalMap, vec2 rayStartTexCoords, float parall
 
 		float heightMapDepth = topDepth - texture2D(normalMap, rayStartTexCoords + displacement * currentDepth).a;
 
+		/*
 		if(bestDepth > 0.996) // if no depth found yet
 		{
 			if(currentDepth >= heightMapDepth)
@@ -309,6 +310,10 @@ vec2 ParallaxTexOffset(sampler2D normalMap, vec2 rayStartTexCoords, float parall
 				bestDepth = currentDepth;
 			}
 		}
+		*/
+
+		bool test = bestDepth > 0.996 && currentDepth >= heightMapDepth;
+		bestDepth = (int(test) * currentDepth) + (int(!test) * bestDepth);
 	}
 
 	currentDepth = bestDepth;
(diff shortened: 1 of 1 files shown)


illwieckz:
well, I unrolled the loops and I get no one extra fps… 😕

slipher:
With regard to the `for`-loop change, this is the same as what the compiler probably does internally. The before/after generated code could very well be identical.

gimhael:
It is quite normal that a shader optimization doesn't directly lead to a measurable FPS increase. The GPU processes the draw calls in a pipeline of several steps, and as in real life, the slowest step determines the overall speed of the pipeline.
If e.g. the fragment shader already produces more data than the ROPs can handle, then making the fragment shader even faster will not help overall performance. (The ROPs are the units that render the fragment shader outputs into the framebuffer.)

illwieckz:
The thing is that using an obviously quicker algorithm in that place without modifying anything else (like in #193) multiplies fps by 3… This look to be the slowest step everything is waiting for. ¯\\_(ツ)_/¯

Without this code at all, a given solarium scene with an active flamethrower produces 200fps, with that code on lightmap and real time light the scene produces 30fps. With that code on lightmap and quicker code from #193 on real time light the scene produces 90fps. Everything else is an unmodified black box at this point. Everything looks to be waiting on this…

slipher:
The point is not that that a conditional is inherently bad, it's that in general you have to execute all possible code paths. So if you have `if(...) { A } else { B }` then the time to execute will be A + B (while it would be at most max(A, B) on a single processor). Or if you try to speed up a loop by having an early exit condition, it won't help because you have to wait for the maximum possible number of iterations for it to finish.

Doing a substitution like `cond ? x : y` -> `cond * x + !cond * y` is a bad idea because such things should be left to the compiler. It will have the knowledge of which way is efficient to achieve the goal.

illwieckz:
Oh thank you, that was the answer I was looking for. Well, I don't know why but my first assumption was wrong, it's not that conditions are inherently more costly than operations, it's conditions does not help to avoid operations.

So if I'm right, conditions are not bad in a way branching would be very costly by itself and must be avoided because of that branching cost itself, but conditions are bad in the way that all the code within the condition block is run even if the condition is false, the result just being discarded. 

So in that way, avoiding the condition by making sure an operation is done in any way is not able to help, first I'm not replacing a costly operation by a non-costly one, second I just make sure that the code is always run to fix the fact the code is always run.

Doing what I did may even be able to slow down the compute if the compute trick that is meant to produce the same result without usage of `if` is slower than the compute within the if block itself.

Note: in that case I noticed no difference, on both R9 390X and Haswell.

So, I got my answer, thank you very much. Closing it.

PR #193, opened 2019-03-26
quick and cheap parallax for real time lights

illwieckz:
It jumps from 31fps to 95fps on hd1080p (110fps in a common scene) on an R9 390X with flame thrower on an open scene that is known to be far slower than the average, I don't see anything wrong noticeable.

Produces 65fps on 2K on that costly scene (90fps in a common scene) and 36fps on 4K (45fps in common scene).

Note that without flame thrower, I get 200~225fps (hd1080p), 185/225fps (2K), 100/140fps (4K).

Basically this quick parallax is only done to be sure the realtime light is not shifted from the usual lightmapped scene (which is parallaxed with a more efficient algorithm).

I'll post screenshots later.

This hack makes a Dæmon based game performant enough to be releasable with parallax enabled or at least available to the user.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -584,6 +584,11 @@ static std::string GenEngineConstants() {
 		AddDefine( str, "r_normalMapping", 1 );
 	}
 
+	if ( r_quickLightParallaxMapping->integer )
+	{
+		AddDefine( str, "r_quickLightParallaxMapping", 1 );
+	}
+
 	if ( r_specularMapping->integer )
 	{
 		AddDefine( str, "r_specularMapping", 1 );
--- a/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
+++ b/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
@@ -948,7 +948,11 @@ void	main()
 
 #if defined(USE_PARALLAX_MAPPING)
 	// compute texcoords offset from heightmap
+#if defined(r_quickLightParallaxMapping)
 	vec2 texOffset = QuickParallaxTexOffset(texCoords, viewDir, tangentToWorldMatrix);
+#else
+	vec2 texOffset = ParallaxTexOffset(texCoords, viewDir, tangentToWorldMatrix);
+#endif
 
 	texCoords += texOffset;
 #endif // USE_PARALLAX_MAPPING
--- a/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
+++ b/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
@@ -948,7 +948,7 @@ void	main()
 
 #if defined(USE_PARALLAX_MAPPING)
 	// compute texcoords offset from heightmap
-	vec2 texOffset = ParallaxTexOffset(texCoords, viewDir, tangentToWorldMatrix);
+	vec2 texOffset = QuickParallaxTexOffset(texCoords, viewDir, tangentToWorldMatrix);
 
 	texCoords += texOffset;
 #endif // USE_PARALLAX_MAPPING
--- a/src/engine/renderer/glsl_source/reliefMapping_fp.glsl
+++ b/src/engine/renderer/glsl_source/reliefMapping_fp.glsl
@@ -108,6 +108,23 @@ vec3 NormalInWorldSpace(vec2 texNormal, mat3 tangentToWorldMatrix)
 }
 
 #if defined(USE_PARALLAX_MAPPING)
+// very poor but very fast parallax
+// do not use for something else than real time lights
+vec2 QuickParallaxTexOffset(vec2 rayStartTexCoords, vec3 viewDir, mat3 tangentToWorldMatrix)
+{
+	// compute view direction in tangent space
+	vec3 tangentViewDir = normalize(viewDir * tangentToWorldMatrix);
+
+	vec2 displacement = tangentViewDir.xy * -u_ParallaxDepthScale / tangentViewDir.z;
+
+	float topDepth = 1.0 - u_ParallaxOffsetBias;
+
+	float depth = topDepth - texture2D(u_NormalMap, rayStartTexCoords).a;
+
+	// depthmap value at current texture coordinates
+	return depth * displacement;
+}
+
 // compute texcoords offset from heightmap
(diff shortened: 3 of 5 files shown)


illwieckz:
This is an Unvanquished map layout for Xonotic solarium map:

```
arm -2250.734375 -1314.114990 -116.574997 0.000000 -155.000000 0.000000 0.000000 0.000000 1.000000 0.000000 0.000000 0.000000
reactor -2251.954590 -1416.356567 -123.875000 0.000000 -97.000000 0.000000 0.000000 0.000000 1.000000 0.000000 0.000000 0.000000
overmind -1768.764893 373.085999 -112.875000 0.000000 -93.000000 0.000000 0.000000 0.000000 1.000000 0.000000 0.000000 0.000000
telenode -2204.000000 -980.000000 -91.875000 0.000000 0.000000 0.000000 0.000000 0.000000 1.000000 0.000000 0.000000 0.000000
medistat -2249.424316 -1235.030640 -127.875000 0.000000 -90.000000 0.000000 0.000000 0.000000 1.000000 0.000000 0.000000 0.000000
eggpod -1905.912476 268.833099 -80.875000 0.000000 137.000000 0.000000 0.000000 0.000000 1.000000 0.000000 0.000000 0.000000
```

put it in `~/.local/share/unvanquished/game/layouts/solarium/base.dat` and load the map this way: `devmap solarium base`.

This makes testing flame-thrower in a parallax enabled map easy.

slipher:
This may be a dumb idea, but I wonder if it's possible to do something like batching several lights together so that the parallax calculation doesn't have to be redone for every light.

illwieckz:
or to do the parallax compute for the given pixel one time whatever the lighting…

illwieckz:
@gimhael or @cmf028 do you know if there is a way to compute the texOffset one time per pixel per frame then to store it so other shaders may use it? for example abusing a texture a that would not be displayed?

illwieckz:
Basically uploading two times the texture if it's a diffuse one and r_parallaxMapping is enabled, compute offset and store the texOffset in r and g (x and y) in an early stage, then bind the given map in stages that would need it…

illwieckz:
I think it looks good enough to be enabled by default. It's known to double the framerate on AMD GPU Radeon R9 390X.

I made a cvar to disable it, but for some reason I'm not able to go through the slow code path when I disable it, what can be happening?

illwieckz:
Hmmm, for some reason the shaders are not re-compiled on vid_restart. If I force shader recompilation by deleting `~/.local/share/unvanquished/glsl/*`, it works.

illwieckz:
option enabled:

[with quick light parallax](https://dl.illwieckz.net/b/unvanquished/shots/quick-light-parallax/unvanquished_2019-07-06_221914_000.jpg)

[without quick light parallax](https://dl.illwieckz.net/b/unvanquished/shots/quick-light-parallax/unvanquished_2019-07-06_222013_000.jpg)

illwieckz:
@slipher do you think this issue can be related to c4772b5a87c7589cbe78ae1419ec0b20a424e2a7 ?

illwieckz:
maybe it's not related to c4772b5a87c7589cbe78ae1419ec0b20a424e2a7, it's expected a shader is invalidated when the checksum does not match, the issue is probably there, maybe the defines are not added to the text on which the checksum is computed?

illwieckz:
#211 ships the fix for the shader not being recompiled on `vid_restart` after tweaking `r_quickLightParallaxMapping` cvar.

Unless someone has a better idea for the cvar name, this PR is ready to merge on my point of view.

illwieckz:
@slipher once _done correctly_, that checksum-based shader invalidation makes `CVAR_SHADER` and `r_recompileShaders` useless, so c4772b5a87c7589cbe78ae1419ec0b20a424e2a7 is OK.

cmf028:
You've implemented parallax mapping which is known to have severe artifacts in the case of occlusion.

The existing implementation is known as "Steep parallax mapping", "Relief mapping", or "parallax occlusion mapping".

Your code change applies only to the very slow forward lighting path (I didn't even know that was still working). The forward lighting path will render the geometry (and compute parallax, access textures, etc) once for each light in the scene. The tiled lighting path will render geometry just once (computing parallax, etc) no matter how many lights are in the scene. Since the flamer spawns double digit counts of lights when in use, I'd image this would be much faster.

SomaZ:
What about doing some kind of hybrid approach? I haven't looked much at the rendering code yet, but do you distinguish between shadowed and unshadowed analytical lights? To solve the bad performance one could write the texturecoordinates (which get altered based on height and the view vector on parallax occlusion mapped surfaces) in a previous pass via mrt. Should be much faster than it is right now and you dont need to worry about packing the shadow samplers in an atlas. Just my two cents.

briancullinan:
Interesting, I was planning on pulling this in https://ioquake3.org/2008/03/04/raytracing-ioquake3/

How is this related? The source patch is further down and I have a branch with it applied at the time of release, just need to pull it forward to latest.

illwieckz:
You mean you're trying to rebase that raytracing patch on latest ioquake3 master and in order to save render time you're interested in doing cheap parallax with raytraced lights the same way I experimented?

briancullinan:
@illwieckz saving on lighting compile time is exactly what I was hoping for! Did you already try this particular patch? I thought the parallel C++ stuff was interesting and could be used to background load some extra deferred models. 

Ioq3 has a note in it that shadows are broken, maybe Xonotic fixed this already? Not sure of this is relevant but it looks nice. https://u.teknik.io/JqNdg.jpg

illwieckz:
The patch saves a lot of time when using a forward renderer as, when using complex effects with many lights like a flame thrower, parallax has to be computed multiple time for the same pixel, one time per light. But that's because the forward renderer is inefficient by design.

When using the tiled renderer (I was not using by mistake at the time I made the patch), the parallax is computed once for the diffuse and reused for all the lights, so our tiled renderer makes this patch absolutely useless.

I you want to a computation-intensive path like raytracing, I'll would recommend you to implement a way to only compute the parallax one for each pixel, not on every ray bounce… Doing many cheap quick parallax computations would be slower than doing a good-quality relief mapping only once for all. So basically I recommend you to look for a design that _do not require_ to use such tweak instead of looking for such tweak.

Our forward renderer is meant to be deleted, so I don't even need to keep that PR open.

PR #195, opened 2019-03-31
tr_shader: do not enable deluxe map on material without normal map

illwieckz:
fix a regression introduced in 3d53075

before regression:  
[deluxe without normal](https://dl.illwieckz.net/b/daemon/bugs/deluxe-without-normal/unvanquished_2019-03-31_093518_000.jpg)

after regression:  
[deluxe without normal](https://dl.illwieckz.net/b/daemon/bugs/deluxe-without-normal/unvanquished_2019-03-31_093634_000.jpg)

after fix:  
[deluxe without normal](https://dl.illwieckz.net/b/daemon/bugs/deluxe-without-normal/unvanquished_2019-03-31_094947_000.jpg)

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -1154,9 +1154,11 @@ static void Render_lightMapping( int stage, bool asColorMap, bool normalMapping,
 
 	GL_State( stateBits );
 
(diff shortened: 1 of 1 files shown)


illwieckz:
@slipher, may be a dumb question, but I guess I can replace this:

```c
normalMapping = normalMapping && deluxeMapping;
```

by this:

```c
normalMapping &= deluxeMapping;
```

right?

(I only did this with bit fields before, not boolean, but I see no reason why it wouldn't work)

slipher:
You can. It doesn't actually look nicer though.

PR #196, opened 2019-03-31
Fix console font resizing

### LINKED ISSUE #162: Can't change console size without quitting the engine first
This is something I discovered while working on #161.

To change console size, people have to tweak `cl_consoleFontSize` then to quit the engine entirely before starting it again. Doing `vid_restart` is not enough.

It means people have to quit the currently played game to resize the console font.

I don't know if it's a bug or if it's a design issue. It may be a design issue since the console is probably one of the earliest things that is spawned but I hope it's not and I hope a fix can be easily implemented.

slipher:
\+ misc tiny changes

Fixes #162.

### DIFF
--- a/src/engine/audio/SoundCodec.cpp
+++ b/src/engine/audio/SoundCodec.cpp
@@ -102,7 +102,7 @@ AudioData LoadSoundCodec(std::string filename)
 		return AudioData();
 	}
 
-	audioLogs.Warn("Sound file %s not found.", filename);
+	audioLogs.Notice("Sound file '%s' not found.", filename);
 	return AudioData();
 
 }
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -1906,8 +1906,6 @@ and determine if we need to download them
 */
 void CL_InitDownloads()
 {
-	char missingfiles[ 1024 ];
-
 	// TTimo
 	// init some of the www dl data
 	clc.bWWWDl = false;
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -3019,6 +3019,10 @@ bool CL_InitRenderer()
 		return false;
 	}
 
+	cl_consoleFont = Cvar_Get( "cl_consoleFont", "fonts/unifont.ttf",  CVAR_LATCH );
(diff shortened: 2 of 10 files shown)


src/engine/audio/SoundCodec.cpp
```diff
@@ -102,7 +102,7 @@ AudioData LoadSoundCodec(std::string filename)
 		return AudioData();
 	}
 
-	audioLogs.Warn("Sound file %s not found.", filename);
+	audioLogs.Notice("Sound file '%s' not found.", filename);
```
illwieckz: I would not be surprised is the given warning is there to tell us some code is trying to load a file without checking if it exists first, so silencing the warning would not be the right fix.

illwieckz: i.e. this is probably better to take care of that in another PR, I was planning to investigate on this but I don't own this, but I know there is some code trying to load optional file the required way in other places (like the model code), it may be a game issue, not an engine one.

illwieckz: see 7774ff80c78f2083e23027c7a24561f07b64cf9f that is expected, the code trying to use a sound file that does not exist has to be fixed instead, so can you remove this commit? If you want I may look myself for why the code is trying to load files that does not exist, I'll was going to do it one day in any way.

src/engine/renderer/tr_local.h
```diff
@@ -1093,8 +1093,6 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 
 		uint32_t        stateBits; // GLS_xxxx mask
 
-		bool            heightMapInNormalMap;
```
illwieckz: my bed!

src/engine/client/cl_main.cpp
```diff
@@ -1906,8 +1906,6 @@ and determine if we need to download them
 */
 void CL_InitDownloads()
 {
-	char missingfiles[ 1024 ];
```
illwieckz: :+1:

src/engine/audio/SoundCodec.cpp
slipher: Yeah I know, I asked for it :) Didn't know it would lead to warning spam though. I've filed https://github.com/Unvanquished/Unvanquished/issues/1110 so the issue will not be buried, but in the meantime before anyone gets to looking at it, let's silence the warning.

PR #198, opened 2019-04-07
tr_shader: ignore inactive stages in collapsible stage lookup

illwieckz:
ignore inactive stages in collapsable stage lookup


slipher:
LGTM

PR #199, opened 2019-04-07
WIP: set a minimum amount of lighting

illwieckz:
The last time I updated [The Dark Mod](http://www.thedarkmod.com/main/), I got a bug: there was no lighting. But I noticed everything was still visible. So I discovered that this stealth game that rely a lot on darky zones never have pure dark zones:

[no light](https://dl.illwieckz.net/b/thedarkmod/bugs/no-light/training_mission_2019-02-03_02.07.17.jpg)

[no light](https://dl.illwieckz.net/b/thedarkmod/bugs/no-light/training_mission_2019-02-03_02.09.46.jpg)

[no light](https://dl.illwieckz.net/b/thedarkmod/bugs/no-light/training_mission_2019-02-03_02.14.36.jpg)

I remembered the usual complain about the Unvanquished game being too darky, and I thought the idea was nice and I wanted the feature in Dæmon too.

Then I was yet another time annoyed by the ugly darky blotches of our non-sRGB lightmaps. And then the light came to my mind. If there is no pure dark zones, there can't be dark blotches.

Dark blotches came from the fact the current color curve does not have so much steps in the dark part, hence the shadows coming from average grey to pure black in one step.

So I implemented this hack. For every shadow color component between `0` and `0.18`, rescale it between `0.06` and `0.18`. The `0.06` value was chosen the empirical way by testing with some maps suffering a lot from the bug, and some other maps to ensure nothing wrong occurs. The `0.18` value was arbitrarily chosen without much thinking.

I really have no idea about the way to name this feature. Currently, the cvar is named `r_floorLight`, I highly recommend yourself to try this.

It's an easy hack to greatly improve the looking of our maps. sRGB lightmaps will tamper the need for such hack for good looking purpose, but since it's a good feature to always keep a minimum amount of light, it may be good to keep the feature and just adjust the numbers for the sRGB curve.

Note: it looks less good in thumbnails because the dark stuff being less contrasted, but it looks better in game this way:

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190038_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190100_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190234_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190257_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190403_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190425_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190701_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190723_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190830_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190847_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_191519_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_191545_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_191841_000.jpg)

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_191905_000.jpg)

There is one issue though, the usual "black" texture hack does not work anymore:

[floorlight](https://dl.illwieckz.net/b/daemon/wip/floorlight/unvanquished_2019-04-07_190440_000.jpg)

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -579,6 +579,11 @@ static std::string GenEngineConstants() {
 		AddDefine( str, "r_showLightTiles", 1 );
 	}
 
+	if ( r_floorLight->integer )
+	{
+		AddDefine( str, "r_floorLight", 1 );
+	}
+
 	if ( r_normalMapping->integer )
 	{
 		AddDefine( str, "r_normalMapping", 1 );
--- a/src/engine/renderer/glsl_source/lightMapping_fp.glsl
+++ b/src/engine/renderer/glsl_source/lightMapping_fp.glsl
@@ -41,6 +41,26 @@ IN(smooth) vec4		var_Color;
 
 DECLARE_OUTPUT(vec4)
 
+#if defined(r_floorLight)
+// returns a pseudo-boolean (1 or 0) vec3 given the component
+// is smaller to the given float or not
+// for example isSmaller(vec3(1, 2, 3), 2) returns vec3(1, 0, 0)
+vec3 isSmaller(vec3 v, float max)
+{
+	return floor(min(v, max) / max) * -1 + 1;
+}
+
+// shift and rescale a vec3 from any [min1, max1] range to [min2 - max2] range
+// for example it can be used to rescale [0, 12] to [6, 32]
+vec3 shiftRange(vec3 v, float min1, float max1, float min2, float max2)
+{
+	v -= min1;
+	v *= (max2 - min2) / (max1 - min1);
+	v += min2;
+	return v;
+}
+#endif // r_floorLight
+
 void	main()
 {
 	// compute view direction in world space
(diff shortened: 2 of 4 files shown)


illwieckz:
Note: for some unknown reason I have to delete `~/.local/share/unvanquished/glsl` between each `vid_restart` when I switch `r_floorLight` value.

illwieckz:
Note that when maps are not really contrasted with this feature it just means the given maps are very low light, since it would mean that the average light of the given scenes would be very near the step from where the grey is not dark…

slipher:
Well this is a great cheat code for the humans :stuck_out_tongue: 

Do we really want to have settings like that though? Which give an advantage to people who know to turn them on?

Alternatively, it could be turned on by default, but it seems bad that the renderer is not faithfully rendering what the output image should be according to the inputs.

necessarily-equal:
I must say I am unable to see a difference between your screenshots. That's probably my fault I guess, between strong redshift, low screen luminosity, and a laptop screen. But that also means those area are still pitch black for me :X

necessarily-equal:
Just as a note, I've been thinking of writing a shader for alien nightvision; and some are considering torchs https://forums.unvanquished.net/viewtopic.php?f=33&t=2097&p=17541#p17682. Maybe that could be a way to go?

illwieckz:
> That's probably my fault I guess

Your hardware fault, you're fine. 😁

1. This experiment was trying to workaround color banding.
2. Also, it can be a feature request for the engine itself (for Unvanquished game or another) as an optional feature.

About color banding, two things are currently evaluated: dithering (fake gradient by mixing pixels), sRGB lightmaps (better colorspace that is less aggressive on low values), HDR lightmaps (more data means less banding).

About the feature itself, games like The Dark Mod reveal some game may want to have such feature, so as a standalone engine it may be good to offer that option, but without an explicit need that's a low priority.

bmorel:
> There is one issue though, the usual "black" texture hack does not work anymore (well, the texture works, but the illusion does not work anymore):

What was that hack (the black texture) used for?

Also, I like the results on the screenshots. The difference is clear, it might even make some maps actually playable (hangar28), but this needs to be on by default, if we let this feature slip in gamelogic.
Is there any progress on the alternatives, non-hacky, ways? If not, maybe this hack is better than nothing?

VReaperV:
> Is there any progress on the alternatives, non-hacky, ways? If not, maybe this hack is better than nothing?

There is now, in #1550. Funnily enough, the arbitrarily chosen 0.18 value here is mid grey, and is used by default as the dark-area point cut-off in #1550. It also doesn't break the black texture thing.

PR #200, opened 2019-04-07
Delete and move around stuff...

slipher:


illwieckz:
- _Remove unused startup code_: LGTM
- _Fix 2 unused variable warnings_: LGTM
- _Move some download functions to a new file_: LGTM
- _Remove unused "binary message" code_: what the code did?

slipher:
I don't know. Apparently it is some hack added in ET.

slipher:
I can say roughly what it does which is some out of band way that the sgame can send data to the cgame.

Issue #201, opened 2019-04-10
Do cgame restarts only at the top level


slipher:
Currently, operations that involve restarting the cgame VM can be called from many different places. All of the shutting down and reinitializing is done, and then the rest of the frame continues executing from wherever it left off. So everything is in a weird state then and there are certain cases where it crashes. 

I am working on changing things so that cgame restarts are only done at the top level. The model is `Sys::Drop` which already works like this: it throws a `DropErr` exception to unwind the stack before doing the cgame restart. My biggest question is: To also use exceptions for other cgame restarts, or store the request for a state change in a global variable and wait until the end of the frame to do it.

Points in favor of an exception:
- Using the same mechanism for all cases
- If you executed a sequence of commands that included e.g. `vid_restart` then some other command, they would be executed in the correct order. (I don't know any interesting use cases for this though).
- No weird case to handle when more than one restart is requested during a frame.

Points in favor of not using an exception:
- Less code that needs to leave things in a correct state when an exception is thrown in it (though most of it would already have this requirement because of `Sys::Drop` being possible)
- It could be confusing/bad style to use an exception for something that's not an error.

slipher:
I should mention this actually applies to the server and sgame restarts as well, not only cgame.

PR #203, opened 2019-04-21
Preliminary work for VM restarts

slipher:
This is some stuff that has come up while working on #201.


DolceTriade:
Very good clean up

PR #204, opened 2019-04-22
Replace CVAR_SHADER with CVAR_LATCH

slipher:


illwieckz:
looks good to me, by the way it's useful to remind that CVAR_LATCH is only supported in legacy cvars like the renderer ones at this time:
https://github.com/DaemonEngine/Daemon/blob/master/src/engine/framework/CvarSystem.h#L101

PR #206, opened 2019-04-28
some tr_shade/glsl code cleanup

illwieckz:
While I was working on a branch to collapse the reflection cube map stage with the common lightmap one, I did this clean-up.

It makes working on this part of the code easier, but don't have to wait for featural changes to be merged.

Note that one commit makes sure reflective specular is not used whith PBR assets since current reflective specular code is not PBR compatible.

### DIFF
--- a/src/engine/renderer/glsl_source/dispersion_C_fp.glsl
+++ b/src/engine/renderer/glsl_source/dispersion_C_fp.glsl
@@ -37,26 +37,26 @@ DECLARE_OUTPUT(vec4)
 void	main()
 {
 	// compute incident ray
-	vec3 I = normalize(var_Position - u_ViewOrigin);
+	vec3 incidentRay = normalize(var_Position - u_ViewOrigin);
 
 	// compute normal
-	vec3 N = normalize(var_Normal);	// FIXME normalize?
+	vec3 normal = normalize(var_Normal);	// FIXME normalize?
 
 	// compute reflection ray
-	vec3 R = reflect(I, N);
+	vec3 reflectionRay = reflect(incidentRay, normal);
 
 	// compute fresnel term
-	float fresnel = u_FresnelBias + pow(1.0 - dot(I, N), u_FresnelPower) * u_FresnelScale;
+	float fresnel = u_FresnelBias + pow(1.0 - dot(incidentRay, normal), u_FresnelPower) * u_FresnelScale;
 
 	// compute reflection color
-	vec3 reflectColor = textureCube(u_ColorMap, R).rgb;
+	vec3 reflectColor = textureCube(u_ColorMap, reflectionRay).rgb;
 
 	// compute refraction color using a refraction ray for each channel
 	vec3 refractColor;
 
-	refractColor.r = textureCube(u_ColorMap, refract(I, N, u_EtaRatio.x)).r;
-	refractColor.g = textureCube(u_ColorMap, refract(I, N, u_EtaRatio.y)).g;
-	refractColor.b = textureCube(u_ColorMap, refract(I, N, u_EtaRatio.z)).b;
+	refractColor.r = textureCube(u_ColorMap, refract(incidentRay, normal, u_EtaRatio.x)).r;
+	refractColor.g = textureCube(u_ColorMap, refract(incidentRay, normal, u_EtaRatio.y)).g;
+	refractColor.b = textureCube(u_ColorMap, refract(incidentRay, normal, u_EtaRatio.z)).b;
 
 	// compute final color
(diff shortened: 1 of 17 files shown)


src/engine/renderer/tr_shade.cpp
```diff
@@ -2806,20 +2802,15 @@ void Tess_StageIteratorGeneric()
...
-									Render_lightMapping( stage, false, true, false );
+									Render_lightMapping( stage, false, true );
 								}
 								else
 								{
-									Render_lightMapping( stage, false, false, false );
+									Render_lightMapping( stage, false, false );
 								}
 							}
 							else if ( backEnd.currentEntity != &tr.worldEntity )
```
slipher: This can be reached if r_vertexLighting = 0 and tess.lightmapNum is invalid which doesn't seem right.

src/engine/renderer/tr_shader.cpp
```diff
@@ -4352,7 +4352,8 @@ static void CollapseStages()
 					ASSERT_GE(lightStage, 0);
 					Log::Debug("found glow map with custom lightmap stage in '%s' shader, not collapsing", shader.name);
 					stages[ glowStage ].type = stageType_t::ST_COLORMAP;
-					stages[ glowStage ].bundle[ TB_COLORMAP ] = stages[ glowStage ].bundle[ 0 ];
+					// TB_COLORMAP == 0
+					// stages[ glowStage ].bundle[ TB_COLORMAP ] = stages[ glowStage ].bundle[ 0 ];
```
slipher: I'd say drop this commit. The compiler can figure out on its own that the assignment is not needed and optimize it out.

src/engine/renderer/glsl_source/vertexLighting_DBS_entity_fp.glsl
```diff
@@ -94,8 +94,8 @@ void	main()
 	// compute normal in world space from normalmap
 	vec3 N = NormalInWorldSpace(texCoords, tangentToWorldMatrix);
 
-	// compute the specular term
-#if defined(USE_REFLECTIVE_SPECULAR)
+	// compute the specular term, not compatible with PBR yet
+#if defined(USE_REFLECTIVE_SPECULAR) && !defined(USE_PHYSICAL_SHADING)
```
slipher: I'm not sure exactly what it does but it might be interesting to look at defining `HasConflictingMacros` for this shader.

src/engine/renderer/glsl_source/dispersion_C_fp.glsl
```diff
@@ -37,26 +37,26 @@ DECLARE_OUTPUT(vec4)
 void	main()
 {
 	// compute incident ray
-	vec3 I = normalize(var_Position - u_ViewOrigin);
+	vec3 incicentRay = normalize(var_Position - u_ViewOrigin);
```
slipher: spelling

src/engine/renderer/tr_shade.cpp
illwieckz: is it something introduced by my change or something that was already there?

src/engine/renderer/tr_shader.cpp
illwieckz: ok, I still think a comment may be useful in anyway…

src/engine/renderer/glsl_source/dispersion_C_fp.glsl
illwieckz: good catch!

src/engine/renderer/glsl_source/vertexLighting_DBS_entity_fp.glsl
illwieckz: hmmm, yes I forgot it, I just remember that it's very cryptic like all the macro code (and I'm even not sure that macro code works correctly at all) :confused:

src/engine/renderer/glsl_source/dispersion_C_fp.glsl
illwieckz: I don't know what can we do to improve glsl debugging, stuff get easily unnoticed…

src/engine/renderer/tr_shade.cpp
slipher: Already there. Leave a TODO?

illwieckz: I just added a `FIXME`

illwieckz:
I just added what was asked to be added and then rebased on current master. This looks ready to merge now.

slipher:
LGTM

PR #209, opened 2019-07-05
More work on initialization/shutdown routines

slipher:


illwieckz:
You can rebase on master to make travis happy.

In any way I have nothing to say against those changes.

illwieckz:
if no one else has something to say, it's ok to me.

PR #211, opened 2019-07-06
gl_shader: also checksum engine constants to invalidate the shader on cvar change

illwieckz:
That's an issue I'm facing since a long time, with problems appearing randomly… Last time I got it was when I was working on #193 and this time it was fully reproducible.

The cached binary shader is expected to be invalidated (then rebuilt) when the content of the shader source change. This is done by computing a checksum at shader compile time and comparing it lately.

Unfortunately, the checksum operation was not checksuming against the header that defines constants like `r_normalMapping`, only the content of the `.glsl` file and some macro. Because of that the shaders where not recompiled when such cvar were modified.

This fixes the issue by concatenating the constant header to the text that is sent to the checksum operation.


slipher:
LGTM

PR #214, opened 2019-07-22
make missing shader image a warning

illwieckz:
I was toying with some maps, I was surprised to see many textures missing without any warning telling me what was wrong. It's better to make this a warning.


slipher:
LGTM

PR #215, opened 2019-07-23
load image and sound file with dot in their extension-independent name

illwieckz:
At Quake 3 time it was common to add the .blend suffix to glow maps
so people made files named like:

- textures/something/somename.blend.jpg

If the shader use the full path, the engine will find the file even
if it's shipped in another format, because the engine will look for:

- textures/something/somename.blend.jpg
- textures/something/somename.blend.crn
- textures/something/somename.blend.png
- etc.

But the shader use the extension-independent path, the engine will
not find the file because the .blend suffix will be mistakenly
removed like if it was an extension, and the engine will look for:

- textures/something/somename.jpg
- textures/something/somename.crn
- textures/something/somename.png
- etc.

The engine may either fail to load the file, or load the wrong file
given the somename.blend.jpg file is usually a sidecar for an
existing somename.jpg file.

With this change, the engine tries to load the file by appending
common file extensions before stripping the extension and retrying
with common file extension on stripped name.

Tremulous mappers usually followed Quake 3 methodology, so there is
hundreds or thousands of maps that use .blend picture suffix out
there. It would have been possible to write a workaround for those
special files, but it's far better to make the engine able to fully
support files with dot in their names.

The change is applied to the sound loader as well.

### DIFF
--- a/src/engine/audio/SoundCodec.cpp
+++ b/src/engine/audio/SoundCodec.cpp
@@ -50,35 +50,16 @@ static const soundExtToLoaderMap_t soundLoaders[] =
 
 static int numSoundLoaders = ARRAY_LEN(soundLoaders);
 
-AudioData LoadSoundCodec(std::string filename)
+static int FindSoundLoader(std::string filename)
 {
+	const FS::PakInfo* bestPak = nullptr;
+	int bestLoader = -1;
 
-	std::string ext = FS::Path::Extension(filename);
-
-	// if filename has extension, try to load it
-	if (ext != "") {
-		// look for the correct loader and use it
-		for (int i = 0; i < numSoundLoaders; i++) {
-			if (ext == soundLoaders[i].ext) {
-				// if file exists, load it
-				if (FS::PakPath::FileExists(filename)) {
-					return soundLoaders[i].SoundLoader(filename);
-				}
-			}
-		}
-	}
-
-	// if filename does not have extension or there is no file with such extension
-	// or if there is no codec available for this file format,
 	// try and find a suitable match using all the sound file formats supported
 	// prioritize with the pak priority
-	int bestLoader = -1;
-	const FS::PakInfo* bestPak = nullptr;
-	std::string strippedname = FS::Path::StripExtension(filename);
-
 	for (int i = 0; i < numSoundLoaders; i++)
 	{
-		std::string altName = Str::Format("%s%s", strippedname, soundLoaders[i].ext);
+		std::string altName = Str::Format("%s%s", filename, soundLoaders[i].ext);
 		const FS::PakInfo* pak = FS::PakPath::LocateFile(altName);
 
 		// We found a file and its pak is better than the best pak we have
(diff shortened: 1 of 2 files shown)


src/engine/renderer/tr_image.cpp
```diff
@@ -1758,12 +1758,21 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
 		}
 	}
 
-	// the file isn't there, try again without the extension
-	COM_StripExtension3( token, filename, MAX_QPATH );
-
+	// if the file isn't there, maybe the file path did not have any extension
+	// but the file has a dot in its name that was mistakenly taken as an extension
```
slipher: This path is also used if the filename doesn't have a dot, right? So s/but/including when/ or something.

src/engine/renderer/tr_image.cpp
```diff
@@ -1758,12 +1758,21 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
...
 	const char *prefix;
 	int bestLoader = R_FindImageLoader( filename, &prefix );
 
+	if ( bestLoader == -1 )
```
slipher: Let's skip this block if `ext` is empty.

src/engine/renderer/tr_image.cpp
```diff
@@ -1758,12 +1758,21 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
...
+	{
+		// if the file isn't there, maybe its filename does not have extension
+		// or there is no file with such extension
+		// or there is no codec available for this file format
+		COM_StripExtension3( token, filename, MAX_QPATH );
```
slipher: Prefer sizeof(filename) instead of hard-coding the buffer size

src/engine/audio/SoundCodec.cpp
```diff
@@ -68,32 +93,23 @@ AudioData LoadSoundCodec(std::string filename)
...
 
-	// if filename does not have extension or there is no file with such extension
-	// or if there is no codec available for this file format,
-	// try and find a suitable match using all the sound file formats supported
-	// prioritize with the pak priority
-	int bestLoader = -1;
-	const FS::PakInfo* bestPak = nullptr;
-	std::string strippedname = FS::Path::StripExtension(filename);
+	// if the file isn't there, maybe the file path did not have any extension
+	// but the file has a dot in its name that was mistakenly taken as an extension
```
slipher: Same comment about "but" comment

src/engine/audio/SoundCodec.cpp
```diff
@@ -68,32 +93,23 @@ AudioData LoadSoundCodec(std::string filename)
...
+	int bestLoader = FindSoundLoader(filename);
 
-	for (int i = 0; i < numSoundLoaders; i++)
+	if (bestLoader == -1)
```
slipher: Same comment about skipping the extra loop in the no-dot case

src/engine/renderer/tr_image.cpp
illwieckz: This code already existed before, it's just moved to a new place.
Isn't it better to postpone such change to another PR?

To be honest I first started to rewrite the code to use std::string like the SoundCodec does, but then I dropped my work in progress because it would not be related to the initial purpose of the PR.

illwieckz:
I fixed stuff almost as requested :smirk:

src/engine/audio/SoundCodec.cpp
```diff
@@ -68,32 +93,24 @@ AudioData LoadSoundCodec(std::string filename)
...
+	// and it may be possible the file has a dot in its name that was mistakenly
+	// taken as an extension
+	int bestLoader = FindSoundLoader(filename);
 
-	for (int i = 0; i < numSoundLoaders; i++)
+	if (!ext.empty() && bestLoader == -1)
 	{
-		std::string altName = Str::Format("%s%s", strippedname, soundLoaders[i].ext);
-		const FS::PakInfo* pak = FS::PakPath::LocateFile(altName);
+		// if the file isn't there, maybe its filename does not have extension
```
slipher: This comment seems out of place since this block can only work if the filename does have an extension.

src/engine/renderer/tr_image.cpp
```diff
@@ -1758,12 +1758,22 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
...
-
+	// if the file isn't there, maybe the file path did not have any extension,
+	// and it may be possible the file has a dot in its name that was mistakenly
+	// taken as an extension
 	const char *prefix;
 	int bestLoader = R_FindImageLoader( filename, &prefix );
 
+	if ( *ext && bestLoader == -1 )
+	{
+		// if the file isn't there, maybe its filename does not have extension
```
slipher: Same here

illwieckz: It seems out of place because the way the code works.

In the case of `texture/name.blend.tga` named internally `texture/name.blend`, the `ext` string contains `.blend`, but the name itself does not have extension. Maybe I can add explicit words telling that in this particular case, `ext` is not empty but is not an extension, it's part of the name.

slipher: In that case (`texture/name.blend` requested when actual file is `texture/name.blend.tga`), it will be found by the first `R_FindImageLoader` call above.

illwieckz: Hmm, I have to sleep, I will re-read that later :grin:

illwieckz: Well, my mind was probably not very clear at this time, I was talking as if it was about the previous line, yes. I just fixed the comment.

PR #217, opened 2019-07-28
Remove disconnected download

slipher:


src/engine/client/cl_download.cpp
```diff
@@ -294,44 +281,26 @@ void CL_WWWDownload()
...
-				strcat( clc.redirectedList, "@" );
-				strcat( clc.redirectedList, cls.originalDownloadName );
-			}
-		}
-	}
-	else
-	{
-		if ( cls.bWWWDlDisconnected )
+		// tracking potential web redirects leading us to wrong checksum - only works in connected mode
+		if ( strlen( clc.redirectedList ) + strlen( cls.originalDownloadName ) + 1 >= sizeof( clc.redirectedList ) )
```
illwieckz: this makes me discovering there is code to attempt to handle redirects… because of course redirects don't work! :smirk:

slipher: There is not really any code to handle redirects. It's just a list of downloaded files.

PR #222, opened 2019-08-12
HTTP download enhancements

slipher:
Refactor the HTTP download code with better handling of libcurl errors.
Introduce the PAKSERVER file security feature (see the comments in dl_main.cpp for explanation).

Reviewers may prefer to look at the combined (as opposed to per-commit) diffs for dl_main.cpp as I moved some stuff around multiple times.

Before merging, we should get everyone running a server with HTTP downloads enabled to add the PAKSERVER file.

### DIFF
--- a/src/engine/client/dl_main.cpp
+++ b/src/engine/client/dl_main.cpp
@@ -81,12 +81,36 @@ void DL_InitDownload()
 	}
 
 	/* Make sure curl has initialized, so the cleanup doesn't get confused */
-	curl_global_init( CURL_GLOBAL_ALL );
+	if ( curl_global_init( CURL_GLOBAL_ALL ) == CURLE_OK && (dl_multi = curl_multi_init()) != nullptr )
+	{
+		downloadLogger.Debug( "Client download subsystem initialized" );
+		dl_initialized = 1;
+	}
+	else
+	{
+		downloadLogger.Warn( "Error initializing libcurl" );
+	}
+}
 
-	dl_multi = curl_multi_init();
+static void DL_StopDownload()
+{
+	if ( !dl_initialized )
+	{
+		return;
+	}
 
-	downloadLogger.Debug( "Client download subsystem initialized" );
-	dl_initialized = 1;
+	if ( dl_request )
+	{
+		curl_multi_remove_handle( dl_multi, dl_request );
+		curl_easy_cleanup( dl_request );
+		dl_request = nullptr;
+	}
+
+	if ( dl_file )
+	{
+		FS_FCloseFile( dl_file );
+		dl_file = 0;
+	}
 }
 
 /*
@@ -102,8 +126,7 @@ void DL_Shutdown()
 		return;
 	}
 
-	curl_multi_cleanup( dl_multi );
-	dl_multi = nullptr;
+	DL_StopDownload();
 
 	curl_global_cleanup();
 
@@ -120,18 +143,7 @@ int DL_BeginDownload( const char *localName, const char *remoteName )
 {
 	char referer[ MAX_STRING_CHARS + URI_SCHEME_LENGTH ];
 
-	if ( dl_request )
-	{
-		curl_multi_remove_handle( dl_multi, dl_request );
-		curl_easy_cleanup( dl_request );
-		dl_request = nullptr;
-	}
-
-	if ( dl_file )
-	{
-		FS_FCloseFile( dl_file );
-		dl_file = 0;
-	}
+	DL_StopDownload();
 
 	if ( !localName || !remoteName )
 	{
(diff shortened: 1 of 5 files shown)


src/engine/client/dl_main.cpp
```diff
@@ -37,40 +37,207 @@ Maryland 20850 USA.
...
+		return unmatchedPrefix_.empty() ? dlStatus_t::DL_DONE : dlStatus_t::DL_CONTINUE;
+	}
+
+public:
+	PakserverCheck(Str::StringRef url) : CurlDownload(url) {}
+};
+
+struct DownloadState {
+	Util::optional<PakserverCheck> pakserverCheck;
+	Util::optional<FileDownload> realDownload;
```
DolceTriade: as opposed to a fake download?

src/engine/client/dl_main.cpp
```diff
@@ -194,13 +194,48 @@ class FileDownload : public CurlDownload {
...
+// before an HTTP download commences, the client checks for a magic file named PAKSERVER in the same
+// directory where the desired file is stored. If the file begins with the right magic string (we
+// check only a prefix to avoid any fiddly newline issues), this is interpreted as permission
+// Daemon clients to download anything from that directory.
+//
+// Note that this is the same issue addressed by the Same-Origin Policy in web browsers. In that
+// model, CORS headers are used to give scripts permission to access resources. We couldn't really
+// implement that here since there is no origin tracking of cgame binaries.
+const Str::StringRef PAKSERVER_FILE_NAME = "PAKSERVER";
+const Str::StringRef PAKSERVER_FILE_CONTENT_PREFIX = "This server is configured to host Daemon engine packages";
```
illwieckz: While I'm 100% in favor of such feature, I really don't like this kind of sentence.

I think it would be better to stick to one case and to avoid whitespace (a bit like [this one before the `(C)` sutff](https://github.com/kholia/OSX-KVM/blob/fb9d9094d74a6283c8c18f94d67b3012af929655/macOS-libvirt.xml#L169)).

Also, I'm not sure that's needed to reference the Dæmon engine name. Like the DPK format, this kind of mechanism can be implemented by other engines. We also don't need such long string. The magic number can be far shorter. All we need is to reduce the probability that a file with the required name and the required content can appear unintentionally. Something like `GRANTED` sounds enough to me.

slipher: I want the magic string to be self-documenting. As such, spaces make it easier to read and I don't see the advantage to removing them. Upper/lower case, no big deal to me. I'd be open to changing it to something that doesn't include "Daemon".

Uppercasing is one possible convention for magic filenames, to make them look distinct from the other files.

src/engine/client/dl_main.cpp
slipher: `PakserverCheck` is also a subclass of `CurlDownload`, so, I guess.

illwieckz:
This is fixed, http://dl.unvanquished.net redirects to https, but http://dl.unvanquished.net is available on both http and https.

src/engine/client/dl_main.cpp
illwieckz: I would prefer to leverage [RFC 5785](https://tools.ietf.org/html/rfc5785) (See for example: [eff.org/.well-known/dnt-policy.txt](https://www.eff.org/.well-known/dnt-policy.txt)) and do something like:

```
.well-known/pak-server.txt
```

It requires an extra step (creating a directory), but that would not reinvent the wheel.

Also, we can imagine a game server provider to automatically add such path as an alias (like most do with letsencrypt's `.well-known/acme-challenge` directory) without requiring customers to handle it themselves.

I prefer a simple word because of the tutorial part, the shortest and the less error prone it is, the better it is, we could tell people to do that:

```sh
mkdir -p .well-known
echo granted > .well-known/pak-server.txt
```

I worry about the uppercase because I'm already annoyed by the uppercase of `DEPS` in dpk files. When we switch from PK3 to DPK extension we had opportunity to rename it (even put it in a folder so we can easily add feature to the format), but I did not see the need for it. Now that I wrote a tool that automatically port tremulous maps by taking care of all issues including casing by lowercasing everything… I had to write a special rule in my tool to exclude README.md and… DEPS, a file we had control over. So, by default I'm a bit unpleased about uppercase files. In any way if we go the `.well-known` way it would be inconsistent to create an uppercase file name in a lowercase directory. I have nothing against uppercase content in anyway, but it's just require extra typing for nothing then it's more error prone, and there would be people to ask if it's required to be uppercase or not.

illwieckz: Hmmm, `.well-known` mechanism is expected to be on `/`, not in a subdirectory, and we can serve paks in web servers subdirectories. Maybe it's a bad idea to do `.well-known` stuff the way `.well-known` stuff was never meant…

So to sum it up, my feeling is:

- it's not bad if it's a dot-hidden file (I have a strong feeling for this)
- it's not bad if there is only one string without space, even better one word (I have strong feeling for this)
- it's not bad if the case is consistent in name and in file (I have strong feeling for this)
- it's not bad if it's lower case (but I would not enforce it)

src/engine/client/dl_main.cpp
illwieckz: `actualDownload`? this seems to less induce the opposite would be "fake" or "wrong"

src/engine/client/dl_main.cpp
slipher: I have some concerns about using a dot name. If there are some HTTP servers that don't serve dot files by default, that could cause headaches. Also I prefer it to show up in the HTTP server's directory listing, so that someone who is setting up a server can easily imitate a working example.

I can see the point in shortening the file contents, but one word is too vague and ungoogleable. How about `ALLOW_UNRESTRICTED_DOWNLOAD` or some case/punctuation variant thereof?

Also, I have realized an annoying flaw in the current version of the PR, which is that if you have paks in subdirectories, you'd have to create a PAKSERVER file in all subdirectories (I didn't realize it's possible to have `/` in pak names when I wrote the code). So now I'm thinking of changing the download protocol to send a base URL plus a relative one, rather than a single URL. That way PAKSERVER only has to be in the base directory.

illwieckz: > How about `ALLOW_UNRESTRICTED_DOWNLOAD` or some case/punctuation variant thereof?

I'm ok with `ALLOW_UNRESTRICTED_DOWNLOAD`, I think it's better to use _same-case-plus-separator_ than mixing case.

At this point we may say _"why not `ALLOW UNRESTRICTED DOWNLOAD` with spaces?"_, but I know that some keyboard layout put a non-breaking space character on [shift+space] (which is crazy, we all agree), that's why I'm not very happy with spaces between upper case letters.

> Also, I have realized an annoying flaw in the current version of the PR, which is that if you have paks in subdirectories, you'd have to create a PAKSERVER file in all subdirectories (I didn't realize it's possible to have / in pak names when I wrote the code). So now I'm thinking of changing the download protocol to send a base URL plus a relative one, rather than a single URL. That way PAKSERVER only has to be in the base directory.

If I'm right, it's already _“base directory plus relative pak path”_ in engine, so it's better to keep that behavior all the way down.

src/engine/client/dl_main.cpp
slipher: Done.

slipher:
Removed the commit changing the default download address.

illwieckz:
I'm not sure you pushed your latest branch

slipher:
> I'm not sure you pushed your latest branch

Fixed. Anyway, what do you think about the potential issues with dot files I mentioned above?

src/engine/client/dl_main.cpp
illwieckz: > If there are some HTTP servers that don't serve dot files by default, that could cause headaches. Also I prefer it to show up in the HTTP server's directory listing, so that someone who is setting up a server can easily imitate a working example.

Hmm, right, that's very strong. I just tested default apache and nginx default listing behavior and they both hide hidden dot files. I truly agree the behavior must be explicit and obvious, and for example people must be able to download the `PAKSERVER` file themselves to do it by imitation. So, let's go for `PAKSERVER` without dot and with caps.

Since `PAKSERVER` name is explicit, I still think the 'ENABLED' content would be enough. The `PAKSERVER` `ENABLED` combination looks explicit enough and self-explaining to me.

illwieckz:
> If there are some HTTP servers that don't serve dot files by default, that could cause headaches. Also I prefer it to show up in the HTTP server's directory listing, so that someone who is setting up a server can easily imitate a working example.

Hmm, right, that's very strong. I just tested default apache and nginx default listing behavior and they both hide hidden dot files. I truly agree the behavior must be explicit and obvious, and for example people must be able to download the `PAKSERVER` file themselves to do it by imitation. So, let's go for `PAKSERVER` without dot and with caps.

Since `PAKSERVER` name is explicit, I still think the 'ENABLED' content would be enough. The `PAKSERVER` `ENABLED` combination looks explicit enough and self-explaining to me.

slipher:
Apparently there is a hosting company called Pakserver. So "pakserver enabled" might not be enough to find relevant information in a search engine...

illwieckz:
Well, it's not like if our engine was named “Daemon” :grin: 

I also thought about DPKSERVER, which is not bad as it is coherent with the whole DPK strategy. I'm not sure it have to be DPK centric in any way, the engine itself being compatible with legacy PK3 and other project may mimic us to avoid the same kind of issue.

I also thought about something: our game does not fetch game content from website but it may be possible in the future to do it (the way War§ow does with blog posts), in that case it would make sense to rely on the same PAKSERVER mechanism, the PAKSERVER would not be very meaningful, but it's probably not a big issue. What is _important_ is that it's meaningful for paks.

And yes, of course PAKistan has SERVERs.

Instead of the « server » meaning, why not something like _directory_, _registry_, _library_, _repository_, collection… in face we don't care about the server, what is important is what is served. The directory containing paks is not a server itself, it's a collection. We don't flag a server (or we would better use the `.well-known` mechanism at server root), we flag a directory, there may be multiple directories on the same server. The “directory” word is not the good one because of the existing pk3dir/dpkdir which has another meaning. But after some thought I think it would be better to use a name about the repository itself than a name about the server. What do you think about it?

illwieckz:
_PAKDEPOT_ ? (we can even explore the « store » meaning, like ine « appstore », there is a lot of words available, _storage_, _cellar_, _vault_)

Of course there is already existing _pakdepot_ entries in search engine results…

slipher:
OK, I changed the message to `ALLOW_UNRESTRICTED_DOWNLOAD` and left the filename as `PAKSERVER`. It does not need to be a totally unique word as long as the content is. Is that all right @illwieckz ?

illwieckz:
If I was doing it myself I would just have put one word but `ALLOW_UNRESTRICTED_DOWNLOAD` is very ok and I don't want to looks like a dictator :grin:

…and this PR remained open for enough time :smirk:

So let's go with `PAKSERVER` file name with content starting with `ALLOW_UNRESTRICTED_DOWNLOAD` :thumbsup: :ok_hand:

PR #225, opened 2019-11-20
Function to get or set console keys

slipher:
There are other parts to this which will need to go into the for-0.52 branch, the Unvanquished repository, and the unvanquished_src.dpkdir repository.


illwieckz:
seems good

Issue #190, opened 2019-03-24
glsl ifdef generator does not set define anytime


illwieckz:
See #189 for a branch that reproduces the bug (you need Xonotic assets since Unvanquished ones run the code used when the define is not defined).

It looks like the define is not defined any time as while I move in map, the wrong code is sometime run instead of the expected one:

[![glsl ifdef](https://dl.illwieckz.net/b/daemon/bugs/glsl-ifdef/unvanquished_2019-03-23_203425_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/glsl-ifdef/unvanquished_2019-03-23_203425_000.jpg)

[![glsl ifdef](https://dl.illwieckz.net/b/daemon/bugs/glsl-ifdef/unvanquished_2019-03-23_203429_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/glsl-ifdef/unvanquished_2019-03-23_203429_000.jpg)

The bug is on the wall texture, not on the blooming reactor.

I understand nothing from the define generator code and I need help to fix it, or someone who understands it to fix it himself.

illwieckz:
If I understand it right, the code computes a program index for each shader variation, so it can switch to those program per texture within one frame. I guess the issue is that that index computation gets some collision.

illwieckz:
[OpenGL swizzle mask](https://www.khronos.org/opengl/wiki/Texture#Swizzle_mask) may allows us to not require the usage of defines to discriminate normalmap shipping heightmap or from one not shipping height map, it would also allow use to add support for loose heightmap without requiring a complicated glsl code.

Basically, I think about this:

```
if normalmap:
	if heightmap in normalmap:
		bind normalmap.rgba as normalmap.rgb0
		bind normalmap.a as heightmap.rgba
	else
		bind normalmap.rgba as normalmap.rgba
		if heightmap:
			bind heightmap.rgba as heightmap.rgba
elif heightmap:
	bind dummynormalmap as normalmap.rgba
	bind heightmap.rgba as heightmap.rgba
```

That would require OpenGL 3.3. The game would still work for older hardware without OpenGL 3.3 support but with normalmap and heightmap support being disabled for normalmap files shipping heightmap.

illwieckz:
I'm not sure there is a bug in that ifdef code, similar artifact can happen with zero normalscale (hence undefined normalscale). I noticed some issues with that other feature, so maybe it's the same. Closing it for the moment.

PR #230, opened 2019-12-07
implement pre-collapsed shader stage, fix #216

### LINKED ISSUE #216: [rant] XreaL material syntax is broken by design
Recently I had some talk with @xycaleth from @JAcoders, and he taughth me that ioq3 and OpenJK relies on `normalFormat` keyword to handle channel flipping (for example: flipping the Y channel to convert from OpenGL format to DirectX format). Basically you can do `normalFormat 2 -2 2` to double all the channels and revert the Y one. This is clever.

Previously I implemented a `normalFormat` keyword in Dæmon with purpose to flip normal map channels with. At the same time, I discovered that normal map scaling using shader keywords was not implemented (only a cvar existed, with only one value for both X and Y).

So I thought about implementing `normalScale` and ditch `normalFormat`. I implemented normalScale on stage level, because it's where it is expected to be done. Then I added the `normalFormat -1 1 1` line on a vega shader that is known to have a wrong normal map with flipped X channel. Then I wanted to test my code and NOOOOOOOOOOO. Please, XreaL, get out of my way! %@#§ø\%!!!

The edited shader was looking like that:

```c
textures/shared_vega/panel01
{
	qer_editorImage textures/shared_vega_src/panel01_d

	diffuseMap  textures/shared_vega_src/panel01_d
	normalMap   textures/shared_vega_src/panel01_n
	specularmap textures/shared_vega_src/panel01_s

	normalScale -1 1 1
}
```

But that can't work because the keyword is not in a stage, but it would be very ugly to implement the keyword on a shader level instead! :'(

So, what do we know about that %@#§ø\% Xreal shader syntax?

We know that all those `diffuseMap`, `normalMap` lines are shortcuts for stages, stages that are then collapsed with some code that is hard to edit without leaving it half broken.

So I would be able to do that:

```c
textures/shared_vega/panel01
{
	qer_editorImage textures/shared_vega_src/panel01_d

	diffuseMap  textures/shared_vega_src/panel01_d
	{
		stage       normalMap
		map         textures/shared_vega_src/panel01_n
		normalScale	-1 1 1
	}
	specularmap textures/shared_vega_src/panel01_s
}
```

But that's ugly.

There is another way to do shaders, and that would magically solve the need for a collapsing code, and would solve my need for the normalScale keyword, and that would make Dæmon able to do terrain blending with normal and specular maps one day!

Because we can do that today:

```c
{
	q3map_nonplanar
	q3map_shadeAngle 90

	{
		map textures/shared_pk02_src/rock01_d
		rgbGen identity
	}
	{
		map textures/shared_pk02_src/sand01_d
		alphaFunc GE128
		rgbGen identity
		alphaGen vertex
	}
	{
		map $lightmap
		rgbGen identity
	}
}
```

But we can only blend colormaps, there is no way to add normalMap since the XreaL-inherited code expects the normalMap to be a stage that would be collapsed by the engine, and there is no way to tell the engine to collapse the normal map to one of the colormaps used in that terrain blending keyword.

What I suggest that instead of doing that:

```c
textures/shared_pk01/door01a_norad
{
	qer_editorImage     textures/shared_pk01_src/door01a_d

	diffuseMap          textures/shared_pk01_src/door01a_d
	normalMap           textures/shared_pk01_src/door01_n
	specularMap         textures/shared_pk01_src/door01_s
	glowMap             textures/shared_pk01_src/door01_a
}
```

We do that:

```c
textures/shared_pk01/door01a_norad
{
	qer_editorImage     textures/shared_pk01_src/door01a_d
	{
		diffuseMap          textures/shared_pk01_src/door01a_d
		normalMap           textures/shared_pk01_src/door01_n
		specularMap         textures/shared_pk01_src/door01_s
		glowMap             textures/shared_pk01_src/door01_a
	}
}
```

That would allow us to do that:

```c
textures/shared_vega/panel01
{
	qer_editorImage textures/shared_vega_src/panel01_d

	{
		diffuseMap  textures/shared_vega_src/panel01_d
		normalMap   textures/shared_vega_src/panel01_n
		specularmap textures/shared_vega_src/panel01_s
		normalScale -1 1 1
	}
}
```

As soon as we don't leave the stage, the engine would collapse stages without having to know clever stuff.

And we would be able to do that:

```c
{
	q3map_nonplanar
	q3map_shadeAngle 90

	{
		diffuseMap  textures/shared_pk02_src/rock01_d
		normalMap   textures/shared_pk02_src/rock01_n
		specularMap textures/shared_pk02_src/rock01_d
		rgbGen identity
		nolightmap
	}
	{
		diffuseMap  textures/shared_pk02_src/sand01_d
		normalMap   textures/shared_pk02_src/sand01_n
		alphaFunc GE128
		rgbGen identity
		alphaGen vertex
		nolightmap
	}
	{
		map $lightmap
		rgbGen identity
	}
}
```

So, yes, we would have to convert our existing shaders, but it's just a matter of adding one `{` and one `}`. I'm very OK to do it.

We would have to edit Sloth to add those `{`, `}` too. It may be not so hard to do.

In the end the code will be cleaner and features would be easier to implement. We would also be able to deduplicate all the `if keyword is diffuseMap or normaMap or…` code that lives both in shader parser and stage parser, to only keep the stage parser.

%@#§ø\%!!!

illwieckz:
This is my attempt to implement pre-collapsed shader stages, as defined in #216.
This will allow to use more than one collapsed (multi-texture) stage per shader.

Historical syntax (Doom3-like) with collapsing heuristic is still there.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1089,7 +1089,7 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 		bool        tcGen_Environment;
 		bool        tcGen_Lightmap;
 
-		bool            disableImplicitLightmap;
+		bool implicitLightmap;
 
 		Color::Color32Bit constantColor; // for CGEN_CONST and AGEN_CONST
 
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1222,8 +1222,6 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 		float          portalRange; // distance to fog out at
 		bool       isPortal;
 
-		int            collapseTextureEnv; // 0, GL_MODULATE, GL_ADD (FIXME: put in stage)
-
 		cullType_t     cullType; // CT_FRONT_SIDED, CT_BACK_SIDED, or CT_TWO_SIDED
 		bool       polygonOffset; // set for decals and other items that must be offset
 		float          polygonOffsetValue;
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1004,6 +1004,7 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	{
 	  TB_COLORMAP,
 	  TB_DIFFUSEMAP = TB_COLORMAP,
+	  TB_REFLECTIONMAP = TB_COLORMAP,
 	  TB_NORMALMAP,
 	  TB_SPECULARMAP,
 	  TB_MATERIALMAP = TB_SPECULARMAP,
@@ -1054,6 +1055,7 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	enum class collapseType_t
 	{
 	  COLLAPSE_none,
+	  COLLAPSE_generic, // used before we know it's another one
 	  COLLAPSE_lighting_PHONG,
 	  COLLAPSE_lighting_PBR,
 	  COLLAPSE_reflection_CB,
@@ -1062,6 +1064,8 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	struct shaderStage_t
 	{
 		stageType_t     type;
+
+		collapseType_t collapseType;
 
 		bool        active;
 
@@ -1218,8 +1222,6 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 		float          portalRange; // distance to fog out at
 		bool       isPortal;
 
-		collapseType_t lightingCollapseType;
-		collapseType_t reflectCollapseType;
 		int            collapseTextureEnv; // 0, GL_MODULATE, GL_ADD (FIXME: put in stage)
 
 		cullType_t     cullType; // CT_FRONT_SIDED, CT_BACK_SIDED, or CT_TWO_SIDED
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -4025,11 +4025,23 @@ static bool ParseShader( const char *_text )
 			continue;
 		}
 		// reflectionMapBlended <image>
-		else if ( !Q_stricmp( token, "reflectionMapBlended" )
-			|| ( *r_dpMaterial && !Q_stricmp( token, "dpreflectcube" ) ) )
+		else if ( !Q_stricmp( token, "reflectionMapBlended" ) )
 		{
 			ParseReflectionMapBlended( &stages[ s ], text );
 			s++;
+			continue;
+		}
+		else if ( !Q_stricmp( token, "dpreflectcube" ) )
+		{
+			if ( *r_dpMaterial )
+			{
+				ParseReflectionMapBlended( &stages[ s ], text, TB_COLORMAP );
+				s++;
+			}
+			else
+			{
+				Log::Warn("disabled DarkPlaces shader parameter '%s' in '%s'", token, shader.name );
+			}
 			continue;
 		}
 		// lightFalloffImage <image>
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -1364,7 +1364,7 @@ static bool ParseMap( const char **text, char *buffer, int bufferSize )
 	return true;
 }
 
-static bool LoadMap( shaderStage_t *stage, const char *buffer )
+static bool LoadMap( shaderStage_t *stage, const char *buffer, const int bundleIndex = TB_COLORMAP )
 {
 	char         *token;
 	int          imageBits = 0;
@@ -1382,19 +1382,19 @@ static bool LoadMap( shaderStage_t *stage, const char *buffer )
 	if ( !Q_stricmp( token, "$whiteimage" ) || !Q_stricmp( token, "$white" ) || !Q_stricmp( token, "_white" ) ||
 	     !Q_stricmp( token, "*white" ) )
 	{
-		stage->bundle[ 0 ].image[ 0 ] = tr.whiteImage;
+		stage->bundle[ bundleIndex ].image[ 0 ] = tr.whiteImage;
 		return true;
 	}
 	else if ( !Q_stricmp( token, "$blackimage" ) || !Q_stricmp( token, "$black" ) || !Q_stricmp( token, "_black" ) ||
 	          !Q_stricmp( token, "*black" ) )
 	{
-		stage->bundle[ 0 ].image[ 0 ] = tr.blackImage;
+		stage->bundle[ bundleIndex ].image[ 0 ] = tr.blackImage;
 		return true;
 	}
 	else if ( !Q_stricmp( token, "$flatimage" ) || !Q_stricmp( token, "$flat" ) || !Q_stricmp( token, "_flat" ) ||
 	          !Q_stricmp( token, "*flat" ) )
 	{
-		stage->bundle[ 0 ].image[ 0 ] = tr.flatImage;
+		stage->bundle[ bundleIndex ].image[ 0 ] = tr.flatImage;
 		return true;
 	}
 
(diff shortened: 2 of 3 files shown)


illwieckz:
that looks ready to me

illwieckz:
I removed the `colorMap` stage keyword as it is useless and seems to have no reason to be used. In the engine, a _colormap_ is a simple `map` without anything else (no lightmap, no normal…).

illwieckz:
I merged into master the very first commit that is not doing anything on the code itself, except moving some text to make subsequent PR less noisy.

src/engine/renderer/tr_local.h
```diff
@@ -1066,6 +1066,8 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 		stageType_t     type;
 
 		collapseType_t collapseType;
+		// TODO: is it used?
+		int            collapseTextureEnv; // 0, GL_MODULATE, GL_ADD
```
slipher: Why would you add an unused field?

illwieckz: I just moved it from shader struct to stage struct with other collapse stuff. Only the comment “is it used?” was new at this point.

I then deleted that variable with a commit deleting leftovers when I was sure it was not used.

illwieckz: oh, right, this is the commit that was supposed to delete it, something bad probably happened on some rebase step.

illwieckz: fixed.

illwieckz:
Note to myself: I said there: https://github.com/DaemonEngine/Daemon/pull/189#issuecomment-485199306

I had an implementation to collapse both lighting stages and reflection stages with an unified glsl shader (I hope I still have the branch somewhere :grin: ) but this is blocked because of #190.

Once a fix is found for #190, many lines would be deletable.

src/engine/renderer/tr_shader.cpp
```diff
@@ -4327,13 +4493,14 @@ static void CollapseStages()
...
 				// merge with diffuse stage
-				stages[ diffuseStage ].bundle[ TB_MATERIALMAP ] = stages[ materialStage ].bundle[ 0 ];
+				stages[ diffuseStage ].bundle[ TB_MATERIALMAP ] = stages[ materialStage ].bundle[ TB_COLORMAP ];
 				// disable since it's merged
 				stages[ materialStage ].active = false;
 			}
 			// always test for this stage before glow stage
 			if ( lightStage != -1 )
 			{
+				// “blendFunc filter” is same as “blendFunc GL_DST_COLOR GL_ZERO” and the default
```
slipher: duplicated line

src/engine/renderer/tr_shader.cpp
```diff
@@ -3985,49 +4116,56 @@ static bool ParseShader( const char *_text )
...
+			Log::Warn("deprecated standalone stage '%s' in shader '%s'", token, shader.name );
+			ParseNormalMap( &stages[ s ], text, TB_COLORMAP );
 			s++;
 			continue;
 		}
 		// specularMap <image>
 		else if ( !Q_stricmp( token, "specularMap" ) )
 		{
-			ParseSpecularMap( &stages[ s ], text );
+			Log::Warn("deprecated standalone stage '%s' in shader '%s'", token, shader.name );
```
slipher: Could you put a link or some kind of clue for users to learn what the alternative syntax is?

src/engine/renderer/tr_shader.cpp
```diff
@@ -5466,19 +5633,34 @@ void R_ShaderList_f()
...
+		for ( int i = 0; i < MAX_SHADER_STAGES; i++ )
 		{
-			str += "reflection_CB  ";
+			if ( shader->stages[ i ]->active )
+			{
+				if ( shader->stages[ i ]->collapseType == collapseType_t::COLLAPSE_lighting_PBR )
+				{
+					str += "lighting_PBR   ";
+					foundCollapseType = true;
+					continue;
```
slipher: Looks like you wanted `break`

src/engine/renderer/tr_shader.cpp
```diff
@@ -1717,7 +1718,7 @@ static void ParseLightFalloffImage( shaderStage_t *stage, const char **text )
 	char buffer[ 1024 ] = "";
 
 	stage->active = true;
-	stage->type = stageType_t::ST_ATTENUATIONMAP_Z;
+	stages->type = stageType_t::ST_ATTENUATIONMAP_Z;
```
slipher: This looks suspicious

src/engine/renderer/tr_shader.cpp
```diff
@@ -1733,15 +1734,17 @@ struct extraMapParser_t
...
 };
 
 static const extraMapParser_t extraMapParsers[] =
 {
-	{ "norm",    "normal map",     ParseNormalMap     },
-	{ "gloss",   "specular map",   ParseSpecularMap   },
-	{ "glow",    "glow map",       ParseGlowMap       },
-	{ "luma",    "glow map",       ParseGlowMap       },
+	// currently uses the collapse code path
+	{ "norm",    "normal map",     ParseNormalMap,     TB_COLORMAP, },
```
slipher: In what circumstance would you want to use a different value besides TB_COLORMAP?

src/engine/renderer/tr_shader.cpp
illwieckz: right

src/engine/renderer/tr_shader.cpp
illwieckz: When you do this:

```
textures/collection/name
{
	qer_editorImage textures/collection/name_d
	{
		diffuseMap textures/collection/name_d
		glowMap textures/collection/name_g
	}
}
```

The engine create a new stage with `ST_COLLAPSED_lighting_PBR` type, and adds `rock_d` texture to `TB_DIFFUSEMAP` and `rock_g` texture to `TB_GLOWMAP`.
And stage is ready to feed the renderer.

When you do this:

```
textures/collection/name
{
	qer_editorImage textures/collection/name_d

	diffuseMap textures/collection/name_d
	glowMap textures/collection/name_g
}
```

or this:

```
textures/collection/name
{
	qer_editorImage textures/collection/name_d
	{
		map textures/collection/name_d
		stage diffuseMap
	}
	{
		map textures/collection/name_g
		stage glowMap
	}
}
```

There is two stages created: one with `ST_DIFFUSEMAP` type with `rock_d` in `TB_COLORMAP`, and one with `ST_NORMLALMAP` type with `rock_g` in `TB_COLORMAP`.

Then a collapsing heuristing creates a new stage with  `ST_COLLAPSED_lighting_PBR` type and stores `rock_d` texture to `TB_DIFFUSEMAP` and `rock_g` texture to `TB_GLOWMAP`, initial stages are disabled.

We can't store `rock_d` and `rock_g` in `TB_DIFFUSEMAP` or `TB_GLOWMAP` in standalone stages in case we can't collapse.

If for some if we can't collapse, the glow map will be loaded as a color map with a blend function, like if it was initially written this way:

```
textures/collection/name
{
	qer_editorImage textures/collection/name_d

	diffuseMap textures/collection/name_d
	{
		map textures/collection/name_g
		blendfunc add
	}
}
```

We have to keep in mind we are attempting to collapse colormaps that has known behaviour to optimize the renderer, but they can all be rendered as separate colormap stages with proper blend function. If we fail to collapse, there must be good old standard colormap stages with their respective blend functions.

src/engine/renderer/tr_shader.cpp
illwieckz: right

src/engine/renderer/tr_shader.cpp
illwieckz: I'll do

illwieckz: I mean, clue in code or in warning text?

slipher: warning text

illwieckz:
@slipher: resolved remarks, rebased on master

slipher:
Code seems all right to me. I don't understand anything about shaders, of course.

illwieckz:
> I don't understand anything about shaders, of course.

Which is unfotunately to expect: that code is probably the most obscure part of dæmon, at least the most obscure I've seen.

Basically XreaL project hacked shader parser of Quake 3 to parse Doom3 shaders the quick and dirty way, expecting multi textures as shader globals (what does Doom 3) while the Quake 3 design allows advanced stage blending and operation. While those textures parts (normal/spec/glow) can perfectly run as separate stages, for performance reason it's better to collapse everything we can to render as must things we can in one pass, hence the convoluted collapse code that has to deal with the fact there is way to define textures both in shader stages and shader globals… :fearful:

To add features it's usually required to redesign from scratch some parts that do not make possible to extend the existing code to add such features. So, hopefully, PR after PR the code becomes less complex.

PR #231, opened 2019-12-08
implement per-stage normalscale

illwieckz:
This PR sits above #230.
This is a demonstration of what ca be achieved thanks to #230.

Previously in #180 I introduced a per-material `normalFormat` option that would be able to flip channels. But then I discovered that other engines relies on a `normalScale` option that can both scale (and flip) channels.

A proper implementation required to be done on the stage scope.

Note that ioq3 seems to only read `normalScale X Y` while we are able to read `normalScale X Y Z`, if Z is missing, default one is used instead, to keep compatibility with material that may already exists.

As an example, we can look at some materials from [tex-vega_src.dpkdir](https://github.com/UnvanquishedAssets/tex-vega_src.dpkdir) that have a flipped X channel.

Before:

[bad normal map](https://dl.illwieckz.net/b/daemon/features/normalscale/unvanquished_2019-12-08_030625_000.jpg)

After:

[fixed normal map](https://dl.illwieckz.net/b/daemon/features/normalscale/unvanquished_2019-12-08_030828_000.jpg)

They can be easily compared this way:

[before/after fixed normal map](https://dl.illwieckz.net/b/daemon/features/normalscale/unvanquished_2019-12-08_030828_001.jpg)

or with a slider there: https://imgsli.com/OTI4Mw

To do fix them, their materials were edited from this:

```c
textures/shared_vega/panel01
{
	qer_editorImage textures/shared_vega_src/panel01_d

	diffuseMap  textures/shared_vega_src/panel01_d
	normalMap   textures/shared_vega_src/panel01_n
	specularMap textures/shared_vega_src/panel01_s
}

textures/shared_vega/trim03
{
	qer_editorImage textures/shared_vega_src/trim03_d

	diffuseMap  textures/shared_vega_src/trim03_d
	normalMap   textures/shared_vega_src/trim03_n
	specularmap textures/shared_vega_src/trim03_s
}
```

to this:

```c
textures/shared_vega/panel01
{
	qer_editorImage textures/shared_vega_src/panel01_d

	{
		diffuseMap  textures/shared_vega_src/panel01_d
		normalMap   textures/shared_vega_src/panel01_n
		normalScale -1 1 1
		specularMap textures/shared_vega_src/panel01_s
	}
}

textures/shared_vega/trim03
{
	qer_editorImage textures/shared_vega_src/trim03_d

	{
		diffuseMap  textures/shared_vega_src/trim03_d
		normalMap   textures/shared_vega_src/trim03_n
		normalScale -1 1 1
		specularmap textures/shared_vega_src/trim03_s
	}
}
```

In the process we also moved from separate stages that has to be collapsed by the engine to a multi texture stage that is collapsed from the start.

----
Also note we inherited from XreaL another  `normalScale` keyword that is currently only used by broken liquid shader, and this keyword is expecting an expression. We have to rename it. It is now named normalIntensity, this applies on XY normal map channels and can be computed on every frame by using expressions.

It would have been better to keep the  `normalScale` keyword for that feature (as exists `fresnelScale` for example) and use `normalChannelScale` for the par-channel-scaling-and-flipping variable, but  `normalScale` is already widely used by other idtech3 derivated engines, while our liquid shader is used by no one yet, even not by us.


illwieckz:
This seems ready to me.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -26,37 +26,16 @@ uniform sampler2D	u_NormalMap;
 uniform int u_HeightMapInNormalMap;
 #endif // r_normalMapping || USE_PARALLAX_MAPPING
 
+#if defined(r_normalMapping)
+// HACK: fourth component tells renderer if normal scale is customized or not
```
slipher: Why not simply put in (1, 1, 1) in that case?

illwieckz: The true reliable test is to test of something is not zero, so everytime we allocate, we already know the default value is right. Otherwise, we have to check all the code to set the default value everytime we allocate, which is calling for future bugs in the future as things would be modified.

Note that patching this code is an iterative process. As second commit moves the test out of the glsl shader (doing it per-material instead of per-pixel), there is no need to hack a fourth component and a bool is used instead.

We maybe would be able to not use a bool at all by hacking the third channel (we know 0 is wrong) but that's something I'm not ready to support and it would add no performance improvement. The purpose of hacking a channel is to avoid to bind an additional variable from CPU to GPU. The second commit entirely remove the need for an additional variable so the bool costs nothing GPU side and costs no less CPU side.

illwieckz:
I added a commit to remove the `r_normalScale` glsl define, since we pass normalScale and compute it in all case, it maybe better to reuse it to feed r_normalScale value if exists.

illwieckz:
I noticed I forgot to make the dead liquid shader using `u_normalScale`, that's now fixed.
This is just to not let dead code rot until we revive it.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: Unfortunately it would be complicated to remove the fourth channel hack in first commit without squashing second commit. And I would prefer to keep the two commits separated as it makes better to understand the iterative process involved in rewriting that code. This is done step by step and sometime some code has to be written that would be removed by the next step, this to avoid huge commits that does everything at once.

illwieckz:
@slipher, I added a small commit I want your advice on.

This commit is just done to reduce boilerplate. But it replaces a definition by an affectation, do you think a compiler is able to optimize it, or, if not, this is negligible?

src/engine/renderer/tr_shade.cpp
```diff
@@ -598,6 +598,12 @@ void SetNormalScale( shaderStage_t *pStage, vec3_t normalScale)
 			normalScale[ 1 ] *= pStage->normalScale[ 1 ];
 			normalScale[ 2 ] *= pStage->normalScale[ 2 ];
 		}
+
+		// never multiply normal Z by zero, it's wrong
+		if ( r_normalScale->integer )
```
slipher: I think this should be a float, not integer.

 I'd drop the zero test also. I imagine the main purpose of this variable is debugging, to make the normals stand out more so you can check them. So 0 would also be a legitimate value, to test that things become flat as expected (although it wouldn't work now I guess, due to testing the z component in glsl).

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
slipher: Testing the z component in GLSL still has a small extra cost in terms of running extra instructions, compared to just setting 1,1,1. The condition has to be checked for every texture in every frame, instead of once at shader load time. Probably not worth worrying about though.

slipher:
Removing r_normalScale define seems good. Not sure what you mean about stuff needing to be optimized as you simply removed the extra instructions from GLSL.

src/engine/renderer/tr_shade.cpp
illwieckz: Hmm, right, I'll see if it's a float.

In fact, I have a strong feeling some of those computes are wrong. At this point I'm just moving all of them at the same place, keeping the legacy behavior. Then once code is clean, it will be time to ask if it makes sense or not. Initially we had a `normalScale` expression that was expected to operate on X and Y,  and a `r_normalScale` cvar that was expected to operate on Z… I'll investigate on those things later. My first priority is to have a cleaner code so we find easily where things are done so we can question if it is the right way to do things or not and fix it easily.

illwieckz:
> @slipher: Not sure what you mean about stuff needing to be optimized

instead of this in 8 place:

```c
		vec3_t normalScale = { 1, 1, 1 };
		SetNormalScale( pStage, normalScale );
```

I do this in 8 place:

```c
		vec3_t normalScale;
		SetNormalScale( pStage, normalScale );
```

And this in 1 place (in `SetNormalScale`):

```c
 		normalScale[ 0 ] = 1;
		normalScale[ 1 ] = 1;
		normalScale[ 2 ] = 1;
```

The code is more compact this way, and less prone to errors, but in first case the vec3 is defined to 1,1,1, in the second case the vec3 is defined to 0,0,0 then set to 1,1,1.

src/engine/renderer/tr_shade.cpp
illwieckz: float fixed.

illwieckz:
Does anyone has something to say against that?

PR #232, opened 2019-12-08
enable PBR for lightMapping rendering and add support for physicalMap

illwieckz:
This PR sits above #230.

Enable PBR for lightMapping rendering, add support for physicalMap.

physicalMap expects ORM support (glTF 2.0 standard).

Test map provided by @SomaZ, I edited the shader, converted the assets and build the package the way this branch expects it:
https://dl.illwieckz.net/b/daemon/wip/pbr/map-test-pbr_0%2B20191212-075241%2B9a1edb3.dpk

I removed the swizzle mask code because as long as we don't get expected result on PBR side, we can't know if experiments with swizzles are working or not.

The PR introduces `r_physicalMapping` cvar, which is currently disabled by default (set to `0`) since PBR is not production ready.

Some part of the PBR rendering looks OK, but the lighting is not. Fixing PBR is out of topic. This PR is about being sure assets are piped through the engine from VFS to GLSL. It would make debugging PBR far easier.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -589,6 +589,11 @@ static std::string GenEngineConstants() {
 		AddDefine( str, "r_specularMapping", 1 );
 	}
 
+	if ( r_physicalMapping->integer )
+	{
+		AddDefine( str, "r_physicalMapping", 1 );
+	}
+
 	if ( r_glowMapping->integer )
 	{
 		AddDefine( str, "r_glowMapping", 1 );
--- a/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
+++ b/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
@@ -975,12 +975,13 @@ void	main()
 	}
 	diffuse.rgb *= u_LightColor * NL;
 
+#if !defined(USE_PHYSICAL_SHADING)
 #if defined(r_specularMapping)
-	// FIXME: ignore if PBR
 	// compute the specular term
 	vec4 spec = texture2D(u_MaterialMap, texCoords).rgba;
 	vec3 specular = spec.rgb * u_LightColor * pow(clamp(dot(normal, H), 0.0, 1.0), u_SpecularExponent.x * spec.a + u_SpecularExponent.y) * r_SpecularScale;
 #endif // r_specularMapping
+#endif // !USE_PHYSICAL_SHADING
 
 	// compute light attenuation
 #if defined(LIGHT_PROJ)
@@ -999,9 +1000,11 @@ void	main()
 	// compute final color
 	vec4 color = diffuse;
 
+#if !defined(USE_PHYSICAL_SHADING)
 #if defined(r_specularMapping)
 	color.rgb += specular;
 #endif // r_specularMapping
+#endif // !USE_PHYSICAL_SHADING
 
 #if !defined(LIGHT_DIRECTIONAL)
 	color.rgb *= attenuationXY;
--- a/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
+++ b/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
@@ -32,7 +32,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #endif
 
 uniform sampler2D	u_DiffuseMap;
-uniform sampler2D	u_SpecularMap;
+uniform sampler2D	u_MaterialMap;
 uniform sampler2D	u_AttenuationMapXY;
 uniform sampler2D	u_AttenuationMapZ;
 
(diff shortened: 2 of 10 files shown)


illwieckz:
I removed the swizzle mask code because as long as we don't get expected result on PBR side, we can't know if experiments with swizzles are working or not. The textures from test map were converted to not require such feature.

illwieckz:
I merged into master the very first commit that is not doing anything on the code itself, except moving some text to make subsequent PR less noisy.

illwieckz:
This PR is no longer WIP because remaining PBR bugs seems related to the implementation itself, not to the proposed changes made to enable it.

illwieckz:
I noticed the PBR compute is not done when there is no normalmap. That's something to keep in mind but that must not prevent the merge of this PR. This PR is there to make life easier to people wanting to work on the PBR part of the engine (which looks incomplete). A GLSL wizard would then just have to focus on fixing/completing GLSL code since PBR maps will be loaded right.

illwieckz:
This now looks ready to me. I have more revamp ideas but that would be another PR. Let's do it step by step.

illwieckz:
I renamed `materialMap` to `physicalMap`, the _material_ word was too much generic and was a creation introduced in e85abec00d70b79e4037e901746de63d4b0e30f6 and never used yet. So there is not problem to set a better name instead.

src/engine/renderer/glsl_source/forwardLighting_fp.glsl
```diff
@@ -975,12 +975,13 @@ void	main()
 	}
 	diffuse.rgb *= u_LightColor * NL;
 
+#if !defined(USE_PHYSICAL_SHADING)
```
slipher: Is this supposed to say r_physicalshading?

Also you could combine these into a single #if.

src/engine/renderer/tr_shader.cpp
```diff
@@ -1845,22 +1845,22 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
 				ParseSpecularMap( stage, text );
 			}
 		}
-		else if ( !Q_stricmp( token, "materialMap" ) )
+		else if ( !Q_stricmp( token, "physicalMap" ) )
```
slipher: This is compatibility breaking, right? Is it the case that zero maps are using this?

illwieckz: There is zero map using it. That's something we created and never used yet.

And if we discover another engine had the same bad idea (and I've checked some of them and found no one), we can re-add this keyword as a _compatibility layer for third party content_.

src/engine/renderer/glsl_source/forwardLighting_fp.glsl
illwieckz: No, both specular and physical textures can be rendered in the same screen. This must generate two GLSL shaders, one for specular path, one for material path, those two paths must be available together.

In fact the real logic is:

```c
#ifdef USE_PHYSICAL_SHADING
#ifdef r_physicalMapping
// some code
#endif // r_physicalMapping
#else !USE_PHYSICAL_SHADING
#ifdef r_specularMapping
// some code
#endif // r_specularMapping
#endif !USE_PHYSICAL_SHADING
```

And I already combined the two ifs for `USE_PHYSICAL_SHADING` and `r_physicalMapping`.

illwieckz: In this glsl code the physical shading is just not implemented (it's the forward renderer we don't use), that's why the `#if !defined(USE_PHYSICAL_SHADING)`: this is just to be sure a possible PBR map is not used by specular computation, but we still have to test if it's PBR map (to never do anything) or a specular map (to do something only if `r_specularMapping` is enabled).

src/engine/renderer/gl_shader.cpp
```diff
@@ -584,6 +584,11 @@ static std::string GenEngineConstants() {
 		AddDefine( str, "r_normalMapping", 1 );
 	}
 
+	if ( r_physicalMapping->integer )
+	{
+		AddDefine( str, "r_physicalMapping", 1 );
```
illwieckz: I can remove this define

illwieckz: done

illwieckz:
rebased on master

src/engine/renderer/glsl_source/forwardLighting_fp.glsl
illwieckz: Well, I discovered we cannot combine `USE_PHYSICAL_SHADING` and `r_physicalMapping` otherwise the PBR map would be used as specular map when `r_physicalMapping` is disabled.

slipher:
looks OK

illwieckz:
I just added a commit to properly route through PBR mapping if r_physicalMapping is enabled and shader is physical _or_, if it's not a physical map and r_specularMapping is disabled or it's not a physical map but r_specularMapping is disabled, produce something. We need to test everything…

src/engine/renderer/glsl_source/forwardLighting_fp.glsl
illwieckz: fixed

src/engine/renderer/gl_shader.cpp
illwieckz: _finally_, the define was required.

Issue #233, opened 2019-12-08
fail to detect alpha channel heightmap in png normal map (seems to be the same with webp)


illwieckz:
Some time ago I added some code to be able to autodetect an heightmap in a normalmap, this is very useful when loading third-party asset from games that relies on such mechanism (Xonotic maps are good example of maps that push our engine forward):

https://github.com/DaemonEngine/Daemon/blob/e63aa213b6593da0021a759f3b9d4cf30705bcde/src/engine/renderer/tr_image.cpp#L1364-L1383

I discovered that code is not able to autodetect the heightmap in png, and if I'm right, webp too. It would be useful to test against tga too.

I verified it works against crn and dds by the way (haven't experimented with ktx at all) which are some kind of “native” GPU formats.

I plan to implement an explicit `normalHeightMap` keyword (as others idtech3 derivated engines did) but it would be good to know why the above code fails on non-GPU textures formats.

illwieckz:
It looks like I fixed it in one of my work-in-progress branch.

The problem is that XreaL was bad designed and expected loose heightmap to be heightmap and embedded heightmap in normalmap to be depth map, so if normalmap had no alpha channel, loading the alpha channel would result in a weird displacement, so the code was disabling normalmap alpha channel if the shader did not allowed it explicitly, making alphachannel hence heightmap detection not possible.

After fixing the design by using an heightmap instead of a depthmap like all the industry around, we can keep normalmap alphachannel even if does not exist: it will result in a flat heightmap.

PR #235, opened 2019-12-09
implement per-stage normalHeightMap

illwieckz:
This PR sits above #230.

The engine usually attempts to autodetect heightmap in normal map, but this keyword is for people who don't trust the engine (no one will blame your for that, see #233).

Some idtech3 derivated engines like OpenJK did it this way. This is not a bad idea.


illwieckz:
this now looks ready to me

illwieckz:
this is very straightforward

PR #238, opened 2019-12-09
Remove need for our patch to Detour (see #130)

slipher:
I imported the current master of the official repository as the 'upstream' branch in our fork, and pointed the submodule reference to that. I plan to rename the current `master` branch to `obsolete/daemonassert` after that.

But maybe instead of all that, it would be better to revert the one commit in our repository that's different, and merge all the commits from upstream into our master branch. WDYT?

### DIFF
--- a/src/engine/botlib/bot_api.h
+++ b/src/engine/botlib/bot_api.h
@@ -33,6 +33,7 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 
 #include "bot_types.h"
 
+void BotInit();
 bool     BotSetupNav( const botClass_t *botClass, qhandle_t *navHandle );
 void         BotShutdownNav();
 
--- a/src/engine/botlib/bot_load.cpp
+++ b/src/engine/botlib/bot_load.cpp
@@ -31,6 +31,10 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 ===========================================================================
 */
 
+#include "common/Common.h"
+
+#include "DetourAssert.h"
+
 #include "bot_local.h"
 #include "nav.h"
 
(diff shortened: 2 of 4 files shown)


illwieckz:
I don't think you can rename a branch, so in any way you would have to push current `master` under a new branch name to keep the submodule commit available to checkout. It is good to keep the commit available since it's registered on dæmon repository history.

That said, force-pushing master with upstream branch will erase the custom commit, either we purge it before or not.

slipher:
> I don't think you can rename a branch, so in any way you would have to push current master under a new branch name to keep the submodule commit available to checkout. It is good to keep the commit available since it's registered on dæmon repository history.

Right, that's what I'm saying. I will push the current `master` branch to `obsolete/daemonassert` so that it is still available.

illwieckz:
Oh, I misread your initial comment. I don't think there is a need to revert the commit and merge things. The submodule just exist for convenience. As soon as the reference is kept, a force push of master seems to be the best way to go.

illwieckz:
If you know this works, you have my go for a merge of this and the force push on detour side.

illwieckz:
As branch name I suggest `obsolete/donotdelete` or  `obsolete/daemon/donotdelete` if you really think the branch name must contains the daemon name.

I think it's good to put a `donotdelete` flag so a distracted mind (like myself in the future or even someone who is not yet born) knows what not to do. This is more important than keeping the daemon name.

I really doubt people would fork our repo in any way, it will just be there to be able to register submodule references, apply some tags or things like that to help source management. 

So, `obsolete/daemon/donotdelete` is a bit verbose but would satisfy any need.

illwieckz:
I just noticed that on other repositories I named such branches `legacy`:
https://github.com/DaemonEngine/daemonmap/tree/legacy
https://github.com/Unvanquished/updater/tree/legacy

So maybe use `legacy/daemon/donotdelete` instead.

PR #249, opened 2019-12-20
appveyor: use Visual Studio 2019 and do 32-bit and 64-bit build

illwieckz:
use Visual Studio 2019 and do 32-bit and 64-bit build

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,19 +2,14 @@ branches:
   except:
     - debian
 
-image: Visual Studio 2017
+image: Visual Studio 2019
 
 environment:
+  generator: Visual Studio 16 2019
+  configuration: Debug
   matrix:
-    - COMPILER: msvc
-      COMPILER_VERSION: "15.0" # Visual Studio 2017
-      TARGET: x86
-      GENERATOR: "NMake Makefiles"
-
-matrix:
-  fast_finish: false
-
-configuration: Debug
+    - platform: win32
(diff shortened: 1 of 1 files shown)


illwieckz:
lol, appveyor build worked, but travis failed because what looks to be a cache proxy just went offline while building…

illwieckz:
It's all green now, but is the test significant?

slipher:
Seems like an *anti*-cargo cult commit to me, as it deletes the `matrix` stuff which did nothing for us on account of there being only one build configuration :)

Are the Travis builds also 64-bit? I think it makes sense to have at least one 32-bit build somewhere, since we still supposedly support 32-bit Windows (although maybe we should stop?)

illwieckz:
OK, it now tests both 32-bit and 64-bit platform build. Note that it does not use NMake generator but Visual Studio one. We may add a test for NMake (or Ninja, or…) if we think we really have to test those generators but in this case we would just test the platform that works and don't worry about any cross-platform build, as we would already know the platform works. We may also test older visual studio or other compilers if available, but for same reason, testing only one platform with others compilers would be enough.

illwieckz:
Note: MSYS2 is available on the selected appveyor VM.

illwieckz:
also, we may use MSYS2 shell to run [git-checkout-modules](https://gitlab.com/illwieckz/git-checkout-modules/) to sync branches.

Issue #164, opened 2019-02-19
Drop VM file handle support?


slipher:
There is code that allows the VMs to open an FS::File that is backed by a real Unix file descriptor, which can be directly operated on by the gamelogic. It works, but is used by nobody (all VM code is using the older API where each operation is an engine call operating on the table in src/engine/qcommon/files.cpp).

Reasons to delete it other than being unused:
- Potential difficulty in porting if we move off of NaCl. 
- Sketchy or at least difficult to understand from a security point of view.

The main reason to keep it would be if we expected to have a use case for high-volume file I/O inside the gamelogic.

DolceTriade:
I wonder if the file support can be extended to file support to allow moving network code into the VM?

slipher:
I don't know, not sure if there's any way to use a socket as a file on Windows.

Even if we do want to have file descriptors for network I/O, it doesn't mean we need to support it for any files in the homepath.

slipher:
Done in #250.

PR #250, opened 2019-12-23
Drop support of OS file handles in gamelogic

slipher:
To make WebAssembly migration easier by getting rid of IPC calls which aren't purely sending bytes.

The feature is currently not used anywhere. There is one bit of code in the cgame which depends on PakPath::ReadFile, but that turns out not to be used also...

### DIFF
--- a/src/common/IPC/CommonSyscalls.h
+++ b/src/common/IPC/CommonSyscalls.h
@@ -246,14 +246,14 @@ namespace VM {
 
     enum EngineFileSystemMessages {
         FS_INITIALIZE,
-        FS_HOMEPATH_OPENMODE,
+        FS_HOMEPATH_OPENMODE, // TODO(0.52) remove
         FS_HOMEPATH_FILEEXISTS,
         FS_HOMEPATH_TIMESTAMP,
         FS_HOMEPATH_MOVEFILE,
         FS_HOMEPATH_DELETEFILE,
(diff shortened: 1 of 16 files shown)


slipher:
Fixes #164 by the way.

src/common/IPC/CommonSyscalls.h
```diff
@@ -246,14 +246,14 @@ namespace VM {
 
     enum EngineFileSystemMessages {
         FS_INITIALIZE,
-        FS_HOMEPATH_OPENMODE,
+        FS_HOMEPATH_OPENMODE, // TODO(0.52) remove
```
illwieckz: I prefer the `TODO:` syntax

illwieckz:
If no one has something against the drop of that unused feature, this looks OK to me.

src/common/IPC/CommonSyscalls.h
slipher: I'm using `TODO(0.52)` as a search string to find stuff that I want to delete later. Do you have an idea what I should use instead?

illwieckz: I usually look for the complete `TODO:` string so I’m not very happy with the missing colon, but that's not very important: those TODO are not expected to last. Make sure the related commit deleting this ends in 0.52 branch and it's OK. :grin:

Issue #183, opened 2019-03-17
normalmap compressed with `crunch -dxn` are loaded with reverted B channel


illwieckz:
This bug was uncovered in #180.

All our normal maps are compressed with `crunch -dxn` and our engine loads them with reverted blue channel (Z component). This was hidden by the bug in the normalmap code that was loading the Z component from alpha channel instead of blue one. Hopefully we did not ship any heightmap in alpha channel and since `crunch -dxn` drops the alpha channel in any way, it looks like the engine was just using an empty Z component). The normalmap loads correctly if loaded from another format (like webp).

With `r_normalMapping` **disabled**:

[![dpoffsetmapping](https://dl.illwieckz.net/b/daemon/bugs/crn-reverted-normalmap/unvanquished_2019-03-17_050224_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/crn-reverted-normalmap/unvanquished_2019-03-17_050224_000.jpg)

With `r_normalMapping` **enabled** and normal map compressed with `crunch -dxn`:

[![dpoffsetmapping](https://dl.illwieckz.net/b/daemon/bugs/crn-reverted-normalmap/unvanquished_2019-03-17_050229_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/crn-reverted-normalmap/unvanquished_2019-03-17_050229_000.jpg)

With `r_normalMapping` **enabled** and normal map compressed with `crunch -dxn` and Z component reverted in engine to workaround the bug:

[![dpoffsetmapping](https://dl.illwieckz.net/b/daemon/bugs/crn-reverted-normalmap/unvanquished_2019-03-17_050324_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/crn-reverted-normalmap/unvanquished_2019-03-17_050324_000.jpg)

With `r_normalMapping` **enabled** and normal map compressed with `webp`:

[![dpoffsetmapping](https://dl.illwieckz.net/b/daemon/bugs/crn-reverted-normalmap/unvanquished_2019-03-17_050744_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/crn-reverted-normalmap/unvanquished_2019-03-17_050744_000.jpg)

[...]

PR #251, opened 2019-12-25
renderer code polishing: better way to write code doing the same things

illwieckz:
Renderer code polishing: better way to write things doing the same things.

This is mainly to make easier to implement things since the code base will be less dirty.

Basically, once this is merged, the lone heightmap support is one commit away.

A special effort is made to make boilerplate code looking the same to make easier to fix code across copy-pasted boiler plate.

Note that parallax offset operation is currently only usable with Darkplaces' `dpoffsettmapping` shader keyword. A per-stage keyword will be implemented one day. That's not a regression since this feature is not yet used outside of xonotic parallax testbed, which uses `dpoffsettmapping`.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -3811,37 +3811,35 @@ static bool ParseShader( const char *_text )
 			continue;
 		}
 		// parallax mapping
-		else if ( !Q_stricmp( token, "parallax" )
-			|| ( *r_dpMaterial && !Q_stricmp( token, "dpoffsetmapping" ) ) )
+		else if ( !Q_stricmp( token, "parallax" ) )
 		{
+			// legacy lone “parallax” XreaL keyword was
+			// never used, it had purpose to enable parallax
+			// for the current shader
+			//
+			// the engine also relied on this to know
+			// that height map was stored in normal map
+			// but there was no other storage options
+			//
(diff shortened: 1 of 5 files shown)


src/engine/renderer/tr_shader.cpp
```diff
@@ -1456,8 +1456,8 @@ static bool LoadMap( shaderStage_t *stage, const char *buffer, const int bundleI
 		}
 	}
 
-	// enable parallax if an heightmap is found
-	// and not explicitely disabled by a shader keyword
+	// tell renderer to enable parallax mapping since an heightmap is found
+	// also tell renderer to not abuse normalmap alpha channel because it's an heightmap
```
slipher: What does it mean to abuse the alpha channel?

illwieckz: https://github.com/DaemonEngine/Daemon/blob/a321153b70157e9d37dd3d2b120ff93b14a8e9f8/src/engine/renderer/glsl_source/reliefMapping_fp.glsl#L45-L58

See also extensive explanation by Fuma there:
https://github.com/DaemonEngine/Daemon/issues/183#issuecomment-473691252

I'll do a wiki page one day.

illwieckz: I pasted the link with Fuma explanation in comment.

Issue #246, opened 2019-12-14
also collapse heightmap with standalone lightmap stage


illwieckz:
Also collapse heightmap with standalone lightmap stage, maybe the same is required with standalone glowmap too. Well, if there is an heightmap, collapse it will all standalone stages.

This to avoid this:

[![flat lightmap on relief](https://dl.illwieckz.net/b/daemon/wip/stagecollapse/unvanquished_2019-12-14_160430_001.jpg)](https://dl.illwieckz.net/b/daemon/wip/stagecollapse/unvanquished_2019-12-14_160430_001.jpg)

illwieckz:
It's better to implement lone heightmap first as implementing lone heightmap may change the design of the engine in the way the fix for this would have to be rewritten.

bmorel:
is it still true? There's no instruction to reproduce

illwieckz:
It is still true. To reproduce it easily one may use a map with visible light/shadow limit on a surface with high displacement.

bmorel:
I know I'm repeating myself, but, could it be possible to get exact instructions to reproduce it on opening comment? It'll make it easier for future ourselves (or other contributors!) to check if that can still be reproduced, and how.

PR #252, opened 2019-12-25
renderer: add per-stage heightMap keyword (loose heightMap support)

illwieckz:
This PR introduces the per-stage `heightMap` keyword to load loose heightMaps.


src/engine/renderer/tr_shader.cpp
```diff
@@ -4533,6 +4571,14 @@ static void CollapseStages()
 				stages[ normalStage ].active = false;
 			}
 
+			if ( heightStage != -1 )
```
slipher: The commit message says the heightmap is a per-stage attribute, but this looks like it's in a stage by itself?

illwieckz: The `dpMaterial` DarkPlaces compatibility layer is not yet ported to the pre-collapsed stage code and I'm using Xonotic assets to test this feature, hence the temporary requirement to collapse a standalone heightmap Stage. There is no keyword available to mapper to use the standalone heightmap stage, only the per-stage attribute in any way, so there is no chance a mapper use the standalone heightmap stage by mistake and we can drop it one day.

illwieckz: I just added comments in code about this.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -134,7 +134,13 @@ vec2 ParallaxTexOffset(vec2 rayStartTexCoords, vec3 viewDir, mat3 tangentToWorld
 	{
 		currentSize *= 0.5;
 
-		float heightMapDepth = topDepth - texture2D(u_NormalMap, rayStartTexCoords + displacement * currentDepth).a;
+#if defined(USE_HEIGHTMAP_IN_NORMALMAP)
+		float depth = texture2D(u_HeightMap, rayStartTexCoords + displacement * currentDepth).a;
```
slipher: This should say `u_normalMap`?

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -116,7 +110,13 @@ vec2 ParallaxTexOffset(vec2 rayStartTexCoords, vec3 viewDir, mat3 tangentToWorld
 	{
 		currentDepth += currentSize;
 
-		float heightMapDepth = topDepth - texture2D(u_NormalMap, rayStartTexCoords + displacement * currentDepth).a;
+#if defined(USE_HEIGHTMAP_IN_NORMALMAP)
+		float depth = texture2D(u_HeightMap, rayStartTexCoords + displacement * currentDepth).a;
```
slipher: same here

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -41,27 +38,24 @@ vec3 NormalInTangentSpace(vec2 texNormal)
...
-	else
-	{
-		// alpha channel contains the height map so do not try to restore the normal map from it
-		normal = texture2D(u_NormalMap, texNormal).rgb;
-		normal = 2.0 * normal - 1.0;
-	}
+#if defined(USE_HEIGHTMAP_IN_NORMALMAP)
+	// alpha channel contains the height map so do not try to reconstruct normal map from it
+	normal = texture2D(u_NormalMap, texNormal).rgb;
+	normal = 2.0 * normal - 1.0;
```
slipher: Don't you still need to reconstruct z in this case?

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -21,16 +21,13 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
-#endif // r_normalMapping || USE_PARALLAX_MAPPING
-
 #if defined(r_normalMapping)
-uniform vec3 u_NormalScale;
+uniform sampler2D	u_NormalMap;
+uniform vec3        u_NormalScale;
 #endif // r_normalMapping
 
 #if defined(USE_PARALLAX_MAPPING)
+uniform sampler2D	u_HeightMap;
```
slipher: Should this not be defined is there is USE_HEIGHTMAP_IN_NORMALMAP?

src/engine/renderer/tr_shade.cpp
```diff
@@ -2437,7 +2531,30 @@ static void Render_liquid( int stage )
 	// depth texture
 	GL_BindToTMU( 2, tr.currentDepthImage );
 
-	if ( normalMapping || hasHeightMapInNormalMap )
+	// bind u_HeightMap u_depthScale u_parallaxOffsetBias
+	if ( parallaxMapping )
```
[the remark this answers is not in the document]
illwieckz: reflection requires heightmap or you can get errors similars to that: https://github.com/DaemonEngine/Daemon/issues/246
cubemap reflections has to be merged with lightmap shader to avoid some bugs, and I have some branch sleeping with that improvements.

liquid shader currently does not grok heightmap yet, but this is just a matter of time before they will.

Also, I have plan to merge the liquid shader with the lightmapping one, people would be able to do that:

```
textures/water/water
{
	qer_editorImage textures/water/water_p
	{
		diffuseMap textures/water/water_d
		liquidMap textures/water/water_n
		materialMap textures/water/water_d

		scroll time * .1, time * .1
		scale .5, .5
		refractionIndex 1.3
		fresnelPower 2
		fresnelScale .85
		fresnelBias .05

		blend blend
	}
}
```

The code would store normalmap in normalmap either the keyword is normalMap or liquidMap but liquidMap would set a boolean to true.
In case that boolean would be true, if r_normalMapping would be enabled the liquid computation would be done, else the diffuse one would be displayed as a fallback, reusing the normal map the classic way.

Basically I'm in a process of boilerplating the code before factorizing it.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: The code is correct because when an heightmap is in normalmap, normalmap is bound as heightmap and glsl code reads heightmap alpha. Well, we may avoid this bind by putting `uniform sampler2D	u_HeightMap;` in a `USE_HEIGHTMAP_IN_NORMALMAP` ifdef as you said. That may be better performance wise.

I wrote this code as if I was able to do swizzle masks in the future, so I would just bind the normalmap as heightmap and swizzle alpha to red, or bind heightmap as heightmap and swizzle nothing, and read heightmap red in all case in glsl. 

By doing swizzle masks, we would also be able to avoid the `USE_HEIGHTMAP_IN_NORMALMAP` macro entirely by setting to 0 the alpha channel when feeding normalmap as normalmap to glsl code.

I'm a bit concerned by the way an additional bound would affect a bit the performance.  On the other hand, implementing swizzle mask would reduce the number of shaders to compile. But since shader compilation is only done once at map loading, maybe I can rewrite that code to avoid that extra bind when heightmap is embedded in normalmap. This is about deciding we would never get rid of the `USE_HEIGHTMAP_IN_NORMALMAP` macro that duplicates shaders.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: no need to reconstruct z which is stored in `normal.b`.

slipher: In Fuma's comment the "DXT1/BC1" case has only r and g defined, no b or a.

illwieckz: Because the compute can reconstruct z without reading it. But we can just read it.
The alternative would be to reconstruct z after overwritten the heightmap with 1 to not break reconstruction.

The reconstuction trick only works for normalmap without heightmap. In one case the normalmap is stored in GBA (DXT5nm), in other case the normalmap is stored in RGB. The trick leverages the fact reading missing alpha from RGB texture gives 1.

We use the trick to avoid to do some `ifdef USE_DXT5NM` when we know there is no heigtmap, and there can't be DXT5nm textures with heightmap because alpha channel is already used (abused) so there is no need to use any trick when we know there is an heightmap, there is only one format to grok whatever the compression (or it is compressed or not).

slipher: > DXT1/BC1 compressed texture
> Assumes R = (x+1.0)/2.0, G =(y+1.0)/2.0, B = const, A = 1.0

I don't understand. Why is this case no longer relevant? Or are you deliberately dropping support?

illwieckz: There is two way to get Z, either read it from file, either reconstruct it from other components.

This is the compatibility matrix for those two methods:

|format               |storage|heightmap in alpha|compatible with trick compute|compatible with standard read|
|---------------------|-------|------------------|-----------------------------|-----------------------------|
|PNG normalMap        |RGB1   |:x:               |:heavy_check_mark:           |:heavy_check_mark:           |
|PNG normalHeightMap  |RGBA   |:heavy_check_mark:|:x:                          |:heavy_check_mark:           |
|DXT1 normalMap       |RGB1   |:x:               |:heavy_check_mark:           |:heavy_check_mark:           |
|DXT1 normalHeightMap |RGBA   |:heavy_check_mark:|:x:                          |:heavy_check_mark:           |
|DXTn normalMap       |1BGA   |:x:               |:heavy_check_mark:           |:x:                          |

NormalMap without HeightMap in alpha can can be read with trick compute or standard read whatever the format.
NormalMap with HeightMap in alpha can't be read with trick compute.

If we know the NormalMap has an HeightMap, we have to avoid the trick compute.
We can use standard read when there is an HeightMap in NormalMap since a DXTn NormalMap can't embed an HeightMap.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: I pushed a commit that does many tests to avoid the extra bind. The code may be faster but the code is far more convoluted. Do you think this is better?

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
slipher: You're right that the stored texture file does have a blue channel, which is a don't-care value in the case of RG normal maps. But if I understand correctly, the blue channel may be set to a constant value in order to get better compression -- the blue channel has nearly zero information, so you get a higher resolution on the red and green ones.

illwieckz: That's probably the reason why people implemented DXn.

I don't see the issue in reusing blue channel when it exists instead of doing an `sqrt`. The correct way to load a normal map without trick is to read RGB.

Also, to make sure the trick works we have to properly re-normalize normalmap while compressing them. We patched crunch for that, see the _rtopmip renormalize_ commit: https://github.com/DaemonEngine/crunch/commit/902f075ac933b7d21c6de3e33cdfa484e1818b54

Well, in any way this is not really on topic about implementing loose heightmaps.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: I'm experimenting on a branch about generating normalmap from heightmap when only heightmap is provided (Xonotic has some textures like that). That other patch I'm working on integrates well with the more-convoluted-code-less-texture-bind implementation so I'm now fully ok for merging that.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
slipher: OK I see, for now you are assuming that no one is combining the RG-only compression trick with height map in normal map.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
slipher: 2nd commit seems good to me

slipher:
LGTM

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: > OK I see, for now you are assuming that no one is combining the RG-only compression trick with height map in normal map.

Yes, since the DXTn compression uses the alpha channel for the normalmap, there is no way to store an heightmap in normalmap alpha channel.

PR #253, opened 2019-12-25
tr_init: make r_normalMapping a latched cvar

illwieckz:
The glsl code expects this to be defined before compilation.

Not having this a latched cvar means different code is
run when `r_normalMapping` is switched off but `vid_restart`
is not yet called, and when `r_normalMapping` is already off
when vid_restart is called.

The previous code was preventing shader compilation error
to be detected in some scenarii, when `r_normalMapping` was
switched off then switched on before doing `vid_restart`.


slipher:
Don't remove the CVAR_ARCHIVE - I believe any cvar that is a menu option needs to have CVAR_ARCHIVE, else it will not be saved properly.

illwieckz:
So I must do `CVAR_ARCHIVE | CVAR_LATCH` instead?

slipher:
Right

illwieckz:
Fixed.

illwieckz:
This looks to be the only mistake doable in that code.

PR #255, opened 2019-12-25
tr_shade: feed heatHaze and liquid shader with flatImage if normalmap is missing

illwieckz:
tr_shade: feed heatHaze and liquid shader with flatImage if normalmap is missing


slipher:
LGTM

PR #256, opened 2019-12-26
renderer: revive liquid shader (currently disabled because it is broken)

illwieckz:
By making this code live again, we make easier to not introduce bugs in this code.

A cvar is available for adventurous people to debug the feature, named `r_liquidMapping`, disabled by default.

The specular code is refactored against the one in `computeLight` library.

### DIFF
--- a/src/engine/renderer/glsl_source/liquid_fp.glsl
+++ b/src/engine/renderer/glsl_source/liquid_fp.glsl
@@ -111,9 +111,15 @@ void	main()
 	vec4 reflectColor;
 	vec4 color;
 
+#if defined(r_liquidMapping)
 	refractColor = texture2D(u_CurrentMap, texScreen).rgb;
 	reflectColor.rgb = texture2D(u_PortalMap, texScreen).rgb;
 	reflectColor.a = 1.0;
+#else // !r_liquidMapping
+	// dummy fallback color
+	refractColor = vec3(0.7, 0.7, 0.7);
+	reflectColor = vec4(0.7, 0.7, 0.7, 1.0);
+#endif // !r_liquidMapping
 
 	color.rgb = mix(refractColor, reflectColor.rgb, fresnel);
 	color.a = 1.0;
--- a/src/engine/renderer/glsl_source/liquid_fp.glsl
+++ b/src/engine/renderer/glsl_source/liquid_fp.glsl
@@ -34,7 +34,6 @@ uniform float		u_FresnelScale;
 uniform float		u_FresnelBias;
 uniform mat4		u_ModelMatrix;
 uniform mat4		u_UnprojectMatrix;
-uniform vec2		u_SpecularExponent;
 
 uniform sampler3D       u_LightGrid1;
 uniform sampler3D       u_LightGrid2;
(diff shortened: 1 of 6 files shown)


illwieckz:
The glsl code itself seems to not be broken, I don't know if the glsl compute is correct, but the glsl shader is not broken.

The glsl shader produce garbage because it receives undefined content as input, from there:
https://github.com/DaemonEngine/Daemon/blob/a321153b70157e9d37dd3d2b120ff93b14a8e9f8/src/engine/renderer/tr_backend.cpp#L4720-L4725

Instead of _“current color buffer”_ this code captures some random video memory (likely filled with leftovers).

illwieckz:
Does anyone have something against something in this PR?

This patch changes nothing on how currently supported things are currently rendered, but this change is required to clean and refactor the code (and maybe fix bugs).

src/engine/renderer/glsl_source/liquid_fp.glsl
```diff
@@ -138,24 +137,17 @@ void	main()
...
-	vec3 H = normalize(L + viewDir);
+		       lightDir, ambientColor, lightColor );
 
-	// compute the light term
-	vec3 light = lgtCol * clamp(dot(normal, L), 0.0, 1.0);
+	vec4 diffuse = vec4(0.0, 0.0, 0.0, 1.0);
 
-#if defined(r_specularMapping)
 	// compute the specular term
-	vec3 specular = reflectColor * lgtCol * pow(clamp(dot(normal, H), 0.0, 1.0), u_SpecularExponent.x + u_SpecularExponent.y) * r_SpecularScale;
```
slipher: I'm not really convinced that replacing 2 lines with a call to a big function with 4 ifdefs is really a simplification, but whatever.

src/engine/renderer/tr_backend.cpp
```diff
@@ -4716,6 +4716,7 @@ static void RB_RenderView( bool depthPass )
 
 	if ( backEnd.viewParms.portalLevel > 0 )
 	{
+		if ( r_liquidMapping->integer )
 		{
```
slipher: How this part related to liquid shading?

src/engine/renderer/glsl_source/liquid_fp.glsl
illwieckz: 1. By reusing the _big function with 4 ifdefs_ we make sure the used specular code is actually used and tested and maintained. This to prevent the code to break without being noticed.
2. I spent a lot of time to be sure the original code did the same. Now that we know it does the same, the best way to keep that information and to make sure it will be always true is to reuse the same.

Also, this is a bit like reusing a boring wall-of-text free software license written by lawyers and writing a short custom license, people will trust more the boring license because it is widely used and it was already successfully defended in courts.

src/engine/renderer/tr_backend.cpp
illwieckz: `tr.portalRenderImage` is used as `u_PortalMap` in `liquid_fp.glsl`

Note that this is the part of the liquid code that is broken.

src/engine/renderer/glsl_source/liquid_fp.glsl
illwieckz: Also, there is hope to merge liquid shader with lightmapping shader one day, so the surface would be displayed with the diffuse and normal textures as fallback if liquidmapping is disabled. The current design paints nothing if liquidmapping is disabled. See #261.

PR #257, opened 2019-12-29
debugging: implement r_showNormalMaps and r_showMaterialMaps

illwieckz:
implement `r_showNormalMaps`

`r_showEntityNormals` is dropped, in any way, `vertexLighting_DBS_entity`
and `vertexLighting_DBS_world` glsl are asking for merge

also beautify some code for other `r_show*` cvar

also introduce `r_showMaterialMaps` to debug specular and physical maps.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -558,6 +558,11 @@ static std::string GenEngineConstants() {
 		AddDefine( str, "r_showNormalMaps", 1 );
 	}
 
+	if ( r_showMaterialMaps->integer )
+	{
+		AddDefine( str, "r_showMaterialMaps", 1 );
+	}
+
 	if ( glConfig2.vboVertexSkinningAvailable )
 	{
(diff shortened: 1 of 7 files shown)


illwieckz:
I was having problems with specular maps, so I added `r_showMaterialMaps`… This to discover a mistake I did… and to somwhat fix PBR in the process…

illwieckz:
I would like to merge #256 before this one so I can also implement `r_normalMapping` in liquid shader.

illwieckz:
rebased on master and implemented `r_showNormalMaps` on `liquidmap_fp.glsl`

illwieckz:
> @slipher A warning message if 2 or more of these mutually exclusive cvars are enabled would be nice. Otherwise looks good to merge.

I would like to have such feature but that seems not be very easy to do, as we may test all conflicting options…

A better implementation would be something like `r_showDebug normalmaps' or something like that, with a number for each type. We may create an issue for this but to me that looks out of topic of this PR.

PR #258, opened 2019-12-29
renderer: compute normalmap from heightmap when missing normalmap, aka bumpmap

illwieckz:
compute normalmap from heightmap when missing normalmap, also known as bumpmapping

two algorithms available:

- one naive that does the job, uses 3 samples
- one advanced that does better job, uses 8 samples (sobel operation)

a new cvar is added: `r_sobelFiltering`, which would be used
to disable or enable various sobel filtering operation including this one

the engine is likely to implement other sobel filters in the future,
for example DarkPlaces has a postprocessing effect that does such kind of compute,
maybe Dæmon based games will want such kind of effect, in this case
a common cvar to enable or disable various sobel compute does not look bad

the `r_sobelFiltering` cvar is enabled by default


illwieckz:
It's better to merge #257 (_debugging: implement r_showNormalMaps_)  before even if this one can technically be merged before. By merging #257 first we make Dæmon able to display computed lightmaps from this commit.

illwieckz:
hmmm, currently that code seems to only run with lightMapping glsl shader, I'll enable vertex ones later. This does not prevent to review the code as it would just require some boilerplate to plug the feature in other glsl shaders.

illwieckz:
Hmm, the normalmap computation itself seems to be OK. _but_, I was testing it with Xonotic's solarium map which has such kind of "bumpmap" and everything looked good, until I remembered I had disabled normalScaling and then, all the classic normalmap (not computed ones) were using DirectX format, hence meaning that if the generated normalmap from heightmap looks good, they are also produced reverted… Which makes sense since I mostly read HLSL examples and found no one GLSL example to read do implement this.

I may edit the [`test-normal`](https://github.com/UnvanquishedAssets/UnvanquishedTestAssets/tree/master/src/map-test-normal_src.dpkdir) test map or create a new test map to get a proper testbed I'm 100% confident with.

illwieckz:
I prefer to merge this after #260 (_renderer: fix normalmap dark blotches_) in order to be able to reuse the normalmap dark-blotch fix instead of duplicating it.

src/engine/renderer/tr_shade.cpp
```diff
@@ -1240,6 +1240,8 @@ static void Render_lightMapping( int stage )
 
 	gl_lightMappingShader->SetHeightMapInNormalMap( hasHeightMapInNormalMap );
 
+	gl_lightMappingShader->SetNormalMapFromHeightMap( hasHeightMap && !hasNormalMap );
```
slipher: Should this be done in more places? I see this attribute being added to a lot of shaders, but it is only switched on in this one place.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -43,7 +44,142 @@ vec3 NormalInTangentSpace(vec2 texNormal)
 	vec3 normal;
 
 #if defined(r_normalMapping)
-#if defined(USE_HEIGHTMAP_IN_NORMALMAP)
+#if defined(USE_NORMALMAP_FROM_HEIGHTMAP)
```
slipher: This section is pretty long, would you mind putting it in a function?

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -43,7 +44,142 @@ vec3 NormalInTangentSpace(vec2 texNormal)
...
+					// Sobel computation for Y component (use the rotated X sobel kernel as Y one):
+					normal.y += heights[i][j] * sobel[j][i];
+				}
+			}
+		}
+	}
+
+	// Reconstruct Z component while making sure to not
+	// take the square root of a negative number.
+	// That may occur because of compression artifacts.
```
slipher: This comment doesn't make sense here, since you are getting the normals from your calculations, rather than an RGTC texture.

slipher:
Any screenshots? What should we see if it is working?

src/engine/renderer/tr_shade.cpp
illwieckz: Yep, it's currently only done on lightMapping shader.
I'll enable it in other places too.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: Right.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: For what I understand, the Z component is still reconstructed from X and Y, the later being computed from heightmap.

I'm getting black artifacts with that sand texture at the same place there is pure white pixels that do not look legit in the original heightmap, I'm not sure to understand everything, but I would expect those pure-white pixels artifacts to produce unexpected normal directions.

illwieckz:
I'll do screenshot, I also want to make a test map.

illwieckz:
The generated normal map from heightmap is the sand on the beach (using `r_showNormalMaps`):

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_105753_000.jpg)

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_105759_000.jpg)

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_105808_000.jpg)

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_105829_000.jpg)

So, this is with sobel filtering:

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_105855_000.jpg)

With the other less good but faster algorithm:

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_105930_000.jpg)

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_105954_000.jpg)

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_105958_000.jpg)

Close look, with sobel:

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_110054_000.jpg)

Without:

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_110125_000.jpg)

Fully textured render, with sobel:

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_110337_000.jpg)

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_110341_000.jpg)

Without:

[bumpmap](https://dl.illwieckz.net/b/daemon/wip/bumpmap/unvanquished_2020-01-18_110302_000.jpg)

src/engine/renderer/tr_shade.cpp
illwieckz: done

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: done

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
illwieckz: made the comment explicit about this.

slipher:
What seems to be is missing is a decision of how steep of an angle a certain height delta represents. So trying to reason this out geometrically... Now so far you've calculated your height differential over 1 pixel in the x direction. So the heightmap should have some physical scale for the difference between a value of 0 and a value of 1 (I don't suppose there's any standard for this?) which you would multiply by to get the physical delta-height. And you have a physical delta-x which is the distance between neighboring pixels on the texture. So (physical delta-x, 0, physical delta-height(x)) would be a tangent vector to your surface. And with reasoning from the y-coordinate calculations, another tangent is (0, physical delta-y, physical delta-height(y)). Then you could cross-product those to get a normal vector.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -29,21 +29,169 @@ uniform sampler2D	u_NormalMap;
...
+	// Set sobel X kernel (beware, line is column with this notation):
+	sobel[0] = vec3( 1.0,  2.0,  1.0);
+	sobel[1] = vec3( 0.0,  0.0,  0.0);
+	sobel[2] = vec3(-1.0, -2.0, -1.0);
+
+	for (int i = 0; i < 3; i++)
+	{
+		for (int j = 0; j < 3; j++)
+		{
+			if (i != 1 && j != 1)
```
slipher: This condition looks wrong

illwieckz: one matrix is:

    N N N
    0 0 0
    N N N

while other is:

    N 0 N
    N 0 N
    N 0 N

`mat[1][1]` is know to be zero in all cases, because only surrounding pixels are read, not the pixel itself.

illwieckz: note that I'm not sure to understand everything… but other implementations I read just did similar thing (they basically unrolled loops, ignoring the useless unrolled step).

slipher: This also ignores [0][1] and [2][1] and [1][0] and [2][0]

illwieckz: Hmm, I think I wanted to do `!( i == 1 && j == 1)`
…which must be equivalent to `(i != 1 || j !=1)`, right.

illwieckz:
> What seems to be is missing is a decision of how steep of an angle a certain height delta represents. So trying to reason this out geometrically... Now so far you've calculated your height differential over 1 pixel in the x direction. So the heightmap should have some physical scale for the difference between a value of 0 and a value of 1 (I don't suppose there's any standard for this?) which you would multiply by to get the physical delta-height. And you have a physical delta-x which is the distance between neighboring pixels on the texture. So (physical delta-x, 0, physical delta-height(x)) would be a tangent vector to your surface. And with reasoning from the y-coordinate calculations, another tangent is (0, physical delta-y, physical delta-height(y)). Then you could cross-product those to get a normal vector.

I'm not sure to have the knowledge to fully understand that (euphemism).

illwieckz:
I rebased this on master after the _great merge_.

Issue #240, opened 2019-12-10
fixing/completing PBR support


illwieckz:
Engine has _work in progress_ PBR support (see #232), may be awesome to get this usable.

illwieckz:
Until someone reports a bug, the PBR support is considered working.

PR #259, opened 2019-12-30
fix materialmaps: properly load specular and physical maps, ref #240

illwieckz:
**gl_shader: load specularmap and physicalmap properly**

fix a bug introduced in 5dd263b9bfa4a65e6bd55e3caacd11cf5f8bc58c
the renaming of a variable was not done everywhere
    
because of that, the memory video for specularmap was not set,
and by kind of luck the glsl was capturing… the diffuse map instead that was just bound before, so the diffusemap was used as a specular map but in 90% of the case this was not really noticeable (except snow on chasm being so much bright and turret texture being a bit mat.
    
this mistake was also breaking physical mapping as the diffusemap was also used as physicalmap, see #240

**tr_shade: fix pbr route in vertex world shader, do not use physical map as specular one**
    
such mistake would be harder to do if vertex entity shader and vertex world shader were only one

This fix required `r_showMaterialMaps` to be implemented to do the required investigations, see #257.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -1495,7 +1495,7 @@ void GLShader_lightMapping::SetShaderProgramUniforms( shaderProgram_t *shaderPro
 {
 	glUniform1i( glGetUniformLocation( shaderProgram->program, "u_DiffuseMap" ), 0 );
 	glUniform1i( glGetUniformLocation( shaderProgram->program, "u_NormalMap" ), 1 );
-	glUniform1i( glGetUniformLocation( shaderProgram->program, "u_SpecularMap" ),  2 );
+	glUniform1i( glGetUniformLocation( shaderProgram->program, "u_MaterialMap" ),  2 );
 	glUniform1i( glGetUniformLocation( shaderProgram->program, "u_LightMap" ), 3 );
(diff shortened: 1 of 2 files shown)


illwieckz:
This is how looks PBR test map now:

with lightmapping:

[pbr test map with lightmapping](https://dl.illwieckz.net/b/daemon/wip/pbr/unvanquished_2019-12-29_234926_000.jpg)

with vertex lighting:

[pbr test map with vertex lighting](https://dl.illwieckz.net/b/daemon/wip/pbr/unvanquished_2019-12-30_000013_000.jpg)

this is the reference (OpenJK):

[pbr test map in OpenJK](https://dl.illwieckz.net/b/blendiant/talks/files/full/shot2019-07-11_20-01-46.png)

Note that this later shot has reflextion mapping enabled (= more light) and a realtime sun lighting the scene (= more light), in any way, I still think our scene is a bit darky. In any way, by looking at the indoor room it background, which is not lit by the sun in any way, that's very similar (maybe a bit darky in Dæmon but it is subtle there).

I don't know if this being a bit darky is a bug or not, but even if that low light was a bug, PBR materials now look physically correct.

illwieckz:
SomaZ who sent me that test map said to me about my screenshot:

> looks ok to me. Unvanquished is missing ambient lighting though, because of that it looks pretty dark. Once you have working cubemaps it will look much more like the screenshot of mine.

So maybe we now have working PBR texturing. :-)

PR #260, opened 2019-12-31
renderer: fix normalmap dark blotches

illwieckz:
I discovered that bug while toying with `r_showNormalMaps` (see #257).

I was like:

Wait,

[weird normal z reconstruction](https://dl.illwieckz.net/b/daemon/bugs/weird-normal-z-reconstruction/unvanquished_2019-12-31_001650_000.jpg)

Wait…

[weird normal z reconstruction](https://dl.illwieckz.net/b/daemon/bugs/weird-normal-z-reconstruction/unvanquished_2019-12-31_001632_000.jpg)

Wait………

[weird normal z reconstruction](https://dl.illwieckz.net/b/daemon/bugs/weird-normal-z-reconstruction/unvanquished_2019-12-31_001747_000.jpg)

As you can see the bug is also visible when diffuse and everything is rendered properly.

So I clamped normal Z component between 0 and 1… and magically the bug disappeared!

Before:

[weird normal z reconstruction](https://dl.illwieckz.net/b/daemon/bugs/weird-normal-z-reconstruction/unvanquished_2019-12-31_002441_000.jpg)

[weird normal z reconstruction](https://dl.illwieckz.net/b/daemon/bugs/weird-normal-z-reconstruction/unvanquished_2019-12-31_002507_000.jpg)

After:

[weird normal z reconstruction](https://dl.illwieckz.net/b/daemon/bugs/weird-normal-z-reconstruction/unvanquished_2019-12-31_002655_000.jpg)

[weird normal z reconstruction](https://dl.illwieckz.net/b/daemon/bugs/weird-normal-z-reconstruction/unvanquished_2019-12-31_002720_000.jpg)

I don't know why this bug happens, and why this fix works, but it works. This is a fix I first discovered when I faced similar bug while generating normalmap from heightmap (see #258). This is why I implemented the fix after all the available ways to load normalmaps, and not only at the end of the Z reconstruction: the fix will be used later by normalmap generation too.

Note that an `abs(normal.z)` is not enough, `clamp(normal.z, 0, 1.0)` is enough.

-----

Edit: the depicted texture is `shared_pk02/rock01`, so the file is `textures/shared_pk02_src/rock01_n`. Note that the normalmap is renormalized at crunch time.

### DIFF
--- a/src/engine/renderer/glsl_source/reliefMapping_fp.glsl
+++ b/src/engine/renderer/glsl_source/reliefMapping_fp.glsl
@@ -59,9 +59,22 @@ vec3 NormalInTangentSpace(vec2 texNormal)
 	normal = texture2D(u_NormalMap, texNormal).rga;
 	normal.x *= normal.z;
 	normal.xy = 2.0 * normal.xy - 1.0;
-	normal.z = sqrt(1.0 - dot(normal.xy, normal.xy));
+	// In a perfect world this code must be enough:
+	// normal.z = sqrt(1.0 - dot(normal.xy, normal.xy));
+	//
+	// Unvanquished texture known to trigger black normalmap artifacts
+	// when doing Z reconstruction:
+	//   textures/shared_pk02_src/rock01_n
+	//
+	// Although the normal vector is supposed to have a length of 1,
+	// dot(normal.xy, normal.xy) may be greater than 1 due to compression
+	// artifacts: values as large as 1.27 have been observed with crunch -dxn.
+	// https://github.com/DaemonEngine/Daemon/pull/260#issuecomment-571010935
+	//
+	// This might happen with other formats too. So we must take care not to
+	// take the square root of a negative number here.
+	normal.z = sqrt(max(0, 1.0 - dot(normal.xy, normal.xy)));
 #endif // !USE_HEIGHTMAP_IN_NORMALMAP
-
 	// HACK: 0 normal Z channel can't be good
 	if (u_NormalScale.z != 0)
 	{


illwieckz:
The bug is reproducible with stock chasm map (outdoor uses that `pk02/rock01` texture) on 0.51.1 release:

[weird normal z reconstruction](https://dl.illwieckz.net/b/daemon/bugs/weird-normal-z-reconstruction/unvanquished_2020-01-06_000109_000.jpg)

slipher:
I checked the DXT5 normal maps on Chasm for out-of-range values. The sum of the squares of the two components (derived from byte values by x / 255 * 2 - 1) should be no greater than 1 to get a valid 3rd component. Many of the normalmap textures had a few hundred or a few thousand pixels with a sum greater than 1. Presumably, this happens when a vector which originally has an L2 norm less than 1 becomes greater than 1 after being lossily compressed.  I confirmed that this is the cause of most of the black areas by replacing any block containing a bad pixel with a solid color (however there is 1 patch that remains mysteriously black which you can see in the screenshot). 
unvanquished_2020-01-05_213323_000

illwieckz:
Wow, that's precious to know! So basically things are going wrong because of compression artifacts…

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -61,12 +61,16 @@ vec3 NormalInTangentSpace(vec2 texNormal)
...
 	// HACK: 0 normal Z channel can't be good
 	if (u_NormalScale.z != 0)
 	{
 		normal *= u_NormalScale;
 	}
+
+	// HACK: prevent artifacts that may produce black normalmap pixels
+	// Unvanquished texture known to trigger the bug when doing Z reconstruction:
+	//   textures/shared_pk02_src/rock01_n
+	normal.z = clamp(normal.z, 0.0, 1.0);
```
slipher: Apparently clamp is not well-defined for NaNs (https://www.shadertoy.com/view/Mtscz2), so we should do it a different way. Like `normal.z = sqrt(max(0, 1.0 - dot(normal.xy, normal.xy)))`

illwieckz:
do you think this bug can appear with normalmaps loaded the RGB way?

slipher:
Instead of this we could also consider trying to make the crunch renormalize option always produce vectors with a norm of less than 1. But this patch is certainly easier :)

illwieckz:
In any way, it's good to have an engine being able to automatically fix compression artifact, even if we patch every software on the planet to not produce problematic compression artifacts.

That said, I'm not against a crunch fix if that's doable.

illwieckz:
note: doing a crunch fix for this would make us eligible for a blog post about it, and we would send it to Binomial/Unity/Khronos so the next Khronos texture standard would take care of this. :wink:

slipher:
Yeah, this PR seems a definite good idea in any case considering the large variety of inputs which can have z-component reconstruction. Also a comment about why the sum of the first 2 components can be too large would be good.

illwieckz:
I'm copy pasting your comment

illwieckz:
@slipher, your `max()` code fixes the artifacts as well, so I'm using it.

Do you know if there would be a need for a similar fix with normalmap read the RGB way? And what it would be?

slipher:
> Do you know if there would be a need for a similar fix with normalmap read the RGB way? And what it would be?

This GLSL code is downstream from texture decompression, so it doesn't matter how the image was obtained. You just get some RGBA values from 0 to 1.

So in the z-reconstruction case, the code would work regardless of the input texture format. And in the non-z-reconstruction case, there is no taking the square root of a possibly negative number, so the problem could never occur at all.

illwieckz:
Thanks, that's good to know.

src/engine/renderer/glsl_source/reliefMapping_fp.glsl
```diff
@@ -56,12 +56,47 @@ vec3 NormalInTangentSpace(vec2 texNormal)
...
-	normal.z = sqrt(1.0 - dot(normal.xy, normal.xy));
+	// In a perfect world this code must be enough:
+	// normal.z = sqrt(1.0 - dot(normal.xy, normal.xy));
+
+	// Unvanquished texture known to trigger black normalmap artifacts
+	// when doing Z reconstruction:
+	//   textures/shared_pk02_src/rock01_n
+	// -- illwieckz
+
+	// I checked the DXT5 normal maps on Chasm for out-of-range values.
```
slipher: All of the following discussion is a bit verbose; how about this:

Although the normal vector is supposed to have a length of 1, dot(normal.xy, normal.xy) may be greater than 1 due to compression artifacts: values as large as 1.27 have been observed with crunch --<insert options here>. This might happen with other formats too. So we must take care not to take the square root of a negative number here.

PR #262, opened 2020-01-05
appveyor: also test Visual Studio 2017 and investigate NMake generator as we support them

illwieckz:
README says the minimum supported VS is VS2017 so it's better to test it

Unvanquished uses `NMake Makefiles` generator so it's better to test it

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,19 +5,14 @@ branches:
 environment:
   configuration: Debug
   matrix:
+    # see 96d5c1f3ed77b09c64ce7c3c7cbd37c70456b3db
+    # for NMake template
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
       generator: Visual Studio 16 2019
       platform: x64
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
-      generator: NMake Makefiles
-      # -Aplatform is not supported by CMake's NMake Makefiles generator and produces an error
-      VCVARSBAT: vcvars64.bat
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
-      generator: Visual Studio 16 2019
-      platform: win32
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       generator: Visual Studio 15 2017
-      platform: x64
+      platform: win32
 
 build:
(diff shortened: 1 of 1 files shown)


illwieckz:
OK, this looks OK to me now.

I don't have solution to build NMake 32-bit build, I don't know why Unvanquished manages to do it.

illwieckz:
The bad thing is that because of testing those various build scenarii, the appveyor run now lasts 20 minutes.

We now test:

- VS2019 amd64
- VS2019 i386
- VS2019 Nmake amd64
- VS2017 amd64

If we do that:

- VS2019 amd64
- VS2019 Nmake amd64
- VS2017 i386

We save 1 run, reducing the total run time to 15min instead of 20min.

slipher:
There's no real reason to care about the NMake Makefiles generator. I always figured that it must be needed because of the Visual Studio generator being only usable with the IDE, but since you can apparently use it on Appveyor, it seems better to just do that.

illwieckz:
OKOK, If you don't mind, I would add a commit to disable NMake and not squash them so we keep knowledge in history. Getting it working was not that easy. Losing the information would be a shame.

What about testing 64-bit on VS2019 and 32-bit on VS2017? it would reduce the build time to 10min.
That would not be a complete test, but eh… 32-bit people are likely to run old OS hence old VS…

illwieckz:
Well, that would be a shame to discover the day of the release we fail to build 32-bit on top-notch setup because of a silent bug…

slipher:
> What about testing 64-bit on VS2019 and 32-bit on VS2017? it would reduce the build time to 10min.

Yes, I think this is good enough. Rather than filling every cell of the build matrix, so to speak, let's try to get at least one entry in each row/column.

> Well, that would be a shame to discover the day of the release we fail to build 32-bit on top-notch setup because of a silent bug…

Well, we've always used Mingw for releases anyway.

illwieckz:
right

illwieckz:
so I merged with VS2017-32 and VS2019-64 being built only

note that there is there a good example of MSYS2-in-Appveyor implementation: https://github.com/mypaint/libmypaint/commit/553f175317c5fa34fc5e3fa700848c0a5a21ac47

and we may also implement appveyor conditional build: https://www.appveyor.com/blog/2018/09/11/how-to-skip-build-matrix-jobs-conditionally/

PR #263, opened 2020-01-07
polishing: unify the way glsl shaders are written

illwieckz:
unify the way glsl shaders are written:

- make easier to compare them while looking for bugs
- make factorisation or merge easier

This is the way I fixed #247

This makes easier to merge vertex entity and vertex world shaders as one, and maybe merge the lightmapping one too, one day. This would be good to avoid duplicates of code living their own life with their own bugs.

### DIFF
--- a/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
+++ b/src/engine/renderer/glsl_source/forwardLighting_fp.glsl
@@ -934,9 +934,9 @@ void	main()
 
 	// compute light direction in world space
 #if defined(LIGHT_DIRECTIONAL)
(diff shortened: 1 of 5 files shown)


illwieckz:
In fact the best way to review this is to do things like that:

```
meld lightMapping_fp.glsl \
     vertexLighting_DBS_entity_fp.glsl \
     vertexLighting_DBS_world_fp.glsl
```

but that's not supported by GitHub.

slipher:
LGTM

illwieckz:
Thanks for the review!

PR #264, opened 2020-01-08
cvar: make r_dpMaterial, fs_legacypaks and fs_maxSymlinkDepth latched cvars

illwieckz:
- reimplement latched cvar in new cvar system
- make `r_dpMaterial` a latched cvar
- make `fs_legacypaks` and `fs_maxSymlinkDepth` latched cvars

All I want is to be told to restart when I change those cvars.

Note that:

1. it seems to work
3. @Kangz wrote in ad04f91 that `CVAR_LATCH` is ”not longer supported” and “will be implemented by the proxy” which is not I done (and I don't know what Kangz expected): https://github.com/DaemonEngine/Daemon/blob/ad04f91fb60508d04a894205f728159abf912681/src/engine/framework/CvarSystem.h#L40

So, they may be dragons!

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -123,8 +123,8 @@ template<> struct SerializeTraits<FS::LoadedPakInfo> {
 namespace FS {
 
 #ifdef BUILD_ENGINE
-static Cvar::Cvar<bool> fs_legacypaks("fs_legacypaks", "also load pk3s, ignoring version", Cvar::NONE, false);
-static Cvar::Cvar<int> fs_maxSymlinkDepth("fs_maxSymlinkDepth", "max depth of symlinks in zip paks (0 means disabled)", Cvar::NONE, 0);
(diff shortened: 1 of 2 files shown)


slipher:
I tried the latched cvar and don't think it is working. Compare the behavior to an old-style latched cvar:

```
]/r_dpMaterial 
"r_dpMaterial" - bool - Enable DarkPlaces material compatibility^* - default: "off^*"
]/r_dpMaterial on
The change will take effect after restart.
]/r_dpMaterial
"r_dpMaterial" - "on" - bool - Enable DarkPlaces material compatibility^* - default: "off^*"
]/r_customheight
"r_customheight" - "700^*"^* - default: "1024^*"
]/r_customheight 702
The change will take effect after restart.
]/r_customheight 
"r_customheight" - "700^*" - latched value "702^*"^* - default: "1024^*"
```

I can look into implementing it for new-style cvars this weekend.

illwieckz:
I see, I admit all I was looking for was a message telling me to restart. It's true that `LATCH` does more than that.

slipher:
See #272 for latched cvars

PR #265, opened 2020-01-08
travis: attempt to make brew more resilient

illwieckz:
attempt to make brew more resilient

by using `BrewFile` mechanism, `brew` does not fail if package is already installed


slipher:
It seems this still happens, e.g. https://travis-ci.org/DaemonEngine/Daemon/jobs/634014616?utm_medium=notification&utm_source=github_status

illwieckz:
The travis job shows the built branch was not yet rebased above this fix (not including it yet).

PR #266, opened 2020-01-08
renderer: make dpMaterials collapsed shader from the start, including non-shader images

illwieckz:
- now useless very ugly `FindMapInStage` is now dropped
- now useless heightmap collapsing code is now dropped
- extra maps loading for non-shader images is now enabled

as DarkPlaces only supports one kind of lightmap blend operation we can blindly collapse lightmap stages in darkplaces pre-collapsed stages.

So, shader parsing and rendering must be faster, and code is cleaner.

Next step is to merge cube Reflection with lightMapping glsl shader, a lot of code will disappear and bugs with it.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -1430,7 +1430,6 @@ static bool LoadMap( shaderStage_t *stage, const char *buffer, const int bundleI
 		case stageType_t::ST_HEATHAZEMAP:
 		case stageType_t::ST_LIQUIDMAP:
 			imageBits |= IF_NORMALMAP;
-
 		default:
 			// silence warning for other types, we don't have to take care of them:
 			//    warning: enumeration value ‘ST_GLOWMAP’ not handled in switch [-Wswitch]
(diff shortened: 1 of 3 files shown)


illwieckz:
Now, the remaining bugs in solarium maps are:

- rendering issue of scrolling collapsed shader, which is a bug in Dæmon we have in Unvanquished if we attempt to do the same: Dæmon only supports proper multi-texturing on textures without effects applied on them. I get the exact same bugs on my test-water map when disabling water shader, and this test map with related assets are built the Unvanquished way from the start;
- wrong light curve because of missing sRGB ligthing, something Unvanquished needs if we want physically correct lighting;
- non-working water shader, something Unvanquished needs if we want good-looking water;
- dynamic light issues on surfaces with cubemap reflection applied over lightmap since only lightmap stage processes the dynamic lights, something that would affect Unvanquished assets if we used that feature, this is asking for a glsl merge to do them in one stage;
- issues when strong displacement is applied on textures with discarded fragments (like grids), something that would affect Unvanquished assets if we uses the feature;
- bad parallax on boxes, which is an issue in xonotic assets (not engine's fault);
- bad translucent on grass, maybe an issue in xonotic assets (they render fine when using another alpha operation…, so maybe the shader does not setup the right operation and darkplaces just not support the other blend mode that may be mistakenly used, like does with lightmap blending).

So basically the remaining issues to render properly the Solarium map are :

1. Bugs or missing features in Dæmon that are unrelated to Xonotic;
2. Bugs in Xonotic assets.

Alongside Solarium, the Erbium map is even more buggy than Solarium as shaders are mostly garbage trying to push the right buttons and triggers the Darkplaces Q3 compatibility mode than writing correct Q3 shaders. ¯\\\_(ツ)_/¯

I'm impressed about the number of bugs and how crippled the design of many Dæmon renderer parts.
Most of bugs are due to design issues, or made easier because of them.

It looks like the XreaL renderer was just cut in pieces to hack some quick and dirty doom3 shader compatibility for showcase purposes, but that was done not the future proof way.

src/engine/renderer/tr_shader.cpp
```diff
@@ -1795,9 +1795,9 @@ void LoadExtraMaps( shaderStage_t *stage, const char *colorMapName )
 
 		COM_StripExtension3( colorMapName, colorMapBaseName, sizeof( colorMapBaseName ) );
 
-		for ( const extraMapParser_t parser: extraMapParsers )
+		for ( const extraMapParser_t parser: dpExtraMapParsers )
 		{
-			std::string extraMapName = Str::Format( "%s_%s", colorMapBaseName, parser.suffix );
+			std::string extraMapName = Str::Format( "%s%s", colorMapBaseName, parser.suffix );
```
slipher: could use `+` instead of `Str::Format`

illwieckz:
```c
tr_shader.cpp:1809:48: error: invalid operands of types ‘char [64]’ and ‘const char* const’ to binary ‘operator+’
 1809 |    std::string extraMapName = colorMapBaseName + parser.suffix );
      |                               ~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~
      |                               |                         |
      |                               char [64]                 const char* const
```

PR #267, opened 2020-01-09
DarkPlaces only supports one alphaFunc operation, GE128

illwieckz:
DarkPlaces only supports one alphaFunc operation:

https://gitlab.com/xonotic/darkplaces/blob/324a5329d33ef90df59e6488abce6433d90ac04c/model_shared.c#L1875-1876

Which is GE128:

https://gitlab.com/xonotic/darkplaces/blob/0ea8f691e05ea968bb8940942197fa627966ff99/render.h#L95

Because of that people may silently introduce regressions in their textures
designed for GT0 by compressing them using a lossy picture format like Jpg.
Xonotic texture known to trigger this bug:

    models/desertfactory/textures/shaders/grass01

Using GE128 instead would hide Jpeg artifacts while not breaking that much
non-Darkplaces GT0.

No one operation other than GT0 an GE128 was found in whole Xonotic corpus,
so if there is other operations used in third party maps, they were broken
on Darkplaces and will work there.

Xonotic assets were fixed to make sure any idTech 3 renderers display the Xonotic
textures the same way Darkplaces displays them by telling renderer to use GE128
operator:

- https://gitlab.com/xonotic/xonotic-maps.pk3dir/merge_requests/140
- https://gitlab.com/xonotic/xonotic-data.pk3dir/merge_requests/753
- https://gitlab.com/xonotic/xonotic-nexcompat.pk3dir/merge_requests/2

But there is thousands of third party maps around there, so the workaround
is required.

Hopefully in case of legit GT0, a GE128 is not that bad, for example with
the grass texture quoted above, it would just be thinner.

Things may be less good in some case, when only the dark frame is expected
to be discarded but there is some dark pixels to keep (typically a road sign).

### DIFF
--- a/src/engine/renderer/gl_shader.h
+++ b/src/engine/renderer/gl_shader.h
@@ -1243,7 +1243,27 @@ public:
 
 		switch( stateBits & GLS_ATEST_BITS ) {
 			case GLS_ATEST_GT_0:
-				value = 1.0f;
+				if ( *r_dpBlend )
+				{
+					// DarkPlaces only supports one alphaFunc operation:
+					//   https://gitlab.com/xonotic/darkplaces/blob/324a5329d33ef90df59e6488abce6433d90ac04c/model_shared.c#L1875-1876
+					// Which is GE128:
+					//   https://gitlab.com/xonotic/darkplaces/blob/0ea8f691e05ea968bb8940942197fa627966ff99/render.h#L95
+					// Because of that, people may silently introduce regressions in their textures
+					// designed for GT0 by compressing them using a lossy picture format like Jpg.
+					// Xonotic texture known to trigger this bug:
+					//   models/desertfactory/textures/shaders/grass01
+					// Using GE128 instead would hide Jpeg artifacts while not breaking that much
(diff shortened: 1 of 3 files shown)


illwieckz:
Before:

[GT0](https://dl.illwieckz.net/b/daemon/features/dpge128/unvanquished_2020-01-09_092417_000.jpg)

After:

[GE128](https://dl.illwieckz.net/b/daemon/features/dpge128/unvanquished_2020-01-09_092622_000.jpg)

illwieckz:
I'll make this using a dedicated cvar, not the `r_dpMaterial` one, maybe `r_dpBlend`, so in the future it would be possible to enable `r_dpMaterial` without this hack being used (when using assets that are fixed, for example).

illwieckz:
I pushed a completely new implementation, introducing `r_dpBlend` which can be set on and off whatever r_dpMaterial is, and can be switched at runtime (not a latched cvar). This paves the way for games like Xonotic to enable or disable this code path at map load by using some heuristics.

slipher:
LGTM

PR #268, opened 2020-01-09
builtin: import to-be-builtin assets in pkg/daemon_src.dpkdir

illwieckz:
The lifegoal of those assets is to be embedded in engine, a bit like glsl shaders are.
Those are the assets required by the engine to run.

Before we find a solution to embed them in engine, shipping them within engine
repository is not only the first step to that objective, it makes easier for
third-party games developing games on Dæmon without forking Unvanquished to start
the engine, by loading the dpkdir from the engine repository.

For example the Xonotic's daemon-glue would be able to do something like that:

    build/daemon -pakpath libs/daemon/pkg -set fs_basepath daemon […]

See https://gitlab.com/xonotic/daemon-glue for Xonotic's Daemon Glue.

See https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir for
the repository those assets come from.


illwieckz:
Because we don't know how to embed those assets yet, games have to ship them themselves, but it would be nice to not require games to fix them. Even before embedding them in engine, shipping them in engine repository makes life easier.

In the past I did this as a temporary solution:
https://gitlab.com/illwieckz/daemon-engine_src.dpkdir/

This PR removes the need for an extra repository.

slipher:
How is `daemon.shader` referenced?

illwieckz:
> How is `daemon.shader` referenced?

There is no need to reference it, the engine loads all `.shader` files it finds in all dpk/dpkdir it loads.

The `shaderlist.txt` mechanism is only used by netradiant/q3map2 to not display everything that is not required for mapping (like engine and game internals).

Those files are currently shipped with [`unvanquished_src.dpkdir](https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir). I now see this file is named [`engine.shader`](https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/blob/master/scripts/engine.shader) in that other repo. I made this PR from a temporary standalone repo I had, maybe I'll rename it to `engine.shader` there too.

PR #272, opened 2020-01-13
Implement CVAR_LATCH for new cvars

slipher:
To do before merging: Prevent `The change will take effect after restart` messages from being printed when the cvars are set from command line/configs.

### DIFF
--- a/src/common/Cvar.h
+++ b/src/common/Cvar.h
@@ -46,6 +46,7 @@ namespace Cvar {
         TEMPORARY  = BIT(8), // The cvar is temporary and is not to be archived (overrides archive flags)
         CHEAT      = BIT(9), // The cvar is a cheat and should stay at its default value on pure servers.
         USER_ARCHIVE = BIT(14), // The cvar is saved to the configuration file at user request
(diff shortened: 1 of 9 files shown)


illwieckz:
thanks for doing this!

illwieckz:
I rebased #262 over this and it works correctly on my end. LGTM.

src/common/Cvar.h
```diff
@@ -46,6 +46,7 @@ namespace Cvar {
         TEMPORARY  = BIT(8), // The cvar is temporary and is not to be archived (overrides archive flags)
         CHEAT      = BIT(9), // The cvar is a cheat and should stay at its default value on pure servers.
         USER_ARCHIVE = BIT(14), // The cvar is saved to the configuration file at user request
+        LATCH      = BIT(15), // The cvar's value can be changed only after a cgame restart. Intentionally different from CVAR_LATCH.
```
illwieckz: Just for my knowledge, why it has to be different? maybe a comment explaining why would be good, that's my only concern.

slipher: Done.

illwieckz:
good.

PR #275, opened 2020-01-24
Fix range checks for numBorders in CM_AddFacetBevels

### LINKED ISSUE #169: ClipWinding: MAX_POINTS_ON_WINDING with moteof (q3) and oasis (wolfet) maps
Both `ioquake3`[🗗](https://github.com/ioquake/ioq3/blob/d068e1dce174d2631d2d760d1d729a9d1d4151b8/code/qcommon/cm_polylib.h#L31) and `daemon` defines `MAX_POINTS_ON_WINDING` to `64`:

https://github.com/DaemonEngine/Daemon/blob/be89425dd92889b0f965d6890a804148b7832711/src/common/cm/cm_polylib.h#L43

But while ioquake3 loads “[The Edge of Forever](http://www.simonoc.com/pages/design/sp/moteof.htm)” with that setting, dæmon requires this value to be set to `134`.

If this value is less or equal to `133`, this message is printed on console and the map loading aborts:

```
Warn: ClipWinding: MAX_POINTS_ON_WINDING 
```

There is no reason this map can't be loaded with a max set to `64`.

On a side note, because the problem aborts the map loading, this problem must be raised as an error instead of a warning.

## How to reproduce

Download map [here](https://dl.illwieckz.net/b/daemon/bugs/max-points-on-winding/map-moteof_0%2Bfinal2.dpk) (it's just a dpk repack of the [original pk3 for quake3](http://www.simonoc.com/files/maps/sp/moteof_final2.zip)), load it with `+devmap moteof`.

illwieckz:
With hope to find the cause of #169 I was comparing our code with the one from other engine and my sharp eyes noticed this. This is not the cause of #169, but it may be useful anyway.

Adapted from patch by @zturtleman, he said at that time it was to fix a bug uncovered by Coverity.

See https://github.com/ioquake/ioq3/commit/ee2541efee2eace042303197a1645013d22fcc34
See https://github.com/etlegacy/etlegacy/commit/0fa531a3a958a287d424a6e56198ac14fc3aec5c

@zturtleman for some reason, compared to ioquake3 the etlegacy patch only has 2 fixes on 3, do you know why? Is it an oversight?

### DIFF
--- a/src/common/cm/cm_plane.cpp
+++ b/src/common/cm/cm_plane.cpp
@@ -495,9 +495,10 @@ void CM_AddFacetBevels( cFacet_t *facet )
 
 			if ( i == facet->numBorders )
 			{
-				if ( facet->numBorders > MAX_FACET_BEVELS )
+				if ( facet->numBorders >= MAX_FACET_BEVELS )
 				{
 					Log::Warn( "too many bevels" );
+					continue;
 				}
 
 				facet->borderPlanes[ facet->numBorders ] = CM_FindPlane2( plane, &flipped );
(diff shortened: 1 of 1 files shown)


zturtleman:
It's probably an oversight. The third check was added to ET:Legacy in a separate commit. https://github.com/etlegacy/etlegacy/commit/f0204f1a72b2f09a1e9a33f9a4732a40714a8cc7

illwieckz:
I see! thanks for your answer 👀

PR #278, opened 2020-01-28
 tr_shade: reduce diff noise between vertex entity, vertex world and lightmap code (and more)

illwieckz:
This is the huge rewording work I made when I was looking for the fix described in #277.

Basically I had identified that the bug was bound to the `Render_lightMapping()` function as swaping it with `Render_generic()` or `Render_vertexLighting_DBS_world` made the feature working.

But at this point I didn't know if the issue was in those C++ functions, in GLSL fragment code, or in GLSL vertex code.

At the same time, I wanted to do that rewording because it would benefit for many other work: collapsing the cube reflection, perhaps merge this vertex/lightmapping code one day if it does not clutter the code too much, etc. etc.

Three noticeable changes:

- The `Render_lightMapping()` C++ function was using some global variable directly while other code from `Render_vertexLighting_DBS_world()` and `Render_vertexLighting_DBS_entity()` was copying the variable to a local one before using it. Without more knowledge, I thought it was safe to make the lightMapping C++ code copy the variable before using it too. Maybe it may fix bugs (SMP anyone?) while it would be very unlikely to introduce some, and the operation is probably minor on CPU since other code does it by default any time.
- The `Render_vertexLighting_DBS_world()` function was calling `GL_CheckErrors()` two times while others were doing it once, once may be enough.
- The lightMapping code was feeding GLSL with a `u_ModelMatrix` matrix that was unused.

Without this work, I would not be able to fix #277, which was basically what remained different for no explained reason after everything looked the same.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -1478,7 +1478,6 @@ GLShader_lightMapping::GLShader_lightMapping( GLShaderManager *manager ) :
 	u_Color( this ),
 	u_AlphaThreshold( this ),
 	u_ViewOrigin( this ),
-	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
 	u_ParallaxDepthScale( this ),
 	u_ParallaxOffsetBias( this ),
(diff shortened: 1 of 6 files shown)


slipher:
This is too much work to review, so just merge it :stuck_out_tongue:

illwieckz:
Yeah, there is no easy tool to review this. To do it properly I extracted those functions as loose file to compare them together.

PR #282, opened 2020-02-02
Add layout to serverinfo

Gireen:
This fixes Unvanquished/Unvanquished#1122

### DIFF
--- a/src/engine/server/sv_init.cpp
+++ b/src/engine/server/sv_init.cpp
@@ -612,7 +612,7 @@ void SV_Init()
 	Cvar_Get( "protocol", va( "%i", PROTOCOL_VERSION ), CVAR_SERVERINFO  );
 	sv_mapname = Cvar_Get( "mapname", "nomap", CVAR_SERVERINFO | CVAR_ROM );
 	Cvar_Get( "pakname", "", CVAR_SERVERINFO | CVAR_ROM );
-	Cvar_Get( "layout", "", CVAR_SERVERINFO | CVAR_ROM );
+	Cvar_Get( "layout", "", CVAR_SERVERINFO );
 	Cvar_Get( "g_layouts", "", 0 ); // FIXME
 	sv_hostname = Cvar_Get( "sv_hostname", UNNAMED_SERVER, CVAR_SERVERINFO  );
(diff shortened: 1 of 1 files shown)


src/engine/server/sv_init.cpp
```diff
@@ -612,7 +612,7 @@ void SV_Init()
 	Cvar_Get( "protocol", va( "%i", PROTOCOL_VERSION ), CVAR_SERVERINFO  );
 	sv_mapname = Cvar_Get( "mapname", "nomap", CVAR_SERVERINFO | CVAR_ROM );
 	Cvar_Get( "pakname", "", CVAR_SERVERINFO | CVAR_ROM );
-	Cvar_Get( "layout", "", CVAR_SERVERINFO | CVAR_ROM );
```
DolceTriade: Why does this need to be removed to make it work? Layout should only be set internally by map loading commands and shouldn't be set directly right?

Gireen: Cause after removing it it works. 😄
If i understand it correct it is the same problem as with g_layouts, that CVAR_ROM can only be changed by daemon?
But the value of layout is [decided by unvanquished](https://github.com/Unvanquished/Unvanquished/blob/e66d2747fa6900aecdc44200582bd4708bba2132/src/sgame/sg_buildable.cpp#L2325) and stays emty without this change.

illwieckz: How it behaves if you only remove the `CVAR_ROM` flag?

Gireen: the cvar "layout" then contains the current layout and it also shows up in the serverinfo

illwieckz: On a side note, I was really surprised that layout would be an engine thing, that really looks like a game thing. Maybe it's an engine thing because of file writing?

Gireen: This is one of two places where layouts are mentioned inside the engine. The other is a check for the map command. Everything else is on the game side.

DolceTriade: Yeah, you're right Gireen. I believe in quake3 calling Cvar_Set allows writing to ROM only values, whereas in Daemon it does not. We can maybe change this in our code.

illwieckz:
I guess we can merge this, right?

DolceTriade:
yes

PR #287, opened 2020-02-09
renderer: fix infuriating grey line

illwieckz:
Maybe I'm a wizard, then.

I've noticed that if the bug is not in SSAO neither FXAA code, using both makes worst artifacts than only one of those. I also noticed that ticking FXAA on and off (the later not requiring an engine restart), I was seeing the screen being more streched with on than off, in real time.

Here is with  and without fxaa (while ssao is enabled):

[with ssao, without fxaa](https://dl.illwieckz.net/b/daemon/bugs/boggy-lines/unvanquished_2020-02-09_122221_000.png)

[with ssao, with fxaa](https://dl.illwieckz.net/b/daemon/bugs/boggy-lines/unvanquished_2020-02-09_122226_000.png)

Those screenshots can also be [seen there with a nice slider](https://imgsli.com/MTE5MTg).

So, I've noticed that the only common code in both shaders was something like that:

```
gl_FragCoord.st * r_FBufScale;
```

Which is the computation for the pixel size.

I then thought: _“maybe those floats in `r_FBufScale` (a `vec2`) have rounding errors”_, I then modified them in shader the arbitrary way and I was able to produce garbage based on vertical and horizontal lines the same way:

[garbage with lines](https://dl.illwieckz.net/b/daemon/bugs/boggy-lines/unvanquished_2020-02-09_123041_000.png)

I then thought: _“wait, those floats are passed from c++ to glsl, those floats are just the quotient of 1 by ints, we can pass them as int instead and translate multiplication by division”_, I tried, it seems to work.

So, this is expected to fix https://github.com/Unvanquished/Unvanquished/issues/858

On my side it does:

[no more boggy lines](https://dl.illwieckz.net/b/daemon/bugs/boggy-lines/unvanquished_2020-02-09_140119_000.png)

### DIFF
--- a/src/engine/renderer/glsl_source/contrast_fp.glsl
+++ b/src/engine/renderer/glsl_source/contrast_fp.glsl
@@ -42,14 +42,13 @@ vec4 f(vec4 color) {
 
 void	main()
 {
-	vec2 scale = r_FBufScale;
-	vec2 st = gl_FragCoord.st;
-
 	// calculate the screen texcoord in the 0.0 to 1.0 range
-	st *= r_FBufScale;
+	vec2 st = gl_FragCoord.st / r_FBufSize;
 
 	// multiply with 4 because the FBO is only 1/4th of the screen resolution
 	st *= vec2(4.0, 4.0);
+
+	vec2 scale = 1 / r_FBufSize;
 
 	// perform a box filter for the downsample
 	vec4 color = f(texture2D(u_ColorMap, st + vec2(-1.0, -1.0) * scale));
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -722,10 +722,7 @@ std::string     GLShaderManager::BuildGPUShaderText( Str::StringRef mainShaderNa
 	AddDefine( env, "MAX_REF_LIGHTS", MAX_REF_LIGHTS );
 	AddDefine( env, "TILE_SIZE", TILE_SIZE );
 
-	float fbufWidthScale = 1.0f / glConfig.vidWidth;
-	float fbufHeightScale = 1.0f / glConfig.vidHeight;
-
-	AddDefine( env, "r_FBufScale", fbufWidthScale, fbufHeightScale );
+	AddDefine( env, "r_FBufSize", glConfig.vidWidth, glConfig.vidHeight );
 
 	AddDefine( env, "r_tileStep", glState.tileStep[ 0 ], glState.tileStep[ 1 ] );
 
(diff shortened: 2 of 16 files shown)


illwieckz:
@DolceTriade the bug was almost 15 years old:

https://sourceforge.net/p/xreal/svn/462

DolceTriade:
Ha! Nice find!

src/engine/renderer/glsl_source/contrast_fp.glsl
```diff
@@ -42,15 +42,14 @@ vec4 f(vec4 color) {
...
-	vec2 st = gl_FragCoord.st;
-
 	// calculate the screen texcoord in the 0.0 to 1.0 range
-	st *= r_FBufScale;
+	vec2 st = gl_FragCoord.st / r_FBufSize;
 
 	// multiply with 4 because the FBO is only 1/4th of the screen resolution
 	st *= vec2(4.0, 4.0);
 
+	vec2 scale = 1 / r_FBufSize;
```
slipher: unused?

slipher:
How can this be so important? The max error of a float is 30 parts per billion. It's hard to see how this could make a difference.

illwieckz:
@slipher, the big difference I see is that when this float is set as a `define`, it is not copied from CPU memory to GPU memory, but stringified by some `Str::Format` then parsed by the glsl renderer. There is high chance to see a truncation there.

In this case the bug was probably hidden until 2008, when those floats were converted from uniforms to defines:

https://sourceforge.net/p/xreal/svn/1963
> `Tue Jan 8 17:26:35 2008 +0000`
> u_FBufScale -> r_FBufScale
> u_NPOTScale -> r_NPOTScale

Also, `1/1024` is rational (`.0009765625 × 1024 = 1`) so vertical line may not occur on `1024×768` screen (very common resolution in 2005, still prevalent in 2008) while `1/1920` is not (`.00052083333333333333 × 1920 ≈ .99999999999999999360`). And in any way, the shortest `Str::Format` truncates, the faster the bug may occur.

src/engine/renderer/glsl_source/contrast_fp.glsl
illwieckz: It is used from line `54` and beyond. Don't worry I did the mistake first, my eyes first failed to notice them until the compiler yelled at me because I had deleted it in my first attempt.

illwieckz:
This also fixed the heatHaze edge bug:

[heatHaze edge bug](https://dl.illwieckz.net/b/daemon/bugs/boggy-lines/unvanquished_2020-02-10_094756_000.jpg)

This was a bug that I don't remember having noticed while playing at `1920×1080`, but this bug started to be noticeable at `2560×1440` and became very obvious and annoying at `3840×2160`.

illwieckz:
This also fixed the fog offset issue I talked about in https://github.com/Unvanquished/Unvanquished/issues/799#issuecomment-582376241 .

[etmain on Daemon](https://dl.illwieckz.net/b/unvanquished/bugs/etmain-regression/unvanquished_2020-02-05_124349_000.jpg)

[etmain on Daemon](https://dl.illwieckz.net/b/unvanquished/bugs/etmain-regression/unvanquished_2020-02-05_125443_000.jpg)

DolceTriade:
Dang, big impact fix!

PR #290, opened 2020-02-13
tr_shader: fix loose texture lighting, fix #289

### LINKED ISSUE #289: Wrong default colormap blend
The bug is not new, but was hidden by the bug fixed in #277 (_glsl/lightmapping: fix terrain alpha blending for collapsed materials_).

This bug was already there before in vertex lighting mode.

This is how it must look:

[wrong colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-11_005235_000.jpg)

This is now it looks now on light map lighting mode:

[wrong colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-11_002823_000.jpg)

This is how it already looked like on vertex lighting mode before #277:

[wrong colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-11_080750_000.jpg)

This is how it looks like on vertex lighting mode if I apply the same buggy workaround that was applied in lightmap lighting mode before #277 

[wrong colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-11_121039_000.jpg)

Don't mind the bizarre artifacts for now, that's another bug I have not yet investigated, but definitely another bug (looks to have been introduced between `0.50.0` and `0.51.1`).

The bug we talk about always has been there, this is how the same map looked like on Unvanquished 0.6.0 and 0.50.0 in vertex lighting mode. That was already too much darky, but we also see our engine is now more darky with newer artifacts that was not there at first, that's probably another yet unknown bug:

[wrong colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-11_115727_000.jpg)

[wrong colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-11_113839_000.jpg)

So the bug fixed in lightmap glsl in #277 looks to have been a workaround for an old bug living in others glsl shaders.

This bug only occurs on some legacy materials. It's related to the various guesses the engine does to detect the right color blend to apply, the problem probably living in that test:

https://github.com/DaemonEngine/Daemon/blob/94823acb06899fb7a5d33214b97742630712e4da/src/engine/renderer/tr_shader.cpp#L2984-L2995

If `stage->rgbGen` is always set to `colorGen_t::CGEN_IDENTITY`, both lightmap lighting and vertex lighting looks ok (but then it breaks some other textures that legitimately relied on this test to set the other type).

illwieckz:
Before (light mapping/vertex lighting):

[wrong colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-13_101512_000.jpg)

[wrong colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-13_101544_000.jpg)

After (light mapping/vertex lighting):

[correct colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-13_101612_000.jpg)

[correct colormap blend](https://dl.illwieckz.net/b/daemon/bugs/wrong-colormap-blend/unvanquished_2020-02-13_101644_000.jpg)

Don't mind the weird artifacts visible with vertex lighting, they come from another issue that is not yet tracked down (introduced between `0.50.0` and `0.51.1`).

For some reason the loose textures were set to `DIFFUSEMAP` stageType (that seems correct) while the colorGen was set to `CGEN_VERTEX` (that seems not correct):

https://github.com/DaemonEngine/Daemon/blob/390a2432a748a29a622c9ed3d2796aa0b8d53399/src/engine/renderer/tr_shader.cpp#L5464-L5470

That seems not correct because the rest of the code sets and expects `CGEN_IDENTITY` colorGen for textures with `DIFFUSEMAP` stageType:

https://github.com/DaemonEngine/Daemon/blob/390a2432a748a29a622c9ed3d2796aa0b8d53399/src/engine/renderer/tr_shader.cpp#L1530-L1536

The issue was reproduced on 0.6.0 release and the wrong colorGen was seen in the very first commit of the repository:

https://github.com/DaemonEngine/Daemon/blob/eb73a11506d0cf83cdcf26aa7cc2e26c1f40b5bf/src/engine/rendererGL/tr_shader.c#L5558-L5564

The bug was not visible before because the diffuse map lighting code had blending disabled (that was disabling alpha blending too), the bug was uncovered when alpha blending was enabled for multitexture materials (they use diffuse map lighting code), see #277 (_glsl/lightmapping: fix terrain alpha blending for collapsed materials_).

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5467,7 +5467,7 @@ shader_t       *R_FindShader( const char *name, shaderType_t type,
 				stages[ 0 ].type = stageType_t::ST_DIFFUSEMAP;
 				stages[ 0 ].bundle[ 0 ].image[ 0 ] = image;
 				stages[ 0 ].active = true;
-				stages[ 0 ].rgbGen = colorGen_t::CGEN_VERTEX;
+				stages[ 0 ].rgbGen = colorGen_t::CGEN_IDENTITY;
 				stages[ 0 ].stateBits = implicitStateBits;
 				break;
 			}


illwieckz:
Related line was introduced 13 years ago, there is high chance the code was already faulty (but the bug was probably already hidden for similar reasons):

https://sourceforge.net/p/xreal/svn/664/
> changed lightmap handling
> `Date:   Mon Feb 13 14:50:20 2006 +0000`

It's fun to see the faulty line was introduced on `2006-02-13` and fixed on `2020-02-13`, for its 14th birthday! :birthday:

illwieckz:
Tremulous 1.1.0 was [released on 30th of May 2006](http://tremulous.net/news/page_4/#news_6), I probably played Tremulous [for the first time on 4th of April 2006](https://linuxfr.org/news/tremulous-version-110). So the bug I just fixed in that Tremulous child that is Unvanquished was introduced before I played Tremulous for the first time.

Any way, Tremulous did not have this bug, neither Unvanquished since the bug was dormant because of the bug is living in a feature Tremulous did not have and Unvanquished/Dæmon never enabled before, but that's crazy to think I just fixed a bug that predates the day I played Tremulous for the first time!

slipher:
LGTM - unless someone who understands this stuff wants to comment, since I don't.

illwieckz:
I'm not sure to understand everything in that stuff but I understand this:

> The loose textures were set to `DIFFUSEMAP` stageType with `CGEN_VERTEX` colorGen while the rest of the code sets and expects `CGEN_IDENTITY` colorGen for textures with `DIFFUSEMAP` stageType.

So the proposed fix looks to fulfill the need for a consistent rendering of diffuse maps in a way they are all wrongly or correctly rendered. We can then expect them to be rendered the same way either they are described in material files or not. Either the default rendering is wrong or not. it would be done the same.

About “is it the right way to render this” question we can see that those textures (both loose textures and ones described in materials) are now rendered the same way it was rendered before alpha blending feature was enabled on material-described diffuse maps so by enabling the feature and implementing this fix we added an optional feature for material-described diffuse maps without regression for material-described diffuse cases not using the feature, and without regression for non-material-described diffuse maps using the feature or not.

PR #294, opened 2020-02-26
WIP: computeLight: rewriting diffuse alpha channel looks suspicious

illwieckz:
Light accumulation was done on all color components including alpha channel, that can't be good.

---
This change is [Reviewable](https://reviewable.io/reviews/daemonengine/daemon/294)

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -189,7 +189,6 @@ void computeDLights( vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse, vec4 mater
         if (numLights > 0.0)
         {
           color.rgb = vec3(numLights/(lightsPerLayer*numLayers));
-          color.a = 1.0;
         }
 #endif
         return;
@@ -207,7 +206,6 @@ void computeDLights( vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse, vec4 mater
   if (numLights > 0.0)
   {
     color.rgb = vec3(numLights/(lightsPerLayer*numLayers));
-    color.a = 1.0;
   }
 #endif
 }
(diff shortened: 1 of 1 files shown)


DolceTriade:
Does this fix anything?

illwieckz:
No, I haven't seen something being broken because of that (or I was not aware of this), but that bug may be uncovered later when adding features or reorganizing code.

slipher:
This is just debugging code, right?

illwieckz:
Hmmm, I feel stupid, the alpha channel is not rewritten with light computation but with 1.0, which does not seems legit by the way, since 1.0 alpha channel is fully transparent.

And yes, it's debug code, I don't know how I managed to miss that. But anyway, setting a texture to be fully transparent to show the rgb channels seems wrong. For some reason, I'm not able to see the bug I expect.

This is the code _before_, this is not what I expect the code to render (and the glass displaying light color on borders if alpha channel is fully transparent?):

[debug alpha light](https://dl.illwieckz.net/b/daemon/bugs/debug-alpha-light/unvanquished_2020-02-26_134907_000.jpg)

This is the code _after_, which is what the code claims to do when there is no alpha channel rewrite:

[debug alpha light](https://dl.illwieckz.net/b/daemon/bugs/debug-alpha-light/unvanquished_2020-02-26_135148_000.jpg)

We may want to **not** blend the debug light color by setting it fully opaque (`0` alpha channel)

This is the code when there is alpha channel rewrite to 0.0 (fully opaque), so at least we know the alpha channel is read:

[debug alpha light](https://dl.illwieckz.net/b/daemon/bugs/debug-alpha-light/unvanquished_2020-02-26_142547_000.jpg)

So, what to do, keep the legacy behavior (set alpha channel to 1), keep the alpha channel unmodified?, set the alpha channel to 0 (fully opaque)?

Note that in later case, the light color is not displayed the same, maybe it's not what we want.

In all case I plan to merge the very first commit that reword stuff to make it more readable.

VReaperV:
Alpha value 1.0 is opaque, not transparent. We would also get incorrect results on semi-transparent surfaces otherwise, because we don't want that affecting how light tiles are shown.

PR #297, opened 2020-02-29
Warnings

DolceTriade:
Allow the engine to build with -Werror.


CMakeLists.txt
```diff
@@ -25,8 +25,8 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 cmake_minimum_required (VERSION 2.8.12)
-if (POLICY CMP0017)
```
slipher: Why remove this one?

src/engine/qcommon/parse.cpp
```diff
@@ -2971,8 +2971,8 @@ static int Parse_Directive_include( source_t *source )
 		{
 			// buffer too small?
 			path[ MAX_QPATH - 1 ] = 0;
-			strncpy( path, source->includepath, MAX_QPATH - 1 );
```
slipher: I don't think you should subtract 1 from the buffer size with the new functions. It doesn't really matter though, since on the 0.52.0 branch I deleted this file (and moved it into the cgame).

illwieckz:
@DolceTriade can you rebase on `0.52.0/sync`?

slipher:
I made a couple of edits:
- Restore the mistakenly deleted CMP0017 policy
- Drop the string functions commit since it is in a file deleted in 0.52 whic will cause a merge conflict.

Issue #299, opened 2020-03-09
deluxemap overflow


illwieckz:
Reported there before: UnvanquishedAssets/map-station15_src.dpkdir#3

I'm now sure it's an engine issue, and I now know the line causing this:

```glsl
	lightColor /= clamp(dot(normalize(var_Normal), lightDir), 0.004, 1.0);
```

There:

https://github.com/DaemonEngine/Daemon/blob/42e408a4f45f6e78b59fe9b7022f57456a7677d6/src/engine/renderer/glsl_source/lightMapping_fp.glsl#L91-L92

We can find trace of this code in first commit of the repository:

https://github.com/DaemonEngine/Daemon/blob/eb73a11506d0cf83cdcf26aa7cc2e26c1f40b5bf/main/glsl/lightMapping_fp.glsl#L165

It was removed in 0bb3550ce16ab7b165a88ebb365edf6b2837d1f5 (_Add tiled light-index renderer_) and re-added in 0949fdad282b42c90e5f444199629c32b7d49d2e (_Lightmapping fixes_) by @gimhael.

I don't know what this code is for, it looks like there is less issues by not using this code than using it.

illwieckz:
If I comment out this code, the artifacts disappear:

Before / after:

[![deluxemap overflow](https://dl.illwieckz.net/b/daemon/bugs/deluxemap-overflow/unvanquished_2020-03-09_200404_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/deluxemap-overflow/unvanquished_2020-03-09_200404_000.jpg)

[![deluxemap overflow](https://dl.illwieckz.net/b/daemon/bugs/deluxemap-overflow/unvanquished_2020-03-09_193523_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/deluxemap-overflow/unvanquished_2020-03-09_193523_000.jpg)

Before / after:

[![deluxemap overflow](https://dl.illwieckz.net/b/daemon/bugs/deluxemap-overflow/unvanquished_2020-03-09_200615_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/deluxemap-overflow/unvanquished_2020-03-09_200615_000.jpg)

[...]

PR #301, opened 2020-03-09
renderer: merge lightmap and vertex (lightgrid) world and entity lighting

illwieckz:
This merges world lightmapping glsl, world lightgrid lighting glsl (wrongly called vertex world) and entity lightgrid glsl (wrongly called vertex entity).

This prevent the risk of mistakes by not copy pasting 90% of the same code between files. This also helps to improve feature parity between all rendering options.

It makes possible to do or to implement:

- normal mapping on maps without deluxe maps (useful to improve legacy maps)
- deluxe mapping disablement to rely on lightgrid direction instead (useful for debugging purposes, this may also use less gpu memory)
- render some world part with lightmap and other world part with lightgrid like moving platforms to avoid moving shadows and to enable ambient lighting on them instead.
- lightgrid lighting with deluxe map-based normal lighting (I don't see any usage, but it would work)
- do reflective specular on world (what is it?)

Also it would be possible to implement rim lighting on world but I don't think there is any usage for that, in any way it would be very easy to do.

The implementation is rendering properly world and entities using either light maps or light grid, and normal maps are rendererd properly using either deluxe maps or light grid.


illwieckz:
Once this is merged, merging cube reflection within diffuse lighting stage would only require to edit one glsl code instead of three, same for any future feature we would want to add.

Maybe one day we would be able to also merge the liquid code.

illwieckz:
The “invisible surface” viewing angle-based bug is probably not in GLSL code.

The bug is probably not in `lightMapping_fp.glsl` file (fragment program) since replacing the whole `main` by a simple 3-èline code displaying the diffuse texture is reproducing the bug. There is very low chance the bug would be in `lightMapping_vp.glsl` file (vertex program) since this file is small and simple.

On C++ side, there is low chance the bug would be in `gl_shader.cpp` or `gl_shader.h` (code is not very complex). So there is high chance this code is in `tr_shade.cpp`.

illwieckz:
One interesting thing is that the code stores lighgrid1 (used to compute light color) as lightmap when lightmap is disabled or unavailable, and stores lightgrid2 (used to compute light direction) as deluxemap when deluxe map is disabled or unavailable. This way there is only two binds for light color and light direction whatever the combination of features used. Then some `#ifdef` enable the related code in GLSL.

illwieckz:
PR is now ready, #309 was requirement to make it works.

There is now a single function in `tr_shade.cpp` and glsl shader for lightmapped world, grid-lit world, and grid-lit entities. They share now the same feature, and they work the same or fail the same.

Those scenarii are now possible:

1. lightmap-based light mapping and deluxemap based normal mapping,
the usual scenario with official map;
2. lightmap-based light mapping and lightgrid based normal mapping  
would allow use to add normal map to already baked legacy maps without deluxe maps;
3. lightgrid-based light mapping and deluxemap based normal mapping
this is a curiosity made possible for free;
4. lightgrid-based light mapping and lightgrid based normal mapping  
for when lightmapping is disabled, also used by entities.

At this point, it is expected to reproduce existing bugs. Scenario 2. currently looks wrong. The fix seems to be easy: apply on lightgrid color used for normal mapping the same tweak we apply on lightmap and discussed in #299 and #306 fixes it, but then entity lighting may be wrong, at least not what we are accustomed to. It looks like light grid lighting is wrong and too much bright (see also #313).

Note: the fact the same code is used for entities and map models would allow us to avoid the bug of moving models with moving shadows by lighting moving models with lightgrid, it would also mean moving map models would have their lighting changing with ambient light, which would be very cool.

illwieckz:
This PR deletes almost 680 lines, does not remove features, allow more feature combinations, make new features possible, make the code more tested, will make maintenance easier, and would prevent some bugs to be introduced or fix forgotten by oversight.

illwieckz:
I added a commit to fix the lighting on the “lightmap light color but lightgrid light direction” scenario. I still thing there is an issue with lightgrid light color, but it's possible to fix the first without modifying the second behaving like it was before.

illwieckz:
So, the current status:

1. lightmap-based light mapping and deluxemap based normal mapping,
the usual scenario with official map; **OK**
2. lightmap-based light mapping and lightgrid based normal mapping,
would allow use to add normal map to already baked legacy maps without deluxe maps; **OK**
3. lightgrid-based light mapping and deluxemap based normal mapping,
this is a curiosity made possible for free; **OK**
4. lightgrid-based light mapping and lightgrid based normal mapping,
for when lightmapping is disabled, also used by entities. **NOT OK _as usual_**

Fix for 4. is out of topic. At least this PR does not change legacy behaviour.

illwieckz:
Lightmap-based light color with deluxemap-based light direction:

[glslmerge](https://dl.illwieckz.net/b/daemon/wip/glslmerge/unvanquished_2020-04-16_195001_000.jpg)

Lightmap-based light color with lightgrid-based light direction:

[glslmerge](https://dl.illwieckz.net/b/daemon/wip/glslmerge/unvanquished_2020-04-16_195113_000.jpg)

Lightgrid-based light color with deluxemap-based light direction:

[glslmerge](https://dl.illwieckz.net/b/daemon/wip/glslmerge/unvanquished_2020-04-16_195200_000.jpg)

Lightgrid-based light color with lightgrid-based light direction:

[glslmerge](https://dl.illwieckz.net/b/daemon/wip/glslmerge/unvanquished_2020-04-16_195256_000.jpg)

This last one does not looks _that bad_, but it's a bit too much bright, that can be seen especially on the building. See #313.

src/engine/renderer/tr_local.h
```diff
@@ -508,6 +508,21 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 		vertexHash_t *next;
 	};
 
+	enum
+	{
```
slipher, line 512: Why `B_`?

illwieckz: For bind, was lacking imagination.

illwieckz: can also use `BIND_`

slipher:
`BIND_` sounds good. Feel free to go ahead. As you can see, the PR is way too complicated for me to understand, so all I could do is bikeshed the enum naming.

src.cmake
```diff
@@ -194,6 +188,8 @@ set(GLSLSOURCELIST
     ${ENGINE_DIR}/renderer/glsl_source/refraction_C_fp.glsl
     ${ENGINE_DIR}/renderer/glsl_source/refraction_C_vp.glsl
     ${ENGINE_DIR}/renderer/glsl_source/reliefMapping_fp.glsl
+    ${ENGINE_DIR}/renderer/glsl_source/lightMapping_fp.glsl
```
Kangz, line 191: u-nit: this breaks alphabetical order

src/engine/renderer/tr_local.h
Kangz: The enum is only defined in gl_shader.cpp to number texture units so maybe it could be defined locally there?

src.cmake
illwieckz: right

src/engine/renderer/tr_local.h
illwieckz: okok

illwieckz: Hmm, no, it's used in both `tr_shade.cpp` and `gl_shader.cpp` (the first one sets, the second one reads).

illwieckz:
I currently disabled the feature to do normal/specular mapping on bsp surfaces because of #324.

It would be cool to fix #324 to revert this change, even if we don't care about that feature: this workaround is very intrusive and I would like to get rid of it.

This workaround makes sure those surfaces are rendered before and after the merge is done.

illwieckz:
I also noticed current master fixed #326 (_blend issue with internal lightmap_), probably thanks to #309 (_tell a surface belongs to the bsp, properly select the renderer_). But the light map/light grid code merge re-introduced a variant of it.

So that was likely an issue related to the fact that when the code for grid lighting and light map lighting is merged, we have to take care of the fact the light grid must not be used if there is no lightmap but light mapping is disabled. Because of the design, no lightmap use lightmap code but with a white image.

So I added a commit to:

- enable lightmapping with whiteimage on nolightmap shader, fix #326;
- enable lightmapping with whiteimage when there is no lightmap neither lightgrid;
- enable deluxemapping with blackimage when there is no deluxemap;
- rework the code and functions around those features.

illwieckz:
Let's do it.

PR #303, opened 2020-03-29
Static analyzer (PVS Studio) fixes

dimhotepus:

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -1571,7 +1571,7 @@ Cmd::CompletionResult CompleteFilename(Str::StringRef prefix, Str::StringRef roo
 		if (!allowSubdirs)
 			return {};
 		prefixDir = prefix;
-		prefixBase = "";
+		prefixBase.clear();
 	} else {
 		prefixDir = Path::DirName(prefix);
 		if (!allowSubdirs && !prefixDir.empty())
@@ -2056,7 +2056,7 @@ Cmd::CompletionResult CompleteFilename(Str::StringRef prefix, Str::StringRef roo
 		if (!allowSubdirs)
 			return {};
 		prefixDir = prefix;
-		prefixBase = "";
+		prefixBase.clear();
 	} else {
 		prefixDir = Path::DirName(prefix);
 		if (!allowSubdirs && !prefixDir.empty())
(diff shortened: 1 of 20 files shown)


src/common/FileSystem.cpp
```diff
@@ -1432,13 +1434,19 @@ void CopyFile(Str::StringRef path, const File& dest, std::error_code& err)
 		char buffer[65536];
 		while (true) {
 			offset_t read = zipFile.ReadFile(buffer, sizeof(buffer), err);
-			if (err)
+			if (err) {
+				std::error_code ignored;
+				zipFile.CloseFile(ignored);
```
slipher: So there's this wacky multi-mode error handling thing in these functions... 
If `&err` equals a magic value, then errors are reported by throwing exceptions, instead of by setting the error code. So CloseFile will not get called in exception mode.

I'd still accept the PR which is an improvement over the current state. Maybe add a TODO about the fact that it doesn't work when using exceptions.

Probably we should get rid of this double paradigm error stuff because it is too hard to write correctly.

src/engine/framework/VirtualMachine.cpp
```diff
@@ -370,7 +370,7 @@ uint32_t VMBase::Create()
 {
 	type = static_cast<vmType_t>(params.vmType.Get());
 
-	if (type < 0 || type >= TYPE_END)
+	if (type < TYPE_NACL || type >= TYPE_END)
```
slipher: I find this change makes the intent less clear. How about adding a TYPE_BEGIN alias for the first enumerator?

slipher:
Thanks for the fixes. I think I ran into that first [overrun] one recently while working on WASM support.

src/common/IPC/CommandBuffer.cpp
```diff
@@ -34,7 +34,7 @@ namespace IPC {
 
     void CommandBuffer::Init(void* memory, size_t size) {
         if (size < DATA_OFFSET) {
-            Sys::Drop("Buffer is too small, size is %i", size);
+            Sys::Drop("Buffer is too small, size is %u", (unsigned)size);
```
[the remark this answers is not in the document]
dimhotepus: Restored to original, thanks!

src/common/FileSystem.cpp
dimhotepus: Done.

src/engine/framework/VirtualMachine.cpp
dimhotepus: Done.

Can you, please, review fixes and added commits?

illwieckz:
Thank you @dimhotepus! That's very precious. :+1:
I remember [Daemon#24](https://github.com/DaemonEngine/Daemon/pull/24), [Unvanquished#1013](https://github.com/Unvanquished/Unvanquished/pull/1013) and [Unvanquished#1014](https://github.com/Unvanquished/Unvanquished/pull/1014), :wink:. Good job! That's nice to see you around. :smiley:

Issue #274, opened 2020-01-24
Switch to "OpenGL" normalmap direction?


slipher:
This is a thread to decide whether we should keep the default interpretation of normalmaps the way it is now (so-called DirectX style), or switch to OpenGL style, which interprets the y-axis (green channel) as pointing the opposite direction. Note that the direction chosen doesn't really have any relation to the rendering library you use - it's up to the GPU shader code to decide how to interpret things.

Advantages of switching to OpenGL direction:
- More intuitive: y-component points up, as expected, and the coordinate system is right-handed
- Matches most OpenGL normal mapping tutorials
- Improves compatibility with Xonotic assets

Advantages of staying on DirectX:
- Our assets continue to work unchanged - we don't have to add q3shaders to reverse the direction of every existing normalmap
- Maybe it's more common among games in general? (unclear)

illwieckz:
About the engine, the internal representation in glsl shaders can be switched to the OpenGL representation or any format we want from now. The issue is about the default format expected from a material described in a file stored as `scripts/<somename>.shader` and described using the syntax we use.

Maps from games using another way to describe materials, another material file naming scheme or another material file syntax would allow us to import their normal map the expected way whatever the internal format we choose and the format the use.

This distinction is very important, because it means we can have an engine having an OpenGL-like internal normal representation, even if we expect our `scripts/<somename>.shader` written with our syntax to describe DirectX-like normal maps.

For other material `.shader`-like formats, it does not prevent us to load normal maps described in hypothetical `materials/<somename>.mtr` the DirectX way and normal maps described in hypothetical `<somepath>/<somename>.gltf` file the OpenGL way.

DarkPlaces does not use `.shader` files to describe normal maps. This engine uses suffixes. So we are free to use any format we want internally, we would be able to read DarkPlaces's files the expected way. The big issue would come if one day we want to add support for assets from another engine which uses the same technique an relies on the same suffix but expects another format.

The biggest issue I see is that other idTech 3 engines may use DirectX normal map format by default and they may use the same syntax we use… If that's true, it would be sad but we may have to accept they chose for us.

I'm doing some investigations.

---

[...]

PR #307, opened 2020-04-07
renderer/tr_shader: introduce engine-agnostic normalFormat material stage keyword

illwieckz:
Introduce engine-agnostic `normalFormat` material stage keyword.

There is already a `normalScale` keyword which allows to revert channel by using a negative value, but this is not a way to describe the _format_ of the file that is loaded. So, people would have to use a negative `normalScale` Y value in their `.shader` file if the normal map file is stored in DirectX way and the engine is computing normal maps the OpenGL way, but a positive Y `normalScale` value if file format and engine format matches.

By introducing the `normalFormat` stage keyword, artist can describe the normal map file format and don't mind about what the engine does. It also means the engine may switch its internal format without having to mind anything, the coder implementing another normal format would know how to convert the normal map file in engine whatever the format, not requiring him to ask artists to edit every existing `.shader` material file.

It was possible to make `normalScale` keyword engine-agnostic but that would be a very bad idea because this keyword exist in other engine of this lineage (idtech3 renderer2, OpenJK…, etc.). Then it's better to keep this keyword behaviour the way it works the same as other engine does.

For making things easier to think of, the engine is currently assuming a default normal map format is OpenGL (+X +Y +Z), because it's easier to multiply every channel with `1.0` by default whatever the channel when there is nothing to do.

For backward-compatibility purposes and to keep compatibility with other engines (idtech3 renderer2, XreaL, ET:Legacy, OpenJK, etc.), the `normalMap`, `normalHeightMap` keywords and `bumpMap` deprecated alias set the normal map format to DirectX one. Historically, the Doom 3 engine is an OpenGL engine using DirectX normal map format convention, and XreaL implemented normal maps in order to be able to load Doom3 formats, then expected DirectX normal maps. Engines like those I just quoted are reusing XreaL material stage keywords like `normalMap` that were Doom3 keywords at first, then are expecting DirectX format. It looks like we would have to live with it.

While the `normalMap` keyword is expecting normal map in the DirectX format because of backward compatibility with historical engines and third-party ones, Dæmon-based games developers are free to encourage the use of normal map using OpenGL format, maybe use contribution guidelines asking for normal map using OpenGL format explicitely. In such case contributors would just have to add the required line to tell the normal map is using OpenGL format.

The material stage keyword syntax is:

OpenGL format:

```
normalMap /textures/castle/brick_n
normalFormat X Y Z
```

DirectX format (`normalFormat` keyword is not required since `nomalMap` keyword already does it):

```
normalMap /textures/castle/brick_n
normalFormat X -Y Z
```

Weird format (like the one used in Vega texture set):

```
normalMap /textures/castle/brick_n
normalFormat -X -Y Z
```

With this work, we may one day change the normal map format used in GLSL code (see #274) without having to mind anything: the material parser is telling the engine which conversion to do, and the assets would not have to be edited if they use the `normalFormat` keyword.

The Vega map and texture set is expected to use the `normalFormat` keyword to fix normal map orientation in upcoming Unvanquished 0.52.

With such design we would be able to add support for alternative material syntax from other games to load normal maps, in such case, the keyword parsing code would just have to set the normal map format this syntax expects. The Darkplaces compatibility layer already does it this way.

Here are the default normal map formats used in engine:

- C++ engine: OpenGL
- XreaL `normalMap` keyword: DirectX
- Darkplaces `_norm` sidecar autoloading: OpenGL
- GLSL renderer as a black box: OpenGL
- GLSL code itself: DirectX (a hack reverts the Y channel on input, we may edit the code to avoid the Y reversion hack)

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -579,37 +579,34 @@ void Tess_Begin( void ( *stageIteratorFunc )(),
 	}
 }
 
-/*
-==============
-SetNormalScale
-==============
-*/
-void SetNormalScale( shaderStage_t *pStage, vec3_t normalScale)
+void SetNormalScale( shaderStage_t *pStage, vec3_t normalScale )
 {
-		float normalIntensity = RB_EvalExpression( &pStage->normalIntensityExp, 1 );
+	float normalIntensity = RB_EvalExpression( &pStage->normalIntensityExp, 1.0 );
 
-		normalScale[ 0 ] = 1;
-		normalScale[ 1 ] = 1;
-		normalScale[ 2 ] = 1;
+	for ( int i = 0; i < 3; i++ )
+	{
+		normalScale[ i ] = pStage->normalScale[ i ];
 
-		normalScale[ 0 ] *= normalIntensity;
-		normalScale[ 1 ] *= normalIntensity;
-
-		if ( pStage->hasNormalScale )
+		// Normal intensity is only applied on X and Y.
+		// This behaviour is inherited.
+		if ( i < 2 )
 		{
-			normalScale[ 0 ] *= pStage->normalScale[ 0 ];
-			normalScale[ 1 ] *= pStage->normalScale[ 1 ];
-			normalScale[ 2 ] *= pStage->normalScale[ 2 ];
+			normalScale[ i ] *= normalIntensity;
 		}
+	}
 
-		// NOTE: normalmap with zero Z component is wrong
-		//
-		// since glsl disables normalScale when Z is zero
-		// setting this to zero has side effect to reset
-		// per-stage normalScale
-		//
-		// there is no intention to keep this undefined feature
-		normalScale[ 2 ] *= r_normalScale->value;
+	/* Note: the GLSL code disables normal map scaling when normal Z scale is
+	equal to zero.
+
+	It means normal map scaling is disabled when r_normalScale is set to zero.
+	This is cool enough to be kept as a feature.
+
+	Normal Z component equal to zero would be wrong anyway.
+
+	r_normalScale is only applied on Z.
+	This behaviour is inherited.
+	*/
+	normalScale[ 2 ] *= r_normalScale->value;
 }
 
 // *INDENT-ON*
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -49,6 +49,27 @@ static cullType_t    implicitCullType;
 
 static char          whenTokens[ MAX_STRING_CHARS ];
 
+/* This parser does not assume any default normal map format.
+
+Syntax variations must define a default normal map format.
+
+Stage keywords like normalMap, bumpMap and normalHeightMap set
+normal format to +X -Y +Z like DirectX, to keep those keywords
+compatible with other idTech3 engines using those keywords.
+
+XreaL implemented normalMap like Doom3 which is an OpenGL engine
+using DirectX normal format convention. Then other renderers like
+Daemon, ioquake3 renderer2, OpenJK, ET:Legacy, etc. did like XreaL.
+
+It's possible to extend this parser with other stage keywords
+from other engines to load normal map with their respective format.
+
+The normalFormat stage keyword can be used in materials to set an
+arbitrary format.
+*/
+static const vec3_t glNormalFormat = {  1.0,  1.0,  1.0 };
+static const vec3_t dxNormalFormat = {  1.0, -1.0,  1.0 };
+
 // DarkPlaces material compatibility
 static Cvar::Cvar<bool> r_dpMaterial("r_dpMaterial", "Enable DarkPlaces material compatibility", Cvar::LATCH, false);
 Cvar::Cvar<bool> r_dpBlend("r_dpBlend", "Enable DarkPlaces blend compatibility, process GT0 as GE128", Cvar::NONE, false);
@@ -1679,6 +1700,54 @@ static void ParseLightFalloffImage( shaderStage_t *stage, const char **text )
 	}
 }
 
+/* SetNormalFormat: set normal format for given stage if normal format
+is not already set or it must be overwritten.
+
+Set force option to true to overwrite existing normalFormat.
+
+The idea is to call SetNormalFormat with force set to true when called by normalFormat keyword
+so existing per-keyword format is overwritten, but keyword like normalMap does not set force
+to true so their per-keyword format is only set if there is no normal format already set.
+
+This ensures those two materials are rendered the same:
+
+```
+textures/castle/brick
+{
+		diffuseMap textures/castle/brick_d
+		// set normalMap-specific normal format
+		normalMap textures/castle/brick_n
+		// overwrite with custom normal format
+		normalFormat X Y Z
+}
+```
+
+```
+textures/castle/brick
+{
+		diffuseMap textures/castle/brick_d
+		// set custom normal format
+		normalFormat X Y Z
+		// do not set normalMap-specific normal format
+		//keep existing custom one
+		normalMap textures/castle/brick_n
+}
+```
+*/
+
+void SetNormalFormat( shaderStage_t *stage, const vec3_t normalFormat, bool force = false )
+{
+	if ( !stage->hasNormalFormat || force )
+	{
+		stage->hasNormalFormat = true;
+
+		for ( int i = 0; i < 3; i++ )
+		{
+			stage->normalFormat[ i ] = normalFormat[ i ];
+		}
+	}
+}
+
 struct extraMapParser_t
 {
 	const char *suffix;
@@ -1771,7 +1840,7 @@ void LoadExtraMaps( shaderStage_t *stage, const char *colorMapName )
 					diffuseMap  textures/map_solarium/water4/water4
 					normalMap   textures/map_solarium/water4/water4_norm
 					specularMap textures/map_solarium/water4/water4_gloss
-					normalScale 1 -1 1
+					normalFormat +X +Y +Z
 					tcmod scale  0.3  0.4
 					tcMod scroll 0.05 0.05
 					blendfunc add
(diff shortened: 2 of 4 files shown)


illwieckz:
This implementation or another one is required for Unvanquished 0.52 because Vega normal maps have to be fixed in a way we don't have to edit materials in the future again. Editing the pictures themselves must be avoided to not uselessly increase the git repository size.

src/engine/renderer/tr_shader.cpp
```diff
@@ -49,6 +49,24 @@ static cullType_t    implicitCullType;
...
+// XreaL implemented normalMap like Doom3 which is an OpenGL engine
+// using DirectX normal format convention. Then other renderers like
+// Daemon, ioquake3 renderer2, OpenJK, ET:Legacy, etc. did like XreaL.
+//
+// It's possible to extend this parser with other stage keywords
+// from other engines to load normal map with their respective format.
+//
+// The normalFormat stage keyword can be used in materials to set an
+// arbitrary format.
+static vec3_t glNormalFormat = {  1.0,  1.0,  1.0 };
```
slipher, line 67: Nit: `const`

src/engine/renderer/tr_shader.cpp
```diff
@@ -2895,26 +2929,79 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
...
+			// replace existing normal format in all case
+			stage->hasNormalFormat = true;
+
+			for ( int i = 0; i < 3; i++ )
+			{
+				token = COM_ParseExt2( text, false );
+
+				if ( token[ 0 ] == '+' )
+				{
+					Log::Warn("Found + normalFormat parm in shader '%s'", shader.name );
```
slipher, line 2967: Why is this a warning? Isn't that the expected input?

src/engine/renderer/tr_shade.cpp
```diff
@@ -583,32 +583,35 @@ void Tess_Begin( void ( *stageIteratorFunc )(),
...
 		{
-			normalScale[ 0 ] *= pStage->normalScale[ 0 ];
-			normalScale[ 1 ] *= pStage->normalScale[ 1 ];
-			normalScale[ 2 ] *= pStage->normalScale[ 2 ];
+			normalScale[ i ] = pStage->normalScale[ i ];
+		}
+		else
+		{
+			// assume OpenGL +X +Y +Z format if unknown
+			normalScale[ i ] = 1.0;
```
slipher, line 599: So this is reversing the default from what it was before?

src/engine/renderer/tr_shader.cpp
illwieckz: Ah, sorry, that was debug code to follow what I was doing, will remove it. :-)

src/engine/renderer/tr_shade.cpp
illwieckz: There was no "default" in such meaning, neither a “-1” neither a “1” was stored. Parser was just feeding files to the renderer without knowing anything about the file. By introducing format management, this introduce a default format that is “1, 1 ,1”.

The default for “normalMap” keyword is  “1, -1, 1”.

In a way, we can say it reverse the default because if an existing normal map is loaded without the “normalMap” keyword (option we don't have anyway), it would be rendered wrong.

I decided to use “1, 1, 1” as a default because it makes things meaningful, things are written in a way they are, not in a way they must be written to be rendered right. The alternative way would be to name “1, -1, 1” OpenGL, and “1, 1, 1” DirectX, which would be confusing and bite us hard in the back in the future.

illwieckz: We can say the default normal format for the material parser is OpenGL (something we did not have), but the default normal format in case of `normalMap` keyword is DirectX (like before).

Currently, this code is dead code, the xreal `normaMap` parser sets the DirectX format, the darkplaces parser sets the OpenGL format. We currently don't offer way to set normal maps without format being set. This code was thought to make it future proof but it's maybe a bad idea to fix potential issues silently, while the fix may be wrong.

I'm thinking about setting an UNREACHABLE assert there instead. This way we would ensure future implementation don't forget to set the format. But to do that we need a reliable way to know there is a normal map, and the code is not ready to do it easily (at this point we still have some materials storing normal map in diffuse slot). I'm not going to do the required refactorisation before that MR. This is planned, but this refactorization is not required for the release, while this normal format management is. I'll add a FIXME.

src/engine/renderer/tr_shader.cpp
illwieckz: done

src/engine/renderer/tr_shade.cpp
illwieckz: Added `FIXME`s.

illwieckz:
After some deep thoughts, this is the syntax:

```
normalFormat X Y Z
```

```
normalFormat X -Y Z
```

Even if I have no plan to implement it in engine for normal maps, this syntax allows to implement swizzling. I would like to implement swizzling for PBR maps (`physicalFormat R M O`, `physicalFormat O R M`, etc.) so this syntax would be consistent with other format keywords. We can also imagine `heightFormat -H` to tell _black is high_. So I feel confident with this syntax. The remaining alternative syntax I would not be against would be `XYZ` and `X-YZ` in one word.

slipher:
For files that provide only X and Y channels and infer Z, what does it mean to put -Z? Does it do anything?

illwieckz:
Hmm, Right! Z must only be modified with `normalFormat` if Z is loaded from file. But it still makes sense to modify Z with `normalScale` or `r_normalScale`. Hmm, that may make code complicated.

illwieckz:
Well, because of all those corner cases requiring a better source code to be targeted properly, I had to do the refactorization I was first hoping to be able to postpone for after the next release. So this PR is now built atop #317.

illwieckz:
New code will not apply custom format transformation on Z if Z is reconstructed from X and Y, but will still apply custom scale on Z.

illwieckz:
Just rebased.

illwieckz:
I forgot to tell it before, but that new code based on the recent factorization (#317) sets an `UNREACHABLE_ASSERT()` when there is no normal format set at the end of the processing of a material. This expects future code implementing new syntax sets a normal format explicitely, without assumption.

src/engine/renderer/tr_shader.cpp
```diff
@@ -1771,7 +1840,7 @@ void LoadExtraMaps( shaderStage_t *stage, const char *colorMapName )
 					diffuseMap  textures/map_solarium/water4/water4
 					normalMap   textures/map_solarium/water4/water4_norm
 					specularMap textures/map_solarium/water4/water4_gloss
-					normalScale 1 -1 1
+					normalFormat + + +
```
slipher, line 1843: Is this right? It looks like the letters are required.

src/engine/renderer/tr_shader.cpp
```diff
@@ -2895,26 +2969,83 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
...
+
+			SkipRestOfLine( text );
+			continue;
+		}
+		/* normalScale <float> <float> <float>
+
+		Scales normal map channels.
+
+		For compatibility purpose, ioquake3 renderer2 syntax is supported:
+		normalScale <float> <float>
```
slipher, line 3044: What is the intended effect of this when the z component is reconstructed?

src/engine/renderer/tr_shader.cpp
```diff
@@ -4652,6 +4792,37 @@ static void CollapseStages()
...
+
+				if ( !stage->hasNormalScale )
+				{
+					stage->normalScale[ i ] = 1.0;
+				}
+
+				// If height map is not stored in normal map alpha channel,
+				// the Z component will be reconstructed and will be fine
+				// from the start, so we must not translate the Z format.
+				if ( i < 2 || stage->isHeightMapInNormalMap )
```
slipher, line 4816: What is the idea of checking isHeightMapInNormalmap?

src/engine/renderer/tr_shade.cpp
```diff
@@ -579,37 +579,30 @@ void Tess_Begin( void ( *stageIteratorFunc )(),
...
 
-		// NOTE: normalmap with zero Z component is wrong
-		//
-		// since glsl disables normalScale when Z is zero
-		// setting this to zero has side effect to reset
-		// per-stage normalScale
-		//
-		// there is no intention to keep this undefined feature
-		normalScale[ 2 ] *= r_normalScale->value;
+	Normal·Z·component·equal·to·zero·would·be·wrong·anyway.
```
slipher, line 603: Is it just me or are there dots between the words

illwieckz: lol, yes, I did a copy-paste from the terminal the wrong way (my vim config prints whitespaces as dots). :see_no_evil:

src/engine/renderer/tr_shader.cpp
illwieckz: when z component is reconstructed, normalformat is not used but normalscaled is used, but here it's 1.0 and it's a factor, it will not revert anything because positive, and do nothing because 1.

src/engine/renderer/tr_shader.cpp
illwieckz: hmm, yes will fix the comment

src/engine/renderer/tr_shader.cpp
illwieckz: note: this may be confusing, but that's the comment for the normalscale not normalformat, I edited comments.

src/engine/renderer/tr_shader.cpp
```diff
@@ -4652,6 +4792,37 @@ static void CollapseStages()
...
+				{
+				stage->normalScale[ i ] *= stage->normalFormat[ i ];
+			}
+			}
```
illwieckz, line 4820: note to myself: rebase space hell there.

src/engine/renderer/tr_shader.cpp
illwieckz: That's related to the fact Z reconstruction is never done when normal map has heightmap in normal map (as reconstruction is destructive on alpha channel). This code applies format translation for X and Y in all case, and Z only when it's not reconstructed. I will make the comment more explicit.

illwieckz:
I also added a warning when people use negative value with `normalScale`, telling `normalFormat` is recommended to swap channels.

illwieckz:
OK, now everything looks fine to me, I think I addressed all the comments.

I'm just waiting for this to be merged to push the normal map format fixes on vega map and texture repositories.

slipher:
LGTM

Issue #302, opened 2020-03-12
need a reliable way to detect if a given surface is one of a bsp model


illwieckz:
Render is basically doing this to pick which glsl code to use to render a given texture:

```python3
if lightmapping is enabled and surface has lightmap:
	render_with_lightmap_as_map()
else if entity is not world:
	render_with_lightgrid_as_model()
else:
	render_with_lightgrid_as_map()
```

Real code there:

https://github.com/DaemonEngine/Daemon/blob/42e408a4f45f6e78b59fe9b7022f57456a7677d6/src/engine/renderer/tr_shade.cpp#L3058-L3070

The issue is:

If a bsp model has light map, light mapping is enabled the model is rendered with the same code used to render the whole map in all case, which is good.

But if such bsp model is not a world model (for example a func_rotating entity), if light mapping is disabled, such bsp model is rendered with the c++ code and glsl code for the non-bsp models (like player models, weapons…).

So basically we have this:

kind|world|lm enabled|lm disabled
----|-----|--------|-------
bsp model|yes|lightmap world|lightgrid world
bsp model|no|lightmap world|lightgrid entity
game model|no|lightgrid entity|lightgrid entity

When bsp models are rendered as game models, such unwanted glsl computation may be done, like rim lighting.

Although, we see non-world bsp model are handled fine by both lightmap, lightgrid world and lightgrid entity shaders, meaning there is maybe not a strong need for so much copypaste (see #301 for ongoing merge effort).

To improve the code I would like to have a reliable way to know if a model is a bsp model or not. I don't think it's reliable to detect if surface has lightmap since map may be built without.

To make the bug obvious, I temporarily modified the lightgrid entity glsl shader (`vertexLighting_DBS_entity_fp.glsl`) to set red channel to full, this to paint in red every surface rendered by this code.

Light map enabled:

[![map entities as models](https://dl.illwieckz.net/b/daemon/bugs/map-entities-as-models/unvanquished_2020-03-10_154112_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/map-entities-as-models/unvanquished_2020-03-10_154112_000.jpg)

Light map disabled:

[![map entities as models](https://dl.illwieckz.net/b/daemon/bugs/map-entities-as-models/unvanquished_2020-03-10_154139_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/map-entities-as-models/unvanquished_2020-03-10_154139_000.jpg)

Light map enabled:

[...]

Issue #296, opened 2020-02-27
dynamic lightmap take precedence over vertex lighting if exists


illwieckz:
In case of dynamic lightmap, surface is lightmapped even if vertex world lighting is enabled.

illwieckz:
How to reproduce:

- load [this map](https://dl.illwieckz.net/b/daemon/bugs/vertex-lighting-artifacts/map-metro_0~b1-2.dpk)
- enable vertex lighting, restart renderer
- go to the alien base

illwieckz:
This seems to come from the fact those lightmaps are standalone lightmaps, we can also see standalone lightmaps (legacy uncollapsed q3 material) being rendered in vertex world lighting.

PR #309, opened 2020-04-11
tell a surface belongs to the bsp, properly select the renderer

illwieckz:
See #302 for explanations about the issue it fixes.

If I color in red what is rendered by `vertexLighting_DBS_entity_fp.glsl`, in green what is rendered by `vertexLighting_DBS_world_fp.glsl` and in blue what is rendered by `lightMapping_fp.glsl`, I get this:

While world vertex lighting is disabled:

[bsp surface](https://dl.illwieckz.net/b/daemon/bugs/map-entities-as-models/unvanquished_2020-04-11_085656_000.jpg)

While world vertex lighting is enabled:

[bsp surface](https://dl.illwieckz.net/b/daemon/bugs/map-entities-as-models/unvanquished_2020-04-11_085727_000.jpg)

Everything looks fine now, bsp non-world models are not anymore rendered as entities.

There is still cases where surfaces are lightmapped while vertex world lighting is enabled, but that looks to be the bug described in #296.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -828,6 +828,7 @@ static void RB_RenderDrawSurfaces( shaderSort_t fromSort, shaderSort_t toSort,
 	shader_t      *shader, *oldShader;
 	int           lightmapNum, oldLightmapNum;
 	int           fogNum, oldFogNum;
+	bool          bspSurface;
 	bool      depthRange, oldDepthRange;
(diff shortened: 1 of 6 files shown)


illwieckz:
The bug was found while working on #301 (_merge lightmap and vertex (lightgrid) world and entity lighting_).

illwieckz:
I think there is very low chance this code introduces bugs, at least it's better than what we had before.

Issue #304, opened 2020-04-05
Condump filesystem issue with OneDrive


slipher:
When OneDrive storage is enabled for the My Documents folder on Windows 10, the `condump` command fails with the message `Failed to open 'condump/' for writing: Filename contains invalid characters`.

[Reported by themuffinator](https://forums.unvanquished.net/viewtopic.php?f=6&t=2176).

PR #310, opened 2020-04-11
Write condump output to passed file name

dimhotepus:
When we pass arg to /condump command, it should be taken as file
name where command output should be written to.

Now if arg is passed, it is taken as file name. If no arg is passed, file name
is generated by the engine.

Fixes https://github.com/DaemonEngine/Daemon/issues/304


slipher:
Thanks! Now I feel silly for falling for the idea that it is related to OneDrive.

PR #312, opened 2020-04-12
polylib: workaround MAX_POINTS_ON_WINDING in ChopWindingInPlace, ref #169

illwieckz:
Tell the workaround is applied (as a warning) so we don't forget we applied a workaround.
See #169 for details about the issue.

### DIFF
--- a/src/common/cm/cm_polylib.cpp
+++ b/src/common/cm/cm_polylib.cpp
@@ -209,9 +209,14 @@ ChopWindingInPlace
 */
 void ChopWindingInPlace( winding_t **inout, vec3_t normal, vec_t dist, vec_t epsilon )
 {
+// FIXME: https://github.com/DaemonEngine/Daemon/issues/169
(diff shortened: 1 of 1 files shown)


slipher:
It seems that this define is only used for the size of the stack-allocated buffers in this one function. This means there is no real cost to making it larger than necessary. I'd just increase the number without adding any TODOs or whatever, because there is nothing that needs to be fixed.

illwieckz:
We know we're not sure there is no bug because Tremulous 1.1.0 loads the _moteof_ map with `MAX_POINTS_ON_WINDING` set to `64` and Unvanquished requires more than twice of that, while we expect they both run the same code on that part.

Either Tremulous had a bug elsewhere disabling the feature that requires more than `64` (and our tree fixed this bug but we don't know this bug), either Unvanquished has a bug we don't know where  and must not require more than `64`, but we don't know this bug.

illwieckz:
Made it `Debug` instead of `Warn` to not spam the average user. Also printed values.

PR #315, opened 2020-04-12
map meta lookup: only check for map in packs starting with “map-” prefix, fix #285

### LINKED ISSUE #285: Unhandled exception (St12out_of_range) when displaying map list window if dpk with small name is in path
I don't know from which server it comes, but if that a dpk file with a small file name is in path, the engine triggers this exception and quits:

```
Warn: Unhandled exception (St12out_of_range): basic_string::substr: __pos (which is 4) > this->size() (which is 2) 
```
The only requirement to reproduce the bug seems to have this kind of file in path to crash the engine while clicking on “_Start local/LAN game_”:

[Unhandled exception (St12out_of_range)](https://dl.illwieckz.net/b/daemon/bugs/dpk-unhandled-exception/20200205-034900-001.unvanquished-dpk-unhandled-exception.png)

I reproduce the bug on released `0.51.1` and `master`.

illwieckz:
Map meta lookup: only check for map in packs starting with `map-` prefix, fix #285.

- don't look for map in packs whose name is too small to have a map-prefix
- don't look for map in packs not starting with `map-` prefix
- don't look for map in packs with empty string after `map-` prefix

---

- fix #285 bug
- may speedup map listing

Note: looking for stuff in `meta/` looks to be game-specific.

### DIFF
--- a/src/engine/qcommon/files.cpp
+++ b/src/engine/qcommon/files.cpp
@@ -637,9 +637,22 @@ void FS_LoadBasePak()
 
 void FS_LoadAllMapMetadata()
 {
+	const std::string pak_map_prefix("map-");
+
 	for (const auto& pak: FS::GetAvailableMapPaks()) {
 		try {
-			FS::PakPath::LoadPakPrefix(pak, va("meta/%s/", pak.name.substr(4).c_str()));
+			// If pak name is long enough to have “map-” prefix,
+			// and pak name without ”map-” prefix is not empty string.
+			if (pak.name.length() > pak_map_prefix.length())
+			{
+				// If pak name starts with “map-” prefix
+				if (Str::IsPrefix(pak_map_prefix, pak.name))
+				{
+					// FIXME: This looks to be game-specific,
+					// not all games use “meta/” mechanism.
+					FS::PakPath::LoadPakPrefix(pak, va("meta/%s/", pak.name.substr(pak_map_prefix.length()).c_str()));
+				}
+			}
 		} catch (std::system_error&) {} // ignore and move on
 	}
 }


src/engine/qcommon/files.cpp
```diff
@@ -639,7 +639,19 @@ void FS_LoadAllMapMetadata()
 {
 	for (const auto& pak: FS::GetAvailableMapPaks()) {
 		try {
-			FS::PakPath::LoadPakPrefix(pak, va("meta/%s/", pak.name.substr(4).c_str()));
+			std::string pak_map_prefix("map-");
```
dimhotepus, line 642: May be make `const` and move out of `for` loop?

src/engine/qcommon/files.cpp
```diff
@@ -639,7 +639,19 @@ void FS_LoadAllMapMetadata()
...
+			// If pak name is long enough to have “map-” prefix
+			// and pak name without "map-" prefix is not empty string.
+			if (pak.name.length() > pak_map_prefix.length())
+			{
+				// If pak name starts with "map-" prefix
+				if (pak.name.rfind(pak_map_prefix, 0) == 0)
```
dimhotepus, line 648: Out of curiosity, why not just `find(pak_map_prefix)` ?

https://en.cppreference.com/w/cpp/string/basic_string/find

src/engine/qcommon/files.cpp
dimhotepus: Especially useful for debug builds, where optimizations are not applied.

src/engine/qcommon/files.cpp
illwieckz: Right, `rfind` would do “startswith without other occurrences”, while `find` allows package to be named `map-map-map_0.dpk`, which is legit. I'm fixing it.

cmake/DaemonFlags.cmake
```diff
@@ -162,7 +162,7 @@ elseif (NACL)
     set_c_cxx_flag("-O2 -DNDEBUG -g3"   RelWithDebInfo)
     set_c_cxx_flag("-O0          -g3"   Debug)
 else()
-    set_c_cxx_flag("-ffast-math")
+#    set_c_cxx_flag("-ffast-math")
```
slipher, line 165: what's this

src/engine/qcommon/files.cpp
```diff
@@ -639,7 +639,19 @@ void FS_LoadAllMapMetadata()
...
 	for (const auto& pak: FS::GetAvailableMapPaks()) {
 		try {
-			FS::PakPath::LoadPakPrefix(pak, va("meta/%s/", pak.name.substr(4).c_str()));
+			std::string pak_map_prefix("map-");
+			// If pak name is long enough to have “map-” prefix
+			// and pak name without "map-" prefix is not empty string.
+			if (pak.name.length() > pak_map_prefix.length())
+			{
+				// If pak name starts with "map-" prefix
+				if (pak.name.find(pak_map_prefix, 0) == 0)
```
slipher, line 648: `Str::IsPrefix`

cmake/DaemonFlags.cmake
illwieckz: You noticed it between two force-push, faster than I notice it myself :sweat_smile:, was an attempt for https://github.com/DaemonEngine/Daemon/issues/169#issuecomment-611310651, was mistakenly committed there.

src/engine/qcommon/files.cpp
illwieckz: Thanks.

src/engine/qcommon/files.cpp
illwieckz: Done.

src/engine/qcommon/files.cpp
```diff
@@ -637,9 +637,22 @@ void FS_LoadBasePak()
 
 void FS_LoadAllMapMetadata()
 {
+	const std::string pak_map_prefix("map-");
+
 	for (const auto& pak: FS::GetAvailableMapPaks()) {
```
slipher, line 642: GetAvailableMapPaks is already supposed to filter for paks that start with `map-`, but the filtering is disabled when fs_legacypaks is on. So either this function should be changed to always return only paks starting with `map-`, or the behavior here should do something different when the map does not start with `map-`.

illwieckz: This is to load meta information to build the create server ui.

We may have to keep the test for the `map-` prefix in both.

This one is to filter maintained maps that were packaged the right we, we are not expecting legacy paks to have meta information. So, since `GetAvailableMapPaks` may list all map packs even the legacy ones, we have to filter in `FS_LoadAllMapMetadata`

We can either drop the `map-` prefix test in `GetAvailableMapPaks`, but we can keep it for the sole purpose of speeding-up things for when support for legacy packs is disabled.

The `map-` prefix for maps is part of DPK format definition, so this is OK to keep the test for it in engine wen looking for map packs when support for legacy packs is disabled.

That `FS_LoadAllMapMetadata` function really looks like a game-specific feature, especially Unvanquished specific feature. We may imagine other games doing things differently than the `meta/` thing (even if we strongly recommend to do so, as best practices and best support). Or we have to decide the meta thing is part of the DPK format.

At this point it's fairly legit to think a game may extends `FS_LoadAllMapMetadata`, extending the existing meta stuff with checks for other games (like legacy `.arena` files). It's not up to the engine to support all existing features.

Also, we can imagine a game wanting to support more formats would rely on this unmodified `FS_LoadAllMapMetadata` but add another function to complete the list with other formats. In this case we may remove the `map-` prefix and make sure this function is never used for other formats.

slipher:
LGTM

dimhotepus:
LGTM

PR #317, opened 2020-04-13
renderer: factorize and move some computation done every frame for every surface in tr_shade to be done only once in tr_shader

illwieckz:
Do a refactorization of some renderer thing, move some computation that are done every frame for every surface in `tr_shade.cpp` to be done only once at map load (or when shader is created) in `tr_shader.cpp`.

This makes also easier to implement some features, pull request #307 (_engine-agnostic normal fomat_) relies heavily on this.

Also deletes 150 lines.


slipher:
Looks OK stylistically. I didn't try to closely review for correctness.

illwieckz:
I have tested with many maps from many games without noticing any regression. The changes are not rocket science anyway, were just boring and annoying to do.

PR #318, opened 2020-04-16
[WIP] PVS Studio fixes #3

### LINKED ISSUE #3: Organization name change.
@DaemonDevelopers/owners, what's your thoughts on renaming this organization to `DaemonEngine`? It's shorter, which is nice because the current name is too long to fully display inside "Repositories you contribute to", and it's also more descriptive, showing that this organization hosts a (game) engine and its tools. (Unfortunately, just `Daemon` is taken by a GitHub user.)

Right now it's easy to make such a rename as the engine wasn't fully outsourced from Unvanquished just yet, but this is bound to happen very soon.

dimhotepus:

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -1152,15 +1152,16 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 	}
 
 	loadedPaks.emplace_back();
-	loadedPaks.back().name = pak.name;
-	loadedPaks.back().version = pak.version;
-	loadedPaks.back().checksum = pak.checksum;
-	loadedPaks.back().type = pak.type;
-	loadedPaks.back().path = pak.path;
+	auto loadedPak = loadedPaks.back();
+	loadedPak.name = pak.name;
+	loadedPak.version = pak.version;
+	loadedPak.checksum = pak.checksum;
+	loadedPak.type = pak.type;
+	loadedPak.path = pak.path;
 
 	// Update the list of files, but don't overwrite existing files, so the sort order is preserved
 	if (pak.type == pakType_t::PAK_DIR) {
-		loadedPaks.back().fd = -1;
+		loadedPak.fd = -1;
 		auto dirRange = RawPath::ListFilesRecursive(pak.path, err);
 		if (err)
 			return;
@@ -1176,14 +1177,14 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 		}
 	} else if (pak.type == pakType_t::PAK_ZIP) {
 		// Open file
-		loadedPaks.back().fd = my_open(pak.path, openMode_t::MODE_READ);
-		if (loadedPaks.back().fd == -1) {
+		loadedPak.fd = my_open(pak.path, openMode_t::MODE_READ);
+		if (loadedPak.fd == -1) {
 			SetErrorCodeSystem(err);
 			return;
 		}
 
 		// Open zip
-		zipFile = ZipArchive::Open(loadedPaks.back().fd, err);
+		zipFile = ZipArchive::Open(loadedPak.fd, err);
 		if (err)
 			return;
 
@@ -1219,19 +1220,19 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 	}
 
 	// Save the real checksum in the list of loaded paks (empty for directories, not used for legacy paks)
-	loadedPaks.back().realChecksum = realChecksum;
+	loadedPak.realChecksum = realChecksum;
 
 	// Get the timestamp of the pak, but only for dpk files. 
 	// Directories (aka a dpkdir) don't need timestamp.
 	// Fixes Windows bug where calling _wstat64i with trailing slash causes "file not found" error.
 	// For future stat calls on directories, trim the trailing slash (if exists)
 	if (pak.type == pakType_t::PAK_ZIP) {
-		loadedPaks.back().timestamp = FS::RawPath::FileTimestamp(pak.path, err);
+		loadedPak.timestamp = FS::RawPath::FileTimestamp(pak.path, err);
 		if (err)
 			return;
 	}
 
-	loadedPaks.back().pathPrefix = pathPrefix;
+	loadedPak.pathPrefix = pathPrefix;
 
 	// Legacy paks don't have version neither checksum
 	if (!isLegacy) {
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -1152,7 +1152,7 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 	}
 
 	loadedPaks.emplace_back();
-	auto loadedPak = loadedPaks.back();
+	auto &loadedPak = loadedPaks.back();
 	loadedPak.name = pak.name;
 	loadedPak.version = pak.version;
 	loadedPak.checksum = pak.checksum;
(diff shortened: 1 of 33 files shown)


src/engine/server/sv_client.cpp
```diff
@@ -673,7 +673,7 @@ when the cvar is set to something, the download server will effectively never us
 */
 static bool SV_CheckFallbackURL( client_t *cl, const char* pakName, msg_t *msg )
 {
-	if ( !sv_wwwFallbackURL->string || strlen( sv_wwwFallbackURL->string ) == 0 )
+	if ( !sv_wwwFallbackURL->string || !*sv_wwwFallbackURL->string )
```
illwieckz, line 676: I'm not very happy with this kind of `!!*!*!!!*!` optimization that affects readability. Is there any alternate option?
If it were me I would replace existing `!(non bool)` tests with explicit tests against `0`, `\0`, `NULL` or `nullptr`…

illwieckz:
If I exclude

- the  `!*!*!!!*!` commit that does not make me happy,
- the `emplace_back` commits I'm not sure to completely understand so I let this to others to check…

that looks good to me.

src/engine/server/sv_client.cpp
dimhotepus: `if (a == nullptr || a[0] == '\0')`?

Also may be refactored to a function, like `bool is_null_or_empty(const char*) `

dimhotepus:
https://stackoverflow.com/questions/23717151/why-emplace-back-is-faster-than-push-back#23717223 answer has a good explanation on `push_back` vs `emplace_back` difference. 


```
Foo x; Bar y; Zip z;

v.push_back(T(x, y, z));  // make temporary, push it back
v.emplace_back(x, y, z);  // no temporary, directly construct T(x, y, z) in place
```

src/engine/server/sv_client.cpp
illwieckz: I like the `a[0] == '\0'` idea, I would like to know if @slipher has an opinion or something against this.

At least, this sounds less obscure to me (if I'm not doing mistake, it's equivalent and it works):

```c++
if ( !sv_wwwFallbackURL->string || !sv_wwwFallbackURL->string[0] )
```

This would cut it in half: not too verbose, explicit enough. That requires less understanding of pointer mechanism to understand what's done, at least it's easy to guess the right part is testing the content. Even a python programmer who would have no clue about c and pointers can infer `if None or if empty string`. This would encourage contributors to write good code the cargo cult way. =D

I would not be against a `is_null_or_empty` function if there is more than one or two occurrences.

illwieckz:
> https://stackoverflow.com/questions/23717151/why-emplace-back-is-faster-than-push-back#23717223 answer has a good explanation on `push_back` vs `emplace_back` difference.

OKOK

src/common/FileSystem.cpp
```diff
@@ -1152,15 +1152,16 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 	}
 
 	loadedPaks.emplace_back();
-	loadedPaks.back().name = pak.name;
-	loadedPaks.back().version = pak.version;
-	loadedPaks.back().checksum = pak.checksum;
-	loadedPaks.back().type = pak.type;
-	loadedPaks.back().path = pak.path;
+	auto loadedPak = loadedPaks.back();
```
dimhotepus, line 1155: Oops, missed ref qualifier, so it is copy of the loaded pack.
Now fixed.

src/engine/server/sv_client.cpp
illwieckz: @dimhotepus I guess `!sv_wwwFallbackURL->string[0]` and `!*sv_wwwFallbackURL->string` produces the same code, right?

dimhotepus:
```
if ( !sv_wwwFallbackURL->string || !sv_wwwFallbackURL->string[0] )
```

illwieckz: Yes, if you do this it would look OK to me. Then also squash the fix for the loaded pack reference and I'll merge the PR.

src/engine/renderer/tr_decals.cpp
```diff
@@ -190,7 +190,6 @@ void RE_ProjectDecal( qhandle_t hShader, int numPoints, vec3_t *points, vec4_t p
 
 	/* basic setup */
 	temp.shader = R_GetShaderByHandle( hShader );  /* debug code */
-	temp.numPlanes = temp.shader->entityMergable;
 	temp.color = color;
 	temp.numPlanes = numPoints + 2;
```
illwieckz, line 194: How do you know there was no bug and `temp.shader->entityMergable` had to be used somehow?

illwieckz: If you don't know, please add a comment like this one (feel free to reword):

```c++
	/* Note: the code was storing this value before overwriting it the next line.
	If something had to be done with this value, there may be bugs around there:
	temp.numPlanes = temp.shader->entityMergable;
	*/
	temp.numPlanes = numPoints + 2;
```

This way if someone reports a bug with decals, it would help.

dimhotepus: Done, thanks!

src/engine/server/sv_init.cpp
```diff
@@ -483,7 +483,7 @@ void SV_SpawnServer(const std::string pakname, const std::string mapname)
 	FS::PakPath::ClearPaks();
 	FS_LoadBasePak();
 	if (!FS_LoadPak(pakname.c_str()))
-		Sys::Drop("Could not load map pak\n");
+		Sys::Drop("Could not load map pak '%s'\n", pakname);
```
illwieckz, line 486: Thank you for that! That's precious!

illwieckz:
Can you squash the commits that reverts, fix or extend something that was introduced in another commit of this PR? Like the copy/ref/copy commits, and the commit adding a comment around a deleted affectation? After that, LGTM.

dimhotepus:
> Can you squash the commits that revert, fix or extend something that was introduced in another commit of this PR? Like the copy/ref/copy commits, and the commit adding a comment around a deleted affectation? After that, LGTM.

Ok, i've added last found potential problems. Going to cleanup commits as you suggested.

illwieckz:
Great. I've reviewed the new ones.

This C++ thing is beyond my knowledge:
https://github.com/DaemonEngine/Daemon/pull/318/commits/3d4d91b9e527a49c0ea1c952ac9b93851c896298
@slipher can you look at this one?

Otherwise, everything else LGTM.

src/engine/qcommon/translation.cpp
```diff
@@ -281,7 +281,7 @@ void Trans_LoadDefaultLanguage()
 		FL_FindLocale( &locale, FL_MESSAGES );
 
 		// Invalid or not found. Just use builtin language.
-		if( !locale->lang || !locale->lang[0] || !locale->country || !locale->country[0] )
```
slipher, line 284: What is the PVS studio finding here?

slipher:
> This C++ thing is beyond my knowledge:
> 3d4d91b (leak] Ensure no mem leak occurs on emplace failure)

It's fine. It doesn't matter for us since we always exit the program on an allocation failure (there's an operator new override that calls Sys::Error), but the change is harmless, and can help to avoid noise next time running the analyzer.

illwieckz:
Thanks, then everything looks OK on my end.

src/engine/qcommon/translation.cpp
dimhotepus: In case `country` is not present in locale info, we reverted locale to "en". But in `else` branch we see a fallback when `country` is not present - use just `lang` member to choose game language.

So was decided to remove `country` check above and just use `lang` for game language selection, if country is not present.

PR #329, opened 2020-04-24
do not crash the engine when using cubemaps

illwieckz:
The cubemap feature and reflective specular thing is not working, anyway, this PR makes sure the engine does not crash when the cubemaps are generated:

1. do not attempt to generate cubemaps (`buildcubemaps` command) when no map is loaded (for example in main menu);
2. always load a `whiteCubeImage` when cubemap pointers are null and do more tests for this.

Note that the renderer attempts to use cubemaps while they are built so maybe the extra tests for null pointers are only required until they all are build successfully, but I have no proof there would not be a bug somewhere trying to bind `nullptr` as image in some other scenario.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -948,6 +948,11 @@ static void Render_lightMapping( int stage )
 		cubemapProbe_t *cubeProbeNearest;
 		cubemapProbe_t *cubeProbeSecondNearest;
 
+		image_t *cubeMap0 = nullptr;
+		image_t *cubeMap1 = nullptr;
+
+		float interpolation = 0.0;
+
 		bool isWorldEntity = backEnd.currentEntity == &tr.worldEntity;
 
 		if ( backEnd.currentEntity && !isWorldEntity )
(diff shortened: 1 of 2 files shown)


illwieckz:
On a side note, that cubemap code is expected to be broken: there is `FIXME` comments that date backs to the very first commit for the repository, I noticed cubemaps are not invalidated when loading a new map, etc.

Environment mapping would be a very cool feature to have, but it looks like no one have looked at it yet. This PR only ensures we can keep the code in the tree until someone can take the time to look at it and fix it to make Dæmon great again.

src/engine/renderer/tr_shade.cpp
```diff
@@ -964,35 +969,21 @@ static void Render_lightMapping( int stage )
...
-			GL_BindToTMU( BIND_ENVIRONMENTMAP0, cubeProbeNearest->cubemap );
-
-			// bind u_EnvironmentMap1
-			// GL_SelectTexture( BIND_ENVIRONMENTMAP1 );
-			// GL_Bind( cubeProbeNearest->cubemap );
-
-			// u_EnvironmentInterpolation
-			gl_lightMappingShader->SetUniform_EnvironmentInterpolation( 0.0 );
+			cubeMap0 = cubeProbeNearest->cubemap;
+			// cubeMap1 = cubeProbeNearest->cubemap;
```
slipher, line 986: remove this line?

illwieckz: I prefer to not delete information from a code that does not work, we never attempted to debug, and we will have to debug one day. That would not be the same if this was working.

slipher: But it's just noise here. It's not difficult to see what the code is trying to do - interpolate between two images if two of them are found, or if only one is found, just use the one image. Moreover, the case where the first one is missing and the case where the 2nd is missing should be the same.

illwieckz: okok

PR #330, opened 2020-04-25
Fix attack3, drop useitem, reorder command array, ref Unvanquished#1182, Unvanquished #544

illwieckz:
Counterpart to https://github.com/Unvanquished/Unvanquished/pull/1182

Fix `attack3`, drop `useitem`, reorder command array, ref Unvanquished#1182, Unvanquished #544

See https://github.com/Unvanquished/Unvanquished/issues/1180
See https://github.com/Unvanquished/Unvanquished/issues/544

### DIFF
--- a/src/engine/qcommon/q_shared.h
+++ b/src/engine/qcommon/q_shared.h
@@ -1810,9 +1810,9 @@ union OpaquePlayerState {
 //                          10
 //                          11
 //                          12
-//                          13
+// BUTTON_DECONSTRUCT       13 // defined in gamelogic
 #define BUTTON_RALLY        14
-#define BUTTON_DODGE        15
+//                          15
 
 #define MOVE_RUN          120 // if forwardmove or rightmove are >= MOVE_RUN,
 // then BUTTON_WALKING should be set
--- a/src/engine/qcommon/q_shared.h
+++ b/src/engine/qcommon/q_shared.h
@@ -1794,8 +1794,8 @@ union OpaquePlayerState {
 #define USERCMD_BUTTONS     16 // bits allocated for buttons (multiple of 8)
 
 #define BUTTON_ATTACK       0
-#define BUTTON_TALK         1  // disables actions
-#define BUTTON_USE_HOLDABLE 2
+#define BUTTON_ATTACK2      1
+#define BUTTON_ATTACK3      2
 #define BUTTON_GESTURE      3
 #define BUTTON_WALKING      4  // walking can't just be inferred from MOVE_RUN
                                // because a key pressed late in the frame will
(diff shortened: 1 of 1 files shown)


slipher:
I think it would better to squash it into a single commit, since in the end all you want to do is change the name of one button.

illwieckz:
yes, now that everything work I can squash that.

slipher:
You know, I think it would be better to rename the one entry in place, without moving the numbers around, which could cause unnecessary difficulties later when trying to test old versions.

Of course in the long run, these should not be defined in the engine at all. There are apparently some hacks related to specific entries (namely BUTTON_TALK, BUTTON_ANY, and BUTTON_WALKING), but for other entries you could already ignore the q_shared definitions and make different ones in the cgame.

illwieckz:
> You know, I think it would be better to rename the one entry in place, without moving the numbers around, which could cause unnecessary difficulties later when trying to test old versions.

I prefer to break it before beta if we want to cleanup things. Trying old versions would be trying old games with newer engine. We properly tag our repository to prevent that, and this commit targets the `next` branch the right way.

> Of course in the long run, these should not be defined in the engine at all. There are apparently some hacks related to specific entries (namely BUTTON_TALK, BUTTON_ANY, and BUTTON_WALKING), but for other entries you could already ignore the q_shared definitions and make different ones in the cgame.

I wonder if those special keybinds keys are related to the ability to handle `+action` and `-action` but the jetpack hold seems to work the other way anyway. I also make supposition about that may be a faster implementation, or less hack friendly (don't know)… For sure `taunt` really don't need to eat-up a precious slot there, maybe `activate` can be moved too.

illwieckz:
If I'm right, this was obsoleted by #341.

There is one commit that is still useful, will be part of another PR.

illwieckz:
No, this was not obsoleted, we still need this.

slipher:
I don't think we need it. It's not important to change the names or numbers of the buttons, and for the long term the buttons should be moved out of the engine anyway.

illwieckz:
The purpose of this work was to implement attack3 (so it can be used for firing weapons, like alien barb or spit instead of abusing useitem), implement weapprev and weapnext, and then drop useitem.

Reordering or not, we need the feature. I don't think it's a bad idea to make the code more readable for now if we gonna drop it later. At least, I know where I'm going.

illwieckz:
At this point it looks OK to me. There is more improvements to do, but it's better to first merge this. Same if we want to do something in another way, it's better to first get a working state and then go forward.

slipher:
Merged as #1208.

illwieckz:
not needed anymore

PR #332, opened 2020-04-25
renderer: use specular code instead of pbr code when there is neither specular neither physical map, it's faster

illwieckz:
renderer: use specular code instead of pbr code when there is neither specular neither physical map, it's faster

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -4473,14 +4473,14 @@ static void CollapseStages()
 			continue;
 		}
 		else if ( stages[ i ].collapseType == collapseType_t::COLLAPSE_generic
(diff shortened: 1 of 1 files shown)


SomaZ:
Swapping between energy conserving code and non energy conserving code will result in an inconsistent look. Shouldnt make a big difference performance wise. Also, using the same shader multiple times against swapping shaders is concidered to be faster overall. My two cents and I'm out. :P

illwieckz:
> Swapping between energy conserving code and non energy conserving code will result in an inconsistent look.

Unvanquished material library is like 99% specular based, 1% plain color map (this use case), 0% pbr at this point (this can change).

>  Shouldnt make a big difference performance wise.

OK, I like the idea that the default rendering mode of Dæmon is PBR.

PR #338, opened 2020-05-01
playerstate: introduce weaponCharge, fix #336

### LINKED ISSUE #336: [Need Help] increase the player stat index
Required by https://github.com/Unvanquished/Unvanquished/pull/1182.

I need help to increase `MAX_STATS` as seen here:

https://github.com/DaemonEngine/Daemon/blob/393e1829d5acffc0f0fde14c87fd8dd9e7141445/src/engine/qcommon/q_shared.h#L1764-L1766

To add one field there:

https://github.com/Unvanquished/Unvanquished/blob/f79ea6e9b10f6d0d0c32b59f9ed371e099c52b9a/src/shared/bg_public.h#L224-L244

@Viech it looks like you added a comment 7 years ago about a very similar array to increase the same way:

https://github.com/Unvanquished/Unvanquished/blame/f79ea6e9b10f6d0d0c32b59f9ed371e099c52b9a/src/shared/bg_public.h#L327

Is it something you know how to do?

illwieckz:
playerstate: introduce weaponCharge, fix #336

Required by https://github.com/Unvanquished/Unvanquished/pull/1188

Unfortunately includes https://github.com/DaemonEngine/Daemon/pull/330 because of the Unvanquished PR including commits also requiring it (it's hard to split the commits on the other side).

### DIFF
--- a/src/engine/qcommon/msg.cpp
+++ b/src/engine/qcommon/msg.cpp
@@ -706,6 +706,7 @@ void MSG_WriteDeltaUsercmd( msg_t *msg, usercmd_t *from, usercmd_t *to )
 	     from->upmove == to->upmove &&
 	     !memcmp( from->buttons, to->buttons, sizeof( from->buttons ) ) &&
 	     from->weapon == to->weapon &&
+	     from->weaponCharge == to->weaponCharge &&
 	     from->flags == to->flags && from->doubleTap == to->doubleTap)
 	{
 		// NERVE - SMF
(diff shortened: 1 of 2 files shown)


illwieckz:
@Amanieu do you know how much this is “breaking change”?

Amanieu:
Yes this is a breaking change to the network protocol.

slipher:
But no more breaking than changing the syscall interface between the engine and a gamelogic module.

illwieckz:
In general I'm OK to break things if that reduces breakage in the future like:
- implementing a very common feature the right way so there is high change people needing this feature too would be happy to discover that the engine already supports it end they don't need to break it themselves.
- redesign some stuff in a way newer changes can be done without breaking things.

Basically for this PR I had the choice to implement buildtimer or weaponcharge in playerstate, but buildtimer is unvanquished-specific (game-specific) while weaponcharge is fps-specific (genre-specific), that's why I did weaponcharge.

I sometime think about the fact maybe some parts of the engine may be doable in a nexe. Those would not be updatable and never read from dpk archive, but a game developer would be able to tweak some things on the engine side without having to taking care of the operating system or the achitecture.

illwieckz:
If I'm right, this was obsoleted by #341.

PR #340, opened 2020-05-04
PVS Studio fixes for used libs

dimhotepus:


illwieckz:
For minizip we may want to rely on https://github.com/nmoinvaz/minizip instead, as a submodule.
Note that the master of this maintained fork is not source-compatible, a port has to be done (see here for details about that: https://github.com/XQF/xqf/issues/223#issuecomment-623173893).

dimhotepus:
For now we can patch `minizip` as is. If we migrate to `minizip2`, I'll check this, too.

Anyway, thank you for clarification!

illwieckz:
Yes, there is nothing against merging this patch. It's better to have it. :-)

PR #341, opened 2020-05-05
Gamelogic-configurable netcode for playerState_t

slipher:
A lot of hacks! There are a few fields referenced in the engine code, so I'm keeping those hard-coded in for now. I have only implemented loading the table from sgame so far, not from cgame (though it's matter of trivial copy and paste), so it only works with a local server.

The point is to be able to change the definition of playerState_t without rebuilding the engine.  If we like this, there's another table for entityState_t that it can also be done for.

### DIFF
--- a/src/common/IPC/CommonSyscalls.h
+++ b/src/common/IPC/CommonSyscalls.h
@@ -34,6 +34,26 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "Primitives.h"
 #include <common/FileSystem.h>
 
+namespace Util {
+    template<> struct SerializeTraits<netField_t> {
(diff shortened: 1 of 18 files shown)


slipher:
This should be in a mergeable state now.

illwieckz:
I pulled `playerstate/sync` on both side and I got that while loading a map:

```
Warn: SGame VM: IPC: Unexpected end of message 
```

And the game returns to main menu.

slipher:
Whoops, accidentally put `QVM` as the syscall major number of `GetNetcodeTablesMsg`, when it was ended to be `MISC`. Now fixed.

illwieckz:
I now see no issue on my end. I'm not sure to understand everything but I'm confident in what you've done.

PR #350, opened 2020-07-11
Reuse window when attempting to create a context, fix #349

### LINKED ISSUE #349: Two windows are created on OpenGL 2.1 hardware
On OpenGL 2.1 hardware, two windows are created, one of them having undefined content, the other one rendering the game itself:

[opengl2.1 two windows](https://dl.illwieckz.net/b/daemon/bugs/opengl2.1-two-windows/20200707-131814-000.unvanquished-two-windows.png)

illwieckz:
For each format attempt, the renderer creates a window and try to create a context using this window.
But anytime a context cannot be created, the next one is tried without destroying the previous window first, or reusing it.

This is an attempt to fix #349.


slipher:
LGTM

PR #354, opened 2020-07-13
renderer: print shader and macro names at glsl build time

illwieckz:
Print shader name and macro at build time, may help to trouble shoot various things.

That debug code prints something like that in console:

```
Debug: building shadowFill shader permutation with macro: none
Debug: building shadowFill shader permutation with macro: USE_VERTEX_SKINNING
Debug: building shadowFill shader permutation with macro: USE_VERTEX_ANIMATION
Debug: building shadowFill shader permutation with macro: LIGHT_DIRECTIONAL
Debug: building shadowFill shader permutation with macro: USE_VERTEX_SKINNING LIGHT_DIRECTIONAL
Debug: building shadowFill shader permutation with macro: USE_VERTEX_ANIMATION LIGHT_DIRECTIONAL
```

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -1015,6 +1015,10 @@ void GLShaderManager::CompileGPUShaders( GLShader *shader, shaderProgram_t *prog
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


illwieckz:
Well, this is just a log, can't break things.

PR #356, opened 2020-07-13
do not build GLSL shaders that will not be used

illwieckz:
After having implemented #354 I discovered something awful: the engine is building all possible shader and permutations even if the game is never gonna use them.

1. GLSL shaders: if you don't use the FXAA effect, you don't need to build the FXAA shader.
2. Permutations are ifdef macro-based variants for the same GLSL shader: for example the `lightMapping` shader is built two times for BSP and non-BSP surface, and the game selects the right one given the surface. The same mechanism is used for reliefMapping and surface having heightmap or not. The thing is, if reliefMapping is disabled, the engine does not need to build the reliefMapping variant of it.

This is an example with an old laptop with slow hardware:  a 13 years old Lenovo ThinkPad X61 Tablet from the year 2007 featuring a Core 2 Duo L7500 @1.60 GHz with an integrated GMA X3100 (GM965) graphic chip with OpenGL 2.1 support, sharing 385MB (on 3072MB) of DDR2-667MHz PC2-5300 ram with the CPU.

I disabled in game any GLSL feature I can to make things go faster.

Before that patch, the engine was building 445 GLSL shader variants for 82.8 seconds (almost a minute!), then after that patch, the engine is building 70 GLSL shader variants for 128msec (yes, almost 1/10 of a second!)…

GLSL building time is divided by 400!

On a computer that can handle all the features (relief mapping etc.), compiling all the shader variants is a matter of second. But when your computer is so old you need to disable feature to make it somewhat playable, you would also want to not spend time to build GLSL shaders you won't use. Also, you may save memory and other resources.

So, on that old laptop, this is before:

```
Debug: building generic shader permutation with macro: none
Debug: building generic shader permutation with macro: USE_VERTEX_SKINNING 
Debug: building generic shader permutation with macro: USE_VERTEX_ANIMATION 
Debug: building generic shader permutation with macro: USE_VERTEX_SPRITE 
Debug: building generic shader permutation with macro: USE_TCGEN_ENVIRONMENT 
Debug: building generic shader permutation with macro: USE_VERTEX_SKINNING USE_TCGEN_ENVIRONMENT 
Debug: building generic shader permutation with macro: USE_VERTEX_ANIMATION USE_TCGEN_ENVIRONMENT 
Debug: building generic shader permutation with macro: USE_VERTEX_SPRITE USE_TCGEN_ENVIRONMENT 
Debug: building generic shader permutation with macro: USE_TCGEN_LIGHTMAP 
[...]

PR #358, opened 2020-07-16
renderer: do not load textures for disabled feature

illwieckz:
Simiar to #356 (to not build shader for disabled feature), do not load texture for disabled feature. For example, if specular mapping is disabled, do not load specular maps.


illwieckz:
We may also not load lightmaps when lightmapping is disabled but #296 must be fixed first (“light styles” dynamic lightmaps are still displayed even when lightmapping is disabled).

illwieckz:
Now lightmaps aren't loaded when light mapping is disabled, same for “light style” light maps.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -447,6 +447,11 @@ R_LoadLightmaps
 static const int LIGHTMAP_SIZE = 128;
 static void R_LoadLightmaps( lump_t *l, const char *bspName )
 {
+	if ( r_vertexLighting->integer )
```
slipher, line 450: does vertex lighting really mean lightmapping is disabled?

illwieckz: Yes. This cvar is really misnamed and the way it is used is really confusing:

- everywhere in engine (and game UI) the purpose of this cvar is to know if light mapping is enabled or not;
- engine never does vertex lighting but grid lighting instead.

I plan to make a PR to fix this later (later to not break any possible work-in-progress branches).

In the future I want to have this:

- `r_lightMapping 0`: light mapping disabled, use grid lighting for everything (map and models)
- `r_lightMapping 1`: light mapping enabled, use light map anytime a surface has one (map), use grid light for other surfaces (models)

This naming and value scheme would not only make code easier to read, but it would also enable alternative light scenario like:

- `r_lightMapping 2`: raytrace the light map at map load.
- `r_lightMapping 3`: raytrace the light in real time.

Or anything else we can imagine or cannot imagine yet.

PR #362, opened 2020-08-22
renderer: refactor extension loading code

illwieckz:
refactor extension loading code

also make sure extension loading is always logged (and the same way)

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1246,7 +1246,13 @@ success:
 
 		glConfig.extensions_string[ 0 ] = '\0';
 		for ( i = 0; i < numExts; ++i )
+		{
+			if ( i != 0 )
+			{
+				Q_strcat( glConfig.extensions_string, sizeof( glConfig.extensions_string ), ( char * ) " " );
+			}
 			Q_strcat( glConfig.extensions_string, sizeof( glConfig.extensions_string ), ( char * ) glGetStringi( GL_EXTENSIONS, i ) );
+		}
 	}
 	else
 	{
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -962,73 +962,57 @@ GLimp_InitExtensions
 ===============
 */
 
-static void RequireExt( bool hasExt, const char* name )
-{
-	if ( hasExt )
-	{
-		logger.WithoutSuppression().Notice("...using GL_%s", name );
-	}
-	else
-	{
-		Sys::Error("Required extension GL_%s is missing", name );
-	}
-}
+/* ExtFlag_CORE means the extension is known to be an OpenGL 3 core extension.
+The code considers the extension is available even if the extension is not listed
+if the driver pretends to support OpenGL Core 3 and we know this extension is part
+of OpenGL Core 3. */
 
-static bool LoadExtWithCvar( bool hasExt, const char* name, bool cvarValue )
+enum {
+	ExtFlag_NONE,
+	ExtFlag_REQUIRED = BIT( 1 ),
+	ExtFlag_CORE = BIT( 2 ),
+};
+
+static bool LoadExt( int flags, bool hasExt, const char* name, bool test = true )
 {
-	if ( hasExt )
+	if ( hasExt || ( flags & ExtFlag_CORE && glConfig2.glCoreProfile) )
 	{
-		if ( cvarValue )
+		if ( test )
 		{
-			logger.WithoutSuppression().Notice("...using GL_%s", name );
+			logger.WithoutSuppression().Notice( "...using GL_%s", name );
 			return true;
 		}
 		else
 		{
-			logger.WithoutSuppression().Notice("...ignoring GL_%s", name );
+			// Required extension can't be made optional
+			ASSERT( !( flags & ExtFlag_REQUIRED ) );
+
+			logger.WithoutSuppression().Notice( "...ignoring GL_%s", name );
 		}
 	}
 	else
 	{
-		logger.WithoutSuppression().Notice("...GL_%s not found", name );
-	}
-	return false;
-}
-
-/*
-	load extensions that were made a required part of OpenGL core profile by version 3.2
-*/
-static bool LoadCoreExtWithCvar(bool hasExt, const char* name, bool cvarValue)
-{
-	if (hasExt || glConfig2.glCoreProfile)
-	{
-		if (cvarValue)
+		if ( flags & ExtFlag_REQUIRED )
 		{
-			logger.WithoutSuppression().Notice("...using GL_%s", name);
-			return true;
+			Sys::Error( "Required extension GL_%s is missing", name );
 		}
 		else
 		{
-			logger.WithoutSuppression().Notice("...ignoring GL_%s", name);
+			logger.WithoutSuppression().Notice( "...GL_%s not found", name );
 		}
-	}
-	else
-	{
-		logger.WithoutSuppression().Notice("...GL_%s not found", name);
 	}
 	return false;
 }
-#define REQUIRE_EXTENSION(ext) RequireExt(GLEW_##ext, #ext)
 
-#define LOAD_EXTENSION_WITH_CVAR(ext, cvar) LoadExtWithCvar(GLEW_##ext, #ext, cvar->value)
+#define LOAD_EXTENSION(flags, ext) LoadExt(flags, GLEW_##ext, #ext)
 
-#define LOAD_CORE_EXTENSION_WITH_CVAR(ext, cvar) LoadCoreExtWithCvar(GLEW_##ext, #ext, cvar->value)
+#define LOAD_EXTENSION_WITH_TEST(flags, ext, test) LoadExt(flags, GLEW_##ext, #ext, test)
 
 static void GLimp_InitExtensions()
 {
 	logger.Notice("Initializing OpenGL extensions" );
 
-	if ( LOAD_EXTENSION_WITH_CVAR(ARB_debug_output, r_glDebugProfile) )
+	if ( LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, ARB_debug_output, r_glDebugProfile->value ) )
 	{
 		glDebugMessageCallbackARB( (GLDEBUGPROCARB)GLimp_DebugCallback, nullptr );
 		glEnable( GL_DEBUG_OUTPUT_SYNCHRONOUS_ARB );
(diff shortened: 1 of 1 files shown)


src/engine/sys/sdl_glimp.cpp
```diff
@@ -1131,51 +1117,31 @@ static void GLimp_InitExtensions()
...
+#else
+	NoticeExtUnsupported("ARB_get_program_binary");
 #endif
-	{
-		logger.Notice("...GL_ARB_get_program_binary not found");
-		glConfig2.getProgramBinaryAvailable = false;
-	}
 
+	glConfig2.bufferStorageAvailable = false;
 #ifdef GL_ARB_buffer_storage
```
slipher, line 1140: Can we just get rid of these ifdefs? I don't see a need for compiling against ancient versions of the GL headers.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -962,67 +962,55 @@ GLimp_InitExtensions
...
-		logger.WithoutSuppression().Notice("...using GL_%s", name );
-	}
-	else
-	{
-		Sys::Error("Required extension GL_%s is missing", name );
-	}
-}
+#define NoticeExtUnsupported(name) logger.WithoutSuppression().Notice("...GL_%s unsupported", #name)
 
-static bool LoadExtWithCvar( bool hasExt, const char* name, bool cvarValue )
```
slipher, line 977: What does `isCore` mean?

illwieckz: The code does not test for the explicit extension if the driver advertises OpenGL Core 3 and we know Core 3 requires this extension.

I don't know which buggy driver would advertise GL Core 3 but not the related extensions, but that was the previous behavior I just refactored: when using the right macro, core extensions are checked by their name only if core is not there, typically the case with GL `Core 2.1 + ARB_framebuffer_object`, the lowest configuration we support.

src/engine/sys/sdl_glimp.cpp
illwieckz: Maybe, I would make a separate commit so if some people with ancient software comes, he can revert or look at the code to know how to workaround.

Or, I may extend those macro to do the `#ifdef` check anytime (does preprocessor do multiple pass?) that would make the code ready for GL 4.7 extensions while avoiding failures when compiling Xenial. 🙈

slipher:
I'm generally having trouble understanding the purpose of this PR. The code on the left side of the diff seems easier to understand. Also it's not good to add macros when there is no compelling reason (see e.g. https://stackoverflow.com/a/14041847).

If there are any bugs that need to be fixed, it would help to add more details in the commit message and/or have smaller focused commits.

illwieckz:
@slpiher, I experimented the need for a refactor when I toyed during my holidays to see if it would be possible to use `EXT_framebuffer_object` when `ARB_framebuffer_object` is missing (full disclosure: it went nowhere). Basically, either this or any feature or new extension we may want to support would benefit from this.

I also faced that need when I toyed with `GL_ARB_texture_swizzle` in the past: adding new extensions reveals current code is very inconsistent, and adding new code just adds more noise.

As it is before:

- the code is not consistent: to do the same thing sometime a function is called, sometime the function body is copy pasted, sometime custom code was written, whith risk to not do things the same way and introducing mistakes or inconsistencies when editing parts of the code;
- not every used extensions had a log line telling it was loaded (or not);
- people writing new code may copy paste one of the various ways to do the same thing with risk for copy-pasting the non-optimal one;
- strings for the same message were duplicated many times, if we re-implement translation later, it would be better to translate them once for all;

I'm not really happy with the macros, but at least I extended a macro that was already there so I'm not introducing macro-related issue. Also, I consider more valuable to get proper in game logging than gdb-aware functions: users will not use gdb and we need their testimony to troubleshoot their issues.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -962,67 +962,55 @@ GLimp_InitExtensions
...
-		if ( cvarValue )
+		if (cvarValue)
 		{
-			logger.WithoutSuppression().Notice("...using GL_%s", name );
+			logger.WithoutSuppression().Notice("...using GL_%s", name);
 			return true;
 		}
+		else if (isRequired)
+		{
+			DAEMON_ASSERT("Required extension can't be made optional");
```
slipher, line 978: asserting a value which is always true

slipher:
Can't you fix the "inconsistencies" without adding new macros?

src/engine/sys/sdl_glimp.cpp
illwieckz: `isRequired` is `true` with `LOAD_REQUIRED_EXTENSION` and `LOAD_REQUIRED_CORE_EXTENSION` but `false` with others.

slipher: The assert macro terminates the program when it was built in debug mode and the argument evaluates to false as a boolean. A string literal is always non-null, meaning it is true as a boolean, so nothing happens.

illwieckz:
> Can't you fix the "inconsistencies" without adding new macros?

It would be very verbose unfortunately and this will not remove the need for at least one or two macros…

src/engine/sys/sdl_glimp.cpp
illwieckz: Ah, I see what you mean. I guess I wanted to call `Sys::Error()` instead :woman_facepalming:

illwieckz: Ah no, this error would be a coding bug, so ASSET would be a good idea anyway, I just have to redo it better.

Anyway, why this file also has an `ASSERT` somewhere else? Are they the same?
https://github.com/DaemonEngine/Daemon/blob/6b94e2e890f893afdc55b7f70bd58cb39f3cafcd/src/engine/sys/sdl_glimp.cpp#L263-L265

slipher: The condition `smpData == nullptr` could be either true or false, depending on the value of `smpData`. So it makes sense.

illwieckz: Sorry, my question was: is `DAEMON_ASSERT` the same as `ASSERT`?

illwieckz:
I may use some `enum` to reduce the number of macros…

slipher:
I don't see what new capabilities are being introduced in the proposed changes, which would require new macros to be created.

src/engine/sys/sdl_glimp.cpp
slipher: `DAEMON_ASSERT` is the same as `ASSERT`

illwieckz:
purpose was readability, to be able to do:

```
LOAD_EXTENSION(ARB_debug_output);
```

instead of:

```c
LOAD_EXTENSION(false, false, ARB_debug_output, false);
```

Note that there seems to be at least one required macro to turn macro name into a string at compile time, so using optional arguments may be tricky to implement.

slipher:
It's possible to make a function with optional arguments, so you could pass the trailing arguments of a macro into such a function.

But for the purpose of getting rid of the ugly boolean arguments here, a flags argument seems better. Something like

```
enum class Extension Flags {
    CORE = 1 << 0,
    REQUIRED = 1 << 1,
};
```

The TEST variants should be removed - just use `if` statements. the CVAR variants are also unnecessary to have a function for, since you only used them one time.

illwieckz:
A flag is nice idea but I still don't see how to mlake a macro with an optional argument, without relying on huge verbose stuff that would not make it better.

The optional argument is for the optional test. I can merge `_WITH_CVAR` and `_WITH_TEST` so when there is a cvar I pass `cvar_name->value` instead of `cvar_name`, this would still reduce to two macros.

slipher:
That's why my suggestion was to get rid of the "test" thing. There is no need to write code like this:

```
		if ( formats == 0 )
		{
			logger.Notice("...GL_ARB_get_program_binary found, but with no binary formats");				
			glConfig2.getProgramBinaryAvailable = false;	
		}	

		glConfig2.getProgramBinaryAvailable = LOAD_EXTENSION_WITH_TEST(ARB_get_program_binary, formats > 0);
```

Instead,

```
		if ( formats == 0 )
		{
			logger.Notice("...GL_ARB_get_program_binary found, but with no binary formats");				
			glConfig2.getProgramBinaryAvailable = false;	
		} else {
			glConfig2.getProgramBinaryAvailable = LOAD_EXTENSION(ARB_get_program_binary);
		}
```

illwieckz:
In both snippets in your comment, the logger is written by the one who write the code, and the extension name has to be hardcoded. The whole purpose of this PR is to make sure those two things are never done by hand by the one who write the code, this to avoid any mistake, inconsistencies, or omission.

slipher:
What? The first snippet in my previous comment is code in the proposed change, written by you yourself.

illwieckz:
Ah, yes, that's the only use case where the logger is custom (we would do a custom logger for the Nvidia 340 test, for example), except those special cases, it's better to not have custom loggers. To be honest I was already looking for ways to get rid of the need to hardcode the ext name (maybe by extending macros) in that special case when you started to talk about reducing macros so I stopped there.

illwieckz:
Looking at `_WITH_TEST` without looking at `_WITH_CVAR` is misleading, because the TEST one is a variation of the CVAR one and they can be merged. Once merged there is 13 occurrences of `_WITH_TEST`.

Look at my work-in progress code that reduces to two macros. I haven't implemented the flag yet, I used enum instead because I wrote it before you suggested the flag.

slipher:
Somehow you have reverted the `r_noBorder` changes. Also please squash the fixup commits.

illwieckz:
OK, I fixed things, this is my current implementation with basic `enum`, I tried to do the flag thing the c++ way but it went nowhere.

illwieckz:
Anyway, about this:

> There is no need to write code like this:
> 
> ```
> 		if ( formats == 0 )
> 		{
> 			logger.Notice("...GL_ARB_get_program_binary found, but with no binary formats");				
> 			glConfig2.getProgramBinaryAvailable = false;	
> 		}	
> 
> 		glConfig2.getProgramBinaryAvailable = LOAD_EXTENSION_WITH_TEST(ARB_get_program_binary, formats > 0);
> ```
> 
> Instead,
> 
> ```
> 		if ( formats == 0 )
> 		{
> 			logger.Notice("...GL_ARB_get_program_binary found, but with no binary formats");				
> 			glConfig2.getProgramBinaryAvailable = false;	
> 		} else {
> 			glConfig2.getProgramBinaryAvailable = LOAD_EXTENSION(ARB_get_program_binary);
> 		}
> ```

The second one does not print what it it expected to be print.

 It is expected the user get this on his log:

```
...GL_ARB_get_program_binary found, but with no binary formats
..ignoring GL_ARB_get_program_binary.
```

The first notice is just the explanation for the log that comes after, I want this log printed in all case. First notice may be rewritten in a way this is printed:

```
...no binary formats found,
..ignoring GL_ARB_get_program_binary.
```

The same way the Nvidia workaround is planned to display something like that:

```
...found buggy Nvidia 340 driver,
..ignoring GL_ARB_texture_gather.
```

I want the `..ignoring` notice to always be printed by `LoadExt()`, if there is something custom to say, it would be said before calling `LoadExt()`. Calling `LoadExt()` in all case ensures the notices are consistent.

illwieckz:
So, basically in the end the code would looks like:

```c++
		if ( formats == 0 )
		{
			logger.Notice("...no binary formats found");
		}

		glConfig2.getProgramBinaryAvailable = LOAD_EXTENSION_WITH_TEST(ARB_get_program_binary, formats > 0);
```

illwieckz:
OK, this now looks like I want it to be (ie. with only two macros) but without the flags thing. I don't know how to do the flags things, it would be better to do it the flags way.

src/engine/sys/sdl_glimp.cpp
illwieckz: I removed those.

slipher:
I think you know how to use flags - I have seen some commits involving them before. For example in https://github.com/DaemonEngine/Daemon/pull/357/files the `IF_`* are a set of flags.

illwieckz:
Yeah, but those are simple C `enum`, not C++ `enum class`. In my WIP branch I add to define some operators etc…

I'm on it anyway, I wanted to push the implementation with booleans earlier so you can have a picture of what I'm doing, while keeping in mind that at some point `ext_NOCORE, ext_NOTREQUIRED` would be replaced by `ExtFlag::NONE` and `ext_CORE, ext_REQUIRED` would be replaced by `ExtFlag::CORE | ExtFlag::REQUIRED` or something like that.

illwieckz:
I just pushed an implementation using flags the C `enum` way. I'll attempt a C++ way but you can already look at the code to get a picture of how it's done.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -962,73 +962,52 @@ GLimp_InitExtensions
...
-	}
-	else
-	{
-		Sys::Error("Required extension GL_%s is missing", name );
-	}
-}
+enum {
+	ExtFlag_NONE,
+	ExtFlag_REQUIRED = BIT( 1 ),
+	ExtFlag_CORE = BIT( 2 ),
```
slipher, line 968: Please write a comment explaining the meaning of the CORE flag.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1131,60 +1108,31 @@ static void GLimp_InitExtensions()
...
-		{
-			logger.Notice("...GL_ARB_get_program_binary found, but with no binary formats");
-			glConfig2.getProgramBinaryAvailable = false;
-		}
-		else
+		if ( formats == 0 )
 		{
-			logger.Notice("...using GL_ARB_get_program_binary");
-			glConfig2.getProgramBinaryAvailable = true;
+			logger.WithoutSuppression().Notice("...no program binary formats");
```
slipher, line 1119: No need for `WithoutSuppression` for something which can only be printed once per renderer restart.

slipher (commented):
This looks pretty good now. Feel free to merge when you address the 2 review comments.

src/engine/sys/sdl_glimp.cpp
illwieckz: What about other notices? like `...using` or `...ignoring`? We don't use hundreds of extension, maybe we don't need to suppress them at all…

PR #363, opened 2020-08-23
renderer: disclose r_noBorder

illwieckz:
it looks like the code was ready but was missing a cvar to enable it

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -185,6 +185,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_subdivisions;
 	cvar_t      *r_stitchCurves;
 
+	cvar_t      *r_noBorder;
 	cvar_t      *r_fullscreen;
 
 	cvar_t      *r_customwidth;
@@ -1063,6 +1064,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_depthbits = ri.Cvar_Get( "r_depthbits", "0",  CVAR_LATCH );
 		r_ext_multisample = ri.Cvar_Get( "r_ext_multisample", "0",  CVAR_LATCH | CVAR_ARCHIVE );
 		r_mode = ri.Cvar_Get( "r_mode", "-2", CVAR_LATCH | CVAR_ARCHIVE );
+		r_noBorder = ri.Cvar_Get( "r_noBorder", "0", CVAR_ARCHIVE );
 		r_fullscreen = ri.Cvar_Get( "r_fullscreen", "1", CVAR_ARCHIVE );
 		r_customwidth = ri.Cvar_Get( "r_customwidth", "1600", CVAR_LATCH | CVAR_ARCHIVE );
 		r_customheight = ri.Cvar_Get( "r_customheight", "1024", CVAR_LATCH | CVAR_ARCHIVE );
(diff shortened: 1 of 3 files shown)


src/engine/renderer/tr_init.cpp
```diff
@@ -1063,6 +1064,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_depthbits = ri.Cvar_Get( "r_depthbits", "0",  CVAR_LATCH );
 		r_ext_multisample = ri.Cvar_Get( "r_ext_multisample", "0",  CVAR_LATCH | CVAR_ARCHIVE );
 		r_mode = ri.Cvar_Get( "r_mode", "-2", CVAR_LATCH | CVAR_ARCHIVE );
+		r_noBorder = ri.Cvar_Get( "r_noBorder", "0", CVAR_ARCHIVE );
```
slipher, line 1067: It seems to me, it should be CVAR_LATCH.

illwieckz: I agree but I remember `CVAR_LATCH` can't be set at startup yet (requires a restart after startup). We got the problem wiht `fs_legacyPaks` before.

This option is very likely to be set on command line of a launcher (users already know how to do it for other games like valve's ones).

slipher: It works fine for all the other window sizing cvars like `r_fullscreen` etc., so why is this one a problem?
Pak loading happens earlier than renderer initialization.

illwieckz: Well, `r_fullscreen` is currently not a `CVAR_LATCH`… this may have fooled me to think it would be like that early filesystem cvar…

Experience proves both `r_noBorder` and `r_fullscreen` can be `CVAR_LATCH`.

I flagged properly `r_noBorder` and added a commit to also flag `r_fullscreen`.

illwieckz: lol, now that `r_fullscreen` is `CVAR_LATCH`, pressing Alt+Enter does not toggle fullscreen…

illwieckz: So, after some investigations I figured that `r_noBorder` can be applied immediately on change (like `r_fullscreen`, so there I removed `CVAR_LATCH` flag and added code to apply it on change (like it is done for `r_fullscreen`).

While I was at it, I edited a bit the code to toggle fullscreen on `r_fullscreen` change, because it was using the cvar value as fullscreen flag, which was working by luck and was not raising any compiler because both were integer, but was wrong.

illwieckz:
I made some research and

- ioq3 (and derivatives like smokinguns, iodfe, etc.)
- etlegacy
- iortcw
- OpenJK
- Valve's source

use `r_noborder` cvar name but calll the option `borderless` in documentation, so we follow the trend (but we can camelCase it anyway).

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1428,6 +1429,23 @@ void GLimp_HandleCvars()
 
 		r_fullscreen->modified = false;
 	}
+
+	if ( r_noBorder->modified )
+	{
+		bool needToToggle = true;
```
slipher, line 1435: Drop this `needToToggle` check. I don't think it's important to avoid doing SetWindowBordered(true) twice if the user sets the cvar to 1 and then 2, and you already have the `modified` check to avoid setting it every frame.

slipher:
Tested the noBorder code and it works on my Windows system.

Changing the value of `r_fullscreen` does NOT work well - if you change it from 1 to 0 (while having r_mode = -1 ?), the application starts to ignore all user inputs. But this is a preexisting bug, so nothing need be done for this PR.

illwieckz:
> Tested the noBorder code and it works on my Windows system.

Thanks

> Changing the value of `r_fullscreen` does NOT work well - if you change it from 1 to 0 (while having r_mode = -1 ?), the application starts to ignore all user inputs. But this is a preexisting bug, so nothing need be done for this PR.

Yep, see #365

src/engine/sys/sdl_glimp.cpp
illwieckz: Oh, I missed this message…

illwieckz: Fixed, merged.

PR #366, opened 2020-08-29
binds: allow to set modcase commands without / like everywhere else, vstr users can use vstr

illwieckz:
Binds: allow to set modcase commands without / like everywhere else, vstr users can use `vstr`,
also make the code simpler, nowhere else the code requires ending `\n`.

Rogue commit stolen from dead #330.


slipher:
LGTM

PR #367, opened 2020-09-04
cleanup: dodge does not exist anymore, document buttons defined in game code

illwieckz:
cleanup: dodge does not exist anymore, document buttons defined in game code


slipher:
LGTM

PR #369, opened 2020-09-05
renderer/tr_mesh: silent “R_AddMDVSurfaces: no such frame -2147483648” debug log spam on md3 models used in particles

illwieckz:
I was really annoyed by such log spam, and I don't want to use `logs.suppression.enabled` for this (I would like to keep logs unsuppressed for other uses):

[md3 particle weird log spam](https://dl.illwieckz.net/b/daemon/bugs/md3-particle-weird-log-spam/unvanquished_2020-09-05_185012_001.png)

While adding some more logging, I noticed that with our md3 models, the `ent->e.frame` thing is either `0` (typically a third person weapon md3 model), either `-2147483648` (it only and always with md3 models used in particles, like weapon  shell or gibs). So, I decided to silent that use case and to keep the debug log for some unknown other situation that may happen.

This is a **HACK**, maybe it is hiding another bug somewhere else. I don't know, but I'm really fed up by this.

Note: we don't have animated md3 models in Unvanquished anymore so I was not able to look at how they behave.

If you know how to fix this issue at its root cause, please do it! At least, now you have an entry point to start investigations…

### DIFF
--- a/src/engine/renderer/tr_mesh.cpp
+++ b/src/engine/renderer/tr_mesh.cpp
@@ -269,6 +269,20 @@ void R_AddMDVSurfaces( trRefEntity_t *ent )
 		ent->e.frame %= tr.currentModel->mdv[ 0 ]->numFrames;
 		ent->e.oldframe %= tr.currentModel->mdv[ 0 ]->numFrames;
 	}
+	else if ( ent->e.frame == -2147483648 )
+	{
+		/* HACK: It looks like the previous test is always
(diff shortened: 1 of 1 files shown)


slipher:
I have a fix for this on the gamelogic side - https://github.com/Unvanquished/Unvanquished/pull/1212 (the trail systems commit).

illwieckz:
Oh nice!!!!

illwieckz:
A proper fix was made instead.

PR #370, opened 2020-09-06
sys/sdl_glimp: rewrite the workaround for Nvidia GT218 on 340 driver, fix #368

### LINKED ISSUE #368: GT218, Nvidia 340 failure is back on Quadro NVS 300
We hardcoded some string detection to fix the same issue with the Geforce 210 (See https://github.com/Unvanquished/Unvanquished/issues/971), but the driver string detection does not match the Quadro NVS 300 and then the bug is reproduceable on current `0.52.0/sync` branch.

We currently test for the `Geforce` string to apply the workaround and then `NVS` does not match.

I noticed both GPU are using the `340` version of the driver, it may be better to look for it instead.

illwieckz:
This may also match Quadro cards and not apply the workaround to all Nvidia products.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1059,11 +1059,36 @@ static void GLimp_InitExtensions()
 	// made required in OpenGL 3.0
 	glConfig2.textureRGAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_texture_rg, r_ext_texture_rg->value );
 
-	// made required in OpenGL 4.0
-	if( Q_stristr( glConfig.renderer_string, "geforce" ) ) {
-		glConfig2.textureGatherAvailable = false; // disabled on nVidia because some driver versions are bugged
-	} else {
-		glConfig2.textureGatherAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, ARB_texture_gather, r_arb_texture_gather->value );
+	{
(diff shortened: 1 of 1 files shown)


src/engine/sys/sdl_glimp.cpp
```diff
@@ -1076,8 +1076,19 @@ static void GLimp_InitExtensions()
...
+	{
+		/* Disable on Nvidia 340 driver because GT218-based GPU with
+		Nvidia 340.108 driver is know to fail on shader compilation.
+		see https://github.com/DaemonEngine/Daemon/issues/368
+		Unfortunately this workaround may also disable the feature for
+		GPUs that are not affected, but there is no known easy way to
+		detect GT218-based cards. Not all cards using 340 driver
+		supports this extension anyway, like the G92 one.
+		We can assume cards not using 340 driver are not GT218 ones
+		and are not affectd. */
```
slipher, line 1090: Typo - "affected"

slipher (approved):

illwieckz:
I want to properly log what happens, and merging #362 first would make it easy.

illwieckz:
Otherwise I just tested the code with an affected GPU and the detection works as expected.

illwieckz:
A `daemon.log` sample generated on Windows by @slipher shows that both `GL_VENDOR` and `GL_VERSION` strings are built-up the same way by the Nvidia proprietary driver on Linux and Windows, so this patch may also fix the issue on Windows if by any chance it also exists on Windows.

PR #374, opened 2020-09-12
Fill window with dark grey at creation time, fix #373

### LINKED ISSUE #373: Dæmon window is empty and undefined until shaders are compiled
Dæmon window is empty and undefined until shaders are compiled

On slow hardware it can take a long time and the window displays garbage for all this time…

[empty window](https://dl.illwieckz.net/b/daemon/bugs/empty-window/20200723-231745-000.unvanquished-empty-window.png)

Maybe there is a way for SDL to fill the window with black before this:

https://github.com/DaemonEngine/Daemon/blob/6b94e2e890f893afdc55b7f70bd58cb39f3cafcd/src/engine/renderer/tr_init.cpp#L333-L338

Or even earlier, right after that:

https://github.com/DaemonEngine/Daemon/blob/6b94e2e890f893afdc55b7f70bd58cb39f3cafcd/src/engine/sys/sdl_glimp.cpp#L702-L719

illwieckz:
Attempt to fill window with black at creation time, to prevent this (and fix #373):

[empty window](https://dl.illwieckz.net/b/daemon/bugs/empty-window/20200723-231745-000.unvanquished-empty-window.png)

It works but I know nothing about OpenGL and SDL, so maybe I'm introducing a time bomb…

@gimhael or @Kangz, is it OK to do that?

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -722,6 +722,11 @@ static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
 			}
 			SDL_GL_SetSwapInterval( r_swapInterval->integer );
 
+			// Fill window with a dark grey (#141414) background.
+			glClearColor( 0.08f, 0.08f, 0.08f, 1.0f );
+			glClear( GL_COLOR_BUFFER_BIT );
+			GLimp_EndFrame();
+
 			glConfig.colorBits = testColorBits;
 			glConfig.depthBits = depthBits;
 			glConfig.stencilBits = stencilBits;


illwieckz:
I've updated the code to not use a pitch black color but a dark grey one (`#141414`), because dark is really intimidating and can also give the impression of something being uninitialized.

The window looks like this until the main menu is displayed:

[empty window](https://dl.illwieckz.net/b/daemon/bugs/empty-window/20201004-055144-000.unvanquished-empty-window.png)

In fact, the best we can do is to just display a nice _“powered with Dæmon engine”_ picture in the middle, with a nice logo. Because this is expected to be displayed as early as possible, it would require to embed the picture in the engine and display it with OpenGL functions by hand. That would be perfect.

Anyway, I consider my PR ready, so please tell if something is wrong with my implementation.

illwieckz:
@cmf028, do you have something to say about this?

DolceTriade:
It's probalby fine to merge

illwieckz:
Yes, for what I now understand, this is not asking to really do something, but to set what must be the default color of something that is left untouched, then asking to not let untouched pixels uncolored, and to finish the frame. This is very unlikely to have any effect on subsequent draw calls, because anyway this code is expected to only do things on pixels that were never drawn.

We can see similar code there (which looks like to share a bit the purpose):

https://github.com/DaemonEngine/Daemon/blob/6b94e2e890f893afdc55b7f70bd58cb39f3cafcd/src/engine/renderer/tr_init.cpp#L881-L886

And there:

https://github.com/DaemonEngine/Daemon/blob/6b94e2e890f893afdc55b7f70bd58cb39f3cafcd/src/engine/renderer/tr_backend.cpp#L190-L201

Issue #376, opened 2020-10-02
Missing faces on Nvidia Curie (Geforce 6000/7000) running with Nvidia 304 driver and ATI R300 ones (regression)


illwieckz:
Only Nvidia Curie (Geforce 6000/7000) running with Nvidia 304 driver are affected, that's a regression.

This bug is a combination of two:

- #375 _Bad normal map channel detection, seems to be related to missing ARB_texture_compression_rgtc_
- a mistake introduced in #356 _Do not build GLSL shaders for disabled features_ especially in c5c18ce _do not build unused GLSL shader macro permutation_.

The bug makes face disappear or being badly rendered.

Unvanquished 0.51:

[![wrong normal format detection disables shader](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-29_011631_000.png)](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-29_011631_000.png)

[![wrong normal format detection disables shader](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-29_011648_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-29_011648_000.jpg)

Unvanquished 0.52:

[![wrong normal format detection disables shader](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-29_010422_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-29_010422_000.jpg)

[![wrong normal format detection disables shader](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-29_010437_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-29_010437_000.jpg)

[...]

Issue #375, opened 2020-10-02
Bad normal map channel detection, seems to be related to missing ARB_texture_compression_rgtc


illwieckz:
I noticed multiple bugs, some being regressions, on Nvidia Geforce 6000 and Geforce 7000 series. Those cards do not work with nouveau (`ARB_framebuffer_object` is missing and that's a requirement to run the game) but those card work with Nvidia 204 driver.

Nvida 304 driver is in really bad shape for Geforce 6000 series (most of them will just crash while the kernel complains they have `fallen off the buse`) but I haven't experienced such issue with a Geforce 7000 one.

Both Geforce 6000 and Geforce 7000 experience the bugs I've noticed.

While looking for the cause of the bug, I noticed this in `daemon.log`:

```
...GL_ARB_texture_compression_rgtc not found
```

Over the 40+ cards [that were tested](https://wiki.unvanquished.net/wiki/GPU_compatibility_matrix), no other one reproduces the bug, but also no one misses that extension. In fact the only other GPUs known to miss that extensions are the ones from the R300 series by ATI and they are experiencing another bug that prevent maps to load at all (see #345) so that other bug is likely to hide on ATI side the one I disscuss in this thread.

What I noticed is while the engine tells this extension is missing, the engine also prints wrong things in logs:

```
Debug: found heightmap embedded in normalmap 'textures/shared_pk02_src/floor12_n'
Debug: found heightmap embedded in normalmap 'textures/shared_pk02_src/floor12_n'
Debug: found heightmap embedded in normalmap 'textures/shared_pk02_src/floor09_n'
Debug: found heightmap embedded in normalmap 'textures/shared_pk02_src/floor06_n'
Debug: found heightmap embedded in normalmap 'textures/shared_pk02_src/floor05_n'
Debug: found heightmap embedded in normalmap 'textures/shared_pk02_src/floor04_n'
Debug: found heightmap embedded in normalmap 'textures/shared_pk02_src/floor03_n'
Debug: found heightmap embedded in normalmap 'textures/shared_pk02_src/floor01_n'
```

The wrong thing in those logs is that those textures do not ship height map in normal map file.
When an height map is shipped in a normal map file, the height map is stored in alpha channel, this is not the case here.

We have to detect if the height map is stored in alpha channel for two reasons:

- to use the height map
- to properly read the normal map

So if either height mapping or normal mapping is enabled, we have to detect if the normal map ships an height map.

[...]

PR #378, opened 2020-10-04
tr_image: add debug log for undocumented formats and add more code names when listing images

illwieckz:
Some improvements to help debugging:

- add debug log for unknown formats when listing images,
- add some code names for RGTC variant when listing images,
- add code name for compressed RGB when listing images.

Note that I made up myself the code names.

This is the way I did them:

- since `a` in `DXT1a` is for `alpha` channel, I made `RGTC1r` and `RGTC2rg` with `r` and `g` for `red` and `green`,
- since `F` in `RG32F` is for `float` type, I made `RGTC1Sr` and `RGTC2Srg` with `S` for `signed`,
- since `C` in `RGTC` if for `compression`, I made `RGBAC` with `C` for `compressed`.

I made this to investigate #375 and #376.

Before this PR:

```
 209:  128  128  yes   2D   ????     s.rept  t.rept   textures/shared_pk02_src/floor01_n
```

With debug logs added:

```
Debug: Undocumented image format 36285 for image textures/shared_pk02_src/floor01_n
 209:  128  128  yes   2D   36285    s.rept  t.rept   textures/shared_pk02_src/floor01_n
```

With new format code names:

```
 209:  128  128  yes   2D   RGTC2rg  s.rept  t.rept   textures/shared_pk02_src/floor01_n
```

Note that I don't know at all what `imageDataSize` is for, also, it looks like the game did not required it to be set for every formats.
So, I just copy-pasted the `DXT5` value, given `RGTC` looks to be a variant of `DXT5`. I flagged every of those lines so we know how to return at this code later. @gimhael, do you know something about it?

I'm in favor of merging this even if those values are not known to be good, because anyway before those commits those values were not set so the chance they were wrong was the same, and anyway those values seem to not be changed from the default one. This is probably not that hurting and it's flagged correctly as `FIXME`. On the other side, the addition of those debug logs and new code names are needed to make easier to debug the game (for example, #375).

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -329,6 +329,30 @@ void R_ImageList_f()
 				imageDataSize *= 4 / 4;
 				break;
 
+			case GL_COMPRESSED_RED_RGTC1:
+				Com_sprintf( buffer, sizeof( buffer ),  "RGTC1r   " );
+				out += buffer;
+				// TODO: find imageDataSize
+				break;
+
+			case GL_COMPRESSED_SIGNED_RED_RGTC1:
+				Com_sprintf( buffer, sizeof( buffer ),  "RGTC1Sr  " );
+				out += buffer;
+				// TODO: find imageDataSize
+				break;
+
+			case GL_COMPRESSED_RG_RGTC2:
+				Com_sprintf( buffer, sizeof( buffer ),  "RGTC2rg  " );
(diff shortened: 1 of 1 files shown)


slipher:
I think imageDataSize is supposed to be the size on disk. For the unknown ones, I'd rather see a simple `TODO: find imageDataSize` rather than the mysterious-looking `4 / 4` calculation.

LGTM

PR #379, opened 2020-10-04
tr_shader: do not detect heightmap in normalmap alpha channel, unless asked to, fix #375

### LINKED ISSUE #375: Bad normal map channel detection, seems to be related to missing ARB_texture_compression_rgtc

illwieckz:
With this PR,

- `normalMap` material keyword loads a file and assume it's a normal map without height map in alpha channel (even if the engine believes there is an alpha channel because of a format being wrongly guessed on missing GL extension, see #375).
- `normalHeightMap` material keyword loads a file it's a normal map with an height map in alpha channel.

This code also introduces an internal `ParseNormalMapDetectHeightMap` function that attempts to detect an alpha channel and declare it's an heightmap. There is no material keyword for that and it must not be implemented as a keyword.

This internal function is only purposed to be used by compatibility code like the Darkplaces compatibility mode. Darkplaces-based games being known to not use formats that would break on detection, it's seems to be safe to use this detection function for their assets.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -1506,16 +1506,6 @@ static bool LoadMap( shaderStage_t *stage, const char *buffer, const int bundleI
 		}
 	}
 
-	// tell renderer to enable relief mapping since an heightmap is found
-	// also tell renderer to not abuse normalmap alpha channel because it's an heightmap
-	// https://github.com/DaemonEngine/Daemon/issues/183#issuecomment-473691252
-	if ( stage->bundle[ bundleIndex ].image[ 0 ]->bits & IF_NORMALMAP
-		&& stage->bundle[ bundleIndex ].image[ 0 ]->bits & IF_ALPHA )
-	{
-		Log::Debug("found heightmap embedded in normalmap '%s'", buffer);
-		stage->isHeightMapInNormalMap = true;
-	}
-
 	return true;
 }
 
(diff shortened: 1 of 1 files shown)


illwieckz:
@slipher do you have any remark on it? I would appreciate some remark on pointer wizardry to keep an unmodified copy of `**text` if there is a better implementation possible.

slipher:
I suggest as follows. Notes:

- DoDebugCode is used to execute the code only if the log level is `debug`. 
- `char buffer[1024] = "";` requires the entire buffer to be zeroed.

```cpp
	const char* initialText = *text;

	ParseNormalMap( stage, text, bundleIndex );

	/* Tell renderer to enable relief mapping since an heightmap is found
	also tell renderer to not abuse normalmap alpha channel because it's an heightmap.
	See https://github.com/DaemonEngine/Daemon/issues/183#issuecomment-473691252 */
	if ( stage->bundle[ bundleIndex ].image[ 0 ]->bits & IF_NORMALMAP
		&& stage->bundle[ bundleIndex ].image[ 0 ]->bits & IF_ALPHA )
	{
		Log::defaultLogger.DoDebugCode([&] {
			char buffer[ 1024 ];
			buffer[ 0 ] = '\0'
			bool parsed = ParseMap( &initialText, buffer, sizeof( buffer ) );
			ASSERT( parsed );
			Log::Debug("Found heightmap embedded in normalmap '%s'", buffer);
		});

		stage->isHeightMapInNormalMap = true;
	}
```

illwieckz:
@slipher, I rewrote the things accordingly. I added a `UNUSED` keyword to silent the compiler complaining about `parsed` being unused on non-debug code. Or I may hack something like that instead:

```c
			#ifdef DEBUG
			bool parsed = 
			#endif
			ParseMap( &initialText, buffer, sizeof( buffer ) );
```

slipher:
There is already a facility for marking an unused variable, `Q_UNUSED`. It's used like `Q_UNUSED(variable);`

Alternatively `if(!parsed) ASSERT(false);` would work.

illwieckz:
Hmm, since all those tricks are there to move the `ParseMap` call out of `ASSERT`, then we can just avoid a `bool` and do:

```c
			if ( !ParseMap( &initialText, buffer, sizeof( buffer ) ) )
			{
				ASSERT( false );
			}
```

slipher (approved):

PR #377, opened 2020-10-03
disable useless heightMapInNormalMap if both normal and relief mapping are disabled, also fix #376

illwieckz:
Disable useless heightMapInNormalMap if both normal and relief mapping are disabled, fix #376.

Before:

[wrong normal format detection disables shader](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-29_010422_000.jpg)

After:

[wrong normal format detection disables shader](https://dl.illwieckz.net/b/daemon/bugs/wrong-normal-format-detection-disables-shader/unvanquished_2020-09-30_234455_000.jpg)

This bug is known to affect Nvidia Curie (Geforce 6000/7000 series) with driver Nvidia 304.

Root cause of #376 is fixed in #385. This commit is hardening the fix anyway.

PR #381, opened 2020-10-05
Do not use lightTile code when tiled renderer is disabled, fix rendering on Radeon R300 with tiny ALU, fix #344

### LINKED ISSUE #344: r300 FP: Compiler Error: r300_fragprog_emit.c::emit_alu(): Too many ALU instructions on Radeon X1050 (RV370)
The [Unvanquished download page](https://unvanquished.net/download/) states the minimal requirement on GPU side is an OpenGL 2.1 GPU.

I've driven some tests to check that, using the to-be-released 0.52 engine. I tested both the Intel GMA965 X3100 (integrated in Core 2 Duo L7500 CPU) and the ATI X1050 (discrete PCIe card). They're both from around the year 2006.

While the test succeeded with the Intel GMA964 chip (I had to reduce texture resolution though), the test failed miserably with the ATI X1050 card.

The ATI X1050 GPU uses the architecture that existed before _TeraScale_ (papers may have named it “_Unified Superscalar Shader Architecture_” at the time).

I got this error in Dæmon's output:

```
r300 FP: Compiler Error:
../src/gallium/drivers/r300/compiler/r300_fragprog_emit.c::emit_alu(): Too many ALU instructions
Using a dummy shader instead.
r300 FP: Compiler Error:
build_loop_info: Cannot find condition for if
Using a dummy shader instead.
```

I've read interesting things about [a similar issue on another project](https://bugs.freedesktop.org/show_bug.cgi?id=35257) and about fixes of similar issues I have [read that](https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/43cf60f5636e7883bd5012e4c89ed5d57e28200c?view=parallel):

> The number instructions in a shader is limited to 64 on r300 hardware,
the fade shader in StScrollViewFade was ending up using 97 instructions
which is way over the limit.
> 
> So refactor the shader to use less instructions by precomputing as many
values as possible outside of the conditionals. The resulting shader
ends up using 34 instructions which is well within the hardware limits.

I still don't understand why in this example doing this avoids the issue:

```
result1 = compute1()
result2 = compute2()
if (something) {
  result = result1;
}
else {
  result = result2;
}
```

but doing this would fail:

```
if (something) {
  result = compute1();
}
else {
  result = compute2();
}
```

So, I don't really know how to fix the issue myself.

Anyway, since Xonotic/DarkPlaces runs on that GPU there is no reason Unvanquished/Dæmon cannot. This issue keeps track of this bug if someone wants to hack on it.

In all case, we would want to catch properly the issue and at least return to main menu with a proper message. That's for another ticket (see #345).

On a side note, I experienced hangs with that GPU but maybe I just need to reduce texture resolution like I had to do with the Intel X3100 GPU, that's for another ticket (see #346).

illwieckz:
Do not call `RB_RenderPostDepth` when the tiled renderer is disabled. It looks like this function added in 0bb3550ce16ab7b165a88ebb365edf6b2837d1f5 is only used when the tiled renderer is enabled.

The light tile code is known to be too large for the tiny ALU of the Radeon R300 GPU. See #344.

With this simple modification, the R300s GPU like the X300, X550HM and X1050 become able to run the game.

Also rename `RB_RenderPostDepth` to `RB_RenderPostDepthLightTile` to make this characteristic more obvious when reading the code.

Also disable more light tile code when the tiled renderer is disabled, for better performance in that case.

With that PR merged, we know no one OpenGL 2.1+ hardware that is not able to render Unvanquished.

I would be happy to get a comment by @gimhael since I touched his code. Anyway that looks safe because I tested this code both when the tiled renderer was enabled and disabled without noticing any issue. Maybe some people has some idea for better implementation…?

To properly render the game on those GPUs, merging #377 is also required (bug fix) since those cards are also affected by #375, hence #376.

Here are some screenshots taken with a Radeon X1050 on Ubuntu 20.04, Linux 5.4.0 and Mesa 20.0.8.

As seen on a typical desktop (performance is greatly affected by a VNC server):

[r300 small alu](https://dl.illwieckz.net/b/daemon/bugs/r300-small-alu/20201005-045740-000.unvanquished-r300-small-alu.png)

The common Plat23 scene:

[r300 small alu](https://dl.illwieckz.net/b/daemon/bugs/r300-small-alu/unvanquished_2020-10-05_065952_000.jpg)

Some other scenes, we can see that like other OpenGL 2.1 cards, performance are heavily affected by models, especially alien ones. See https://github.com/Unvanquished/Unvanquished/issues/1207 about that.

[r300 small alu](https://dl.illwieckz.net/b/daemon/bugs/r300-small-alu/unvanquished_2020-10-05_070000_000.jpg)

[r300 small alu](https://dl.illwieckz.net/b/daemon/bugs/r300-small-alu/unvanquished_2020-10-05_070006_000.jpg)

[r300 small alu](https://dl.illwieckz.net/b/daemon/bugs/r300-small-alu/unvanquished_2020-10-05_071154_000.jpg)

[r300 small alu](https://dl.illwieckz.net/b/daemon/bugs/r300-small-alu/unvanquished_2020-10-05_081200_000.jpg)

[r300 small alu](https://dl.illwieckz.net/b/daemon/bugs/r300-small-alu/unvanquished_2020-10-05_072501_000.png)

[r300 small alu](https://dl.illwieckz.net/b/daemon/bugs/r300-small-alu/unvanquished_2020-10-05_072615_000.jpg)

[r300 small alu](https://dl.illwieckz.net/b/daemon/bugs/r300-small-alu/unvanquished_2020-10-05_072619_000.jpg)

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -2758,7 +2758,7 @@ void RB_RunVisTests( )
 	}
 }
 
-void RB_RenderPostDepth()
+void RB_RenderPostDepthLightTile()
 {
 	static vec4_t quadVerts[4] = {
 		{ -1.0f, -1.0f, 0.0f, 1.0f },
@@ -2769,7 +2769,31 @@ void RB_RenderPostDepth()
 	vec3_t zParams;
 	int w, h;
 
-	GLimp_LogComment( "--- RB_RenderPostDepth ---\n" );
+	GLimp_LogComment( "--- RB_RenderPostDepthLightTile ---\n" );
+
+	if ( r_dynamicLight->integer != 2 )
+	{
+		/* Do not run lightTile code when the tiled renderer is not used.
+
+		This computation is part of the tiled dynamic lighting renderer,
+		it's better to not run it and save CPU cycles when such effects
+		are disabled.
+
+		Disabling this code also make possible to not compile the related
+		GLSL shaders at all when such effects are disabled.
+
+		Not running the related GLSL shaders also helps older hardware to
+		run the game, for example the Radeon R300 Arithmetic Logic Unit is
+		too small to run the related GLSL code even if the shader itself
+		can be compiled. Such GPU are so old and slow that	any kind of
+		dynamic lighting including the tiled implementation is expected to
+		be disabled anyway. Saving CPU cycles when a feature is not used is
+		welcome in any case.
+
+		See https://github.com/DaemonEngine/Daemon/issues/344 */
+
+		return;
+	}
 
 	if ( ( backEnd.refdef.rdflags & RDF_NOWORLDMODEL ) )
 	{
@@ -2843,7 +2867,10 @@ void RB_RenderPostDepth()
 		if( !glConfig2.glCoreProfile )
 			glEnable( GL_POINT_SPRITE );
 		glEnable( GL_PROGRAM_POINT_SIZE );
+
+		// Radeon R300 small ALU is known to fail on this.
 		Tess_DrawArrays( GL_POINTS );
+
 		glDisable( GL_PROGRAM_POINT_SIZE );
 		if( !glConfig2.glCoreProfile )
 			glDisable( GL_POINT_SPRITE );
@@ -4656,7 +4683,7 @@ static void RB_RenderView( bool depthPass )
 	if( depthPass ) {
 		RB_RenderDrawSurfaces( shaderSort_t::SS_DEPTH, shaderSort_t::SS_DEPTH, DRAWSURFACES_ALL );
 		RB_RunVisTests();
-		RB_RenderPostDepth();
+		RB_RenderPostDepthLightTile();
 		return;
 	}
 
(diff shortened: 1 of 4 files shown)


src/engine/renderer/tr_backend.cpp
```diff
@@ -2769,7 +2769,20 @@ void RB_RenderPostDepth()
 	vec3_t zParams;
 	int w, h;
 
-	GLimp_LogComment( "--- RB_RenderPostDepth ---\n" );
+	GLimp_LogComment( "--- RB_RenderPostDepthLightTile ---\n" );
+
+	/* Do not run lightTile code when the tiled renderer is not used.
+
+	Especially since the related GLSL code cannot run on the small
```
slipher, line 2776: It would be more relevant to say why the code accomplishes nothing/is not needed when the tiled renderer is not used, rather than some information about old hardware. It goes without saying that it is good to skip the code if it doesn't do anything.

illwieckz:
rewrote comments

illwieckz:
@slipher do you have something more to say?

slipher:
go ahead

Issue #365, opened 2020-08-28
Engine often fails to grab input on vid_restart or r_fullscreen toggling


illwieckz:
I experience this bug a lot were the game fails to grab input after `vid_restart`. The grab failure is easy to reproduce by putting the game window behind another unrelated window while the reloading is going on, waiting enough to make sure the whole reload process complete (and map is displayed again) before putting the game window on foreground.

Also, toggling fullscreen / windowed mode is likely to reproduce the bug, especially going from fullscreen to windowed mode.

Better a load a map first to reproduce the bug.

A `vid_restart` call while keeping the game in foreground usually grabs input again (but I'm afraid it may only be done on a TTY console in some case).

illwieckz:
It looks like the bug also occurs when Alt-Tabbing from fullscreen: input is lost when returning to the game.

illwieckz:
This bug is critical. Players quit the game in a middle of playing because of this. @slipher you seems to know a bit the input code, do you have any idea about that bug?

If we don't know how to fix it we better workaround it by calling `vid_restart` on Alt-Tab. Releasing something with such bug in it would be a disaster.

bofh80:
Right, according to this : https://discourse.libsdl.org/t/problem-with-switching-between-full-screen-to-window-mode/24689

> You’ll need to use: SDL_SetWindowFullscreen(mWindow, 0); to get out of fullscreen mode, and then use: SDL_SetWindowSize(SDL_Window* window, int w, int h) to set the size of the screen. 

So it seems you need to tell SetWindowSize and it should start working properly. Note a comment from the bottom

> Calling SDL_SetWindowSize(SDL_Window* window, int w, int h) in the start of the next frame was the solution.

I had a look at adding the SetWindowSize, but i'm not sure how it relates to GLimp_StartDriverAndSetMode or GLimp_SetMode. Maybe you want the setwindowsize in one of those functions, and call it when the fullscreen switches back to windowed mode.

Best I could find, and it does seem quite relevant.

illwieckz:
Related issues:

- https://github.com/DaemonEngine/Daemon/issues/234
- https://github.com/DaemonEngine/Daemon/issues/408
- https://github.com/Unvanquished/Unvanquished/issues/1136

@bofh80 interesting!

illwieckz:
Fixed by DaemonEngine/Daemon#409.

illwieckz:
[...]

PR #382, opened 2020-10-06
Fix fullscreen toggling bug (#365)

slipher:
This makes it possible to toggle r_fullscreen more than once without getting stuck.

Still want to look at a couple other things:

- Delete the code that makes (undesirable) changes to window cvars in response to SDL window resize events
- I noticed that toggling r_noBorder repeatedly causes a growing black space in the window.

### DIFF
--- a/src/engine/framework/CvarSystem.cpp
+++ b/src/engine/framework/CvarSystem.cpp
@@ -81,7 +81,7 @@ namespace Cvar {
                 modified = true;
             } else {
                 if (Q_stricmp(var.string, cvar.value.c_str())) {
-                    Log::Notice("The change will take effect after restart.");
+                    Log::Notice("The change to %s will take effect after restart.", var.name);
                     if (var.latchedString) Z_Free(var.latchedString);
                     var.latchedString = CopyString(cvar.value.c_str());
                     modified = true;
@@ -283,7 +283,7 @@ namespace Cvar {
                         OnValueChangedResult undo = cvar->proxy->OnValueChanged(cvar->value);
                         ASSERT(undo.success);
                         if (setLatchedValuesCalled) {
-                            Log::Notice("The change will take effect after restart.");
+                            Log::Notice("The change to %s will take effect after restart.", cvarName);
                         }
                         cvar->latchedValue = value;
                         return;
--- a/src/engine/null/null_renderer.cpp
+++ b/src/engine/null/null_renderer.cpp
@@ -194,7 +194,6 @@ bool RE_BeginRegistration( glconfig_t *config, glconfig2_t *glconfig2 )
 	Com_Memset( config, 0, sizeof( glconfig_t ) );
 	config->vidWidth = 640;
 	config->vidHeight = 480;
-	config->windowAspect = 1.0f;
 	Com_Memset( glconfig2, 0, sizeof( glconfig2_t ) );
 
 	return true;
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -190,7 +190,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 	cvar_t      *r_customwidth;
 	cvar_t      *r_customheight;
-	cvar_t      *r_customaspect;
 
 	cvar_t      *r_debugSurface;
 	cvar_t      *r_simpleMipMaps;
@@ -469,7 +468,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	};
 	static const int s_numVidModes = ARRAY_LEN( r_vidModes );
 
-	bool R_GetModeInfo( int *width, int *height, float *windowAspect, int mode )
+	bool R_GetModeInfo( int *width, int *height, int mode )
 	{
 		const vidmode_t *vm;
 
@@ -486,13 +485,11 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 		if( mode == -2)
 		{
 			// Must set width and height to display size before calling this function!
-			*windowAspect = ( float ) *width / *height;
 		}
 		else if ( mode == -1 )
 		{
 			*width = r_customwidth->integer;
 			*height = r_customheight->integer;
-			*windowAspect = r_customaspect->value;
 		}
 		else
 		{
@@ -500,7 +497,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 			*width = vm->width;
 			*height = vm->height;
-			*windowAspect = ( float ) vm->width / ( vm->height * vm->pixelAspect );
 		}
 
 		return true;
@@ -1068,7 +1064,6 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_fullscreen = ri.Cvar_Get( "r_fullscreen", "1", CVAR_ARCHIVE );
 		r_customwidth = ri.Cvar_Get( "r_customwidth", "1600", CVAR_LATCH | CVAR_ARCHIVE );
 		r_customheight = ri.Cvar_Get( "r_customheight", "1024", CVAR_LATCH | CVAR_ARCHIVE );
-		r_customaspect = ri.Cvar_Get( "r_customaspect", "1", CVAR_LATCH );
 		r_simpleMipMaps = ri.Cvar_Get( "r_simpleMipMaps", "0", CVAR_LATCH );
 		r_subdivisions = ri.Cvar_Get( "r_subdivisions", "4", CVAR_LATCH );
 		r_dynamicLightCastShadows = ri.Cvar_Get( "r_dynamicLightCastShadows", "1", 0 );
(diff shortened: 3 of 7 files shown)


illwieckz:
I'm still experiencing issues, maybe differently.

I noticed that when I'm windowed after having been fullscreen, alt-tabbing out and in restore input, but alt-tabbing out and in does not restore input when I'm fullscreen after having been windowed.

I noticed I get this log everytime I go windowed from fullscreen:

```
The change to r_customwidth will take effect after restart. 
The change to r_customheight will take effect after restart. 
```

illwieckz:
Note: My game height and width is not my desktop resolution. My desktop resolution is 4K and my game resolution is 2K. So when I go fullscreen, the game changes the desktop mode, and restore desktop mode when going windowed again.

illwieckz:
I have some ideas for Dæmon/Unvanquished.

- having a window height / width different than fullscreen (which requires to call vid_restart when going windowed and back)
- leveraging the fake SDL fullscreen (fullscreen window atop everything)

But implementing those two features may hide the bug. So I want the bug to be fixed before someone investigate this.

I want you to know this because it may change the way your fixes are implemented. For example we can decide to drop classic fullscreen entirely or things like that. To be honest I've experienced no performance difference between windowed and classic fullscreen, even on 16 years old hardware. But I want to be sure we don't hide bugs because we delete a feature that uncovers it.

See https://wiki.libsdl.org/SDL_SetWindowFullscreen and `SDL_WINDOW_FULLSCREEN_DESKTOP`.

slipher:
> alt-tabbing out and in does not restore input when I'm fullscreen after having been windowed.

Oh. Alt-Tab has always worked on Windows, so when you said this in the bug I assumed it was a typo for Alt-Enter (which toggles r_fullscreen).

> I noticed I get this log everytime I go windowed from fullscreen:

Yes, this is the first bullet point in the first post.

> My desktop resolution is 4K and my game resolution is 2K. So when I go fullscreen, I imagine the game changes the desktop mode
It does not. The resolution cannot be changed without a restart. So it can only stretch or truncate when changing size.

slipher:
How about this for Alt+Enter (which currently just toggles r_fullscreen):

- If r_fullscreen is toggled to 1, set r_mode to -2 and vid_restart.
- If r_fullscreen is toggled to 0, set r_mode to -1, vid_restart and use the previously stored value of r_customwidth/height. If there is no previous value we could use half the screen dimensions.

illwieckz:
To me dealing with things like that is for after 0.52, but we need input fixed for 0.52. I have many ideas about user experience that would decide the implementations, but I don't focus on them for the moment.

Can you remind me what does `r_mode -2`?

slipher:
r_mode -2 makes the resolution the same as the screen resolution. It is the default.

illwieckz:
Oh, so no, we really have to keep the ability to go fullscreen with a smaller resolution, this is one of the first need for full screen along with displaying nothing else on screen. For example people may play the game in 2K in full screen on a 4K screen to get higher framerate or enhanced graphical effects if their GPU don't keep up.

illwieckz:
So, to return to the topic: yes, there was three bugs (at least on Linux with GNOME Shell):

- input is lost when going from fullscreen to window (or back) [usually with Alt+Enter],
- input is lost when switching between apps when the game is fullscreen [usually with Alt+Tab],
- input is lost when the game is entering map after the map load while the window is not focused.

Alt-tabbing with windowed game does not lose input. but having the game windowed and not in focus while loading the map will make the game loses input.

There is some interesting combinations, with this PR, if I go from fullscreen to window (with [Alt+Enter]) I lose input. But if I cycle between windows and come back to the windowed game (with [Alt+Tab]), I recover input.

But if go from window to fullscreen (with [Alt+Enter]), I cannot recover input, and if I cycle between windows, instead of recovering input all I got is a painted desktop pointer painted over the screen (which I can control instead of the game itself):

[fullscreen window lost input](https://dl.illwieckz.net/b/daemon/bugs/fullscreen-window-lost-input/20201006-214546-000.unvanquished-fullscreen-window-lost-input.png)

It looks like your PR fixes the third point (input is lost when the game is entering map after the map load while the window is not focused).

illwieckz:
While running code from your PR, I noticed than when I cycle between windowed and fullscreen using [Alt+Enter], I can recover input by cycling between game and in game UI, for example when being spectator, clicking to open the team selection menu and pressing [Escape] to close that menu recovers input. If I'm already in a team, pressing [Escape] to display the main menu and [Escape] again to hide it recovers input.

Those tweaks do not work at all when input is lost when cycling between windows with [Alt+Tab].

slipher:
I fixed the thing that sets cvars when you toggle r_fullscreen. I'm happy enough with how it works now on my own machine. Later I'll try it in a Linux VM.

illwieckz:
So, to sum up the current status:

- input is lost when going from fullscreen to window (or back) [usually with Alt+Enter],
→ half fixed (display/hide UI trick required)
- input is lost when cycling between apps when the game is fullscreen [usually with Alt+Tab],
→ still broken
- input is lost when the game enters map after map load while the window is not focused.
→ fixed

I also noticed that the game frequently resets my custom height / width to my screen size on startup. It seems to be related to the fact I've cycled between windowed and fullscreen mode on the last run. Maybe this bug was introduced with that PR.

Note that I'm using GNOME Shell on Xorg on Ubuntu 20.04. I guess there is no need to get a real GPU in a virtual machine, llvmpipe-based OpenGL may be enough to reproduce the big.

illwieckz:
Hmmm, I managed to load the game with a map on command line with a “fullscreen” game being minimized, and I cannot move the mouse. It's probably just the [Alt-Tab] bug, but it's interesting to notice I did not switched to any other window first, game started on fullscreen, but minimized.

About the current status of input issue, it looks like keyboard keys work, mouse button work, it's the mouse movement that is not captured, only that.

slipher:
I did a lot of Alt+Enter, Alt+Tab and `vid_restart` on this branch in a Debian bullseye VM with GNOME desktop, but nothing bad happened. I also tried starting a devmap while minimized. How exactly do you produce bugs?

> I also noticed that the game frequently resets my custom height / width to my screen size on startup.

This PR should fix that, as long as you have `r_allowResize` at the default of 0.

illwieckz:
> How exactly do you produce bugs?

Exactly the same way you attempted to: Alt+Enter, Alt+Tab and `vid_restart`… :confused: 

I'm using Ubuntu 20.04 with GNOME Shell…

> > I also noticed that the game frequently resets my custom height / width to my screen size on startup.
> This PR should fix that, as long as you have r_allowResize at the default of 0.

I experiment the opposite, I started to experiment that bug starting with this PR. Also `r_allowResize` is `0` on my end.

To reproduce it I can do this way:

- start the game windowed with a resolution smaller than the screen itself (for example 2k resolution on 4k screen), typically the config comes from a previous run
- switch to fullscreen with Alt+Enter (original resolution seems to be kept)
- return to windowed mode with Alt+Enter (original resolution is kept)
- quit the game
- restart the game

The game starts windowed but with fullscreen resolution.

slipher:
Are you sure you are testing with the right code? I pushed the `r_allowResize`-related commit later than the other ones.

illwieckz:
Hmm, I already knew GitHub does nasty things with commit ordering etc. but well, I missed two commits. I don't know yet for that resize issue, but input issue is still there. I just started the game both on background and fullscreen, I had no input in game when I alt-tabbed to display the game (UI trick did not recovered it). When I went windowed with Alt+Enter I had input in window mode, when I went fullscreen again I lost input (but UI trick recovered it). I Alt-Tabbed in and out and then lost input again (without being to recover it with UI trick). This was a brand new profile.

illwieckz:
Note: I use GNOME Shell with Xorg, not Wayland, maybe there is a difference between both?

slipher:
I tried with X instead of Wayland, but didn't managed to reproduce your bugs.

I did find one other way for r_custom* to get changed. If you set r_mode to something other than -1 (are you doing that?), then the engine sets r_custom* get set to whatever the actual resolution is. The way this is done now is a bit buggy as the currently used resolution is displayed as the "latched value", even though it is already in effect. I'm not sure what the best behavior is for when you use an r_mode not equal to -1:

1. Set r_custom* in a way that overrides CVAR_LATCH to be immediately displayed as the current value. This might be preferred so that the user can always inspect the current resolution by looking at the cvars.
2. Do not set r_custom*. This might be preferred so that the windowed resolution can be "remembered" while you are using `r_mode -2`.
3. Set the current value of the cvars to the current resolution, and the latched value to whatever the user manually set last. This is an attempt to get the best of both worlds.

I also found one other input bug. If I alt-tab out during a restart and alt-tab in after it finishes, I get an unwanted system cursor displayed (in addition to the librocket one). This condition can be cleared by cycling alt-tab again.

slipher:
I implemented option (1) from the previous comment.

It would be interesting to see if any remaining bugs are improved by the SDL update in #401.

illwieckz:
I get the exact same behaviour with #401 alone and #410 plus this one. Most of the time the only issue is the one when I have to click to recover input (that is less bad than before).

But while toying with a fresh profile, I reproduced the bug of I losing mouse input entirely and having my desktop mouse move over the game window… :-/

Anyway you can merge both PR as they are both steps forward, maybe rename this one to “Fix some fullscreen toggling bugs” because there may still be dragons, harder to find though.

illwieckz:
Uh, forget what I said about “harder to find”, it's still really easy to reproduce that issue on my end, fresh or non-fresh profile, I just have to Alt-Tab from fullscreen then Alt-Tab back.

PR #383, opened 2020-10-08
renderer: rewrite 8ba2505 in a better way, do not build unused GLSL shaders

illwieckz:
This is just 8ba2505033014de931b5a00525b6b04b0b6d0e08 rewritten in a better way: no need to pop from a list something that is never added to the list to begin with.

Also maybe the `bloom` string for bloom was wrong since the shader is named `screen`.


illwieckz:
This was tested on a large range of hardware and works as expected. Part of #381 can be rewritten in a better way once this is merged.

PR #389, opened 2020-10-20
some revamp on IQM and MD5 model code

illwieckz:
I was looking to see if there was a way to reduce engine side the impact of some model slowing down a lot the game on some older hardware (see https://github.com/Unvanquished/Unvanquished/issues/1207) so I incrementally did minor changes over minor changes to make it easier to read, understandable and refactorisable and in the end that was what I got.

Now I know why the performance drops so much, and any improvement to this code can't do magic. Anyway, this code now looks better and may even save some cycles (or rely less on compiler optimization). Also, two loops were rewritten at once in a way the code only iterate once for both.

I reported such changes to the MD5 model code and then it gave me some example to clean-up the IQM code a bit more in return.

There is two changes I have committed separately (see commits about `memcpy` and `memcmp`), that's two alternative implementation for a given code, those two commits are there for a question: is there something to gain from them? The idea why I thought about those alternate implementation is that we may assume `memcmp` and `memcpy` may have specific optimized code from standard library. I would like to know if it is better or not, so I can squash them or drop them.

Also, I made `r_vboModels` cvar a latched one, because it is required to restart the engine to see a change.

Basically, people can set `r_vboModels` to `0` to reproduce https://github.com/Unvanquished/Unvanquished/issues/1207 on any hardware.

The issue is that some Unvanquished models have too much bones for the hardware so some fast OpenGL feature cannot be used, despite the feature is supported.

### DIFF
--- a/src/engine/renderer/tr_surface.cpp
+++ b/src/engine/renderer/tr_surface.cpp
@@ -1053,145 +1053,160 @@ Tess_SurfaceMD5
 */
 static void Tess_SurfaceMD5( md5Surface_t *srf )
 {
-	int             j;
-	int             numIndexes = 0;
-	int             numVertexes;
-	md5Model_t      *model;
-	md5Vertex_t     *v;
-	srfTriangle_t   *tri;
+	md5Model_t *model = srf->model;
 
 	GLimp_LogComment( "--- Tess_SurfaceMD5 ---\n" );
 
-	Tess_CheckOverflow( srf->numVerts, srf->numTriangles * 3 );
+	int numIndexes = srf->numTriangles * 3;
 
-	model = srf->model;
-
-	numIndexes = srf->numTriangles * 3;
-
-    tri = srf->triangles;
-	for (unsigned i = 0; i < srf->numTriangles; i++, tri++ )
-	{
-		tess.indexes[ tess.numIndexes + i * 3 + 0 ] = tess.numVertexes + tri->indexes[ 0 ];
-		tess.indexes[ tess.numIndexes + i * 3 + 1 ] = tess.numVertexes + tri->indexes[ 1 ];
-		tess.indexes[ tess.numIndexes + i * 3 + 2 ] = tess.numVertexes + tri->indexes[ 2 ];
-	}
+	Tess_CheckOverflow( srf->numVerts, numIndexes );
 
 	tess.attribsSet |= ATTR_POSITION | ATTR_TEXCOORD;
 
-	if ( tess.skipTangentSpaces )
+	if ( !tess.skipTangentSpaces )
 	{
-		// convert bones back to matrices
-		for (unsigned i = 0; i < model->numBones; i++ )
-		{
-			if ( backEnd.currentEntity->e.skeleton.type == refSkeletonType_t::SK_ABSOLUTE )
-			{
-				refBone_t *bone = &backEnd.currentEntity->e.skeleton.bones[ i ];
+		tess.attribsSet |= ATTR_QTANGENT;
+	}
 
-				TransInitRotationQuat( model->bones[ i ].rotation, &bones[ i ] );
-				TransAddTranslation( model->bones[ i ].origin, &bones[ i ] );
-				TransInverse( &bones[ i ], &bones[ i ] );
-				TransCombine( &bones[ i ], &bone->t, &bones[ i ] );
-				TransAddScale( backEnd.currentEntity->e.skeleton.scale, &bones[ i ] );
-			}
-			else
-			{
-				TransInitRotationQuat( model->bones[ i ].rotation, &bones[i] );
-				TransAddTranslation( model->bones[ i ].origin, &bones[ i ] );
-			}
-			TransInsScale( model->internalScale, &bones[ i ] );
+	vec_t entityScale = backEnd.currentEntity->e.skeleton.scale;
+	float modelScale = model->internalScale;
+	transform_t *bone = bones;
+	transform_t *lastBone = bones + model->numBones;
+
+	// Convert bones back to matrices.
+	if ( backEnd.currentEntity->e.skeleton.type == refSkeletonType_t::SK_ABSOLUTE )
+	{
+		refBone_t *entityBone = backEnd.currentEntity->e.skeleton.bones;
+		md5Bone_t *modelBone = model->bones;
+
+		for ( ; bone < lastBone; bone++,
+			modelBone++, entityBone++ )
+		{
+			TransInitRotationQuat( modelBone->rotation, bone );
+			TransAddTranslation( modelBone->origin, bone );
+			TransInverse( bone, bone );
+			TransCombine( bone, &entityBone->t, bone );
+			TransAddScale( entityScale, bone );
+			TransInsScale( modelScale, bone );
 		}
+	}
+	else if ( tess.skipTangentSpaces )
+	{
+		md5Bone_t *modelBone = model->bones;
 
-		// deform the vertices by the lerped bones
-		numVertexes = srf->numVerts;
-
-		for ( j = 0, v = srf->verts; j < numVertexes; j++, v++ )
+		for ( ; bone < lastBone; bone++,
+			modelBone++ )
 		{
-			vec3_t tmp; vec3_t pos;
-
-			VectorClear( pos );
-			for (unsigned k = 0; k < v->numWeights; k++ ) {
-				TransformPoint( &bones[ v->boneIndexes[ k ] ],
-						v->position, tmp );
-				VectorMA( pos,
-					  v->boneWeights[ k ], tmp,
-					  pos );
-
-			}
-
-			VectorCopy(pos, tess.verts[tess.numVertexes + j].xyz);
-			tess.verts[ tess.numVertexes + j ].texCoords[ 0 ] = floatToHalf( v->texCoords[ 0 ] );
-			tess.verts[ tess.numVertexes + j ].texCoords[ 1 ] = floatToHalf( v->texCoords[ 1 ] );
+			TransInitRotationQuat( modelBone->rotation, bone );
+			TransAddTranslation( modelBone->origin, bone );
+			TransInsScale( modelScale, bone );
 		}
 	}
 	else
 	{
-		tess.attribsSet |= ATTR_QTANGENT;
-
-		// convert bones back to matrices
-		for (unsigned i = 0; i < model->numBones; i++ )
+		for ( ; bone < lastBone; bone++ )
 		{
-			if ( backEnd.currentEntity->e.skeleton.type == refSkeletonType_t::SK_ABSOLUTE )
-			{
-				refBone_t *bone = &backEnd.currentEntity->e.skeleton.bones[ i ];
-				TransInitRotationQuat( model->bones[ i ].rotation, &bones[ i ] );
-				TransAddTranslation( model->bones[ i ].origin, &bones[ i ] );
-				TransInverse( &bones[ i ], &bones[ i ] );
-
-				TransCombine( &bones[ i ], &bone->t, &bones[ i ] );
-				TransAddScale( backEnd.currentEntity->e.skeleton.scale, &bones[ i ] );
-			}
-			else
-			{
-				TransInitScale( backEnd.currentEntity->e.skeleton.scale, &bones[ i ] );
-			}
-			TransInsScale( model->internalScale, &bones[ i ] );
+			TransInitScale( entityScale, bone );
+			TransInsScale( modelScale, bone );
 		}
+	}
 
-		// deform the vertices by the lerped bones
-		numVertexes = srf->numVerts;
+	glIndex_t *tessIndex = tess.indexes + tess.numIndexes;
+	srfTriangle_t *surfaceTriangle = srf->triangles;
+	srfTriangle_t *lastTriangle = surfaceTriangle + srf->numTriangles;
+	md5Vertex_t *surfaceVertex = srf->verts;
 
-		for ( j = 0, v = srf->verts; j < numVertexes; j++, v++ )
+	for ( ; surfaceTriangle < lastTriangle; surfaceTriangle++,
+		tessIndex += 3 )
+	{
+		tessIndex[ 0 ] = tess.numVertexes + surfaceTriangle->indexes[ 0 ];
+		tessIndex[ 1 ] = tess.numVertexes + surfaceTriangle->indexes[ 1 ];
+		tessIndex[ 2 ] = tess.numVertexes + surfaceTriangle->indexes[ 2 ];
+	}
+
+	shaderVertex_t *tessVertex = tess.verts + tess.numVertexes;
+	shaderVertex_t *lastVertex = tessVertex + srf->numVerts;
+
+	// Deform the vertices by the lerped bones.
+	if ( tess.skipTangentSpaces )
+	{
+		for ( ; tessVertex < lastVertex; tessVertex++,
+			surfaceVertex++ )
 		{
-			vec3_t tangent, binormal, normal, tmp, pos;
+			vec3_t position, tmp;
 
-			VectorClear( pos );
+			VectorClear( position );
+
+			float *boneWeight = surfaceVertex->boneWeights;
+			float *lastWeight = boneWeight + surfaceVertex->numWeights;
+			uint32_t *boneIndex = surfaceVertex->boneIndexes;
+			vec4_t *surfacePosition = &surfaceVertex->position;
+
+			for ( ; boneWeight < lastWeight; boneWeight++,
+				boneIndex++ )
+			{
+				TransformPoint( &bones[ *boneIndex ], *surfacePosition, tmp );
+				VectorMA( position, *boneWeight, tmp, position );
+			}
+
+			VectorCopy( position, tessVertex->xyz );
+
+			Vector2Set( tessVertex->texCoords,
+			floatToHalf( surfaceVertex->texCoords[ 0 ] ),
+			floatToHalf( surfaceVertex->texCoords[ 1 ] ) );
+		}
+	}
+	else
+	{
+		for ( ; tessVertex < lastVertex; tessVertex++,
+			surfaceVertex++ )
+		{
+			vec3_t tangent, binormal, normal, position, tmp;
+
+			VectorClear( position );
 			VectorClear( normal );
 			VectorClear( binormal );
 			VectorClear( tangent );
 
-			for(unsigned k = 0; k < v->numWeights; k++ ) {
-				TransformPoint( &bones[ v->boneIndexes[ k ] ],
-						v->position, tmp );
-				VectorMA( pos,
-					  v->boneWeights[ k ], tmp,
-					  pos );
+			float *boneWeight = surfaceVertex->boneWeights;
+			float *lastWeight = boneWeight + surfaceVertex->numWeights;
+			uint32_t *boneIndex = surfaceVertex->boneIndexes;
+			vec4_t *surfacePosition = &surfaceVertex->position;
+			vec4_t *surfaceNormal = &surfaceVertex->normal;
+			vec4_t *surfaceTangent = &surfaceVertex->tangent;
+			vec4_t *surfaceBinormal = &surfaceVertex->binormal;
 
-				TransformNormalVector( &bones[ v->boneIndexes[ k ] ],
-						       v->normal, tmp );
-				VectorMA( normal, v->boneWeights[ k ], tmp, normal );
+			for ( ; boneWeight < lastWeight; boneWeight++,
+				boneIndex++ )
+			{
+				TransformPoint( &bones[ *boneIndex ], *surfacePosition, tmp );
+				VectorMA( position, *boneWeight, tmp, position );
 
-				TransformNormalVector( &bones[ v->boneIndexes[ k ] ],
-						       v->tangent, tmp );
-				VectorMA( tangent, v->boneWeights[ k ], tmp, tangent );
+				TransformNormalVector( &bones[ *boneIndex ], *surfaceNormal, tmp );
+				VectorMA( normal, *boneWeight, tmp, normal );
 
-				TransformNormalVector( &bones[ v->boneIndexes[ k ] ],
-						       v->binormal, tmp );
-				VectorMA( binormal, v->boneWeights[ k ], tmp, binormal );
+				TransformNormalVector( &bones[ *boneIndex ], *surfaceTangent, tmp );
+				VectorMA( tangent, *boneWeight, tmp, tangent );
+
+				TransformNormalVector( &bones[ *boneIndex ], *surfaceBinormal, tmp );
+				VectorMA( binormal, *boneWeight, tmp, binormal );
 			}
+
 			VectorNormalize( normal );
 			VectorNormalize( tangent );
 			VectorNormalize( binormal );
+			VectorCopy( position, tessVertex->xyz );
 
-			VectorCopy(pos, tess.verts[tess.numVertexes + j].xyz);
-			R_TBNtoQtangents( tangent, binormal, normal, tess.verts[ tess.numVertexes + j ].qtangents );
+			R_TBNtoQtangents( tangent, binormal, normal, tessVertex->qtangents );
 
-			tess.verts[ tess.numVertexes + j ].texCoords[ 0 ] = floatToHalf( v->texCoords[ 0 ] );
-			tess.verts[ tess.numVertexes + j ].texCoords[ 1 ] = floatToHalf( v->texCoords[ 1 ] );
+			Vector2Set( tessVertex->texCoords,
+			floatToHalf( surfaceVertex->texCoords[ 0 ] ),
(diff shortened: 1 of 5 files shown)


illwieckz:
About the question on `memcpy` against per-component copy, it looks like compilers produce the exact same code:
https://godbolt.org/z/hK98hK

basically I compared:

```c++
    a[0] = b[0];
    a[1] = b[1];
    a[2] = b[2];
    a[3] = b[3];
```

with:

```c++
    memcpy( a, b, 4 * sizeof( a[0] ));
```

And they both produce the same code:

```asm
  mov esi, dword ptr [rsp]
  mov edx, dword ptr [rsp + 4]
  mov ecx, dword ptr [rsp + 8]
  mov r8d, dword ptr [rsp + 12]
```

So I'll remove that memcpy commit.

illwieckz:
About the `memcmp` stuff to check something is zeroed, I tested:

https://godbolt.org/z/7P6dnx (clang 6)

```c++
    if ( a[ 0 ] == 0 && a[ 1 ] == 0 && a[ 2 ] == 0 && a[ 3 ] == 0 )
``` 

with:

```c++
    if ( memcmp( a, b, 4 * sizeof( b[ 0 ] ) ) )
```

and:

```c++
    if ( !( a [ 0 ] | a[ 1 ] | a [ 2 ] | a [ 3 ] ) )
```

And they produced:

```asm
  mov eax, dword ptr [rsp]
  or eax, dword ptr [rsp + 4]
  or eax, dword ptr [rsp + 8]
  or eax, dword ptr [rsp + 12]
  jne .LBB0_2
```

and:

```asm
  movdqa xmm0, xmmword ptr [rsp]
  pcmpeqb xmm0, xmmword ptr [rsp + 16]
  pmovmskb eax, xmm0
  cmp eax, 65535
  je .LBB1_2
```

and:

```asm
  mov esi, dword ptr [rsp]
  mov edx, dword ptr [rsp + 4]
  mov eax, edx
  or eax, esi
  mov ecx, dword ptr [rsp + 8]
  or eax, ecx
  mov r8d, dword ptr [rsp + 12]
  or eax, r8d
  jne .LBB2_2
```

Third is probably way slower, but I don't know if there is something better in first or second. Since first and second implementation have 4 instructions (not including the jump), it looks like the `memcmp` implementation starts to become useful with 4 components.

Interestingly, with clang11, the first implementation is done the `memcmp` way:

https://godbolt.org/z/G7xWoz

illwieckz:
Hmm, for the “compare with zero” code (question about `memcmp`), GCC seems to produce better code:

https://godbolt.org/z/bbxd3o (gcc 10.2)

and way:

```asm
  mov eax, DWORD PTR [rsp]
  or eax, DWORD PTR [rsp+4]
  or eax, DWORD PTR [rsp+8]
  or eax, DWORD PTR [rsp+12]
  je .L5
```

`memcmp` way:

```asm
  mov rdx, QWORD PTR [rsp+8]
  mov rax, QWORD PTR [rsp]
  xor rdx, QWORD PTR [rsp+24]
  xor rax, QWORD PTR [rsp+16]
  or rdx, rax
  je .L6
```

bitwise way:

```asm
  mov eax, DWORD PTR [rsp]
  or eax, DWORD PTR [rsp+4]
  or eax, DWORD PTR [rsp+8]
  or eax, DWORD PTR [rsp+12]
  je .L15
```

So with GCC, the and way and the bitwise way produce the same code, and the `memcmp` code use one more instruction and one more register. So it's better to use the and way, it is more consistent between compilers.

For the “set all component” code (question about `memcpy`), unfortunately gcc is optimizing so much the first implementation in my sample code is even not done at all, but for the second implementation (the `memcmp` one), GCC produces this:

https://godbolt.org/z/dT9qG4 (gcc 10.4)

```
  mov rsi, QWORD PTR [rsp]
  mov rcx, QWORD PTR [rsp+8]
```

Which looks very efficient.

slipher:
I believe it works something like this:

When the engine main thread calls an OpenGL function, the GL driver copies the data into a graphics operation queue. The GL driver runs all the time, in parallel with the engine thread, to perform any operations in the queue. If the queue is too full, then when the engine thread calls an OpenGL function, it may have to wait for some operations to be completed before there is room to insert the operations into the queue. I saw something similar while investigating https://github.com/Unvanquished/Unvanquished/issues/1109. The GL driver will block the main thread for a while if the graphics processing pipeline gets too far behind.

https://www.khronos.org/opengl/wiki/Synchronization#Asynchronous_action

When the GPU is by far the bottleneck, you will get zero increase in overall throughput by speeding up the engine code.

illwieckz:
Yes, that's what I'm afraid of. What I would like to understand is why when I double framerate when I return right before the `// Deform the vertices by the lerped bones.` comment in IQM code. At this point the data is garbage so surfaces are badly rendered, but this doubles the framerate…

slipher:
> What I would like to understand is why when I double framerate when I return right before the // Deform the vertices by the lerped bones. comment in IQM code. 

Good question... maybe the garbage data in `tess.verts` results in points outside the player's field of view or something, which allows them to be thrown out without further processing?

Perhaps a fairer comparison would be to always take the `else` clause corresponding to `if( model->num_joints > 0 && model->blendWeights && model->blendIndexes )`, rather than returning if the condition is true.

illwieckz:
> Perhaps a fairer comparison would be to always take the `else` clause corresponding to `if( model->num_joints > 0 && model->blendWeights && model->blendIndexes )`, rather than returning if the condition is true.

If I do the scene runs at `12fps` instead of `7fps`.

Anyway, that's how it looks on such hardware:

[slow models](https://dl.illwieckz.net/b/unvanquished/bugs/highmodels/20201022-162429-000.unvanquished-slow-models.png)

slipher:
OK now I understand, the code in question only runs if r_vbomodes=0 or some capabilities are missing.

The Tess_SurfaceIQM commit looks good - it should help to reduce pointer aliasing.

With the Tess_SurfaceMD5 commit, it does not seem helpful to me to combine the `srf->numTriangles`- and `srf->numVerts`-bounded loops into one. Why do you want to do that?

illwieckz:
Yeah, those loops were some experiments, I plan to revert that on md5 side too, and to implement freem suggests as well, it's just that at this point I only redid properly the IQM code after my first attempt.

illwieckz:
For the loop merge, at first I was looking for a way to reduce the number of loops, but the conditions may just destroy prediction, and anyway I did not noticed any gain, so I'm just reverting them.

illwieckz:
@slipher: I reproduced the pointer optimization on md5mesh code, and moved tests outside of loops.

illwieckz:
Code now looks ready to merge to me. Any other optimization would be to be done on separate PR, based on this work.

src/engine/renderer/tr_surface.cpp
```diff
@@ -1053,145 +1053,159 @@ Tess_SurfaceMD5
...
+	tess.attribsSet |= ATTR_POSITION | ATTR_TEXCOORD;
 
-    tri = srf->triangles;
-	for (unsigned i = 0; i < srf->numTriangles; i++, tri++ )
+	if ( tess.skipTangentSpaces )
 	{
-		tess.indexes[ tess.numIndexes + i * 3 + 0 ] = tess.numVertexes + tri->indexes[ 0 ];
-		tess.indexes[ tess.numIndexes + i * 3 + 1 ] = tess.numVertexes + tri->indexes[ 1 ];
-		tess.indexes[ tess.numIndexes + i * 3 + 2 ] = tess.numVertexes + tri->indexes[ 2 ];
+		tess.attribsSet |= ATTR_QTANGENT;
```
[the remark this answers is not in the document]
illwieckz: Wow, nice catch, I fixed that.

src/engine/renderer/tr_surface.cpp
```diff
@@ -1053,145 +1053,159 @@ Tess_SurfaceMD5
...
+
+	glIndex_t *tessIndex = &tess.indexes [ tess.numIndexes ];
+	srfTriangle_t *surfaceTriangle = srf->triangles;
+	srfTriangle_t *lastTriangle = surfaceTriangle + srf->numTriangles;
+	md5Vertex_t *surfaceVertex = srf->verts;
+
+	for ( ; surfaceTriangle < lastTriangle; surfaceTriangle++,
+		tessIndex += 3 )
+	{
+		VectorSet( tessIndex, tess.numVertexes, tess.numVertexes, tess.numVertexes );
```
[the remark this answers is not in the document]
illwieckz: Right, I used that Vector thing because there is a real need for function/macro that are usable for any duet/triplet/quadruplet of data. Copypasting is really annoying. So I did copypasting for MD5 instead.

Also, I just noticed that on IQM code, we can just iterate by one because both triangles and indexes are iterables the same way, and that helped me to reach 10fps on the test machine (from 9fps).

illwieckz:
That's not the speed gain I was looking for, but at least we can notice the code is more efficient. On the test machine, the scene went from solid 7fps to solid 10fps (and it flirts with 11).

[highmodels](https://dl.illwieckz.net/b/unvanquished/bugs/highmodels/unvanquished_2020-10-22_123251_000.jpg)

[highmodels](https://dl.illwieckz.net/b/unvanquished/bugs/highmodels/unvanquished_2020-10-27_162507_000.jpg)

slipher:
I don't understand the meaning of `tess.skipTangentSpaces`. Why is the `implement iqm skipTangentSpace alternative like md5` commit needed? Is it faster?

illwieckz:
I don't know the meaning of `tess.skipTangentSpaces`, but thing is that engine is flagging some surfaces with that, and MD5 function had a faster code for such situation. I've seen no reason why IQM models can't be in exact same situation as MD5 ones, so there is no reason that situation cannot occurs with IQM if it can occurs with MD5, so I assume this faster code can be used in such situation. I don't know how to reproduce the situation so I can't really measure the difference, but I've seen no one issue with any of our models.

Basically I'm doing with MD5 and IQM what I did with GLSL shaders: I make sure the same thing is done for the same situation.

Now the functions are like 95% similar. What prevents to merge the two functions is the data structure being different. Otherwise operations are now exactly the same on MD5 and IQM code.

illwieckz:
Can I merge this?

src/engine/renderer/tr_surface.cpp
```diff
@@ -1257,6 +1262,14 @@ void Tess_SurfaceIQM( srfIQModel_t *surf ) {
 			TransInsScale( modelScale, bone );
 		}
 	}
+	else
+	{
+		for ( ; bone < lastBone; bone++ )
```
slipher, line 1265: These two cases can be merged. `TransInit(bone); TransAddScale(entityScale, bone);` is equivalent to `TransInitScale(entityScale, bone);`.

src/engine/renderer/tr_surface.cpp
```diff
@@ -1053,145 +1053,160 @@ Tess_SurfaceMD5
...
+		tessIndex[ 1 ] = tess.numVertexes + surfaceTriangle->indexes[ 1 ];
+		tessIndex[ 2 ] = tess.numVertexes + surfaceTriangle->indexes[ 2 ];
+	}
 
-		// convert bones back to matrices
-		for (unsigned i = 0; i < model->numBones; i++ )
+	shaderVertex_t *tessVertex = tess.verts + tess.numVertexes;
+	shaderVertex_t *lastVertex = tessVertex + srf->numVerts;
+
+	// Deform the vertices by the lerped bones.
```
illwieckz, line 1128: The lost comment is there.

src/engine/renderer/tr_surface.cpp
illwieckz: Oh, nice, and now you say that's the same, I notice I did the commit “implement iqm skipTangentSpace alternative like md5” absolutely wrong: on that part I imported what was already done instead of what was not done yet. That makes me notice I had another `TransInitRotationQuat( modelBone->rotation, bone ); TransInverse( &modelBone->joint, bone );` occurrence to optimize in ` precompute md5 bone transform like iqm`. That also means I can do `TransInitScale(entityScale, bone);` in md5 code as well. Impressive sight!

slipher:
Is the original code wrong in `Tess_SurfaceMD5`?  
In the not SK_ABSOLUTE case, with skipTangentSpaces=true there's `TransInitRotationQuat( model->bones[ i ].rotation, &bones[i] ); TransAddTranslation( model->bones[ i ].origin, &bones[ i ] );`.
For skipTangentSpaces=false there's `TransInitScale( backEnd.currentEntity->e.skeleton.scale, &bones[ i ] );`.

I thing skipTangentSpaces=true is some thing where the model is not actually drawn on the screen, so it might be hard to tell if it is wrong. Is it even noticeable if you do something completely wrong for that case?

illwieckz:
> I thing skipTangentSpaces=true is some thing where the model is not actually drawn on the screen

If true, is there interest into taking care of that? How are taken in account things that are not in direct sight, like seen through a mirror?

slipher:
Based on `Tess_Begin` calls, it appears to happen for some lighting stuff, some debug utils, and a function called `SurfIsOffscreen`.

illwieckz:
> Is the original code wrong in `Tess_SurfaceMD5`?
> In the not SK_ABSOLUTE case, with skipTangentSpaces=true there's `TransInitRotationQuat( model->bones[ i ].rotation, &bones[i] ); TransAddTranslation( model->bones[ i ].origin, &bones[ i ] );`.
> For skipTangentSpaces=false there's `TransInitScale( backEnd.currentEntity->e.skeleton.scale, &bones[ i ] );`.

I'm not sure to understand what you are meaning on that part (or what you are asking for).

illwieckz:
I added some commits to rewrite the common part of MD5 VBO code and unify it the same way the common MD5 non-VBO and IQM VBO/non-VBO is.

Because for some reason there is one function for MD5 non-VBO and one for MD5 VBO while IQM code has one function to rule them all and a test to do this or that.

illwieckz:
I would like to know what this is for:

```
	tess.attribsSet |= ATTR_POSITION | ATTR_TEXCOORD;

	if ( !tess.skipTangentSpaces )
	{
		tess.attribsSet |= ATTR_QTANGENT;
	}
```

At first

- MD5 non-VBO code had the above,
- MD5 VBO code did not have it at all,
- IQM non-VBO code had what is below,
- IQM VBO was doing the same as IQM non-VBO:

```
tess.attribsSet |= ATTR_POSITION | ATTR_TEXCOORD | ATTR_QTANGENT;
```

I see no reason why this would be different between MD5 and IQM neither I see why it would be different between VBO and non-VBO in MD5 case while it would be the same between VBO and non-VBO in IQM case.

While it may be possible for VBO code to be different, I would expect MD5 VBO to be like IQM VBO and MD5 non-VBO to be like IQM non-VBO.

slipher:
> I'm not sure to understand what you are meaning on that part (or what you are asking for).

In all cases talking about the original code before this PR, in the case where skeleton.type != refSkeletonType_t::SK_ABSOLUTE:

In Tess_SurfaceIQM, and in Tess_SurfaceMD5 with skipTangentSpaces=false, it does the same thing in the loop over bones. In Tess_SurfaceMD5 with skipTangentSpaces=true, it does a different thing (and a more complicated one, so it can hardly be an optimization). It seems like these 3 cases should all be in the same situation, so it might be a mistake that one of them is different.

illwieckz:
Maybe commit `tr_model_md5,tr_surface: precompute md5 bone transform like iqm` is an answer to your question?

Previously the MD5 code was doing `TransInitRotationQuat( model->bones[ i ].rotation, &bones[i] ); TransAddTranslation( model->bones[ i ].origin, &bones[ i ] );` code at render time (in `tr_surface.cpp` file) while the IQM code was doing it at load time (in `tr_model_iqm.cpp` file).

Now they both do it a load time and use a similar code at run time (this must be more efficient on MD5 side).

illwieckz:
I added some code to precompute weight numbers, so the IQM non-VBO code is almost the same as the MD5 non-VBO code, the only difference is now the data model. Basically the iqm code iterates multiples arrays for a, b, c, d, e (more iterations, less pointer dereference) when the md5code iterates one single array of a structure containing a, b , c, d, e (more pointer dereference, less iterations).

On a more recent CPU (FX-9590), I now get the exact same FPS between md5 and iqm code when vbo models are entirely disabed.

So, I did some extensive tests.

Scene:  `devmap plat23`, `setviewpos 2055 1920 12 0 0`

Note: only acid tubes models are converted to md5 for the test, so the iqm test only process iqm models when the md5 test process at least six md5 models and the remaining ones are still iqm: 6 acid tubes (4 fully visible, two hidden), 2 eggs (hidden), 1 barricade (partially visible), 1 overmind (hidden).

Configuration:

FX9n:
CPU: AMD FX-9590
GPU: AMD Radeon R9 390X
Resolution: 2K (2560×1440)
Preset: medium
Rendering: `r_vboModels 0`

ATHn:
CPU: AMD Athlon64 3200+
GPU: ATI Radeon X1950 PRO
Resolution: VGA (640×480)
Preset: lowest
Rendering: `r_vboModels 0`

ATHv:
CPU: AMD Athlon64 3200+
GPU: ATI Radeon X1950 PRO
Resolution: VGA (640×480)
Preset: lowest
Rendering: `r_vboModels 1` but most models can't do VBO.

Result:

code    |format       |FX9n|ATHn|ATHv
--------|-------------|----|----|----
original|6 iqm + 4 iqm|29  |7   |7
new     |6 iqm + 4 iqm|37  |10  |10


code    |format       |FX9n|ATHn|ATHv
--------|-------------|----|----|----
original|6 md5 + 4 iqm|33  |8   |14
new     |6 md5 + 4 iqm|37  |9   |17

Interestingly:

- the MD5 non-VBO code was given more fps than IQM non-VBO code on all computers, but while both MD5 and IQM code got both a performance boost, the MD5 boost was slightly lower so on the more recent computer the fps is now the same for both formats, but on the older computer this speed boost difference is enough to make IQM a bit faster than MD5.

- What is really surprising and unexplained yet is that if I enable VBO models but the hardware can't handle most of them as VBO ones (like acid tubes), using MD5 acid tubes is faster than using IQM acid tubes (while there is other IQM models in scene) in all case, getting almost twice the performance with both older and new code.

illwieckz:
Along the data model, the remaining difference between MD5 and IQM code is that MD5 VBO and non-VBO code are split into two functions, while IQM VBO and non-VBO code is in the same function with some `if`(less duplication of code but less predictable), also the IQM VBO code seems more complexe than the MD5 VBO code, but in my test the MD5 VBO code is not expected to be used at all.

illwieckz:
I noticed something though:

In `tr_animation.cpp`, the `R_AddMD5Interactions` function has two alternatives codes for when this test is true or not:

```
	if ( !r_vboModels->integer || !model->numVBOSurfaces ||
	     ( !glConfig2.vboVertexSkinningAvailable && ent->e.skeleton.type == refSkeletonType_t::SK_ABSOLUTE ) )
```

While the `R_AddIQMInteractions` code only has one. I don't know if it's related.

Also `R_AddMD5Surfaces` in this file also have this kind of test while `R_AddIQMSurfaces` in `tr_iqm_models.cpp` doesn't have it (but there is another test not entirely similar in `R_LoadIQModel`, I don't know what it means).

I don't know if there is anything useful here.

slipher:
> Maybe commit tr_model_md5,tr_surface: precompute md5 bone transform like iqm is an answer to your question?

No. The question is why the code is different in these two places:

https://github.com/DaemonEngine/Daemon/blob/6b94e2e890f893afdc55b7f70bd58cb39f3cafcd/src/engine/renderer/tr_surface.cpp#L1098
https://github.com/DaemonEngine/Daemon/blob/6b94e2e890f893afdc55b7f70bd58cb39f3cafcd/src/engine/renderer/tr_surface.cpp#L1145

illwieckz:
I don't know, my effort was to move around code and rewrite it in a way it is more efficient, trying to make sure it does the same thing before and after but I don't know what it does. This is pure meta, like simplifying an equation by moving around elements by applying associativity and commutativity until terms just cancels each other out.

That's also why the only optimizations have done are only about “doing better C code for that algorithm” and not “doing a better algorithm for that problem”.

illwieckz:
So, when rendering the acid tube, the MD5 code processes 6294 weights, and the IQM code processes 6266 (while previously it was processing 15452 ones). I don't know why the IQM model seems to have a bit less weights. For information the IQM model was converted from md5 in the past (the md5 model is the original one). But then, both code now process ~6200 weights, so the speed difference is not there anymore.

illwieckz:
I rebased on current master. To remind the context, this work was motivated by https://github.com/Unvanquished/Unvanquished/issues/1207 when I discovered that some GPUs may not be able to process the model transformations because of some limits and then, the engine would rely on a fallback that is too slow to make the game playable.

Most of that PR is to rewrite a lot of code to help compiler optimize better the code, for example by reducing some pointer indirections and reduce branching (so less branching prediction uncertainty). Other parts of that PR is to rewrite both IQM and MD5 code to reduce the most possible differences in their implementations, in a way to make possible to port back and forth optimizations or identify differences that may explain performance differences (maybe in the future it would be possible to use the same code for both? who knows).

It's believed there is still big rooms for optimizations, given:

- the same scene with md5 acid tubes instead of iqm acid tubes is known to have twice the performances,
- ZTM has done heavy optimizations in ioquake3 involving agressive precomputations and we may want to import those one day.

But, this PR is already very big, and I would like to see this merged before I look at other optimizations. I basically need this to be merged so I can count on it.

To test this code, the `r_vboModel` cvar must be set to `0`. Note: when `r_vboModel` is set to `0` there is a bug where third-person MD3 weapon models are not displayed, that's not related to this code, and when non-VBO model animation is done because of hardware limit (not because that cvar is disabled), they are properly displayed.

While this does not bring the twice performance increase I was looking for, the performance gain is still measurable:

On plat23 map, using `lowest preset` and `r_vboModel`, I get on my desktop computer:

Scene: `/setviewpos 2314 2124 308 52 45` (camera on alien overmind, two eggs, leech and barricade)

- Before: 59 fps
- After : 74 fps

Scene: `/setviewpos 1893 1920 -5 0 0` (camera from alien base entry corridor, even if not everything is seen, the engine renders 6 acid tubes, 3 eggs, 1 overmind, 1 leech and 1 barricade)

- Before: 28
- After: 38

So that's already a 20~25% gain.

I remember that on a very old computer the fps went from 10 to 12 on that complex scene (while the same old computer can do 40 fps when there is none of such non-hardware accelerated models), which confirms the 20% gain. I'm still dreaming of doubling the performances though…

illwieckz:
Note: given comment there: https://github.com/DaemonEngine/Daemon/pull/389#issuecomment-742222570

The MD5 code is also a bit faster when done on CPU.

illwieckz:
I tested on an old Intel integrated (Core 2 Duo L7500 + X3100 / GMA965), on this hardware, the Linux driver emulates the GPU vertex animation, which is the slowest implementation, as we would got only 8 fps on the heavy plat23 alien base scene reference:

[iqm slow on cpu](https://dl.illwieckz.net/b/daemon/bugs/iqm-slow-on-cpu/unvanquished_2021-07-10_123620_000.jpg)

With our own CPU alternative code (`r_vboVertexSkinning 0`), current `master` gets 13 fps:

[iqm slow on cpu](https://dl.illwieckz.net/b/daemon/bugs/iqm-slow-on-cpu/unvanquished_2021-07-10_134242_000.jpg)

With this IQM revamp, 15 fps is reached:

[iqm slow on cpu](https://dl.illwieckz.net/b/daemon/bugs/iqm-slow-on-cpu/unvanquished_2021-07-10_143546_000.jpg)

With this IQM revamp plus q_math inlinine (#505), 16 fps is reached:

[iqm slow on cpu](https://dl.illwieckz.net/b/daemon/bugs/iqm-slow-on-cpu/unvanquished_2021-07-11_183227_000.jpg)

This scene (`+devmap plat23 +delay 100f setviewpos 1893 1920 0 0 0`) is very heavy to render, there are twelve animated buildables to process: one overmind, one leech, one barricade, six tubes, and three eggs.

After this PR and #505, there are two more optimizations paths to explore (both can be done), turning `q_math` inline functions into SSE code, and porting ZTM's optimizations from ioquake3 as described there: https://github.com/DaemonEngine/Daemon/issues/390#issuecomment-719093758

Note: to test this code path, one can set `r_vboVertexSkinning` to `0`.

illwieckz:
Status of the PR: ready to merge.

The proposed changes are meant to improve performance without intention to change the underlying logic. Except some compute being moved from run time to load time, there was no intention to rethink the way it is done.

The proposed changes are now deeply tested over various hardware and using compilers we use for release (gcc and clang) but I only tested on Linux, though I doubt there would be operating-system specificity there.

Proposed changes were addressed and suggests applied.

Even if not everything got a comment, at this point I even don't know what can be left for a review.

I see no reason to delay the merge more so if no one has something to say I'll merge it tomorrow.

illwieckz:
People wanting to use that code can set `r_vboVertexSkinning` to `0`.

In the past I talked about `r_vboModels 0` but this other one also has unwanted effects on other things.

The exact cvar to enable the CPU path for model animation is `r_vboVertexSkinning 0`.

PR #391, opened 2020-10-25
Drop support for cinematic mode and videoMap

slipher:
The immediate motivation for this arose because while updating the external_deps build, I noticed the Theora project seems kind of dead. When compiling, I saw a gcc warning that the decoder code indexes out of bounds. Although this was [fixed in 2012](https://github.com/xiph/theora/commit/0ae66d565e6bead8604d312bc1a4e9dccf245c88), there hasn't been a new release since 2010 so we didn't have the fix.

I wanted to test `/cinematic`, but had trouble finding any ogm file or converters to test with. I suspect it would crash, considering the aforementioned bug.


illwieckz:
I'm sorry I wasn't available today so I was not available to answer your question before.

Outside of _this_ specific “cinematic” feature, I'm a bit worried on multiple features that were dropped last years. I see the issues around keeping code we don't use and I'm concerned about them. Like this one, there is many feature Unvanquished as a game does not use. The fact I don't know what it is about just reveals how much we don't need it for Unvanquished at least on our current state, but at some point maybe we are just not feature complete enough?

Was such “cinematic” used for things like the id Logo intro and stage video in Quake 3 Arena (`.RoQ` videos)? Or the video played on the background of Wolf:ET? If we drop this, do we have a way to play a video as an intro for the game, a video as a background in a menu, or as a credit screen? Because even if we don't use that for Unvanquished, this looks like a legit feature for a game engine, especially for a game engine that may run games ported from those older idtech3 engines.

What I'm a bit concerned about is also that we basically use an engine that was intially a merge of quake3, xreal and and wolfet features, and years after years we are just reverting the wolfet merge and maybe even drop original quake3 features.

On January of 2019, we nuked IRC support. Really, we don't use it, and I seen no usage for it, I probably gave my GO or just approved by my silence. But then in November of 2019, I participated in the Xonotic Defrag World Championship, and then I opened my eyes. During that event, all the participating servers were connected to a single irc channel and then, broadcasting talks, events like joins and departures, and some scores. The experience was really good, whatever the server you chose, the one the closest to you, people from all over the world on other servers were greeting you, and you were congratulating them or just talking them. You can experience that yourself because the XDWC 2020 is running now (first week of the competition ends today, five weeks to come), just join and XDWC server in Xonotic, or join #xdwc @ quakenet. Now I'm sad we cannot do that.

Recently I looked to see if we can mutualise development effort on the renderer with other teams, like ET:Legacy, then I discovered we nuked MDM model and MDX animation format in 2013 (here some unofficial [spefications](https://mino-git.github.io/rtcw-wet-blender-model-tools/publications/MDMMDXFileFormat.html) about those Wolf:ET formats)…

To me it's OK to disable this code to not postpone release 0.52, It's better to postpone this instead: priority is Unvanquished.
If we keep this cinematic stuff, for sure we can drop theora, better use VP8 at this point. If we go this route we can really postpone this. Can we disable by default the theora related code in cmake?

illwieckz:
So, I've checked in Quake3, `/cinematic video/idlogo.RoQ` is playing the id logo intro, you can do the same with `video/intro.RoQ`, `video/tier1.RoQ`, `video/end.RoQ`, etc. It would be bad to drop such feature in engine even if we don't use it today. Being able to map a video on a surface with `videoMap` is really good too, even if don't use it for Unvanquished. So I'm worried about such deletion.

slipher:
> So, I've checked in Quake3, /cinematic video/idlogo.RoQ is playing the id logo intro, you can do the same with video/intro.RoQ, video/tier1.RoQ, video/end.RoQ, etc.

1. ROQ support was already removed in the past.
2. The `CA_CINEMATIC` state is incompatible with being connected to a server or having gamelogic running. So it requires hardcoding game-specific logic into the engine to use. It would require a redesign to be usable with our current engine design.

`videoMap` on the other hand could be useful in theory, I guess, but I'd be surprised if it actually works.

> If we keep this cinematic stuff, for sure we can drop theora, better use VP8 at this point. If we go this route we can really postpone this. Can we disable by default the theora related code in cmake?

I don't understand, OGM is the only supported format. I'm not going to add support for a new one when there isn't any demand for the feature. It harms development to keep dead code in the build; much better to delete it and then revert when a need arises.

illwieckz:
Code is currently broken:

```
]/cinematic video/idlogo.ogm 
Debug: CL_PlayCinematic_f 
Debug: CIN_PlayCinematic( video/idlogo.ogm ) 
Debug: OGM-Init done (video/idlogo.ogm) 
Debug: [Theora(ogg)]new resolution 1024x512 
]malloc(): unsorted double linked list corrupted
```

I used `video/idlogo.ogm` file from [World of Padman](https://worldofpadman.net/en/) (you can find it in `wop_004.pk3`).

 I know RoQ support was dropped long time ago. I'm just assuming that OGM support was an attempt for newer (cough, cough) format replacement for this feature. Also, I don't know if this is really OGM. OGM is no OGV, OGM is a non-standardized hack on OGG format to embed non-theora data bcause that predated theora, or the incompatible standardized OGV format for theora, VP8, VP9 etc. On the other hand, OGV and OGA are just different file extensions for the same OGG format.

Loading an OGV converted from a RoQ file taken from Smokin' Guns (Smokin' Guns have both intro and credit video) gave me another error (I modified the engine locally to allow this file extension):

```
]/cinematic video/idlogo.ogv 
Debug: CL_PlayCinematic_f 
Debug: CIN_PlayCinematic( video/idlogo.ogv ) 
Debug: OGM-Init done (video/idlogo.ogv) 
Debug: [Theora(ogg)]new resolution 512x512 
]double free or corruption (!prev)
```

Loading an OGM file produced with VLC (while I cannot be sure it's OGM and not OGV) gave me the same issue I got with the OGM from World of Padman:

```
Debug: [Theora(ogg)]new resolution 512x512 
]malloc(): unsorted double linked list corrupted
```

There is a small chance I broke videoMap when revamping the renderer, because the [XreaL shader manual](https://tremap.xtr3m.net/__Games/Xreal/Manual_Shader_1/ShaderManual.htm#videoMap.3Cmap.3E) says “this is only supported by colormap stage types”, and I removed the `colorMap` stage keyword, but I removed it because it was doing nothing: a colormap is already a legacy stage using `map` keyword, without normal, lightmaps and other textures.

I know how it's problematic to keep dead code, and here, the code is broken. But I'm really concerned about features being removed and removed and removed… I'm more in favor of building test assets (see [UnvanquishedTestAssets](https://github.com/UnvanquishedAssets/UnvanquishedTestAssets)).

illwieckz:
World of Padman:
- https://dl.illwieckz.net/b/daemon/features/cinematic/test-assets/worldofpadman/idlogo.ogm

Smokin' Guns RoQ converted to OGV:
- https://dl.illwieckz.net/b/daemon/features/cinematic/test-assets/smokinguns/idlogo.ogv
- https://dl.illwieckz.net/b/daemon/features/cinematic/test-assets/smokinguns/credit.ogv

Why not fixing the OGM loader to load OGV properly? we already have OGG, Vorbis and Theora support, and we know OGG and Vorbis already work.

slipher:
> Why not fixing the OGM loader to load OGV properly? 

I don't want to put major development efforts into something which at present, no one wants to use. And if I did implement some video loader, I'd want to use one with a maintained library.

illwieckz:
Well, in fact I know one use for video: the Smokin' Guns game has two videos, but I know the port is dormant for now.

I don't know why you talk about maintained library like if I said the contrary, I proposed the OGG format that is maintained, and we already use.

There is some xvid references in OGM code comments, so it's very likely to be a code not for OGG but for that old OGM hack that existed before Theora existed and before OGG was modified to support videos. OGM was created to be able to merge xvid and vorbis, in fact, many format with vorbis, but the main purpose was xvid and vorbis. It looks like OGM could ship theora and vorbis but I know no tool that works today to produce such files.

So, if my guess are right, we have a code that supports a deprecated and dead container and an obsolete but working video format and a maintained audio format.

Maybe if Dæmon is ported to support OGG container for videos instead of OGM, even by keeping theora for the moment, we would get something working and testable. Replacing theora by something maintained (vpx?) would be up to the one wanting it.

Unless I'm wrong, the issue seems to be the Dæmon engine expects an old and deprecated container, and that's not related to the video format itself.

XreaL was a demonstrator with ugly quircks, so I would not be surprised to see an OGM + Theora + Vorbis thing in it, even it's the only one on earth to do it.

After some investigation, it looks like the `idlogo.ogm` file from World Of Padman is an OGG file,not an OGM.

I managed to produce an OGM file (not with theora) and `ogminfo`, `file` and `mediainfo` shows things the same way for OGV (OGG) files and that `idlogo.ogm` file while the output is different with OGM files. So we may look at World of Padman code to see the OGG implementation for video.

I'm really worried about deleting features that looked pretty standard (here, video support, whatever the format) because the code is broken. In fact, Every quake3-derivated game I can think about right now used videos. Tremulous, Smokin' Guns, OpenArena, Urban Terror, Q3 Rally, Reaction, they all had videos (usually intro videos)… That's a really standard feature for a game engine, I see the same usage on other engines like Source with GoldenEye source or QFusion with Warsow. So it's very likely a port of a game to Dæmon may look at this feature (video support, whatever the format), which is pretty standard and something anyone can fairly expect from an engine.

slipher:
> I don't know why you talk about maintained library like if I said the contrary

Theora is the unmaintained library I was referring to. 

An Ogg+Vorbis+VPx player would be a good thing to have, yes. But a video in the splash screen is a polish feature, a cherry on top. I'm not concerned by the idea that someone will reject the Daemon engine because they can't have a video in the main menu. I saw your point about Smokin' Guns, but AFAICT the Daemon port hasn't advanced beyond the stage of speculation. If there were some gamelogic going in Daemon, then we could say that we have a potential user for the feature. Meanwhile, as long as no one is going to use it in the near future, I don't want to expend scarce development resources on it.

illwieckz:
I'vde diffed our code with the one from [World of Padman](https://github.com/PadWorld-Entertainment/wop-gamesource/blob/master/code/client/cin_ogm.c), the only difference (except WoP still has XviD support), is on this line:

```c++
				Audio::StreamData( 0, rawBuffer, i, g_ogm.vi.rate, 2, 2, 1.0f, 1);
```

Which was modified from:

```c++
				S_RawSamples( 0, i, g_ogm.vi.rate, 2, 2, rawBuffer, 1.0f, 1.0f );
```

In bc838e1e664a533a6138f17c5a0b412195edda9a (https://github.com/Unvanquished/Unvanquished/commit/6b970ecdab483c8c80c8930048de4e448428f0fd).

On the other hand, WoP has this:

```
				S_RawSamples( 0, i, g_ogm.vi.rate, 2, 2, rawBuffer, 1.0f, -1 );
```

I don't know why the difference between `1`/`-1` but that does not seem to be the culprit, our code worked with 1.

That commit was introduced between `v0.23.0` and `v0.24.0` and the code still worked in `v0.26.0`. The started to fail between `v0.26.0` and `v0.27.0`. Audio code got huge revamp at the time.

Anyway, the code is now failing on the `Audio::StreamData`:

```gdb
]malloc(): unsorted double linked list corrupted
#0  __GI_raise(sig = 6, sig@entry = 6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  __GI_abort() at abort.c:79
#2  __libc_message(action = do_abort, action@entry = do_abort, fmt = 0x7ffff6f92285 "%s\\n", fmt@entry = 0x7ffff6f92285 "%s\\n") at ../sysdeps/posix/libc_fatal.c:155
#3  malloc_printerr(str = 0x7ffff6f94ad8 "malloc(): unsorted double linked list corrupted", str@entry = 0x7ffff6f94ad8 "malloc(): unsorted double linked list corrupted") at malloc.c:5347
#4  _int_malloc(av = 0x7ffff6fc3b80 <main_arena>, av@entry = 0x7ffff6fc3b80 <main_arena>, bytes = 4096, bytes@entry = 4096) at malloc.c:3744
#5  __GI___libc_malloc(bytes = 4096) at malloc.c:3066
#6  alBufferStorageSOFT() at :0
#7  Audio::AL::Buffer::Feed(this = 0x7fffffffd3e8, audioData = ) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/audio/ALObjects.cpp:138
#8  Audio::StreamData(streamNum = 0, data = 0x555576e9af80, numSamples = 1024, rate = 44100, width = 2, channels = 2, volume = 1, entityNum = 1) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/audio/Audio.cpp:367
#9  loadAudio() at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/client/cin_ogm.cpp:232
#10  loadFrame() at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/client/cin_ogm.cpp:454
#11  Cin_OGM_Run(time = 17) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/client/cin_ogm.cpp:713
#12  CIN_RunCinematic(handle = 0) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/client/cl_cin.cpp:276
#13  SCR_RunCinematic() at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/client/cl_cin.cpp:603
#14  CL_Frame(msec = 23) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/client/cl_main.cpp:2564
#15  Com_Frame() at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/qcommon/common.cpp:1015
#16  Application::ClientApplication::Frame(this = 0x5555560296c0 <Application::GetApp()::app>) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/client/ClientApplication.cpp:84
#17  Application::Frame() at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/framework/Application.cpp:73
#18  main(argc = 53, argv = 0x7fffffffd8a8) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/framework/System.cpp:688
```

We know the WoP `idlogo.ogm` worked, and after more experiments I'm now convinced it's a real OGM (not OGG) with theora, that's very rare.

What I understand from this is that at some point we introduced a bug in the audio code, this OGM code is known to trigger it, I'm not sure that's a good idea to hide this bug.

illwieckz:
Here are two PK3 that makes possible to test OGM support with unvanquished 0.26:

The WoP OGM file (can be played with `/cinematic video/idlogo.ogm`, to store in `~/.unvanquished/pkg`):

https://dl.illwieckz.net/b/daemon/features/cinematic/test-assets/unvanquished_0.26.0/000-ogm_0.pk3

A fix for some GLSL code that was broken at the time (`~/.unvanquished/glsl` may have to be deleted before running the engine):

https://dl.illwieckz.net/b/daemon/features/cinematic/test-assets/unvanquished_0.26.0/000-glsl_0.pk3

This can be played this way:

```
./daemon +nocurses +set fs_extrapaks '000-glsl 000-ogm' +cinematic 'video/idlogo.ogm'
```

illwieckz:
Or the bug is not in the audio bakend but was introduced while revamping that backend.

Commit 02206c82937646ae483f22cff390864cfc2dd14a (https://github.com/Unvanquished/Unvanquished/commit/6276f9d1a09ef6791b48023bd680a328fad428a7) touched cinematic files between `v0.26.0` and `v0.27.0` and was part of that audio revamp. That seems to be the other non-whitespace difference with WoP code.

slipher:
Please don't waste your time on this investigation. I'm pretty sure I already know the cause of the heap corruption, which would be the out-of-bounds indexing I mentioned in the OP. Debugging a bug is not a problem here. The issue is for this feature to be any good, we would need to get new container and video stream formats and redesign the cinematic interface so that it can be used by a cgame.

illwieckz:
My concern is to keep the contribution gap low. I don't really expect a messiah to come and write for us a complete brand new video implementation using ogg, opus and vpx from scratch. I just compare the random contributor to be another myself. I don't have the skills to write such things, but I can test and identify the commit introducing the issue in a code that already exists. Even if this code is far from perfect, this code exists, and if it exists I can attempt things, and if I can attempt things, others can do.

And I'm really concerned by all those really perfectible amount of code that are just nuked and nuked and nuked in favor of a better implementations that never come. IRC got nuked. VoIP got nuked, and when I make hard effort to get players, first thing newcomer ask for is VoIP…

Almost all the fixes I do on this code base is based on micro-edits. I don't have the skill to fix features with a wide understanding of the whole, so instead I rewrite things steps by steps with the lowest scope possible, and testing my micro edits on every step.

Such method makes anyone able to contribute to any part of the code.

Anyway, regarding this bug, I may have reached my limits. I see a byte array that was replaced with a short array while seeing no adaption of the code after that, but that's all I see and I don't know what's next.

In all case I prefer an imperfect implementation that barely works than a wide welcoming place for something better no one believes in. Getting a test bed for the OGM code is a small step, finding what may be the root cause of the OGM code malfunction is a small step, fixing the bizarre OGM code is a medium step, implementing OGG instead is an high step, implementing VPX is another high step, redesigning the cinematic interface is a huge step.

Redoing everything from scratch is not a step, it's a slick and bald cliff. I know for sure people like me can't even contribute a part of such effort. My concern is to keep the contribution gap low.

slipher:
To me, keeping a broken feature is by far the worst option. If I see a broken feature in some piece of software, it makes me think the developers are incompetent and I should avoid that software. But what exactly are you suggesting I do?

As for removing IRC or VoIP, it seems like a reasonable decision when the feature was there for years with no evidence of anyone having used it. Anyway, the code is still right there in the repository for anyone who wants to do something with it. Besides VoIP, there is also some Mumble code, which is still there. Do we really need both? Rumors were, the VoIP code didn't really work right.

Judging by https://wiki.xiph.org/index.php/MIMETypesCodecs, the Ogg container doesn't actually support any mainstream alternatives to Theora. So maybe the plan to keep the Ogg container is not so great after all...

illwieckz:
Well, so, there is probably not so much thing to save from that OGM thing then. :-/

We really need to track down what was removed and why, if possible with references to nuking commits.
Also we really need to get more test assets. When this OGM code broke, no one noticed it… The reason why I did not touched the MDV code when I revamped MD5 and IQM is that because I don't have MDV file to test at this time, but we may have other things that may have been broken without knowing it.

Also we have many features Unvanquished does not use so they are rotting in engine, but also they are rotting because Unvanquished does not use them, and it's a vicious circle. For example the sun lighting from the forward renderer, we don't use it so it rots, it rots so we don't use it…

illwieckz:
I guess the future way to go for video would be Webm with Opus and VPX. Anyway, if a better implementation than our current one does Theora and Vorbis with proper OGG container (not OGM), I will not put a veto on it because that seems good enough to me. I see no real problem per se with Theora even if it's dated, at least support is wide enough and there is all the tools people needs for this format.

If you really think there is nothing to save from that cinematic code then drop it, but to me the reason is not Theora but OGM, especially OGM with Theora, that's insane. Theora per se is good enough.

src/engine/renderer/tr_shader.cpp
```diff
@@ -2214,26 +2214,8 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
...
-			}
-
-			stage->bundle[ 0 ].videoMapHandle = ri.CIN_PlayCinematic( token, 0, 0, 512, 512, ( CIN_loop | CIN_silent | CIN_shader ) );
-
-			if ( stage->bundle[ 0 ].videoMapHandle != -1 )
-			{
-				stage->bundle[ 0 ].isVideoMap = true;
-				stage->bundle[ 0 ].image[ 0 ] = tr.scratchImage[ stage->bundle[ 0 ].videoMapHandle ];
-			}
+			Log::Warn("videoMap unsupported");
```
illwieckz, line 2217: Can you put a comment explaining a bit why this was removed, with a link to this PR?

slipher:
The problem with Theora is that the library is dead and there is no usable release. The last release was 10 years ago and contains the very bad out of bounds indexing bug. This is blocking me from creating a new build for the external_deps libraries (which will fix the MacOS Catalina crashing bug, among others).

illwieckz:
What would prevent us to download a patch to apply with the build script? That's basically what does most distro on such situation.

slipher:
To keep around a bit-rotted library that we don't use, never receives updates, and probably has a bunch of security bugs brings us only harm.

illwieckz:
OKOK, so let's nuke it and kill the child of the child on seven generations… :-/

src/engine/renderer/tr_shader.cpp
slipher: Added a link here. I put a description of why it is done in the commit message.

slipher:
Thanks. If you find a map with `videoMap` or something, I'd be happy to help recreate such functionality.

src/engine/renderer/tr_shader.cpp
illwieckz: Thank you very much.

PR #392, opened 2020-10-26
deprecate forward renderer, requires r_dynamicLight to be -1

illwieckz:
With this, positive value for `r_dynamicLight ` would enable the tiled renderer, which is the only verified one for dynamic lights.

- This ensure people already having this value set to 1 will get the best we have on 0.52.0 release,
- This ensure people setting this value to 1 will not get what they don't expect.

In the future, we may do “if value equals to 2 then set 1”, but for the moment we can live with any positive value to enable the tiled renderer (and it allows 0.51 and 0.52 cohabitation).

At some point in the future we would want to nuke the forward renderer. In fact it already started to rot: the dynamic lighting seems to be broken. I'm a bit concerned about such huge code deletion since some features are not yet ported (things like real time shadows which is really buggy anyway, and sun lighting, I don't know how to test but I would like to be used).

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -546,7 +546,7 @@ static std::string GenEngineConstants() {
 			AddDefine( str, "r_ShowParallelShadowSplits", 1 );
 	}
 
-	if ( r_dynamicLight->integer )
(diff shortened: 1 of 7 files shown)


slipher:
LGTM

PR #393, opened 2020-10-27
Remove DEPS_DIR from Mac rpath

### LINKED ISSUE #387: Redundancy in Mac shared library search paths
For the Mac build, the binary's rpath (search path for shared libraries) is set to `@executable_path/;${DEPS_DIR};${DEPS_DIR}/lib`, meaning it looks for them in both the binary's own directory, and in `external_deps`. However, the build script also copies the deps into the same directory as the binary, as it does for Windows builds. So the deps can be found at two different places for a Mac build. I encountered this confusing situation when testing with an updated version of a dependency.

We should either stop copying the deps into the build dir, or remove the external_deps entries from the rpath. I would prefer the latter, since it is better to avoid having the external_deps rpath entries in the official binaries. And it reduces the number of different ways that things are done (by making it work the same as Windows).

slipher:
It is redundant since we have the executable's directory in the rpath
also, and we copy the dependencies to this directory in the build.

Fixes #387.

### DIFF
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -335,7 +335,7 @@ endif()
 if (APPLE)
     add_definitions(-DMACOS_X)
     set_linker_flag("-Wl,-no_pie")
-    set(CMAKE_INSTALL_RPATH "@executable_path/;${DEPS_DIR};${DEPS_DIR}/lib")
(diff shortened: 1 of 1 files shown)


illwieckz (approved):
LGTM.

PR #394, opened 2020-10-27
WIP: resize textures using GL_PROXY mechanism (and other things)

illwieckz:
1. Resize textures using GL_PROXY mechanism
2. Introduce picMin and picMax

The ATI X1950 PRO can display the game with medium preset including full size textures except the skybox that is `1024^6` size:

[picmax](https://dl.illwieckz.net/b/daemon/features/picmax/unvanquished_2020-10-27_184412_000.jpg)

With `r_picMax` set to `512`, everything looks good:

[picmax](https://dl.illwieckz.net/b/daemon/features/picmax/unvanquished_2020-10-27_184517_000.jpg)

The `r_picMax` cvar expects the size of an edge, so `r_picMax` 512 will scale down every image larger than a square image of `512×512` size.

Using `r_picMax 512` will scale down textures from `tex-vega` that are `1024×1024` large but will not scale down textures from `tex-pk01` that are `512x512` large. This way a map using both vega and pk01 textures will look highly textured with `512×512` textures everywhere while using `r_picmip 1` would have displayed a mix of nice `512x512` and ugly `256×256` images.

Both `r_picmip` and `r_picMax` can be used at the same time. For example, by using `r_picmip 1` and `r_picMax 512`, once minimap is set as nopicmip (see https://github.com/Unvanquished/Unvanquished/pull/1230 ), the game will scale down all textures but the minimap to keep it clear, unless it is larger than 512, _but minimaps with size up to 512 will not be scaled down_…

I added two shader keywords: `picMin` and `picMax`.

The `picMin` one allows an artist to tell the renderer to never downscale a texture under this size. This is useful for models for example, since seeing details like the eyes and others is better.

The `picMax` allows an artist to tell the renderer to downscale the texture to this size, even if the player is playing full size.

I added another cvar, `r_picMin`, which allows to override the `picMin` value set in shader, and when it is set to `-1`, never downscale images that has `picMin` value (it would be like making them `nopicmip`).

### DIFF
--- a/src/engine/renderer/InternalImage.cpp
+++ b/src/engine/renderer/InternalImage.cpp
@@ -761,6 +761,163 @@ int R_GetInternalImageSize( const image_t *image )
 	return 8 > imageSize ? 8 : imageSize;
 }
 
+int R_GetBlockSize( const image_t *image )
+{
+	switch ( image->internalFormat )
+	{
+		// S3TC formats
+
+		case GL_COMPRESSED_RGB_S3TC_DXT1_EXT:
+		case GL_COMPRESSED_SRGB_S3TC_DXT1_EXT:
+			return 8;
+
+		case GL_COMPRESSED_RGBA_S3TC_DXT3_EXT:
+		case GL_COMPRESSED_SRGB_ALPHA_S3TC_DXT3_EXT:
+			return 16;
+
+		case GL_COMPRESSED_RGBA_S3TC_DXT5_EXT:
+		case GL_COMPRESSED_SRGB_ALPHA_S3TC_DXT5_EXT:
+			return 16;
+
+		// RGTC formats
+
+		case GL_COMPRESSED_RED_RGTC1:
+			return 8;
+		case GL_COMPRESSED_RG_RGTC2:
+			return 16;
+
+		/* Others
+
+		Assume 16 so it may be large enough. */
+
+		default:
+			Log::Debug( "Unknown block size for undocumented internal format %i for image %s", image->internalFormat, image->name );
+			return 16;
+	}
+}
+
+int R_GetMipSize( int imageWidth, int imageHeight, int blockSize )
+{
+	return ( ( imageWidth + 3 ) >> 2 ) * ( ( imageHeight + 3 ) >> 2 ) * blockSize;
+}
+
+int R_GetMipSize( const image_t *image )
+{
+	int imageWidth = image->uploadWidth;
+	int imageHeight = image->uploadHeight;
+	int blockSize = R_GetBlockSize( image );
+
+	return R_GetMipSize( imageWidth, imageHeight, blockSize );
+}
+
+int R_GetImageHardwareScalingStep( image_t *image, GLenum format )
+{
+	/* From https://www.khronos.org/opengl/wiki/GLAPI/glTexImage3D
+
+	> If target is GL_PROXY_TEXTURE_3D, no data is read from data,
+	> but all of the texture image state is recalculated, checked
+	> for consistency, and checked against the implementation's
+	> capabilities. If the implementation cannot handle a texture
+	> of the requested texture size, it sets all of the image state
+	> to 0, but does not generate an error (see glGetError).
+	> To query for an entire mipmap array, use an image array level
+	> greater than or equal to 1.
+
+	From https://www.khronos.org/opengl/wiki/GLAPI/glTexImage3D
+
+	> To then query this state, call glGetTexLevelParameter. */
+
+	void ( *functionCompressed3D ) (GLenum, GLint, GLenum, GLsizei, GLsizei, GLsizei, GLint, GLsizei, const GLvoid*);
+	void ( *function3D ) (GLenum, GLint, GLint, GLsizei, GLsizei, GLsizei, GLint, GLenum, GLenum, const GLvoid*);
+	void ( *functionCompressed2D ) (GLenum, GLint, GLenum, GLsizei, GLsizei, GLint, GLsizei, const GLvoid*);
+	void ( *function2D ) (GLenum, GLint, GLint, GLsizei, GLsizei, GLint, GLenum, GLenum, const GLvoid*);
+
+	GLenum target;
+
+	constexpr GLint border = 0;
+	constexpr GLenum type = GL_UNSIGNED_BYTE;
+
+	bool texture3D = false;
+	bool textureCompressed = format == GL_NONE;
+
+	int mipSize = R_GetMipSize( image );
+
+	switch ( image->type )
+	{
+		case GL_TEXTURE_3D:
+			texture3D = true;
+			functionCompressed3D = glCompressedTexImage3D;
+			function3D = glTexImage3D;
+			target = GL_PROXY_TEXTURE_3D;
+			break;
+
+		case GL_TEXTURE_2D:
+			functionCompressed2D = glCompressedTexImage2D;
+			function2D = glTexImage2D;
+			target = GL_PROXY_TEXTURE_2D;
+			break;
+
+		case GL_TEXTURE_CUBE_MAP:
+			functionCompressed2D = glCompressedTexImage2D;
+			function2D = glTexImage2D;
+			target = GL_PROXY_TEXTURE_CUBE_MAP;
+			break;
+
+		default:
+			ASSERT_UNREACHABLE();
+			return 0;
+	}
+
+	if ( texture3D )
+	{
+		if ( textureCompressed )
+		{
+			functionCompressed3D( target, 0, image->internalFormat, image->uploadWidth, image->uploadHeight, 0, border, mipSize, nullptr );
+
+			GL_CheckErrors();
+		}
+		else
+		{
+			function3D( target, 0, image->internalFormat, image->uploadWidth, image->uploadHeight, 0, border, format, type, nullptr );
+
+			GL_CheckErrors();
+		}
+	}
+	else
+	{
+		if ( textureCompressed )
+		{
+			functionCompressed2D( target, 0, image->internalFormat, image->uploadWidth, image->uploadHeight, border, mipSize, nullptr );
+
+			GL_CheckErrors();
+		}
+		else
+		{
+			function2D( target, 0, image->internalFormat, image->uploadWidth, image->uploadHeight, border, format, type, nullptr );
+
+			GL_CheckErrors();
+		}
+	}
+
+	GLint finalWidth;
+	glGetTexLevelParameteriv( target, 0, GL_TEXTURE_WIDTH, &finalWidth );
+
+	if ( finalWidth == 0 )
+	{
+		image->uploadWidth >>= 1;
+		image->uploadHeight >>= 1;
+
+		if ( image->uploadWidth <= 1 || image->uploadHeight <= 1 )
+		{
+			return 1;
+		}
+
+		return 1 + R_GetImageHardwareScalingStep( image, format );
+	}
+
+	return 0;
+}
+
 int R_GetImageCustomScalingStep( const image_t *image, imageParams_t *imageParams )
 {
 	int scalingStep = 0;
--- a/src/engine/renderer/InternalImage.cpp
+++ b/src/engine/renderer/InternalImage.cpp
@@ -1,0 +1,129 @@
+/*
+===========================================================================
+
+Daemon BSD Source Code
+Copyright (c) 2013-2016 Daemon Developers
+All rights reserved.
+
+This file is part of the Daemon BSD Source Code (Daemon Source Code).
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of the Daemon developers nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL DAEMON DEVELOPERS BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+===========================================================================
+*/
+// InternalImage.cpp
+#include "tr_local.h"
+
+// Comments may include short quotes from other authors.
+
+int R_GetImageCustomScalingStep( const image_t *image, imageParams_t *imageParams )
+{
+	int scalingStep = 0;
+
+	// Perform optional picmip operation.
+	if ( !( image->bits & IF_NOPICMIP ) )
+	{
+		int imageMinDimension = std::max( -1, r_imageMinDimension->integer );
+		int imageMaxDimension = std::max( 0, r_imageMaxDimension->integer );
+		int shaderMinDimension = 0;
+		int shaderMaxDimension = 0;
+
+		if ( imageParams != nullptr )
+		{
+			shaderMinDimension = std::max( 0, imageParams->minDimension );
+			shaderMaxDimension = std::max( 0, imageParams->maxDimension );
+		}
+
+		// If shader has imageMaxDimension value set, enforce downscaling using this value.
+		if ( shaderMaxDimension != 0 )
+		{
+			// Except if r_imageMaxDimension is set and smaller.
+			if ( imageMaxDimension == 0 || imageMaxDimension >= shaderMaxDimension )
+			{
+				imageMaxDimension = shaderMaxDimension;
+			}
+		}
+
+		// If r_imageMinDimension == -1 and shader has imageMinDimension value set, keep original size.
+		if ( imageMinDimension == -1 && shaderMinDimension != 0 )
+		{
+			imageMaxDimension = 0;
+		}
+
+		if ( imageMaxDimension != 0 )
+		{
+			// If image has imageMinDimension value set and r_imageMinDimension > 0,
+			// use greater value between imageMinDimension and imageMaxDimension.
+			if ( shaderMinDimension > 0 && imageMinDimension > 0 )
+			{
+				imageMaxDimension = std::max( imageMinDimension, imageMaxDimension );
+			}
+
+			// Do not downscale this image below imageMinDimension size.
+			if ( imageMaxDimension < shaderMinDimension )
+			{
+				imageMaxDimension = shaderMinDimension;
+			}
+
+			// Downscale image to imageMaxDimension size.
+			int scaledWidth = image->width;
+			int scaledHeight = image->height;
+
+			while ( imageMaxDimension < scaledWidth || imageMaxDimension < scaledHeight )
+			{
+				scaledWidth >>= 1;
+				scaledHeight >>= 1;
+				scalingStep++;
+			}
+		}
+
+		scalingStep = std::max( scalingStep, std::max( 0, r_picmip->integer ) );
+	}
+
+	return scalingStep;
+}
+
+void R_DownscaleImageDimensions( int scalingStep, int *scaledWidth, int *scaledHeight, const byte ***dataArray, int numLayers, int *numMips )
+{
+	if ( scalingStep > 0 )
+	{
+		*scaledWidth >>= scalingStep;
+		*scaledHeight >>= scalingStep;
+
+		if( *dataArray && *numMips > scalingStep ) {
+			*dataArray += numLayers * scalingStep;
+			*numMips -= scalingStep;
+		}
+	}
+
+	// Clamp to minimum size.
+	if ( *scaledWidth < 1 )
+	{
+		*scaledWidth = 1;
+	}
+
+	if ( *scaledHeight < 1 )
+	{
+		*scaledHeight = 1;
+	}
+}
--- a/src/engine/renderer/InternalImage.cpp
+++ b/src/engine/renderer/InternalImage.cpp
@@ -36,6 +36,731 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // Comments may include short quotes from other authors.
 
+/*
+===============
+GetInternalImageSize
+
+Returns internal size in GPU memory in bytes.
+===============
+*/
+int R_GetInternalImageSize( const image_t *image )
+{
+
+	constexpr int s8 = 8;
+	constexpr int s16 = 16;
+	constexpr int i8 = 8;
+	constexpr int i16 = 16;
+	constexpr int i32 = 32;
+	constexpr int ui2= 2;
+	constexpr int ui8 = 8;
+	constexpr int ui10 = 10;
+	constexpr int ui16 = 16;
+	constexpr int ui32 = 32;
+	constexpr int f10 = 10;
+	constexpr int f11 = 11;
+	constexpr int f16 = 16;
+	constexpr int f32 = 32;
+
+	int texelPerBlock = 1;
+	int texelDivisor = 1;
+
+	int texelSize;
+	int imageSize;
+
+	int texelCount = image->uploadWidth * image->uploadHeight;
+
+	switch ( image->internalFormat )
+	{
+		/* Sized formats
+
+		Values described as:
+			red + green + blue + alpha + shared
+		borrowed from:
+			https://www.khronos.org/opengl/wiki/GLAPI/glTexStorage3D */
+
+		case GL_R8:
+			texelSize = 8;
+			break;
+		case GL_R8_SNORM:
+			texelSize = s8;
+			break;
+		case GL_R16:
+			texelSize = 16;
+			break;
+		case GL_R16_SNORM:
+			texelSize = s16;
+			break;
+		case GL_RG8:
+			texelSize = 8 + 8;
+			break;
+		case GL_RG8_SNORM:
+			texelSize = s8 + s8;
+			break;
+		case GL_RG16:
+			texelSize = 16 + 16;
+			break;
+		case GL_RG16_SNORM:
+			texelSize = s16 + s16;
+			break;
+		case GL_R3_G3_B2:
+			texelSize = 3 + 3 + 2;
+			break;
+		case GL_RGB4:
+			texelSize = 4 + 4 + 4;
+			break;
+		case GL_RGB5:
+			texelSize = 5 + 5 + 5;
+			break;
+		case GL_RGB8:
+			texelSize = 8 + 8 + 8;
+			break;
+		case GL_RGB8_SNORM:
+			texelSize = s8 + s8 + s8;
+			break;
+		case GL_RGB10:
+			texelSize = 10 + 10 + 10;
+			break;
+		case GL_RGB12:
+			texelSize = 12 + 12 + 12;
+			break;
+		case GL_RGB16_SNORM:
+			texelSize = 16 + 16 + 16;
+			break;
+		case GL_RGBA2:
+			texelSize = 2 + 2 + 2 + 2;
+			break;
+		case GL_RGBA4:
+			texelSize = 4 + 4 + 4 + 4;
+			break;
+		case GL_RGB5_A1:
+			texelSize = 5 + 5 + 5 + 1;
+			break;
+		case GL_RGBA8:
+			texelSize = 8 + 8 + 8 + 8;
+			break;
+		case GL_RGBA8_SNORM:
+			texelSize = s8 + s8 + s8 + s8;
+			break;
+		case GL_RGB10_A2:
+			texelSize = 10 + 10 + 10 + 2;
+			break;
+		case GL_RGB10_A2UI:
+			texelSize = ui10 + ui10 + ui10 + ui2;
+			break;
+		case GL_RGBA12:
+			texelSize = 12 + 12 + 12 + 12;
+			break;
+		case GL_RGBA16:
+			texelSize = 16 + 16 + 16 + 16;
+			break;
+		case GL_SRGB8:
+			texelSize = 8 + 8 + 8;
+			break;
+		case GL_SRGB8_ALPHA8:
+			texelSize = 8 + 8 + 8 + 8;
+			break;
+		case GL_R16F:
+			texelSize = f16;
+			break;
+		case GL_RG16F:
+			texelSize = f16 + f16;
+			break;
+		case GL_RGB16F:
+			texelSize = f16 + f16 + f16;
+			break;
+		case GL_RGBA16F:
+			texelSize = f16 + f16 + f16 + f16;
+			break;
+		case GL_R32F:
+			texelSize = f32;
+			break;
+		case GL_RG32F:
+			texelSize = f32 + f32;
+			break;
+		case GL_RGB32F:
+			texelSize = f32 + f32 + f32;
+			break;
+		case GL_RGBA32F:
+			texelSize = f32 + f32 + f32 + f32;
+			break;
+		case GL_R11F_G11F_B10F:
+			texelSize = f11 + f11 + f10;
+			break;
+		case GL_RGB9_E5:
+			texelSize = 9 + 9 + 9 + 0 + 5;
+			break;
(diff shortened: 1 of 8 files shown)


slipher:
It looks like there is some code right next to the code you added which is supposed to do exactly the same thing. What is `glConfig2.maxCubeMapTextureSize` on your test machine?

illwieckz:
I get `4096` on the for `glConfig2.maxCubeMapTextureSize` on the ATI X1950 PRO.

illwieckz:
Interestingly those are the values on the original code to scale down lightmaps:

```
scaledWidth: 1024, scaledHeight: 1024, glConfig2.maxCubeMapTextureSize: 4096
```

Is `glConfig2.maxCubeMapTextureSize` the size of a single face or the size of the whole cube? If it's the size of a single face then the driver is buggy, if it's the size of the whole cube then we have to compare `scaledWidth` and `scaledHeight` with `glConfig2.maxCubeMapTextureSize / 6` instead.

That said, this hardware is not meant to support `4096` sized standard flat textures, so it's probably the summed size of all the faces of the skybox.

Testing against `glConfig2.maxCubeMapTextureSize / 6` fixes the issue.

I still think the `r_picMax` feature is good even if unneeded to fix bugs. For example setting `r_picMax` to `512` still produces high definition rendering and it's better than doing `r_picmip 1` that would reduce `4096` sided large textures to still-heavy `2048` ones and ok-ish `512` sided textures to ugly `256` ones.

Basically `r_picmip` looks a relic from a past where all textures had the same size and were low definition anyway, and then user would somewhat expect the resulting images having the same size. To me this `r_picMax` feature is more useful than `r_picmip`, this is what I would naturally look for instead.

illwieckz:
I added a commit to fix skybox scaling according to the hardware limits.

src/engine/renderer/tr_image.cpp
```diff
@@ -865,7 +865,7 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips,
 	// deal with a half mip resampling
 	if ( image->type == GL_TEXTURE_CUBE_MAP )
 	{
```
slipher, line 880: This is the existing code I was talking about. Instead of creating yet another downscaling mechanism, I suggest using this one by, say, taking the min of picmax and the glconfig's max size value.
Although I think the existing code is broken for compressed textures, since it doesn't set `picmip`.

slipher (commented):
I doubt you are supposed to divide the cubemap max size by 6, but who knows. According to [this](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetError.xhtml) the max size is supposed to be at least 1024, so reporting the correct value would be illegal... I wonder if 4096 is a default value when the implementation doesn't provide a max size.

I just noticed something really stupid... GL error checking is disabled by default (`r_ignoreGLErrors 1`). Turning it on, however, terminates the program the first time as an error. Of course the sensible thing to do would be a non-fatal log message. This should help diagnose things, since using a texture larger than the max size is one of the GL errors which can be produced.

I agree `r_picmip` is a puzzling feature. I guess you could use it to test the downsampling code, or as a brute force way to reduce GPU memory. Both of which could be done with picmax as well. A quick google suggests it has been mostly popular as a cheat to make enemies more visible in various games.

illwieckz:
I rewrote entirely the code to compute an image size, or to be more precise, to compute a pixel size in memory (see #378).

I added two shader keywords, `picMin` and `picMax`.

The `picMin` one allows an artist to tell the renderer to never downscale a texture under this size. This is useful for models for example, since seeing details like the eyes and others is better.

The `picMax` allows an artist to tell the renderer to downscale the texture to this size, even if the player is playing full size. This can be useful for example if a texture is shipped in `4096×4096` size but is only displayed on a tiny surface. This allow to ship full size textures even if they are used downscaled in game. It's better for conservancy.

I added another cvar, `r_picMin`, which allows to override the `picMin` value set in shader.

So basically, texture is downscaled using  `picMax` (or `r_picMax` if positive) unless a shader uses `picMin`. If a shader uses `picMin`, the texture is downscaled to the largest value between `picMin` or `picMax` unless `r_picMin` is set. If a shader uses `picMin` and `r_picMin` is set to something greater than zero, then texture is downscaled using `r_picMin`. There is a special value for `r_picMin`: if a shader uses `picMin` and `r_picMin` is set to `-1`, then texture is not downscaled at all. This makes possible to downscale a lot what is known to be downscalable safely, and to not downscale at all what is known to not downscale properly. This is useful for competitive scenes: it makes possible to make world surfaces using only one color but models would be rendered properly.

Pictures may help to understand, in all those screenshots, models shaders have `picMin 128` instruction.

`r_picMax 512` and `r_picMin 0`, every texture is loaded with `512×512` size maximum.

[picmax](https://dl.illwieckz.net/b/daemon/features/picmax/unvanquished_2020-10-31_005229_000.jpg)

`r_picMax 64` and `r_picMin 0`, every texture is loaded with `64×64` size maximum, except models that are loaded with `128×128` size.

[picmax](https://dl.illwieckz.net/b/daemon/features/picmax/unvanquished_2020-10-31_005252_000.jpg)

`r_picMax 1` and `r_picMin 0`, every texture is loaded with `1×1` size maximum, except models that are loaded with `128×128` size.

[picmax](https://dl.illwieckz.net/b/daemon/features/picmax/unvanquished_2020-10-31_005313_000.jpg)

`r_picMax 1` and `r_picMin -1`, every texture is loaded with `1×1` size maximum, except models that are loaded full size.

[picmax](https://dl.illwieckz.net/b/daemon/features/picmax/unvanquished_2020-10-31_005350_000.jpg)

slipher:
> The picMax allows an artist to tell the renderer to downscale the texture to this size, even if the player is playing full size. This can be useful for example if a texture is shipped in 4096×4096 size but is only displayed on a tiny surface. This allow to ship full size textures even if they are used downscaled in game.

This doesn't sound like a great practice to encourage, since you would still pay full price in file size and load times. Wouldn't something like this be better as part of the map build process?

illwieckz:
This is better as part of a map build process, but a mapper can still write shaders based on third-party texture set and other situations.

Note that I did not added that because of a urge need. I see a real need for picMin so I did it, and right after than I've seen picMax could be done for free. There would be no issue to not have it, but once picMin is implemented, picMax costs nothing, so why not.

src/engine/renderer/tr_image.cpp
```diff
@@ -143,6 +143,417 @@ void GL_TextureMode( const char *string )
 	}
 }
 
+int GetFormatPixelSize( GLenum internalFormat )
+{
+
+	int s8 = 8;
```
slipher, line 149: const / constexpr?

src/engine/renderer/tr_local.h
```diff
@@ -597,6 +597,7 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 
 		int            frameUsed; // for texture usage in frame statistics
 
+		int            formatSizeFactor;
```
slipher, line 608: unused

src/engine/renderer/tr_image.cpp
```diff
@@ -143,6 +143,417 @@ void GL_TextureMode( const char *string )
...
+			return 3 + 3 + 2;
+		case GL_RGB4:
+			return 4 + 4 + 4;
+		case GL_RGB5:
+			return 5 + 5 + 5;
+		case GL_RGB8:
+			return 8 + 8 + 8;
+		case GL_RGB8_SNORM:
+			return s8 + s8 + s8;
+		case GL_RGB10:
```
slipher, line 199: Are you sure all these ones which are not divisible by 8 actually take up a non-integer number of bytes?

Probably the engine can't actually load any of the ones like that, so one option would be to just remove ones which are too much work to figure out.

src/engine/renderer/tr_image.cpp
```diff
@@ -170,286 +580,270 @@ void R_ImageList_f()
 	{
 		image = (image_t*) Com_GrowListElement( &tr.images, i );
 		char buffer[ MAX_TOKEN_CHARS ];
+		char imageType[ MAX_TOKEN_CHARS ];
```
slipher, line 583: Just use `std::string`

src/engine/renderer/tr_image.cpp
```diff
@@ -1057,6 +1431,58 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips,
...
+	The OpenGL Red Book says:
+
+	> glGetIntegerv(GL_MAX_TEXTURE_SIZE,...) tells you the largest dimension (width or height,
+	> without borders) of a texture image, typically the size of the largest square texture
+	> supported. However, GL_MAX_TEXTURE_SIZE does not consider the effect of the internal
+	> format of a texture. A texture image that stores texels using the GL_RGBA16 internal
+	> format may be using 64 bits per texel, so its image may have to be 16 times smaller
+	> than an image with the GL_LUMINANCE4 internal format. (Also, images requiring borders
+	> or mipmaps may further reduce the amount of available memory.) 
+	-- http://www.glprogramming.com/red/chapter09.html
```
slipher, line 1500: How do you know which texture format GL_MAX_TEXTURE_SIZE is assuming?

It seems better to use the the method described in the next section, *Texture Proxy*. In that case, it is not necessary to have a database of the number of bits per texel, or to guess at what bit density the maximum value assumes, or whether one should consider the edge length or the total number of pixels.

src/engine/renderer/tr_image.cpp
```diff
@@ -143,6 +143,417 @@ void GL_TextureMode( const char *string )
 	}
 }
 
+int GetFormatPixelSize( GLenum internalFormat )
```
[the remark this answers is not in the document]
illwieckz: In Memory, I can rename it right. Also, it can't return perfect answer in all case, for example it is somewhat wrong with DXT textures when a picture has less than 4 pixels (maybe I can add a comment about that).

src/engine/renderer/tr_local.h
illwieckz: ah yes, a leftover from work-in-progress implementation. I noticed storing the result of division as an int was bad about precision XD.

src/engine/renderer/tr_image.cpp
illwieckz: Values taken from https://www.khronos.org/opengl/wiki/GLAPI/glTexImage2D

I think for such situation it's better to be exhaustive even if it's unused: if we start to remove some, it will be hard to know if something is missing and what.

And yes, there is formats with non-integer numbers of bytes. It's even worse with ASTC since pixel size is a float.

src/engine/renderer/tr_image.cpp
illwieckz: Things like `GL_LUMINANCE4` that is taken as example and others are not described in current OpenGL wiki, or I did not found them:
https://www.khronos.org/opengl/wiki/GLAPI/glTexImage2D

But we can find some of them in OpenGL 2.1 documentation:
https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glTexImage2D.xml
Or in OpenGL 1.1 documentation:
https://web.archive.org/web/20011231003042/http://www.opengl.org/developers/documentation/Specs/glspec1.1/node79.html

It still don't find formal definition for `GL_RGB`.

I have read people saying GL_RGB is 24 bit, interlaced (`RGBRGBRGB`), like [here](https://forum.openframeworks.cc/t/gl-rgb/72/2). That's not from Khronos though, but that's an example of an answer to solve someone's need to load data directly as `GL_RGB`.

While using `imagelist` command I discovered some of our textures are listed as using `GL_RGBA`, so I may add those formats too:

```
      -w-- -h-- -px- -size------ -mm- -type- -format- -wrap.s- -wrap.t- -name------- 
Debug: Unknown pixel size for undocumented image format 6408 
Debug: Undocumented image format 6408 for image fonts/unifont_0_0_16.png 
  35:  512  256   32      174762 no   2D      6408    s.clmp   t.clmp   fonts/unifont_0_0_16.png 
```

`6408` is `0x1908` wich is `GL_RGBA` in `gl.h`…

So, unless proven wrong,it looks like `GL_RGB` uses 8 bit per pixel, then I see no reason why `GL_RGBA` would not use 8 bit per pixel, but documentation about them is harder to find. Also, unless proven wrong, it looks like `GL_RGB` is the default format.

illwieckz: This code does not exist anymore.

src/engine/renderer/tr_image.cpp
illwieckz: done.

src/engine/renderer/tr_image.cpp
illwieckz: glconfig max size value is not used anymore.

src/engine/renderer/tr_init.cpp
```diff
@@ -1057,6 +1059,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 
 		r_picmip = ri.Cvar_Get( "r_picmip", "0",  CVAR_LATCH | CVAR_ARCHIVE );
 		AssertCvarRange( r_picmip, 0, 3, true );
+		r_picMax = ri.Cvar_Get( "r_picMax", "0",  CVAR_LATCH | CVAR_ARCHIVE );
```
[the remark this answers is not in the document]
illwieckz: I've basically done `max(0, r_picMax)`, and `r_picMin` is meant to accept `-1`. I see no issue with too large values, except they are useless. To me that does not sound enough to require an assert.

src/engine/renderer/tr_image.cpp
illwieckz: Now the code lives in its own file. I've made it as exhaustive as I could like if it was a repurposable library, even if we don't use every format on earth.

src/engine/renderer/tr_image.cpp
illwieckz: Still `TOTO`.

illwieckz:
I've added a commit temporarily named `wat` that may fix an issue… While reading the original code, I found this part looking weird and that sounded like a typo. I would appreciate opinions on that.

src/engine/renderer/tr_image.cpp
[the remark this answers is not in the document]
illwieckz: Added comments about it.

src/engine/renderer/tr_image.cpp
illwieckz: Hmm, there is still usage of `Com_sprintf`, not sure how to handle that with `std::string`.

illwieckz:
This is basically what does the `wat` commit, before:

```
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 0, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 1, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 2, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 3, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 4, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 5, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 6, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 7, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 8, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 9, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 10, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 11, layer 0
```

after:

```
Debug: Loading compressed image models/buildables/barricade/body_d with 2048×2048 original size, mip 0, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 1024×1024 size, mip 1, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 512×512 size, mip 2, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 256×256 size, mip 3, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 128×128 size, mip 4, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 64×64 size, mip 5, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 32×32 size, mip 6, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 16×16 size, mip 7, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 8×8 size, mip 8, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 4×4 size, mip 9, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 2×2 size, mip 10, layer 0
Debug: Loading compressed image models/buildables/barricade/body_d with 1×1 size, mip 11, layer 0
```

Original behavior sounded wrong, right?

illwieckz:
Unfortunately the new code to detect images being too large for the hardware does not work. I never see them being downscaled.

illwieckz:
It works, but not the way we expect it for cube maps. If I double width and height when passing them to `CheckImage2D`, the obviously too large size is detected and proper order is given to downscale, and the skybox is rendered properly (because effectively downscaled):

```
Debug: Uploading image env/shared_space_src/sky07 with 256×256 size, downscaled from 1024×1024 size, layer 0
Debug: Uploading image env/shared_space_src/sky07 with 256×256 size, downscaled from 1024×1024 size, layer 1
Debug: Uploading image env/shared_space_src/sky07 with 256×256 size, downscaled from 1024×1024 size, layer 2
Debug: Uploading image env/shared_space_src/sky07 with 256×256 size, downscaled from 1024×1024 size, layer 3
Debug: Uploading image env/shared_space_src/sky07 with 256×256 size, downscaled from 1024×1024 size, layer 4
Debug: Uploading image env/shared_space_src/sky07 with 256×256 size, downscaled from 1024×1024 size, layer 5
```

illwieckz:
Well, if I do this (notice the `<< 3`):

```c++
	scalingStep = CheckImage2D( internalFormat, scaledWidth << 3, scaledHeight << 3, 0, format, GL_UNSIGNED_BYTE );
```

it works:

```
Warn: Image env/shared_space_src/sky07 too large, downscaling 1 time(s) to fit the hardware limits 
Debug: Uploading image env/shared_space_src/sky07 with 512×512 size, downscaled from 1024×1024 size, layer 0 
Debug: Uploading image env/shared_space_src/sky07 with 512×512 size, downscaled from 1024×1024 size, layer 1 
Debug: Uploading image env/shared_space_src/sky07 with 512×512 size, downscaled from 1024×1024 size, layer 2 
Debug: Uploading image env/shared_space_src/sky07 with 512×512 size, downscaled from 1024×1024 size, layer 3 
Debug: Uploading image env/shared_space_src/sky07 with 512×512 size, downscaled from 1024×1024 size, layer 4 
Debug: Uploading image env/shared_space_src/sky07 with 512×512 size, downscaled from 1024×1024 size, layer 5 
```

illwieckz:
I noticed that, on this computer reporting `1024` max texture size, if I multiply a `1024` sized texture by `4`, it does not scale down, then display garbage, meaning `4096` is expected to be OK but is not. While multiplying by `5` downscaling happens (and then displays the skybox properly).

illwieckz:
The current branch works, but I have no explanation for the multiplier. I've set `6` like the amount of faces a cube has, and a comment explaining there is no explanation.

illwieckz:
So, I renamed the `wat` commit to `upload image mips using mipWidth and mipHeight size`, that makes sense to me.

illwieckz:
At this point I think it's ready.

src/engine/renderer/tr_image.cpp
```diff
@@ -785,6 +785,81 @@ R_ConvertBC5Image(const byte **in, byte **out, int numMips, int numLayers,
...
+	glTexImage2D( GL_PROXY_TEXTURE_2D, 0, internalFormat, width, height, border, format, type, nullptr );
+
+	GLint finalWidth;
+	glGetTexLevelParameteriv( GL_PROXY_TEXTURE_2D, 0, GL_TEXTURE_WIDTH, &finalWidth );
+
+	if ( finalWidth == 0 )
+	{
+		width >>= 1;
+		height >>= 1;
+		return 1 + CheckImage2D( internalFormat, width, height, border, format, type );
```
slipher, line 857: These functions need to return if width and height get down to 1, to make sure there can't be infinite recursion.

src/engine/renderer/tr_image.cpp
slipher: `.c_str()` to get a char*

src/engine/renderer/tr_image.cpp
```diff
@@ -833,13 +896,61 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips,
 	scaledWidth = image->width;
 	scaledHeight = image->height;
 
+	int picMin = r_picMin->integer;
```
slipher, line 899: The logic to decide mip level is getting very long. Can you put it in a separate function?

src/engine/renderer/tr_image.cpp
```diff
@@ -796,6 +785,81 @@ R_ConvertBC5Image(const byte **in, byte **out, int numMips, int numLayers,
...
+	/* It have been noticed ATI X1950 PRO reports 4096 max cube map size
+	and it does not work with cubemaps having 1024 sized edge.
+	It was also noticed than when multiplying width and height by at least 5,
+	faces get downscaled one time to 512 and they work.
+
+	At this point there is no explanation for the need of such multiplier,
+	and if there is one, what is the multiplier to use,
+	so let's just use the amount of faces, 6. */
+
+	glTexImage2D( GL_PROXY_TEXTURE_CUBE_MAP, 0, internalFormat, width * 6, height * 6, border, format, type, nullptr );
```
slipher, line 831: This calculation doesn't seem right, since if you multiply both the width and height by 6, then the image uses 36 times as much memory.

But in any case it doesn't seem good to add a hack like this, since it could cause images to be wrongly downscaled on hardware that actually supports it.

slipher:
Have you checked the OpenGL error codes to see if there's anything relevant in the cubemap bug case? I opened https://github.com/DaemonEngine/Daemon/pull/395 to enable this checking all the time.

src/engine/renderer/tr_image.cpp
```diff
@@ -811,13 +811,61 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips, image_t
...
+			picMax = imagePicMin;
+		}
+
+		// Downscale image to picMax size.
+		int maxImageSize = picMax * picMax;
+		int imageSize = image->width * image->height;
+
+		while ( maxImageSize < imageSize )
+		{
+			imageSize >>= 2;
```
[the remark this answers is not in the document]
illwieckz: Right, I modified it to compare against width and height independently, like everything else in the existing code does. Also it looks like GL compare edges.

src/engine/renderer/tr_image.cpp
illwieckz: Done

src/engine/renderer/tr_image.cpp
```diff
@@ -811,13 +811,61 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips, image_t
...
+		}
+
+		// Do not downscale this image below picMin size.
+		if ( picMax < imagePicMin )
+		{
+			picMax = imagePicMin;
+		}
+
+		// Downscale image to picMax size.
+		int maxImageSize = picMax * picMax;
```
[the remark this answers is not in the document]
illwieckz: Modified it to compare against width and height independently.

src/engine/renderer/tr_image.cpp
illwieckz: Done

src/engine/renderer/tr_image.cpp
illwieckz: Deleted the bizarre `6`. Also, I noticed I was doing the check before internalFormat is determined (hence always using the same default one).

illwieckz:
I made an extra function to downscale dimensions that not only applies on `scaleHeight` and `scaleWidth` but also on some other array and mip data. I noticed the extra computation was only done with `picmip` (and the variants I introduced), but not when the image is larger than the hardware can do.

We may technically do the custom downscale (picmip & all) and and the hardware-limited downscale at the same time after `internalFormat` is determined, and call the function to do the downscaling computation, but we have code to convert formats in the code determining internalFormat, so doing the custom downscale before determining the internal format may make conversion faster.

src/engine/renderer/tr_init.cpp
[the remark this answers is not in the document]
illwieckz: Now it does not compare `picMax * picMax` anymore but `picMax` itself, is there still a risk of integer overflow?

illwieckz:
> Have you checked the OpenGL error codes to see if there's anything relevant in the cubemap bug case? I opened #395 to enable this checking all the time.

If I run your PR (i.e. without my own patch from this PR), I get `Warn: OpenGL error GL_INVALID_ENUM detected` in `GLShaderManager::buildPermutation`, but no error when loading the skybox (that displays garbage).

src/engine/renderer/tr_image.cpp
illwieckz: I still don't know how to do this with `std::string`:

```c
Com_sprintf( imageType, sizeof( imageType ),  "%5i ", image->type );
```

illwieckz:
Except those two questions:

https://github.com/DaemonEngine/Daemon/pull/394/files#r519607355
https://github.com/DaemonEngine/Daemon/pull/394/files#r519510857

I consider this PR ready again.

Thanks to `GL_CheckErrors` I spotted some issues I introduced in my code, now fixed.

Note: The skybox issue is still not fixed but I give up on getting it fixed for 0.52. Anyway, hardware reproducing the issue runs better with `low` preset and this PR now introduces a nice feature that would make the game look better with `low` preset while workarounding the issue.

illwieckz:
Also, `r_picMin`, `r_picMax`, `picMin` and `picMax` keywords are temporary.

I'm looking for better cvar name than `r_picMin` and `r_picMax`, some base name that would fit well when ported to “dot” naming. Something like `maximumEdge` and `minimumEdge`, so we can do something like `renderer.textures.maximumEdge` or something like that. Material keywords would be edited accordingly.

Does anyone has an idea about that?

src/engine/renderer/tr_image.cpp
slipher: Com_sprintf -> Str::Format

src/engine/renderer/InternalImage.h
```diff
@@ -0,0 +1,48 @@
...
+
+===========================================================================
+*/
+// InternalImage.h
+
+#ifndef INTERNAL_IMAGE_H
+#define INTERNAL_IMAGE_H
+#include "tr_local.h"
+
+int GetInternalImageSize( image_t *image );
```
slipher, line 40: Use `const image_t *` on all these

slipher:
For cvar naming how about `maxImageDimension`?

Did you ever find any cases where the texture proxy successfully identified a texture that couldn't be loaded?

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,1007 @@
...
+			break;
+
+		case GL_TEXTURE_CUBE_MAP:
+			functionCompressed2D = glCompressedTexImage2D;
+			function2D = glTexImage2D;
+			target = GL_PROXY_TEXTURE_CUBE_MAP;
+			break;
+
+		default:
+			ASSERT( false );
```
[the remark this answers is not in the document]
illwieckz: I got this:

```
error: ‘ASSERT_UNREACHABLE’ was not declared in this scope
```

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,1007 @@
...
+
+			while ( picMax < scaledWidth || picMax < scaledHeight )
+			{
+				scaledWidth >>= 1;
+				scaledHeight >>= 1;
+				scalingStep++;
+			}
+		}
+
+		scalingStep = std::max( scalingStep, std::max( 0, r_picmip->integer ) );
```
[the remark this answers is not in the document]
illwieckz: I think we better just remove the enforced range of 0 to 3, I see no reason for this enforced range, especially since we may reduce an image more than `r_picmip 3` would do (even, automatically when checking against hardware capabilities). `r_picmip 4` on a 8K texture is still huge. This range, like the whole `r_picmip` feature, assumes we use `128×128` textures everywhere.

src/engine/renderer/tr_local.h
```diff
@@ -584,6 +584,14 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	static inline bool operator ==( const wrapType_t &a, const wrapType_t &b ) { return a.s == b.s && a.t == b.t; }
 	static inline bool operator !=( const wrapType_t &a, const wrapType_t &b ) { return a.s != b.s || a.t != b.t; }
 
+	struct imageParams_t
+	{
+		int bits;
```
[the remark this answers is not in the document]
illwieckz: Do I have something to do for types as well?

src/engine/renderer/InternalImage.cpp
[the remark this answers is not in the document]
illwieckz: Uh I feel stupid, there is no argument to `ASSERT_UNREACHABLE`… 🤦‍♀️

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,1007 @@
...
+		if ( imageParams != nullptr )
+		{
+			imagePicMin = std::max( 0, imageParams->picMin );
+			imagePicMax = std::max( 0, imageParams->picMax );
+		}
+
+		// If image has picMax value set, enforce downscaling using this value.
+		if ( imagePicMax != 0 )
+		{
+			picMax = imagePicMax;
```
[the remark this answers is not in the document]
illwieckz: I added a test to not use the shader value when the cvar value is defined and smaller.

src/engine/renderer/InternalImage.cpp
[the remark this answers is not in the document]
illwieckz: I added a commit to not enforce that range anymore.

src/engine/renderer/InternalImage.h
illwieckz: Done, except on `R_GetImageHardwareScalingStep` which updates `*image`.

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,1007 @@
...
+
+	GLint finalWidth;
+	glGetTexLevelParameteriv( target, 0, GL_TEXTURE_WIDTH, &finalWidth );
+
+	if ( finalWidth == 0 )
+	{
+		image->uploadWidth >>= 1;
+		image->uploadHeight >>= 1;
+
+		if ( image->uploadWidth == 1 || image->uploadHeight == 1 )
```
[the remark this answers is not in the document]
illwieckz: done

illwieckz:
I did `imageMinDimension`/`imageMaxDimension`, `r_imageMinDimension`/`r_imageMaxDimension`.

illwieckz:
I got this on debug build, I don't know how to fix it:

```gdb
#0  0x00005555558d618e in Str::AssertOnTinyFormatError (reason="\"tinyformat: Not enough conversion specifiers in format string\"") at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/common/String.cpp:37
#1  0x0000555555650b73 in tinyformat::detail::streamStateFromFormat (out=..., spacePadPositive=@0x7fffffff4181: false, ntrunc=@0x7fffffff4188: -1, fmtStart=0x5555559bcf38 "", formatters=0x7fffffff4410, argIndex=@0x7fffffff4184: 5, numFormatters=7) at /home/illwieckz/dev/buildme-unvanquished/Daemon/libs/tinyformat/tinyformat.h:607
#2  0x00005555556518e2 in tinyformat::detail::formatImpl (out=..., fmt=0x5555559bcf38 "", formatters=0x7fffffff4410, numFormatters=7) at /home/illwieckz/dev/buildme-unvanquished/Daemon/libs/tinyformat/tinyformat.h:807
#3  0x0000555555651da9 in tinyformat::vformat (out=..., fmt=0x5555559bcef8 "Uploading compressed image %s with %d×%d size, mip %d, layer %d", list=...) at /home/illwieckz/dev/buildme-unvanquished/Daemon/libs/tinyformat/tinyformat.h:963
#4  0x00005555558223c1 in tinyformat::format<char [64], int, int, unsigned short, unsigned short, int, int> (out=..., fmt=0x5555559bcef8 "Uploading compressed image %s with %d×%d size, mip %d, layer %d") at /home/illwieckz/dev/buildme-unvanquished/Daemon/libs/tinyformat/tinyformat.h:973
#5  0x0000555555821a7f in tinyformat::format<char [64], int, int, unsigned short, unsigned short, int, int> (fmt=0x5555559bcef8 "Uploading compressed image %s with %d×%d size, mip %d, layer %d") at /home/illwieckz/dev/buildme-unvanquished/Daemon/libs/tinyformat/tinyformat.h:982
#6  0x0000555555820e8b in Str::Format<char (&) [64], int&, int&, unsigned short&, unsigned short&, int&, int&> (format=...) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/common/String.h:315
#7  0x000055555581faa9 in Log::Logger::Debug<char (&) [64], int&, int&, unsigned short&, unsigned short&, int&, int&> (this=0x555556236c80 <Log::defaultLogger>, format=...) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/common/Log.h:220
#8  0x000055555581e932 in Log::Debug<char (&) [64], int&, int&, unsigned short&, unsigned short&, int&, int&> (format=...) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/common/Log.h:272
#9  0x00005555558189c2 in R_UploadImage (dataArray=0x7fffffff4b18, numLayers=1, numMips=8, image=0x7fffd5f3b200, imageParams=0x7fffffffcf70) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/engine/renderer/tr_image.cpp:1127
```

illwieckz:
I close this as commits will be cherry-picked in dedicated PR.

PR #395, opened 2020-11-08
Enable OpenGL error checking

slipher:
Also fix the existing errors which arise on my machine.


illwieckz (approved):
LGTM.

illwieckz:
Now I see it is still fatal there: https://github.com/DaemonEngine/Daemon/blob/6b94e2e890f893afdc55b7f70bd58cb39f3cafcd/src/engine/renderer/tr_cmds.cpp#L863

illwieckz:
If I replace all that custom code with:

```c++
	// check for errors
	if ( !r_ignoreGLErrors->integer )
	{
		R_SyncRenderThread();

		GL_CheckErrors();
	}
```

Now I don't get the error in `RE_BeginFrame`… :confused:

slipher:
Thanks for spotting the duplicated code in RE_BeginFrame.  You didn't enable `r_smp` did you? If so, definitely turn it off while debugging. The error checking would be entirely disabled by the `smpActive` condition.

illwieckz:
Ah yes `r_smp` is enabled.

PR #396, opened 2020-11-09
Make GL error check non-fatal in SMP mode

slipher:
Also remove duplicated code for printing names of GL errors.


illwieckz (approved):
I haven't tested the macro thing, but LGTM so if it works on your end, let's merge!

PR #397, opened 2020-11-12
Build fixes

slipher:
Fix a few build errors that can occur with different combinations of options.

Also removed support for unused platforms and an unused cvar.


illwieckz:
LGTM

PR #399, opened 2020-11-22
Various fixes and an optimization

slipher:


illwieckz:
LGTM, special mention for the shader cache invalidation, I'm sure I faced it.

I've experienced random crashes when quitting games while doing development with shared libs, does the commit you added may be related to them?

slipher:
The last commit fixes an issue where a crash in the gamelogic causes the engine to shut down, instead of stopping and restarting the gamelogic as expected.

There have always been problems with librocket shutdown (which only affect shared libs since when the cgame is a process you can just terminate it), so maybe you're seeing those.

PR #401, opened 2020-12-03
Update external_deps

### LINKED ISSUE #145: x86 WebP DLL wrongly assumes stack alignment
The libwebp-5.dll shipped in the msvc32-4 external_deps seems to be actually incompatible with MSVC. On the x86 Windows platform, only 4-byte stack alignment is guaranteed. But the code in this DLL apparently assumes that the stack will be 16-byte aligned when using SSE instructions on stack argument (causing it to crash).

I tried to find a compile switch or attribute to force 16-byte stack alignment, but no luck.

Where does this code come from anyway? Is it built by us or from a binary release?

slipher:
The updated libraries are now available on the server. I have tested them for all supported toolchains. Thanks @DolceTriade for making the final build of the Mac ones.

Fixes #145 and https://github.com/Unvanquished/Unvanquished/issues/1228.


slipher:
Make sure to delete your old build directory before building with the new deps. Many of the paths are stored in cached variables, whose values won't be recalculated if they are already present. So if you reuse an existing CMake cache, you won't actually get the new dependencies.

illwieckz:
The prebuilt things work on my end (Ubuntu 20.04 LTS).

slipher:
> The prebuilt things work on my end (Ubuntu 20.04 LTS).

Well, there is really nothing to test there. The only thing shipped for Linux is NaCl stuff, which I did not update. (I did suggest testing SDL earlier, which was a nonsensical suggestion; sorry to waste your time!)

illwieckz:
Anyway, I don't really know how that works but I attempted to build anything I can using the script and it worked (even if it is useless).

slipher:
I see, I assumed you meant you tried the binary packages downloaded from the server.

It is possible to build and use some non-NaCl libs for Linux, but I didn't try any of these builds since they were not necessary to refresh our binary packages. If you built some of these inside the `external_deps` directory, you may want to run `./build.sh linux clean` (note: the clean command is added on this branch) to ensure that you uninstall them and go back to using ones from your package manager.

illwieckz:
Is there something remaining preventing to merge this?

PR #403, opened 2020-12-07
Nicer names for binding Left Shift, etc.

slipher:
Note that these still can't be made using the menu, because the menu will try a keyNum_t-based one first (e.g. `SHIFT`) before it does a scancode-based one (e.g. `hw:LSHIFT`).

I added some extra logging so that you can more easily figure out, for example, that the "^" key of AZERTY can be referred to with `hw:[`.


illwieckz:
It works but if I set `hw:LALT` as `reload` bind, the bind menu displays `0xe2`.

illwieckz:
So I guess this one can be closed in favor of #404?

PR #404, opened 2020-12-09
Nicer names for binding Left Shift, etc

slipher:
Continuation of #403


slipher:
Is this good to merge?

https://github.com/Unvanquished/Unvanquished/pull/1241 adds support for displaying (but not binding) these in the menu.

illwieckz:
I'm not at home to test it but that looks good to me.

PR #405, opened 2020-12-20
sdl_windowevent_restored adjustment

bofh80:
According to the docs I could find, restored is the same as unminimised and should be treated as such. Also the missing break; No idea how important that is/ was.

Also I was wondering something more like extending the others as a Just in case, like so: ?
`case SDL_WINDOWEVENT_MAXIMIZED:    Cvar_SetValue( "com_minimized", 0 ); Cvar_SetValue( "com_unfocused", 0 ); break;`


slipher:
This change makes no difference in behavior. If you have multiple `case` labels in a row like
```
case A:
case B:
...
```
This means that the program will jump to the same place for both values A and B.

bofh80:
thanks, sorry for the noise..

Issue #408, opened 2020-12-29
Locally built engine seems not to be able to capture mouse


necessarily-equal:
When I start the game, if I keep it focused normally, I can play. But as soon as I select another window and switch back, it doesn't seem to catch the mouse again and I just see a fancy, useless cursor. Button press still shot or do whatever action is bound to the key, but I can't look around. I've seen the problem on Ubuntu 20.04 LTS, both with 0.52.0/sync, v0.51.1 and master branches.

may be related to #234.

illwieckz:
Fixed by #409.

Issue #234, opened 2019-12-08
when running master the engine sometime fails to capture my mouse


illwieckz:
When running master the engine sometime fails to capture my mouse, Like 1 time on 50.

illwieckz:
maybe related to https://github.com/Unvanquished/Unvanquished/issues/1136

illwieckz:
Fixed by #409.

PR #409, opened 2020-12-30
Detect that window isn't minimized more often

necessarily-equal:
Fixes https://github.com/DaemonEngine/Daemon/issues/408 and https://github.com/Unvanquished/Unvanquished/issues/1136.

Could fix https://github.com/DaemonEngine/Daemon/issues/365 and https://github.com/DaemonEngine/Daemon/issues/234.
If it doesn't, it will at least give an easy workaround: switch away from and back to the window.

### DIFF
--- a/src/engine/sys/sdl_input.cpp
+++ b/src/engine/sys/sdl_input.cpp
@@ -1157,7 +1157,18 @@ static void IN_ProcessEvents( bool dropInput )
 					case SDL_WINDOWEVENT_RESTORED:
 					case SDL_WINDOWEVENT_MAXIMIZED:    Cvar_SetValue( "com_minimized", 0 ); break;
 					case SDL_WINDOWEVENT_FOCUS_LOST:   Cvar_SetValue( "com_unfocused", 1 ); break;
-					case SDL_WINDOWEVENT_FOCUS_GAINED: Cvar_SetValue( "com_unfocused", 0 ); break;
+					case SDL_WINDOWEVENT_FOCUS_GAINED:
+
+						Cvar_SetValue( "com_unfocused", 0 );
+
+						// HACK: if the window is focused, it can't be minimized.
+						// fixes
+						//  * https://github.com/DaemonEngine/Daemon/issues/408
+						//  * https://github.com/Unvanquished/Unvanquished/issues/1136
+						// and maybe others
+						Cvar_SetValue( "com_minimized", 0 );
+
+						break;
 				}
 				break;
 			case SDL_QUIT:
@@ -1202,7 +1213,7 @@ void IN_Frame()
 		// Console is down in windowed mode
 		IN_SetFocus( false );
 	}
-	else if ( !( SDL_GetWindowFlags( window ) & SDL_WINDOW_INPUT_FOCUS ) )
+	else if ( com_unfocused->integer )
 	{
 		// Window doesn't have focus
 		IN_SetFocus( false );


necessarily-equal:
To expand a bit about why this didn't work, the window was reported as unfocused in the cvar. Because of that, `IN_Frame` calls `IN_SetFocus(false)` and that's apparently what changes input handling, blocking grab and showing a cursor instead.

src/engine/sys/sdl_input.cpp
```diff
@@ -1202,7 +1211,7 @@ void IN_Frame()
 		// Console is down in windowed mode
 		IN_SetFocus( false );
 	}
-	else if ( !( SDL_GetWindowFlags( window ) & SDL_WINDOW_INPUT_FOCUS ) )
+	else if ( com_unfocused->integer )
```
necessarily-equal, line 1214: (this one is for consistency)

illwieckz:
I don't have a way to test right now, but now while you talk about this the previous code is looking to be waiting for that! Nice finding! :tada:

src/engine/sys/sdl_input.cpp
```diff
@@ -1157,7 +1157,16 @@ static void IN_ProcessEvents( bool dropInput )
 					case SDL_WINDOWEVENT_RESTORED:
 					case SDL_WINDOWEVENT_MAXIMIZED:    Cvar_SetValue( "com_minimized", 0 ); break;
 					case SDL_WINDOWEVENT_FOCUS_LOST:   Cvar_SetValue( "com_unfocused", 1 ); break;
-					case SDL_WINDOWEVENT_FOCUS_GAINED: Cvar_SetValue( "com_unfocused", 0 ); break;
+					case SDL_WINDOWEVENT_FOCUS_GAINED: Cvar_SetValue( "com_unfocused", 0 );
+
+					                                   // HACK: if the window is focused, it can't be minimized.
```
DolceTriade, line 1162: Is it just me or does the formatting here look completely wrong?

DolceTriade (changes requested):

src/engine/sys/sdl_input.cpp
illwieckz: Well, I guess it was meant to be aligned with `Cvar_SetValue`, but one tab would be enough instead (at least I would prefer a tab to this ascii-art).

necessarily-equal: yes, it was for consistency with the other lines. Do you prefer if I put the other case on a newline for consistency?

illwieckz: If you want to restyle other lines just for the sake of doing it, I would prefer doing another commit.

illwieckz:
I confirm it fixes the input issue on my end (I still have to sometime click the windows to get my mouse get captured again, but that's not a big deal, no tty console-based vid_restart is required to unstuck the game anymore).

src/engine/sys/sdl_input.cpp
necessarily-equal: I don't how to unmark the PR as "changes requested", but I changed that.

src/engine/sys/sdl_input.cpp
illwieckz: > [from there](https://github.com/DaemonEngine/Daemon/pull/409#discussion_r551039241)
> I don't how to unmark the PR as "changes requested", but I changed that.

You have to also click “resolve” on this thread.

DolceTriade (approved):

PR #413, opened 2021-01-03
toolchain-pnacl: Use the FindPython2 module

### LINKED ISSUE #411: Make the NaCl Toolchain detect Python2 automatically
When following the instructions to build Unvanquished I got errors in the CMake toolchain python wrapper script because it is using python2-isms that were rejected by Python 3.

Daemon's build should automatically find Python2 if it is available on the system and use it. I'll make a PR for this.

Kangz:
This makes the toolchain automatically find python2 installation even on
systems with /usr/bin/python that's python3.

The FindPython2 module was intrudoced in CMake 3.12 so Daemon's CMake
requirement is bump to that version. Note that it is an older version
than what's in Debian stable and oldstable's backports.

Fixes #411

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,9 +24,12 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-cmake_minimum_required (VERSION 2.8.12)
-if (POLICY CMP0017)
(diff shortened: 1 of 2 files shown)


Kangz:
PTAL! The travis run failed on macOS but it looks to have been failing for a long time.

slipher:
Is the goal to be able to use Python 3 for CBSE while using Python 2 for NaCl?

Kangz:
The goal is to make the CMake configuration "just" work if you have at least python2 installed. It failed on my system because I have both 2 and 3 installed and 3 is used by default.

DolceTriade (approved):

PR #414, opened 2021-01-03
Fix deprecation warnings in toolchain-nacl.cmake

### LINKED ISSUE #412: Various loud CMake warnings when compiling.
Compiling Unvanquished with a recent CMake version (3.19) produces a lot of warnings about behavior being deprecated soon. I'd like to take a look at this to make the build nicer on my machine. Assigning to myself as I'll make a PR for this shortly.

Kangz:
Note that the first commit is #413 

This bumps the CMake version to 3.13 and fixes deprecations warnings
that were raised in toolchain-nacl.cmake as well as deletes some code
paths that were for older CMake versions. Not that 3.13 is what's in
Debian stable and oldstable has backports for 3.16.

CMakeForceCompiler is deprecated so instead we set CMAKE_C/CXX_COMPILER
directly. This causes CMake to not set CMAKE_C/CXX_COMPILER_ID (because
it can't determine it automatically) so the CMakeLists.txt had to
support an empty CMAKE_CXX_COMPILER_ID and a special case for PCH
support on NaCl.

Getting a target's location as done in pnacl_translate is deprecated so
instead we compute the paths to the various targets/nexe/pexe directly
from the build dir and the gamemodule dir.

Fixes #412


DolceTriade (approved):

slipher:
Well it does succeed in allowing me to use Python 3 for the CBSE deps and Python 2 for NaCl.

There is no longer a cache variable like PYTHON_EXECUTABLE to control which Python is used (for either cbse or nacl). Is there any way to see or change this? It seems kind of annoying if there's no way to set the Python interpreter if it doesn't find the one you want.

Kangz:
I tried the following but it doesn't allow overriding the python executable from the cmake call: `cmake .. -DPython2_EXECUTABLE=...`. CMake should find the Python executable, and has over 3000 lines of code to do so in its builtin modules. Using a virtualenv would also make it detect Python.

But if worse comes to worse and someone need a variable to set Python, then we could add `CBSE_PYTHON_EXECUTABLE` and `NACL_PYTHON_EXECUTABLE` variables that you can override. It just seems unlikely to happen so if you don't mind I'll leave this out of this PR.

illwieckz:
I'm OK to postpone `CBSE_PYTHON_EXECUTABLE` and `NACL_PYTHON_EXECUTABLE` to another PR (if such need surfaces one day).

One question: does implementing `CBSE_PYTHON_EXECUTABLE` and `NACL_PYTHON_EXECUTABLE` would help to test various python interpreters? That may create the need.

Kangz:
Do you mean that it would help testing python2 vs python3 for CBSE (NaCl being forced to Python2 anyway)? Yeah that could be a good rational for adding these. I'll look at it tomorrow.

illwieckz:
> Do you mean that it would help testing python2 vs python3 for CBSE?

Yes.

I feel no urge for that but I can see why we may need it at some point.

DolceTriade:
Python2 is dying soon. Hopefully we can deprecate NaCl before then.

Kangz:
Too late, Python2 is already unsupported :) but yeah my goal is to move to WASM (I know @slipher is already looking at it)

slipher:
Can you use a pattern like this?
```
set(CBSE_PYTHON_PATH "CBSEPYTHO-NOTFOUND" CACHE FILEPATH "decription")
if (NOT CBSE_PYTHON_PATH)
  FindPython()
  set(CBSE_PYTHON_PATH "${Python_Executable}" CACHE FILEPATH "decription" FORCE)
endif()
```

Kangz:
Yup

Kangz:
This turns out harder than I thought for the NaCl toolchain because it is used in a different CMake invocation and would need piping of variables around etc. That one always uses Python2 so we don't need to allow configuring the Python executable as much so I left it as is. Does that sound ok?

cmake/DaemonCBSE.cmake
```diff
@@ -24,7 +24,13 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-find_package(PythonInterp REQUIRED)
+# Allow overriding the python path with a cached variable.
+set(DAEMON_CBSE_PYTHON_PATH "NOTFOUND" CACHE FILEPATH "")
+if (NOT ${DAEMON_CBSE_PYTHON_PATH})
+    find_package(Python REQUIRED)
+set(DAEMON_CBSE_PYTHON_PATH "${Python_EXECUTABLE}" CACHE FILEPATH "" FORCE)
```
slipher, line 31: indentation

slipher:
> 
> 
> This turns out harder than I thought for the NaCl toolchain because it is used in a different CMake invocation and would need piping of variables around etc. That one always uses Python2 so we don't need to allow configuring the Python executable as much so I left it as is. Does that sound ok?

OK makes sense - there is not much need to change the NaCl one since it doesn't have any dependencies.

Should fix the message that says `Please install python2 and/or set the PYTHON_EXECUTABLE CMake variable.` though.

Kangz:
Done, I'll merge this after CI results are in.

illwieckz:
I got some merge conflict while attempting to merge `master` to `0.52.0/sync` (which is upcoming version), @Kangz can you merge yourself `master` to `0.52.0/sync` Dæmon branch and then sync the new `daemon` submodule reference in Unvanquished `0.52.0/sync` branch?

Kangz:
Ok I'll take a look

PR #416, opened 2021-01-08
Rework team damage handling

necessarily-equal:
see https://github.com/Unvanquished/Unvanquished/pull/1276.


illwieckz:
I changed branch target to `master` now that 0.52 is merged. 😉

slipher:
The tinygettext library hasn't been updated in several years. So I would suggest to try getting the latest upstream, instead of making ad hoc fixes.

PR #425, opened 2021-02-07
introduce r_imageMaxDimension (from #394)

illwieckz:
I imported `imageMinDimension`/`imageMaxDimension`, `r_imageMinDimension`/`r_imageMaxDimension` from #394.

What's interesting for the release is `r_imageMaxDimension` (engine cvar) and `imageMinDimension` (material keyword).

### DIFF
--- a/src/engine/renderer/InternalImage.cpp
+++ b/src/engine/renderer/InternalImage.cpp
@@ -1,0 +1,181 @@
+/*
+===========================================================================
+
+Daemon BSD Source Code
+Copyright (c) 2013-2016 Daemon Developers
+All rights reserved.
+
+This file is part of the Daemon BSD Source Code (Daemon Source Code).
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of the Daemon developers nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL DAEMON DEVELOPERS BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+===========================================================================
+*/
+// InternalImage.cpp
+#include "tr_local.h"
+
+// Comments may include short quotes from other authors.
+
+/*
+===============
+R_GetImageCustomScalingStep
+
+Returns amount of downscaling step to be done on the given image
+to perform optional picmip or other downscaling operation.
+
+- the r_picMip cvar tells the renderer to downscale textures as many time as this cvar value;
+- the imageMaxDimension material keyword tells the renderer to downscale related textures
+  to that dimension if natively larger than that dimension;
+- the r_imageMaxDimension cvar tells the renderer to downscale all textures to that dimension
+  if natively larger than that dimension;
+- the r_ignoreMaterialMaxDimension cvar tells the renderer to ignore the imageMaxDimension material
+  keyword, r_imageMaxDimension still applies;
+- the imageMinDimension material keyword tells the renderer to not downscale related textures
+  below that dimension, has priority over r_picMip and r_imageMaxDimension;
+- the r_ignoreMaterialMinDemension cvar tells the renderer to ignore the imageMinDimension material
+  keyword;
+- the r_replaceMaterialMinDimensionIfPresentWithMaxDimension cvar tells the renderer to use the
+  maximum texture size for textures having a material keyword setting a minimum dimension;
+
+Examples of scenarii:
+
+- r_picMip set 3 with some materials having imageMinDimension set to 128:
+  attempt to reduce three times the images, but for materials having that keyword,
+  stop before the image dimension would be smaller than 128;
+- r_imageMaxDimension set to 64 with some materials having imageMinDimension set to 128:
+  reduce images having dimension greater than 64 to fit 64 dimension, but for materials
+  having that keyword, reduce image to fit 128 dimension;
+- r_imageMaxDimension set to 64 with some materials having imageMaxDimension set to 32:
+  reduce images having dimension greater than 64 to fit 64 dimension, but for materials
+  having that keyword, reduce image to fit 32 dimension;
+- r_imageMaxDimension set to 64 and r_replaceMaterialMinDimensionIfPresentWithMaxDimension
+  enabled: reduce images having dimension greater than 64 to fit 64 dimension, but for
+  materials having that keyword, keep the native dimension, or the one set by material
+  imageMaxDimension keyword;
+
+Examples of use cases:
+
+- Movie producer wanting to record a movie from a demo using the highest image definition
+  possible even if the game developper configured them to be downscaled for performance
+  purpose:
+    set r_ignoreMaterialMaxDimension on
+- Low budget player wanting to reduce all textures to low definition but keep configured
+  minimum definition on the textures the game developer configured to not be reduced to a
+  given size to keep the game playable:
+    set r_imageMaxDimension 64
+- Very low budget player wanting to reduce all textures to low definition in all case,
+  taking risk stuff may not look as the game developers expect anyway:
+    set r_imageMaxDimension 64
+    set r_ignoreMaterialMinDimension on
+- Competitive player wanting to reduce all textures to one flat color but keep configured
+  minimum definition on the textures the game developer configured to not be reduced to a
+  given size to keep the game playable:
+    set r_imageMaxDimension 1
+- Competitive player wanting to reduce all textures to one flat color but keep high definition
+  on the textures the game developer configured to not be reduced to keep the game playable:
+    ser r_imageMaxDimension 1
+    set r_replaceMaterialMinDimensionIfPresentWithMaxDimension on
+ 
+===============
+*/
+int R_GetImageCustomScalingStep( const image_t *image, const imageParams_t &imageParams )
+{
+	if ( image->bits & IF_NOPICMIP )
+	{
+		return 0;
+	}
+
+	int materialMinDimension = r_ignoreMaterialMinDimension->integer ? 0 : imageParams.minDimension;
+	int materialMaxDimension = r_ignoreMaterialMaxDimension->integer ? 0 : imageParams.maxDimension;
+
+	int minDimension;
+
+	if ( materialMinDimension <= 0 )
+	{
+		minDimension = 1;
+	}
+	else if ( r_replaceMaterialMinDimensionIfPresentWithMaxDimension->integer )
+	{
+		minDimension = materialMaxDimension > 0 ? materialMaxDimension : std::numeric_limits<int>::max();
+	}
+	else
+	{
+		minDimension = materialMinDimension;
+	}
+
+	int maxDimension = materialMaxDimension > 0 ? materialMaxDimension : std::numeric_limits<int>::max();
+
+	if ( r_imageMaxDimension->integer > 0 )
+	{
+		maxDimension = std::min( maxDimension, r_imageMaxDimension->integer );
+	}
+
+	// Consider the larger edge as the "image dimension"
+	int scaledDimension = std::max( image->width, image->height );
+	int scalingStep = 0;
+
+	// 1st priority: scaledDimension >= minDimension
+	// 2nd priority: scaledDimension <= maxDimension
+	// 3rd priority: scalingStep >= r_picMip->integer
+	// 4th priority: scalingStep as low as possible
+	while ( scaledDimension > maxDimension || scalingStep < r_picMip->integer )
+	{
+		scaledDimension >>= 1;
+
+		if ( scaledDimension < minDimension )
+		{
+			break;
+		}
+
+		++scalingStep;
+	}
+
+	return scalingStep;
+}
+
+void R_DownscaleImageDimensions( int scalingStep, int *scaledWidth, int *scaledHeight, const byte ***dataArray, int numLayers, int *numMips )
+{
+	scalingStep = std::min(scalingStep, *numMips - 1);
+
+	if ( scalingStep > 0 )
+	{
+		*scaledWidth >>= scalingStep;
+		*scaledHeight >>= scalingStep;
+
+		if( *dataArray && *numMips > scalingStep ) {
+			*dataArray += numLayers * scalingStep;
+			*numMips -= scalingStep;
+		}
+	}
+
+	// Clamp to minimum size.
+	if ( *scaledWidth < 1 )
+	{
+		*scaledWidth = 1;
+	}
+
+	if ( *scaledHeight < 1 )
+	{
+		*scaledHeight = 1;
+	}
+}
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -808,8 +808,7 @@ level 1 has only numLayers/2 layers. There are still numLayers pointers in
 the dataArray for every mip level, the unneeded elements at the end aren't used.
 ===============
 */
-void R_UploadImage( const byte **dataArray, int numLayers, int numMips,
-		    image_t *image )
+void R_UploadImage( const byte **dataArray, int numLayers, int numMips, image_t *image )
 {
 	const byte *data;
 	byte       *scaledBuffer = nullptr;
@@ -1434,8 +1433,7 @@ static void R_ExportTexture( image_t *image )
 R_CreateImage
 ================
 */
-image_t        *R_CreateImage( const char *name, const byte **pic, int width, int height,
-			       int numMips, int bits, filterType_t filterType, wrapType_t wrapType )
+image_t *R_CreateImage( const char *name, const byte **pic, int width, int height, int numMips, const imageParams_t &imageParams )
 {
 	image_t *image;
 
@@ -1451,9 +1449,9 @@ image_t        *R_CreateImage( const char *name, const byte **pic, int width, in
 	image->width = width;
 	image->height = height;
 
-	image->bits = bits;
-	image->filterType = filterType;
-	image->wrapType = wrapType;
+	image->bits = imageParams.bits;
+	image->filterType = imageParams.filterType;
+	image->wrapType = imageParams.wrapType;
 
 	R_UploadImage( pic, 1, numMips, image );
 
@@ -1513,9 +1511,7 @@ image_t *R_CreateGlyph( const char *name, const byte *pic, int width, int height
 R_CreateCubeImage
 ================
 */
-image_t        *R_CreateCubeImage( const char *name,
-                                   const byte *pic[ 6 ],
-                                   int width, int height, int bits, filterType_t filterType, wrapType_t wrapType )
+image_t *R_CreateCubeImage( const char *name, const byte *pic[ 6 ], int width, int height, const imageParams_t &imageParams )
 {
 	image_t *image;
 
@@ -1525,15 +1521,15 @@ image_t        *R_CreateCubeImage( const char *name,
 	{
 		return nullptr;
 	}
-
+	
 	image->type = GL_TEXTURE_CUBE_MAP;
 
 	image->width = width;
 	image->height = height;
 
-	image->bits = bits;
-	image->filterType = filterType;
-	image->wrapType = wrapType;
+	image->bits = imageParams.bits;
+	image->filterType = imageParams.filterType;
+	image->wrapType = imageParams.wrapType;
 
 	R_UploadImage( pic, 6, 1, image );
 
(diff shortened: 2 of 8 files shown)


src/engine/renderer/tr_image.cpp
```diff
@@ -808,8 +808,7 @@ level 1 has only numLayers/2 layers. There are still numLayers pointers in
 the dataArray for every mip level, the unneeded elements at the end aren't used.
 ===============
 */
-void R_UploadImage( const byte **dataArray, int numLayers, int numMips,
-		    image_t *image )
+void R_UploadImage( const byte **dataArray, int numLayers, int numMips, image_t *image, imageParams_t *imageParams )
```
slipher, line 812: A new argument is added but not used anywhere?

illwieckz: Hmmm, probably a leftover of all those will rebases…

illwieckz: After all those rebase, this change belonged to the next commit, so I moved it to the next commit.

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "tr_local.h"
+
+// Comments may include short quotes from other authors.
+
+int R_GetImageCustomScalingStep( const image_t *image, imageParams_t *imageParams )
+{
+	int scalingStep = 0;
+
+	// Perform optional picmip operation.
+	if ( !( image->bits & IF_NOPICMIP ) )
```
slipher, line 44: For readability, return early for the easy case like `if(bits & IF_NOPICMIP) return 0;`

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,129 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+===========================================================================
+*/
+// InternalImage.cpp
+#include "tr_local.h"
+
+// Comments may include short quotes from other authors.
+
```
slipher, line 38: Please add a full explanation in one place of how the rules are expected to work (like in https://github.com/DaemonEngine/Daemon/pull/394#issuecomment-719865589).

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+void R_DownscaleImageDimensions( int scalingStep, int *scaledWidth, int *scaledHeight, const byte ***dataArray, int numLayers, int *numMips )
+{
+	if ( scalingStep > 0 )
+	{
+		*scaledWidth >>= scalingStep;
+		*scaledHeight >>= scalingStep;
+
+		if( *dataArray && *numMips > scalingStep ) {
```
slipher, line 113: What if numMips isn't greater than scalingStep? I think you need to check if there are enough mips first before attempting to downscale.

src/engine/renderer/InternalImage.cpp
illwieckz: done

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+	// Perform optional picmip operation.
+	if ( !( image->bits & IF_NOPICMIP ) )
+	{
+		int imageMinDimension = std::max( -1, r_imageMinDimension->integer );
+		int imageMaxDimension = std::max( 0, r_imageMaxDimension->integer );
+		int shaderMinDimension = 0;
+		int shaderMaxDimension = 0;
+
+		if ( imageParams != nullptr )
```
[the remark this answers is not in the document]
illwieckz: Done, except for functions calling `R_LoadImage` (`R_FindImageFile` and `R_FindCubeImage`) because it would have required to also fix all image loaders (`LoadWEBP`, `LoadPNG`, `LoadTGA`, `LoadJPG`, `LoadDDS`, `LoadCRN`, `LoadKTX`) and many of the function they call, and this PR would then edit a dozen of files…

src/engine/renderer/tr_shader.cpp
```diff
@@ -3961,6 +3973,30 @@ static bool ParseShader( const char *_text )
...
 			continue;
 		}
+		// imageMinDimension enforcement
+		else if ( !Q_stricmp( token, "imageMinDimension" ) )
+		{
+			token = COM_ParseExt2( text, false );
+
+			if ( token[ 0 ] )
+			{
+				shader.imageMinDimension = atof( token );
```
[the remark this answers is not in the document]
illwieckz: done

src/engine/renderer/InternalImage.cpp
illwieckz: I modified the code to test for `( scalingStep > *numMips )` instead of `( scalingStep > 0 )` at the very beginning of that function, is that OK?

slipher: How about like `scalingStep = min(scalingStep, numMips - 1)` to scale it is as much as possible?

src/engine/renderer/tr_shader.cpp
```diff
@@ -3961,6 +3972,30 @@ static bool ParseShader( const char *_text )
...
+			continue;
+		}
+		// imageMaxDimension enforcement
+		else if ( !Q_stricmp( token, "imageMaxDimension" ) )
+		{
+			token = COM_ParseExt2( text, false );
+
+			if ( token[ 0 ] )
+			{
+				shader.imageMaxDimension = atof( token );
```
slipher, line 3994: missed this atof

illwieckz:
What if we drop `r_picmip`? That would make the code simpler, and anyway the purpose of `r_imageMaxDimension` is to produce better results than `r_picmip`.

src/engine/renderer/InternalImage.cpp
illwieckz: Yes, this is better, my suggestion was just completely wrong (it just disabled my code).

illwieckz: done

src/engine/renderer/tr_image.cpp
```diff
@@ -1513,9 +1511,7 @@ image_t *R_CreateGlyph( const char *name, const byte *pic, int width, int height
 R_CreateCubeImage
 ================
 */
-image_t        *R_CreateCubeImage( const char *name,
-                                   const byte *pic[ 6 ],
-                                   int width, int height, int bits, filterType_t filterType, wrapType_t wrapType )
+image_t *R_CreateCubeImage( const char *name, const byte *pic[ 6 ], int width, int height, imageParams_t &imageParams )
```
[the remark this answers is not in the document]
illwieckz: done

src/engine/renderer/tr_shader.cpp
illwieckz: done

illwieckz:
I rewrote the code testing the different cvar value and material keyword, seems more efficient now.

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,209 @@
...
+===============
+*/
+int R_GetImageCustomScalingStep( const image_t *image, const imageParams_t &imageParams )
+{
+	if ( image->bits & IF_NOPICMIP )
+	{
+		return 0;
+	}
+
+	int cvarMinDimension = std::max( -1, r_imageMinDimension->integer );
```
slipher, line 114: I think you should get rid of `r_imageMinDimension`, it is too confusing and has little practical use.

If you want to support the use case ` Very low budget player wanting to reduce all textures to low definition in all case`, you might add a boolean cvar for whether r_imageMaxDimension overrides the shader imageMaxDimension or not. (I don't think this is actually possible in your current version of the code.)

illwieckz: `r_imageMinDimension` has a legit usage, look at those examples:

This is what a “low budget” player would do (except the 2K screen) with `r_imageMaxDimension 64`, all textures would be reduced to `64`, but the buildables that would be reduced to `128` because I added those minimum limits in materials (with `128` eyes are still there):

[image max dimension](https://dl.illwieckz.net/b/daemon/features/image-max-dimension/unvanquished_2021-02-08_033526_000.jpg)

This is what a competitive player may want to do, with `r_imageMaxDimension 1`:

[image max dimension](https://dl.illwieckz.net/b/daemon/features/image-max-dimension/unvanquished_2021-02-08_033551_000.jpg)

But the same competitive player may still want to produce good looking things (especially if he streams his games), he would then use `r_imageMaxDimension 1` with `r_imageMinDimension 512`:

[image max dimension](https://dl.illwieckz.net/b/daemon/features/image-max-dimension/unvanquished_2021-02-08_033621_000.jpg)

Such features are common in arena FPS, like this:

[xonotic competitive casual](https://xonotic.org/static/img/carousel_competitive_casual.jpg)

Usually, this can be done with agressive `r_picmip` value on renderer side, and by using `noPicmip` keyword in materials that must be kept readable. But as we see, `noPicmip` would keep a 4K texture in 4K size in all case, while the image Min/Max Dimension feature (cvar and material keyword) would allow both mappers and players to target the exact value they want.

Surfaces to be used with `imageMinDimension` are typically player and weapon models, panels to be read, grids in which the fact it looks like a grid is required for gameplay, etc. Materials with meaningful glow maps may rely on it too. The `imageMinDimension` material keyword comes as a better replacement for `noPicmip` for the mapper to provide default good values, while the `r_imageMinDimension` enables the player to have control on it.

Note: the `-1` values are mostly there for debugging purposes, we probably don't need to add boolean cvars for disabling those features, and we have no need to put those values in GUI.

slipher: OK so the semantics of a positive r_imageMinDimension are: do nothing if shader imageMinDimension is unset; overrwrite the shader value if it is set. That behavior just seems so random. Why is such a behavior needed? You already have the better version of nopicmip by using the shader parameter alone. Also this behavior means you can get a texture drawn with higher resolution than imageParams.imageMaxDimension if you set r_imageMinDimension to a positive value.

I would expect setting `imageMinDimension 1` in a shader to have the same effect as not setting it - i.e. arbitrarily down-scalable. The strange behavior of r_imageMinDimension means the effect would be different.

illwieckz:
I added a small fix to the code selecting the final size to fix a bug I introduced while reworking that code. I believe I have tested all combinations by now.

src/engine/renderer/InternalImage.cpp
illwieckz: > Also this behavior means you can get a texture drawn with higher resolution than imageParams.imageMaxDimension if you set r_imageMinDimension to a positive value.

I don't really see as an issue. It's expected a material keyword provides the common behavior and the cvar can override it (even to do things that do not make much sense for the general use case).

slipher: The way I see it, the shader parameters define a window of allowable dimensions for the image. If the minimum dimension is omitted it should be equivalent to writing 1. If the maximum dimension is omitted it should be equivalent to writing an arbitrarily large number. So the use case "Competitive player wanting to reduce all textures to one flat color but keep high definition
  on the textures the game developer configured to not be reduced" is philosophically wrong and should not be supported.

In the movie producer use case, the desire is to ignore shader max dimensions. In the very low-budget player use case, the desire is to ignore shader min dimensions. So I propose to get rid of `r_imageMinDimension` (and the -1 special case of `r_imageMaxDimension`) and add instead two booleans, `r_ignoreShaderMinDimension` and `r_ignoreShaderMaxDimension`. The code could be as follows:
```

if ( ( image->bits & IF_NOPICMIP ) )
{
    return 0;
}
int minDimension = imageParams.minDimension > 0 && !r_ignoreShaderMinDimension->integer
                   ? imageParams.minDimension
                   : 1;
int maxDimension = imageParams.maxDimension > 0 && !r_ignoreShaderMaxDimension->integer
                   ? imageParams.maxDimension
                   : std::numeric_limits<int>::max();
if ( r_imageMaxDimension->integer > 0 )
{
    maxDimension = std::min( maxDimension, r_imageMaxDimension->integer );
}
// Consider the larger edge as the "image dimension"
int scaledDimension = std::max( image->width, image->height );
int scalingStep = 0;
// 1st priority: scaledDimension >= minDimension
// 2nd priority: scaledDimension <= maxDimension
// 3rd priority: scalingStep >= r_picMip->integer
// 4th priority: scalingStep as low as possible
while ( scaledDimension > maxDimension || scalingStep < r_picMip->integer )
{
    scaledDimension >>= 1;
    if ( scaledDimension < minDimension ) break;
    ++scalingStep;
}
return scalingStep;
```

illwieckz: > So the use case "Competitive player wanting to reduce all textures to one flat color but keep high definition on the textures the game developer configured to not be reduced" is philosophically wrong and should not be supported.

If you believe this need (which is a serious and real need) is philosophically wrong, maybe that means the cvar is misnamed. Maybe they must be named `r_imageOverrideMinDimension` and `r_imageOverrideMaxDimension`.

Because it's perfectly legit for some non-exotic use case like an arena shooter to flat out textures but wanting at the same time to look good. The “competitive player” may want to flat out textures for two reasons: 1. performance, 2, readability. The `imageMinDimension` material keyword is provided for the mapper to ensure the texture is properly rendered for gameplay or readability purpose (like seeing grates as grates and not as glasses, distinguishes a face, reading sign posts…) so it has an implicit meaning of “this texture suffers from downscaling” and the value means the texture suffers too much from downscaling under this value.

So it's perfectly valid for a player who wants to flat out textures for visibility purpose to prevent textures known to suffer from downscaling (and marked accordingly) to not use the minimum-size known (which can be pretty ugly) but an higher size.

An image not marked with `imageMinDimension` means it may not affect readability or gameplay even when reduced to 1×1, an image marked with `imageMinDimension` means it may affect readability or gameplay if reduced under the given value, it does not mean it looks good at this value.

I first designed this for Unvanquished but because that's an engine feature I want to make sure the option is not only designed for Unvanquished special use case. For example for Unvanquished we may want to forbid the reduction of textures size for visibility purpose, but that can be done by another cvar (like `r_imageRequiredMinDimension` a game can rely on or not, with a `CVAR_CHEAT` flag, independent to the core feature of downscaling control.

What is said to be “philosophically wrong and should not be supported” is the reason why `noPicmip` exists maybe since 20 years, and this is a pretty valid need. What brings this PR is more control on those pictures known to suffer from downscaling, the user would be able to ask for a quality he considers to be high enough for good looking, even if smaller than the default size of the image.

At some point we may even want to add a cvar to set a maximum size for images flagged with nopicmip for more control, like `r_imageOverridePicmipMaxDimension`, but that would mainly be for legacy assets not relying on `imageMinDimension`. This is just to tell about the possibilities (it's needed to take this eventuality in account when designing what is currently proposed) I have not plan on it at this point but I take care of not closing that door. I can see another need for this though: some images are flagged `RSF_NOMIP` without having a materials attached to them (so `imageMinDimension` can't be used), such cvar would control them.

Also it looks OK to me if cvar can make things renders in unexpected way (we can even disable glow mapping which can have negative impact on gameplay or readability), what is important is that assets looks good by default and provided presets use safe options.

slipher: The other competitive player use case is still supported in my version - the one where you put things with a minimum dimension to the minimum, and everything else to 1x1. `r_imageMaxDimension 1` (or `r_picmip 99` would also do the trick).

It is not the case that `noPicmip` is philosophically wrong in my interpretation. `noPicmip` is equivalent to setting the minimum and maximum dimensions of the window to the actual size of the texture.

The thing that I say is philosophically wrong is that a shader with no dimension specifications is treated differently than one with `imageMinDimension 1` and `imageMaxDimension 999999`.

illwieckz: I have tested your code suggestion and I have not found a way to flat out the textures (reduce to `1×1`)  while keeping “textures that are known to suffer from downscaling” in their native resolution (or one arbitrarily high enough).
Also I was able to set `r_imageMaxDimension` to `1` at the same time I've set `r_imageMinDimension` to `1024`.

If we're going to split the cvars for easier understanding, then we also need `r_overrideShaderMinDimension` and `r_overrideShaderMaxDimension`.

illwieckz: I may have an implementation that can fulfill all the needs, with 6 cvars.

- `r_imageMinDimension`
- `r_imageMaxDimension`
- `r_ignoreShaderMinDimension`
- `r_ignoreShaderMaxDimension`
- `r_overrideShaderMinDimension`
- `r_overrideShaderMaxDimension`

Note: I'm not sure `overrideShaderMinDimension` is a good name, maybe `force`, `replace` instead of `override`? I don't know.

I'll push my implementation later.

slipher: > I have tested your code suggestion and I have not found a way to flat out the textures (reduce to `1×1`) while keeping “ètextures that are known to suffer from downscalingè” in their native resolution (or one arbitrarily high enough).

Right, you can't reduce non-dimension-spec textures to 1x1 while keeping others at max resolution in my version. You can have some at 1x1 while reducing others to configured `minDimension`.

> Also I was able to set `r_imageMaxDimension` to `1` at the same time I've set `r_imageMinDimension` to `1024` (the later was not used in any way and that was confusing).

My implementation does not have `r_imageMinDimension`.

slipher:
I don't think your latest patch does exactly what you would want for the use case you describe. For that use case I think what you would really want is to use `imageParams.maxDimension` as the resolution in cases where it is set. The patch just lets you set an arbitrary dimension which may be larger or smaller than the ideal max dimension.

I thought of something that might achieve that while being hopefully not too complicated. Add `r_defaultShaderMaxDimension` which would be used only when the shader param is *not* set, and set this to 1. For me a default value would be way more palatable than an "anti-default" that takes effect only when there is an explicitly configured value!

```
if ( ( image->bits & IF_NOPICMIP ) )
{
    return 0;
}
int minDimension = imageParams.minDimension > 0 && !r_ignoreShaderMinDimension->integer
                   ? imageParams.minDimension
                   : 1;
int shaderMaxDimension = imageParams.maxDimension > 0 ? imageParams.maxDimension : r_defaultShaderMaxDimension->integer;
int maxDimension = shaderMaxDimension > 0 && !r_ignoreShaderMaxDimension->integer
                   ? shaderMaxDimension
                   : std::numeric_limits<int>::max();
if ( r_imageMaxDimension->integer > 0 )
{
    maxDimension = std::min( maxDimension, r_imageMaxDimension->integer );
}
// Consider the larger edge as the "image dimension"
int scaledDimension = std::max( image->width, image->height );
int scalingStep = 0;
// 1st priority: scaledDimension >= minDimension
// 2nd priority: scaledDimension <= maxDimension
// 3rd priority: scalingStep >= r_picMip->integer
// 4th priority: scalingStep as low as possible
while ( scaledDimension > maxDimension || scalingStep < r_picMip->integer )
{
    scaledDimension >>= 1;
    if ( scaledDimension < minDimension ) break;
    ++scalingStep;
}
return scalingStep;
```

src/engine/renderer/InternalImage.cpp
illwieckz: > My implementation does not have `r_imageMinDimension`.

Oh yes, I'm dumb.

Well, with the `r_overrideShaderMinDimension` cvar I added, I can set an arbitrary high value, but not unlock the unknown default resolution for _the one known to suffer from downscaling_ while downscaling the others. People may still set `9999` but that's not clean. That's not a big problem though, but that's not clean.

The problem was about:

- getting a fixed dimension target for downscaling (like, if target is `256`, do nothing with `128` textures, divide once `512` textures, twice `1024` textures, instead of dividing once both `4096` textures and `128`, ending with a mix of textures with `2048` and `64` dimensions,
- making possible for artists to set boundaries on select textures to ensure they are readable in game (or look good enough)
- giving fine control to the user on the engine side for generic textures
- giving fine control to the user on the engine side for textures with boundaries

At least the current implementation of the PR (I still have to edit the comments in code to reflect current code) seems to fulfill the need.

illwieckz:
> I thought of something that might achieve that while being hopefully not too complicated. Add `r_defaultShaderMaxDimension`

Or maybe drop `r_overrideShaderMaxDimension` if your proposition is an alternative to that, I'm not sure, I'm not sure to understand everything.

Anyway, maybe that is enough:

- `r_imageMaxDimension`, `uint`:
  downscale textures to this dimension or to shader `imageMinDimension` if reached,
- `r_ignoreShaderMinDimension`, `bool`:
  ignore shader `imageMinDimension` when downscaling texture to `r_imageMaxDimension`,
- `r_ignoreShaderMaxDimension`, `bool`:
  ignore shader `imageMaxDimension` but `r_imageMaxDimension` still has the last word if set,
- `r_imageHighMinDimension`, `bool`:
  textures with `imageMinDimension` use shader `imageMaxDimension` or image native one if unset when using `r_imageMaxDimension`.
  
I'm open to any better name suggestion for the last one.

slipher:
Instead of "default max dimension", an option following the discussion on IRC, `r_replaceShaderMinDimensionIfPresentWithMaxDimension` for a UI option possibly called "high-resolution models". In addition to specifying the minimum acceptable size, existence of `imageMinDimension` parameter is treated as a hint that high resolution for this texture is more important than for other textures.

```
if ( ( image->bits & IF_NOPICMIP ) )
{
    return 0;
}
int shaderMinDimension = r_ignoreShaderMinDimension->integer ? 0 : imageParams.minDimension;
int shaderMaxDimension = r_ignoreShaderMaxDimension->integer ? 0 : imageParams.maxDimension;
int minDimension;
if ( shaderMinDimension <= 0 )
{
    minDimension = 1;
} else if ( r_replaceShaderMinDimensionIfPresentWithMaxDimension->integer )
{
    minDimension = shaderMaxDimension > 0 ? shaderMaxDimension : std::numeric_limits<int>::max();
}
else
{
    minDimension = shaderMinDimension;
}
int maxDimension = shaderMaxDimension > 0
                   ? shaderMaxDimension
                   : std::numeric_limits<int>::max();
if ( r_imageMaxDimension->integer > 0 )
{
    maxDimension = std::min( maxDimension, r_imageMaxDimension->integer );
}
// Consider the larger edge as the "image dimension"
int scaledDimension = std::max( image->width, image->height );
int scalingStep = 0;
// 1st priority: scaledDimension >= minDimension
// 2nd priority: scaledDimension <= maxDimension
// 3rd priority: scalingStep >= r_picMip->integer
// 4th priority: scalingStep as low as possible
while ( scaledDimension > maxDimension || scalingStep < r_picMip->integer )
{
    scaledDimension >>= 1;
    if ( scaledDimension < minDimension ) break;
    ++scalingStep;
}
return scalingStep;
```

illwieckz:
Thanks! That seems to work. I'll do more tests later (and rewrite comment, clean up the branch…).

illwieckz:
that cvar name… though… 🤭

src/engine/renderer/InternalImage.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+Examples of scenarii:
+
+- r_picMip set 3 with some materials having imageMinDimension set to 128:
+  attempt to reduce three times the images, but for materials having that keyword,
+  stop before the image dimension would be smaller than 128;
+- r_imageMaxDimension set to 64 with some materials having imageMinDimension set to 128:
+  reduce images having dimension greater than 64 to fit 64 dimension, but for materials
+  having that keyword, reduce image to fit 128 dimension;
+- r_imageMaxDimension set to 64 with some materials having imageMinDimension set to 32:
```
slipher, line 70: This example seems wrong - I don't see why anything is reduced to 32.

illwieckz: Hmm, I meant `imageMaxDimension set to 32`:

```
- r_imageMaxDimension set to 64 with some materials having imageMaxDimension set to 32:
  reduce images having dimension greater than 64 to fit 64 dimension, but for materials
  having that keyword, reduce image to fit 32 dimension;
  ```

slipher:
LGTM

illwieckz:
I just rebased on current `0.52.0/sync` branch, this push has no change.

illwieckz:
Yipee!

Issue #334, opened 2020-04-25
relief mapping is not applied on depth computation


illwieckz:
Relief mapping is not applied on depth computation, hence alpha blended textures using depth write (like grids) displays a bit of void. It's possible to workaround this by disabling the depth computation on those textures, but it would be better to fix the bug.

illwieckz:
Another proposed solution (and easier to implement at first) would be to just redo the relief computation on the depth map as well…

illwieckz:
To illustrate the problem, the white garbage is filling the space between the displaced diffuse and the non-displaced depth mask:

[![broken relief alpha mask](https://dl.illwieckz.net/b/daemon/bugs/broken-relief-alpha-mask/unvanquished_2021-02-09_020924_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/broken-relief-alpha-mask/unvanquished_2021-02-09_020924_000.jpg)

VReaperV:
> Another proposed solution (and easier to implement at first) would be to just redo the relief computation on the depth mask as well…

Checking the alpha channel in `lightMapping_fp` would probably be faster. Some extra texture reads, but no extra divergence introduced in depth shader then. It would also avoid writing depth in it, which would otherwise disable early-z optimisations.

slipher:
Instead of disabling relief mapping, we should disable the depth shader for the affected shader. Depth shaders are not essential to render correctly: see #1412.

illwieckz:
@slipher the screenshot in the comment above was taken in `test-pbr` map from https://github.com/UnvanquishedAssets/UnvanquishedTestAssets/releases/tag/v2.1

VReaperV:
I'd prefer a proper fix here rather than disabling things.

illwieckz:
A proper fix is likely slower: what I understand is that it would require to compute relief mapping in the depth shader…

illwieckz:
I mean, if by disabling we mean disablign depth.

If we mean disabling relief mapping, we better not do that yes.

slipher:
Since relief mapping is known to be very expensive, maybe skipping the depth pass could actually be optimal in any case? For an sufficiently expensive shader, the cost of running it twice would outweigh the savings of using less memory bandwidth.

illwieckz:
> Since relief mapping is known to be very expensive, maybe skipping the depth pass could actually be optimal in any case?

I do believe it, yes.

VReaperV:
[...]

PR #426, opened 2021-02-09
tr_shader: workaround relief not applied on alpha mask by disabling relief, ref #334

illwieckz:
To not delay 0.52 more, I propose to just disable relief mapping on grids for now, see #334.


slipher:
By "grids" do you mean like gratings?

LGTM I guess.

illwieckz:
> By "grids" do you mean like gratings?

Yes.

PR #428, opened 2021-02-11
common: misnamed common.framerate.fixed cvar is about frametime

illwieckz:
common: misnamed `common.framerate.fixed` cvar is about frametime


illwieckz:
Also `common.framerate.fixed` is likely to not be about renderer frametime/framerate like `common.framerate.max` so `common.framerate.fixed` may still not be good enough as a name.

slipher:
`common.frametime` is probably not going to be a whole category, so maybe just `common.fixedFrameTime` ?

illwieckz:
Rewritten as suggested.

Issue #71, opened 2018-02-18
3 shaders in engine code requiring game assets


illwieckz:
There are 3 shaders in engine code that are sets from game assets

```
Daemon/src/engine/client/cl_main.cpp:	cls.charSetShader = re.RegisterShader( "gfx/2d/bigchars", RSF_DEFAULT );
Daemon/src/engine/client/cl_main.cpp:	cls.whiteShader = re.RegisterShader( "white", RSF_NOMIP );
Daemon/src/engine/client/cl_main.cpp:	cls.consoleShader = re.RegisterShader( "console", RSF_DEFAULT );
```

The first one "gfx/2d/bigchars" looks to be a fallback if the ttf font is not found. Do we have to keep that? If the ttf found is not found it means both system font and game pack are missing, and if game pack is missing, the fallback will never be found. It looks to fallback to nothing.

https://github.com/DaemonEngine/Daemon/blob/8c1c497498e133ba24d4148603413717488e5a13/src/engine/client/cl_main.cpp#L3054-L3069

The other ones are just shaders for pure white and pure black shaders:

https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/blob/8bd935eca289ab3e55ef21d6122eab884378fd6d/scripts/misc.shader#L106-L114
```c
white
{
	cull none
	{
		map *white
		blendfunc	GL_SRC_ALPHA GL_ONE_MINUS_SRC_ALPHA
		rgbgen vertex
	}
}
```

https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/blob/8bd935eca289ab3e55ef21d6122eab884378fd6d/scripts/misc.shader#L38-L45
```c
console
{
	nopicmip
	nomipmaps
	{
		map gfx/colors/black
	}
}
```

The `white` one only relies on pure text, it can be embedded. The `console` one (black one) relies on a black texutres, it can probably be generated at run time or we can probably embed a 1×1px xpm or somethin like that (it's currently a pure black 8×8 jpg)…

Note that both `white` and `console` shaders are redefined by the game code itself and game packages so they are basically fallback for when game code does not define them and when game packages are not there, so the purpose of fallback is entirely defeated: the fallback for missing game requires game packs (!!!).

illwieckz:
Well, the engine also requires:

```
lights/defaultPointLight
lights/defaultProjectedLight
lights/defaultDynamicLight
```

and they rely on some pixmaps.

[...]

PR #432, opened 2021-02-23
Generate GLSL includes with pure CMake

### LINKED ISSUE #122: Cross-compilation is broken
Hi guys,

I might have found a bug in the new `include-bin` generator process for glsl when cross-compiling:
basically, since the tool is cross-compiled you cannot run it on the host during the build process and that makes CMake complaining about missing binary and it doesn't generate the glsl include files.

From a quick search on the internet it seems there is not an easy solutinon to this (they suggest a 2-pass build).

slipher:
For hex-dumping the GLSL source files into character arrays, use a CMake
script instead of a C++ program. This fixes #122 (include-bin can't be
used with cross-compilation).

Also, put the GLSL generated headers directory in the build directory
instead of the source tree. This makes the build work with the source on
a read-only filesystem.


illwieckz:
This is welcomed, I wanted this! At some point we will be able to rename `glsl_source` to just `glsl` in source tree.

Also at some point we may reuse similar mechanisms to embed more engine-related assets in binary to fix #71.

illwieckz:
Why not doing something like that instead:

```cmake
set(EMBEDDED_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/embedded)
outpath ${EMBEDDED_INCLUDE_DIR}/glsl-${filename_no_ext}.h
```

With files being produced as `embedded/glsl-skybox_vp.h`.

At the time we start embedding more assets, they would be stored as `embedded/gfx-2d-bigchars.h` (for `gfx/2d/bigchars.png`).

illwieckz (approved):
LGTM

Issue #423, opened 2021-02-06
Crash when loading a custom build of plat23 map


illwieckz:
File to reproduce the bug: [map-plat23_0~crash.dpk](https://dl.illwieckz.net/b/daemon/bugs/plat23-crash/map-plat23_0~crash.dpk)

Note: there is no lighting computed at all (neither lightmap, neither lightgrid), neither visibility step is done but this should not be a problem.

I'm running `0.52.0/sync` game and engine, here is the `gdb` backtrace:

```gdb
Warn: light grid mismatch, default light grid used
 
Thread 1 "daemon" received signal SIGSEGV, Segmentation fault.
0x00005555557f49b1 in R_CreateWorldVBO () at …/Daemon/src/engine/renderer/tr_bsp.cpp:2903
2903				if ( surf1->viewCount != -1 )

Thread 34 (Thread 0x7fff5ffff700 (LWP 670800)):
#0  __libc_recvmsg (flags=0, msg=0x7fff5fff4470, fd=35) at ../sysdeps/unix/sysv/linux/recvmsg.c:28
#1  __libc_recvmsg (fd=35, msg=0x7fff5fff4470, flags=0) at ../sysdeps/unix/sysv/linux/recvmsg.c:25
#2  0x00007fff5de7854a in NaClReceiveDatagram (handle=35, message=0x7fff5fff4570, flags=0) at /home/illwieckz/dev/buildme-unvanquished/Daemon/libs/nacl/native_client/src/shared/imc/linux/nacl_imc.cc:168
#3  0x00007fff5dce5429 in IPC::InternalRecvMsg (handle=35, reader=...) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/common/IPC/Primitives.cpp:380
#4  0x00007fff5dce59ca in IPC::Socket::RecvMsg (this=0x7fff5f2a1e40 <VM::rootChannel>) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/common/IPC/Primitives.cpp:464
#5  0x00007fff5dc5a12c in IPC::Channel::RecvMsg (this=0x7fff5f2a1e40 <VM::rootChannel>) at /home/illwieckz/dev/buildme-unvanquished/Daemon/src/common/IPC/Channel.h:111
[...]

PR #437, opened 2021-03-29
Fix crash with invalid light grid data size

slipher:
Fixes the crash in #423.


illwieckz:
I confirm it fixes the crash reported in #423.

PR #438, opened 2021-04-05
Enable ASLR on Mac

slipher:
See issue #439


illwieckz:
I haven't tried to run it, but otherwise I see no issue while reading it.

PR #441, opened 2021-04-08
make r_fastsky cvar an archive one

illwieckz:
- make `r_fastsky` cvar an archive one.


slipher:
I don't think `r_imageMaxDimension` should be limited by default.

illwieckz:
Hmm, yes, the fact this is an engine option and not a game option does not make it a good candidate for a default. :-/

illwieckz:
I removed the commit about `r_imageMaxDimension`, what about the other one?

Anyway that brings a question: how can a game (not the engine) can set specific defaults that may not be the engine ones on first start…

slipher:
LGTM

> Anyway that brings a question: how can a game (not the engine) can set specific defaults that may not be the engine ones on first start…
I guess you could have a cvar like how `set cg_welcome 1` is used to prevent the name dialog from being shown again after the first time.

illwieckz:
> I guess you could have a cvar like how `set cg_welcome 1` is used to prevent the name dialog from being shown again after the first time.

I was thinking about this as well, it would also be interesting to look at the mechanism that loads default binds.

PR #443, opened 2021-04-11
NaCl build fix

slipher:
I guess there is probably some issue like if `CMAKE_CXX_FLAGS_RELEASE` is already defined at the beginning then you can read and set it in a case-insensitive manner, but if it does not exist yet then the variable gets created with the wrong case like `CMAKE_CXX_FLAGS_Release` and then CMake does not consider it when building.

I have no idea why different arguments to cmake would change whether this occurs.


illwieckz:
It built! 👍

Issue #449, opened 2021-04-18
Improve support to cvar namespaces


peoro:
I'd like to group cvars into "namespaces" more often. For instance, instead of using cvar called `g_doWarmup` and `g_warmup` I'd prefer to name them something like `game.warmup.enabled` and `game.warmup.time`. Daemon already uses a similar style for some cvars, like the logging ones, e.g. `logs.logLevel.sgame.default`.

It would be awesome if Daemon had better support for similar cvars. For instance, when tab-completing, it could list and autocomplete the namespaces without expanding them.

Would such a feature be welcome? We could even use a cvar to control this behavior, if people have different preferences.

illwieckz:
Yes, I would like to see it as well. I also prefer the dot naming.

slipher:
Fixed by #723.

PR #450, opened 2021-04-19
Namespace aware autocompletion. Resolves DaemonEngine/Daemon#449

peoro:
Some example of how autocompletion behaves with this patch (`↹` is shown where I pressed tab):

```
]se↹
-> /se
   server.{x4}
   set                      sets the value of a cvar
   seta                     sets the value of a cvar and marks the cvar as archived
   sets                     sets the value of a cvar
   setu                     sets the value of a cvar
]ser↹
-> /server.
   server.challenge.{x3}
   server.private           cvar - "3" - server private - Controls how much the server advertises: 0 - Advertise everything, 1 - Don't advertise but reply to status queries, 2 - Don't reply to status queries but accept connections, 3 - Only accept LAN connections.
]server.c↹
-> /server.challenge.
   server.challenge.count   cvar - "1024" - uint - Maximum number of active challenges kept in memory - between 1 and 115292150460684697
   server.challenge.length  cvar - "8" - uint - Length in bytes of the challenge data (The hexadecimal representation will be twice as long) - between 1 and 9223372036854775807
   server.challenge.timeout cvar - "5" - int - Timeout (in seconds) of a challenge - between 1 and 2147483647
```

```
]a↹
-> /a
   alias     creates or view an alias
   arg_all   ""
   arg_count "0"
]/set a.a 1
]a↹
-> /a
   a.a       "1"
   alias     creates or view an alias
   arg_all   ""
   arg_count "0"
]/set a.b 1
]a↹
-> /a
   a.{x2}
   alias     creates or view an alias
   arg_all   ""
   arg_count "0"
]a.↹
-> /a.
   a.a "1"
   a.b "1"
```

```
]log↹
-> /logs.
   logs.logFile.{x4}
   logs.logLevel.{x8}
   logs.suppression.{x4}
```

```
]/set b.b.b.b.0
]/set b.b.b.b.1
]b.↹
-> /b.b.b.b.
   b.b.b.b.0 ""
   b.b.b.b.1 ""
```

### DIFF
--- a/src/engine/framework/ConsoleField.cpp
+++ b/src/engine/framework/ConsoleField.cpp
@@ -129,9 +129,57 @@ namespace Console {
         //Print the matches if it is ambiguous
         if (candidates.size() >= 2) {
             Log::CommandInteractionMessage(Str::Format("^3-> ^*%s", Str::UTF32To8(GetText())));
-            for (const auto& candidate : candidates) {
+            
+            auto showCandidate = [&]( Cmd::CompletionItem candidate ) {
                 std::string filler(maxCandidateLength - candidate.first.length(), ' ');
                 Log::CommandInteractionMessage(Str::Format("   %s%s %s", candidate.first, filler, candidate.second));
+            };
+            
+            // we only group candidates by namespace for commands, not for arguments
+            if( argNum > 1 ) {
+                // print everything
+                for (const auto& candidate : candidates) {
+                    showCandidate( candidate );
+                }
+            }
+            else {
+                using CompletionIterator = Cmd::CompletionResult::iterator; 
+                
+                // for every candidate (`it`)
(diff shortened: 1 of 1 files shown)


slipher:
This logic should probably go in the `CompleteCommandNames` function

slipher:
What if we treated `_` the same way as `.`? Then it would also work with cvars having Quake naming conventions such as `r_fullscreen`.

illwieckz:
> What if we treated `_` the same way as `.`? Then it would also work with cvars having Quake naming conventions such as `r_fullscreen`.

That sounds good!

bmorel:
Are the 2 remarks here blocking? Talking about those:

> This logic should probably go in the CompleteCommandNames function

> What if we treated _ the same way as .? Then it would also work with cvars having Quake naming conventions such as r_fullscreen.

slipher:
Done in #723.

PR #451, opened 2021-04-20
restore mover content flag like ioq3/tremulous, avoid collision with nobuild content flag

illwieckz:
Restore `mover` content flag to the value used in Tremulous, was probably mistakenly set to Wolf:ET one when merging Wolf:ET code.

This flag seems to be set by the engine to be used by the game, there is no mention of it either in map compiler source code, neither in the `custinfoparms.txt` the map compiler reads, so the `.bsp` file is not expected to carry it.

As long as you recompile the game code with this revision of the engine, there is no regression. Even movers in Wolf:ET maps would work as expected if Wolf:ET code was built and run with this engine.

Restoring this flag to the Quake3/Tremulous value is required to avoid collision with the `nobuild` content flag that is (currently) using the same value. The `nobuild` content flag is set by the map compiler (defined in [`custinfoparms.txt`](https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/blob/master/scripts/custinfoparms.txt)) and written in stone at map compilation, carried by the `.bsp` file. Keeping compatibility with legacy maps without source or without requiring to recompile the `.map` or use `esquirel` to rewrite the flags in the `.bsp` requires to keep the `nobuild` value as it was set in Tremulous back in the day.

Not restoring this flag means:

- the game code mistakenly considers every `nobuild` brush is a `mover`;
- the spectator can fly through `nobuild` brushes like if it was a mover, even if clipping is set for this brush (including `playerclip`);
- we cannot provide a `common/outside` brush to mappers to both clip player, let weapons projectile pass, and forbid building, requiring mappers to do two brushes instead of one anytime they want to delimit the playable, or relying on undocumented behaviors by attempting to merge into one brush the `surfaceparm` or `contentparm` values from the various surfaces of the brush;
- we cannot implement a renderer feature using grid lighting on movers to adjust the lighting while they move (relatively to the ambient lighting of the actual place) instead of light mapping to prevent train or doors to move with their hardcoded shadow from the static place they were compiled, shadow that is wrong as soon as the mover starts to move, otherwise some `nobuild` surfaces neighbouring buildables surfaces (even from the same plane and expected to just be part of the same surface) would be lit like movers but not like the world, with shadow discontinuation and light visually mismatching among triangles of the same surface.

Not restoring this flag for 0.52 means we would have to skip entirely the 0.52 lifecycle and wait for 0.53 release date to start implementing features requiring that fix (easier support for mappers, the mover lightgrid lighting…).

I have tested this patch with maps with movers like spacetracks (trains, doors, elevators), uss tremor (doors, complex lift), etc. without noticing any issue. Anyway getting a regression would be surprised because this is precisely fixing a regression when compared to tremulous. Getting it for RC4 sounds really important to me.

### DIFF
--- a/src/engine/qcommon/SurfaceFlags.h
+++ b/src/engine/qcommon/SurfaceFlags.h
@@ -57,7 +57,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // undefined                        BIT( 11 ) // 0x800
 // undefined                        BIT( 12 ) // 0x1000     // Unvanquished stores CONTENTS_NOALIENBUILD there
 // undefined                        BIT( 13 ) // 0x2000     // Unvanquished stores CONTENTS_NOHUMANBUILD there
-#define CONTENTS_MOVER              BIT( 14 ) // 0x4000     // Unvanquished stores CONTENTS_NOBUILD there
+#define CONTENTS_MOVER              BIT( 14 ) // 0x4000     // Wolf:ET stores CONTENTS_MOVER there; Unvanquished stores CONTENTS_NOBUILD there
 #define CONTENTS_AREAPORTAL         BIT( 15 ) // 0x8000
 #define CONTENTS_PLAYERCLIP         BIT( 16 ) // 0x10000
 #define CONTENTS_MONSTERCLIP        BIT( 17 ) // 0x20000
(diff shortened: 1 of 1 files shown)


illwieckz:
This PR is expected to fix https://github.com/Unvanquished/Unvanquished/issues/1373 and to make possible to merge more commits from https://github.com/UnvanquishedAssets/tex-common_src.dpkdir/pull/3 especially the `common/outside` material.

A similar fix was already done for `nobuild` in the past: [Unvanquished/Unvanquished@`56b629d`](https://github.com/Unvanquished/Unvanquished/commit/56b629dbf3fe0e89d00a4baa8dd2a3ca2dcd6ea3?branch=56b629dbf3fe0e89d00a4baa8dd2a3ca2dcd6ea3#diff-1912114c75224fbbb87906e450445975329dd4f15b3f577b11f849c697acbdb6L73-L76) where various `nobuild` flags were colliding with `nodamage`, `slick` and `sky`.

illwieckz:
Reference: https://github.com/darklegion/tremulous/blob/c862a5340c8de44dcc7abaff170c20c04f9340e8/src/qcommon/surfaceflags.h#L51

```c
#define CONTENTS_MOVER			0x800000
```

PR #453, opened 2021-04-20
renderer: tiled renderer is r_dynamicLight > 0, see be7229a

### LINKED ISSUE #386: Dynamic lights not removed on GeForce GTX 1070
Whenever I fire one of the weapons with a muzzle flash (e.g. rifle, lasgun, mdriver), the light effect persists indefinitely, until the next time the weapon is fired, upon which it moves to the current location. There is the same issue with the lights produced on the impact of the pulse rifle and lucifer cannon projectiles, which similarly don't disappear until the next projectile hits.

The bug occurs alike with the 0.51 release and a 0.52 branch build. I haven't seen it before, so there may be a recently introduced driver bug.

The bug does not occur if I use the Intel integrated driver.

The bug does not occur with `r_dynamicLights 0`.

image
image
image

### System info
Windows 10

```
]/gfxinfo
GL_VENDOR: NVIDIA Corporation
GL_RENDERER: GeForce GTX 1070 with Max-Q Design/PCIe/SSE2
GL_VERSION: 4.6.0 NVIDIA 452.06
GL_SHADING_LANGUAGE_VERSION: 4.60 NVIDIA
^x0f0Using OpenGL 3.x context
```

> Driver Date: 8/12/2020
> Driver Version: 27.21.14.5206

illwieckz:
The forwared renderer is deprecated and now uses -1 value.
The tiled renderer is meant to use 1 value one day, but both 1 and 2
can be usable for tiled renderer in 0.52.0, in order to keep
compatibility with 0.51.0 config files when testing releases candidates.

Those test rewrite were missing in be7229a.

See https://github.com/DaemonEngine/Daemon/issues/386#issuecomment-823196130

The issue was spotted thanks to https://github.com/DaemonEngine/Daemon/issues/386#issuecomment-823196130

I don't expect this to fix #386 if people experiencing problems with Nvidia used `r_dynamicLight` `2`.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -688,7 +688,7 @@ static void Render_generic( int stage )
 		GL_BindToTMU( 1, tr.currentDepthImage );
 	}
 
-	if ( r_dynamicLight->integer == 2 )
+	if ( r_dynamicLight->integer > 0 )
 	{
 		GL_BindToTMU( 8, tr.lighttileRenderImage );
 	}
@@ -907,7 +907,7 @@ static void Render_lightMapping( int stage )
 	}
 
 	// bind u_LightTiles
-	if ( r_dynamicLight->integer == 2 )
+	if ( r_dynamicLight->integer > 0 )
 	{
 		GL_BindToTMU( BIND_LIGHTTILES, tr.lighttileRenderImage );
 	}


illwieckz:
I introduced this in fd8c6f56807fae11b755b3890cb6187c559da543 (not binding some tiled light when tiled renderer is not used) because I probably forgot I merged be7229a2489971e14a3a636d962117012449478a (making forward renderer using -1) first, or maybe that was two PR that was concurrent and were both right before both were merged.

PR #454, opened 2021-04-21
Fix dynamic light bug

### LINKED ISSUE #386: Dynamic lights not removed on GeForce GTX 1070

slipher:
Fixes #386.
The off-by-one commit didn't do anything; it's the one with u_numLights.

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -209,7 +209,7 @@ void computeDLights( vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse, vec4 mater
     for( int i = 0; i < lightsPerLayer; i++ ) {
       int idx = numLayers * nextIdx( idxs ) + layer;
 
-      if( idx > u_numLights )
+      if( idx >= u_numLights )
       {
 #if defined(r_showLightTiles)
         if (numLights > 0.0)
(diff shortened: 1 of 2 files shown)


illwieckz:
I don't have hardware right now to reproduce the bug and test the fix, and I don't really know what that line does. Though the patch is very simple and I noticed no regression on radeon and intel on linux, so if it fixes nvidia, let's go!

PR #459, opened 2021-04-26
Renderer: never cull 1st-person IQM models

slipher:
To fix https://github.com/Unvanquished/Unvanquished/issues/1393.


illwieckz:
That's a clever trick! And because the first person weapon is always expected to be rendered on screen, the hack seems to be harmless…

It looks like that fix will see the pretty lights of the 0.52 release!

PR #462, opened 2021-04-26
[windows] Include OS / DPI Aware manifests

dimhotepus:
Include manifests for:
* OS support info to signal Windows our app supports it as is without quirks.
* DPI aware v.2 (with v.1 fallback) scaling to disallow Windows to scale our window.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -684,6 +684,17 @@ function(AddApplication)
     set_property(TARGET ${A_Target} APPEND PROPERTY COMPILE_DEFINITIONS ${A_Definitions})
     set_target_properties(${A_Target} PROPERTIES OUTPUT_NAME "${A_ExecutableName}" PREFIX "" FOLDER "engine")
 
+    # Append Windows specific manifests.
+    if (WIN32)
+        set(WINDOWS_MANIFESTS_DIR ${ENGINE_DIR}/sys/manifests)
+        target_sources(${A_Target} PRIVATE
+            # App is DPI aware, so Windows doesn't try to scale UI.
+            ${WINDOWS_MANIFESTS_DIR}/dpi-aware.manifest
+            # App supports Windows Vista+, no quirks required.
+            ${WINDOWS_MANIFESTS_DIR}/supported-os.manifest
+        )
+    endif()
+
     message(STATUS ${A_Target})
     ADD_PRECOMPILED_HEADER(${A_Target})
 endfunction()
--- a/src/engine/sys/manifests/supported-os.manifest
+++ b/src/engine/sys/manifests/supported-os.manifest
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+    <application>
+      <!-- See https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#supportedos -->
+      <!-- Application supports Windows Vista and Windows Server 2008
+           functionality. -->
+      <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}"/>
+      <!-- Application supports Windows 7 and Windows Server 2008 R2
+           functionality. -->
+      <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}"/>
+      <!-- Application supports Windows 8 and Windows Server 2012
+           functionality. -->
+      <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
+      <!-- Application supports Windows 8.1 and Windows Server 2012 R2
+           functionality. -->
+      <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}"/>
+      <!-- Application supports Windows Windows 10, Windows Server 2016 and
+           Windows Server 2019 functionality. -->
+      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
+    </application>
+  </compatibility>
+</assembly>
--- a/src/engine/framework/System.cpp
+++ b/src/engine/framework/System.cpp
@@ -550,15 +550,6 @@ static void Init(int argc, char** argv)
 #ifdef _WIN32
 	// Don't let SDL set the timer resolution. We do that manually in Sys::Sleep.
 	SDL_SetHint(SDL_HINT_TIMER_RESOLUTION, "0");
-
-	// Mark the process as DPI-aware on Vista and above, ignore any errors
-	std::string errorString;
-	Sys::DynamicLib user32 = Sys::DynamicLib::Open("user32.dll", errorString);
-	if (user32) {
-		auto pSetProcessDPIAware = user32.LoadSym<BOOL WINAPI()>("SetProcessDPIAware", errorString);
-		if (pSetProcessDPIAware)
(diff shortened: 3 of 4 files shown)


CMakeLists.txt
```diff
@@ -684,6 +693,34 @@ function(AddApplication)
...
 
+    # Append Windows specific manifests.
+    if (WIN32)
+      set(WINDOWS_MANIFESTS_DIR ${ENGINE_DIR}/sys/manifests)
+      target_sources(${A_Target} PRIVATE
+          # App is DPI aware, so Windows doesn't try to scale UI.
+          ${WINDOWS_MANIFESTS_DIR}/dpi-aware.manifest
+          # App supports Windows Vista+, no quirks required.
+          ${WINDOWS_MANIFESTS_DIR}/supported-os.manifest
+          # Use visual styles for Windows UI controls (message boxes, buttons, etc.).
```
slipher, line 704: The UI is all drawn with OpenGL, so this doesn't seem relevant. The only native control we have is the message box when the engine encounters a fatal error.

CMakeLists.txt
```diff
@@ -684,6 +693,34 @@ function(AddApplication)
     set_property(TARGET ${A_Target} APPEND PROPERTY COMPILE_DEFINITIONS ${A_Definitions})
     set_target_properties(${A_Target} PROPERTIES OUTPUT_NAME "${A_ExecutableName}" PREFIX "" FOLDER "engine")
 
+    # Append Windows specific manifests.
+    if (WIN32)
+      set(WINDOWS_MANIFESTS_DIR ${ENGINE_DIR}/sys/manifests)
+      target_sources(${A_Target} PRIVATE
+          # App is DPI aware, so Windows doesn't try to scale UI.
+          ${WINDOWS_MANIFESTS_DIR}/dpi-aware.manifest
```
slipher, line 701: So we could remove this part: https://github.com/DaemonEngine/Daemon/blob/63ef00aa7b9bbf388a015cd9ffffc6d09a6429f5/src/engine/framework/System.cpp#L554

src/engine/sys/manifests/supported-os.manifest
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+    <application>
+      <!-- See https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#supportedos -->
+      <!-- Application supports Windows Vista and Windows Server 2008
+           functionality. -->
+      <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}"/>
```
slipher, line 8: What are the practical consequences of these?

CMakeLists.txt
dimhotepus: Yep, but still message box can be styled as well.

Since this doesn't break anything, suggest to integrate manifest.

slipher: The visual styles one in particular seems to significantly increase the complexity of the build system, so I don't think it is worth it.

dimhotepus: Good point. Removed.

CMakeLists.txt
dimhotepus: Done, thanks!

src/engine/sys/manifests/supported-os.manifest
dimhotepus: If `supportedOS ` missed, Windows treats app as not OS ready.
See [Windows and Windows Server compatibility cookbook: Windows 8, Windows 8.1, and Windows Server 2012](https://www.microsoft.com/en-us/download/details.aspx?id=27416)

Manifestation
Apps without a compatibility section in their manifest will have Windows Vista
behavior by default on Windows 7 and Windows 8 and future Windows versions. Be
aware that Windows XP and Windows Vista ignore this manifest section and it
has no impact on them.

These Windows components provide divergent behavior based on the compatibility
section:
Remote procedure call (RPC) default thread pool
⦁ Windows 8 and Windows 7: To improve scalability and reduce thread
 counts, RPC switched to the NT thread pool (default pool). For Windows Vista,
 RPC used a private thread pool:
⦁ For binaries compiled for Windows 7 and later versions of Windows, the
  default pool is used.
⦁ If `I_RpcMgmtEnableDedicatedThreadPool` is called before any RPC API is
  called, the private thread pool is used (Vista behavior).
⦁ If `I_RpcMgmtEnableDedicatedThreadPool` is called after an RPC call, the 
  default pool is used, `I_RpcMgmtEnableDedicatedThreadPool` returns the 
  error `1764`, and the requested operation is not supported. 
⦁ Windows Vista (**default**): For binaries compiled for Windows Vista and
 earlier versions of Windows, the private pool is used.

DirectDraw lock 
⦁ Windows 8 and Windows 7: Apps manifested for Windows 7 and later versions of
  the operating system cannot call Lock API in DDRAW to lock the primary
  desktop video buffer; doing so will result in an error, and a `NULL` pointer
  for the primary is returned.
  This behavior is enforced even if Desktop Window Manager Composition is not
  turned on. Apps with compatibility declared for Windows 7 and later must not
  lock the primary video buffer to render. 
⦁ Windows Vista (**default**): Apps can acquire a lock on the primary video buffer,
  as legacy apps depend on this behavior; running the app turns off Desktop
  Window Manager. 

DirectDraw bit-block transfer (bitblt) to primary without clipping window 
⦁ Windows 8 and Windows 7: Apps manifested for Windows 7 and later versions of
  Windows are prevented from performing a bitblt to the primary Desktop video
  buffer without a clipping window; doing so results in an error and the bitblt
  area will not be rendered.
  Windows enforces this behavior even if you do not turn on Desktop Window
  Manager Composition. Apps with compatibility declared for Windows 7 and later
  must perform a bitblt to a clipping window. 
⦁ Windows Vista (**default**): Apps must be able to perform a bitblt to the primary
  without a clipping window, as legacy apps depend on this behavior; running
  this app turns off the Desktop Window Manager. 

`GetOverlappedResult` API 
⦁ Windows 8 and Windows 7: Resolves a race condition where a multithreaded app
  using `GetOverlappedResult` can return without resetting the event in the
  overlapped structure, causing the next call to this function to return
  prematurely.
⦁ Windows Vista (**default**): Provides the behavior with the **race condition** that
  apps might have a dependency on.
  Apps that must avoid this race prior to the Windows 7 behavior should wait
  on the overlapped event and, when signaled, call `GetOverlappedResult` with
  `bWait == FALSE`. 

Shell themes status in high contrast mode 
⦁ Windows 8: Returns the real theming status for when in high contrast mode.
⦁ Windows 7: Returns theming as unavailable when in high contrast mode because
  DWM is still on.
⦁ Windows Vista (**default**): Returns theming as unavailable when in high contrast
  mode because DWM is still on.

Shell `iPersistFile::Save` method
⦁ Windows 8: `CShellLink::Save` now determines if the `IPersistFile` handler is
  called with a relative path argument and fails the call if it is.
  The public documentation describing this behavior indicates that path argument
  has to be an absolute path: http://go.microsoft.com/fwlink/?LinkId=228369
⦁ Windows 7 and earlier (**default**): `CShellLink::Save` does not determine if the
  `iPersistFile` handler sends a relative path check and allows apps to continue
  working with absolute or relative paths.

Program Compatibility Assistant (PCA) 
⦁ Windows 8: Apps with the compatibility section do not get the PCA mitigation.
⦁ Windows 7: Apps with the compatibility section are tracked for potential
  compatibility issues for Windows 8 changes (described in this document).
⦁ Windows Vista (**default**): **Apps that fail to install properly or crash during
 runtime under some specific circumstances get the PCA mitigation.**

slipher:
Looks good, will merge later.

PR #463, opened 2021-04-26
[server] Ensure pak size fit MAX_INT instead of silent truncate

dimhotepus:
`downloadSize` is `int`, but `FS::offset_t` is `int64_t`.

So pak of size > `INT_MAX` is silently truncated to `int` one and may contain bad data.

`MSG_WriteLong` accepts `int`, so can't just change `downloadSize` to `int64`. Also client contract uses `int`.


slipher:
Looks OK, will merge when 0.52 release is finialized.

PR #464, opened 2021-04-26
CMake build warnings fixes

dimhotepus:
Warning-free during CMake project generation.


slipher:
Thanks for fixing the PARENT_SCOPE thing.

I checked whether CMake versions can change under us as mentioned [here](https://github.com/DaemonEngine/Daemon/pull/457#issuecomment-826566576):

- Appveyor has moving CMake versions. For example the Visual Studio 2017 image currently has CMake 3.16.2, while in January 2019 it had 3.12.2. 

- Azure Pipelines has moving [software versions](https://github.com/actions/virtual-environments/commits/main/images/macos/macos-10.15-Readme.md).

- With Travis we might possibly be safe from CMake upgrades, as long as we use the previous stable Ubuntu version. (Currently, we are using an even older one.)

So for Appveyor and Azure Pipelines, I would not want to have '-Werror' CMake warnings, because most new warnings are of the "policy warning" type which is just a notification that some default will change in the future. It's annoying for the build to break out of nowhere when there is not a real problem.

Also, please edit the commit message to reflect the fact that new flags are added (not only reordered).

dimhotepus:
@slipher Done, thanks!

PR #465, opened 2021-04-26
[renderer] Static array bounds checking for frustum calculators

dimhotepus:
In C++ array passed to function as pointer, so no bound checks
performed. For example:

int a[3] = {1, 2, 3};

void accept(int v[4]) {
   v[3] = 1;
}

accept(a);   // Compiles ok, UB in runtime.

To statically check bounds array should be passed as reference:

int a[3] = {1, 2, 3};

void accept(int (&v)[4]) {
   v[3] = 1;
}

accept(a);   // Doesn't compile.

This sample still has problems, as arrays of greater than reference
array length can't be passed in:

int a[5] = {1, 2, 3, 4, 5};

void accept(int (&v)[4]) {
   v[3] = 1;
}

accept(a);   // Doesn't compile, semantically valid.

To allow passing of arrays with length >= acceptable one and do
compile time bounds checks SFINAE can be used:

int three[]{1, 2, 3};
int four[]{1, 2, 3, 4};
int five[]{1, 2, 3, 4, 5};

template<size_t length>
std::enable_if_t<(length >= 4)> accept(int (&v)[length]) {
   v[3] = 1;
}

accept(three);   // Doesn't compile
accept(four);     // Ok
accept(five);      // Ok

Apply ^ in this commit to check frustum / corners array bounds
in compile time.

### DIFF
--- a/src/engine/renderer/tr_main.cpp
+++ b/src/engine/renderer/tr_main.cpp
@@ -1086,7 +1086,7 @@ void R_SetupFrustum2( frustum_t frustum, const matrix_t mvp )
 
 // *INDENT-ON*
 
-void R_CalcFrustumNearCorners( const vec4_t frustum[ FRUSTUM_PLANES ], vec3_t corners[ 4 ] )
+void R_CalcFrustumNearCornersUnsafe( const vec4_t frustum[ FRUSTUM_NEAR + 1 ], vec3_t (&corners)[ 4 ] )
 {
 	PlanesGetIntersectionPoint( frustum[ FRUSTUM_LEFT ], frustum[ FRUSTUM_TOP ], frustum[ FRUSTUM_NEAR ], corners[ 0 ] );
 	PlanesGetIntersectionPoint( frustum[ FRUSTUM_RIGHT ], frustum[ FRUSTUM_TOP ], frustum[ FRUSTUM_NEAR ], corners[ 1 ] );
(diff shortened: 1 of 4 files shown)


necessarily-equal:
I'm not approving because I can't make 100% sure everything works like before, but it looks fine otherwise. :+1:

src/engine/renderer/tr_main.cpp
```diff
@@ -1086,15 +1086,15 @@ void R_SetupFrustum2( frustum_t frustum, const matrix_t mvp )
...
-void R_CalcFrustumNearCorners( const vec4_t frustum[ FRUSTUM_PLANES ], vec3_t corners[ 4 ] )
+void R_CalcFrustumNearCornersUnsafe( const vec4_t frustum[ FRUSTUM_NEAR + 1 ], vec3_t (&corners)[ 4 ] )
 {
 	PlanesGetIntersectionPoint( frustum[ FRUSTUM_LEFT ], frustum[ FRUSTUM_TOP ], frustum[ FRUSTUM_NEAR ], corners[ 0 ] );
 	PlanesGetIntersectionPoint( frustum[ FRUSTUM_RIGHT ], frustum[ FRUSTUM_TOP ], frustum[ FRUSTUM_NEAR ], corners[ 1 ] );
 	PlanesGetIntersectionPoint( frustum[ FRUSTUM_RIGHT ], frustum[ FRUSTUM_BOTTOM ], frustum[ FRUSTUM_NEAR ], corners[ 2 ] );
 	PlanesGetIntersectionPoint( frustum[ FRUSTUM_LEFT ], frustum[ FRUSTUM_BOTTOM ], frustum[ FRUSTUM_NEAR ], corners[ 3 ] );
 }
 
-void R_CalcFrustumFarCorners( const vec4_t frustum[ FRUSTUM_PLANES ], vec3_t corners[ 4 ] )
```
illwieckz, line 1097: @dimhotepus is that `+1` related to this comment?

https://github.com/DaemonEngine/Daemon/blob/a4fbf1829d6a84e05abfde546a57bc454c501bb0/src/engine/renderer/tr_local.h#L243-L257

dimhotepus: We read from `frustum[ FRUSTUM_NEAR ]` inside `R_CalcFrustumNearCornersUnsafe`, so passed `frustum` array should be at least of `FRUSTUM_NEAR (4) + 1 == 5` size.

PR #466, opened 2021-04-26
[client,common,qcommon] Mark allocators with allocator / malloc like

dimhotepus:
Allows compiler to infer pointer usage and optimize better.


illwieckz:
Looks to be fine.

PR #468, opened 2021-04-28
reduce number of warnings

bmorel:


src/engine/renderer/tr_local.h
```diff
@@ -246,7 +246,7 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	  FRUSTUM_BOTTOM,
 	  FRUSTUM_TOP,
 	  FRUSTUM_NEAR,
-	  FRUSTUM_FAR,
+	  FRUSTUM_FAR = 5,
```
bmorel, line 249: I suspect this might be actually incorrect, but that's what the compiler really does.

src/engine/renderer/gl_shader.h
```diff
@@ -589,7 +589,10 @@ class GLUniform4f : protected GLUniform
 #if defined( USE_UNIFORM_FIREWALL )
 		vec4_t *firewall = ( vec4_t * ) &p->uniformFirewall[ _firewallIndex ];
 
-		if ( Vector4Compare( *firewall, v ) )
+		if ( !( *firewall[ 0 ] != v[ 0 ] ||
```
bmorel, line 592: Moving this outside of q_shared.h might fasten a bit compilation, but I initially found it because of a warning about comparing floats. I think there is room for improvements, here.

src/engine/renderer/tr_scene.cpp
```diff
@@ -700,6 +700,7 @@ qhandle_t RE_RegisterVisTest()
 			break;
 		}
 	}
+	ASSERT ( !test->registered );
```
bmorel, line 703: I would prefer to return an error code, but since I don't know which value shuold be considered invalid, I simply put an ASSERT so that at least, the code would crash instead of probably overwriting content.

illwieckz: Maybe add this as a comment?

src/engine/renderer/tr_local.h
illwieckz: Maybe ad this as a comment?

src/engine/renderer/gl_shader.h
illwieckz: I'm not very happy with replacing reusable syntax with custom one as well…
What was the issue with floats?

src/engine/server/sv_client.cpp
```diff
@@ -528,7 +528,7 @@ the same as cl->downloadClientBlock
 */
 void SV_NextDownload_f( client_t *cl, const Cmd::Args& args )
 {
-	int block;
```
slipher, line 531: What's wrong here? It's initialized by `Str::ParseInt`.

If compilers are going to give bogus 'maybe unitialized' warnings, it's better to turn off those warnings rather than adding nonsensical assignments everywhere.

src/engine/renderer/tr_local.h
slipher: It seems to be intentional that FRUSTUM_FAR is often skipped in arrays or loops of length FRUSTUM_PLANES. Comparing our R_RecursiveWorldNode to [ioq3's](https://github.com/ioquake/ioq3/blob/bc7842e301f39df5cbe1bbb1da29831021fd7a7f/code/renderergl2/tr_world.c#L420), we can see that their (unrolled) version of the loop has 5 iterations, same as ours.

There is at least one minor mistake in the usage of FRUSTUM_PLANES though. The signature of R_CalcFrustumFarCorners suggests that an array of length 5 is a good input, but it reads from the 6th element.

src/engine/renderer/gl_shader.h
slipher: You could use `memcmp` here

src/engine/renderer/tr_scene.cpp
illwieckz: > Due to the 0 handle confusion it looks like this feature is probably broken right now. Apparently the only way it can be activated is if there is a `misc_light_flare` or `gfx_light_flare` entity in a map. Maybe @illwieckz knows how to find a map with such an entity.

Among Unvanquished stock maps, _Parpax_ and _Station15_ have some. Among Tremulous ones, _Nexus 6_ and _Niveus_ have some of those entities as well. Among community maps, _USS Tremor_ has some.

If I'm right, the flares are obvious in _Niveus_ (if what i would name flares in games are those flare entities in map).

Note: _I know there is a flare feature in the q3map2 map compiler we don't use, I don't know if it's related or not._

src/engine/renderer/gl_shader.h
illwieckz: See that comment as I did some comparison on similar code before:

https://github.com/DaemonEngine/Daemon/pull/389#issuecomment-713729634

src/engine/framework/BaseCommands.cpp
```diff
@@ -392,33 +391,38 @@ namespace Cmd {
...
-                    result = intValue1 >= intValue2;
+                    } else if (relation == ">=" or relation == "≥") {
+                        result = intValue1 >= intValue2;
 
-                } else {
-                    Print( "invalid relation operator in if command. valid relation operators are = != ≠ < > ≥ >= ≤ <= eq ne in !in" );
-                    Usage(args);
-                    return;
+                    } else {
+                        Print( "invalid relation operator in if command. valid relation operators are = != ≠ < > ≥ >= ≤ <= eq ne in !in" );
```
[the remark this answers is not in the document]
bmorel: Yes, in an upper-level if. My guess was probably that this was intentional, maybe to shave some cycles or some other reason.

src/engine/server/sv_client.cpp
bmorel: oh, right, value is guaranteed to be written into

src/engine/framework/BaseCommands.cpp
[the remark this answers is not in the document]
bmorel: was from a useless commit in the end.

src/engine/renderer/gl_shader.h
bmorel: Well, the point here is that the `Vector4Compare` inline is used only once in the whole daemon code base, and is in a header which is included from everywhere. So I don't think it's that useful. Maybe there are other functions doing the same thing around, though, and if yes, it would be interesting to use those instead of memcmp() (that I just pushed, this is a nice trick I'll remember it)

src/engine/renderer/gl_shader.h
```diff
@@ -589,7 +589,7 @@ class GLUniform4f : protected GLUniform
 #if defined( USE_UNIFORM_FIREWALL )
 		vec4_t *firewall = ( vec4_t * ) &p->uniformFirewall[ _firewallIndex ];
 
-		if ( Vector4Compare( *firewall, v ) )
+		if ( !memcmp( *firewall, v, sizeof( v ) ) )
```
slipher, line 592: Thanks to the wacky C behavior with arrays as function arguments, they type of `v` is `float*` so its size will be pointer size. Use `sizeof( *firewall )` instead.

Alternatively you could use the Math::Vector lib we have. Like `Vec4::Load(firewall) == Vec4::Load(v)`

src/engine/renderer/tr_scene.cpp
```diff
@@ -684,22 +684,24 @@ so the result may be available delayed.
 */
 qhandle_t RE_RegisterVisTest()
 {
-	int hTest;
-	visTest_t *test;
+	visTest_t *test = nullptr;
```
slipher, line 687: This assignment shouldn't be needed to avoid warnings because it is guaranteed to be assigned in the first iteration of the loop.

src/engine/renderer/tr_local.h
bmorel: > There is at least one minor mistake in the usage of FRUSTUM_PLANES though. The signature of R_CalcFrustumFarCorners suggests that an array of length 5 is a good input, but it reads from the 6th element.

So, what should I do? I don't know what those FRUSTUM_PLANES planes are about, I only added that change to show explicitly what the code, in practice, does, and which might not be that obvious.

slipher: You don't have to do anything.

src/engine/renderer/gl_shader.h
slipher: You didn't fix it

bmorel: woopsie. Got messed in my rebases.

PR #469, opened 2021-05-15
Handle null locale country

slipher:
The bug was introduced about 1 year ago (859709d6c46cbdcad001b65fabae885821cb5fd3), so it is a regression in 0.52.


mboerwinkle:
@slipher Thanks! Sorry for the poorly-placed original report.
Martin

PR #476, opened 2021-06-01
Support older and newer CMake at the same time, fix #475

### LINKED ISSUE #475: CMake requirement is too high for Ubuntu Xenial (16.04)
During pre-0.52 development cycle I regularly tested building and running the game on Ubuntu Xenial (16.04), it worked fine.

That's why I written in [Dæmon 0.52 blog post](https://unvanquished.net/announcing-daemon-engine-beta-0-52/):

> We build the Unvanquished release on Debian Buster and the produced Linux build must work on distributions from the Ubuntu Bionic era (2018, we made specific efforts for this), and the engine is still buildable on Ubuntu Xenial (2016): our continuous integration system verifies it for us and we tested such binaries were playable.

But it looks like I missed something, I tried to build it today and got this:

```
cmake '/home/illwieckz/dev/buildme-unvanquished/Daemon' -B'/home/illwieckz/dev/buildme-unvanquished/build/engine/fast' \
	-DUSE_CURSES=OFF -D'USE_BREAKPAD=OFF' -D'CMAKE_BUILD_TYPE=Release' -D'USE_DEBUG_OPTIMIZE=OFF' -D'USE_LTO'='OFF' \
	 \
	-D'EXTERNAL_DEPS_DIR=/home/illwieckz/dev/buildme-unvanquished/build/deps' \
	-D'BUILD_SERVER=OFF' -D'BUILD_CLIENT=ON' -D'BUILD_TTY_CLIENT=OFF' \
	-G'Unix Makefiles'
CMake Error at CMakeLists.txt:27 (cmake_minimum_required):
  CMake 3.12 or higher is required.  You are running version 3.5.1
```

The requirement for 3.12 was added in f2fe511fdeebc34ab39a2e78edbd58eed371841e as part of #413 to fix #411.

Due to the proprietary nature of Nvidia drivers, some old Nvidia GPUs may require Xenial, that's why I tested Xenial before releasing (and those GPUs doesn't have featureful enough free open source drivers). Though, if they are supposedly able to run the game in term of supported features, they are old and slow.

illwieckz:
Instead of raising the CMake version requirement, the version is compared in CMakeList to check for older ways to do things, or to do the modern way we recommend if supported.

See f2fe511fdeebc34ab39a2e78edbd58eed371841e
See 54ea3cff20d8de768b4c82695d04e2530f4d01d1 and e062145f41836259e0ac156a4afe8b6113bda415
See c1c5d592bcdca90c070eae2bd54f4be653a546fd

Those very recent commits were those who introduced the incompatibilities without fallback.

A CMake warning is displayed for people building with CMake older than 3.12.

The engine builds (and run) on Ubuntu Xenial again. Fix #475.


illwieckz:
Before:

```
CMake Error at CMakeLists.txt:27 (cmake_minimum_required):
  CMake 3.12 or higher is required.  You are running version 3.5.1
```

After:

```
CMake Warning at CMakeLists.txt:32 (message):
  CMake 3.12 or higher is recommended.  You are running version 3.5.1
```

illwieckz:
There is solid solutions to get 3.12 CMake on older system, both Ubuntu repositories and universal installer scripts that are verified to work on Xenial (16.04) and Trusty (14.04) that are still under ESM. So we can avoid making the CMake files more complicated and keep CMake 3.12 as an hard requirement.

PR #478, opened 2021-06-06
renderer/glimp: new GL detection and selection code

illwieckz:
_Update:_ 2022-02-16

~~This PR is low priority (existing code works).~~ This fix many bugs, this is now highly wanted.

The effort was initially motivated by #477 to provide a new GL selection code that first detects the higher GL version supported before trying any custom configuration, and if no custom configuration (or it fails), loads the higher GL version possible. This way `/gfxinfo` displays the higher GL supported by the card (Nvidia proprietary driver rewrites the GL and GLSL version strings with currently requested GL version).

With default configuration, the code only tries OpenGL 3.2 core and 2.1 compatibility for success (and all versions below on error to provide user a meaningful error message). By setting `r_glExtendedValidation` to `1`, an user can also detect (at the cost of a slower startup) the best OpenGL profile his hardware and software support, by iterating all known OpenGL versions, which may produce more useful logs on demand.

Subsequent `vid_restart` calls are also now faster, by skipping validation test that was already done before. There are also more reuse of window and context when possible.

Some bugs were also fixed in the meantime, like fullscreen not being flagged as borderless while borderless option is enabled, meaning switching from fullscreen to windowed would provide user a bordered window instead of a borderless window.

Most error messages are now more meaningful.

Diving in the code and rewriting it also uncovered some unused variables that were not detected as unused because of the code convolution. Unless I did a mistake, the value of `r_ext_multisample` and `r_alphabits` appears to be unused. The first one may explain why I haven't seen multisampling working in years… And also some bugs were caught and fixed, including crashes. In the end, this PR is more about making the code more robust and to fix bugs than to improve logging, logging is just made better in the process (and is easier to make it better).

An extra commit enables extra compatibility in Intel drivers for some Intel cards on Linux (at this point, it helps to have better meaningful error messages).

About logging, It looks like there is no way to know the available GL profiles and versions without testing all of them. Even  [`glxinfo`](https://cgit.freedesktop.org/mesa/demos/tree/src/xdemos/glinfo_common.h#n87) does it. This is because you need to set the version to create a context, and you need the context to be created to query the version… Hence the need for looping every GL version and profile to know what's the best supported ones. 🤦‍♀️

---

## sdl_glimp,tr_init: rewrite the original GL selection code, improve gfxinfo

This commit squashes multiple commits by illwieckz and slipher.

# Squashed commits by illwieckz

## sdl_glimp: rewrite the original GL selection code

- Detect best configuration possible.
- Try custom configuration if exists.
- If no custom configuration or if it fails,
  load the recommended configuration if
  possible (OpenGL 3.2 core) or the best
  one available.
- Reuse window and context when possible.
- Display meaningful popup telling user
  OpenGL version is too low or required
  extensions are missing when that happens.
- Rely on more return codes for GLimp_SetMode().
- Test for negative SDL_Init return value,
  not just -1.

## sdl_glimp,tr_init: do not test all OpenGL versions unless r_glExtendedValidation is set

When r_glExtendedValidation is enabled, the engine
tests if OpenGL versions higher than 3.2 core
are supported for logging and diagnostic purpose,
but still requestes 3.2 core anyway. Some drivers
may provide more than 3.2 when requesting 3.2, this
is not our fault.

## sdl_glimp,tr_init: rewrite logging, improve gfxinfo

- Move GL query for logging purpose from tr_init to sdl_glimp.
- Do not split MODE log message.
- Unify some log.
- Add more debug log when building GL extension list.
- Also increase the extensions_string length to not
  truncate the string, 4096 is not enough, there can
  be more than 6000 characters on an OpenGL 4.6 driver.
- Also log missing extensions to make gfxinfo more useful.
- Rewrite gfxinfo in more useful way.
- List enabled and missing GL extensions.

## sdl_glimp: silence the GL error when querying if context
is core on non-core implementation

- Silence the error that may happen when querying if the
  OpenGL context uses core profile when core profile is not
  supported by the OpenGL implementation to begin with.

For example this may happen on implementations not supporting
higher than OpenGL 2.1, while forcing OpenGL 2.1 on
implementations supporting higher versions including
core profiles may not raise an error.

## sdl_glimp: make GLimp_StartDriverAndSetMode only return true on RSERR_OK

- Only return true on OK, don't return true on unknown errors.

## sdl_glimp: catch errors from GLimp_DetectAvailableModes to prevent further segfault

It may be possible to create a valid context that is unusable.

For example the 3840×2160 resolution is too large for the
Radeon 9700 and the Mesa r300 driver may print this error
when the requested resolution is higher than what is supported
by hardware:

> r300: Implementation error: Render targets are too big in r300_set_framebuffer_state, refusing to bind framebuffer state!

It will unfortunately return a valid but unusable context that will
make the engine segfault when calling GL_SetDefaultState().

## sdl_glimp: flag fullscreen window as borderless when borderless is enabled

Flag fullscreen window as borderless when borderless is enabled
otherwise the window will be bordered when leaving fullscreen
while the borderless option would be enabled.

## sdl_glimp: remove SDL_INIT_NOPARACHUTE

In GLimp_StartDriverAndSetMod() the SDL_INIT_NOPARACHUTE flag was
removed from SDL_Init( SDL_INIT_VIDEO ) as this flag is now
ignored, see https://wiki.libsdl.org/SDL_Init

# Squashed commits by slipher

## Rework GL initialization

Have GLimp_ValidateBestContext() validate both the highest-numbered
available context (if r_glExtendedValidation is enabled), and the
highest context that we actually want to use (at most 3.2). This means
most of the code in GLimp_ApplyPreferredOptions can be removed because
it was duplicating the knowledge about version preferences and the code
for instantiating them in GLimp_ValidateBestContext.

## Fix glConfig.colorBits logging

- show actual not requested
- don't log it twice at notice level

---

## system: enable GL 2.1 on Intel Gen 3 hardware, <3 papap

This may not be enough to make the game run on such
hardware anyway.

The Mesa i915 driver for GMA Gen 3 disabled GL 2.1 on such
hardware to force Google Chrome to use its CPU fallback
that was faster but we don't implement such fallback.
See https://gitlab.freedesktop.org/mesa/mesa/-/commit/a1891da7c865c80d95c450abfc0d2bc49db5f678

Only Mesa i915 on Linux supports GL 2.1 for GMA Gen 3,
so there is no similar tweak being required for Windows
and macOS.

Enabling those options would at least make the engine
properly report missing extension instead of missing
GL version, for example the Intel GMA 3100 G33 (Gen 3)
will report missing GL_ARB_half_float_vertex extension
instead of missing OpenGL 2.1 version.

The GMA 3150 is known to have wider OpenGL support than
GMA 3100, for example it has OpenGL version similar to
GMA 4 on Windows while being a GMA 3 so the list of
available GL extensions may be larger.

Thanks papap and its LanPower association for the kind
report and availability for testing.
http://asso.lanpower.free.fr

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -38,6 +38,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_glDebugProfile;
 	cvar_t      *r_glDebugMode;
 	cvar_t      *r_glAllowSoftware;
+	cvar_t      *r_glExtendedValidation;
 
 	cvar_t      *r_verbose;
 	cvar_t      *r_ignore;
@@ -101,7 +102,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_checkGLErrors;
 	cvar_t      *r_logFile;
 
-	cvar_t      *r_stencilbits;
 	cvar_t      *r_depthbits;
 	cvar_t      *r_colorbits;
 	cvar_t      *r_alphabits;
@@ -801,11 +801,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		GLimp_LogComment( "--- GL_SetDefaultState ---\n" );
 
 		GL_ClearDepth( 1.0f );
-
-		if ( glConfig.stencilBits >= 4 )
-		{
-			GL_ClearStencil( 0 );
-		}
+		GL_ClearStencil( 0 );
 
 		GL_FrontFace( GL_CCW );
 		GL_CullFace( GL_FRONT );
@@ -905,44 +901,142 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		Log::Notice("GL_VENDOR: %s", glConfig.vendor_string );
 		Log::Notice("GL_RENDERER: %s", glConfig.renderer_string );
 		Log::Notice("GL_VERSION: %s", glConfig.version_string );
-		Log::Debug("GL_EXTENSIONS: %s", glConfig.extensions_string );
-		Log::Debug("GL_MAX_TEXTURE_SIZE: %d", glConfig.maxTextureSize );
+		Log::Debug("GL_EXTENSIONS: %s", glConfig2.glExtensionsString );
+		Log::Notice("GL_MAX_TEXTURE_SIZE: %d", glConfig.maxTextureSize );
 
 		Log::Notice("GL_SHADING_LANGUAGE_VERSION: %s", glConfig2.shadingLanguageVersionString );
 
 		Log::Notice("GL_MAX_VERTEX_UNIFORM_COMPONENTS %d", glConfig2.maxVertexUniforms );
-		Log::Debug("GL_MAX_VERTEX_ATTRIBS %d", glConfig2.maxVertexAttribs );
+		Log::Notice("GL_MAX_VERTEX_ATTRIBS %d", glConfig2.maxVertexAttribs );
 
 		if ( glConfig2.occlusionQueryAvailable )
 		{
-			Log::Debug("%d occlusion query bits", glConfig2.occlusionQueryBits );
+			Log::Notice("Occlusion query bits: %d", glConfig2.occlusionQueryBits );
 		}
 
 		if ( glConfig2.drawBuffersAvailable )
 		{
-			Log::Debug("GL_MAX_DRAW_BUFFERS: %d", glConfig2.maxDrawBuffers );
+			Log::Notice("GL_MAX_DRAW_BUFFERS: %d", glConfig2.maxDrawBuffers );
 		}
 
 		if ( glConfig2.textureAnisotropyAvailable )
 		{
-			Log::Debug("GL_TEXTURE_MAX_ANISOTROPY_EXT: %f", glConfig2.maxTextureAnisotropy );
+			Log::Notice("GL_TEXTURE_MAX_ANISOTROPY_EXT: %f", glConfig2.maxTextureAnisotropy );
 		}
 
-		Log::Debug("GL_MAX_RENDERBUFFER_SIZE: %d", glConfig2.maxRenderbufferSize );
-		Log::Debug("GL_MAX_COLOR_ATTACHMENTS: %d", glConfig2.maxColorAttachments );
+		Log::Notice("GL_MAX_RENDERBUFFER_SIZE: %d", glConfig2.maxRenderbufferSize );
+		Log::Notice("GL_MAX_COLOR_ATTACHMENTS: %d", glConfig2.maxColorAttachments );
 
-		Log::Debug("\nPIXELFORMAT: color(%d-bits) Z(%d-bit) stencil(%d-bits)", glConfig.colorBits,
-		           glConfig.depthBits, glConfig.stencilBits );
-		Log::Debug("MODE: %d, %d x %d %s hz:", r_mode->integer, glConfig.vidWidth, glConfig.vidHeight,
-		           fsstrings[ r_fullscreen->integer == 1 ] );
+		Log::Notice("PIXELFORMAT: color(%d-bits)", glConfig.colorBits );
 
-		if ( glConfig.displayFrequency )
 		{
-			Log::Debug("%d", glConfig.displayFrequency );
+			std::string out;
+			if ( glConfig.displayFrequency )
+			{
+				out = Str::Format("%d", glConfig.displayFrequency );
+			}
+			else
+			{
+				out = "N/A";
+			}
+
+			Log::Notice("MODE: %d, %d x %d %s hz: %s",
+				r_mode->integer,
+				glConfig.vidWidth, glConfig.vidHeight,
+				fsstrings[ r_fullscreen->integer == 1 ],
+				out
+				);
+		}
+
+		if ( !!r_glExtendedValidation->integer )
+		{
+			Log::Notice("Using OpenGL version %d.%d, requested: %d.%d, highest: %d.%d",
+				glConfig2.glMajor, glConfig2.glMinor, glConfig2.glRequestedMajor, glConfig2.glRequestedMinor,
+				glConfig2.glHighestMajor, glConfig2.glHighestMinor );
 		}
 		else
 		{
-			Log::Debug("N/A" );
+			Log::Notice("Using OpenGL version %d.%d, requested: %d.%d", glConfig2.glMajor, glConfig2.glMinor, glConfig2.glRequestedMajor, glConfig2.glRequestedMinor );
+		}
+
+		if ( glConfig.driverType == glDriverType_t::GLDRV_OPENGL3 )
+		{
+			Log::Notice("%sUsing OpenGL 3.x context.", Color::ToString( Color::Green ) );
+
+			/* See https://www.khronos.org/opengl/wiki/OpenGL_Context
+			for information about core, compatibility and forward context. */
+
+			if ( glConfig2.glCoreProfile )
+			{
+				Log::Notice("%sUsing an OpenGL core profile.", Color::ToString( Color::Green ) );
+			}
+			else
+			{
+				Log::Notice("%sUsing an OpenGL compatibility profile.", Color::ToString( Color::Red ) );
+			}
+
+			if ( glConfig2.glForwardCompatibleContext )
+			{
+				Log::Notice("OpenGL 3.x context is forward compatible.");
+			}
+			else
+			{
+				Log::Notice("OpenGL 3.x context is not forward compatible.");
+			}
+		}
+		else
+		{
+			Log::Notice("%sUsing OpenGL 2.x context.", Color::ToString( Color::Red ) );
+		}
+
+		if ( glConfig2.glEnabledExtensionsString.length() != 0 )
+		{
+			Log::Notice("%sUsing OpenGL extensions: %s", Color::ToString( Color::Green ), glConfig2.glEnabledExtensionsString );
+		}
+
+		if ( glConfig2.glMissingExtensionsString.length() != 0 )
+		{
+			Log::Notice("%sMissing OpenGL extensions: %s", Color::ToString( Color::Red ), glConfig2.glMissingExtensionsString );
+		}
+
+		if ( glConfig.hardwareType == glHardwareType_t::GLHW_R300 )
+		{
+			Log::Notice("%sUsing ATI R300 approximations.", Color::ToString( Color::Red ));
+		}
+
+		if ( glConfig.textureCompression != textureCompression_t::TC_NONE )
+		{
+			Log::Notice("%sUsing S3TC (DXTC) texture compression.", Color::ToString( Color::Green ) );
+		}
+
+		if ( glConfig2.vboVertexSkinningAvailable )
+		{
+			/* Mesa drivers usually support 256 bones, Nvidia proprietary drivers
+			usually support 233 bones, OpenGL 2.1 hardware usually supports no more
+			than 41 bones which may not be enough to use hardware acceleration on
+			models from games like Unvanquished. */
+			if ( glConfig2.maxVertexSkinningBones < 233 )
+			{
+				Log::Notice("%sUsing GPU vertex skinning with max %i bones in a single pass, some models may not be hardware accelerated.", Color::ToString( Color::Red ), glConfig2.maxVertexSkinningBones );
+			}
+			else
+			{
+				Log::Notice("%sUsing GPU vertex skinning with max %i bones in a single pass, models are hardware accelerated.", Color::ToString( Color::Green ), glConfig2.maxVertexSkinningBones );
+			}
+		}
+		else
+		{
+			Log::Notice("%sMissing GPU vertex skinning, models are not hardware-accelerated.", Color::ToString( Color::Red ) );
+		}
+
+		if ( glConfig.smpActive )
+		{
+			Log::Notice("Using dual processor acceleration." );
+		}
+
+		if ( r_finish->integer )
+		{
+			Log::Notice("Forcing glFinish." );
 		}
 
 		Log::Debug("texturemode: %s", r_textureMode->string );
@@ -951,62 +1045,6 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		Log::Debug("ignoreMaterialMinDimension: %d", r_ignoreMaterialMinDimension->integer );
 		Log::Debug("ignoreMaterialMaxDimension: %d", r_ignoreMaterialMaxDimension->integer );
 		Log::Debug("replaceMaterialMinDimensionIfPresentWithMaxDimension: %d", r_replaceMaterialMinDimensionIfPresentWithMaxDimension->integer );
-
-		if ( glConfig.driverType == glDriverType_t::GLDRV_OPENGL3 )
-		{
-			int contextFlags, profile;
-
-			Log::Notice("%sUsing OpenGL 3.x context", Color::ToString( Color::Green ) );
-
-			// check if we have a core-profile
-			glGetIntegerv( GL_CONTEXT_PROFILE_MASK, &profile );
-
-			if ( profile == GL_CONTEXT_CORE_PROFILE_BIT )
-			{
-				Log::Debug("%sHaving a core profile", Color::ToString( Color::Green ) );
-			}
-			else
-			{
-				Log::Debug("%sHaving a compatibility profile", Color::ToString( Color::Red ) );
-			}
-
-			// check if context is forward compatible
-			glGetIntegerv( GL_CONTEXT_FLAGS, &contextFlags );
-
-			if ( contextFlags & GL_CONTEXT_FLAG_FORWARD_COMPATIBLE_BIT )
-			{
-				Log::Debug("%sContext is forward compatible", Color::ToString( Color::Green ) );
-			}
-			else
-			{
-				Log::Debug("%sContext is NOT forward compatible", Color::ToString( Color::Red  ));
-			}
-		}
-
-		if ( glConfig.hardwareType == glHardwareType_t::GLHW_R300 )
-		{
-			Log::Debug("HACK: ATI R300 approximations" );
-		}
-
-		if ( glConfig.textureCompression != textureCompression_t::TC_NONE )
-		{
-			Log::Debug("Using S3TC (DXTC) texture compression" );
-		}
-
-		if ( glConfig2.vboVertexSkinningAvailable )
-		{
-			Log::Notice("Using GPU vertex skinning with max %i bones in a single pass", glConfig2.maxVertexSkinningBones );
-		}
-
-		if ( glConfig.smpActive )
-		{
-			Log::Debug("Using dual processor acceleration" );
-		}
-
-		if ( r_finish->integer )
-		{
-			Log::Debug("Forcing glFinish" );
-		}
 	}
 
 	static void GLSL_restart_f()
@@ -1032,6 +1070,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_glDebugProfile = ri.Cvar_Get( "r_glDebugProfile", "", CVAR_LATCH );
 		r_glDebugMode = ri.Cvar_Get( "r_glDebugMode", "0", CVAR_CHEAT );
 		r_glAllowSoftware = ri.Cvar_Get( "r_glAllowSoftware", "0", CVAR_LATCH );
+		r_glExtendedValidation = ri.Cvar_Get( "r_glExtendedValidation", "0", CVAR_LATCH );
 
 		// latched and archived variables
 		r_ext_occlusion_query = ri.Cvar_Get( "r_ext_occlusion_query", "1", CVAR_CHEAT | CVAR_LATCH );
@@ -1059,7 +1098,6 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_colorMipLevels = ri.Cvar_Get( "r_colorMipLevels", "0", CVAR_LATCH );
 		r_colorbits = ri.Cvar_Get( "r_colorbits", "0",  CVAR_LATCH );
 		r_alphabits = ri.Cvar_Get( "r_alphabits", "0",  CVAR_LATCH );
-		r_stencilbits = ri.Cvar_Get( "r_stencilbits", "8",  CVAR_LATCH );
 		r_depthbits = ri.Cvar_Get( "r_depthbits", "0",  CVAR_LATCH );
 		r_ext_multisample = ri.Cvar_Get( "r_ext_multisample", "0",  CVAR_LATCH | CVAR_ARCHIVE );
 		r_mode = ri.Cvar_Get( "r_mode", "-2", CVAR_LATCH | CVAR_ARCHIVE );
@@ -1428,7 +1466,6 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 
 		// print info
 		GfxInfo_f();
-		GL_CheckErrors();
 
 		Log::Debug("----- finished R_Init -----" );
 
--- a/src/engine/renderer/tr_public.h
+++ b/src/engine/renderer/tr_public.h
@@ -37,6 +37,7 @@ Maryland 20850 USA.
 #define __TR_PUBLIC_H
 
 #include "tr_types.h"
+#include <string>
 
 #define REF_API_VERSION 10
 
@@ -44,7 +45,21 @@ struct glconfig2_t
 {
 	bool textureCompressionRGTCAvailable;
 
+	int glHighestMajor;
+	int glHighestMinor;
+
+	int glRequestedMajor;
+	int glRequestedMinor;
+
+	int glMajor;
+	int glMinor;
+
 	bool glCoreProfile;
+	bool glForwardCompatibleContext;
+
+	std::string glExtensionsString;
+	std::string glEnabledExtensionsString;
+	std::string glMissingExtensionsString;
 
 	int      maxCubeMapTextureSize;
 
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -36,9 +36,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 static Log::Logger logger("glconfig", "", Log::Level::NOTICE);
 
-SDL_Window         *window = nullptr;
+SDL_Window *window = nullptr;
 static SDL_GLContext glContext = nullptr;
-static int colorBits = 0;
 
 #ifdef USE_SMP
 static void GLimp_SetCurrentContext( bool enable )
@@ -316,6 +315,7 @@ enum class rserr_t
 
   RSERR_INVALID_FULLSCREEN,
   RSERR_INVALID_MODE,
+  RSERR_MISSING_GL,
   RSERR_OLD_GL,
 
   RSERR_UNKNOWN
@@ -325,6 +325,9 @@ cvar_t                     *r_allowResize; // make window resizable
 cvar_t                     *r_centerWindow;
 cvar_t                     *r_displayIndex;
 cvar_t                     *r_sdlDriver;
+
+static void GLimp_DestroyContextIfExists();
+static void GLimp_DestroyWindowIfExists();
 
 /*
 ===============
@@ -347,17 +350,8 @@ void GLimp_Shutdown()
 
 #endif
 
-	if ( glContext )
-	{
-		SDL_GL_DeleteContext( glContext );
-		glContext = nullptr;
-	}
-
-	if ( window )
-	{
-		SDL_DestroyWindow( window );
-		window = nullptr;
-	}
+	GLimp_DestroyContextIfExists();
+	GLimp_DestroyWindowIfExists();
 
 	SDL_QuitSubSystem( SDL_INIT_VIDEO );
 
@@ -407,7 +401,7 @@ static int GLimp_CompareModes( const void *a, const void *b )
 GLimp_DetectAvailableModes
 ===============
 */
-static void GLimp_DetectAvailableModes()
+static bool GLimp_DetectAvailableModes()
 {
 	char     buf[ MAX_STRING_CHARS ] = { 0 };
 	SDL_Rect modes[ 128 ];
@@ -421,7 +415,7 @@ static void GLimp_DetectAvailableModes()
 	if ( SDL_GetWindowDisplayMode( window, &windowMode ) < 0 )
 	{
 		logger.Warn("Couldn't get window display mode: %s", SDL_GetError() );
-		return;
+		return false;
 	}
 
 	for ( i = 0; i < SDL_GetNumDisplayModes( display ); i++ )
@@ -436,7 +430,7 @@ static void GLimp_DetectAvailableModes()
 		if ( !mode.w || !mode.h )
 		{
 			logger.Notice("Display supports any resolution" );
-			return;
+			return true;
 		}
 
 		if ( windowMode.format != mode.format || windowMode.refresh_rate != mode.refresh_rate )
@@ -473,35 +467,122 @@ static void GLimp_DetectAvailableModes()
 		logger.Notice("Available modes: '%s'", buf );
 		ri.Cvar_Set( "r_availableModes", buf );
 	}
+
+	return true;
 }
 
-/*
-===============
-GLimp_SetMode
-===============
-*/
-static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
-{
-	const char  *glstring;
-	int         perChannelColorBits;
-	int         alphaBits, depthBits, stencilBits;
-	int         samples;
-	int         i = 0;
-	SDL_Surface *icon = nullptr;
-	SDL_DisplayMode desktopMode;
-	Uint32      flags = SDL_WINDOW_SHOWN | SDL_WINDOW_OPENGL;
-	int         x, y;
-	GLenum      glewResult;
-	int         GLmajor, GLminor;
-	int         GLEWmajor, GLEWminor, GLEWmicro;
+enum class glProfile {
+	UNDEFINED = 0,
+	COMPATIBILITY = 1,
+	CORE = 2,
+};
 
-	logger.Notice("Initializing OpenGL display" );
+struct glConfiguration {
+	int major;
+	int minor;
+	glProfile profile;
+	int colorBits;
+};
+
+static bool operator!=(const glConfiguration& c1, const glConfiguration& c2) {
+	return c1.major != c2.major
+		|| c1.minor != c2.minor
+		|| c1.profile != c2.profile
+		|| c1.colorBits != c2.colorBits;
+}
+
+static const char* GLimp_getProfileName( glProfile profile )
+{
+	ASSERT(profile != glProfile::UNDEFINED);
+	return profile == glProfile::CORE ? "core" : "compatibility";
+}
+
+static void GLimp_SetAttributes( const glConfiguration &configuration )
+{
+	int perChannelColorBits = configuration.colorBits == 24 ? 8 : 4;
+
+	SDL_GL_SetAttribute( SDL_GL_RED_SIZE, perChannelColorBits );
+	SDL_GL_SetAttribute( SDL_GL_GREEN_SIZE, perChannelColorBits );
+	SDL_GL_SetAttribute( SDL_GL_BLUE_SIZE, perChannelColorBits );
+	SDL_GL_SetAttribute( SDL_GL_DOUBLEBUFFER, 1 );
+
+	if ( !r_glAllowSoftware->integer )
+	{
+		SDL_GL_SetAttribute( SDL_GL_ACCELERATED_VISUAL, 1 );
+	}
+
+	SDL_GL_SetAttribute( SDL_GL_CONTEXT_MAJOR_VERSION, configuration.major );
+	SDL_GL_SetAttribute( SDL_GL_CONTEXT_MINOR_VERSION, configuration.minor );
+
+	if ( configuration.profile == glProfile::CORE )
+	{
+		SDL_GL_SetAttribute( SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE );
+	}
+	else
+	{
+		SDL_GL_SetAttribute( SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_COMPATIBILITY );
+	}
+
+	if ( r_glDebugProfile->integer )
+	{
+		SDL_GL_SetAttribute( SDL_GL_CONTEXT_FLAGS, SDL_GL_CONTEXT_DEBUG_FLAG );
+	}
+}
+
+static bool GLimp_CreateWindow( bool fullscreen, bool bordered, const glConfiguration &configuration )
+{
+	/* The requested attributes should be set before creating
+	an OpenGL window.
+
+	-- http://wiki.libsdl.org/SDL_GL_SetAttribute */
+	GLimp_SetAttributes( configuration );
+
+	Uint32 flags = SDL_WINDOW_SHOWN | SDL_WINDOW_OPENGL;
 
 	if ( r_allowResize->integer )
 	{
 		flags |= SDL_WINDOW_RESIZABLE;
 	}
 
+	SDL_Surface *icon = nullptr;
+
+	icon = SDL_CreateRGBSurfaceFrom( ( void * ) CLIENT_WINDOW_ICON.pixel_data,
+		CLIENT_WINDOW_ICON.width,
+		CLIENT_WINDOW_ICON.height,
+		CLIENT_WINDOW_ICON.bytes_per_pixel * 8,
+		CLIENT_WINDOW_ICON.bytes_per_pixel * CLIENT_WINDOW_ICON.width,
+#ifdef Q3_LITTLE_ENDIAN
+		0x000000FF, 0x0000FF00, 0x00FF0000, 0xFF000000
+#else
+		0xFF000000, 0x00FF0000, 0x0000FF00, 0x000000FF
+#endif
+	);
+
+	const char *windowType = nullptr;
+
+	if ( fullscreen )
+	{
+		flags |= SDL_WINDOW_FULLSCREEN;
+		windowType = "fullscreen";
+	}
+
+	/* We need to set borderless flag even when fullscreen
+	because otherwise when disabling fullscreen the window
+	will be bordered while the borderless option is enabled. */
+
+	if ( !bordered )
+	{
+		flags |= SDL_WINDOW_BORDERLESS;
+
+		/* Don't tell fullscreen window is borderless,
+		it's meaningless. */
+		if ( ! fullscreen )
+		{
+			windowType = "borderless";
+		}
+	}
+
+	int x, y;
 	if ( r_centerWindow->integer )
 	{
 		// center window on specified display
@@ -514,32 +595,168 @@ static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
 		y = SDL_WINDOWPOS_UNDEFINED_DISPLAY( r_displayIndex->integer );
 	}
 
-	icon = SDL_CreateRGBSurfaceFrom( ( void * ) CLIENT_WINDOW_ICON.pixel_data,
-			        CLIENT_WINDOW_ICON.width,
-			        CLIENT_WINDOW_ICON.height,
-			        CLIENT_WINDOW_ICON.bytes_per_pixel * 8,
-			        CLIENT_WINDOW_ICON.bytes_per_pixel * CLIENT_WINDOW_ICON.width,
-#ifdef Q3_LITTLE_ENDIAN
-			        0x000000FF, 0x0000FF00, 0x00FF0000, 0xFF000000
-#else
-			        0xFF000000, 0x00FF0000, 0x0000FF00, 0x000000FF
-#endif
-					);
+	window = SDL_CreateWindow( CLIENT_WINDOW_TITLE, x, y, glConfig.vidWidth, glConfig.vidHeight, flags );
+
+	if ( window )
+	{
+		int w, h;
+		SDL_GetWindowPosition( window, &x, &y );
+		SDL_GetWindowSize( window, &w, &h );
+		logger.Debug( "SDL %s%swindow created at %d,%d with %d×%d size",
+			windowType ? windowType : "",
+			windowType ? " ": "",
+			x, y, w, h );
+	}
+	else
+	{
+		logger.Warn( "SDL %d×%d %s%swindow not created",
+			glConfig.vidWidth, glConfig.vidHeight,
+			windowType ? windowType : "",
+			windowType ? " ": "" );
+		logger.Warn("SDL_CreateWindow failed: %s", SDL_GetError() );
+		return false;
+	}
+
+	SDL_SetWindowIcon( window, icon );
+
+	SDL_FreeSurface( icon );
+
+	return true;
+}
+
+static void GLimp_DestroyContextIfExists()
+{
+	if ( glContext != nullptr )
+	{
+		SDL_GL_DeleteContext( glContext );
+		glContext = nullptr;
+	}
+}
+
+static void GLimp_DestroyWindowIfExists()
+{
+	// Do not let orphaned context alive.
+	GLimp_DestroyContextIfExists();
+
+	if ( window != nullptr )
+	{
+		int x, y, w, h;
+		SDL_GetWindowPosition( window, &x, &y );
+		SDL_GetWindowSize( window, &w, &h );
+		logger.Debug("Destroying %d×%d SDL window at %d,%d", w, h, x, y );
+		SDL_DestroyWindow( window );
+		window = nullptr;
+	}
+}
+
+static bool GLimp_CreateContext( const glConfiguration &configuration )
+{
+	GLimp_DestroyContextIfExists();
+	glContext = SDL_GL_CreateContext( window );
+
+	const char* profileName = GLimp_getProfileName( configuration.profile );
+	if ( glContext != nullptr )
+	{
+		logger.Debug( "Valid context: %d-bit OpenGL %d.%d %s",
+			configuration.colorBits,
+			configuration.major,
+			configuration.minor,
+			profileName );
+	}
+	else
+	{
+		logger.Debug( "Invalid context: %d-bit OpenGL %d.%d %s",
+			configuration.colorBits,
+			configuration.major,
+			configuration.minor,
+			profileName );
+	}
+
+	return glContext != nullptr;
+}
+
+/* GLimp_DestroyWindowIfExists checks if window exists before
+destroying it so we can call GLimp_RecreateWindowWhenChange even
+if no window is created yet.
+
+It is assumed width, height and other things like that are unchanged,
+given vid_restart is called when changing those and then destroying
+the window before calling this. */
+static bool GLimp_RecreateWindowWhenChange( const bool fullscreen, const bool bordered, const glConfiguration &configuration )
+{
+	static bool currentFullscreen = false;
+	static bool currentBordered = false;
+	static glConfiguration currentConfiguration = {};
+
+	if ( window == nullptr
+		|| configuration != currentConfiguration )
+	{
+		currentConfiguration = configuration;
+
+		GLimp_DestroyWindowIfExists();
+
+		if ( !GLimp_CreateWindow( fullscreen, bordered, configuration ) )
+		{
+			return false;
+		}
+	}
+
+	if ( fullscreen != currentFullscreen )
+	{
+		Uint32 flags = fullscreen ? SDL_WINDOW_FULLSCREEN : 0;
+		int sdlToggled = SDL_SetWindowFullscreen( window, flags );
+
+		if ( sdlToggled < 0 )
+		{
+			GLimp_DestroyWindowIfExists();
+
+			if ( !GLimp_CreateWindow( fullscreen, bordered, configuration ) )
+			{
+				return false;
+			}
+
+			const char* windowType = fullscreen ? "fullscreen" : "windowed";
+			logger.Debug( "SDL window recreated as %s.", windowType );
+		}
+		else
+		{
+			const char* windowType = fullscreen ? "fullscreen" : "windowed";
+			logger.Debug( "SDL window set as %s.", windowType );
+		}
+	}
+
+	if ( bordered != currentBordered )
+	{
+		SDL_bool sdlBordered = bordered ? SDL_TRUE : SDL_FALSE;
+
+		SDL_SetWindowBordered( window, sdlBordered );
+
+		const char* windowType = bordered ? "bordered" : "borderless";
+		logger.Debug( "SDL window set as %s.", windowType );
+	}
+
+	currentFullscreen = fullscreen;
+	currentBordered = bordered;
+
+	return true;
+}
+
+static rserr_t GLimp_SetModeAndResolution( const int mode )
+{
+	SDL_DisplayMode desktopMode;
 
 	if ( SDL_GetDesktopDisplayMode( r_displayIndex->integer, &desktopMode ) == 0 )
 	{
 		displayAspect = ( float ) desktopMode.w / ( float ) desktopMode.h;
-
-		logger.Notice("Display aspect: %.3f", displayAspect );
 	}
 	else
 	{
 		memset( &desktopMode, 0, sizeof( SDL_DisplayMode ) );
-
-		logger.Notice("Cannot determine display aspect (%s), assuming 1.333", SDL_GetError() );
+		displayAspect = 1.333f;
+		logger.Warn("Cannot determine display aspect, assuming %.3f: %s", displayAspect, SDL_GetError() );
 	}
 
-	logger.Notice("...setting mode %d:", mode );
+	logger.Notice("Display aspect: %.3f", displayAspect );
 
 	if ( mode == -2 )
 	{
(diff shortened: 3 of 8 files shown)


illwieckz:
Note: on [this page of Khronos wiki](https://www.khronos.org/opengl/wiki/Creating_an_OpenGL_Context_(WGL)) I have read:

> Warning: Unfortunately, Windows does not allow the user to change the pixel format of a window. You get to set it exactly once. Therefore, if you want to use a different pixel format from the one your fake context used (for sRGB or multisample framebuffers, or just different bit-depths of buffers), you must destroy the window entirely and recreate it after we are finished with the dummy context.

And in that code, I never destroy a windows. when destroying a context before creating another. The thing is, with SDL, you create a context from a window, so at no time we set the pixel format when creating the window, it is set when creating the context from the window which already exists. So I would like to see some Windows user testing this code on Windows.

Assuming Windows would just create 2-bit window by default, it's also possible that trying to create a 16-bit context on a 24-bit window will just fail, which would be very OK since the next attempt (24-bit context) will succeed. But better test it on Windows.

illwieckz:
Before:

```
/gfxinfo
GL_VENDOR: NVIDIA Corporation
GL_RENDERER: Quadro K1100M/PCIe/SSE2
GL_VERSION: 3.2.0 NVIDIA 340.108
GL_SHADING_LANGUAGE_VERSION: 1.50 NVIDIA via Cg compiler
Using OpenGL 3.x context
```

After:

```
/gfxinfo
GL_VENDOR: NVIDIA Corporation 
GL_RENDERER: Quadro K1100M/PCIe/SSE2 
GL_VERSION: 4.4.0 NVIDIA 340.108 
GL_SHADING_LANGUAGE_VERSION: 4.40 NVIDIA via Cg compiler 
Using OpenGL 3.x context 
Having a core profile 
```

When forcing configuration:

```
/r_glMajorVersion 3 ; r_glMinorVersion 2 ; vid_restart
/gfxinfo
GL_VENDOR: NVIDIA Corporation 
GL_RENDERER: Quadro K1100M/PCIe/SSE2 
GL_VERSION: 3.2.0 NVIDIA 340.108 
GL_SHADING_LANGUAGE_VERSION: 1.50 NVIDIA via Cg compiler 
Using OpenGL 3.x context 
Having a core profile 
```

```
/r_glMajorVersion 4 ; r_glMinorVersion 1 ; vid_restart
/gfxinfo
GL_VENDOR: NVIDIA Corporation 
GL_RENDERER: Quadro K1100M/PCIe/SSE2 
GL_VERSION: 4.1.0 NVIDIA 340.108 
GL_SHADING_LANGUAGE_VERSION: 4.10 NVIDIA via Cg compiler 
Using OpenGL 3.x context 
Having a core profile 
```

```
/r_glMajorVersion 4 ; r_glMinorVersion 1 ; r_glProfile compat ; vid_restart
/gfxinfo
GL_VENDOR: NVIDIA Corporation 
GL_RENDERER: Quadro K1100M/PCIe/SSE2 
GL_VERSION: 4.4.0 NVIDIA 340.108 
GL_SHADING_LANGUAGE_VERSION: 4.40 NVIDIA via Cg compiler 
Using OpenGL 3.x context 
Having a compatibility profile 
```

illwieckz:
Also:

```
/r_glMajorVersion 1 ; r_glMinorVersion 3 ; vid_restart
The change to r_glMajorVersion will take effect after restart. 
The change to r_glMinorVersion will take effect after restart. 
Debug: Shutting down OpenGL subsystem 
Debug: Destroying 1920×1080 SDL window at 0,27 
Calling GetRefAPI… 
SDL_Init( SDL_INIT_VIDEO )...  
Using SDL Version 2.0.10 
SDL using driver "x11" 
Initializing OpenGL display 
Display aspect: 1.778 
...setting mode -1: 
 1920 1080 
Debug: SDL borderless window created at 0,0 with 1920×1080 size 
Debug: Valid context: 16-bit GL 2.1 compat 
Debug: Valid context: 24-bit GL 2.1 compat 
Debug: Invalid context: 16-bit GL 2.2 compat 
Debug: Invalid context: 24-bit GL 2.2 compat 
Debug: Valid context: 16-bit GL 3.0 compat 
Debug: Valid context: 24-bit GL 3.0 compat 
Debug: Valid context: 16-bit GL 3.1 compat 
Debug: Valid context: 24-bit GL 3.1 compat 
Debug: Valid context: 16-bit GL 3.2 core 
Debug: Valid context: 24-bit GL 3.2 core 
Debug: Valid context: 16-bit GL 3.3 core 
Debug: Valid context: 24-bit GL 3.3 core 
Debug: Invalid context: 16-bit GL 3.4 core 
Debug: Invalid context: 24-bit GL 3.4 core 
Debug: Valid context: 16-bit GL 4.0 core 
Debug: Valid context: 24-bit GL 4.0 core 
Debug: Valid context: 16-bit GL 4.1 core 
Debug: Valid context: 24-bit GL 4.1 core 
Debug: Valid context: 16-bit GL 4.2 core 
Debug: Valid context: 24-bit GL 4.2 core 
Debug: Valid context: 16-bit GL 4.3 core 
Debug: Valid context: 24-bit GL 4.3 core 
Debug: Valid context: 16-bit GL 4.4 core 
Debug: Valid context: 24-bit GL 4.4 core 
Debug: Invalid context: 16-bit GL 4.5 core 
Debug: Invalid context: 24-bit GL 4.5 core 
Debug: Invalid context: 16-bit GL 5.0 core 
Debug: Invalid context: 24-bit GL 5.0 core 
Best context: 24-bit GL 4.4 core 
Warn: OpenGL 1.3 is not supported, trying 4.4 instead 
Debug: Created best context: 24-bit GL 4.4 core 
Using 24 Color bits, 0 depth, 8 stencil display. 
Using GLEW 2.1.0 
Using enhanced (GL3) Renderer in GL 3.x mode... 
Available modes: '640x360 720x405 864x486 960x540 1024x576 1280x720 1600x900 1920x1080 1368x768 1360x768 1280x800 1440x900 1680x1050 1600x1024 1400x900 640x480 800x600 1024x768 1152x864 1280x960 1400x1050 1280x1024 ' 
GL_RENDERER: Quadro K1100M/PCIe/SSE2 
Detected graphics driver class 'OpenGL 3+' 
Detected graphics hardware class 'generic' 
Initializing OpenGL extensions 
...ignoring GL_ARB_debug_output 
...found shading language version 440 
...using GL_ARB_half_float_pixel 
...using GL_ARB_texture_float 
...using GL_EXT_gpu_shader4 
...using GL_EXT_texture_integer 
...using GL_ARB_texture_rg 
...found buggy Nvidia 340 driver 
...ignoring GL_ARB_texture_gather 
...using GL_EXT_texture_compression_s3tc 
...using GL_ARB_texture_compression_rgtc 
...using GL_EXT_texture_filter_anisotropic 
...using GL_ARB_half_float_vertex 
...using GL_ARB_framebuffer_object 
...using GL_ARB_get_program_binary 
...using GL_ARB_buffer_storage 
...using GL_ARB_uniform_buffer_object 
...using GL_ARB_map_buffer_range 
...using GL_ARB_sync 
glsl shaders took 123 msec to build 
Warn: Couldn't find image file 'flareShader' 
Warn: Couldn't find image file 'sun' 
GL_VENDOR: NVIDIA Corporation 
GL_RENDERER: Quadro K1100M/PCIe/SSE2 
GL_VERSION: 4.4.0 NVIDIA 340.108 
GL_SHADING_LANGUAGE_VERSION: 4.40 NVIDIA via Cg compiler 
Using OpenGL 3.x context 
Having a core profile 
```

And:

```
/r_glMajorVersion 4 ; r_glMinorVersion 6 ; vid_restart
The change to r_glMajorVersion will take effect after restart. 
The change to r_glMinorVersion will take effect after restart. 
Debug: Shutting down OpenGL subsystem 
Debug: Destroying 1920×1080 SDL window at 0,27 
Calling GetRefAPI… 
SDL_Init( SDL_INIT_VIDEO )...  
Using SDL Version 2.0.10 
SDL using driver "x11" 
Initializing OpenGL display 
Display aspect: 1.778 
...setting mode -1: 
 1920 1080 
Debug: SDL borderless window created at 0,0 with 1920×1080 size 
Debug: Valid context: 16-bit GL 2.1 compat 
Debug: Valid context: 24-bit GL 2.1 compat 
Debug: Invalid context: 16-bit GL 2.2 compat 
Debug: Invalid context: 24-bit GL 2.2 compat 
Debug: Valid context: 16-bit GL 3.0 compat 
Debug: Valid context: 24-bit GL 3.0 compat 
Debug: Valid context: 16-bit GL 3.1 compat 
Debug: Valid context: 24-bit GL 3.1 compat 
Debug: Valid context: 16-bit GL 3.2 core 
Debug: Valid context: 24-bit GL 3.2 core 
Debug: Valid context: 16-bit GL 3.3 core 
Debug: Valid context: 24-bit GL 3.3 core 
Debug: Invalid context: 16-bit GL 3.4 core 
Debug: Invalid context: 24-bit GL 3.4 core 
Debug: Valid context: 16-bit GL 4.0 core 
Debug: Valid context: 24-bit GL 4.0 core 
Debug: Valid context: 16-bit GL 4.1 core 
Debug: Valid context: 24-bit GL 4.1 core 
Debug: Valid context: 16-bit GL 4.2 core 
Debug: Valid context: 24-bit GL 4.2 core 
Debug: Valid context: 16-bit GL 4.3 core 
Debug: Valid context: 24-bit GL 4.3 core 
Debug: Valid context: 16-bit GL 4.4 core 
Debug: Valid context: 24-bit GL 4.4 core 
Debug: Invalid context: 16-bit GL 4.5 core 
Debug: Invalid context: 24-bit GL 4.5 core 
Debug: Invalid context: 16-bit GL 5.0 core 
Debug: Invalid context: 24-bit GL 5.0 core 
Best context: 24-bit GL 4.4 core 
Debug: GL version 4.6 is forced by r_MajorVersion and r_MinorVersion 
Warn: Failed custom context: 24-bit GL 4.6 core 
Warn: SDL_GL_CreateContext failed: Could not create GL context: BadMatch (invalid parameter attributes) 
Debug: Created best context: 24-bit GL 4.4 core 
Using 24 Color bits, 0 depth, 8 stencil display. 
Using GLEW 2.1.0 
Using enhanced (GL3) Renderer in GL 3.x mode... 
Available modes: '640x360 720x405 864x486 960x540 1024x576 1280x720 1600x900 1920x1080 1368x768 1360x768 1280x800 1440x900 1680x1050 1600x1024 1400x900 640x480 800x600 1024x768 1152x864 1280x960 1400x1050 1280x1024 ' 
GL_RENDERER: Quadro K1100M/PCIe/SSE2 
Detected graphics driver class 'OpenGL 3+' 
Detected graphics hardware class 'generic' 
Initializing OpenGL extensions 
...ignoring GL_ARB_debug_output 
...found shading language version 440 
...using GL_ARB_half_float_pixel 
...using GL_ARB_texture_float 
...using GL_EXT_gpu_shader4 
...using GL_EXT_texture_integer 
...using GL_ARB_texture_rg 
...found buggy Nvidia 340 driver 
...ignoring GL_ARB_texture_gather 
...using GL_EXT_texture_compression_s3tc 
...using GL_ARB_texture_compression_rgtc 
...using GL_EXT_texture_filter_anisotropic 
...using GL_ARB_half_float_vertex 
...using GL_ARB_framebuffer_object 
...using GL_ARB_get_program_binary 
...using GL_ARB_buffer_storage 
...using GL_ARB_uniform_buffer_object 
...using GL_ARB_map_buffer_range 
...using GL_ARB_sync 
glsl shaders took 125 msec to build 
Warn: Couldn't find image file 'flareShader' 
Warn: Couldn't find image file 'sun' 
GL_VENDOR: NVIDIA Corporation 
GL_RENDERER: Quadro K1100M/PCIe/SSE2 
GL_VERSION: 4.4.0 NVIDIA 340.108 
GL_SHADING_LANGUAGE_VERSION: 4.40 NVIDIA via Cg compiler 
Using OpenGL 3.x context 
Having a core profile 
```

src/engine/sys/sdl_glimp.cpp
```diff
@@ -575,8 +572,29 @@ static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
 		logger.Warn( "GLEW version < 2.0.0 doesn't support GL core profiles" );
 	}
 
+	int colorBits = std::max( 0, r_colorbits->integer );
+	// int alphaBits = std::max( 0, r_alphabits->integer );
```
slipher: commented-out code

src/engine/renderer/tr_init.cpp
```diff
@@ -935,16 +935,24 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
...
+			else
+			{
+				out = "N/A";
+			}
+
+			Log::Debug("MODE: %d, %d x %d %s hz: %s",
+				r_mode->integer,
+				glConfig.vidWidth, glConfig.vidHeight,
+				fsstrings[ r_fullscreen->integer == 1 ],
+				out.c_str()
```
slipher, line 954: don't need `c_str()`

src/engine/sys/sdl_glimp.cpp
```diff
@@ -653,17 +653,16 @@ static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
...
 		displayAspect = ( float ) desktopMode.w / ( float ) desktopMode.h;
-
 		logger.Notice("Display aspect: %.3f", displayAspect );
 	}
 	else
 	{
 		Com_Memset( &desktopMode, 0, sizeof( SDL_DisplayMode ) );
 
-		logger.Notice("Cannot determine display aspect (%s), assuming 1.333", SDL_GetError() );
+		logger.Notice("Cannot determine display aspect, assuming 1.333: %s", SDL_GetError() );
```
slipher, line 662: "Assuming 1.333" doesn't make sense because this number is never actually used anywhere. Maybe just remove the display aspect logs entirely?

slipher (commented):
What is the motivation of `reuse already created SDL window`?

I still have not reviewed `rewrite the original GL selection code`.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -575,8 +572,29 @@ static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
...
+
+	if ( colorBits > 24 )
+	{
+		logger.Warn( "Color bitness %d set by r_colorbits can't be greater to 24, fallback to 24", colorBits );
+		colorBits = 24;
+	}
+
+	int samples;
+
 	do
```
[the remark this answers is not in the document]
illwieckz: Better review the whole PR as a single commit because this is already removed, and GitHub already mark this comment as outdated without me pushing anything.

src/engine/sys/sdl_glimp.cpp
illwieckz: This is the code I mention in my PR description:

> Unless I did a mistake, the value of `r_ext_multisample` and `r_alphabits` appears to be unused.

Also this code was rewritten. It's possible I kept that comment just to remember that is the place of this code if we need it (maybe not, I don't know). I believe multisample has to be implemented, but I don't know what alphabit is for.

src/engine/sys/sdl_glimp.cpp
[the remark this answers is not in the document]
slipher: If you don't want some commits to be reviewed, then squash those I guess?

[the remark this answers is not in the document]
illwieckz: The problem is that it's like the rewrite steps of an equation in order to reduce it to solve it. Some steps are done with the only purpose to move things around in a way the behavior must be the same (that's why some unused stuff may still be there) but makes easier to do the actual stuff after that. The big problem in squashing those steps is that it makes impossible to bisect the various steps in case a mistake was introduced at some point… The bisect would just identify the squashed commit, making hard to identify the cause. Exactly like reducing an equation after having wrongly assumed something is equal, and not having reported that assumption was made…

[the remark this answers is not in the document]
illwieckz: There is many PR were I would expect separate steps to be reviewed as single changes, but for this special PR I wrote in purpose:

> The various commits reflect the iterative work I did in rewriting that code. So there is a rewrite… that is entirely rewritten again after that. I'll probably squash in the end, but keeping those commit separated can help to debug some things (like to figure out at which point of my effort I may have introduced an issue if I did). So, better review the changes as if it was a single commit.

illwieckz:
On thing I would like to see is to see that code being tested on Windows. And if that's still possible to set a 16 bit display on Windows 10, tested on a 16 bit Windows display as well.

src/engine/sys/sdl_glimp.cpp
illwieckz: That `displayAspect` variable seems to be a global one, used there:

https://github.com/DaemonEngine/Daemon/blob/a971371ef5d1df8a7ece8961f4f3865046635aba/src/engine/sys/sdl_glimp.cpp#L377-L402

Though, it is set by default to `0.0f`:

https://github.com/DaemonEngine/Daemon/blob/a971371ef5d1df8a7ece8961f4f3865046635aba/src/engine/renderer/tr_init.cpp#L31

So maybe we have a bug to not set it to `1.333` when we say we assume it's `1.333`. Also maybe we want to assume something else than `4/3` in 2021. I don't know what's it's for anyway.

src/engine/renderer/tr_init.cpp
illwieckz: Right.

illwieckz:
> What is the motivation of `reuse already created SDL window`?

To avoid the engine cycling by creating and destroying and creating again a window when testing various configuration, potentially triggering window manager animations. Before that PR we would have tested 4 configuration in worst situation, after this PR the code is expected to test all configurations that works until the highest is found, so, we may not want to create and destroy 10 windows in a row…

We need a window being already created to create a GL context, and we need to create a context to test a configuration, but we don't need to create a new window each time we need to create a GL context to test if such context is valid.

illwieckz:
That push only fixed the unnecessary `.c_str()`.

illwieckz:
I added a new commit setting some environment variable to enable GL 2.1 on Intel GMA Gen 3 on Linux.

The Mesa i915 driver for GMA Gen 3 disabled GL 2.1 on such hardware to force Google Chrome to use its CPU fallback that was faster but we don't implement such fallback.

See https://gitlab.freedesktop.org/mesa/mesa/-/commit/a1891da7c865c80d95c450abfc0d2bc49db5f678

Only Mesa i915 on Linux supports GL 2.1 for GMA Gen 3, so there is no similar tweak being required for Windows and macOS.

Mesa i915 and macOS also supports GL 2.1 on GMA Gen 4 while windows drivers don't but those tweaks are not required as the related features are enabled by default.

First Intel hardware range expected to have drivers supporting GL 2.1 on Windows is GMA Gen 5.

Enabling those options would at least make the engine properly report missing extension instead of missing GL version, for example the Intel GMA 3100 G33 (Gen 3) will report missing `GL_ARB_half_float_vertex` extension instead of missing OpenGL 2.1 version.

The GMA 3150 is known to have wider OpenGL support than GMA 3100, for example it has OpenGL version similar to GMA 4 on Windows while being a GMA 3 so the list of available GL extensions may be larger. Though I have not put my hand on it.

Thanks papap and its LanPower association ( http://asso.lanpower.free.fr ) for the kind report and availability for testing.

Before:

```
Debug: ----- R_Init -----
SDL_Init( SDL_INIT_VIDEO )... 
Using SDL Version 2.0.10
SDL using driver "x11"
Initializing OpenGL display
Display aspect: 1.250
...setting mode -1: 640×480
Debug: SDL borderless window created at 320,272 with 640×480 size
Debug: Invalid context: 16-bit GL 2.1 compat
Debug: Invalid context: 24-bit GL 2.1 compat
Debug: Invalid context: 16-bit GL 3.0 compat
Debug: Invalid context: 24-bit GL 3.0 compat
Best context: 16-bit GL 0.0 compat
Debug: Destroying 640×480 SDL window at 320,272
Warn: OpenGL is too old.

You need a graphic card with drivers supporting at least OpenGL 3.2
or OpenGL 2.1 with ARB_half_float_vertex and ARB_framebuffer_object.
Debug: ----- CL_Shutdown -----
```

After:

```
Debug: ----- R_Init -----
SDL_Init( SDL_INIT_VIDEO )... 
Using SDL Version 2.0.10
ATTENTION: default value of option fragment_shader overridden by environment.
ATTENTION: default value of option stub_occlusion_query overridden by environment.
SDL using driver "x11"
Initializing OpenGL display
Display aspect: 1.250
...setting mode -1: 640×480
Debug: SDL borderless window created at 320,272 with 640×480 size
Debug: Valid context: 16-bit GL 2.1 compat
Debug: Valid context: 24-bit GL 2.1 compat
Debug: Invalid context: 16-bit GL 2.2 compat
Debug: Invalid context: 24-bit GL 2.2 compat
Debug: Invalid context: 16-bit GL 3.0 compat
Debug: Invalid context: 24-bit GL 3.0 compat
Best context: 24-bit GL 2.1 compat
Debug: Created best context: 24-bit GL 2.1 compat
Using 24 Color bits, 0 depth, 8 stencil display.
Using GLEW 2.1.0
Using GL3 Renderer in GL 2.x mode...
Available modes: '1280x1024 640x480 800x600 1024x768 '
GL_RENDERER: Mesa DRI Intel(R) G33 
Detected graphics driver class 'integrated'
Detected graphics hardware class 'generic'
Initializing OpenGL extensions
...ignoring GL_ARB_debug_output
...found shading language version 120
...using GL_ARB_half_float_pixel
...GL_ARB_texture_float not found
...GL_EXT_gpu_shader4 not found
...GL_EXT_texture_integer not found
...GL_ARB_texture_rg not found
...GL_ARB_texture_gather not found
...using GL_EXT_texture_compression_s3tc
...GL_ARB_texture_compression_rgtc not found
...using GL_EXT_texture_filter_anisotropic
Warn: Required extension GL_ARB_half_float_vertex is missing
Debug: ----- CL_Shutdown -----
```

illwieckz:
I rewrote the commit “new GL detection and selection code”, after some time I was not happy with the way I did it first, for example I discovered that if an hypothetic driver had 3.3 support but not 2.1, the code may not test for 3.3. So now the code is just testing every supported GL version, reminds the best supported one to use it, but does not stop on any unsupported version. The behaviour is now very similar to what does the `glxinfo` command line tool on Linux. Not only such rewrite make the code more strong and less convoluted, it makes logging more useful when debugging. One can ask an user to enable debug log, run the game and then send the `daemon.log` file, we would get a good picture of the OpenGL support.

Also, when an user has a too old OpenGL version, the error message tells the user which old OpenGL version he has.

I fixed a bug where the available OpenGL extension list was truncated.

The code now only test for existing GL versions to this day, so it will not test for a future GL 4.7 if it gets released one day, without patching this code. This is the way usual tools (like `glxinfo`) work.

About specific commit:

- `rewrite the original GL selection code to help to understand it`  
  The code modified by this commit get entirely rewritten later, this is a commit that is expected to work, but main purpose was to clean-up the code in order to make me understand it before rewriting it.  
  It is not needed to do extensive review like checking for unused variables, this commit does not claim to fix unused variables. What's important to review is that the commit is not introducing mistakes or obvious wrong thing, because further rewrite is based on the assumption this commit does not break anything.
- `reuse already created SDL window`  
  This commit is another rewrite that helps to make the rewrite easier after that.  
  Review should focus on testing this code on 16-bit and 24-bit Display on Windows and verify it works (and render is hardware accelerated).
- `new GL detection and selection code`  
  The actual GL detection and selection code to be used after this PR is merged.  
  The diff is not really meaningfull, the review would be more efficient by looking at the new code. It's not a deviation of the old implementation, it's a new implementation to replace the old one.
  
Other commits are simple and small and can be reviewed as usual.

On Linux with Mesa, one may want to try those various environment variables to trigger some code path:

- `export LIBGL_DRIVERS_PATH=''` \
  SDL error for failing to load OpenGL Driver (Dæmon shutdowns before being able  to call `SDL_GetError`).
- `export DISPLAY='none'` \
  SDL error for failing to open an X11 display (Dæmon shutdowns before being able  to call `SDL_GetError`).
- `export MESA_GL_VERSION_OVERRIDE='0.1'` \
  SDL error for `badValue` (Dæmon shutdowns before being able  to call `SDL_GetError`).
- `export MESA_GL_VERSION_OVERRIDE='1.0'` \
  Dæmon error about OpenGL not available (GL 1.0 does not exist but major 1 is valid).
- `export MESA_GL_VERSION_OVERRIDE='1.1'` \
  Dæmon error about OpenGL being too old.
- `export MESA_GL_VERSION_OVERRIDE='2.1'` \
  Dæmon using GL3 renderer in 2.x compatibility mode.
- `export MESA_GL_VERSION_OVERRIDE='3.1'` \
  Dæmon using GL3 renderer with 3.1 OpenGL version even if higher is available.
- `export MESA_GL_VERSION_OVERRIDE='4.6'` \
  Dæmon using GL3 renderer with 4.6 OpenGL version.
- `export MESA_GL_VERSION_OVERRIDE='2.1'; export MESA_EXTENSION_OVERRIDE='-GL_ARB_framebuffer_object'` \
  Dæmon error about missing `ARB_framebuffer_object` OpenGL extension.
- `export MESA_GL_VERSION_OVERRIDE='2.1'; export MESA_EXTENSION_OVERRIDE='-GL_ARB_half_float_vertex'` \
  Dæmon error about missing `ARB_half_float_vertex` OpenGL extension.

I noticed the engine gets a crash when overriding a GL version that does not exist (like `MESA_GL_VERSION_OVERRIDE='5.7'`) but the root cause looks to be a bug in the underlying tech. An invalid pointer to GL extension string is returned by ` instead of either something valid or `0` (meaning that would be an error), so the engine attempts to access this invalid pointer. I'm not sure we can do anything to catch this error, except catching the segfault itself.

illwieckz:
Some samples of various logs:

```
Debug: SDL borderless window created at 0,0 with 3840×2160 size 
Debug: Invalid context: 16-bit GL 1.1 compat 
Debug: Invalid context: 24-bit GL 1.1 compat 
Debug: Invalid context: 16-bit GL 1.2 compat 
Debug: Invalid context: 24-bit GL 1.2 compat 
Debug: Invalid context: 16-bit GL 1.3 compat 
Debug: Invalid context: 24-bit GL 1.3 compat 
Debug: Invalid context: 16-bit GL 1.4 compat 
Debug: Invalid context: 24-bit GL 1.4 compat 
Debug: Invalid context: 16-bit GL 1.5 compat 
Debug: Invalid context: 24-bit GL 1.5 compat 
Debug: Invalid context: 16-bit GL 2.0 compat 
Debug: Invalid context: 24-bit GL 2.0 compat 
Debug: Invalid context: 16-bit GL 2.1 compat 
Debug: Invalid context: 24-bit GL 2.1 compat 
Debug: Invalid context: 16-bit GL 3.0 compat 
Debug: Invalid context: 24-bit GL 3.0 compat 
Debug: Invalid context: 16-bit GL 3.1 compat 
Debug: Invalid context: 24-bit GL 3.1 compat 
Debug: Invalid context: 16-bit GL 3.2 core 
Debug: Invalid context: 24-bit GL 3.2 core 
Debug: Invalid context: 16-bit GL 3.3 core 
Debug: Invalid context: 24-bit GL 3.3 core 
Debug: Invalid context: 16-bit GL 4.0 core 
Debug: Invalid context: 24-bit GL 4.0 core 
Debug: Invalid context: 16-bit GL 4.1 core 
Debug: Invalid context: 24-bit GL 4.1 core 
Debug: Invalid context: 16-bit GL 4.2 core 
Debug: Invalid context: 24-bit GL 4.2 core 
Debug: Invalid context: 16-bit GL 4.3 core 
Debug: Invalid context: 24-bit GL 4.3 core 
Debug: Invalid context: 16-bit GL 4.4 core 
Debug: Invalid context: 24-bit GL 4.4 core 
Debug: Invalid context: 16-bit GL 4.5 core 
Debug: Invalid context: 24-bit GL 4.5 core 
Debug: Invalid context: 16-bit GL 4.6 core 
Debug: Invalid context: 24-bit GL 4.6 core 
Debug: Destroying 3840×2160 SDL window at 0,0 
Warn: OpenGL is not available.

You need a graphic card with drivers supporting at least OpenGL 3.2
or OpenGL 2.1 with ARB_half_float_vertex and ARB_framebuffer_object.
```

```
Debug: SDL borderless window created at 0,0 with 3840×2160 size 
Debug: Valid context: 16-bit GL 1.1 compat 
Debug: Valid context: 24-bit GL 1.1 compat 
Debug: Invalid context: 16-bit GL 1.2 compat 
Debug: Invalid context: 24-bit GL 1.2 compat 
Debug: Invalid context: 16-bit GL 1.3 compat 
Debug: Invalid context: 24-bit GL 1.3 compat 
Debug: Invalid context: 16-bit GL 1.4 compat 
Debug: Invalid context: 24-bit GL 1.4 compat 
Debug: Invalid context: 16-bit GL 1.5 compat 
Debug: Invalid context: 24-bit GL 1.5 compat 
Debug: Invalid context: 16-bit GL 2.0 compat 
Debug: Invalid context: 24-bit GL 2.0 compat 
Debug: Invalid context: 16-bit GL 2.1 compat 
Debug: Invalid context: 24-bit GL 2.1 compat 
Debug: Invalid context: 16-bit GL 3.0 compat 
Debug: Invalid context: 24-bit GL 3.0 compat 
Debug: Invalid context: 16-bit GL 3.1 compat 
Debug: Invalid context: 24-bit GL 3.1 compat 
Debug: Invalid context: 16-bit GL 3.2 core 
Debug: Invalid context: 24-bit GL 3.2 core 
Debug: Invalid context: 16-bit GL 3.3 core 
Debug: Invalid context: 24-bit GL 3.3 core 
Debug: Invalid context: 16-bit GL 4.0 core 
Debug: Invalid context: 24-bit GL 4.0 core 
Debug: Invalid context: 16-bit GL 4.1 core 
Debug: Invalid context: 24-bit GL 4.1 core 
Debug: Invalid context: 16-bit GL 4.2 core 
Debug: Invalid context: 24-bit GL 4.2 core 
Debug: Invalid context: 16-bit GL 4.3 core 
Debug: Invalid context: 24-bit GL 4.3 core 
Debug: Invalid context: 16-bit GL 4.4 core 
Debug: Invalid context: 24-bit GL 4.4 core 
Debug: Invalid context: 16-bit GL 4.5 core 
Debug: Invalid context: 24-bit GL 4.5 core 
Debug: Invalid context: 16-bit GL 4.6 core 
Debug: Invalid context: 24-bit GL 4.6 core 
Best context: 24-bit GL 1.1 compat 
Debug: Created best context: 24-bit GL 1.1 compat 
Using 24 Color bits, 0 depth, 8 stencil display. 
Debug: Provided OpenGL 1.1 version. 
Debug: Destroying 3840×2160 SDL window at 0,0 
Warn: OpenGL 1.1 is too old.

You need a graphic card with drivers supporting at least OpenGL 3.2
or OpenGL 2.1 with ARB_half_float_vertex and ARB_framebuffer_object. 
```

```
Debug: SDL borderless window created at 0,0 with 3840×2160 size 
Debug: Valid context: 16-bit GL 1.1 compat 
Debug: Valid context: 24-bit GL 1.1 compat 
Debug: Valid context: 16-bit GL 1.2 compat 
Debug: Valid context: 24-bit GL 1.2 compat 
Debug: Valid context: 16-bit GL 1.3 compat 
Debug: Valid context: 24-bit GL 1.3 compat 
Debug: Valid context: 16-bit GL 1.4 compat 
Debug: Valid context: 24-bit GL 1.4 compat 
Debug: Valid context: 16-bit GL 1.5 compat 
Debug: Valid context: 24-bit GL 1.5 compat 
Debug: Valid context: 16-bit GL 2.0 compat 
Debug: Valid context: 24-bit GL 2.0 compat 
Debug: Valid context: 16-bit GL 2.1 compat 
Debug: Valid context: 24-bit GL 2.1 compat 
Debug: Invalid context: 16-bit GL 3.0 compat 
Debug: Invalid context: 24-bit GL 3.0 compat 
Debug: Invalid context: 16-bit GL 3.1 compat 
Debug: Invalid context: 24-bit GL 3.1 compat 
Debug: Invalid context: 16-bit GL 3.2 core 
Debug: Invalid context: 24-bit GL 3.2 core 
Debug: Invalid context: 16-bit GL 3.3 core 
Debug: Invalid context: 24-bit GL 3.3 core 
Debug: Invalid context: 16-bit GL 4.0 core 
Debug: Invalid context: 24-bit GL 4.0 core 
Debug: Invalid context: 16-bit GL 4.1 core 
Debug: Invalid context: 24-bit GL 4.1 core 
Debug: Invalid context: 16-bit GL 4.2 core 
Debug: Invalid context: 24-bit GL 4.2 core 
Debug: Invalid context: 16-bit GL 4.3 core 
Debug: Invalid context: 24-bit GL 4.3 core 
Debug: Invalid context: 16-bit GL 4.4 core 
Debug: Invalid context: 24-bit GL 4.4 core 
Debug: Invalid context: 16-bit GL 4.5 core 
Debug: Invalid context: 24-bit GL 4.5 core 
Debug: Invalid context: 16-bit GL 4.6 core 
Debug: Invalid context: 24-bit GL 4.6 core 
Best context: 24-bit GL 2.1 compat 
Debug: Created best context: 24-bit GL 2.1 compat 
Using 24 Color bits, 0 depth, 8 stencil display. 
Debug: Provided OpenGL 2.1 version. 
Using GL3 Renderer in GL 2.x mode... 
```

```
Debug: SDL borderless window created at 0,0 with 3840×2160 size 
Debug: Valid context: 16-bit GL 1.1 compat 
Debug: Valid context: 24-bit GL 1.1 compat 
Debug: Valid context: 16-bit GL 1.2 compat 
Debug: Valid context: 24-bit GL 1.2 compat 
Debug: Valid context: 16-bit GL 1.3 compat 
Debug: Valid context: 24-bit GL 1.3 compat 
Debug: Valid context: 16-bit GL 1.4 compat 
Debug: Valid context: 24-bit GL 1.4 compat 
Debug: Valid context: 16-bit GL 1.5 compat 
Debug: Valid context: 24-bit GL 1.5 compat 
Debug: Valid context: 16-bit GL 2.0 compat 
Debug: Valid context: 24-bit GL 2.0 compat 
Debug: Valid context: 16-bit GL 2.1 compat 
Debug: Valid context: 24-bit GL 2.1 compat 
Debug: Invalid context: 16-bit GL 3.0 compat 
Debug: Invalid context: 24-bit GL 3.0 compat 
Debug: Invalid context: 16-bit GL 3.1 compat 
Debug: Invalid context: 24-bit GL 3.1 compat 
Debug: Invalid context: 16-bit GL 3.2 core 
Debug: Invalid context: 24-bit GL 3.2 core 
Debug: Invalid context: 16-bit GL 3.3 core 
Debug: Invalid context: 24-bit GL 3.3 core 
Debug: Invalid context: 16-bit GL 4.0 core 
Debug: Invalid context: 24-bit GL 4.0 core 
Debug: Invalid context: 16-bit GL 4.1 core 
Debug: Invalid context: 24-bit GL 4.1 core 
Debug: Invalid context: 16-bit GL 4.2 core 
Debug: Invalid context: 24-bit GL 4.2 core 
Debug: Invalid context: 16-bit GL 4.3 core 
Debug: Invalid context: 24-bit GL 4.3 core 
Debug: Invalid context: 16-bit GL 4.4 core 
Debug: Invalid context: 24-bit GL 4.4 core 
Debug: Invalid context: 16-bit GL 4.5 core 
Debug: Invalid context: 24-bit GL 4.5 core 
Debug: Invalid context: 16-bit GL 4.6 core 
Debug: Invalid context: 24-bit GL 4.6 core 
Best context: 24-bit GL 2.1 compat 
Debug: Created best context: 24-bit GL 2.1 compat 
Using 24 Color bits, 0 depth, 8 stencil display. 
Debug: Provided OpenGL 2.1 version. 
Using GL3 Renderer in GL 2.x mode... 
…
Initializing OpenGL extensions 
...ignoring GL_ARB_debug_output 
...found shading language version 460 
...using GL_ARB_half_float_pixel 
...using GL_ARB_texture_float 
...using GL_EXT_gpu_shader4 
...using GL_EXT_texture_integer 
...using GL_ARB_texture_rg 
...using GL_ARB_texture_gather 
...using GL_EXT_texture_compression_s3tc 
...using GL_ARB_texture_compression_rgtc 
...using GL_EXT_texture_filter_anisotropic 
Warn: Required extension GL_ARB_half_float_vertex is missing. 
```

```
Debug: SDL borderless window created at 0,0 with 3840×2160 size 
Debug: Valid context: 16-bit GL 1.1 compat 
Debug: Valid context: 24-bit GL 1.1 compat 
Debug: Valid context: 16-bit GL 1.2 compat 
Debug: Valid context: 24-bit GL 1.2 compat 
Debug: Valid context: 16-bit GL 1.3 compat 
Debug: Valid context: 24-bit GL 1.3 compat 
Debug: Valid context: 16-bit GL 1.4 compat 
Debug: Valid context: 24-bit GL 1.4 compat 
Debug: Valid context: 16-bit GL 1.5 compat 
Debug: Valid context: 24-bit GL 1.5 compat 
Debug: Valid context: 16-bit GL 2.0 compat 
Debug: Valid context: 24-bit GL 2.0 compat 
Debug: Valid context: 16-bit GL 2.1 compat 
Debug: Valid context: 24-bit GL 2.1 compat 
Debug: Valid context: 16-bit GL 3.0 compat 
Debug: Valid context: 24-bit GL 3.0 compat 
Debug: Valid context: 16-bit GL 3.1 compat 
Debug: Valid context: 24-bit GL 3.1 compat 
Debug: Valid context: 16-bit GL 3.2 core 
Debug: Valid context: 24-bit GL 3.2 core 
Debug: Invalid context: 16-bit GL 3.3 core 
Debug: Invalid context: 24-bit GL 3.3 core 
Debug: Invalid context: 16-bit GL 4.0 core 
Debug: Invalid context: 24-bit GL 4.0 core 
Debug: Invalid context: 16-bit GL 4.1 core 
Debug: Invalid context: 24-bit GL 4.1 core 
Debug: Invalid context: 16-bit GL 4.2 core 
Debug: Invalid context: 24-bit GL 4.2 core 
Debug: Invalid context: 16-bit GL 4.3 core 
Debug: Invalid context: 24-bit GL 4.3 core 
Debug: Invalid context: 16-bit GL 4.4 core 
Debug: Invalid context: 24-bit GL 4.4 core 
Debug: Invalid context: 16-bit GL 4.5 core 
Debug: Invalid context: 24-bit GL 4.5 core 
Debug: Invalid context: 16-bit GL 4.6 core 
Debug: Invalid context: 24-bit GL 4.6 core 
Best context: 24-bit GL 3.2 core 
Debug: Created best context: 24-bit GL 3.2 core 
Using 24 Color bits, 0 depth, 8 stencil display. 
Debug: Provided OpenGL 3.2 version. 
Using GL3 Renderer in GL 3.x mode... 
```

```
Debug: SDL borderless window created at 0,0 with 3840×2160 size 
Debug: Valid context: 16-bit GL 1.1 compat 
Debug: Valid context: 24-bit GL 1.1 compat 
Debug: Valid context: 16-bit GL 1.2 compat 
Debug: Valid context: 24-bit GL 1.2 compat 
Debug: Valid context: 16-bit GL 1.3 compat 
Debug: Valid context: 24-bit GL 1.3 compat 
Debug: Valid context: 16-bit GL 1.4 compat 
Debug: Valid context: 24-bit GL 1.4 compat 
Debug: Valid context: 16-bit GL 1.5 compat 
Debug: Valid context: 24-bit GL 1.5 compat 
Debug: Valid context: 16-bit GL 2.0 compat 
Debug: Valid context: 24-bit GL 2.0 compat 
Debug: Valid context: 16-bit GL 2.1 compat 
Debug: Valid context: 24-bit GL 2.1 compat 
Debug: Valid context: 16-bit GL 3.0 compat 
Debug: Valid context: 24-bit GL 3.0 compat 
Debug: Valid context: 16-bit GL 3.1 compat 
Debug: Valid context: 24-bit GL 3.1 compat 
Debug: Valid context: 16-bit GL 3.2 core 
Debug: Valid context: 24-bit GL 3.2 core 
Debug: Valid context: 16-bit GL 3.3 core 
Debug: Valid context: 24-bit GL 3.3 core 
Debug: Valid context: 16-bit GL 4.0 core 
Debug: Valid context: 24-bit GL 4.0 core 
Debug: Valid context: 16-bit GL 4.1 core 
Debug: Valid context: 24-bit GL 4.1 core 
Debug: Valid context: 16-bit GL 4.2 core 
Debug: Valid context: 24-bit GL 4.2 core 
Debug: Valid context: 16-bit GL 4.3 core 
Debug: Valid context: 24-bit GL 4.3 core 
Debug: Valid context: 16-bit GL 4.4 core 
Debug: Valid context: 24-bit GL 4.4 core 
Debug: Valid context: 16-bit GL 4.5 core 
Debug: Valid context: 24-bit GL 4.5 core 
Debug: Valid context: 16-bit GL 4.6 core 
Debug: Valid context: 24-bit GL 4.6 core 
Best context: 24-bit GL 4.6 core 
Debug: Created best context: 24-bit GL 4.6 core 
Using 24 Color bits, 0 depth, 8 stencil display. 
Debug: Provided OpenGL 4.6 version. 
Using GL3 Renderer in GL 3.x mode... 
```

illwieckz:
Unless there is a bug, I consider this PR ready.

I squashed the commits rewriting the stuff multiple time in an iterative way.

I'm still looking for a 16-bit depth test on Windows (I assume one would have to use some compatibility mode).

slipher:
I tested on a Windows Nvidia machine with a timer that starts running after the `else if ( major == 4 && minor == 7 )` block and ends at the next iteration of the major/minor loop. Results:
```
glversiontest N=18 min=8094us max=11935us avg=9550us tot=171912us`
```
So we spent 172 ms on testing versions. Testing unused versions is a major performance hit to engine startup, which seems not worth it for some debug logs.

Anyway, I'd think that for us the *best* version is 3.2 core. This means we get just the functions we want from our target version, without any deprecated stuff or future changes, so that developers can't accidentally use them.

Is there any scenario under which testing extra versions concretely benefits the user as compared to the current code? Or is it solely for logging purposes?

slipher:
I don't think it makes sense to to create any contexts higher than 3.2. It takes a lot of time and doesn't do anything useful for the user. It's not very relevant what the driver's highest supported version is, and if you really wanted to know you could probably look it up knowing the driver string. I ask again:

> Is there any scenario under which testing extra versions concretely benefits the user as compared to the current code? Or is it solely for logging purposes?

illwieckz:
> Is there any scenario under which testing extra versions concretely benefits the user as compared to the current code? Or is it solely for logging purposes?

There is no scenario under which testing extra versions concretely primarily benefits the user outside of providing them a cross-platform `gfxinfo`, it secondarily benefits the user by improving logging and then, reporting. This is for logging purpose.

> If you really wanted to know you could probably look it up knowing the driver string.

The various strings we fetch from drivers are not reliable. It looks like Mesa (tested with radeon) always prints the highest GL version, while Nvidia prints the current version for the current context. The only reliable way to know the driver's highest supported version is to actually test it, this is what does tools like `glxinfo` on Linux but such tool isn't necessarily installed by default, and is not cross-platform, there is third-party `wglinfo` on Windows (not installed by default), and I don't know if there is an alternative on macOS.

Part of the purpose of that logging improvement is to provide users with Dæmon's `gfxinfo` a reliable and cross platform tool to know more about their system and do better reports.

Also at some point we may want to do optional reporting (disabled by default) like [0.A.D.](https://play0ad.com/) does: https://feedback.wildfiregames.com/report/opengl/

For now I do it by hand on https://wiki.unvanquished.net/wiki/GPU_compatibility_matrix

Note that I would like to make Dæmon's `gfxinfo` query and print more information (like GPU memory) on that purpose.

slipher:
I tried doing version detection inside `/gfxinfo` and it was fairly simple. In fact, I believe the overall code would be considerably smaller doing it this way than trying to mix it together with real initialization. Proof of concept: https://github.com/slipher/Daemon/tree/gldetect-gfxinfo

This follows the PR in considering 24-bit color on OpenGL 1.1 "better" than 16-bit color on OpenGL 4.5. I still don't believe that "highest OpenGL version" is a good concept or particularly useful for debugging... just because the Linux `glxinfo` command does it, doesn't mean that it's good or useful for our situation.

slipher:
To me it would make sense to have `gfxinfo` report all information, while only a smaller set would be printed during renderer initialization. (In contrast to how it is now, where GfxInfo_f is called from tr_init. I noticed in your current version you have a bunch of debug-level logs in gfxinfo which a typical user wouldn't be able to see even if they called the command.

illwieckz:
Note that I'm not against doing detection in `gfxinfo` if that works but I have some concerns:

> I tried doing version detection inside /gfxinfo and it was fairly simple.

How creating a secondary window deals with having the game window fullscreen, does it make it leave full screen? That's my biggest fear about doing that. If we're sure it does not make fullscreen game leaving fullscreen (or being minimized or other annoyance) on all system, I'm OK to do the version testing in gfxinfo. I went the way to do the testing at initialization time because it's before the game is actually rendered, so it ruled out that risk.

> I believe the overall code would be considerably smaller doing it this way than trying to mix it together with real initialization.

Initialization code will not be smaller if we do this testing in `gfxinfo`. The code in initialization is hardly compressible, it's lenghty for good reasons.

If we do the “greatest version detection” in `gfxinfo`, the only change we would do to the code in this PR code is to reduce the array of version to test, keeping the exact same code I wrote but iterating a smaller version array.

The code I wrote is lengthy to many good reasons, including preventing crashes, and including make it faster, it's possible to make it very much more concise but very much more slower or prone to crashes.

> This follows the PR in considering 24-bit color on OpenGL 1.1 "better" than 16-bit color on OpenGL 4.5.

This PR does not consider 24-bit color on OpenGL 1.1 is better than 16-bit color on OpenGL 4.5, it assumes that a 16-bit display would invalidate all 24-bit OpenGL contexts, and it assumes the same OpenGL driver would not provide an higher OpenGL version on 16-bit display and a lower OpenGL version on 24-bit display.

Testing all 24-bit display then all 16-bit display and the way the code was made lengthy in this PR is done avoid poorly documented pitfalls and get the most speed possible.

Some system may create window with a specific bitness, requiring to re-create a window when switching from 24-bit to 16-bit. For example on Windows a window is tided to a device context, and an OpenGL context is tied to a device context, and pixel format availability is given by the device context.

For example we can read that on [Khronos Wiki](https://www.khronos.org/opengl/wiki/Creating_an_OpenGL_Context_(WGL)):

> Each window in MS Windows has a Device Context (DC) associated with it. This object can store something called a Pixel Format. This is a generic structure that describes the properties of the default framebuffer that the OpenGL context you want to create should have.

I'm not sure to understand everything and it may be possible to create device contexts while keeping the window but we use SDL abstraction anyway and I didn't wanted to go deep into it, especially because it becomes display-server specific (Windows, X11, Wayland, the thing of macOS…), for example [here someone explains](https://stackoverflow.com/a/43292466/9131399) it would be possible on Windows to untie a context from a DC of a window than to tie it to another DC of another Window, but this is entirely system-specific.

When using SDL, we first create a window, then we create a context from it. This gives the impression we can reuse the window for everything, even different pixel format, experiences seems to prove it's wrong. I don't know the internal of SDL, but after having read some documentation about pixel formats on Windows I started to doubt it is a good idea to reuse the same window when switching from 24-bit to 16-bit pixel format.

My first iteration of this PR was reusing the same window whatever the pixel format, and it was working fine on my 24-bit X11 display to create a 24-bit context then create a 16-bit context reusing the same window for every GL version testing, meaning the final 24-bit context would be done using a window from which both 24-bit and 16-bit context were created before. But then I used a virtual X11 to test 16-bit display (Xephyr or Xvnc or something like that, I don't remember which one), and then I got crashes of the engine when I reused the same SDL window to create both 24-bit and 16-bit context then use the 16-bit contex to render the game.

I don't know the internals of SDL, but it looks like creating a GL context from a window is somewhat modifying the window and we better recreate the window for every bitness pixel change. It appeared creating window is super slow on my end, I didn't have to time it: I was able to notice it visually (my display looked frozen by the empty windows for the time the testing was done).

This why the GL version loop is inside the pixel bitness loop and not the other way, to reuse an N-bit window the most possible, and this is why the current code is lenhgty and a bit convoluted: the primary constraint was to destroy/create window the less possible, then the code had to be written around that constraint, even if that means it's not concise.

> I noticed in your current version you have a bunch of debug-level logs in gfxinfo which a typical user wouldn't be able to see

Yes, I decided some information were relevant to be by default, but for things I was not able to make a decision myself, I just postponed that, we can decide to change the debug level later and do more PR on this topic. It can be postponed.

One thing to know is that if we reuse the same window for both 24-bit and 16 bit context but never do anything with that window and that context (because it's just a version testing), we may not face any crash, crashes I experienced when reusing the same window for both 24-bit and 16-bit context happened later in the engine code.

illwieckz:
Also one reason why I went doing the testing at initialisation time is that `gfxinfo` is called on every startup so even if we move the version testing to gfxinfo we would still spend the time at startup.

In fact the `gfxinfo` command is called on every `vid_restart` and every map load, meaning that if we do the version testing in `gfxinfo`, the extra time would be spent not only at startup, but also on every `vid_restant` and on every map load.

On the contrary in the current PR the version testing is only done once at startup because result is cached for next `vid_restart` calls and map load does not run initialization code anyway.

Of course we can implement caching in `gfxinfo`, but then implementing caching of detection results, taking care of not reusing the same window when changing pixel bitness, taking care of specific GLEW version limitations, writing code for speed first instead of writing for conciseness, and the code would suddenly become fare more lengthy and complex.

slipher:
> Also one reason why I went doing the testing at initialisation time is that `gfxinfo` is called on every startup so even if we move the version testing to gfxinfo we would still spend the time at startup.

I'm aware of that. In the proof-of-concept I disable this calculation on renderer init and only do it when the user actually issues the `gfxinfo` command.

illwieckz:
When doing more tests, I also noticed the game can crash when reusing the same window for a OpenGL context that does not use the same OpenGL version than before:

```
X Error of failed request:  BadMatch (invalid parameter attributes)
  Major opcode of failed request:  130 (MIT-SHM)
  Minor opcode of failed request:  3 (X_ShmPutImage)
  Serial number of failed request:  231
  Current serial number in output stream:  232
```

I did the test with Xvfb and 16-bit depth.

It looks like we better recreate the window everytime the OpenGL context properties change.

illwieckz:
So, for now I recreate a window for every test, but I also start testing with OpenGL 3.2 version and skip everything greater than that for faster game launch.

Even if some rewrite may be done later (I even plan to ask some questions to SDL guys), at some point I want this being merged as it is better in many aspects than what we currently have.

illwieckz:
I added a `r_glExtendedValidation` that is set to `0` by default, in such case, the init code only test for OpenGL 3.2 core and 2.1 compatibility profiles. If this cvar is set to `1`, the init code tests for all versions known, reporting the best version supported by the system. This is what did the previous code.

Note: if neither 3.2 core neither 2.1 compatibility profile works, the init code still tests all versions below 2.1 compatibility to properly report to the user the version his system supports in the window saying the available OpenGL is too low.

Note: there is an unwanted side effect of an optimization I did, on vid_restart, the new code skips OpenGL version and profile detection code and reuses the already detected one. It means setting `r_glExtendedValidation` to `1` after startup will _not_ look-up the best available OpenGL version and profile on next `vid_restart` call. I wonder if we can abuse latched cvar mechanism to know the cvar value has been changed and then force the look-up again…

illwieckz:
I consider the PR ready.

I tested it successfully on Linux and macOS.

I rewrote  the first post

src/engine/sys/sdl_glimp.cpp
```diff
@@ -558,196 +774,542 @@ static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
...
+	}
 
-			if( !testCore && !Q_stricmp(r_glProfile->string, "core") )
-				continue;
+	return rserr_t::RSERR_OK;
+}
 
-			if( testColorBits > colorBits )
-				continue;
+static rserr_t GLimp_ApplyPreferredOptions( const bool fullscreen, const bool bordered, const glConfiguration &bestConfiguration, glConfiguration &requestedConfiguration )
```
slipher, line 1067: What is the point of this function?

src/engine/sys/sdl_glimp.cpp
```diff
@@ -558,196 +774,542 @@ static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
...
+	if ( bestValidatedConfiguration.major == 0 )
+	{
+
+		return rserr_t::RSERR_MISSING_GL;
+	}
+
+	return rserr_t::RSERR_OLD_GL;
+}
+
+static rserr_t GLimp_ApplyCustomOptions( const int GLEWmajor, const bool fullscreen, const bool bordered, const glConfiguration &bestConfiguration, glConfiguration &requestedConfiguration, bool &customOptions )
```
slipher, line 923: It seems better for this function to only return a new configuration struct. The caller can do window/context creation calls.

Also can you just use the `operator!=` between the bestConfiguration and the customized options to get the answer to whether custom options were applied?

slipher (commented):
I pushed one commit to the branch improving type safety which I made for my own reviewing benefit. Feel free to squash if needed.

I tested on Windows and didn't find anything obviously broken.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -473,35 +467,121 @@ static void GLimp_DetectAvailableModes()
...
-	int         samples;
-	int         i = 0;
-	SDL_Surface *icon = nullptr;
-	SDL_DisplayMode desktopMode;
-	Uint32      flags = SDL_WINDOW_SHOWN | SDL_WINDOW_OPENGL;
-	int         x, y;
-	GLenum      glewResult;
-	int         GLmajor, GLminor;
-	int         GLEWmajor, GLEWminor, GLEWmicro;
+	return profile == glProfile::CORE ? "core" : "compatibility";
```
illwieckz, line 496: Now that I think about it, maybe I may add an ASSERT to make sure it's not undefined.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -558,196 +774,542 @@ static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
...
+
+	/* FIXME: It looks like MSAA was only implemented in legacy renderer.
+	int samples = std::max( 0, r_ext_multisample->integer ); */
+
+	glConfig.colorBits = requestedConfiguration.colorBits;
+	glConfig.depthBits =  std::max( 0, r_depthbits->integer );
+	glConfig.stencilBits = std::max( 0, r_stencilbits->integer );
+
+	logger.Notice("Using %d color bits, %d depth, %d stencil display.",
+		glConfig.colorBits,
```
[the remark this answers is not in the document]
illwieckz: Ah, great, I can remove log stuff I didn't understood what it was about. 😄

src/engine/sys/sdl_glimp.cpp
illwieckz: It is to select OpenGL 3.2 if a greater OpenGL version is validated when “extended validation” is enabled. When “extended validation” is disabled, if 3.2 is validated this would just do nothing and keep 3.2.

I remember some talk we had when you said there using greater version wasn't doing anything useful, and you probably also said it was better to only use what we need, which I agree with.

Hence I wrote this function to force using a 3.2 context if any greater core version is validated.

When extended validation is enabled, that would add at least one context recreation to downgrade from higher validated version (example, 4.6) to 3.2. With default behaviour, the engine would test 3.2 first, validate 3.2, but then when this function is called nothing will be done since context would already be 3.2 and then context will be kept as is.

src/engine/sys/sdl_glimp.cpp
illwieckz: I added log for dimension and “actually set it to 1.333”.

src/engine/renderer/tr_public.h
```diff
@@ -37,14 +37,29 @@ Maryland 20850 USA.
...
+	int glRequestedMajor;
+	int glRequestedMinor;
+
+	int glMajor;
+	int glMinor;
+
 	bool glCoreProfile;
+	bool glForwardCompatibleContext;
+
+	std::string glExtensionsString;
```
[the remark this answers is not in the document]
illwieckz: done

src/engine/sys/sdl_glimp.cpp
```diff
@@ -804,11 +1525,26 @@ static bool GLimp_StartDriverAndSetMode( int mode, bool fullscreen, bool noborde
 		SDL_GetVersion( &v );
 
 		logger.Notice("SDL_Init( SDL_INIT_VIDEO )... " );
-		logger.Notice("Using SDL Version %u.%u.%u", v.major, v.minor, v.patch );
+		logger.Notice("Using SDL version %u.%u.%u", v.major, v.minor, v.patch );
+
+		/* SDL_INIT_NOPARACHUTE flag is removed.
```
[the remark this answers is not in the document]
illwieckz: done

src/engine/renderer/tr_init.cpp
```diff
@@ -905,108 +906,148 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
...
-			if ( contextFlags & GL_CONTEXT_FLAG_FORWARD_COMPATIBLE_BIT )
+			if ( glConfig2.glForwardCompatibleContext )
 			{
-				Log::Debug("%sContext is forward compatible", Color::ToString( Color::Green ) );
+				Log::Notice("%sOpenGL 3.x context is forward compatible.", Color::ToString( Color::Green ) );
 			}
 			else
 			{
-				Log::Debug("%sContext is NOT forward compatible", Color::ToString( Color::Red  ));
+				Log::Notice("%sOpenGL 3.x context is not forward compatible.", Color::ToString( Color::Red  ));
```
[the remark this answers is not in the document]
illwieckz: I removed all colors from that line, I don't know what is bad or good in it.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -558,196 +774,542 @@ static rserr_t GLimp_SetMode( int mode, bool fullscreen, bool noborder )
...
+		// Check if context is forward compatible.
+		int contextFlags;
+		glGetIntegerv( GL_CONTEXT_FLAGS, &contextFlags );
 
-	SDL_FreeSurface( icon );
+		glConfig2.glForwardCompatibleContext = contextFlags & GL_CONTEXT_FLAG_FORWARD_COMPATIBLE_BIT;
+
+		if ( glConfig2.glForwardCompatibleContext )
+		{
+			logger.Debug( "Provided OpenGL core context is forward compatible." );
```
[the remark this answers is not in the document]
illwieckz: Unless I miss something the other place is in `gfxinfo`, the engine may stop before `gfxinfo` call is reached.

src/engine/sys/sdl_glimp.cpp
illwieckz: @slipher I added an assert in your commit.

slipher:
I tried to remove a lot of code duplication: PTAL at the `slipher/gldetect` branch in this repo. The biggest change is that when trying versions from the version table, it keeps track of the highest version we actually want to use, in addition to the absolutely highest version. So the ApplyPreferredContext function doesn't have to do a bunch of logic to find the highest version we want to use.

Most of these commits can probably be squashed.

illwieckz:
OK thanks, for what I see in [`slipher/gldetect`](https://github.com/DaemonEngine/Daemon/commits/slipher/gldetect), there are many good things there, thanks a lot !

One comment though, on “_unnecessary glExtendedValidation variable_”:

- https://github.com/DaemonEngine/Daemon/commit/6f5f2baa873b4c62b97eb88761b5b7c857d4c064#r66640359

> If I remember correctly this was to detect the event of r_glExtendedValidation having been modified after startup and before vid_restart to not use cached results of previous validation when the cvar is modified, as if r_glExtendedValidation was disabled before and become enabled, code may still assume 3.2 is best version available without redoing tests starting from 4.6.

So whatever the way it is implemented, I want to be sure modifying `r_glExtendedValidation` redo the validation from scratch.

slipher:
> I want to be sure modifying r_glExtendedValidation redo the validation from scratch on next vid_restart

Is there some reason to do extended validation from scratch every time instead of only the first time it is requested?

illwieckz:
I'm not sure to get the question. Do you mean you have the idea to reset the cvar to zero once the extended validation is done.

slipher:
No. I'm doing extended validation only the *first* time that the cvar is turned on, rather than every time it is changed.

illwieckz:
I have two needs:

- be able to ask someone to do extended validation so he can send me a complete log (the cvar does not have to be persistent);
- be able to do extended validation anytime when validating bunch of hardware/software combinations (the cvar being persistent can be useful, but I can workaround by scripting).

If I remember correctly, my initial code was redoing non-extended validation from scratch on `vid_restart` after switching `r_glExtendedValidation` off and was redoing extended validation from scratch on `vid_restart` after switching `r_glExtendedValidation` on.

There may be other ways to fulfill my needs, but at least this one had predictable behaviour: after a `vid_restart` the selected profile was one produced accordingly to `r_glExtendedValidation` even if it was one selected at a previous `vid_restart` call:

```
initial config: r_glExtendedValidation 0
engine start: basic validation
/vid_restart: reuse cached profile version from basic validation
/vid_restart: reuse cached profile version from basic validation
/set r_glExtendedValidation 1
/vid_restart: extended validation
/vid_restart: reuse cached profile version from extended validation
/vid_restart: reuse cached profile version from extended validation
/set r_glExtendedValidation 0
/vid_restart: basic validation
/vid_restart: reuse cached profile version from basic validation
/vid_restart: reuse cached profile version from basic validation
```

slipher:
In my version, the max extended version no longer flows into the rest of the program, as the normal max version is stored in a separate variable. The max extended version is only used for the diagnostic log statements.

The validation function always populates the normal max version, and additionally populates the extended max version if the cvar is on. Both of these persist across vid_restarts.

The validation function is called if the normal max version is not yet known or if the cvar is on and the extended max version is not yet known.

illwieckz:
>  as the normal max version is stored in a separate variable.

Ah OK, I missed that, it looks like that does the job.

So once the max version is known, it's known until the end of the program, whatever the current value of the cvar, that's good. It looks to fulfill all my needs.

illwieckz:
@slipher I merged your fixes and squashed them. I added some `Co-authored-by:` lines in commit message to reflect that.

LGTM.

slipher:
I missed a couple chances to use the ContextDescription function in GLimp_CreateContext and with Sys::Error "GLEW initialization failed".

LGTM

illwieckz:
> I missed a couple chances to use the ContextDescription function in GLimp_CreateContext and with Sys::Error "GLEW initialization failed".

Well, I'm not sure to entirely get it but it's always possible for someone to add another patch above this work one day. Thanks a lot for the deep review(s) your precious advises and the patience!

PR #481, opened 2021-06-09
readme: recastnavigation is not a submodule

illwieckz:
readme: recastnavigation is not a submodule

Also unify the wording with Unvanquished readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ cd Daemon/
 git submodule update --init --recursive
 ```
 
-If cmake complains about missing files in `recastnavigation/` folder or similar issue then you have skipped this step.
+If cmake complains about missing files in `libs/crunch/` folder or similar issue then you have skipped this step.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,7 +72,7 @@ cd Daemon/
 git submodule update --init --recursive
 ```
 
-If cmake complains about missing files in `recastnavigation/` folder or similar issue then you have skipped this step.
```
slipher, line 75: The old wording seems more accurate - it checks for specific files, not a directory; I think the message is something like "Missing source file".

`libs/crunch/` would be a better example since Breakpad is off by default.

illwieckz: OK, I can unify the other way. Good point for breakpad/crunch, I just picked the first one on alphabetical order, but crunch is better indeed.

slipher (approved):

Issue #474, opened 2021-05-31
“could not load OpenGL subsystem” bug with Nvidia when running a non-release build…


illwieckz:
I usually build the game on my workstation running Ubuntu 20.04 (no Nvidia stuff on it, at all), then run the game on various hosts with an Ubuntu 20.04 running on an USB key. Usually, I rsync my build and run it, but while testing 473 I went into troubles…

On the USB key, running Nvidia driver, I got this:

```
`^3Warn: GLW_StartOpenGL() - could not load OpenGL subsystem: Missing GL version
```

One thing to notice is that, if I'm right, I was running the old 340 driver on old 5.4 kernel.

I then rebuilt the game from the system on the USB key, got the same error.

At the same time the 0.52 release build works on this system.

illwieckz:
So, the same custom non-release build, on the same computer (with Quadro 1100M), on the same system, I get the error with the 340 driver but ont with the 390 driver:

|Nvidia →   |340.108|390.143|
|-----------|-------|-------|
|Linux 5.4.0|error  |       |
|Linux 5.8.0|error  |works  |

But with the release build, built in docker, it always work:

|Nvidia →   |340.108|390.143|
|-----------|-------|-------|
|Linux 5.4.0|works  |       |
|Linux 5.8.0|works  |works  |

illwieckz:
Hmm, the 340 driver is the buggy one we implemented a detection to disable an optional feature to workaround a driver bug, see #340 and #370.

illwieckz:
And we read the version to do the detection…

illwieckz:
An `ldd daemon` on the self-built binary binary shows `libGLdispatche.so.0` but not with the release build.

illwieckz:
With:

```sh
cmake -DOpenGL_GL_PREFERENCE=LEGACY ..
```

and calling `make`, it relinks `daemon` and I can run it.

If this variable is unset or set to `GLVND`, I cannot run the game on 340 driver.

illwieckz:
From: https://cmake.org/cmake/help/latest/module/FindOpenGL.html#linux-specific

[...]

Issue #482, opened 2021-06-11
Tracking issue: native Wayland support


necessarily-equal:
Wayland is the new fancy display protocol on Linux. It should bring security improvements, kill tearing and bring higher performance to the Linux world. It would be nice to support it.

To try Unvanquished on Wayland, set the `SDL_VIDEODRIVER=wayland` env var. It will force the wayland backend.

## Bugs

* [ ] Dæmon doesn't start: `Warn: GLW_StartOpenGL() - could not load OpenGL subsystem: Unknown error` at start (PR incoming)
* [ ] Refresh rate seems lower on wayland

… And more issues to come as we drive this forward.

PR #483, opened 2021-06-11
Don't fail on non-GLX backends

necessarily-equal:
Not having a GLX context is not a bug: wayland uses EGL for example.
This is inspired by:
https://github.com/endless-sky/endless-sky/pull/5230/commits/5f5fb100422685f2eed443bb5bf7f5dbadef2aa9

This may help in https://github.com/DaemonEngine/Daemon/issues/474.

With this commit, the game will start on wayland, see https://github.com/DaemonEngine/Daemon/issues/482.


illwieckz:
For information, this does not fix #474 on my end.

PR #485, opened 2021-06-18
tr_shade: small rewrite of a test I just stumbled on

illwieckz:
just a small rewrite of a test I just stumbled on.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -495,8 +495,6 @@ void Tess_Begin( void ( *stageIteratorFunc )(),
 		Tess_MapVBOs( false );
 	}
 
-	bool isSky = ( state != nullptr && state->isSky != false );
-
 	tess.lightShader = lightShader;
 
 	tess.stageIteratorFunc = stageIteratorFunc;
@@ -507,17 +505,16 @@ void Tess_Begin( void ( *stageIteratorFunc )(),
 		Sys::Error( "tess.stageIteratorFunc == NULL" );
 	}
 
-	if ( tess.stageIteratorFunc == &Tess_StageIteratorGeneric )
+	bool isSky = ( state != nullptr && state->isSky != false );
+
+	if ( isSky )
 	{
-		if ( isSky )
+		if ( tess.stageIteratorFunc == &Tess_StageIteratorGeneric )
 		{
 			tess.stageIteratorFunc = &Tess_StageIteratorSky;
 			tess.stageIteratorFunc2 = &Tess_StageIteratorGeneric;
 		}
-	}
-	else if ( tess.stageIteratorFunc == &Tess_StageIteratorDepthFill )
-	{
-		if ( isSky )
+		else if ( tess.stageIteratorFunc == &Tess_StageIteratorDepthFill )
 		{
 			tess.stageIteratorFunc = &Tess_StageIteratorSky;
 			tess.stageIteratorFunc2 = &Tess_StageIteratorDepthFill;


slipher:
This seems more like code golf than simplification... it's not easier to understand when stuff is swapped among variables multiple times.

illwieckz:
I noticed that, if I'm right, we can even remove one block, this will probably not make compiler more happy, but now code is just doing:

```c
if ( this && that && that )
{
	a = b;
	b = c;
}
```

illwieckz:
I just put the code in goldbot and the source code is not only simpler to follow, the produced code is indeed far less convoluted, the compiler is definitely more happy this way: [clang 12.0.0](https://godbolt.org/z/6nWc88vPj), [gcc 11.1](https://godbolt.org/z/b11ExGKPM), [msvc 19.29](https://godbolt.org/z/hG65TEx9x).

slipher:
The original version is by far simpler for a reader to understand.

illwieckz:
Original code was:

```c
	if ( tess.stageIteratorFunc == &Tess_StageIteratorGeneric )
	{
		if ( isSky )
		{
			tess.stageIteratorFunc = &Tess_StageIteratorSky;
			tess.stageIteratorFunc2 = &Tess_StageIteratorGeneric;
		}
	}
	else if ( tess.stageIteratorFunc == &Tess_StageIteratorDepthFill )
	{
		if ( isSky )
		{
			tess.stageIteratorFunc = &Tess_StageIteratorSky;
			tess.stageIteratorFunc2 = &Tess_StageIteratorDepthFill;
		}
	}
```

It then looked obvious to me this would be simpler:

```c
	if ( isSky )
	{
		if ( tess.stageIteratorFunc == &Tess_StageIteratorGeneric )
		{
			tess.stageIteratorFunc = &Tess_StageIteratorSky;
			tess.stageIteratorFunc2 = &Tess_StageIteratorGeneric;
		}
		else if ( tess.stageIteratorFunc == &Tess_StageIteratorDepthFill )
		{
			tess.stageIteratorFunc = &Tess_StageIteratorSky;
			tess.stageIteratorFunc2 = &Tess_StageIteratorDepthFill;
		}
	}
```

Then given it would be exactly the same as:

```c
	if ( isSky )
	{
		if ( tess.stageIteratorFunc == &Tess_StageIteratorGeneric )
		{
			tess.stageIteratorFunc2 = &Tess_StageIteratorGeneric;
			tess.stageIteratorFunc = &Tess_StageIteratorSky;
		}
		else if ( tess.stageIteratorFunc == &Tess_StageIteratorDepthFill )
		{
			tess.stageIteratorFunc2 = &Tess_StageIteratorDepthFill;
			tess.stageIteratorFunc = &Tess_StageIteratorSky;
		}
	}
```

It would be far simpler like this:

```c
	if ( isSky )
	{
		if ( tess.stageIteratorFunc == &Tess_StageIteratorGeneric
			|| tess.stageIteratorFunc == &Tess_StageIteratorDepthFill )
		{
			tess.stageIteratorFunc2 = tess.stageIteratorFunc;
			tess.stageIteratorFunc = &Tess_StageIteratorSky;
		}
	}
```

I have hard time to see how the first one can be simpler… This uselessly convoluted code just cried at me, especially the same test being done within all alternate blocks of a `if else` pattern…

Then there is the trick of moving the second test to an ASSERT that can save some test, and I can understand this one being controversial especially because I may miss something on this part. I don't mind that much the ASSERT trick though I wanted to get comment on its. In fact I was expecting comments on the ASSERT trick.

I still don't get how the later code in this comment would be less simple than the first one.

Especially, there is no reason to test for `tess.stageIteratorFunc` value when not processing sky at this point.

slipher:
I agree that 1->2 is an improvement. 3->4 is not really an improvement for me but not terrible either. The steps after that are making it worse.

illwieckz:
It's also possible to do it this way:

```c
	if ( isSky )
	{
		if ( tess.stageIteratorFunc == &Tess_StageIteratorGeneric )
		{
			tess.stageIteratorFunc2 = &Tess_StageIteratorGeneric;
		}
		else if ( tess.stageIteratorFunc == &Tess_StageIteratorDepthFill )
		{
			tess.stageIteratorFunc2 = &Tess_StageIteratorDepthFill;
		}
		tess.stageIteratorFunc = &Tess_StageIteratorSky;
	}
```

slipher:
Personally I might prefer

```
	if ( isSky && tess.stageIteratorFunc == &Tess_StageIteratorGeneric )
	{
		tess.stageIteratorFunc = &Tess_StageIteratorSky;
		tess.stageIteratorFunc2 = &Tess_StageIteratorGeneric;
	}
	else if ( isSky && tess.stageIteratorFunc == &Tess_StageIteratorDepthFill )
	{
		tess.stageIteratorFunc = &Tess_StageIteratorSky;
		tess.stageIteratorFunc2 = &Tess_StageIteratorDepthFill;
	}
```

To reduce nesting.

illwieckz:
With that code it looks like [gcc](https://godbolt.org/z/sh3jc9PTc) is just redoing the nesting back. So better do the nesting in source as well. I'm not against extra nesting if it makes things easier to read, but the tests were just done.

Note [clang](https://godbolt.org/z/Go3GheM5K) does something a bit weird when repeating `if isSky` on each test…

slipher:
> With that code it looks like [gcc](https://godbolt.org/z/sh3jc9PTc) is just redoing the nesting back. 

Of course. An optimizing compiler should be smart enough not to repeat a test it already knows the outcome of.

> So better do the nesting in source as well.

That does not follow. No one would want the source code to look like optimized assembly.

illwieckz:
I meant I see no reason to save a nesting by doing copy pasting if the produced code is not better.

Like I would say there is no reason to save a loop by unrolling it if the produced code is the same.

To me that kind of copy pasting to save a nesting looks to an attempt match optimized assembly (like unrolling a loop), and “_no one would want the source code to look like optimized assembly_” for nothing. And more copy paste is calling for more mistake on future edits.

illwieckz:
so I did this way:

```c
	if ( isSky )
	{
		if ( tess.stageIteratorFunc == &Tess_StageIteratorGeneric )
		{
			tess.stageIteratorFunc2 = &Tess_StageIteratorGeneric;
		}
		else if ( tess.stageIteratorFunc == &Tess_StageIteratorDepthFill )
		{
			tess.stageIteratorFunc2 = &Tess_StageIteratorDepthFill;
		}
		tess.stageIteratorFunc = &Tess_StageIteratorSky;
	}
```

slipher:
LGTM

PR #488, opened 2021-06-25
Fix download with backup URL

slipher:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1445


illwieckz:
What is the underlying bug?

slipher:
The code sending the backup URL was not updated to the new protocol which supports PAKSERVER checks.

illwieckz:
Ah, that's not good then… What is the commented out `resetconfig`?

slipher:
It's a convention for writing the names of unused arguments. To help the reader know what the argument is intended for, without triggering the unused variable warning.

illwieckz:
If it works, LGTM.

PR #489, opened 2021-07-01
gitignore: ignore removed libs/recastnavigation/ submodule

illwieckz:
When bisecting, it may be pulled again then left over.


slipher:
This change is also found within https://github.com/DaemonEngine/Daemon/pull/457, which I was coincidentally just reviewing.

Issue #470, opened 2021-05-21
Modifying cg_shadows when running dll gamecode crashes at vid_restart


illwieckz:
Setting cg_shadows with value greater than 1 crashes the game (engine quits, and the calling terminal is left in a garbage state). Options greater than 1 were riddled with bugs and were not working, and the Unvanquished graphic option UI does not make possible to set values higher than 1, but it's possible to do it on command line and some players may have an old config using them.

illwieckz:
The crash also happens with value `0`.

The crash happens on `ASSERT_NQ(cvar->proxy, nullptr);` line in `Cvar::SetLatchedValues()`:

https://github.com/DaemonEngine/Daemon/blob/075128eaaf1f9eb7a177892ac3b186b5f16ff794/src/engine/framework/CvarSystem.cpp#L479-L498

```gdb
#0  Cvar::SetLatchedValues() at …/Daemon/src/engine/framework/CvarSystem.cpp:489
#1  CL_StartHunkUsers() at …/Daemon/src/engine/client/cl_main.cpp:2659
#2  CL_Vid_Restart_f() at …/Daemon/src/engine/client/cl_main.cpp:1487
#3  ProxyCmd::Run(this = 0x5555559fe060 <myProxyCmd>, args = ) at …/Daemon/src/engine/qcommon/cmd.cpp:446
#4  Cmd::ExecuteCommand(command = , parseCvars = true, env = 0x0) at …/Daemon/src/engine/framework/CommandSystem.cpp:215
#5  Cmd::ExecuteCommandBuffer() at …/Daemon/src/engine/framework/CommandSystem.cpp:93
#6  Com_Frame() at …/Daemon/src/engine/qcommon/common.cpp:975
#7  Application::ClientApplication::Frame(this = 0x555555e90100 <Application::GetApp()::app>) at …/Daemon/src/engine/client/ClientApplication.cpp:84
#8  Application::Frame() at …/Daemon/src/engine/framework/Application.cpp:73
#9  main(argc = 43, argv = 0x7fffffffdb58) at …/Daemon/src/engine/framework/System.cpp:717
```

illwieckz:
This cvar is defined both set by Dæmon engine and Unvanquished game code.

Dæmon:

```
src/engine/renderer/tr_init.cpp
112:	cvar_t      *r_shadows;
1249:		r_shadows = ri.Cvar_Get( "cg_shadows", "1", 0 );
```

```
src/cgame/cg_local.h
[...]

PR #491, opened 2021-07-01
renderer: use legacy weight calculation for GLSL 1.20, fix #490 ref #470 #479

### LINKED ISSUE #490: Nvidia fix likely broke GLSL 1.20 GPUs
I wanted to try some optimization code and to test them on my old Thinkpad X61 Tablet on which Unvanquished was running at 0.52.0 time (slow but running), and then the engine reported failing to compile the `generic` shader:

```
Warn: Compile log:
0:44(20): error: bit-wise operations are forbidden in GLSL 1.20 (GLSL 1.30 or GLSL ES 3.00 required)
0:44(20): error: bit-wise operations are forbidden in GLSL 1.20 (GLSL 1.30 or GLSL ES 3.00 required)
```

The GPU is an Intel `GM965` one, that's one of the OpenGL 2.0 /  GLSL 1.20 ones.

This regression was probably introduced in #479 to fix #472.

On a side note: that's also why I asked for a comment telling why the code had to be written that way, because the motivation of this specific _shape_ of the code was not to implement something we want but to workaround something we don't want.

illwieckz:
Code introduced in #479 fixes bug described in #470 but broke compatibility with GLSL 1.20 as described in #490. This commit brings back the legacy weight calculation code as a fallback when running GLSL 1.20 GPUs (OpenGL 2.1). The new code is still used with newer GPUs as expected.

### DIFF
--- a/src/engine/renderer/glsl_source/vertexSkinning_vp.glsl
+++ b/src/engine/renderer/glsl_source/vertexSkinning_vp.glsl
@@ -41,8 +41,20 @@ void VertexFetch(out vec4 position,
 	const float scale = 1.0 / 256.0;
 	localBasis inLB;
 
-	ivec4 idx = 2 * ( ivec4(attr_BoneFactors) >> 8 );
-	vec4 weights = ( ivec4(attr_BoneFactors) & 0xFF ) / 255.0;
+	#if __VERSION__ > 120
+		/* This code works on GL 3 GPUs and later including the
+		Nvidia proprietary driver but does not work on GL 2 GPUs.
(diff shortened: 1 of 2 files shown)


slipher:
I think we could find still other ways to express the calculation, without entering the unpleasant state of multiple implementations.

illwieckz:
I'm not that against multiple implementations if they are rare and short, but what annoys me the most is that it adds two tests in CPU code we can avoid (on GPU code it's pre-processed anyway).

slipher:
Superseded by #498.

PR #493, opened 2021-07-02
better cmake option descriptions + faster default builds for devs and players

bmorel:
"Faster default builds" is basically, don't build stuff which are used for server and mod hosters. Is a draft for 2 reasons:

1) not sure that this "faster default builds" is really a good idea
2) I want to give a look at cmake's presets, which might improve things for packagers.


CMakeLists.txt
```diff
@@ -70,19 +70,32 @@ project(Daemon C CXX)
...
-    else()
-        set(BUILD_GAME_NACL 0)
-        set(BUILD_GAME_NACL_NEXE 0)
+        option(BUILD_GAME_NACL ON)
+        option(BUILD_GAME_NACL_NEXE ON)
     endif()
-    option(BUILD_TTY_CLIENT "Build Daemon headless client" 1)
-    option(BUILD_DUMMY_APP "Build the dummy app for daemon" 0)
+    option(BUILD_TTY_CLIENT "Engine «client» with no graphical display, used to control servers without logging in" ON)
+    option(BUILD_DUMMY_APP "Stripped-down engine client, mostly used to ease incremental porting and debugging" OFF)
```
slipher, line 98: This can't be called a client, as it cannot connect to a server.

CMakeLists.txt
```diff
@@ -71,7 +71,7 @@ project(Daemon C CXX)
 
 if(NOT DAEMON_EXTERNAL_APP)
     option(BUILD_CLIENT "Engine client executable, required to play" ON)
-    option(BUILD_SERVER "Engine server executable, required to host servers" ON)
+    option(BUILD_SERVER "Engine server executable, required to host servers" OFF)
```
slipher, line 74: I don't think server or or ttyclient should be turned off.

The nacl and native exe changes are good.

CMakeLists.txt
```diff
@@ -92,11 +105,11 @@ if(NOT DAEMON_EXTERNAL_APP)
     elseif(WIN32)
         set(USE_CURSES ON)
     else()
-	option(USE_CURSES "Compile with curses support" 1)
+        option(USE_CURSES "Enable fancy colors in terminal's output (does not detect if output is a file)" ON)
     endif()
     cmake_dependent_option(USE_SMP "Compile with support for running the renderer in a separate thread" 1 BUILD_CLIENT 0)
-    cmake_dependent_option(USE_GEOIP "Use libgeoip" 1 "BUILD_SERVER OR BUILD_CLIENT OR BUILD_TTY_CLIENT" 0)
-    option(USE_BREAKPAD "Use Breakpad crash logging" 0)
+    cmake_dependent_option(USE_GEOIP "Allows to show from which coutry a player logs from" 1 "BUILD_SERVER OR BUILD_CLIENT OR BUILD_TTY_CLIENT" 0)
```
slipher, line 111: "Server prints which country a player connects from"

CMakeLists.txt
```diff
@@ -70,19 +70,32 @@ project(Daemon C CXX)
...
+    ## About the different ways to host/play games:
+    ## Native DLL: no sandboxing, no cleaning up but debugger support. Use for dev.
+    ## NaCl exe: sandboxing, no leaks, slightly slower, hard to debug. Use for regular players.
+    ## Native exe: no sandboxing, no leaks, hard to debug. Might be used by server owners for perf.
+    ## see VirtualMachine.h for code
+
+    # can be loaded by daemon with vm.[sc]game.type 0 or 1
+    # TODO (maybe): those could probably be merged as a single option "BUILD_DISTRIBUTED_FILES"
+    option(BUILD_GAME_NACL "Build the «pexe» files, required to build the «nexe» files which are required to host servers." ON)
+    option(BUILD_GAME_NACL_NEXE "Build the «nexe» files from the «pexe» files, required to host servers." ON)
```
slipher: Why not make this one advanced? It is an extremely niche option.

CMakeLists.txt
```diff
@@ -70,19 +70,32 @@ project(Daemon C CXX)
...
+    option(BUILD_GAME_NATIVE_DLL "Build the shared library files, mostly useful for debugging changes locally." ON)
+
+    # can be loaded by daemon with vm.[sc]game.type 2
+    option(BUILD_GAME_NATIVE_EXE "Build «native executable», which might be used for better performances by server owners" ON)
+
     if (CMAKE_SYSTEM_NAME STREQUAL CMAKE_HOST_SYSTEM_NAME)
-        option(BUILD_GAME_NACL "Build PNaCl game logic" 1)
-        option(BUILD_GAME_NACL_NEXE "Build NaCl game logic" 1)
-    else()
-        set(BUILD_GAME_NACL 0)
```
slipher, line 81: I believe the option definitions are guarded by this check because the build system only downloads the external_deps package for the target system; it's not smart enough to fetch another external_deps package to get the nacl compiles for the host system.

Is there some reason you're moving the options out of there?

CMakeLists.txt
```diff
@@ -92,11 +105,11 @@ if(NOT DAEMON_EXTERNAL_APP)
     elseif(WIN32)
         set(USE_CURSES ON)
     else()
-	option(USE_CURSES "Compile with curses support" 1)
+        option(USE_CURSES "Enable fancy colors in terminal's output (does not detect if output is a file)" ON)
```
slipher, line 108: Curses is not used unless you pass the `-curses` flag at runtime. So it's not good to imply that just having this option on could break something with the output.

CMakeLists.txt
```diff
@@ -70,19 +70,32 @@ project(Daemon C CXX)
...
+    # can be loaded by daemon with vm.[sc]game.type 0 or 1
+    # TODO (maybe): those could probably be merged as a single option "BUILD_DISTRIBUTED_FILES"
+    option(BUILD_GAME_NACL "Build the «pexe» files, required to build the «nexe» files which are required to host servers." ON)
+    option(BUILD_GAME_NACL_NEXE "Build the «nexe» files from the «pexe» files, required to host servers." ON)
+
+    # can be loaded by daemon with vm.[sc]game.type 3
+    option(BUILD_GAME_NATIVE_DLL "Build the shared library files, mostly useful for debugging changes locally." ON)
+
+    # can be loaded by daemon with vm.[sc]game.type 2
+    option(BUILD_GAME_NATIVE_EXE "Build «native executable», which might be used for better performances by server owners" ON)
```
slipher: Please remove all random quotation marks :)

CMakeLists.txt
```diff
@@ -70,19 +70,32 @@ project(Daemon C CXX)
...
-        option(BUILD_GAME_NACL_NEXE "Build NaCl game logic" 1)
-    else()
-        set(BUILD_GAME_NACL 0)
-        set(BUILD_GAME_NACL_NEXE 0)
+        option(BUILD_GAME_NACL ON)
+        option(BUILD_GAME_NACL_NEXE ON)
     endif()
-    option(BUILD_TTY_CLIENT "Build Daemon headless client" 1)
-    option(BUILD_DUMMY_APP "Build the dummy app for daemon" 0)
+    option(BUILD_TTY_CLIENT "Engine «client» with no graphical display, used to control servers without logging in" ON)
```
slipher: Remove "used to control servers without logging in" - there are multiple use cases.

bmorel: Are there others? This is what I supposed, I'd like to know if there are other uses so that the wording can be improved.

CMakeLists.txt
bmorel: well, it was because I have doubts on the wording, really. I don't think it's really a native executable, that I would understand as a standalone basically.

CMakeLists.txt
bmorel: Alright.

CMakeLists.txt
bmorel: No, it just seemed cleaner to me to have stuff in one block. If there are technical reasons to keep them at this place, then sure. Well, I would still argue about the fact it would be nice to have build system not having a strong dependency on fast internet access, but this is another topic I suppose.

CMakeLists.txt
bmorel: It seems less advanced than BUILD_DUMMY_APP to me, and also looks like something that should not be "hidden" into advanced (advanced options are really a mess, I'm not sure project's options should go there honestly)

CMakeLists.txt
bmorel: this imply that this dependency is server-side only. I don't know if that's true or wrong.

CMakeLists.txt
bmorel: what are server and tty client useful for when you don't build the shared modules anyway?

CMakeLists.txt
bmorel: well, I didn't knew that. But then right, needs better wording.

CMakeLists.txt
slipher: Hosting a server with DPKs from the release.

CMakeLists.txt
slipher: Yes it is server-side only. It takes the client IP and looks it up in the database

CMakeLists.txt
slipher: There is no reason for anyone to ever use it at this point, and it is confusing to have both optoins, so the user shouldn't have to see it. (At one point it might have been a basis for developing runtime pexe->nexe translation).  I have used BUILD_DUMMY_APP several times, but never this one.

CMakeLists.txt
slipher: It is an executable, which just means it has a certain file format that can be executed. You could just as well say daemon is not standalone, since it can't run without a cgame.

CMakeLists.txt
slipher: It's useful in a way similar to the dummy app, having binaries with incrementally larger portions of functionality included.

It can be used to script some debugging/testing scenarios.

CMakeLists.txt
bmorel: building the NEXE files is probably not really "advanced" as long as one wants to build a mod. Now, I have no idea what would happen with only one of those defined. I suppose there should be a single option enabling both of them?

slipher: Building pexe without nexe is advanced. The user would have to modify BUILD_GAME_NACL_NEXE to do that. Only BUILD_GAME_NACL should be shown to the user since it is the one which has the expected effect of turning on/off both pexe and nexe (given that BUILD_GAME_NACL_EXE is on). So BUILD_GAME_NACL_NEXE is the one which should be marked as advanced.

bmorel: From "quick" testing:

* `BUILD_GAME_NACL_NEXE` requires `BUILD_GAME_NACL` to be of any use at all.
* `BUILD_GAME_NACL` generates only the `*.pexe`

So if I'm correct, what should be done is to only show `BUILD_GAME_NACL_NEXE` by default, and when it's enabled, enable `BUILD_GAME_NACL`. I would even say that being able to toggle NEXE without the other and without a warning  in cmake is a bug.

slipher: There is a cmake_dependent_option function that does what you want (makes an option not show up unless another one is enabled). It has the serious drawback that the option is hidden (in ccmake/cmake-gui) until you hit "Configure" again. This can make an option hard to find, but since this option does not need to be seen by anyone, I guess it would be OK here...

bmorel: I gave it a try after reading cmake's doc and can't really understood how I got the results I had. This would have been fine if the thing what I wanted, but it does not, by far.

I suppose the better at this point I'll just do a big option which sets both the variables to the same value. Easy and reliable, at least. I can even change the name to something which speaks more this way.

bmorel:
Now that I think about it, since some options are changed, it will probably require unvanquished and daemon CI systems to force all options to true when accepted?

CMakeLists.txt
```diff
@@ -112,21 +112,21 @@ if(NOT DAEMON_EXTERNAL_APP)
...
+option(USE_PEDANTIC "Tell the compiler to be pedantic" OFF)
+option(USE_DEBUG_OPTIMIZE "Try to optimize the debug build" ON)
+
+option(USE_PRECOMPILED_HEADER "Improve build times by using a precompiled header" ON)
 if (APPLE)
     # Turn off by default because it is reportedly broken
-    option(USE_PRECOMPILED_HEADER "Improve build times by using a precompiled header" 0)
-else()
-    option(USE_PRECOMPILED_HEADER "Improve build times by using a precompiled header" 1)
+    option(USE_PRECOMPILED_HEADER OFF)
```
slipher, line 128: I believe defining the option twice won't work. Undo the change getting rid of the "else".

CMakeLists.txt
```diff
@@ -92,11 +105,11 @@ if(NOT DAEMON_EXTERNAL_APP)
...
         set(USE_CURSES ON)
     else()
-	option(USE_CURSES "Compile with curses support" 1)
+        option(USE_CURSES "Enable fancy colors in terminal's output" ON)
     endif()
     cmake_dependent_option(USE_SMP "Compile with support for running the renderer in a separate thread" 1 BUILD_CLIENT 0)
-    cmake_dependent_option(USE_GEOIP "Use libgeoip" 1 "BUILD_SERVER OR BUILD_CLIENT OR BUILD_TTY_CLIENT" 0)
-    option(USE_BREAKPAD "Use Breakpad crash logging" 0)
+    cmake_dependent_option(USE_GEOIP "Allows server to show from which coutry a player logs from" 1 "BUILD_SERVER OR BUILD_CLIENT OR BUILD_TTY_CLIENT" 0)
+    option(USE_BREAKPAD "Allows to generate Daemon-specific crash dumps, requires specific tooling to exploit them" OFF)
```
slipher, line 115: Could we make this more concise: "Generate Daemon crash dumps (which require Breakpad tools to read)"

It's good to have short descriptions since in ccmake the part of the description text that doesn't fit on the screen is seemingly impossible to view...

slipher (commented):
I think there are more issues with the latest BUILD_NACL commit. I don't want to go into it - we've spent enough time at this point on something that doesn't matter much and will soon be removed anyway. Please just remove this commit. Also don't do any changes that define the options in different places or whatever (I think there might be some in an earlier commit) because it might break stuff.

bmorel:
> I think there are more issues with the latest BUILD_NACL commit.

Indeed, that's why I moved back the status to draft before going to sleep.


> we've spent enough time at this point on something that doesn't matter much and will soon be removed anyway. 

Exactly my thinking.

CMakeLists.txt
bmorel: Agree, I also prefer shorter descriptions when all information is there.

slipher:
Merged here: abf17b5a7e9fd952093a64f8f2b669d92129fd73. Thanks!

PR #504, opened 2021-07-10
Add Cvar Reset method

slipher:
@necessarily-equal


necessarily-equal (approved):
I find it a bit weird to have `AssertCvarRange` in `src/engine/renderer/`, but that's fine for me otherwise.

PR #509, opened 2021-07-18
some small improvements on sky code

illwieckz:
1. use `memcpy` to initialize arrays.

Using memcpy produces SSE code on both GCC, clang,
MSVC and ICC.

Alternative implementations like `for` loops or bunch
of `VectorSet` calls may not produce as good code, and
produced code may vary much more across compilers.

See [this comparison on Godbolt](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZugHZEDKqAhgDbYgCMALOXUYBVAM7YACgA8QAcgAMM8gCse5dq3qhUAUgBMAIT37yYzqiIEG1bPUwBhdOwCuAW3ogAzN3I3MAGQJ6bAA5VwAjbFIeD3IAB3QRYkt6B2c3T294xIsGAKDQlwio7hjTbHNkpiJWUiJU13cvE2wzHPoqmqI8kPDI6JNq2vr0ppFBrsCewr6SgEoTdCdSVC4ZPQ9A1GccAGptDzsx0kDgADokfdxtOQBBa7vdDxwaSZ2ANXKSUgdYgE8IHaschhHazPYAVjs92uAE4AQCQbNtOD9Ds5BCACJ7DxYgGsUHI1Ho5EY8gIgkonbcTHY3GAimo6kksk7RGEna6Gn7On4pGUzkkgm3e4AN3QBEwOzsnBqTAA1r99OhJLoATR2OhWEQdiIFQB9FyBESE3QkwkANmZO3Vmu1ut%2BBtYkmNKNN4IxFsFSOFAHZDMK5DCxlqCKgdhh6GNrRqtTtDZHCYLuXs/dDA7CYZmYeQdlns7mszm80XCwXM%2BQ0zCM6Xi2X87WG1nK9ofR6PP67gGg9ULGGI1GbbGXE6XYZ3bSUx3YbCALS1uelhflnZL/OrnPr5uB9cr%2Bd7xf7zPN1v7KddlzYFyoP4A%2B0Go05%2BMiHOJABe2HQNABT9BBPblYvK8bx1fVh2dR8RxfAh30/b8R1/JF/19D0AxuMUJSlGVSHlRVlW4NUYztUCjRNM0UUtd0c0HIiHTA0c3Q9civWhVMu2DXtwwYAdCLjEiUSTHFJzrEtlzzFM2zPbsQz7LjtWouMR0TcdkxbVEd3UsSWwktNOxhD5zDIH5/l4yMcz0c07yfdliXdayaUQyT9K%2BIzvwfPZdAs4iE0pGzGNRDx7NPSsnMM9BgKfMyPMsvjGUxOzmKQ3SQu%2BMLjIi9zPNomKqTiykAoSxzPlC8LIIyyzFJ83KiUCxLYWSlyFPAsrQIq6qyP8mrCoMlKSqa8zyuddkmVsyqCuCoqerS0r%2BpawbKWGvydny8cHJYlDOzQ8VJWlbBZQVJVJDkAjbRArLvP0BjPUo6MToG%2Bj2oorFvTuVjdLoUgdgBQJtQICc5FPHZfv2OwdktdtAaMIwhV01StyDLzRzRKrAa5QTovOnL2pRgSsTzIKu1hO74pG1EgeUtHZsRhb2TJrFk1XfGYZPZCZHmdhZHBRR3HkRR0FkKEDChkRFmWbB3K8RQiFkBRZnmOUMlOXhdAADnBH0%2BBVtWNZ9NRZF4RQXBAcE5HIbmFHIPmZEUEQQBNqWefmOBYBQRYiFiJwiEoagMBcWICE4KJwR8QgvlUQRmDYTgeH4cPRAkaQefIAB3UhWFiWRuFZ3WZE503pd52QAHkPfd7VPx2SRlfNGdzV4cN1E0KldFOORTmpCAHF9/3InF8FZkl/PZfIJBdpwKIIDZvWDaNk2zYLq2TFt8h7ZluWFaV1X1Y3rWlZ19mZA8Ln84t2QB4dslECd5A0HQLuA69iAfb9gOQGAbh34Ef2iEiG2IDCY%2BwiBBqL8DOigfYXkYIXeg7AQGJxwMOTQUc4EEFIJ8AgIpsA20TtgSQ5QParEzpQRgLRj7sAIGEVOpBfgOBwKA5exxDYyEzvMcOLAOBcD4AIBgRA45SGPiobwDctCQwMGochNtIDzDCm0LBM5C66Gti0NBVgIC%2BGGO4bwvhugFCKKoLISQGDqL0QkAx9BtG9GKM0VolRxhGO8GUCoDAOi1HMdMSxwYhiOAaKoDxEx8gWJ4PMYWSwViBOzrnOeJ8ZCV2rrXHYwBUBhnfm3T6%2BBiBkHFt4KUt9n493WNwMEAtDAGDPqvcg8tdBB33vrcghtjZ50Tpba2S8V5ZyvhAFAT9u4UCoI/HJ3S0BCO4LoOQ3gXjsG/qQX%2B/9E6APoMAuh4CbBECgTA4%2B8CNDACQebQgqCKgYKwebHBeDv50O%2BiQxOZCKHAJoQQyWDDQEsO4WwqOnDY5iD4YnFQVTNloBEcYK5EiJ4W1iDI2QciFFWOUe4VRtg7E%2BFsK43RmQTFtHhfotoSKZhQsce0WxXj0j2KUbi5xfipjIoGJ0eFvisXFCCSLUJBTwlH0abIGJNc65bE2U3FuKSIBpK%2BJknMndckfXWH3KU/zSms3mCPVgY9qCTxkDUups9j5NMXnbQea8PC8FOJrLeBqNaq2zofBp5sNWtMdpfS%2BnT%2Bn316V0l%2BoZUAjN0CUE24zJnTIAUAqhizb4QJWdA2BOzLybO2YoXZaCDnH2OagfBZziH73Nlcyh1CsB3PoQQRhzCuGMBeRwmO3DeEJ3NioGIQi/mC1EYC%2BAUjQXJFkfIxR1iVFqIJRohF/hJg6OxRi5I6LUXJFpT44lbRSXUvHTYzoo77H4rSF2mlvaAlMuCaLMJ%2B8InqrZVXDlgNEkcjkO604HgW6pJDhk/Jwr7V5MeB4QpUrl7auHqPPowLqnT14DCU4mZeC6phMrOQHhzTgnBOaZW5r57NK1efdpdq76RAfk6voIpryxD1CKbgMI9S6Bw28Jg3BzR6m4HIPUkha6fwmT/agMzzZzIWUwsBgblmrNDVG8NiDVg7JQTGzBcbcEJtOUxoh38U2KDTTczNdCiAPKY08gtkci35p4R8stigVDmjUL8nQNaAXiPrSCsFAB6eROK2jWDhZ21QWiV1uOMdkQd1mUWOdyHZilDiJ0Lu8UStteLZ3uexb46l4w530pCRwpV27WXRLiS4EQaGdhYZ/Xh94hHzRt3RPyy9YrHhZJFd08W5pH16elUPOVCrgXyw8GemrdX6sNa05%2B2pM8W5yHA8bbD5pdX/piJEjVNtYNlOq3qo129DUmv3ma/rp9n0OyVZCmbC8rXzAwVM5IIBeBAA%3D%3D).

Also inline `ClearSkyBox()`.

2. reuse `M_ROOT3` define.


illwieckz:
Maybe we can write the to-be-copied static arrays with `{ { }, { } }`…

slipher:
The Godbolt demonstration is flawed because you used two function arguments for the arrays, instead of global variables. The two arguments can *alias* each other, i.e. point to overlapping segments of memory. With global variables as in the real code, the compiler knows they cannot overlap and won't have any problems to optimize this very simple function.

illwieckz:
Ah yes, [this link is better](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZugHZEDKqAhgDbYgCMALOXUYBVAM7YACgA8QAcgAMM8gCse5dq3qhUAUgBMAIT37yYzqiIEG1bPUwBhdOwCuAW3ohdAZnI3MAGQJ6bAA5VwAjbFIQbwAHdBFiS3oHZzcPWPjEhgCg0JcIqO9TbHMkpiJWUiIU13cvE2wzCwZyyqIckPDI6JMKqpq0%2BpE%2B9sDO/O7PAEoTdCdSVC4ZPU9A1GccAGptTzth0kDgADokHdxtOQBBC8uadnRWIk2RAGsATwB9F0CRbQBWfS6f4AEX%2B%2BgAbCDyM93l9WJJfgCgX9QQDISidoYrjcVjgaGNNgA1EokUgOGJvCCbVjkMKbKbbP52HFyACcVKpdKmYM2ckZwO2ngFVNY9J5fKhnLFAM23H5guF1Ol%2Bll/PIUu5Mt08p2itFmpV2pBYuxVwAbugCJhNnZOJUmO99OhJLoINzsQB2LHXNnDR4EVCbDD0YabO4PJ7fENg4267Zelmsi6slMp6Gp1np1NZtObDM5zOJ5P5vPZ0u5kuV1NFj2gzzexN%2BiyB4Oh8OPTYueGIwwohXxhts5MAWhLo7L49zk8zm2n0LnRbZc9nY9XE7XKZrdcHPtZLmwLlQFKpr0%2BUZE0PP0ISAC9sOgaFTz/SxfXE/vD8eYWfu5ff88CDvB8n27F9uTfT1QVNS4LStG07VIB03idSRuDdHEE1NVkmwDIMGDbe4O3PGM%2BzjbQvXLGcqxTeNt0bCpmzwkMnnbSNuxIgUyIo5ceIzWjMRZXdiXMMhyUpTsfmhPRwVPL4fnFfkFONcDB1ZYTSTEp9JO2XQZNhYiZQlDEZU8eUVMTdTRPQL8rx0vSz3kmU5RBJS%2B3MrDLLJazxNs6TZIMlVnOMlVTOUgSPJJKybP/Pz9PYwzFISsKIKEyKvOihEpN0/z4pVIzURCszwtSkT0p8mLsrihEeSCgreSKlLk08zTOwq%2By4WqpzEsK5KG1rQTYOtW1sHtR1nTkdDPVUuhSE2KlAieAh%2BzkTFNiWnY7E2SF6zWowjBNXdyNU5N/McvLurWnUhW/OToy6lyZXW0jrozYrkxOqqe3qh6VSezjrpyzrAouv7%2B2nN6h36yCZBmdhZD%2BRR3HkRR0FkZkDH2kQ5gWbAdM8bhFCIWQFCmGYXmibgjl4XQAA4/g9PhafpxmPTUWReEUFwQA9GmjmZ6m6YZgX6fIJGFHIVGZEUEQQDkcgieRmY4FgFA5iIGInCIShqAwFwYgITgoj%2BHxCFJVRBGYNhOB4fgLdECRpGR8gAHdSFYGJZAJ2H4cR4mUdkAB5TWNaeB9NkkGnwWHcFeCDdRNFlXQjjkI45QgBw9YNyI8b%2BKZCb90nyCQEacCiN02ZkDnyC53RKeFoWmaF7wxf9qWTFl%2BWC7JimqcbxnBZZivPF9p3JfzxX1UQZXkDQdBM8N7WIF1/XDZAYBuA3gQDaISIZYgMI/dpQJKjeT3FF1/dGAD%2Bh2FPp2cC7TRrfvghSBJAgzWwGWnewSQSk1pYBNKCMEaIfdgBAwhu1IG8BwOAz7ywOFzGQXsBAMEthwLgfBUHCDEFIQ%2BKhuBqA0FoPaBg1AQJlpAGY1lmghlkMOAOuhpaNHflYCAvgBjuEIb4DoeQCiqDiAkWhnCBGZFobwroURCHFFKC0EYIjpEsNkfQVoVQJETCkb0NoCitFqLGHw7o3AZhY3mIsHg3sZAI1FofSWEco4x02MAVAgYN6pzmvgYgZA8aEJtHPFe2cVhGJtKQ/Q48Sbd10MbOGldOYgD%2BHLFuEtZDSw7grEmk94AQBQMvLOFAqBLz8bktA8d166DkIQ/E7Ad6kD3gfJ2YRj7QPgRfGwRBr630Pg/Yhz9xaEDfqUT%2B39xa/3/jveBC1QFO3AZAk%2BsDAGE0QWfGYFsWAYJttgog9s8FOxUFE4haAQnkLCJQ8uNCkjfwYUwhoTQkjWFsDonh%2BjJGiKEUkHRgisj0HUfwxRNy5HaMcLUVQMjaGqNGLkZ50j5GArSFCto3zDHGOxmYoxFcrGJNsZHaOsd1jEMTsnNxEAPGkm8dCDO/jZorFzsEjGBgwkwxmMXVgpdqAWKrlzeJ1jR7JPbnLNJDLyDk08LwI4fd679yibIYeXLxZj07hPaeirsmFIXvknJq8AyoDKbXTwctKnVNqYfBp9AT7NLnpfNpN8769IPN0pYvTX7v0GYfEZqAAHjJAdE8W0yoEwKwPMhBBAkEoJWVbTBts0FbMduLFQ3gSkHNpcYaZJzqExFoRcxhzC/nuHYfcmFXCfC2ARZoj5wj80vM%2BcW4FSjQXQtSAWkFZQRhVrhf0ctrbwXjB%2BUi0xmCLHopsbIOx2K1rOM2Nqymnhk7uNNl4wJZKVUBK8NMGlhg6XyvCUXEu3Ry7RPZSAXgrIjgpl4MK1kNM5CeHBH8P44IaYytbikvlBcMkq1nvPSIi91XdDNEeGIHwzTcFZB8XQwHCRMG4OCD43A5AfEkDHLeVTd7UDqeLY1prkHn3Na09p1rFBdKfva/DjqBlfxdX/N1YzMPAJ3l6xQPrZn%2BvgUQRZmHlloNWdbLBdtcHRsUCocERDNAJrXUmih8BU3ppAAAekYdc1hOaOHtsLf4J5GiK1lvrRppILb5PKLBToxt/y9EQvUx2wzza1PdtmL28xaKR6yqHQ4lwIhf2bEA0e0DRIIPglTnyIls7KVeB8eS3JeNwQMnRqJ%2BlhcmUsvLkKqdnhkspdS6lwTe7YnxOTnIW98SgPgmFae5ug624y2fYrbuwrRUD3FcLIeDnH0boFdEq5GKeX8sLp/GptzeBAA%3D) (also updated in first post).

illwieckz:
I also just modified a weird loop that did:

```c
for( int i = 0; i < 6; i++ ) {
  if ( i == 5 ) {
    continue;
  }
  …
}
```

to:

```c
for( int i = 0; i < 5; i++ ) {
  …
}
```

With thunder map and `/setviewpos 2315 3145 903 128 0` I go from ~225 fps to ~230 fps.

slipher:
> 
> 
> I also just modified a weird loop that did:
> 
> ```c
> for( int i = 0; i < 6; i++ ) {
>   if ( i == 5 ) {
>     continue;
>   }
>   …
> }
> ```
> 
> to:
> 
> ```c
> for( int i = 0; i < 5; i++ ) {
>   …
> }
> ```
> 
> With thunder map and `/setviewpos 2315 3145 903 128 0` I go from ~225 fps to ~230 fps.

There is no way this makes any difference in an optimizing compiler. FPS differences are just noise.

illwieckz:
well, at least this looks better

slipher:
> 
> 
> well, at least this looks better

OK. The ClearSkyBox one looks worse though :stuck_out_tongue:

illwieckz:
> OK. The ClearSkyBox one looks worse though

The human brain even don't have to check for the range or parse the various things in the `a[n][i] = b[n][i] = c` form when reading the `memcpy` variant…

slipher:
> The human brain even doesn't have to check for the loop range or parse the various things in the `a[n][i] = b[n][i] = c` form when reading the `memcpy` variant…

It does. You have to check that the array being copied from is the correct size and correct type, which is more annoying to me.

In any case it's annoying to see commits just changing style things which would be a 50/50 on whether people prefer one or the other...

illwieckz:
> It does. You have to check that the array being copied from is the correct size and correct type, which is more annoying to me.

This is global variable problem, not `memcpy` problem. And because of that global variable problem you can't know if the former loop is iterating all the expected component as well. There is no way to know the global variables have `[2][6]` size by looking at that function. The loop may be iterating outside the array boundaries, looking at that functions can't help to figure it out. I see no difference on that topic there.

Anyway, that `memcpy` thing seems to produce better code and the “cost” of its “look” does not seem that bad. Though maybe we would want to use a `{ {, , , }, {, , , } }` syntax just to give a hint about the dimensions of the arrays.

In fact I would prefer to use the `VectorCopy` variant, but gcc just does not optimize that unfortunately (clang does well on the contrary). This may change if one day we a library with explicit SIMD calls for such operations.

Also gcc does not unroll the current loop variant with `-O2`. It does it with `-O3`, and translate it to SSE code but the `DebWithRelInfo` profile we use sets `-O2`. MSVC unrolls the loop but don't turns it into SSE code even with `/Ox` (I'm less concerned by that since we don't build releases with it).

slipher:
If you want to do it in a way that really doesn't require reading fine details, how about

```
for (auto& a : mins)
  for (float& b: a)
   b = 9999f;
```

illwieckz:
> ```c++
> for (auto& a : mins)
>   for (float& b: a)
>    b = 9999f;
> ```

That would be super cool if we had C++ compilers!

Joke aside, [this looks too much modern for compilers in 2021](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZugHZEDKqAhgDbYgCMALOXUYBVAM7YACgA8QAcgAMM8gCse5dq3qhUAUgBMAIT37yYzqiIEG1bPUwBhdOwCuAW3qqbmADIF62AHKuAEbYpCAA7OQADugixJb0Ds5uqjFxFgw%2BfoEuIWGRptjmCUxErKRESa7u3CbYZhn0peVEWQHBoREmZRVVKbUiPa2%2B7bmd4QCUJuhOpKhcMnoAzAQ0ANRy2nIAgstRpKzALqxrAOJ2dmtRTiJIAProUY0iW7u6S/uHx2cXa48WLgIAC9sGsIHpdE56KRHOwALTsdCPF66XQTV7LTyrDE7GiI1hENYiADWAE87oD6C8AKz6XTaakAEQZ%2BgAbAzGeQiWSKaxJDS6RyWeymdoloYdq8AG7oAiYNZ2TjlJhk/ToSQQdGS8IS3ZyACcdFIYLWrCcJD0rNNaxA3PJlJEay1ev12h1ONdBqNJrx6AJlrWQVtJ2dW09rvdkoNYbDQTWYsZa31ydd4o9MfCzKjEazepd3ogpvN6ADJ1tJPtfMdoejkZdYYLa19/t0VqD1pr4bdupj0YNcYTazhKdTPb7OfTmYxmZkU3Ysmpinc8kU6FkdiMRiJMzmoOW3EURFkCgmU2JICW3AAdLxdAAOanhPj3x/PyLzmS8RQuCJ3q%2Bv28HyfQDH3IZcFHINcZEUEQQDkcgjxXKY4FgFAZiIa4iEoagMBcKICE4MJqXIbBCBIMJakEZg2E4Hh%2BCo0QJGkFdyAAdwOKJZAPOcFyXY9V1kAB5c1ML%2BdZJDvVk4VZXg1lQdRNDWbhdCvOQr24MEHDwgjQnjd5qQmQ9%2BNPcgkGwVgcDCTU1FkL9yB/XRrxA4CX2ApYwP4yDZBguCEOMs8Lyc1znyAt8bJkJY%2BJYqCjKQ8gUIQeAIBQXD8MI7CIFSnSwmAbg8oEAiiFCWCICDFigl8cpSS4xRcJcGwiEE%2Bh2GqlicGOTRaLaghSCKCwpWwWCWOwSQinNBYD0oRh6k89gCCCA5SFJBwcBqhDSAIH8ZG4gQGGojguD4XbhDEKRPJUWoFK0TcDDUebYMgKZ/gSIa4UE3QYPqPqEmsWw%2BhqEjbDaHI8lSWJ4gYf6wfSBJgY6Ci6gaEohihgYvuKBhmgqOGxgRwYWlR7oWhx0HuCmEQd3mHgeJkRcPOi2QJKkmS1mAVBUCU68NIgfBiDIPTLy5LS0t0/cJgVG79Fik8At0YiPzsn9byvcJ9WU7g5F0Xg73Ce973piCYpMXzEJPeLEGStB0G09KqEy62RZyvLKMK4rqDKiCKvoKq1rqhqmpazz2o0YAuogwheuKAahogkaxqKtbfCKj8ILmhaqpWibDw2radqolgDro46iEYs6WJUXQ1BDtBJbuoIHus56GFewTIsR76rAgTxCc8EnOlqNIIcSRxqmhoe%2B7x9HGixyoR/6duMaaIYJ9UfHejngG1%2BGbJ4ep6ZZipsnwrp8CBJkJnpNktmOby9SwV58iBdqBUHeygW0QlgwjGl2cpjMizOjWQVt%2BEA7wrz6jkOEVkSx9ThCWLwDWut9TEVPl5aCxt4Km1nObJKKVX62xwvgzo8kQ7KTkC7dgRVSAlQ9ooL2Pttq1WtvVRgAdWrh2wB1UOCxw49W%2BtHTycdUDjUTtNFOig06LWWlgLO61No1SmPnGih16J7VLsxCCKh5bVx0F/W6acG5PSeC9WQb0PoL0aL9ewG8PBAxGCDfu0RwaNEJoPRoK80ZI0xijGxniO5L2JvY3eAwfHJE3svIJuM94UwPodGmJ9PJQQvizEhillKqTvjzMi/N9xCyIcaZYBlP6GAMD/Ey/9LLUBporC8SwrxLHIdwakchyFPhkrwdyqCjawUwf5cg554FXmCs5EK8tZBty6d5PycUEqzLwTbUIGUsqERAAQdmuhNbcAaQVShbtSqeXoUtX2zD/bNXYYoYOnUeEXL4VHQagjRrCITowqaydZrzSkZnNaRAc4KOLgXWiR0GKnQ0YoFQ7kro1z0cYAx8AjHPFMa3T6Xj3Bdz%2Br4wG3hImkycTDSGGK3Gw2xY4woi8Z6E1JdPCJO8okhIJhireHjyaUzicfKKhtGaSUvmsNZHMNmOXqape%2B2SCnvGfsLN%2Bywljiw3NCspf9zKVKAbZEBHSbxLEvKrXWSxmnvDkKyA2Z8fK9JmRbeZjsln5JAFKVAUQoh3ClNwfUdxdDOoAGpMG4KyO4Gs7iSBkjsqhNCDmVSOS8v2rCzlB04SHMONzI79XucNR5IiXlJxmixSRGcZHfN%2BYwxRe0AUqOLuo86IADWQt0SUmF904WQWMc3EAAB6QSkgLE/TRdYsJtisU0pxYS/F3aB7OKJX2klU9kb0qHe27xgSx141CaPOl2NiUUWZbEveH4EkM3PizFwIgbVrEdfqIZ%2Bo1geq9epOQwq%2BaisFi/BZd7WQyslvK0yirAEBQ1fUjVv6/1/orSq%2ByIBqSsn/KyVkd5%2BVQN4LAp8hq0HGumTLfpF5eBDNCiMkC4UJmJKmVgkyH5zGTPQQRqYA1qE/V4EAA%3D%3D%3D).

Only clang knows how to optimize this with `-O2`. GCC starts to optimize this with `-O3` but `RelWithDebInfo` sets `-O2`. With `-O2` GCC keeps the 2×2 loops, and if using ”`unroll-loops`” optimization with O2 it does not produce SSE code. MSVC keeps the 2×2 loops as well, both with `/O2` and `/Ox`. The very fun part is ICC, I'm glad we don't use that, it produces ~150 instructions with 20+ conditional jumps…

I would really appreciate this kind of writing, but compilers look to not be ready for that yet.

This is also why I'm looking for third party libraries that would just pick the best case for us, like calling `fill(sky_mins, 9999)` and don't worry.

For completeness I also tried `std::fill`:

```c++
#include <algorithm>
#include <iostream>

float sky_mins[2][6], sky_maxs[2][6];

void ClearSkyBox()
{
	std::fill( sky_mins[0], sky_maxs[0]+6, 9999 );
	std::fill( sky_mins[1], sky_maxs[0]+6, 9999 );
	std::fill( sky_maxs[0], sky_maxs[0]+6, -9999 );
	std::fill( sky_maxs[1], sky_maxs[0]+6, -9999 );
}
```

And unless I did a mistake the produced code is just [totally unpredictable given the compiler](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZugHZEDKqAhgDbYgCMALOXUYBVAM7YACgA8QAcgAMM8gCse5dq3qhUAUgBMAIT37yYzqiIEG1bPUwBhdOwCuAW3ohd5G5gAyBetgAcq4ARtikPNzkAA7oIsSW9A7Obh4xcQkMfgHBLmER3FGm2OaJTESspETJru6exaUM5ZVE2UGh4ZEmFVU1qfU9rf7teZ2FAJQm6E6kqFwyegDM/qjOOADU2ot2HMBkxEguW7jacgCCSytr2JvbliJEpNisR4sn56dnNOzorETrIgA1gBPAD6Ln8Im0AFZ9LoYQARGH6ABsiPIAJB4NYkihsPh0KRsLRhK2hg%2B5wAbugCJh1nZOJUmCD9OhJBBxp9tAB2ckXOQATgemBAIBoBHY7AgmLBEPoeP0cnRMuxuORStJBhRGIFuoF605iz5pyFRBFYolUpVcoV3GVQNlOIVGqRWp1eoNZK5guFovFkulDtVzvtWJcTvViKM2vWAFo9frDcafWa/ZbA2GI7C7YSMUHw2rYS7oxj4x6k1zuQiZJN2LJoYp3PJFOhZHYjEYAdNZjcltxFERZApxpNASBFtwAHS8XQADmh3L4c4XS%2B5alkvEULhA3Nnk5XM/ni8PC/ITYU5FbMkUIhAcnIg%2BbkzgsBQ0yI0ScREo1AwLmiEqdNCXiECQBQCAwzBsJwPD8IIRCiBI0jNuQADupCsNEsj9rW9aNkOLayAA8l%2Bn7/OgNDrJIs4orGKK8OsqwaMA6zcLok5yJO3DrBADj/oBpCbLoizQuMA4ESO5BIM8OARBy64yJu5DbroU4nsey7HosZ4EZesg3neD4SaO45qZpS5HquCmLPhKFXuJT7kC%2BCDwBAKB/gBnAUFQEAeQJIDAIUUT%2BkQ4S3hAIS6SE/iVMC2GKH%2BLg2EQRH0OwcUoTg4aaDBmUEE8pSUtgt4odgkglF%2B8z9pQjDYHWKHsAQIQYaQwIODg8UPqQBDbjIOEQYwLAcFwfADQhYhSLpKhROomhoB2BhqE1t6QJM6DRBYDAlbGRG6DedUlJt7gQN4fTuFE3htLk%2BSqLE8RHWdt0ZEdV0dOBDRHc0vSOLUqgfWUgyvaM72DI9RSA8M11jJMIjdnMPC4TIDY6XZsjUbR9HrMAqCoKxU7cRA%2BDEGQQkThifGeeEpPcOM9ILfoDnDiZujAfVSnbtCnECtCiy6HIKLCQK3AorOhQoxe9kmIZj7Dk5iBuWg6D8V5P6%2BUrlMRExmhsXIwUSqFpDhZFKHRfQsWdYlyWpelulZcxuUXoQBUWEVJUXmVFWhZ1/ihfVF6Nc1sXtVVA7db1/XwUNMGjfBiGTShKis8x80GEYS0hCt8nrUd21ETZJgHY0x2nT9qQXbYQM3VEd2ZEkpfnek92JJXYwF2Yn2g/Xf2Fx3LQtyDLRg90feQ29CNTDM8M0wpyPnoRMjo3RDFayxbEcVxPFE2B1Pk%2BrAmk6JdOpwYjM1pM0msLJ1CI%2BzIC8AKHG8Ci3ICos3IonIs4TrOAri/PBn3hlmfcgY5Fi8EnOZdSFlWayHznPPS14jKOWcig9ye8VY%2BT8l5EABAcZ81Uose8IUwrUGNheU25s%2BoJSVklRg1sMqO2wNlYADtFBO0OgQV2ukPaoEqt7WqftFABxam1LAIcuo9XipMSO0ERpwUgnHZCF4VDaVmloem6dM5rQ2okXO%2Bd/pWBOrYIel1R7Ayek3BgQ8a4vTMVXNuHCmidxSA3Ax9AvpDByGPcGg8u4%2BKqP3cesNJ4jURrPXSV5F6Y1wbjfBU5FgcU3qBEmfZd7KypksRYtN2zHwZkgpmUkZKdHkmzLcIBOaTi/u/FEH9Zw/xRNCAULM/4IIAfkmsctXJoPSd5X86DOiUlQNEaIoJKTcAFKCXQEyABqTBhagm4HIUEkh6ICH1iQiKUUYqtQtjQq2aUGFsKYfbeYjt8ocK4aVcqvCvZUJqr7XSwig5iM6o8SRVDpGQSjnIsaiipogG1GolOhhFoBy0ZeHRW0QAAHoiKSAcUXawxi/FeArnY1uNjEjWOes3dF70e4A18S47u7dCUBLxX9Zxv1/GeJGFXGGcNQkz1shLNGmMXAiEGesMZD8pnrFmcLLicgknE0Eqk%2Bk/SxXCRRNk%2Bmp9JIXyvvJUBCTFhqvVRqjVgKNxlPvlxAW6qhaLLAYKFpktbyAOMiA8c4DIF2vnNZFl/92mSXqntM1%2BkXWTCKobRId8gA%3D%3D%3D).

Anyway, compiler writers seems to care about making memcpy as faster as they can, this looks to be the first class array initializer out there. Also this is a pattern compilers can very easily detect (`memcpy` call is obvious, not like all array idiomatic variant that exist). Because of that I would vote for recommending using `memcpy` to initialize arrays by default.

For example Wasm people [are already planning to implement `memcpy` the best way](https://github.com/WebAssembly/proposals/issues/18), for skybox that does not count since we run native code, but we may want to use `memcpy` in game code when that makes sens for that purpose, it is likely they will optimize that before any other copy variant.

One good argument that may come up in the future to go against `memcpy` is if we use a third-party optimized library and we better want to use a provided function to be sure subsequent calls of other provided functions blend well. But we would still be able to rewrite those `memcpy` calls if such thing happen.

Extra question: is there some generic `memset` function but for every type whatever the array depth and size?

necessarily-equal:
Is there any advantage about using memcpy over this?
```C++
	sky_min = {
		{ -9999, -9999, -9999, -9999, -9999, -9999 },
		{ -9999, -9999, -9999, -9999, -9999, -9999 }
	};
```

illwieckz:
Unless I'm wrong, [I get](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZugHZEDKqAhgDbYgCMALOXUYBVAM7YACgA8QAcgAMM8gCse5dq3qhUAUgBMAIT37yYzqiIEG1bPUwBhdOwCuAW3ogAbORuYAMgXpsADlXACNsUhAADnIAB3QRYkt6B2c3TziEpIZ/QJCXcMiY02xzZKYiVlIiVNd3LxKyhgqqolzgsIjok0rq2vSG3raAjoKuqIBKE3QnUlQuGT0AZgDUZxwAam0luxEiUgDgADokbdxtOQBBC8uadnRWIg2RAGsATwB9FwCRbQBWfS6f4AEX%2B%2Bg8IPIz3eX1Ykl%2BAKBf1BAIhyO2hiuN2WOBoIw2ADVSiRSA5Ym8IBtWORQhsJls/nZsXIAJyUym0iZgjZyBnArZLfmU1h07m8yEc0UAjbcPkCoVUqX6GV88iSrnS3Ry7YKkUa5VakGirFY3QrGgym4AN3QBEwGzsnCqTHe%2BnQkiWEC5WIA7JjrqzXp9vvQRPKtn7mSyLtG/RsWQmWVDE0n44nk%2Bm0wmMwmI8DyFGY9o4ymc6nS1ny5mU3nCz7QUt/VGg7D4eHi03WUW4wBaCt9zMD7MbIep0dQ0d5gsm6Nd3v9heDxfDyfF0EztcY7H17FmnwEGjMy42u0Op2kF1vN2SXRe7edll7R4EVAbDChp53B5PEMIg187kPG1QUIwfbtlQrSDq2g4cazXacA1nWMIJgqtYNQstaw3HdGyPGMnwsV93z2DYv0eDYXDhP8NkNdFpSAo0dVAws52VccR2XMdOInOD6wQotWI4pdhJXbihNzTdsIbTtEJcbAXFQclKRbX8oVU54CAAL2wdAaEpX86VFXCZzkhSlOhYMqLUqyNO03T9KowyuWM64dxNY9bXtR1sGdV13W4O9fQfAiXzfBgSLIn8fjBRiQI7Ssy0wzcXPwypCLCj9SPucjKPhGLkXbecRK4xNaxSmciXMMgyQpCifihPQPBU6LpXFOjlTa/lnIfSqSRq/T6q2XQmphX8xQA6Uljlbqo166r0HM9TGua0NuVlEFxqNGaKuJebFsG5bRpa5V1vajYpq2rcdqq0kFtq3KRAa4aVKozazs6oyet227Fpsw7LLy1qJuVC6Cu2xC5p%2B%2B6/ue0bXulU6UQ66aroh77%2BoomGRoB6jEe5UGutRySAw809vN8q93TkQLXIfOhSA2SkAieAhwzkDENlZ7Y7A2CFG05owjGNRCOxYx8jtWoGNulLmCqYlbceBzngP5FMia7QM4cB5HpeVWX%2BXlrXFd15XYv5Ud1djddXOBGQpnYWQ/kUdx5EUdBZCZAwhZEGY5mwIalm4RQiFkBQJimF4QEDo5eF0KI/h9Pg44TpOfTUWReEUFwQB9KIjhT2P48TwuE/IF2FHId2ZEUEQQDkcgQ9dqY4FgFAZiIWInCIShqAwFxYgIThIj%2BbxCBJVRBGYNhOB4fhJ9ECRpFd8gAHdSFYWJZCD%2B3Hed0O3dkAB5LvO6eXSNkkKIPB7DxeDfdRNBlXQjjkI5ZQgBx%2B8HiIA7%2BCZg/3uHcgSAfI4EiF6dOMhM7kGzrobgMdk7F0QXwJYZd96V1kDXOuDdAERyjvAkuSCi6p0gUsPey8q4AKbmqRALdkBoHQF/IePcIB9wHkPEAwBuDcIEIPIgERa4QFCOg0IAQqhvC3ooPuclGCH3oOwCRy8cCUU0DPJRBBSDEgIFabAtdl7YEkKULuCwg6UEYNgB2y92AEFCOvUgbwHA4EkQ3A42cZDbwEAwKeHAuB8E8cIMQUh0EqG4GoDQWhBYGDUDY2ukApgLQsOFWQPZD66BrhYrRVgIA%2BH6O4UJPh2j5EKKoeIiREkpEcHUEpWRymFM6JEUJjRyktD6JU9IjSMlNHoC04YeR6mqCfK0tIeSeitDqWMBpUwfazHmDwHeMgnZoIobIS%2B19b4bGAKgV83C36M3wMQMgAdQkOkYewn%2ByxuD0k9oYAwVCw54N0CPSx0Ds5/HruXA%2B1cTDYMbmHGh8AIAoDYd/CgVBWGnJBWgB%2BXDdByFCXidg/DSCCOEcvUR9BxHOOkTYIgciFHoOUeEtRFdCCaLKDovRFcDFGP4c45mFj0HWNseIxxJjg6uMkVMSeLAfGz38UQBeQTl4qCeeEtAkTjBMtiRAhJyQ9EpLSSYTp5TrC2FyaoApIwildFCaU7IFThnVLKckcZxSOlmGaUMdV5rMndKGKanVoyhlVMafarV/TLnTBmb4%2BZiyPkYJkKsm%2Bd81jhKfi/XZEB9kkiOVCT%2BZyGbLD/g6CVdy7ZTBAawMB1B5kvJAG8pZFdKHfPrr89N5BI5LF4EcZBhCk7x1IeQotmCcHULoe2oFELmFguBRwl8qBYVwKWPXBFSKUUiLEfYrFjCZG4vkYokl8kiULBJRorRFL0HUtQMYul5jLEVyZXYhxWA2UuIIG4jx3Lp6%2BLnl4wVS8K4qFQdC8VXsolSvgPE2I5T5WpPSRa5Iqr7BtJGZqvpEyjX6utZkY1OR3UQZtV0np0GmnlDdeBs1TqaggYGeh0YZqpm%2B1mZ6yxfr0FVyDes/tNE5BwKOEsF%2Beyx6HIuXGrt5yzRLCuam1t9zgGgK6BA55WcQC8BZEcBMvAq0siiHIJYHg/h/A8DEf1xba6lsAf81uDCmERBYb2roVpFKxA%2BFabgLIPi6AswSJg3APAfG4HID4khb68MRQI6gqKK7osxe4qRM6cV4oXYoQlqiV0hbXeS3Rm7DHbtpX5sx/D92KEPSyk9zj9jns5fynlM8/Hz0CQ%2BxQKgvAvp0G%2ByVMTP2V2/XKkAAB6VJSqANZJyTh/JtgHUNJg1B9rPXanwcw6h5oVq%2BvDbtWMwbjrBnYcNa6ybGGdWEe9XMyBZHlmBvWS4EQRmNhmfE1ZwktmPBv15FG5jiazTHPjSCgOHhuMVbTUAzN2aIGVoY0sT7X3vvfdKxnETbyX5yCU288zHgq1SdQaplt6nePlsrdW2tSOG2WLIYWz5T35mKuh18stQCdHIsA7wIAA%3D%3D%3D):

```c
<source>:11:11: error: array type 'float [2][6]' is not assignable
        sky_mins = {
        ~~~~~~~~ ^
<source>:16:11: error: array type 'float [2][6]' is not assignable
        sky_maxs = {
        ~~~~~~~~ ^
2 errors generated.
Compiler returned: 1
```

I did:

```c++
void ClearSkyBox3()
{
	sky_mins = {
		{ 9999, 9999, 9999, 9999, 9999, 9999 },
		{ 9999, 9999, 9999, 9999, 9999, 9999 }
	};

	sky_maxs = {
		{ -9999, -9999, -9999, -9999, -9999, -9999 },
		{ -9999, -9999, -9999, -9999, -9999, -9999 }
	};
}
```

necessarily-equal (approved):
Ok then, memcpy it is I guess. I have a slight preference for constexpr over const, but it shouldn't matter anyway.

illwieckz:
Any comment on this change?

```diff
- boxSize = backEnd.viewParms.zFar / 1.75F; // div sqrt(3)
+ boxSize = backEnd.viewParms.zFar / M_ROOT3;
```

I replace `1.75F` by `1.732050808f`. Though the comment made it explicit the intention was to do `sqrt(3)` and the second is a far much precise approximation for `sqrt(3)`, but the difference is not that minor. I have not seen any problems in the rendering anyway.

illwieckz:
Just for reference, about `memcpy`, that's a result of a benchmark we discussed on IRC comparing a default `memcpy` and a non-default `std::fill` tricked to get the best out of it (standard `std::fill` call would likely to result in separate operations for every sub array instead of doing stuff on the whole array at a time), this was an idea of freem.

So I run the code and recorded the execution with [orbit](https://github.com/google/orbit).

To make it easy (Orbit records an existing process), instead of timing each function I've just put the code in an infinite loop and recorded the process running for some dozen of seconds. Then I looked at how much CPU time was spent in one function or another.

The provided assembly is what displays Orbit.

```c++
#include <string.h>
#include <algorithm>

float sky_mins[2][6], sky_maxs[2][6];

__attribute__((noinline)) void ClearSkyBox1()
{
	std::fill( *sky_mins, (*sky_mins)+2*6, 9999 );
	std::fill( *sky_maxs, (*sky_maxs)+2*6, -9999 );
}

__attribute__((noinline)) void ClearSkyBox2()
{
	static const float mins[ 2 ][ 6 ] = {
		{ 9999, 9999, 9999, 9999, 9999, 9999 },
		{ 9999, 9999, 9999, 9999, 9999, 9999 }
	};

	static const float maxs[ 2 ][ 6 ] = {
		{ -9999, -9999, -9999, -9999, -9999, -9999 },
		{ -9999, -9999, -9999, -9999, -9999, -9999 }
	};

	memcpy( sky_mins, mins, sizeof( mins ) );
	memcpy( sky_maxs, maxs, sizeof( maxs ) );
}

int main()
{
	while ( true )
	{
		ClearSkyBox1();
		ClearSkyBox2();
	}
}
```

# gcc -O2

* ClearSkyBox1: 51.26%
* ClearSkyBox2: 43.13%

```asm
asm: /* ClearSkyBox1() */
Platform: X86 64 (Intel syntax)
0x55d41bc4b150:	endbr64      
0x55d41bc4b154:	lea          rax, [rip + 0x2f25]
0x55d41bc4b15b:	movss        xmm0, dword ptr [rip + 0xf2d]
0x55d41bc4b163:	lea          rdx, [rax + 0x30]
0x55d41bc4b167:	nop          word ptr [rax + rax]
0x55d41bc4b170:	movss        dword ptr [rax], xmm0
0x55d41bc4b174:	add          rax, 4
0x55d41bc4b178:	cmp          rax, rdx
0x55d41bc4b17b:	jne          0x55d41bc4b170
0x55d41bc4b17d:	lea          rax, [rip + 0x2ebc]
0x55d41bc4b184:	movss        xmm0, dword ptr [rip + 0xf08]
0x55d41bc4b18c:	lea          rdx, [rax + 0x30]
0x55d41bc4b190:	movss        dword ptr [rax], xmm0
0x55d41bc4b194:	add          rax, 4
0x55d41bc4b198:	cmp          rax, rdx
0x55d41bc4b19b:	jne          0x55d41bc4b190
0x55d41bc4b19d:	ret  
0x55d41bc4b19e:
```

```asm
asm: /* ClearSkyBox2() */
Platform: X86 64 (Intel syntax)
0x55d41bc4b1a0:	endbr64      
0x55d41bc4b1a4:	movdqa       xmm0, xmmword ptr [rip + 0xeb4]
0x55d41bc4b1ac:	movdqa       xmm1, xmmword ptr [rip + 0xebc]
0x55d41bc4b1b4:	movdqa       xmm2, xmmword ptr [rip + 0xec4]
0x55d41bc4b1bc:	movdqa       xmm3, xmmword ptr [rip + 0xe5c]
0x55d41bc4b1c4:	movdqa       xmm4, xmmword ptr [rip + 0xe64]
0x55d41bc4b1cc:	movdqa       xmm5, xmmword ptr [rip + 0xe6c]
0x55d41bc4b1d4:	movaps       xmmword ptr [rip + 0x2ea5], xmm0
0x55d41bc4b1db:	movaps       xmmword ptr [rip + 0x2eae], xmm1
0x55d41bc4b1e2:	movaps       xmmword ptr [rip + 0x2eb7], xmm2
0x55d41bc4b1e9:	movaps       xmmword ptr [rip + 0x2e50], xmm3
0x55d41bc4b1f0:	movaps       xmmword ptr [rip + 0x2e59], xmm4
0x55d41bc4b1f7:	movaps       xmmword ptr [rip + 0x2e62], xmm5
0x55d41bc4b1fe:	ret          
0x55d41bc4b1ff:
```

# gcc -O3

* ClearSkyBox1: 68.17%
* ClearSkyBox2: 22.81%

```asm
asm: /* ClearSkyBox1() */
Platform: X86 64 (Intel syntax)
0x55b02e8b1150:	endbr64      
0x55b02e8b1154:	movaps       xmm0, xmmword ptr [rip + 0xf35]
0x55b02e8b115b:	movaps       xmmword ptr [rip + 0x2f1e], xmm0
0x55b02e8b1162:	movaps       xmmword ptr [rip + 0x2f27], xmm0
0x55b02e8b1169:	movaps       xmmword ptr [rip + 0x2f30], xmm0
0x55b02e8b1170:	movaps       xmm0, xmmword ptr [rip + 0xf29]
0x55b02e8b1177:	movaps       xmmword ptr [rip + 0x2ec2], xmm0
0x55b02e8b117e:	movaps       xmmword ptr [rip + 0x2ecb], xmm0
0x55b02e8b1185:	movaps       xmmword ptr [rip + 0x2ed4], xmm0
0x55b02e8b118c:	ret          
0x55b02e8b118d:
```

```asm
asm: /* ClearSkyBox2() */
Platform: X86 64 (Intel syntax)
0x55b02e8b1190:	endbr64      
0x55b02e8b1194:	movdqa       xmm0, xmmword ptr [rip + 0xec4]
0x55b02e8b119c:	movdqa       xmm1, xmmword ptr [rip + 0xecc]
0x55b02e8b11a4:	movdqa       xmm2, xmmword ptr [rip + 0xed4]
0x55b02e8b11ac:	movdqa       xmm3, xmmword ptr [rip + 0xe6c]
0x55b02e8b11b4:	movdqa       xmm4, xmmword ptr [rip + 0xe74]
0x55b02e8b11bc:	movdqa       xmm5, xmmword ptr [rip + 0xe7c]
0x55b02e8b11c4:	movaps       xmmword ptr [rip + 0x2eb5], xmm0
0x55b02e8b11cb:	movaps       xmmword ptr [rip + 0x2ebe], xmm1
0x55b02e8b11d2:	movaps       xmmword ptr [rip + 0x2ec7], xmm2
0x55b02e8b11d9:	movaps       xmmword ptr [rip + 0x2e60], xmm3
0x55b02e8b11e0:	movaps       xmmword ptr [rip + 0x2e69], xmm4
0x55b02e8b11e7:	movaps       xmmword ptr [rip + 0x2e72], xmm5
0x55b02e8b11ee:	ret          
0x55b02e8b11ef:
```

src/engine/renderer/tr_sky.cpp
```diff
@@ -310,15 +310,20 @@ static void ClipSkyPolygon( int nump, vec3_t vecs, int stage )
 ClearSkyBox
 ==============
 */
-static void ClearSkyBox()
+inline void ClearSkyBox()
```
necessarily-equal, line 313: `static inline`, if you want the inline

necessarily-equal (changes requested):

PR #514, opened 2021-07-22
Initial support for WASM build in CMake

Kangz:


cmake/toolchain-wasi.cmake
```diff
@@ -0,0 +1,51 @@
+# Daemon BSD Source Code
```
slipher, line 1: Can you use the one at `wasi-sdk/share/cmake/wasi-sdk.cmake`?

cmake/DaemonGame.cmake
```diff
@@ -102,7 +102,50 @@ function(GAMEMODULE)
...
+        if (NOT FORK AND BUILD_GAME_WASM)
+            if (CMAKE_GENERATOR MATCHES "Visual Studio")
+                set(VM_GENERATOR "NMake Makefiles")
+            else()
+                set(VM_GENERATOR ${CMAKE_GENERATOR})
+            endif()
+            set(FORK 1 PARENT_SCOPE)
+            include(ExternalProject)
+            set(vm wasm-vms)
+            set(inherited_option_args)
```
slipher, line 115: This bit could be moved out of the nacl and wasm blocks

PR #516, opened 2021-07-23
Small things

necessarily-equal:


necessarily-equal:
Note I've updated the const fixes commit and sneaked in two other commits. This should be ready for review.

necessarily-equal:
Also, I've tested this with a the released 0.52 assets+gamelogic, and it works, but I must say I'm a bit wary as this could be an ABI break.

src/engine/qcommon/q_math.cpp
```diff
@@ -200,7 +200,7 @@ int DirToByte( vec3_t dir )
 	bestd = 0;
 	best = 0;
 
-	for ( i = 0; i < NUMVERTEXNORMALS; i++ )
+	for ( size_t i = 0; i < NUMVERTEXNORMALS; i++ )
```
slipher, line 203: change back to int?

src/engine/qcommon/q_math.cpp
```diff
@@ -559,6 +559,18 @@ void ExponentialFade( float *value, float target, float lambda, float timedelta
...
 }
 
+/*
+=============
+ExponentialFade
+
+Fades one value towards another exponentially
+=============
+/
+inline float EaseInCubic(float percentage, float from, float to)
```
slipher, line 585: what's this

necessarily-equal: another thing not meant to be commited :eyes:

PR #518, opened 2021-07-30
Drop old things

necessarily-equal:
(This needs https://github.com/Unvanquished/Unvanquished/pull/1468 to be merged in unvanquished for both to compile)


slipher:
Thanks for the cleanups

PR #520, opened 2021-08-08
FileSystem: do not look for DEPS file in pk3dir

illwieckz:
FileSystem: do not look for DEPS file in pk3dir.

The code already ignores DEPS file in pk3 archives.

I found this by attempting to implement a way for partial dpk and dpkdir to set a list of deleted files to not load from deps.


slipher:
LGTM

PR #521, opened 2021-08-08
FileSystem: ignore files listed in DELETED from delta pak dependencies

illwieckz:
FileSystem: ignore files listed in `DELETED` in delta pak dependencies

The code is expected to be only reliable for ignoring deleted files in dependencies. For example if the unvanquished_0.52.2.dpk pak lists the scripts/colors.shader file in DELETED file and has unvanquished_0.52.dpk in DEPS, the scripts/colors.shader file from unvanquished_0.52.dpk will be ignored.

But if the scripts/colors.shader file is still shipped in unvanquished_0.52.2.dpk and listed in DELETED from unvanquished_0.52.2.dpk you cannot expect the file to be ignored, it is expected the user delete it for real in the parent pak.

It only ignores deleted files from dependency paks sharing the same base name. For example if the unvanquished_0.52.2.dpk pak lists the scripts/colors.shader file in DELETED file, this file will only be ignored in older unvanquished paks, not in tex-common pak and newer tex-common pak can ship the new scripts/colors.shader file instead.

It means it's possible to not only delete a file from a repository but also to move a file from a pak repository to another pak repository and not get the older version of the file being loaded from the old pak instead of the new file from the new pak.

See https://github.com/DaemonEngine/Urcheon/pull/40 for a branch of Urcheon able to produce such `DELETED` file.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -146,6 +146,9 @@ std::error_code throw_err;
 // Dependencies file in packages
 #define PAK_DEPS_FILE "DEPS"
 
+// Deleted file list in packages
+#define PAK_DELETED_FILE "DELETED"
+
 // Whether the search paths have been initialized yet. This can be used to delay
 // writing log files until the filesystem is initialized.
 static bool isInitialized = false;
@@ -1052,11 +1055,91 @@ struct LoadedPakGuard {
 };
 static LoadedPakGuard loadedPaksGuard;
 
+// std::unordered_set uses std::hash which does not
+// hash pair of std::string.
+struct stdStringPairHasher
+{
+	std::size_t operator()(const std::pair<std::string, std::string>& p) const
+	{
+		std::hash<std::string> string_hasher;
+		return string_hasher(p.first) ^ string_hasher(p.second);
+	}
+};
+
+// List of deleted filenames per pak files.
+// first string is pak name, second string is deleted file name
+static std::unordered_set<std::pair<std::string, std::string>, stdStringPairHasher> deletedFileSet;
+
 // Map of filenames to pak files. The size_t is an offset into loadedPaks and
 // the offset_t is the position within the zip archive (unused for PAK_DIR).
 static std::unordered_map<std::string, std::pair<uint32_t, offset_t>> fileMap;
 
 #ifndef BUILD_VM
+/* Parse the deleted file list file of a package.
+
+Each line of the file is the pak basename a file must
+be deleted from, followed by the deleted file name.
+
+Example:
+  map-chasm chasm-level0.navMesh
+
+This would prevent loading any chasm-level0.navMesh file
+from any dependencies with map-chasm basename.
+
+Current pak name is known as parent.name so it would be
+possible to implement a format variant that assumes pak name
+to be the same as the one containing the DELETED file. */
+static void ParseDeleted(const PakInfo& parent, Str::StringRef deletedData)
+{
+	auto lineStart = deletedData.begin();
+	int line = 0;
+	while (lineStart != deletedData.end()) {
+		// Get the end of the line or the end of the file.
+		line++;
+		auto lineEnd = std::find(lineStart, deletedData.end(), '\n');
+
+		// Skip spaces.
+		while (lineStart != lineEnd && Str::cisspace(*lineStart)) {
+			lineStart++;
+		}
+
+		if (lineStart == lineEnd) {
+			lineStart = lineEnd == deletedData.end() ? lineEnd : lineEnd + 1;
+			continue;
+		}
+
+		// Read the pak name.
+		std::string pakName;
+		while (lineStart != lineEnd && !Str::cisspace(*lineStart)) {
+			pakName.push_back(*lineStart++);
+		}
+
+		// Skip spaces.
+		while (lineStart != lineEnd && Str::cisspace(*lineStart)) {
+			lineStart++;
+		}
+
+		if (lineStart != lineEnd) {
+			// Read the file name.
+			std::string fileName;
+
+			while (lineStart != lineEnd) {
+				fileName.push_back(*lineStart++);
+			}
+
+			// If this is the end of the line, add the path to the deleted file list.
+			if (lineStart == lineEnd) {
+				fsLogs.Debug("Deleted file %s from %s in %s", fileName, pakName, parent.path);
+				deletedFileSet.emplace(std::pair<std::string, std::string>(pakName, fileName));
+				continue;
+			}
+		}
+
+		fsLogs.Warn("Invalid deleted file list specification on line %d in %s", line, Path::Build(parent.path, PAK_DELETED_FILE));
+		lineStart = lineEnd == deletedData.end() ? lineEnd : lineEnd + 1;
+	}
+}
+
 // Parse the dependencies file of a package
 // Each line of the dependencies file is a name followed by an optional version
 static void ParseDeps(const PakInfo& parent, Str::StringRef depsData, std::error_code& err)
@@ -1130,9 +1213,34 @@ static void ParseDeps(const PakInfo& parent, Str::StringRef depsData, std::error
 	}
 }
 
+/* The code is expected to be only reliable for ignoring deleted files
+in dependencies. For example if the unvanquished_0.52.2.dpk pak lists the
+scripts/colors.shader file in the DELETED file and has unvanquished_0.52.1.dpk
+in DEPS, the scripts/colors.shader file from unvanquished_0.52.1.dpk will
+be ignored.
+
+If the scripts/colors.shader file is still shipped in unvanquished_0.52.2.dpk
+and listed in the DELETED file from unvanquished_0.52.2.dpk you cannot expect
+the file to be ignored, it is expected the user deletes the file for real in
+the parent pak. Making sure the DELETED file applies on files of the same pak
+would increase code complexity while packager can just delete the file, in
+some cases the file may be ignored (if DELETED file is read first) but you
+must not rely on it and not expect it.
+
+This feature means it's possible to only delete a file from a repository but
+also to move a file from a pak repository to another pak repository and not
+get the older version of the file being loaded from the old pak instead of
+the new file from the new pak. */
+static bool FileIsDeleted(const PakInfo& pak, Str::StringRef filename)
+{
+	return deletedFileSet.find(std::pair<std::string, std::string>(pak.name, filename)) != deletedFileSet.end();
+}
+
 static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expectedChecksum, Str::StringRef pathPrefix, std::error_code& err)
 {
 	Util::optional<uint32_t> realChecksum;
+	bool hasDeleted = false;
+	offset_t deletedOffset = 0;
 	bool hasDeps = false;
 	offset_t depsOffset = 0;
 	ZipArchive zipFile;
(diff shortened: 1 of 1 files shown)


slipher:
Why should there be some rule about dependencies? With non-deleted files, the first file loaded for a given name takes precedence, regardless of dependencies. It seems simpler and more consistent for the first thing loaded for a given file to take precedence, whether that thing is a regular file or a tombstone.

illwieckz:
Let's take this example.

I created some basic color textures in `tex-common_src.dpkdir`: plain white, plain red, plain green… many maps just redo custom variants of them, also the pixmaps can be used for blending. Alongside the images, I added the related materials in a file named `scripts/colors.shader`.

But… unvanquished dpk already had a `scripts/colors.shader` file I named that way long time ago because I lacked imagination. I can rename it in unvanquished dpk but then the future `unvanquished_0.52.2` package will have this `DEPS` file:

```
unvanquished 0.52.1
tex-common
res-players
res-weapons
res-buildables
res-voices
res-soundtrack
res-legacy
```

So in future Unvanquished 0.52.0, to use `textures/colors/red` material in a map one would need `scripts/colors.shader` to be loaded from future `tex-common` dpk, but what will happen is that `unvanquished_0.52.1.dpk` will be loaded first and old `scripts/colors.shader`, and things will be broken.

I also have branch to move the `scripts/null.shader` file containing the `gfx/null` material from `unvanquished` dpk to `tex-common` dpk and rename `gfx/null` material to `textures/null` material because some format may not support images, and then we can apply `textures/null` material on some map models. But, same problem, future unvanquished_0.52.2.dpk will load unvanquished 0.52.1.dpk before future tex-common dpk, and then the wrong `scripts/null.shader` file will be loaded and then `textures/null` will be undefined.

Those are real examples of pull requests I can't submit if we don't have such feature. Well, I currently make use of noisy workarounds like copying back the new file from the new place to the old place to get it loaded properly from the wrong old place, but I want this to be the very last time I do this.

illwieckz:
Also we need to ignore the deleted file only from dependencies of the same basename, otherwise, when moving `scripts/null.shader` from `unvanquished` dpk to `tex-common` dpk, the future `unvanquished_0.52.2.dpk` would list `scripts/null.shader` in `DELETED` file and then the engine would ignore it from both `unvanquished_0.52.1.dpk` and future `tex-common` dpk shipping the new version…

illwieckz:
It is expected debug logs produce this:

```
Debug: Deleted file scripts/colors.shader in /home/illwieckz/Mapping/Unvanquished/UnvanquishedAssets/build/test/unvanquished_test.dpkdir/
Debug: Ignoring deleted file scripts/colors.shader from /home/illwieckz/.local/share/unvanquished/base/pkg/unvanquished_0.52.1.dpk
```

slipher:
Considering how names of shader files don't matter, why can't you name the new version something different? In any case it seems like a bad idea to use a very short and generic name like `colors.shader` which multiple people could easily think of.

illwieckz:
That's ugly workarounds.

Also it won't fix the problem of, for example, having an old package shipping `textures/brick.tga` and a new package shipping `textures/brick.crn` and then have the engine still loading obsolete `textures/brick.tga` when a map or model makes use of `textures/brick` material. The same happens for other extension-agnostic paths like sound paths, even some models paths are format agnostic (loading a .md5 file if present while the config file sets a .md3 one instead).

We are facing multiple variants of this issue for a very long time now, we (intentionally or unintentionally) worked around them until today by:

- not releasing the game for years
- shipping no delta paks in release

slipher:
I read the code and it appears to work just like I was saying, so I guess there was no real disagreement. Apparently with the "dependencies" thing you just meant it doesn't work to use DELETED on a file that exists inside the same pak.

Well, there's one difference which is that it only works if the pak base name is the same. This seems limiting; what if a modder wants to delete a file from the official paks?

illwieckz:
> I read the code and it appears to work just like I was saying, so I guess there was no real disagreement.

Maybe I have some misinterpretation of some things because I don't see what disagreement you refer to.

Anyway, there is no code to prevent to ignore a file that is listed in DELETED from the same pak, it's just the submitted code is unreliable on that case, and I assume there is no reason to make it reliable (in one way or another way) if we just declare we don't expect ignoring file from the same pak having the DELETED file listing the file to ignore is not supported, especially since the modder can just not ship the file to begin with.

For what I understand, if the DELETED file is iterated before the file listed in DELETED, the file will be ignored from the same pak, but if the DELETED file is iterated after the file listed in DELETED, the file will not be ignored from the same pak. But the DELETED file from the parent pak will always be read before the dependencies pak, so ignoring files is only properly supported in dependency paks.

We can just declare it's only supported to ignore files in dependencies and just rely in real deletion in the to-be-released paks. The purpose of such DELETED file is because we can't modify the already released paks we rely on, while we can modify the pak we are going to release, and then don't need such DELETED mechanism.

illwieckz:
> Well, there's one difference which is that it only works if the pak base name is the same. This seems limiting; what if a modder wants to delete a file from the official paks?

This is not the purpose of this feature, the purpose being to be able to provide a delta paks named `something` that acts like a non-delta pak also named `something` with the same version, so deleting files from older versions of itself.

Deleting files from other paks is another feature request. Though it may not be that hard to implement. The same way the DEPS file can set an optional version after the pak name, we may add an optional pak name on the same line of the file name in DELETED or other variants. But that's another feature request. My only concern is that the format for that other feature request may decide the format for the current one if we want to make it easy to use.

We can imagine multiple variant, given the current pak is `unvanquished` and the foreign pak is `tex-common`:

```
unvanquished filename1
tex-common filename2
```

```
. filename1
tex-common filename2
```

```
filename1
tex-common filename2
```

```
filename1
filename2 tex-common
```


The various variants of ”pak name first, file name first” may make things easier to read for users, because the pakname would be the same size for multiple lines for the same pak name.

The last variant would just allow us to extend the current format later.

slipher:
Yes, the  `<pakname> <filename>` format seems good. Then you won't have to change anything about your implementation except to make the parsing bit take two tokens rather than one.

illwieckz:
@necessarily-equal with this engine, you're expected to be able to solve navmesh issue for the nightly server this way:

```bash
urcheon prepare src/map-*_src.dpkdir

urcheon build -r unvanquished/0.52.1 src/map-*_src.dpkdir

for map in $(ls -d src/map-*_src.dpkdir | xargs -n1 -P1 basename | cut -c5- | cut -d_ -f1)
do
cat > "build/test/map-${map}_test.dpkdir/DELETED" <<-EOF
map-${map} maps/${map}-builder.navMesh
map-${map} maps/${map}-human_bsuit.navMesh
map-${map} maps/${map}-human_naked.navMesh
map-${map} maps/${map}-level0.navMesh
map-${map} maps/${map}-level1.navMesh
map-${map} maps/${map}-level1upg.navMesh
map-${map} maps/${map}-level2.navMesh
map-${map} maps/${map}-level2upg.navMesh
map-${map} maps/${map}-level3.navMesh
map-${map} maps/${map}-level3upg.navMesh
map-${map} maps/${map}-level4.navMesh
EOF
done

urcheon package src/map-*_src.dpkdir
```

illwieckz:
@slipher I assume the PR is now ready, I implemented the `<pakname> <filename>` and I tested it this way.

I built a delta pak of `map-chasm_src.dpkdir` having `map-chasm 1.2.1` ((the one from Unvanquished 0.52.1 release) in `DEPS`.

I added a `DELETED` file in produced `map-chasm_test.dpkdir` containing that:

```
map-chasm maps/chasm-builder.navMesh
map-chasm maps/chasm-human_bsuit.navMesh
map-chasm maps/chasm-human_naked.navMesh
map-chasm maps/chasm-level0.navMesh
map-chasm maps/chasm-level1.navMesh
map-chasm maps/chasm-level1upg.navMesh
map-chasm maps/chasm-level2.navMesh
map-chasm maps/chasm-level2upg.navMesh
map-chasm maps/chasm-level3.navMesh
map-chasm maps/chasm-level3upg.navMesh
map-chasm maps/chasm-level4.navMesh
```

I ran the game with a pakpath to the directory containing `map-chasm_test.dpkdir`, loaded chasm map, the game did not find the navmesh from older chasm pak and generated them.

On the other hand, without the `DELETED` file (or when running the engine with `master` branch), the game finds and loads the navmesh from older chasm pak and does not generate them.

src/common/FileSystem.cpp
```diff
@@ -1260,31 +1390,60 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
...
-				return;
-			depsData.resize(length);
-			auto read = zipFile.ReadFile(&depsData[0], length, err);
-			depsData.resize(read);
-			if (err)
-				return;
-		} else {
-			ASSERT_UNREACHABLE();
+	if (!isLegacy && !expectedChecksum) {
+		if (hasDeleted) {
```
slipher, line 1394: I think you don't want this to be guarded by `!expectedChecksum`. I presume the `!expectedChecksum` thing is for the case where the full list of loaded paks is transmitted from the server. You would still need to process the deleted items in that case.

src/common/FileSystem.cpp
```diff
@@ -1130,9 +1213,34 @@ static void ParseDeps(const PakInfo& parent, Str::StringRef depsData, std::error
...
+scripts/colors.shader file in the DELETED file and has unvanquished_0.52.1.dpk
+in DEPS, the scripts/colors.shader file from unvanquished_0.52.1.dpk will
+be ignored.
+
+If the scripts/colors.shader file is still shipped in unvanquished_0.52.2.dpk
+and listed in the DELETED file from unvanquished_0.52.2.dpk you cannot expect
+the file to be ignored, it is expected the user deletes the file for real in
+the parent pak. Making sure the DELETED file applies on files of the same pak
+would increase code complexity while packager can just delete the file, in
+some cases the file may be ignored (if DELETED file si read first) but you
```
slipher, line 1227: Typo: si -> is

slipher:
This should be to the 0.53 branch since clients/servers with different pak loading would be incompatible

src/common/FileSystem.cpp
illwieckz: fixed (please check this is what expected!).

src/common/FileSystem.cpp
illwieckz: fixed

necessarily-equal:
(nitpick: the commit message is a bit confusing)

illwieckz:
> This should be to the 0.53 branch since clients/servers with different pak loading would be incompatible

Hmm, the behavior differs but on the other hand there is no pak loading incompatibilities (this does not prevent to join, for example), and no one package ships any `DELETED` file yet.

The only usage before 0.53 is to enable @necessarily-equal to blacklist old navmeshes to force server-side in-game regeneration, and blacklisting navmeshes would be just ignored by released client or by released client-as-server…

This is mostly adding a feature released games don't implement than breaking something.

I added a special define `ENABLE_FUTURE` so @necessarily-equal can enable the feature in his nightly (non-0.53) server, but that means the CI does not build it.

I'm not sure we need such define anyway given in this situation the impact is very low.

necessarily-equal:
To be honest, I don't see a *need* for such a define

src/common/FileSystem.cpp
```diff
@@ -1258,33 +1391,69 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
...
-			if (err)
-				return;
-			depsData.resize(length);
-			auto read = zipFile.ReadFile(&depsData[0], length, err);
-			depsData.resize(read);
-			if (err)
-				return;
-		} else {
-			ASSERT_UNREACHABLE();
+	if (!isLegacy && !expectedChecksum) {
```
slipher, line 1430: Undo useless change making if statement into 2 if statements

src/common/FileSystem.cpp
slipher: The logic is OK. These 2 if statements could be a single if statement

slipher (commented):
The 2nd commit has junk changes to submodules.

If you really want that feature now, just merge it to master I guess; the #define thing seems worse to me. But @necessarily-equal, note that since the navmesh lazy initialization was broken and had to be reverted for now, you will have to do an sgame restart before using any generated navmeshes.

illwieckz:
I addressed comments.

slipher (commented):
LGTM

illwieckz:
(I just edited the commit message)

PR #523, opened 2021-09-15
Miscellaneous small stuff

slipher:


external_deps/build.sh
```diff
@@ -653,6 +655,10 @@ build_install() {
 		find "${PKG_PREFIX}/lib" -name '*.exp' -execdir rm -f -- {} \;
 		;;
 	esac
+
+	rmdir "${PKG_PREFIX}/bin" 2> /dev/null || true
```
illwieckz, line 659: Maybe doing it that way would be better?

```sh
[ -d "${PKG_PREFIX}/bin" ] && rmdir "${PKG_PREFIX}/bin"
```

illwieckz: Hmmm, wait, it would still return `false` if it does not exist and stop the execution…

Better do the opposite:

```
! [ -d "${PKG_PREFIX}/bin" ] || rmdir "${PKG_PREFIX}/bin"
```

Or also we may just do a reusable function for that (and proper `if` call) to avoid copy pasting.

illwieckz:
For those GLSL shaders being deleted because of not being used, we may want to do a wiki page to reference them, with link to the deletion commit, so one may know where some previous code (even if incomplete) for the feature is if we want to implement it again.

external_deps/build.sh
slipher: The directory should always exist (thanks to the other lines added in this commit); the idea is to remove it if it's empty.

I see there's an `rmdir` option on my machine `--ignore-failure-on-non-empty` doing exactly this, but it probably doesn't work on Mac or something

slipher:
> 
> 
> For those GLSL shaders being deleted because of not being used, we may want to do a wiki page to reference them, with link to the deletion commit, so one may know where some previous code (even if incomplete) for the feature is if we want to implement it again.

You can do `git log --diff-filter=D --summary src/engine/renderer/glsl_source/ | grep 'delete mode'`

illwieckz:
> > For those GLSL shaders being deleted because of not being used, we may want to do a wiki page to reference them, with link to the deletion commit, so one may know where some previous code (even if incomplete) for the feature is if we want to implement it again.
> 
> You can do `git log --diff-filter=D --summary src/engine/renderer/glsl_source/ | grep 'delete mode'`

Yes but one one will know we had some `volumetricCloud` code if we don't list it somewhere.

One can find the deletion commit with `git`, but for that, one must know we had that code and deleted it.

The related `tr_shade.cpp` code is disabled, that explains why it was already removed in Dæmon engine before, in https://github.com/DaemonEngine/Daemon/commit/580aad9a75c108323caf7fdcbce973e9f8126319 .

One may fin it there in our tree:  https://github.com/DaemonEngine/Daemon/blob/8c29cded09d83e0faf61a5e84697deb8c1ee9fdb/src/engine/renderer/tr_shade.cpp#L2557-L2678 (there was also code in `gl_shader.cpp`).

One may also find related `tr_shade.cpp` code in XreaL/RBQUAKE3: https://github.com/RobertBeckebans/RBQUAKE-3/blob/d8b9f193a4b312dc39af524e12109e028f7d5c71/code/engine/renderer/tr_shade.cpp#L3060-L3205

That's part of the code we deleted because it's either dead code, broken, or we don't make any use of it, but one may want to look at it again one day, and maybe just finish it if it was just waiting for completion. Over the years we have pruned many of those incomplete or dead code, but a list of what was deleted would be really useful.

slipher:
> Yes but no one will know we had some volumetricCloud code if we don't list it somewhere.

Not true, the command I gave lists all deleted GLSL shaders. You don't need to know which ones existed.

If you see some specific thing in Xreal and want to find out when it was deleted in our code, you can use `git log -S` with the specific string.

There is already another "volumetric fog" shader by the way.

illwieckz:
> There is already another "volumetric fog" shader by the way.

If I'm right `gl_fogQuake3Shader` is about fog that is not map-wise but brush-wise. This is the `textrures/common/fog` material if I'm right. That's basically a fog that fills a cube area, like the fog in the chasm in chasm map. Some games only support map-wide fog on the contrary. This is very different to what people usually call “volumetric clouds” which is expected to be an atmospheric effect.

external_deps/build.sh
illwieckz: Ah ok, then you can do that:

```sh
find "${PKG_PREFIX}/bin" "${PKG_PREFIX}/include" "${PKG_PREFIX}/lib" -maxdepth 0 -empty -exec rmdir {} \;
```

I verified it works on macOS (and yes macOS does not support `rmdir --ignore-failure-on-non-empty`).

slipher: I replaced it with a command to delete all empty subdirectories, which seems like a good idea and is also shorter.

illwieckz: Great! 👍

illwieckz (approved):
LGTM.

necessarily-equal:
I'm not sure dropping the shaders are an improvement. I have the impression the only benefit here is to allow for a slower compilation time, but that could also be solved with the on-the-fly compilation?

otherwise, looks good to me, and I'm fine with you merging that as I don't know much in the area.

slipher:
I'm mainly concerned with the development burden dead code imposes just by existing. Let's say you want to figure out how fog is implemented an search for "fog". You find this file and start reading it, not understanding it is dead code. Or you want to make changes to some #define/uniform/attr etc. which is referenced in this shader and then waste your time refactoring the shader to "work" with the new definition.

PR #524, opened 2021-09-16
audio: do not abort on missing or busy sound device

illwieckz:
Do not abort on missing or busy sound device.

fix https://github.com/Unvanquished/Unvanquished/issues/1031

There was already a lot of

```c++
        if (not initialized) {
            return;
        }
```

I added a bunch of others, and turned the `Sys:Error` calls into `Log::Warn` ones.

### DIFF
--- a/src/engine/audio/Audio.cpp
+++ b/src/engine/audio/Audio.cpp
@@ -33,6 +33,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "AudioData.h"
 
 namespace Audio {
+    /* When adding an entry point to the audio subsystem,
+    make sure the non-null implementation returns early
+    when audio subsystem is not initialized.
+    See https://github.com/DaemonEngine/Daemon/pull/524 */
 
     Log::Logger audioLogs("audio");
 
@@ -229,15 +233,27 @@ namespace Audio {
     }
 
     void BeginRegistration() {
+        if (not initialized) {
+            return;
+        }
+
         BeginSampleRegistration();
     }
 
     sfxHandle_t RegisterSFX(Str::StringRef filename) {
+        if (not initialized) {
+            return 0;
+        }
+
         // TODO: what should we do if we aren't initialized?
         return RegisterSample(filename)->GetHandle();
     }
 
     void EndRegistration() {
+        if (not initialized) {
+            return;
+        }
+
         EndSampleRegistration();
     }
 
@@ -340,6 +356,10 @@ namespace Audio {
     }
 
     void StopAllSounds() {
+        if (not initialized) {
+            return;
+        }
+
         StopMusic();
         StopSounds();
     }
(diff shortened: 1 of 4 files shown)


illwieckz:
Before:

```
ALSA lib pcm_dmix.c:1089:(snd_pcm_dmix_open) unable to open slave
AL lib: (EE) ALCplaybackAlsa_open: Could not open playback device 'default': Device or resource busy
Warn: Could not open an OpenAL device. 
Warn: Couldn't initialize the audio subsystem. 
Debug: ----- CL_Shutdown ----
```

After:

```
ALSA lib pcm_dmix.c:1089:(snd_pcm_dmix_open) unable to open slave
AL lib: (EE) ALCplaybackAlsa_open: Could not open playback device 'default': Device or resource busy
Warn: Could not open an OpenAL device. 
Warn: Couldn't initialize the audio subsystem. 
Loading VM module /home/illwieckz/dev/buildme-unvanquished/build/vms/default-gcc-nolto-reldeb-dll/cgame-native-dll.so... 
```

In Sound options, the dropdown list for “Audio output device/card” is empty, we may want to add some explicit wording about no device being available.

necessarily-equal:
On principle, sure. I don't really know how to test this though.

necessarily-equal (approved):

illwieckz:
> On principle, sure. I don't really know how to test this though.

You can use `ALSOFT_DRIVERS='none'` environment variable to set a non-existing OpenAL driver.

illwieckz:
I noticed that with released game, one can use `ALSOFT_DRIVERS='null'` to use a dummy audio OpenAL backend and workaround the `Sys::Error`.

We may even chose to set this environment variable when failing and reloading OpenAL as an alternate implementation, but I prefer not doing it.

With current implementation, it is expected that, when loading the game, if the sound device is unavailable the sound is missing in game, but if the user notices it, he would be able to shutdown the concurrent application, and on the next map load, return to main menu or `vid_restart`, the sound will be there again, without quitting the game.

slipher:
Do you think that playing with sound is something non-dev users ever want to do? It's really difficult to play without sound, for me anyway. If this is not something desired by non-dev users, aborting right away might be better so that the problem is surfaced as soon as possible, and can be distinguished from other problems like volume settings too low. And there can be a cvar to enable playing without audio for dev users.

illwieckz:
The one reporting the issue had a broken sound device and wanted to play the game anyway.

illwieckz:
As a side topic, we need a way to print significant error messages on main menu. We have some other errors that can happen, for example by failing to load a map where the game returns to main menu without anything told to the user (it's only written in console).

What we need is a big red string displayed on main menu (and in-map main menu) saying there is no sound card.

slipher:
You missed `SetReverb`. LGTM otherwise

illwieckz:
Ah yes, with `not initialized` early return in `SetReverb` it looks like I don't need such test in `UpdateReverbSlot` to silent a missing preset log.

@slipher what do you think it is better, to keep the test in `UpdateReverbSlot` as an extra security, or remove the test to catch a potential issue in the future if more code calls `UpdateReverbSlot`?

slipher:
I prefer to not add anything in `UpdateReverbSlot`, especially since that allows the warning to be seen. I would rather only put the checks on the entry points to the audio subsystem. You can easily find which those are by seeing which functions are defined in NullAudio.cpp

illwieckz:
Right, I modified accordingly. I also added some comments to tell future ourselves to not forget to add early return when audio is not initialized when adding entry points to the audio subsystem.

PR #526, opened 2021-09-22
CI: get USE_WERROR working with MSVC

slipher:


illwieckz:
So we drop Visual Studio 2017?

We went testing VS2017 because our `README` says VS2017 is supported, see https://github.com/DaemonEngine/Daemon/pull/262 and https://github.com/DaemonEngine/Daemon/blob/83c0a039f5c590b4ddd0bcb54425ce37decb6908/README.md#L23

We went testing amd64 on VS2019 and i686 on VS2017 to avoid doing 4 tests and assuming this would be enough to catch both VS2017, VS2019, amd64 and i686 issues.

slipher:
> We went testing VS2017 because our `README` says VS2017 is supported

My commit updates the README to stay in sync.

The latest Visual Studio works just as well on older versions of Windows, so there's no motivation to keep testing with one that's a few years old.

illwieckz:
Ok that's fine then. 👍

I'm OK with dropping VS2017, I just wanted to be sure we don't miss anything.

illwieckz:
I remind we still don't test Linux build since Travis was dropped, so I don't know if there can be a collateral regression using the build checks. On the other hand the macOS build check was passed so it's very likely the Linux one is OK, and it builds on Ubuntu 20.04 on my end.

illwieckz (approved):
LGTM, though I'm not a Visual Studio specialist.

PR #528, opened 2021-09-27
cg_api: make possible for game to know if an extension-independent image path exists

illwieckz:
I wanted to know if it's possible for the game to query the engine to know if an extension-independent image path exists, it looks like this code make it possible. It would also be possible to do the same for sounds if we have a need.

What this make possible is for example to check if Unvanquished's `meta/<name>/<mapname>` levelshot exists, but if it does not exist, try to load Quake3-like `levelshot/<name>` instead.


necessarily-equal (dismissed):

slipher:
Ignoring possible warning spam, is there some use case that you can't do by attempting to load the image and seeing if the handle is 0?

illwieckz:
Right.

Anyway the usage of it is obsoleted by:

- https://github.com/Unvanquished/Unvanquished/pull/3214

PR #529, opened 2021-09-28
Small fixes

necessarily-equal:
see commits

### DIFF
--- a/src/engine/sys/sdl_input.cpp
+++ b/src/engine/sys/sdl_input.cpp
@@ -492,9 +492,17 @@ static keyNum_t hat_keys[ 16 ] =
 	K_JOY19, K_JOY20
 };
 
+// TODO(C++14): remove and use std::max instead
+static constexpr int max(int a, int b) { return a>b ? a : b; }
+
 struct
 {
-	bool     buttons[ 16 ]; // !!! FIXME: these might be too many.
+	// FIXME: this used to be 16, and was looped over up to
+	// SDL_CONTROLLER_BUTTON_MAX but neither look correct. At least this
+	// avoid memory corruption, but I'd be surprised if the code around
+	// this works correctly.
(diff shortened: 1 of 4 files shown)


src/engine/renderer/tr_main.cpp
```diff
@@ -2686,7 +2686,7 @@ void R_AddLightBoundsToVisBounds()
 static std::vector<char> botDebugDrawCommands;
 void RE_SendBotDebugDrawCommands( std::vector<char> commands )
 {
-	botDebugDrawCommands = std::move(commands);
+	botDebugDrawCommands.swap(commands);
```
slipher, line 2689: Why?

src/engine/sys/sdl_input.cpp
```diff
@@ -679,15 +679,8 @@ static void IN_JoyMove()
 		{
 			// !!! FIXME: is this good for stick balls, or just mice?
 			// Scale like the mouse input...
-			if ( abs( balldx ) > 1 )
```
slipher, line 674: This is not a simplification. It does something different.

src/engine/sys/sdl_input.cpp
```diff
@@ -492,9 +492,17 @@ static keyNum_t hat_keys[ 16 ] =
...
 
+// TODO(C++14): remove and use std::max instead
+static constexpr int max(int a, int b) { return a>b ? a : b; }
+
 struct
 {
-	bool     buttons[ 16 ]; // !!! FIXME: these might be too many.
+	// FIXME: this used to be 16, and was looped over up to
+	// SDL_CONTROLLER_BUTTON_MAX but neither look correct. At least this
+	// avoid memory corruption, but I'd be surprised if the code around
```
slipher, line 502: K_JOY1 to K_JOY32 are defined so it looks like it should be OK iff the number is <= 32.

PR #530, opened 2021-10-01
Move translation stuff to cgame

slipher:
Also fix the `print_tr` format string parser which is used in both the engine and gamelogic.


slipher:
Moved to #531

PR #545, opened 2021-11-11
Don't delete master servers with DNS failure

slipher:
Instead, establish a 1-minute time limit between attempts to do DNS
lookups of master servers. This also prevents a possible DoS issue in
that it might be possible to force DNS lookups at an unlimited rate by
sending info packets to the server.

### DIFF
--- a/src/engine/server/server.h
+++ b/src/engine/server/server.h
@@ -250,6 +250,7 @@ struct serverStatic_t
 	int           nextSnapshotEntities; // next snapshotEntities to use
 	std::unique_ptr<entityState_t[]> snapshotEntities; // [numSnapshotEntities]
 	int           nextHeartbeatTime;
+	int nextMasterResolutionTime;
 	receipt_t     infoReceipts[ MAX_INFO_RECEIPTS ];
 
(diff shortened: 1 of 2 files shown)


illwieckz:
Hmm, rate-limiting DNS lookup is a nice idea for such DoS prevention indeed.

I wonder if DNS lookup timeout can affect the game or not… and then if it's OK to experience a DNS timeout every minute if a master can't be resolved.

slipher:
> I wonder if DNS lookup timeout can affect the game or not… and then if it's OK to experience a DNS timeout every minute if a master can't be resolved.

This normally won't affect clients, since they have `server.private 3` by default. It would only come into play if you were trying to host an online game from your PC with port forwarding or something.

illwieckz:
I don't talk about clients, I talk about servers. If for some reasons a server freezes, client connected to it may lag. So my questions is: is a DNS timeout can freeze a server people are playing on?

slipher:
> So my questions is: is a DNS timeout can freeze a server people are playing on?

Yes, that will still be possible, as it is now.

slipher:
Superseded by #579.

PR #546, opened 2021-11-13
Drop custom Com_Memset, Com_Memcpy, Com_Allocate and Com_Dealloc

necessarily-equal:
They are only silencing compiler warnings about their incorrect usage


slipher:
LGTM

PR #548, opened 2021-12-04
use std::offsetof

bmorel:
Replaces a hack to get offsets by the built-in C "offsetof". This reduces UBsan's noise and makes things more readable.


slipher:
LGTM

PR #550, opened 2021-12-07
Error message for VM timeout

slipher:
Fix the bad error message reported by @bmorel


bmorel:
I don't know what's this cg_viewsize thing is, but in any case, LGTM

PR #554, opened 2021-12-28
filesystem: only load paks with map name when testing if map name has valid map pak

illwieckz:
Only load paks with map name when testing if map name has valid map pak (has bsp file with same name).

Also, only load legacy paks whatever the name when legacy paks are enabled instead of loading all paks whatever the name when legacy paks are enabled.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -2768,21 +2768,45 @@ const std::vector<PakInfo>& GetAvailablePaks()
 	return availablePaks;
 }
 
-
-std::vector<PakInfo> GetAvailableMapPaks()
+// mapName is optional and set to "" by default (list paks for all map names)
+std::vector<PakInfo> GetAvailableMapPaks( std::string mapName )
 {
 	std::vector<PakInfo> infos;
 	for ( const auto& pak : FS::GetAvailablePaks() )
 	{
-		if (UseLegacyPaks() || Str::IsPrefix("map-", pak.name))
+		// Legacy pak has empty version string
+		if ( pak.version.empty() )
 		{
-			infos.push_back(pak);
+			if ( UseLegacyPaks() )
+			{
+				infos.push_back(pak);
+			}
+		}
+		else
+		{
+			std::string mapPrefix = "map-";
+			if ( mapName.empty() )
+			{
+				if ( Str::IsPrefix( mapPrefix, pak.name ) )
+				{
+					infos.push_back(pak);
+				}
+			}
+			else
+			{
+				std::string mapBasename = mapPrefix + mapName;
+				if ( mapBasename == pak.name )
+				{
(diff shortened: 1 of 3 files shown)


slipher:
There do exist old maps with an underscore in the name, which means you can't rename the archive to match `map-$NAME.dpk`. E.g. `pushcannon_b3a`

illwieckz:
I'm not sure to get your comment. There is no support for DPK map with underscore in name, only legacy maps support them.

- If `map-pushcannon_b3a.dpk` exists, base name is `map-pushcannon`.
- If `map-pushcannon_b3a.pk3` exists, base name is `map-pushcannon_b3a`

Maybe you mean someone can't rename `map-pushcannon_b3a.pk3` to `map-pushcannon_b3a.dpk` without renaming `maps/pushcannon_b3a.bsp` to `maps/pushcannon.bsp` (and related map files), this patch has no impact on this. This is already true, one cannot load from a dpk a bsp that has an underscore in its basename (because the map dpk base name can't have an underscore in its basename to begin with).

illwieckz:
I rewrote some code because I noticed I must not compare pak basename has `map-${name}` prefix, but that pak basename is equal to `map-${name}`.

slipher:
> There is no support for DPK map with underscore in name

Yes there is. I can take the pushcannon pk3, rename it to e.g. `map-pushcannon_zzz.dpk`, and use it with `devmap pushcannon_b3a`.

illwieckz:
Huu, I would consider such undocumented feature as a bug to begin with… I don't think we must take care of keeping this working. Such map is even not listable as client. There may be also ways to load `aaa.bsp` from `map-bbb_0.dpk`, I think that should be considered a bug.

There is a serious topic though: in the past games usually provided a large pk3 with all their maps in, without `map-` prefix (like `pak0.pk3` (quake3), or `xonotic-maps.pk3` (xonotic), and other variants). Do we want this to be supported with DPK? The purpose of enforcing the `map-` prefix was to not have to load the paks themselves to list maps, listing filesystem would be enough.

I'm even surprised we actually check for `.bsp` being present in `map-*.dpk` archive (I think it's OK to list the map and abort at loading if the dpk is malformed and does not contain a bsp). I would not be against simplifying the map listing code even more (and maybe make it common between engine and game code thanks to that).

The game code to list maps actually works by listing dpk with `map-` prefix in basename and stripping such prefix (it means it actually does not list legacy maps anyway).

slipher:
In the current state of this PR, I feel like the different parts of the system are only becoming more inconsistent with one another... I opened #557 for more discussion of the issue.

illwieckz:
I rebased on current master.

slipher:
Implemented this for all parts of the system in https://github.com/DaemonEngine/Daemon/pull/601.

PR #559, opened 2021-12-30
Document code layout with READMEs

slipher:


illwieckz:
Maybe name them `README.md` so we can use markdown and are more code forge friendly (GitHub, GitLab, etc. know how to display a `README.md` at the root of a folder)?

slipher:
I don't want to use markdown, because I want it to be readable on the developer's own filesystem.

illwieckz:
The purpose of Markdown is meant to be readable without Markdown viewer.

slipher:
OK I added the extensions so it won't be displayed with typewriter font...

src/common/IPC/README.md
```diff
@@ -0,0 +1,2 @@
+This directory contains code relating to Inter-Process Communication (i.e. between the engine and a
```
illwieckz, line 1: Why not removing the repetitive and obvious `This directory contains`? It looks like all directories are directories containing something, so maybe we can just start with what it contains, this way:

```
Code relating to Inter-Process Communication (i.e. between the engine and a gamelogic VM).
```

src/common/cm/README.md
```diff
@@ -0,0 +1,3 @@
+This directory contains the library for loading, traversing, and querying BSPs. A .bsp file is the
+compiled form of a map. The renderer, the server, and the server and client gamelogics all use it.
```
illwieckz, line 2: Are those two `server` mentions about different things, or is it a duplicate?

src/engine/README.md
```diff
@@ -0,0 +1,2 @@
+This directory primarily contains code for the engine. That is, the main program that runs
```
illwieckz, line 1: If we remove `This directory primarily`, we may also remove `That is,`:

```
Code for the engine: the main program that runs […]
```

src/common/cm/README.md
slipher: Yes server and server gamelogic are different things.

src/common/IPC/README.md
slipher: I don't consider it so obvious as to go without saying. Plus, it changes the meaning. `Code for the engine` suggests the *entire* source of the engine, while `This directory contains code for the engine` makes it clearer that it doesn't necessarily mean all engine code.

illwieckz: OK

illwieckz (approved):
LGTM.

bmorel:
LGTM. Adding doc, without any risk of introducing bugs, that's a great PR!

PR #561, opened 2021-12-31
external_deps/build.sh improvements

illwieckz:
Some WIP improvements for `external_deps/build.sh`:

- fallback on `wget` if `curl` fails (macOS curl fails to download ogg tarball from Xiph, same with Homebrew curl and own build curl);
- fix `jpeg` build on macOS by using own-build `nasm` instead of system one;
- basic introspection (each platform declares itself, each package declares for which platform it is enabled by default, etc.);
- make possible to build `defaults` which build all required packages for the given platform (introspection);
- detect and report properly unsupported platform and unsupported packages (introspection);
- autodocumented built-in help (introspection).

Note: macOS users require a file system with locking supported. This is required for mounting sdl2 `.dmg` files and to build geoip (job compilation control and temporary file creation done by geoip build system relies on locking). When building over NFS, if locking does not work, one may mount with `-o locallocks` to emulate the required file system feature.

Unfortunately the introspection feature (both used for the `defaults` feature, troubleshooting and built-in help) makes use associative array feature from bash, which is believed to be shipped in bash starting with bash 4 but macOS only ship bash 3. Homebrew ships bash 5 anyway.


illwieckz:
This:
```sh
	sed -e 's/\\t/\t/g' <<-EOF
	Usage: $(basename "${0}") <platform> <package[s]…|command[s]…>

	Script to build dependencies for platforms which do not provide them.

	Platforms:
	$(print_platforms)

	Packages:
	$(print_packages)

	Packages required for each platform:
	$(print_packages_per_platform)

	Meta-package (build all required packages for the given platform):
	\tdefaults

	Commands:
	$(print_commands)
	EOF
```

Produces this:

```
Usage: build.sh <platform> <package[s]…|command[s]…>

Script to build dependencies for platforms which do not provide them.

Platforms:
	linux64: Linux amd64 native compilation
	macosx64: macOS amd64 native compilation
	mingw32: Windows i686 cross-compilation from Linux
	mingw64: Windows amd64 cross-compilation from Linux
	msvc32: Windows i686 native compilation
	msvc64: Windows amd64 native compilation

Packages:
	nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis speex opus opusfile lua ncurses wasisdk wasmtime naclsdk naclports gendef

Packages required for each platform:
	linux64: zlib naclsdk naclports
	msvc32: nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis speex opus opusfile lua naclsdk naclports
	mingw64: nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis speex opus opusfile lua naclsdk naclports
	msvc64: nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis speex opus opusfile lua naclsdk naclports
	mingw32: nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis speex opus opusfile lua naclsdk naclports
	macosx64: nasm gmp nettle sdl2 glew png jpeg webp freetype openal ogg vorbis speex opus opusfile lua naclsdk naclports
	unused: ncurses wasisdk wasmtime gendef

Meta-package (build all required packages for the given platform):
	defaults

Commands:
	clean: Remove products of build process, excepting download cache. Must be last.
	package: Create a zip/tarball of the dependencies so they can be distributed.
	install: Create a stripped down version of the built packages that CMake can use.
```

It even detects packages that are not enabled by any known platform.

illwieckz:
So, I'm now able to fully build macOS deps myself (something I never achieved before).

And I'm also able to fully build Unvanquished (something I never achieved before) using the built deps, that means building and run both engine, dll game, and nexe game.

@slipher do you think it's OK to require macOS users to upgrade their bash with Homebrew or should we find a way to workaround this, even if that means going super ugly like emulating associative arrays with temporary files?

One thing is that I didn't managed to run any of my own engine build in repository build directory using prebuilt deps files (but it works if I move the built engine binary to where the updater installed the game). So maybe I'm missing some instructions to make me able to run  in repository build directory my own engine build with deps built by something else, but maybe macOS coders may want to build their own deps as well.

illwieckz:
I remember @cu-kai had trouble to build s/cgame as nexe on macOS, I was able to build and run some by compiling against deps built using that modifed script. I don't know if I fixed something, or if the deps files were actually outdated. For example the current macOS `macosx64-5.tar.bz2` deps archive provide libOpenAL 1.18.2 but the script builds libOpenAL 1.21.1. So maybe NaCl was also outdated?

slipher:
I think you can do the 'defaults' feature with a simple implementation instead.

Like 
```
macosx64_defaults='foo bar'
mingw32_defaults='foo bar baz'
def build_defaults() {
  for p in ${PLATFORM}_defaults
    build_${p}
  done
}
```

> One thing is that I didn't managed to run any of my own engine build in repository build directory using prebuilt deps files (but it works if I move the built engine binary to where the updater installed the game). So maybe I'm missing some instructions to make me able to run in repository build directory my own engine build with deps built by something else, but maybe macOS coders may want to build their own deps as well.

Run the `build.sh macosx64 install` command, which will organize the deps into `${PWD}/macosx64-5`. Then configure CMake with `-DEXTERNAL_DEPS_DIR=${PWD}` where PWD is the current directory when you built the deps.

illwieckz:
> I think you can do the 'defaults' feature with a simple implementation instead.
> Like […]

Yeah, I know about that, but then you have to hardcode any platform in a variable, and then you have to maintain a package list per platform instead of enabling/disabling a platform per package. This looks more prone to errors and mistake to me, that's why I went the associative array way at first.

> Run the […]

Hmm, there is misinterpretation. I successfully built and run my own dæmon/unvanquished build with my own deps. What does not work is running a build done using autodownloaded deps. It builds, but it doesn't run. I'm probably facing some issues about the deps being self-signed by @DolceTriade's computer or something like that, and I don't know how to disable the security for the build folder (if it's possible).

slipher:
> I'm probably facing some issues about the deps being self-signed by @DolceTriade's computer or something like that, and I don't know how to disable the security for the build folder (if it's possible).

I think that stuff only happens for stuff downloaded in a browser, not with e.g. curl. But if you are having that stuff somehow, it is related to some file attributes on the directory. You have to list the extended file attributes and check if there is one like `com.apple.`something and remove that.

Note that the caution about CMake cache variables still applies. If you used your own deps in a build at first and then switched to prepackaged ones, then old paths will be stored in the cache and you need to wipe your build dir.

slipher:
> Yeah, I know about that, but then you have to hardcode any platform in a variable, and then you have to maintain a package list per platform instead of enabling/disabling a platform per package.

OK but this minor point of taste is not worth adding a new dependency over. Note that the per-platform lists are "hard-coded" into the help string, so if you use the variables in the help string then you have no more such lists than before :)

illwieckz:
(my latest push was just some bikeshedding)

> Note that the per-platform lists are "hard-coded" into the help string.

Not anymore, all the lists in the help message are now fully generated.

illwieckz:
I was able to avoid the usage of associative arrays by doing `eval` magic to create the equivalent of your suggested `macosx64_defaults` variables on the fly. So the script now works with macOS old GPLv2 bash 3.2.57.

slipher:
This kind of stuff is awfully unreadable:

>  ```
> register_command() {
> 	local command_name="${1}"
> 	local command_string="${2}"
> 
> 	if ! echo "${command_list_string}" \
> 		| to_list \
> 		| egrep -q "^${command_name}"
> 	then
> 		if ! [ -z "${command_list_string}" ]
> 		then
> 			command_list_string+=' '
> 		fi
> 
> 		command_list_string+="${command_name}"
> 		eval "command_${command_name}_string='${command_string}'"
> 	fi
> }
> ```

It's really not worth adding hundreds of lines of code to avoid writing required dependencies in a single list. It's just over-engineering.

illwieckz:
It looks like using strings allows to write much simpler code… After all unix shell is all about strings. For example there are many situations where I don't need to convert arrays to string again, since they are strings built properly from the beginning. So I simplified a lot the code. The sum of all registration functions are now 32 lines long.

> avoid writing required dependencies

@slipher the purpose is not to avoid writing dependencies, the purpose is to have the code checks for mistakes by itself, and prevent list rotting. Also there are less platforms than packages, so it's more safe to list platforms per packages than packages per platforms: it means more lists but those lists are less long to read and those lists are local to the package they're about.

Even if we were listing packages per platform, I would reuse the same functions to validate the lists properly (they would just read lists the other way).

slipher:
Personally I'm not a fan of `register_command` or `register_package`; I'd rather have the script be smaller and simpler. Let's get a third opinion?

Supposing we went forward with register_package, I'd want to see a distinction between a package being unavailable on a given platform, and available but not part of the default build. I'd find it confusing that a package may be implemented for other platforms than the ones listed in the register_package line.

cu-kai:
> 

I ran the build recently on macOS Monterey (x86_64) and was able to produce working binaries, not sure if this helps now?

illwieckz:
> Supposing we went forward with register_package, I'd want to see a distinction between a package being unavailable on a given platform, and available but not part of the default build. I'd find it confusing that a package may be implemented for other platforms than the ones listed in the register_package line.

With latest commit you can build `required`, `optional` or `all`.

Note that “available but not required” was not fully documented, I tried to guess.

illwieckz:
This looks ready to me.

slipher:
I guess we're still waiting for that third opinion on registration stuff.

necessarily-equal:
To be honest, I couldn't easily grasp what this `register_package` was about. i think it could at least use a comment in the source code.

illwieckz:
@necessarily-equal I added comments explaining what the `register_*` commands do.

illwieckz:
Question:

```
	linux64: Linux amd64 native compilation
	macosx64: macOS amd64 native compilation
	mingw32: Windows i686 cross-compilation from Linux
	mingw64: Windows amd64 cross-compilation from Linux
	msvc32: Windows i686 native compilation
	msvc64: Windows amd64 native compilation
```

When building on Windows without MSVC, for example with MSYS2/MingW, is it mingw that is used?

slipher:
> When building on Windows without MSVC, for example with MSYS2/MingW, is it mingw32/mingw64 package that is used?

Yes. So your quoted text is inaccurate.

illwieckz:
The initial text was “Linux to Windows cross-compile”, I'll fix that.

illwieckz:
With following changes this PR is supposed to fail until we rebuild new packages (we may also investigate rebuilding the packages instead of downloading them when the package build script is modified).

I propose renaming package using triplets and using `_` as version separator. The triplet I suggest is `<system>-<architecture>-<compiler>`, where `system` can be `linux`, `windows` or `macos`, where architecture can be `amd64` or `i686`, and where `compiler` can be `msvc` or `mingw` on Windows, `default` on other systems.

I also removed speex building as it seems to not be used anymore, removed theora references as it is not used anymore, and removed geoip building and made it disabled by default (but code to use it is still there, nuking it is outside of the scope of this PR).

```
./build.sh -h
Usage: build.sh [PLATFORM] [SELECTION]… [PACKAGE]… [COMMAND]…

Script to build dependencies for platforms which do not provide them.

Platforms:
	linux-amd64-default: Linux amd64 native compilation
	macos-amd64-default: macOS amd64 native compilation
	windows-amd64-mingw: Windows amd64 MingW compilation or cross-compilation from Linux
	windows-amd64-msvc: Windows amd64 native compilation
	windows-i686-mingw: Windows i686 MingW compilation or cross-compilation from Linux
	windows-i686-msvc: Windows i686 native compilation

Selections:
	required  (build required packages for the given platform)
	optional  (build optional packages for the given platform)
	all       (build all available packages for the given platform)

Packages:
	pkgconfig nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis opus opusfile lua ncurses wasisdk wasmtime naclsdk naclports gendef

Required packages per platform:
	linux-amd64-default: zlib naclsdk naclports
	macos-amd64-default: pkgconfig nasm gmp nettle sdl2 glew png jpeg webp freetype openal ogg vorbis opus opusfile lua naclsdk naclports
	windows-amd64-mingw: nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis opus opusfile lua naclsdk naclports
	windows-amd64-msvc: pkgconfig nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis opus opusfile lua naclsdk naclports
	windows-i686-mingw: nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis opus opusfile lua naclsdk naclports
	windows-i686-msvc: pkgconfig nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis opus opusfile lua naclsdk naclports

Optional packages per platform:
	linux-amd64-default: pkgconfig gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis opus opusfile lua ncurses wasisdk wasmtime
	macos-amd64-default: zlib curl ncurses wasisdk wasmtime
	windows-amd64-mingw: pkgconfig ncurses wasisdk wasmtime
	windows-amd64-msvc: ncurses wasisdk wasmtime gendef
	windows-i686-mingw: pkgconfig ncurses wasisdk wasmtime
	windows-i686-msvc: ncurses wasisdk wasmtime gendef

Unused packages:
	ncurses wasisdk wasmtime gendef

Commands:
	install: Create a stripped down version of the built packages that CMake can use.
	package: Create a zip/tarball of the dependencies so they can be distributed.
	clean: Remove products of build process, excepting download cache. Must be last.

Example:
	build.sh linux-amd64-default required ncurses wasisdk wasmtime install package clean

```

slipher:
The geoip library should not be deleted from build.sh before it is removed from the engine.

illwieckz:
Is it a problem to not ship GeoIP if building against it is disabled by default?
Anyway it's better to remove GeoIP in another PR. If we prefer to keep it, I can drop the commit here in all case.

illwieckz:
I removed the commit to not build and package geoip for now.

illwieckz:
One of the changes I may have forget to list there is that the linux package will use `xz` instead of `bzip2` to compress the archive.

illwieckz:
Good to know: 7z (which compresses zip better than `zip`) adds symlinks to zip without dereferencing them, which seems to make CMake fail on Windows:

```
-- Downloading dependencies from 'https://dl.unvanquished.net/deps/windows-amd64-msvc_6.zip'
[…]
-- [download 100% complete]
-- Download completed successfully
CMake Error: Problem with archive_read_next_header(): Unsupported ZIP compression method during decompression of link entry (8: deflation)
CMake Error: Problem extracting tar: C:/projects/daemon/build/CMakeFiles/windows-amd64-msvc_6.zip
CMake Error at CMakeLists.txt:450 (message):
Could not extract windows-amd64-msvc_6.zip
-- Configuring incomplete, errors occurred!
See also "C:/projects/daemon/build/CMakeFiles/CMakeOutput.log".
See also "C:/projects/daemon/build/CMakeFiles/CMakeError.log".
Command exited with code 1
```

It looks like 7z doesn't have option to dereference symlinks, but symlinks can be dereferenced at `install` time anyway.

illwieckz:
Now on Windows I get a Zlib that is both seen and unseen at the same time: “_Could NOT find ZLIB (missing: ZLIB_LIBRARY) (found version "1.2.11")_”.

[Example](https://ci.appveyor.com/project/DolceTriade/daemon/builds/43410825/job/kijo20s2n8d6vtma):

```
-- Downloading dependencies from 'https://dl.unvanquished.net/deps/windows-amd64-msvc_6.zip'
[…]
-- [download 100% complete]
-- Download completed successfully
-- Found OpenGL: opengl32   
CMake Error at C:/Program Files/CMake/share/cmake-3.23/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find ZLIB (missing: ZLIB_LIBRARY) (found version "1.2.11")
  C:/Program Files/CMake/share/cmake-3.23/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files/CMake/share/cmake-3.23/Modules/FindZLIB.cmake:120 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:547 (find_package)
-- Configuring incomplete, errors occurred!
See also "C:/projects/daemon/build/CMakeFiles/CMakeOutput.log".
See also "C:/projects/daemon/build/CMakeFiles/CMakeError.log".
Command exited with code 1
```

illwieckz:
It looks like when I build msvc stuff on Linux it produces `.a` files instead of `.lib` files and then they get removied at install time.

illwieckz:
In fact I also get this behavior with old `build.sh` from before this PR, this is not a bug I introduced. 🤔️

illwieckz:
It looks like the only `.lib` files I get are the ones from SDL2 because the script downloads a prebuilt SDL2 for Windows. Every lib I build myself is producing `.a` files instead.

illwieckz:
@slipher do you know how to build the windows-msvc archives? Is it even possible on Linux or it is required to do it on Windows, and if Windows is required, can it be done with mingw or msvc is required?

slipher:
> @slipher do you know how to build the windows-msvc archives? Is it even possible on Linux or it is required to do it on Windows, and if Windows is required, can it be done with mingw or msvc is required?

It's required to do it on Linux, except there's this one little part that you have to do on Windows IIRC.

illwieckz:
> It's required to do it on Linux, except there's this one little part that you have to do on Windows IIRC.

Do you have more information about that one little part. I have a Windows vm and if I can do it that would be nice (also I would just like to know about it anyway).

illwieckz:
I continue **there** the thread started here https://github.com/DaemonEngine/Daemon/pull/629#issuecomment-1138270937 because this thread talks about the present PR:

> @illwieckz:
> I'm gonna merge https://github.com/DaemonEngine/Daemon/pull/561 in all cases, so cherry-picking instead of rebasing makes everyone's life harder with the only benefit of increasing the risk of introducing mistakes.

> @slipher
> Why? AFAIK no one else said they liked the package registration stuff.

> @illwieckz
> No one else expressed any opinion against it as well.
>
> And it fulfills my need of a script that acts consistently with the documentation by design because the documentation is produced by the same code that does the actions.

> @slipher
> Usually we consider that a pull request should have a favorable review before being merged. More people should be for the change than against it.

It looks like no one cared about this script before I took it for a ride. I'm unhappy that bash doesn't make it easy to make it clean and more unhappy that macos distributes a very obsolete bash making things more hackish, but that drawback is worth the reliability acquired. Also this is really appreciable to be able to build `required`, `optional` or `all`. I don't like the bash wizardry involved but that's a cost I cannot avoid.

So basically there is ~60 lines of complex code working around bash (and obsolete macos bash) limitations to implement kind of associative arrays, and then we just add nice annotations for every platform and packages at the exact same place they are defined, and then the script becomes aware of what is for what, being able to check for platform mistakes, package mistakes, select per-platform packages required or not, list platforms and packages, etc.

Note: I would not be against porting this to another language like python in the future to get rid of the bash wizardry.

slipher:
OK, so we can write down a list of optional packages, and make `required` and `optional` pseudo-packages. And this will take, say, 10 lines of simple code to give the desired functionality of `register_package`.

AFAICT `register_command` and `register_platform` don't give us any functionality we don't already have.

Seeing the package name `gendef`, I recall that it is related to the bit that has to run on Windows, but I don't recall exactly how it works.

illwieckz:
The hardcoded list solution, while looking simple at first glance, carry many issues:

- it breaks the principle of locality, the way things are used are not defined where they exist (that was the very first problem that makes me deciding to investigate this, as I had no way to trust the lists).
- there is no validation of what's in them.
- the build order has to be replicated as much as there are platforms, with the big drawback each platform doesn't have the same package list so adding or removing one package requires a very different edit (that's the thing that is calling for trouble the most).

For the last point, the build order may be important (for example we build nasm, and we may need nasm to build other things), with his implementation to change the build order of a package we would just move up or down the related package block, same for inserting a new package in the build order, just insert it once after the one it needs and/or before the one that needs it, and forget everything about platforms, and optional, all, required lists…

If we hardcode `all`, `optional` and `required` lists, those problems skyrocket, we would litterally add levels of complexity the human mind have to process when editing the lists, like adding a package or changing the build order, or make some optional. That ugly bash code exists to delegate this tasks from human's mind responsibility to an automaton and prevent the human to do mistakes by solving multi-level complexity problem (basically an equation with several unknowns). Doing the lists by hand is like considering every package is an unknown, and every (`all`, `optional`, `required`) set per platform is an equation line, and solving that equation by hand.

slipher:
register_package doesn't do anything at all to solve the dependency problem! It does not declare any dependencies between packages and just works the same way as the current hard-coded list in the help - just listing packages in the order they are defined in the file. The only difference being the user has to wade through a soup of inscrutable Bash nonsense to tell what order the packages will be built in. And maintaining a topological sort of the packages (i.e. a sequence respecting dependencies) is *better* how it is now - you edit a list at a single place in the file, while with register_package it would apparently involve looking at the whole file and copy-pasting around the entire `build_xxx` functions (assuming you wanted to keep them next to their register_package statements).

The `all` list can absolutely be the concatenation of `required` and `optional`, because a required package obviously shouldn't depend on an optional package. (The 'optional' concept doesn't even exist for platforms other than Linux, by the way. Unless you count WASM which simply isn't used for anything.)

bmorel:
Since I was asked to comment on this PR... let's go.

For a start, I think this PR could easily be split. The improved feedback, the removal of geoip and speex, the use of `$(...)` instead of the backticks (`), the comments...
Those changes are not subject to any questioning I think.

As  for the  stuff subject to questions, I'm not sure which problem exactly this script is meant to solve so I'm hardly able to express any useful opinion.
From what I read from comments, it's about implementing features which are missing in Apple's OSes (apple, again giving troubles...).
Would it not be easier to just install a non-outdated bash on those instead?

A diff shows 3 big blocks of added code. Reading bash code is not really my fun, but I certainly find code like this:

```
register_command() {
  if ! to_lines "${commands:-}" | egrep -q "^${1}"; then 
    commands+="${commands:+ }${1}"
    eval "command_${1}='${2}'"
  fi
}
```

Not readable. The reason is probably that I can't decipher what `to_lines` does. No, I'm not able to understand what `echo "${@// /$'\n'}";` does easily. I only know it's some kind of substitution...
Most of the times, the pattern is `! to_lines $something | egrep -q "$pattern"`.
The other times, it's: `to_lines $something | sort -u`.
To me, those should be 2 different functions, which would make things much more readable.

I could even push the idea further, as `register_command()` and `register_platform()` contains the exact same lines, only some values in there change.
A notable fact is, there's example of how to use register_package, but examples are not good documentation. Why some stuff are prefixed with ':'?

Basically, the only thing I can say about the additional code is that, if it's useful, then it could probably be more generic, to reduce code duplication.

Again, I do not understand which problem this thing is supposed to solved.

illwieckz:
> register_package doesn't do anything at all to solve the dependency problem!

It removes the need to solve it by simply ordering the per-package function in the source code, then the calling of register_package the closest possible to those function is solving the problem. This is _solving problems by design_.

> The only difference being the user has to wade through a soup of inscrutable Bash nonsense

You are aware that some other people would be able to say such words for some other code of the project you write in some other language you have much more experience than others, let's say c++? (I know this discussion happened). If we don't want bash we can change the language, but this is probably the less intrusive way to implement it with bash.

> to tell what order the packages will be built in.

The user doesn't have to read any code to tell what order the package will be built in, part of the motivation for this code to be written is to tell the user the order without requiring from him to read code. That is in fact one of the argument I listed in my previous comment : the code tells the user the build order without having to read code. And for the one wanting to read code, the build order is defined by the order of the calling of the register commands, which are by design meant to be called the closest possible to the function building a package (this was designed for this).

This code was precisely written to solve the problem you use to argument against that code.

> The all list can absolutely be the concatenation of required and optional, because a required package obviously shouldn't depend on an optional package.

This may be true for our current packages, I don't know and I don't want to look at this as this would only lead us to make assumptions about a situation that will be forgotten and to make this code relying on an omission justified by an untold reason. But Linux distributions are full of required packages that can be built differently given this or that optional package. The “_all package list cannot be the concatenation of required and optional lists_” is just one of the major rules that must be considered when dealing with packages, especially when building packages. For example Gentoo is fully based on this principle (this is in fact one of reason why it exists), and even binary distros like Debian allows to rebuild a required package with an optional package. A real-life example is to rebuild Squid proxy with OpenSSL (on Debian you have to install OpenSSL then rebuild the Squid package using the Debian tools as OpenSSL is an optional dependency to Squid), so if you require a Squid with OpenSSL, your required Squid actually depends on an optional package. Here we need nasm on macOS to build Jpeg, we probably assume it is already provided by the system on Linux, but we may still want to make possible to optionally build nasm on Linux to build Jpeg if we need to require to build a specific version of Jpeg one day. Such assumption is just wrong, it only works by luck.

illwieckz:
@freem:

> For a start, I think this PR could easily be split.

There is multiple commits, but some changes are done after others changes, and you know how that can be painful to extract or reorder them.

About this:

```bash
if ! to_lines "${commands:-}" | egrep -q "^${1}"
```

The `to_lines` function turns a space-separated list into a line-separated list, then the `egrep` call tells if an argument exists as a line or not, so this is equivalent to:

```bash
if ! echo "${commands:-}" | tr ' ' '\n' | egrep -q "^${1}"
```

Which is equivalent to (in Python):

```python
if arg1 not in commands:
```

when `commands` is a list.

> Why some stuff are prefixed with ':'?

This is described in the example but you're right I can add a line to say it explicitly. This is an arbitrary prefix to mark packages as optional for the named platform. I may also do `opt:` or something else, like this:

```
register_package pkgconfig opt:linux-amd64-default opt:windows-i686-mingw opt:windows-amd64-mingw windows-i686-msvc windows-amd64-msvc macos-amd64-default
```

> I could even push the idea further, as `register_command()` and `register_platform()` contains the exact same lines, only some values in there change.

> Basically, the only thing I can say about the additional code is that, if it's useful, then it could probably be more generic, to reduce code duplication.

Making things more generic would require to fully implement ordered associative arrays in bash, this will require much more code than just duplicating those. Here is an example of associative arrays implementation (I haven't tried it):

-  https://github.com/dmitrinove/associative_arrays/blob/master/hash.sh

slipher:
I can't believe you're claiming that it's a good thing that to change the build order of the packages, you'd have to move the build_xxx functions around. This is a a much worse developer experience than editing a one-line list. Just to ascertain the order of the packages, you have to scroll through hundreds of lines. And to edit it you have to awkwardly copy and paste big blocks.


> > The only difference being the user has to wade through a soup of inscrutable Bash nonsense
> 
> You are aware that some other people would be able to say such words for some other code of the project you write in some other language you have much more experience than others, let's say c++?

Sure, I'm thinking relative to a typical Unvanquished code contributor. To do anything here the basic requirement is C++ knowledge; it's to some degree unfortunate when something else is required (e.g. Python + YAML for CBSE).


> This code was precisely written to solve the problem you use to argument against that code.

What?


> > The all list can absolutely be the concatenation of required and optional, because a required package obviously shouldn't depend on an optional package.
> 
> This may be true for our current packages, I don't know and I don't want to look at this as this would only lead us to make assumptions about a situation that will be forgotten and to make this code relying on an omission justified by an untold reason. But Linux distributions are full of required packages that can be built differently given this or that optional package. The “_all package list cannot be the concatenation of required and optional lists_” is just one of the major rules that must be considered when dealing with packages, especially when building packages. For example Gentoo is fully based on this principle (this is in fact one of reason why it exists), and even binary distros like Debian allows to rebuild a required package with an optional package. A real-life example is to rebuild Squid proxy with OpenSSL (on Debian you have to install OpenSSL then rebuild the Squid package using the Debian tools as OpenSSL is an optional dependency to Squid), so if you require a Squid with OpenSSL, your required Squid actually depends on an optional package. Here we need nasm on macOS to build Jpeg, we probably assume it is already provided by the system on Linux, but we may still want to make possible to optionally build nasm on Linux to build Jpeg if we need to require to build a specific version of Jpeg one day. Such assumption is just wrong, it only works by luck.

I'm aware of this kind of optional dependency. For example, when creating the Unvanquished release Docker script I included a lot of audio and video libraries so that the SDL build would enable a lot of optional backends. But I consider it unlikely that such an optional dependency will be relevant for our build script. I know that JPEG will build when NASM is missing despite our having explicitly having a path, but I consider this a bug. I'd set an option to make it error in the absence of NASM if there were one.

bmorel:
Out of topic, but allow me:

> it's to some degree unfortunate when something else is required (e.g. Python + YAML for CBSE).

Do that means that replacing this with something not requiring another lang would be a good idea?

Also, I think you should try to keep the heat for temp medium like IRC, which feels way more adapted to that kind of contents/tones to me.

illwieckz:
@bmorel I managed to remove `to_lines` by simply using `\n` as separator from the start.

illwieckz:
On the root topic of improving the deps building whatever our opinions on the way the script is written, I wonder if the gendef stuff required to run on Windows can be done on Wine.

bmorel:
>> For a start, I think this PR could easily be split.
>
There is multiple commits, but some changes are done after others changes, and you know how that can be painful to extract or reorder them.

I do know. But it's how things are done in unvanquished, rebasing all the time to reduce number of commits and make things easier to review by others. To be honest, in some rare occasions, those rebases also helps the writer, as long as there's no other branch depending on it (which happens to me almost all the time sadly).
In the end I've found it's sometimes easier to checkout a different branch, to keep things safe, soft-reset on master and do the commits as if they were fresh.

I have the feeling that, as long as this PR is not merged since there's some stuff related to geoip there, there will be no `unvanquished` release, though, and this annoys me a lot.
I really do not imagine this discussion to be solved quickly (considering the tone in some messages), and blocking 0.53 longer means no work on GUI stuff (because nobody wants to do cherry picking of _three_ repos)  which is imo a rather critical thing: chat system, vote system (including for next map), selection of mod in solo games, making /devmap accessible to people without too much console knowledge, updating rmlui to the last version...

At a first glance, most of the changes in #629, except [eb83b58](https://github.com/DaemonEngine/Daemon/pull/629/commits/eb83b58a6e212dcb299064a615691ff32d56fc7b) seems compatible with this PR.
@illwieckz actually, it looks like if @slipher did the split job, and if the conflicting part could be removed it would allow to simplify this PR and refocus on those, and only those. Won't solve any problem, ofc, but let's try to move what can be moved.

illwieckz:
Some things were implemented in #629.

Other changes would get other PRs if we want them.

PR #562, opened 2022-01-01
FileSystem: do not iterate string null terminator, fix Unvanquished/Unvanquished#1659

illwieckz:
FileSystem: do not iterate string null terminator, fix https://github.com/Unvanquished/Unvanquished/issues/1659

I don't know why this fixes the bug, maybe the iterator pointer to null terminator points to a common null character that is not part of the string we iterate but part of a common empty string somewhere?

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -508,6 +508,9 @@ std::string Extension(Str::StringRef path)
 
 	// Find a dot or slash, searching from the end of the string
 	for (const char* p = path.end(); p != path.begin(); p--) {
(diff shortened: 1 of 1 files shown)


slipher:
The original code doesn't even dereference the null terminator, as it only reads the character before the pointer. (Not that there would be anything wrong if it did - `StringRef` is supposed to be used on null-terminated strings only.0

The patch also makes the functions not work correctly.

illwieckz:
Yes, this now looks stupid to me. And why this works around the bug is probably as magic as the fact adding a log works around the bug as well.

PR #568, opened 2022-01-10
tr_shade: workaround to not crash on liquidMap when liquidMapping is disabled

illwieckz:
liquidMapping is not functional and not used yet, but we need to not
crash when loading liquidMap test assets while liquidMapping is disabled.

When liquidMapping is disabled this patch just display the liquidMap as
a diffuseMap with a white opaque texture and the liquidMap as normalMap.

No playable map uses liquidMap yet and liquidMap syntax is likely to be
entirely redone in the future so I see no need to implement more complex
workaround (like adding translucency).


slipher:
LGTM

PR #570, opened 2022-01-15
Warn unused and clamp

necessarily-equal:


slipher:
I would rather keep `a = Clamp(a,b,c)`. I think ClampArg just makes it harder to read.

Other commits look good.

bmorel:
I agree with slipher: the only reason I erroneously used Math::Clamp() without reading return value was because this is what unvanquished's API usually does, especially regarding vec3_t. I prefer, by far, to have return value.

necessarily-equal:
I've dropped the ClampArg thing.

bmorel:
LGTM

src/common/Math.h
```diff
@@ -36,7 +36,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace Math {
 
     // Returns min if value is NaN
-    template<typename T> T Clamp(T value, T min, T max)
+    template<typename T> inline WARN_UNUSED_RESULT
```
slipher, line 39: FYI templates are automatically inline.

slipher (commented):
Any reason not to merge this?

necessarily-equal:
You can merge it, I've just not connected to github for a least a week now

Le 1 février 2022 01:46:05 GMT+01:00, slipher ***@***.***> a écrit :
***@***.*** commented on this pull request.

PR #571, opened 2022-01-16
Remove duplicate IN_Init call

slipher:


src/engine/qcommon/common.cpp
```diff
@@ -576,10 +576,9 @@ Com_Crash_f
 A way to force a bus error for development reasons
 =================
 */
-static void NORETURN Com_Crash_f()
+static void Com_Crash_f()
 {
 	* ( volatile int * ) 0 = 0x12345678;
```
bmorel, line 581: maybe document what's this line is while you're at it?

bmorel (approved):
LGTM

src/engine/qcommon/common.cpp
slipher: Crashing of course :)

bmorel: Ah, well, yes didn't noticed the *0, I was focused on the 0x12346.... silly me

PR #572, opened 2022-01-18
Enable C++14 for RmlUi

slipher:
_No description provided._

---

_Edit by illwieckz._

Needed by:

- https://github.com/Unvanquished/Unvanquished/pull/1715

See also:

- https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/95

### DIFF
--- a/src/common/System.cpp
+++ b/src/common/System.cpp
@@ -416,7 +416,13 @@ void* operator new(size_t n)
 		Sys::Error("Out of memory");
 	return p;
 }
+
 void operator delete(void* p) NOEXCEPT
+{
+	free(p);
+}
+
+void operator delete(void* p, size_t) NOEXCEPT
 {
 	free(p);
 }
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -167,7 +167,7 @@ if (MSVC)
     set_c_cxx_flag("/wd4996")
 elseif (NACL)
     set_c_flag("-std=c11")
-    set_cxx_flag("-std=gnu++11")
+    set_cxx_flag("-std=gnu++14")
 
     set_c_cxx_flag("-ffast-math")
     set_c_cxx_flag("-fvisibility=hidden")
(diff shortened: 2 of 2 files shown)


necessarily-equal:
I'll trust you on that one

src/common/System.cpp
```diff
@@ -391,9 +391,17 @@ void* operator new(size_t n)
...
 	return p;
 }
+
 void operator delete(void* p) NOEXCEPT
 {
 	if (!Sys::processTerminating) {
 		free(p);
 	}
 }
+
```
bmorel, line 401: makes me thinking about 3 (not related to C++14) things:

1. there is no new[] / delete[] operators
2. builds don't pass when using USE_ADDRESS_SANITIZER

Not really the point of the PR, though.

bmorel:
LGTM. I don't know the relations with NaCl, so like necessarily-equal I'll trust you on that point.

src/common/System.cpp
necessarily-equal: This makes me think that for ASAN we could simply encapsulate the whole block like [this](https://clang.llvm.org/docs/AddressSanitizer.html#conditional-compilation-with-has-feature-address-sanitizer):
```C++
#if !defined(__has_feature) || !__has_feature(address_sanitizer)
// code that builds only when AddressSanitizer is disabled
#endif
```

slipher: Yeah that's what I did on my inactive ASAN branch. https://github.com/slipher/Daemon/commit/237ff1b764d4eb56ef6ac64888d881992971575f

illwieckz (approved):

illwieckz:
_(I just reverted the branch to its original state after `master` was mistakenly merged into it by @bmorel)._

illwieckz:
I rebased on current `master`.

illwieckz:
Well no, I reverted to before the rebase, something very wrong happened while rebasing. @slipher I'll let you fix the conflict.

illwieckz:
The code works for me but it breaks as soon as I rebase this branch to master.

I also noticed that on Unvanquished side the PR targets `0.53.0/sync` branch, but this one targets `master`.

Otherwise, everything looks good to me, it seems to work as expected.

PR #575, opened 2022-01-24
Fast engine shutdown

slipher:
Shut down the engine quickly instead of doing nonsense like reloading the filesystem before exit. If you want to do all that stuff for debugging purposes (e.g. to detect memory leaks) you can use the `common.pedanticShutdown` cvar.

Things I identified as important shutdown tasks which shouldn't be skipped:
- Flush files
- Terminate gamelogic VMs
- Send "dead heartbeat" to master server (not done because it might block on DNS)
- Restore terminal state
- Close the single-instance lock file
- SDL_Quit to clean up full screen stuff
- For clients, send disconnection packet to the server
- Delete temp files from downloads (not implemented)
- Remove temp files from FS_WRITE_VIA_TEMPORARY
- Finish demo recording
- Finish video recording

Also do proper shutdown when clicking X on the Windows curses window.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -2297,6 +2297,8 @@ void Initialize(Str::StringRef homePath, Str::StringRef libPath, const std::vect
 
 void FlushAll()
 {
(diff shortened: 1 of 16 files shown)


DolceTriade (approved):
nice

PR #577, opened 2022-01-28
Fix DLL unloading on MacOS

slipher:
On MacOS, dlclose has no effect if the shared library contains a
thread_local variable. Get rid of an instance of thread_local
in gamelogic DLLs to fix this. It is important for gamelogic
DLLs to be unloaded and reloaded so that memory is freed and
variables reinitialized.

### DIFF
--- a/src/common/IPC/Primitives.cpp
+++ b/src/common/IPC/Primitives.cpp
@@ -321,7 +321,7 @@ static void FreeHandles(const NaClHandle* h)
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


slipher:
Added back the thread_local if `BUILD_ENGINE` because I realized this code can be used in the engine too, which does have multiple VM connections (though it's generally single-threaded, so there should not be any concern right now).

DolceTriade (approved):

PR #578, opened 2022-01-31
Implement music volume (audio.volume.music)

slipher:


src/engine/audio/Sound.cpp
```diff
@@ -206,6 +205,16 @@ namespace Audio {
         return currentGain;
     }
 
+    void Sound::SetVolumeModifier(const Cvar::Range<Cvar::Cvar<float>>& volumeModifier)
```
bmorel, line 208: why use a shadowing name? I find this an annoying thing and always have a doubt on scope. This will also trigger warnings.

bmorel (approved):
I put some remarks, but those are really unimportant details which don't have to be addressed at all.

src/engine/audio/Sound.cpp
slipher: Done.

src/engine/audio/Sound.cpp
```diff
@@ -175,7 +173,8 @@ namespace Audio {
 
     // Implementation of Sound
 
-    Sound::Sound(): positionalGain(1.0f), soundGain(1.0f), currentGain(1.0f), playing(false), source(nullptr) {
+    Sound::Sound()
+        : positionalGain(1.0f), soundGain(1.0f), currentGain(1.0f), playing(false), volumeModifier(&effectsVolume), source(nullptr) {
```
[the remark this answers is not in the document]
slipher: Added a line break.

bmorel:
Still LGTM

PR #581, opened 2022-02-18
DirToByte does not modify it's param + warning fixes

bmorel:


src/engine/qcommon/q_math.cpp
```diff
@@ -187,7 +187,7 @@ signed char ClampChar( int i )
 }
 
 // this isn't a real cheap function to call!
-int DirToByte( vec3_t dir )
+int DirToByte( vec3_t const dir )
```
necessarily-equal, line 190: Open question: doesn't this break compat between older game and newer daemon, and newer game and old daemon?

src/engine/renderer/tr_light.cpp
```diff
@@ -565,6 +565,7 @@ void R_TessLight( const trRefLight_t *light, const Color::Color& color, bool use
 					Tess_AddQuadStamp2( quadVerts, use_default_color ? Color::Red : color );
 				}
 			}
+			DAEMON_FALLTHROUGH;
```
necessarily-equal, line 568: I would just put

```suggestion
			break;
```

src/engine/renderer/tr_backend.cpp
```diff
@@ -1968,6 +1968,7 @@ static void RB_SetupLightForLighting( trRefLight_t *light )
 				}
 			}
 
+			DAEMON_FALLTHROUGH;
```
necessarily-equal, line 1971: I would just put

```suggestion
			break;

```

src/engine/renderer/tr_shader.cpp
```diff
@@ -1464,6 +1464,7 @@ static bool LoadMap( shaderStage_t *stage, const char *buffer, const int bundleI
 		case stageType_t::ST_HEATHAZEMAP:
 		case stageType_t::ST_LIQUIDMAP:
 			imageParams.bits |= IF_NORMALMAP;
+			DAEMON_FALLTHROUGH;
```
necessarily-equal, line 1467: ditto

```suggestion
			break;
```

src/common/System.cpp
```diff
@@ -278,7 +278,7 @@ void SetupCrashHandler()
     nacl_minidump_set_callback(CrashHandler);
 }
 #else
-static void CrashHandler(int sig)
+[[noreturn]] static void CrashHandler(int sig)
```
necessarily-equal, line 281: we have `NORETURN` instead. Either it isn't needed anymore, and it should probably be replaced, or it is still needed, and you would need to use it.

necessarily-equal (commented):
I still have to review « fix -Wcomma » (which also does two things I think?), here my review of the rest

src/engine/qcommon/q_math.cpp
bmorel: Making constness explicit does not break compat.

necessarily-equal: I've never studied ABI compat, but I would guess that since it's possible to define both `f(int*)` and `f(const int*)` at the same time, the function names must be mangled differently

necessarily-equal: (I'll defer to @slipher on that one)

slipher: There is no ABI boundary here, or anywhere in normal C++ functions. Our binary interfaces are the IPC serialized stuff and anything going on the network.

src/engine/qcommon/cmd.cpp
```diff
@@ -253,12 +264,7 @@ static void Tokenise( const char *text, char *textOut, bool tokens, bool ignoreQ
 
 				while ( *text && *text != '"' )
 				{
-					if ( *text == '\\' && (++text, (*text >= 0 && *text < ' ') ) )
```
slipher, line 256: These two blocks don't do the same thing. The first one is for an unquoted token which can end at a space. The second is for a string literal which should end with a quote.

src/engine/framework/System.cpp
```diff
@@ -549,7 +549,7 @@ static void Init(int argc, char** argv)
 #endif
 
 	// Print a banner and a copy of the command-line arguments
-	Log::Notice(Q3_VERSION " " PLATFORM_STRING " " ARCH_STRING " " __DATE__);
```
slipher, line 552: Can't you fix that with compiler flags or something?

bmorel: With compiler flags? I do not see how. Maybe it could be made a build option though. TBH this commit is not the most interesting to me, so I would even be fine for a removal.

src/engine/qcommon/cmd.cpp
bmorel: Right. I do not remember all I had in mind at this time, so for now I'll just move the commit out of the PR.

bmorel:
@slipher I've removed the -Wcomma and the -Wreproducibility from the stack

bmorel:
note that all of `DirToByte()`, `ByteToDir()`, and `bytedirs[]` are *never* used by daemon-engine itself. As such I think they should be *removed* from it, when games depending on it have integrated them (easy: there's only unvanquished).

bmorel:
From IRC:

> 16:02:08 <afontain[m]> LGTM on #581 btw

slipher:
LGTM

PR #587, opened 2022-02-26
tr_surface: avoid some VectorClear

illwieckz:
Just a minor possible improvement I spotted while reading the code.


slipher:
LGTM

Issue #590, opened 2022-02-28
sv_master* cvars are considered modified even if they carry their default value


illwieckz:
`sv_master*` cvars are considered modified even if they carry their default value and configuration does not have any `set sv_master…` command.

I have not checked other cvars.

illwieckz:
With #591:

```
/sv_master1  
"sv_master1" - text - address of a master server - default: "master.unvanquished.net" (modified)
```

slipher:
#591 probably doesn't do what you think it does. I don't think there's any bug here.

illwieckz:
Calling `/reset` on the cvar turns the `modified` bool to `false` anyway.

slipher:
> Calling `/reset` on the cvar turns the `modified` bool to `false` anyway.

Huh, that's a bug. That means if you use /reset on a cvar using the `modified` mechanism, it won't notice the change.

PR #591, opened 2022-02-28
framework/CvarSystem: print when a cvar is modified

illwieckz:
It helped me to verify:

- #590

### DIFF
--- a/src/engine/framework/CvarSystem.cpp
+++ b/src/engine/framework/CvarSystem.cpp
@@ -210,7 +210,10 @@ namespace Cvar {
                 cvarRecord_t* var = cvars[name];
 
                 if (args.Argc() < 2) {
-                    Print("\"%s\" - %s^* - default: \"%s^*\"", name.c_str(), var->description.c_str(), var->resetValue.c_str());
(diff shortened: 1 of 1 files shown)


slipher:
`modified` is a flag for legacy cvars which can be used to perform an action only when a cvar changes. Something like

```
// Change the fullscreen state only when the cvar changes
if (!r_fullscreen->modified)
    SetFullscreen(r_fullscreen->integer);
    r_fullscreen->modified = false;
}
```

Most cvars don't use this mechanism, so `modified` is always true for them.

What did you think the `modified` flag means?

illwieckz:
> What did you think the `modified` flag means?

I was expecting to know if a cvar was modified. For example, is it equal to empty string because the user emptied it or because nothing was never written there.

slipher:
> I was expecting to know if a cvar was modified.

You mean if it's not equal to the default value?

illwieckz:
> You mean if it's not equal to the default value?

Yes, of course I can compare or make a `modified()` function for that.

slipher:
The value and default value are shown, so it should be obvious whether they are not equal to each other. We don't need to print `modified` which would further increase the amount of text in some lines which are often very long.

PR #595, opened 2022-03-03
Lower maximum FPS to 333 to avoid various issues

Calinou:
This prevents the first-person weapon model from flickering and also avoids persistent "Connection Interrupted" screens. 333 FPS is still enough to feed almost any monitor out there (excluding the currently elusive 360 Hz and 390 Hz monitors), provided your hardware is fast enough.

Note: I don't have a development setup, so I didn't test this.

This closes https://github.com/Unvanquished/Unvanquished/issues/1734.

### DIFF
--- a/src/engine/qcommon/common.cpp
+++ b/src/engine/qcommon/common.cpp
@@ -912,25 +912,28 @@ void Com_Frame()
 		{
 			if ( com_minimized->integer && maxfpsMinimized.Get() > 0 )
 			{
-				minMsec = 1000 / maxfpsMinimized.Get();
+				minMsec = 1000 / std::min(maxfpsMinimized.Get(), 333);
 			}
 			else if ( com_unfocused->integer && maxfpsUnfocused.Get() > 0 )
 			{
-				minMsec = 1000 / maxfpsUnfocused.Get();
+				minMsec = 1000 / std::min(maxfpsUnfocused.Get(), 333);
 			}
 			else if ( maxfps.Get() > 0 )
 			{
-				minMsec = 1000 / maxfps.Get();
+				minMsec = 1000 / std::min(maxfps.Get(), 333);
 			}
 			else
 			{
-				minMsec = 1;
+				minMsec = 3;
 			}
 		}
 	}
 	else
 	{
-		minMsec = 1; // Bad things happen if this is 0
+		// Bad things happen if this is 0.
+		// At 1 or 2, the game still runs but exhibits various issues
+		// (such as the first-person weapon model flickering).
+		minMsec = 3;
 	}
 
 	Com_EventLoop();


src/engine/qcommon/common.cpp
```diff
@@ -912,25 +912,28 @@ void Com_Frame()
...
 			}
 		}
 	}
 	else
 	{
-		minMsec = 1; // Bad things happen if this is 0
+		// Bad things happen if this is 0.
+		// At 1 or 2, the game still runs but exhibits various issues
+		// (such as the first-person weapon model flickering).
+		minMsec = 3;
```
illwieckz, line 936: If I'm right timedemo is to benchmark things, I wonder if that can cap benchmarks?

src/engine/qcommon/common.cpp
```diff
@@ -912,25 +912,28 @@ void Com_Frame()
...
 			else if ( maxfps.Get() > 0 )
 			{
-				minMsec = 1000 / maxfps.Get();
+				minMsec = 1000 / std::min(maxfps.Get(), 333);
 			}
 			else
 			{
-				minMsec = 1;
+				minMsec = 3;
 			}
```
illwieckz, line 928: Why not just doing `minSec = std::min(minSec, 3);` after all this group of `if`s?

slipher:
The animation glitches are most likely due to bugs in gamelogic. I wouldn't want to limit Daemon just to work around cgame bugs.

Connection issues on the other hand may be real engine limitations, so it might make sense to limit FPS for those.

illwieckz:
I had a quick look at this yesterday on gamecode side, and I noticed that some frameLerp was also dividing some value to `1000` (`tr_animation.cpp`, `cg_animation.cpp`) but that's all.

Maybe this line is interesting: https://github.com/Unvanquished/Unvanquished/blob/9738ad7a8cc46e7367715bcaa90e68996637b162/src/cgame/cg_animation.cpp#L45

```
cg.time should be between oldFrameTime and frameTime after exit
```

What happens if time between oldFrameTime and frameTime is too small to be significant?

src/engine/qcommon/common.cpp
Calinou: I'm not sure if timedemo influences the maximum effective FPS that can be reached (since the game will run faster than normal). Logically, I'd say yes, but I haven't tested this in any id Tech 3 game.

illwieckz:
I'm also a bit worried at working around this in engine when the bug is in game code, but if we're going to do it, why not something like that instead?

```c++
	// we may want to spin here if things are going too fast
	if ( !cvar_demo_timedemo.Get() )
	{
		if ( Com_IsDedicatedServer() )
		{
			minMsec = SV_FrameMsec();
		}
		else
		{
			int max;

			if ( com_minimized->integer )
			{
				max = maxfpsMinimized.Get();
			}
			else if ( com_unfocused->integer )
			{
				max = maxfpsUnfocused.Get();
			}
			else
			{
				max = maxfps.Get();
			}

			switch(max) {
				case -1:
					// Bad things happen if this is 0.
					minMsec = 1;
				break;
				case 0:
					// At 1 or 2, the game still runs but exhibits various issues
					// (such as the first-person weapon model flickering).
					minMsec = 3;
				break;
				default:
					minMsec = std::min(max / 1000, 3);
			}
		}
	}
	else
	{
		// Bad things happen if this is 0.
		// At 1 or 2, the game still runs but exhibits various issues
		// (such as the first-person weapon model flickering).
		minMsec = 3;
	}
```

Note: it also allows to use `-1` to really unlock the fps on purpose (debugging, benchmarking…).

src/engine/qcommon/common.cpp
illwieckz: I tested with a demo I had on my storage:

With 3:
```
7932 frames, 28.8s: 275.6 fps
7932 frames, 28.6s: 277.6 fps
```

With 1:
```
7932 frames, 25.9s: 305.7 fps
7932 frames, 25.9s: 306.1 fps
```

Note: I have not noticed weapon disappearing issue with timedemo.

src/engine/qcommon/common.cpp
illwieckz: See [this other comment](https://github.com/DaemonEngine/Daemon/pull/595#issuecomment-1058405894) for better suggestion.

src/engine/qcommon/common.cpp
illwieckz: I also recorded a demo while weapon was flickering and another demo while weapon wasn't flickering (capped at 333fps), and both timedemoed demo looked fine with `minMsec 1`.

illwieckz:
@Calinou see #599

PR #599, opened 2022-03-06
Lower maximum FPS to 333 to avoid various issues (rewrite)

illwieckz:
This prevents the first-person weapon model from flickering
and also avoids persistent "Connection Interrupted" screens.

Make possible to really unlock FPS with maxfps -1.

This rewrite a bit:

- https://github.com/DaemonEngine/Daemon/pull/595

---

Fix #595
Fix Unvanquished/Unvanquished#1734

### DIFF
--- a/src/engine/qcommon/common.cpp
+++ b/src/engine/qcommon/common.cpp
@@ -910,18 +910,36 @@ void Com_Frame()
 		}
 		else
 		{
-			if ( com_minimized->integer && maxfpsMinimized.Get() > 0 )
+			int max;
+
+			if ( com_minimized->integer )
 			{
-				minMsec = 1000 / maxfpsMinimized.Get();
+				max = maxfpsMinimized.Get();
 			}
-			else if ( com_unfocused->integer && maxfpsUnfocused.Get() > 0 )
+			else if ( com_unfocused->integer )
 			{
-				minMsec = 1000 / maxfpsUnfocused.Get();
+				max = maxfpsUnfocused.Get();
 			}
-			else if ( maxfps.Get() > 0 )
+			else
 			{
-				minMsec = 1000 / maxfps.Get();
+				max = maxfps.Get();
 			}
+
+			// A positive maxfps caps the fps to the given number, with an implicit
+			// cap at 333fps to avoid bugs. Above 333fps minMsec is less than 3.
+			// At 1 or 2 minMsec, the game still runs but exhibits various issues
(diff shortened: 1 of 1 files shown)


illwieckz:
I would still appreciate to know why first person weapons disappear anyway. I also noticed that increasing `timescale` hides the bug, of course it also increases timescale so that's speeding-up the game which is unwanted, but it means that it's possible to render more than 333 frames per seconds without issue, and that may partially explain why playing a demo with `cvar_demo_timedemo` enabled is not affected by the bug.

illwieckz:
On a related topic, I also noticed that enabling `cvar_demo_timedemo` is uncapping FPS even if there is no demo played. But I don't know how to test if a demo is currently played or not.

src/engine/qcommon/common.cpp
```diff
@@ -910,18 +910,36 @@ void Com_Frame()
...
+
+			// A positive maxfps caps the fps to the given number, with an implicit
+			// cap at 333fps to avoid bugs. Above 333fps minMsec is less than 3.
+			// At 1 or 2 minMsec, the game still runs but exhibits various issues
+			// such as first-person weapon model flickering, or client having
+			// connection issues with server.
+			if (max > 0)
 			{
-				minMsec = 1000 / maxfpsUnfocused.Get();
+				minMsec = std::min(1000/max, 3);
```
Calinou, line 935: The previous code used spaces between division operators:
```suggestion
				minMsec = std::min(1000 / max, 3);
```

Calinou (approved):

src/engine/qcommon/common.cpp
illwieckz: Right.

src/engine/qcommon/common.cpp
```diff
@@ -910,18 +910,36 @@ void Com_Frame()
...
+
+			// A positive maxfps caps the fps to the given number, with an implicit
+			// cap at 333fps to avoid bugs. Above 333fps minMsec is less than 3.
+			// At 1 or 2 minMsec, the game still runs but exhibits various issues
+			// such as first-person weapon model flickering, or client having
+			// connection issues with server.
+			if ( max > 0 )
 			{
-				minMsec = 1000 / maxfpsUnfocused.Get();
+				minMsec = std::min( 1000 / max, 3 );
```
slipher, line 935: This should be std::max

PR #601, opened 2022-03-08
Map paks must be named map-<mapname>

### LINKED ISSUE #557: Finding maps requires loading all map paks
For various map operations, the engine opens all packages beginning with `map-`. This is pretty slow and scales poorly, so it would be better if we could avoid it. Different parts of the code use different algorithms to determine what a map is:

- The UI map list reads `meta/*.arena` files from all `map-` packages.
- Map command tab completion and the *dedicated server* command `listmaps` load all map packages and determine that there is a map `$NAME` for each `maps/$NAME.bsp`.
- Map commands decide which pak is the map pak for map `$NAME` by loading all `map-*.dpk` and searching for a `maps/$NAME.bsp` file among loaded paks.
- The `listmaps` networked client command uses the (sometimes incorrect) heuristic of assuming there is a map named `$NAME` if there is a pak whose basename is `map-$NAME`.

One possibility is to enforce that `maps/$NAME.bsp` must be located in a pak named `map-$NAME`. This would mean that map commands no longer need to load all packages, and prevent incorrect results from the `listmaps` client command. However, it would not solve the problems with the UI map list, because it depends on the arena files to determine the map's display name in the dropdown. Assuming we continue to rely on arena files, it seems the only solution there would be some sort of caching.

slipher:
Implement the restriction that a map named $NAME must be in a pak named map-$NAME. Fixes #557.

This allows us to avoid a big filesystem hit of loading all map paks upon loading a map or listing maps.

Companion: https://github.com/Unvanquished/Unvanquished/pull/1782

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -1140,6 +1140,10 @@ static void ParseDeleted(const PakInfo& parent, Str::StringRef deletedData)
 	}
 }
 
+static void InternalLoadPak(
+	const PakInfo& pak, Util::optional<uint32_t> expectedChecksum, Str::StringRef pathPrefix,
+	bool loadDeps, std::error_code& err);
+
 // Parse the dependencies file of a package
 // Each line of the dependencies file is a name followed by an optional version
 static void ParseDeps(const PakInfo& parent, Str::StringRef depsData, Str::StringRef prefix, std::error_code& err)
@@ -1176,7 +1180,7 @@ static void ParseDeps(const PakInfo& parent, Str::StringRef depsData, Str::Strin
 				SetErrorCodeFilesystem(err, filesystem_error::missing_dependency);
 				return;
 			}
-			LoadPakPrefix(*pak, prefix, err);
+			InternalLoadPak(*pak, Util::nullopt, prefix, true, err);
 			if (err)
 				return;
 			lineStart = lineEnd == depsData.end() ? lineEnd : lineEnd + 1;
@@ -1200,7 +1204,7 @@ static void ParseDeps(const PakInfo& parent, Str::StringRef depsData, Str::Strin
 				SetErrorCodeFilesystem(err, filesystem_error::missing_dependency);
 				return;
 			}
-			LoadPakPrefix(*pak, prefix, err);
+			InternalLoadPak(*pak, Util::nullopt, prefix, true, err);
 			if (err)
 				return;
 			lineStart = lineEnd == depsData.end() ? lineEnd : lineEnd + 1;
@@ -1236,7 +1240,9 @@ static bool FileIsDeleted(const PakInfo& pak, Str::StringRef filename)
 	return deletedFileSet.find(std::pair<std::string, std::string>(pak.name, filename)) != deletedFileSet.end();
 }
 
-static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expectedChecksum, Str::StringRef pathPrefix, std::error_code& err)
+static void InternalLoadPak(
+	const PakInfo& pak, Util::optional<uint32_t> expectedChecksum, Str::StringRef pathPrefix,
+	bool loadDeps, std::error_code& err)
 {
 	Util::optional<uint32_t> realChecksum;
 	bool hasDeleted = false;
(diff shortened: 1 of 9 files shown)


illwieckz:
There is no support for listing legacy maps when `fs_legacyPaks` is enabled?

slipher:
> There is no support for listing legacy maps when fs_legacyPaks is enabled?

You mean you want to be able to list pk3's without the `map-` prefix?

illwieckz:
> You mean you want to be able to list pk3's without the `map-` prefix?

Yes, map prefix and basename being equal to bsp names is only trustable on DPK maps, and many tremulous maps didn't have map-prefix or may have mixed case between bsp name and package name (and sometime extra characters). Loading pk3 is very useful to evaluate a Tremulous map before attempting to port it.

When `fs_legacyPaks is enabled` I would just list all pk3 files and look for maps/*.bsp in them the old way. That can be an extra function only used when the feature is enable.

I don't know if it's possible to list pk3 explicitly, but a good way to know a pak is a legacy pak is to look for version string, the internal struct must have empty version string if not dpk, so one may want to iterate all the paks like this:

```c++
	std::vector<PakInfo> infos;
	for ( const auto& pak : FS::GetAvailablePaks() )
	{
		// Legacy pak has empty version string
		if ( pak.version.empty() )
		{
			if ( UseLegacyPaks() )
			{
				infos.push_back(pak);
			}
		}
	}
	return infos;
```

I don't mind if pk3 map listing is slower than dpk map listing (it is disabled by default anyway).

slipher:
I restored the ability to *load* pk3 maps. This is only used for debugging though, so do you really need to be able to *list* them? If so which listing methods do you require to work? (tab completion, `listmaps` sgame command, `listmaps` server command, UI dropdowns)

illwieckz:
I need tab completion and listmaps (otherwise it would be very painful). There is no need for UI dropdown at all, as we consider pk3 maps as to-be-ported when talking about Unvanquished. If one day a Dæmon-based game wants to also list them, that's their problem.

illwieckz:
About listmaps: I don't think I ever ran a pk3 map on a server. I always do that on a client, so I only need listmaps on a client.

slipher:
Are you sure you need (the non-server) `listmaps`? It's an sgame command, so it only works if you are in game, not on the main menu.

illwieckz:
> Are you sure you need (the non-server) `listmaps`? It's an sgame command, so it only works if you are in game, not on the main menu.

I don't remember what is what. What I need is to be able to do listmaps from console whatever a map is loaded or not.

slipher:
If you are in a dedicated server console, `listmaps` is an engine command.

If you are in a client, the `listmaps` command is forwarded by the cgame to the server and the sgame sends back a paged list.

illwieckz:
I never loaded a pk3 from a dedicated server console, so engine command support is low priority. On the other hand since the engine is assumed to support pk3, it would be inconsistent if engine command does not list maps from them.

The higher priority to me is the ability to do listmaps from a client console (the sgame one, then).

What listmaps command is called when being on client main menu without any map being loaded yet?

slipher:
> What listmaps command is called when being on client main menu without any map being loaded yet?

None, it doesn't work there. We may want to redesign this somehow, since it's stupid that you can't do it there.

illwieckz:
Ah… I see. Now that rings a bell.

slipher:
Added back the legacy maps for the daemonded `listmaps` and `map`/`devmap` tab completion. For the sgame `listmaps` and the UI dropdowns I think it's preferable to exclude them because those maps are not allowed in votes or admin commands (according to the G_MapExists function).

illwieckz:
On my end I'm unable to list a single map (even non-legacy dpk one) from client main menu console when using `listmaps` (but UI list works), I can only list maps with `listmaps` once I actually loaded a map.

illwieckz:
What I understand it that when running `daemonded` the command is engine one, while when running disconnected `daemon`, the command is the one querying the server, but there is no server.

slipher:
> On my end I'm unable to list a single map (even non-legacy dpk one) from client main menu console when using listmaps (but UI list works), I can only list maps with listmaps once I actually loaded a map.

We already talked about that above, see 


> What I understand it that when running `daemonded` the command is engine one, while when running disconnected `daemon`, the command is the one querying the server, but there is no server.

That's right, as we discussed in earlier comments. I didn't change anything there.

illwieckz:
Then I see no regression.

illwieckz (approved):
LGTM.

illwieckz:
Also this PR makes me happy because it removes some game-specific stuff from engine (`meta/` is Unvanquished-specific).

PR #602, opened 2022-03-08
Improve SDL joystick error handling and error reporting

necessarily-equal:
SDL_Init doesn't necessarily return -1, but any negative value on error.

There was a bit more changes initially, but illwieckz fixed the other issues :eyes:

### DIFF
--- a/src/engine/sys/sdl_input.cpp
+++ b/src/engine/sys/sdl_input.cpp
@@ -545,9 +545,9 @@ static void IN_InitJoystick()
 	{
 		Log::Debug( "Calling SDL_Init(SDL_INIT_JOYSTICK)..." );
 
-		if ( SDL_Init( SDL_INIT_JOYSTICK ) == -1 )
+		if ( SDL_Init( SDL_INIT_JOYSTICK ) < 0 )
 		{
-			Log::Debug( "SDL_Init(SDL_INIT_JOYSTICK) failed: %s", SDL_GetError() );
+			Log::Warn( "SDL_Init(SDL_INIT_JOYSTICK) failed: %s", SDL_GetError() );
 			return;
 		}
 
@@ -575,7 +575,7 @@ static void IN_InitJoystick()
 
 	if ( stick == nullptr )
 	{
-		Log::Debug( "No joystick opened." );
+		Log::Debug( "No joystick opened: %s.", SDL_GetError() );
 		return;
 	}
 


src/engine/sys/sdl_input.cpp
```diff
@@ -545,9 +545,9 @@ static void IN_InitJoystick()
 	{
 		Log::Debug( "Calling SDL_Init(SDL_INIT_JOYSTICK)..." );
 
-		if ( SDL_Init( SDL_INIT_JOYSTICK ) == -1 )
+		if ( SDL_Init( SDL_INIT_JOYSTICK ) < 0 )
 		{
-			Log::Debug( "SDL_Init(SDL_INIT_JOYSTICK) failed: %s", SDL_GetError() );
+			Log::Notice( "SDL_Init(SDL_INIT_JOYSTICK) failed: %s", SDL_GetError() );
```
illwieckz, line 550: Would it error out if there is no joystick, hence the non-usage of `Warn` to avoid an useless warning if something not required is missing ?

necessarily-equal:
I don't have a joystick and the message doesn't show up. I think it can 
only (without a bug) if the .so or .dll that contains the code can't be 
loaded by sdl.

So that got me thinking: is it likely to happen? I've looked a bit and 
it seems Debian, Arch, NixOS and Alpine have it in the base SDL 
package. So I've turned it into a warning.

necessarily-equal:
I've confirmed it's indeed the base package and not a SDL subpackage 
that holds the joystick code. The command belows specifically asks for 
SDL2 package only, without its plugins like SDL2_image or SDL2_gfx.

 > readelf --syms $(nix eval --raw nixpkgs#SDL2)/lib/libSDL2.so | grep 
-i joystick

   424: 000000000004be40     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickNumAxes
   442: 000000000004dfc0     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickIsVirtual
…
   739: 000000000004be80     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickUpdate
…
   897: 000000000004e0d0     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickHasLED
   918: 000000000004e0e0    16 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickSetLED
   925: 000000000004bbd0     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickIsHaptic
   927: 000000000004d770     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickGetP[...]
   933: 000000000004bed0     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickGetButton
   972: 000000000004be50     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickNumBalls
   978: 000000000004be60     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickNumHats
   984: 000000000004e100    12 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickRumb[...]
  1004: 000000000004be10     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickGetG[...]
  1017: 000000000004bdc0     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickOpen
  1024: 000000000004bea0     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickGetAxis
  1027: 000000000004dcb0     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickGetP[...]
  1031: 000000000004be70     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickNumB[...]
  1060: 000000000004d760     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickGetP[...]
  1069: 000000000004d7f0     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickGetD[...]
  1082: 000000000004d800     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickGetType
  1091: 000000000004bec0     6 FUNC    GLOBAL DEFAULT   10 
SDL_JoystickGetBall
… [there are 299 such lines]

So yeah, I think a warning is fine.

illwieckz (approved):
LGTM.

necessarily-equal:
I'll let you merge it

Issue #398, opened 2020-11-20
CMake issues with Visual Studio Ninja generator


slipher:
There is a Ninja-based generator for Visual studio, which is used if you create a project using Visual Studio's CMake integration. There are issues with it because the `CONFIG` generator expression has a different meaning than in other generators. Normally it is equal to one of the possible values for `CMAKE_BUILD_TYPE`. But in the VS Ninja generator, it is an arbitrary name.

This issue is not very important to solve, but I am writing some notes in case it comes up.

Possible workarounds:
1. Use an external CMake install to create VS projects with the classic generator (I am doing this).
2. Set `USE_PRECOMPILED_HEADERS` and `BUILD_GAME_NACL` to 0, which will avoid the two places where this generator expression variable is used.

It might be possible to fix by making variables to evaluate to whether the build type is equal to [foo] in generator expressions, which would be `$<CONFIG:foo>` or the result of `${CMAKE_BUILD_TYPE} EQ foo` depending on which generator is being used.

slipher:
I suppose the precompiled headers error is a moot point, since they don't work on Ninja anyway.

slipher:
Besides the NaCl bit there's one more use of CONFIG generator expressions - adding the `DEBUG_BUILD` define.

PR #603, opened 2022-03-10
Turn off precompiled headers by default as it is not handling recompilation properly

necessarily-equal:


slipher:
Proofs? How can you reproduce? I've never had such issues.

necessarily-equal:
Here's some example steps:

1. Build everything once
2. Remove half of daemon/src/common/Cvar.h in a way that the file 
compiles to a precompiled header, but everything else would be missing 
things from that file
3. Notice how it still builds
4. Notice that if you `make clean` then build again, it fails.

https://github.com/DaemonEngine/Daemon/issues/398 suggests that you are 
not using precompiled headers, which may be why you didn't notice yet.

After some time of wondering, I've got into an habbit of running `find 
~/unv/Unvanquished -name '*.h.gch' -delete` and modifying the file I 
changed (ex Math.h). Freem apparently has a similar command on his 
history, from what he said on IRC. So we are two to have noticed the 
problem independently.

slipher:
Need more details like which OS, compiler, and CMake generator (build system) you are using.

I use precompiled headers most of the time and have built on all of our supported operating systems.

necessarily-equal:
I'm building on Linux with GCC, ninja cmake backend, `cmake -D'BUILD_GAME_NACL=OFF' -D'BUILD_GAME_NACL_NEXE=OFF' -D'BUILD_GAME_NATIVE_EXE=OFF' -D'BUILD_GAME_NATIVE_DLL=ON' -GNinja -DCMAKE_BUILD_TYPE=Debug -Bbuild`. Hopefully this should be enough to reproduce

slipher:
OK, I was able to reproduce. The problem is Ninja-specific. Maybe Ninja fails to automatically detect header dependencies for the PCH.

You can change the line that says `if (APPLE)` to also check if CMAKE_GENERATOR or whatever is equal Ninja.

slipher:
https://cmake.org/cmake/help/latest/command/add_custom_command.html 

> Note that the IMPLICIT_DEPENDS option is currently supported only for Makefile generators and will be ignored by other generators.

That's the problem.

illwieckz:
I implemented per-generator PCH disablement instead:

- #610

PR #605, opened 2022-03-15
Avoid some crashes in cubemap code (reflection mapping).

illwieckz:
Reflection mapping isn't working yet, but those two patches avoid two crashes related to reflection mapping code.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -3948,7 +3948,8 @@ static void RB_RenderDebugUtils()
 		GL_PopMatrix();
 	}
 
-	if ( r_showCubeProbes->integer )
+	// GLSL shader isn't built when reflection mapping is disabled.
+	if ( r_showCubeProbes->integer && gl_reflectionShader != nullptr )
 	{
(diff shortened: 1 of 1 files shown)


src/engine/renderer/tr_backend.cpp
```diff
@@ -3955,6 +3955,13 @@ static void RB_RenderDebugUtils()
 		static const vec3_t mins = { -8, -8, -8 };
 		static const vec3_t maxs = { 8,  8,  8 };
 
+		/* GLSL shader isn't built,
+		likely because reflection mapping is disabled. */
+		if ( gl_reflectionShader == nullptr )
```
slipher, line 3960: How about skipping only the `if ( r_showCubeProbes->integer )` block, and not the whole rest of the function

illwieckz: Ah yes, I was cheated by those long blocks and I forgot they were not functions, also the next `return` cheated me. One day I'll probably turn those blocks into functions because that 1300 line function is just insane.

illwieckz: Fixed.

slipher:
LGTM

PR #606, opened 2022-03-15
Also load gl_forwardLightingShader_projXYZ when dynamic shadows are enabled to prevent a crash

illwieckz:
tr_shade: also load gl_forwardLightingShader_projXYZ when dynamic shadows are enabled to prevent a crash

Dynamic shadows are believed to be part of the old forward renderer,
and they are known to be very buggy.

Anyway commit message of a09f03bc8e775d83ac5e057593eff4e88cdea7eb
seems to imply they may be usable with tiled renderer as well:

> Use conventional shadow mapping code for inverse lights.
> This re-enables shadows for players in the tiled renderer.

After this commit no dynamic shadow is rendered but at least there
is no crash when setting to cg_shadows a value greater than 1 to
enable dynamic shadowing.

Note: for some reasons while the cvar is named r_shadows on renderer code,
the cvar is named cg_shadows in game console, probably because of some
quake3 legacy.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -50,14 +50,28 @@ static void GLSL_InitGPUShadersOrError()
 	// standard light mapping
 	gl_shaderManager.load( gl_lightMappingShader );
 
+	/* Deprecated forward renderer uses r_dynamicLight -1
+
+	Dynamic shadowing code also needs this shader.
+	This code is not well known, so there may be a bug,
+	but commit a09f03bc8e775d83ac5e057593eff4e88cdea7eb mentions this:
+
+	> Use conventional shadow mapping code for inverse lights.
+	> This re-enables shadows for players in the tiled renderer.
+	> -- @gimhael
+
+	See also https://github.com/DaemonEngine/Daemon/pull/606#pullrequestreview-912402293 */
+	if ( r_dynamicLight->integer < 0 || ( r_shadows->integer >= Util::ordinal(shadowingMode_t::SHADOWING_ESM16) ) )
+	{
+		// projective lighting ( Doom3 style )
+		gl_shaderManager.load( gl_forwardLightingShader_projXYZ );
+	}
+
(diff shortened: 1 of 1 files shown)


src/engine/renderer/tr_shade.cpp
```diff
@@ -50,15 +50,20 @@ static void GLSL_InitGPUShadersOrError()
 	// standard light mapping
 	gl_shaderManager.load( gl_lightMappingShader );
 
+	// Deprecated forward renderer uses r_dynamicLight -1
+	// Dynamic shadowing code also needs this shader.
```
slipher, line 54: To me it seems likely that it's a mistake that the code using the shader even runs (like there should have been an if (r_dynmaicLight == -1 around it). The comment could say something like "The shadowing code also accesses this shader. Possibly, that's a bug and it should only be used when the forward renderer is enabled."

illwieckz: That was my thought and I was gonna add another `r_dynamicLight < 0` test until I see @gimhael comment saying in https://github.com/DaemonEngine/Daemon/commit/a09f03bc8e775d83ac5e057593eff4e88cdea7eb

> This re-enables shadows for players in the tiled renderer

So it looks like this was intentional.

My priority is to fix the crash, I haven't seen shadows working in years anyway but we should not crash. I'll add a comment.

slipher: Maybe the change is intended to make cg_shadows 1 work in the tiled renderer but not higher values.

illwieckz: That's a good point!

On the other hand a09f03bc8e775d83ac5e057593eff4e88cdea7eb also modify light stuff, and if I'm right `cg_shadows 1` (blob shadows) are unrelated to lights, they're just a darker blob painted at player's feet…

slipher: OK, LGTM

illwieckz: I added a comment anyway.

PR #610, opened 2022-03-19
Turn off precompiled headers with Ninja generator

illwieckz:
The `IMPLICIT_DEPENDS` feature is currently unsupported with Ninja.

> Note that the IMPLICIT_DEPENDS option is currently supported only for Makefile generators and will be ignored by other generators.
> -- https://cmake.org/cmake/help/latest/command/add_custom_command.html

Supersedes:

- #603

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,6 +122,18 @@ option(USE_DEBUG_OPTIMIZE "Try to optimize the debug build" ON)
 if (APPLE)
     # Turn off by default because it is reportedly broken
     option(USE_PRECOMPILED_HEADER "Improve build times by using a precompiled header" OFF)
+elseif (CMAKE_GENERATOR EQUAL "Ninja")
+    # It isn't currently supported with Ninja yet:
+    #
+    # > Note that the IMPLICIT_DEPENDS option is currently supported only for
+    # > Makefile generators and will be ignored by other generators. Note that
+    # > the IMPLICIT_DEPENDS option is currently supported only for Makefile
+    # > generators and will be ignored by other generators.
+    # > -- https://cmake.org/cmake/help/latest/command/add_custom_command.html
+    #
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -122,6 +122,12 @@ option(USE_DEBUG_OPTIMIZE "Try to optimize the debug build" ON)
 if (APPLE)
     # Turn off by default because it is reportedly broken
     option(USE_PRECOMPILED_HEADER "Improve build times by using a precompiled header" OFF)
+elseif (CMAKE_GENERATOR EQUAL "Ninja")
+    # It isn's currently supported with Ninja yet. It would be better to be able to know
+    # if IMPLICIT_DEPENDS is supported by the current generator.
+    # > Note that the IMPLICIT_DEPENDS option is currently supported only for Makefile generators and will be ignored by other generators.Note that the IMPLICIT_DEPENDS option is currently supported only for Makefile generators and will be ignored by other generators.
```
slipher, line 128: Line break?

Otherwise LGTM

necessarily-equal:
This doesn't seem to work unfortunately

```shell
$ rm -r build
$ cmake -D'BUILD_GAME_NACL=OFF' -D'BUILD_GAME_NACL_NEXE=OFF' -D'BUILD_GAME_NATIVE_EXE=OFF' -D'BUILD_GAME_NATIVE_DLL=ON' -GNinja -DCMAKE_BUILD_TYPE=Debug -DDAEMON_CBSE_PYTHON_PATH=$(which python) -DCMAKE_CXX_FLAGS="" -Bbuild
$ ccmake -Bbuild
 …
 USE_PRECOMPILED_HEADER           ON
 …
```

and cmake actually creates the pch files on compilation.

To be honest, since precompiled header totally break ccache, they actually make the compilation time much longer on my end, so I wonder if they are really worth it.

PR #611, opened 2022-03-21
Compat-breaking deletions for 0.53

slipher:


illwieckz:
I was thinking about something these days: do some of them break demo? Phoronix reintegrated Unvanquished in their benchmark library and use Unvanquished in their official benchmark (cool way to remind we exist), I would prefer if we don't break this. I also wonder how much such demo is tied to our game code or unvanquished dpk version anyway…

slipher:
A demo requires loading the cgame which the recorded player was using, so yes engine ABI changes break compat with demos. We have a lot of ABI changes on every release (including 0.53 just counting stuff that's already merged), so it's not a realistic option to keep compatibility with every old engine version.

illwieckz:
Is it possible to keep compatibility with 0.52 demo for some time? This page says PTS unvanquished benchmark runs 0.52.1, so I assume the demo Michael Larrabel created is 0.52 compatible: https://openbenchmarking.org/test/pts/unvanquished-1.6.0
It would be a shame to invalidate this.

If I'm right before 0.52 Michael Larrabel stopped doing Unvanquished benchmarks because of compatibility problem… I don't want to make him angry at us by wasting his effort.

slipher:
Keeping ABI compatibility is echnically possible, but it would be a real burden. The most annoying part would probably be the syscalls that have struct parameters. For example, you recently wanted to delete some fields from `glconfig_t`. But if we want to maintain compatibility, then we must never add or delete fields from the struct. The only way would be to make a 2nd copy of the struct definition and add a new syscall using the 2nd struct definition.

Besides that, we would need to avoid deleting any syscalls (except ones that were never used). And if we wanted to change the arguments of a syscall, we'd have to make a 2nd version of the syscall and implement 2 versions of it. Finally none of the IPC framework could ever be changed (e.g. serialization code, or using WASM instead of NaCl).

illwieckz:
Hmmm… Things like the `glconfig_t` can be kept for compatibility (there is already an `unused` field like that). I'm not looking for an infinite support, but maybe not break it on every version.

On the other hand, we may want to provide a demo benchmarkers may rely on, making sure it always work at release time. And then it would be useful to have some tools to assist the creation of demos for such purpose.

illwieckz:
I wrote a script to record a demo purposed for benchmark, so we can upload a new demo at release time and not focus on compatibility.

illwieckz (approved):
LGTM.

PR #612, opened 2022-03-21
Delay GLSL shader build until first map load.

illwieckz:
This makes possible for the user to quickly reach the game main menu without building all GLSL shaders and set a low graphics preset before joining a game and loading a map.

Doing so prevents building unwanted or unsupported GLSL shaders on slow and/or old hardware and drastically reduce first startup time.

On Unvanquished side, the UI was already modified to display a message to the user to make him wait for the GLSL shaders being compiled at map load:

- https://github.com/Unvanquished/Unvanquished/pull/1470

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -1107,7 +1107,16 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_exportTextures = ri.Cvar_Get( "r_exportTextures", "0", 0 );
 		r_heatHaze = ri.Cvar_Get( "r_heatHaze", "1", CVAR_LATCH | CVAR_ARCHIVE );
 		r_noMarksOnTrisurfs = ri.Cvar_Get( "r_noMarksOnTrisurfs", "1", CVAR_CHEAT );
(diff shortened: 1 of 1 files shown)


src/engine/renderer/tr_init.cpp
```diff
@@ -1107,7 +1107,16 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_exportTextures = ri.Cvar_Get( "r_exportTextures", "0", 0 );
 		r_heatHaze = ri.Cvar_Get( "r_heatHaze", "1", CVAR_LATCH | CVAR_ARCHIVE );
 		r_noMarksOnTrisurfs = ri.Cvar_Get( "r_noMarksOnTrisurfs", "1", CVAR_CHEAT );
-		r_lazyShaders = ri.Cvar_Get( "r_lazyShaders", "0", 0 );
+
+		/* Delay GLSL shader build until first map load.
```
slipher, line 1111: It's not clear that this comment is talking about the effect of the "1" value of the cvar.

Otherwise LGTM

illwieckz: I updated the comment. I also switched target branch to `master` as it is compatible with master and it makes easier to test it with public games.

PR #614, opened 2022-03-22
cl_main: rename demo file

illwieckz:
Name demo file as date_time_host_map
instead of host_map_date_time.

This make it easy to sort by date in file browser.
Also, better use _ as field separator in all cases.

### DIFF
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -392,11 +392,11 @@ std::string GenerateDemoName()
         map_name.erase(0, last_slash + 1);
 
     return Str::Format(
-        "%s-%s_%04i-%02i-%02i_%02i%02i%02i",
-        NET_AdrToString(clc.serverAddress),
-        map_name,
+        "%04i-%02i-%02i_%02i%02i%02i_%s_%s",
         1900 + time.tm_year, time.tm_mon + 1, time.tm_mday,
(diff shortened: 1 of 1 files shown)


slipher:
> This make it easy to sort by date in file browser.

Why wouldn't you sort by date to sort by date?

illwieckz:
One problem is that creation date used by file browsers to sort files is a metadata that is not part of the file neither the file name, and it can be lost on copy and transfer.

Another problem is that demos are currently listed by IP address when sorted alphabetically, which is probably the worst thing to do.

I usually name all my important files starting with date the `YYYYmmdd-HHMMSS` way and it's very convenient. Especially, I make it for everything I record: photos, audio, video… I see demos as recording, the primary reliable information, even before knowing _where_ and _how_, is the date. Because time is sequential, a single source can't produce documents at the same time in multiple places, so places, events, people, etc. may be reconstructed from the date.

So, I feel like starting with the date is more useful than some other things, especially IP addresses. If a player records demo while roaming across various random servers in the same day, there is high chance such user prefers to see such demos being grouped by date and time.

I'm not very concerned by screenshots being named the `gamename_date` way because the game isn't expected to change, so the prefix does not prevent to sort by date by sorting alphabetically, but if the prefix was variable I would have submitted a similar patch in that PR to make it a suffix.

slipher:
LGTM

PR #616, opened 2022-03-22
Add cvar to shutdown on drop, fix #487

### LINKED ISSUE #487: Option to quit the game after a demo is played
To make Dæmon-based games more benchmark friendly, it would be good to have a convenient way to shutdown the game at the end of a demo playing.

See https://github.com/Unvanquished/Unvanquished/issues/1440

illwieckz:
Add cvar to shutdown on drop, fix #487.

By setting `common.shutdownOnDrop` to `on`, the engine will quit when the game quit. This is useful to get the engine quit after having timed a demo for benchmarking purpose.

See:

- #487
- https://github.com/Unvanquished/Unvanquished/issues/1440#issuecomment-864646180

### DIFF
--- a/src/engine/framework/System.cpp
+++ b/src/engine/framework/System.cpp
@@ -52,6 +52,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 namespace Sys {
+Cvar::Cvar<bool> cvar_common_shutdownOnDrop("common.shutdownOnDrop", "shut down engine on game drop", Cvar::TEMPORARY, false);
 
 #ifdef _WIN32
 static HANDLE singletonSocket;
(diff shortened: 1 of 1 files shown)


illwieckz:
I named it `common.shutdownOnDrop` because we already have `common.pedanticShutdown`, but this code does not live in `common` folder.

Maybe we would prefer naming them `sys.shutdownOnDrop` and `sys.pedanticShutdown` instead? They both live in `Sys` namespace.

src/engine/framework/System.cpp
```diff
@@ -52,6 +52,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 namespace Sys {
+Cvar::Cvar<bool> shutdownOnDrop("common.shutdownOnDrop", "shutdown engine on game drop", Cvar::TEMPORARY, false);
```
slipher, line 55: Grammar nit: `shut down engine on game drop`

src/engine/framework/System.cpp
```diff
@@ -748,6 +749,11 @@ ALIGN_STACK_FOR_MINGW int main(int argc, char** argv)
 					Log::Notice(err.what());
 				}
 				Application::OnDrop(err.is_error(), err.what());
+
+				if (Sys::shutdownOnDrop.Get()) {
+					Sys::Shutdown(false, "Shutdown on drop");
+					Sys::OSExit(0);
```
slipher, line 755: It might be useful to make the return code different based on `err.is_error()`

slipher (commented):
The general idea seems good to me.

> but this code does not live in common folder.

Neither do many of the other `common.` cvars.

src/engine/framework/System.cpp
illwieckz: See my new commit.

slipher:
LGTM

PR #618, opened 2022-03-23
add some time units to `delay` command

illwieckz:
Add time units like `s` (second), `m` (minute), `h` (hour), so someone can do:

```
delay 10s say "10 second has passed!"
```

Also improve a bit the builtin help.

### DIFF
--- a/src/engine/framework/BaseCommands.cpp
+++ b/src/engine/framework/BaseCommands.cpp
@@ -641,7 +641,10 @@ namespace Cmd {
                 int argc = args.Argc();
 
                 if (argc < 3) {
-                    PrintUsage(args, "delay (name) <delay in milliseconds> <command>\n  delay (name) <delay in frames>f <command>", "executes <command> after the delay" );
+                    PrintUsage(args, "(name) <delay>(unit) <command>\n"
+                        "  executes <command> after the delay,\n"
+                        "  no unit means a time in millisecond,\n"
+                        "  unit can be ms (millisecond), s (second), m (minute), h (hour), f (frame)");
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


illwieckz:
I added an extra commit with another way of writing things, what do you think about it? I can drop or squash this extra commit it I don't mind.

necessarily-equal:
I'm not fond that it breaks if you pass it "100s " or something similar, or that it gives something quite wrong for "10ms". If I read the code correctly, "10ms" would be parsed as "10s". It's kinda a pitfal with the current code, but if I think 10ms would have been parsed like "10" before, and thus parsed correctly.

I prefer the first implementation because it is more easily extensible by the way. The other one is a bit hacky to be honest.

illwieckz:
```
]/delay 10ms say hello 
delay: the delay must be a positive integer 
]/delay "10s " say hello 
delay: optional unit suffixes are f (frame), s (second), m (minute), h (hour) 
```

First one is because since `s` stripped, `10m` is attempted to be parsed and it is not a number. I may extend the tests to print better messages, but at least it is not parsed as `10s`.

illwieckz:
@necessarily-equal I improved the code to catch unsupported two-chars units like `ms`. In fact my “second implementation” made it very easy to avoid duplicate code. I'll probably keep going this way.

illwieckz:
Note that I can easily add optional `ms` suffix, even if it not needed, for those who likes to be explicit.

illwieckz:
I added a commit for optional `ms` suffix support, we don't need it, it just allows people to be explicit about the unit in their scripts.

src/engine/framework/BaseCommands.cpp
```diff
@@ -650,10 +653,45 @@ namespace Cmd {
                 const std::string& name = isNamed ? args.Argv(1) : "";
                 const std::string& command = args.EscapedArgs(2 + +isNamed);
                 std::string delay = args.Argv(1 + +isNamed);
-                bool frames = delay.back() == 'f';
 
-                if (frames) {
-                    delay.erase(--delay.end()); //FIXME-gcc-4.6 delay.pop_back()
+                const char front = delay.front();
```
slipher, line 657: This may read outside the array in case of an empty string. Same goes for some other places reading `back()`

src/engine/framework/BaseCommands.cpp
```diff
@@ -662,10 +700,24 @@ namespace Cmd {
...
+                            target *= 60;
+                            DAEMON_FALLTHROUGH;
+                        case delayUnit::MINUTE:
+                            target *= 60;
+                            DAEMON_FALLTHROUGH;
+                        case delayUnit::SECOND:
+                            target *= 1000;
+                            DAEMON_FALLTHROUGH;
+                        default:
+                            ;
```
slipher, line 718: `break;`

src/engine/framework/BaseCommands.cpp
```diff
@@ -650,10 +653,45 @@ namespace Cmd {
...
+
+                switch(delayUnit(back)) {
+                    case delayUnit::SECOND:
+                    case delayUnit::MINUTE:
+                    case delayUnit::HOUR:
+                    case delayUnit::FRAME:
+                        unit = delayUnit(back);
+                        delay.erase(--delay.end()); //FIXME-gcc-4.6 delay.pop_back()
+                        back = delay.back();
+                        if (back == 'm' && unit == delayUnit::SECOND)
```
slipher, line 683: How about `Str::IsSuffix` here

illwieckz:
I believe I addressed comments, I'm not sure using `Str::IsSuffix` would produce shorter code.

src/engine/framework/BaseCommands.cpp
```diff
@@ -650,10 +653,45 @@ namespace Cmd {
...
+                    SECOND = 's',
+                    MINUTE = 'm',
+                    HOUR = 'h',
+                    FRAME = 'f',
+                };
+
+                delayUnit unit = delayUnit::MILLISECOND;
+                char back = !!delay.size() ? delay.back() : '\0';
+
+                switch(delayUnit(back)) {
```
necessarily-equal, line 675: I find this more readable, otherwise I'm looking everywhere where the function "delayUnit" is declared
```suggestion
                switch(static_cast<delayUnit>(back)) {
```

src/engine/framework/BaseCommands.cpp
necessarily-equal: I agree with slipher that this would be more readable.

necessarily-equal (approved):
I've added two style comments but ultimately it's good enough for a merge IMO, assuming you tested it a bit

src/engine/framework/BaseCommands.cpp
illwieckz: I don't think it would be more readable, at this point the `s` of `ms` is already stripped out. We can do `Str::IsSuffix` before stripping it anyway, but in all cases that still look weird to me when half the code uses one way to do it and other half doing it another way. Here we just do “we already detected unit is second, we now look at multiplier”.

illwieckz:
Yep I tested any combinations possible.

PR #619, opened 2022-03-24
common: make demo.timedemo cvar temporary

illwieckz:
That cvar has very specific usage and we may want to prevent it to be stored. People making use of it know what they do, and they're usually scripting it already.

Also timedemo being enabled has side effects when not playing demo (like defeating the fps cap), though, the cvar being cheat prevents this to happen in regular games.

This kind of cvar being stored can bite in the back at some unexpected moment.


necessarily-equal (approved):
sure

PR #620, opened 2022-04-01
Fix command forwarding on Linux + Breakpad

### LINKED ISSUE #78: broken `Forwarding commands to existing instance` feature
the “_Existing instance found / Forwarding commands to existing instance_” feature looks to be broken on `for-0.51.0` branch.

slipher:
The socket listener thread was being started before the Breakpad
initialization which forks. Unusually, the exec() is done in the parent
process. As a result any non-main threads are killed by the forking.

Fixes #78.

### DIFF
--- a/src/engine/framework/System.cpp
+++ b/src/engine/framework/System.cpp
@@ -671,22 +671,26 @@ static void Init(int argc, char** argv)
 		OSExit(0);
 	}
 
-	// Create the singleton socket and a thread to watch it
+	// Create the singleton socket
 	CreateSingletonSocket();
-	try {
-		std::thread(ReadSingletonSocket).detach();
-	} catch (std::system_error& err) {
-		Sys::Error("Could not create singleton socket thread: %s", err.what());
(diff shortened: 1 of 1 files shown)


illwieckz (approved):
I find the usage of `may` verb in “so no threads may be created before this point” sentence ambiguous there with meaning being too much dependent on context, either we consider the code creates a thread or we consider a coder create a threads by writing code, and then the verb is either about the code produced by the coder or the verb is about the coder. I would prefer another wording that is more explicit about the actual advice to never create a thread before this point, otherwise LGTM about the logic.

Edit: `may` is also ambiguous in the way it looks like `may not` can carry either the meaning of physical impossibility (something you cannot do even if you try) or moral impossibility (something you don't want to do).

slipher:
Changed to `must not`.

illwieckz:
Good! This patch is really wanted. 🙂️

illwieckz (approved):
LGTM.

PR #623, opened 2022-04-11
Fix DLL unloading on Linux

slipher:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1792

### DIFF
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -197,6 +197,10 @@ else()
     # Use hidden symbol visibility if possible
     try_c_cxx_flag(FVISIBILITY_HIDDEN "-fvisibility=hidden")
 
+    # Prevent the generation of STB_GNU_UNIQUE symbols on templated functions on Linux.
+    # STB_GNU_UNIQUE renders dlclose() inoperative.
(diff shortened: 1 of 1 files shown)


necessarily-equal:
Some of these apparently come from crunch:

`daemon/libs/crunch/crnlib/crn_math.h`
```C++
inline uint count_trailing_zero_bits(uint v) {
  uint c = 32;  // c will be the number of zero bits on the right

  static const unsigned int B[] = {0x55555555, 0x33333333, 0x0F0F0F0F, 0x00FF00FF, 0x0000FFFF};
  static const unsigned int S[] = {1, 2, 4, 8, 16};  // Our Magic Binary Numbers
```

Out of curiosity, how did you find it? Is this useful for the dll bin or for the engine?

If you've tested it, LGTM.

slipher:
I recently fixed the same problem (dlclose doesn't unload the DLL) for Mac and recognized the symptom of duplicate emoticon warnings. After confirming that was the problem, I googled for "dlclose no effect" or something like that and learned about STB_GNU_UNIQUE. And ran the commands suggested to list STB_GNU_UNIQUE symbols (I didn't see anything from Crunch, presumably because it was optimized out in a release build).

The flag is only needed for objects that go into DLLs. I just applied it to everything because that's easiest and has no ill effects.

PR #624, opened 2022-04-23
Only center the mouse if we get a mouse event with relative motion

slouken:
For more information, see the discussion in https://github.com/libsdl-org/SDL/issues/5569

Fixes https://github.com/DaemonEngine/Daemon/issues/600

### DIFF
--- a/src/engine/sys/sdl_input.cpp
+++ b/src/engine/sys/sdl_input.cpp
@@ -1083,7 +1083,7 @@ static void IN_ProcessEvents( bool dropInput )
 					{
 						Com_QueueEvent( Util::make_unique<Sys::MouseEvent>(e.motion.xrel, e.motion.yrel) );
 #if defined( __linux__ ) || defined( __BSD__ )
-						if ( !in_nograb->integer )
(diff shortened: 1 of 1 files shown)


smcv:
This would probably allow #621 to be reverted, too.

slipher:
@necessarily-equal claimed this will not fix it when I suggested the same. https://github.com/DaemonEngine/Daemon/pull/621#issuecomment-1086951636

illwieckz:
Even if it may be not enough (are you sure?), maybe this is still wanted anyway? @slipher @necessarily-equal

illwieckz:
```
<illwieckz> perturbed, do you thing something prevents that to be merged? https://github.com/DaemonEngine/Daemon/pull/624
<illwieckz> even if it's not 100% better, if it's already 50% better we may still merge it?
<perturbed> i don't have a problem
<afontain> I'd say we should merge it
<illwieckz> okok
```

illwieckz:
Thank you @slouken! 👍️👍️👍️👍️👍️

slouken:
You're welcome!

necessarily-equal:
NixOS 22.05 updated to SDL2.0.20, I can confirm that this PR fixes the issue

PR #625, opened 2022-05-01
disable geoip by default, remove it as release build requirement, fix #492

### LINKED ISSUE #492: libgeoip legal, licencing and privacy concerns
This issue is not per-se a technical one, even if I think technical answers can help. I am sorry that I can't find a better title for it.

My concerns are about a change in maxmind's geoip database licence, which makes it no longer FOSS.
This change only affects GeoLite2, not "legacy GeoLite" from my understanding. Naive answer could be "don't care, old DB is still under correct licence", but according to Maxmind, they changed licence to comply with some country's laws, which were made to protect people's privacy.

Currently, daemon build the libgeoip feature by default, it's opt-out at build time. There is no way for a client to disable the feature, same for server hosters.
The only use of it is to show a message like "$player connected from $country".

The questions it brings are:

* what does it means for server hosters to have this feature enabled (legally speaking)?
* should it be made it runtime-optional, through a variable which would be exposed in unvanquished's GUI (or any other game based on daemon, ofc) and a Cvar for hosters?
* should the option (either build-time, run-time or both) be opt-in or opt-out? Opt-in is the easier way to have people benefiting from it.
* what about the future? The DB is likely no longer maintained.
* is that feature worth the troubles at all?

illwieckz:
Disable geoip by default, remove it as release build requirement, fix #492

After this commit support for GeoIP will not be built by default, meaning libgeoip will not be a requirement for release builds anymore.

Since libgeoip will not be a release build requirement, we would be able to remove libgeoip from prebuilt dependencies.

The geoip code is still there and people can still build their own server with geoip support enabled, but that's up to them to decide if it's good or not and it's up to them to deal with the questions regarding to the geoip database.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,7 @@ if(NOT DAEMON_EXTERNAL_APP)
         option(USE_CURSES "Enable fancy colors in terminal's output" ON)
     endif()
     cmake_dependent_option(USE_SMP "Compile with support for running the renderer in a separate thread" 1 BUILD_CLIENT 0)
-    cmake_dependent_option(USE_GEOIP "Allows server to show from which coutry a player logs from" 1 "BUILD_SERVER OR BUILD_CLIENT OR BUILD_TTY_CLIENT" 0)
+    option(USE_GEOIP "Allows server to show from which coutry a player logs from" OFF)
     option(USE_BREAKPAD "Generate Daemon crash dumps (which require Breakpad tools to read)" OFF)
 endif()
 


illwieckz:
Or we can just nuke it entirely:

https://github.com/slipher/Daemon/commit/37c5527471844e3921b3ae9e080fbeb58fba2338

illwieckz:
I prefer to just disable first because I don't want to rebase #561.

illwieckz:
I believe myself, @slipher, @bmorel and @necessarily-equal are OK to disable or nuke GeoIP.

PR #627, opened 2022-05-25
Nuke GeoIP

### LINKED ISSUE #492: libgeoip legal, licencing and privacy concerns

slipher:
Fixes #492. I noticed that since the last GeoIP release was in 2018, it is either not working at all, or we are using an ancient database (since no one actually knows how to configure an external database). After realizing this it seems better to nuke it right away instead of after a release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,6 @@ if(NOT DAEMON_EXTERNAL_APP)
         option(USE_CURSES "Enable fancy colors in terminal's output" ON)
     endif()
     cmake_dependent_option(USE_SMP "Compile with support for running the renderer in a separate thread" 1 BUILD_CLIENT 0)
-    option(USE_GEOIP "Allows server to show from which coutry a player logs from" OFF)
     option(USE_BREAKPAD "Generate Daemon crash dumps (which require Breakpad tools to read)" OFF)
 endif()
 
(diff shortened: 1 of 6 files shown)


necessarily-equal:
sounds fine

superdupont-fr:
And, this GeoIP was used for what purpose?

illwieckz:
LGTM.

PR #629, opened 2022-05-26
New external_deps build

slipher:
This is WIP - so far I just tested the build script and the game on Mac.

I included some of the commits from #561 that weren't too entangled with the package registration stuff. @illwieckz Let me know if there are any other commits there that would be important to include for the new build.


illwieckz:
I'm gonna merge #561 in all cases, so cherry-picking instead of rebasing makes everyone's life harder with the only benefit of increasing the risk of introducing mistakes.

The M1 mac stuff are very welcome anyway ! 👍️

slipher:
> I'm gonna merge https://github.com/DaemonEngine/Daemon/pull/561 in all cases

Why? AFAIK no one else said they liked the package registration stuff.

By the way, I found that the JPEG NASM stuff still works without any changes. CMAKE_ASM_NASM_COMPILER or whatever gets populated from the NASM environment variable.

illwieckz:
> Why? AFAIK no one else said they liked the package registration stuff.

No one else expressed any opinion against it as well.

And it fullfills my need of a script that acts consistently with the
documentation by design because the documentation is produced by the same
code that does the actions.

slipher:
Usually we consider that a pull request should have a favorable review before being merged. More people should be for the change than against it.

necessarily-equal (approved):

external_deps/build.sh
```diff
@@ -5,34 +5,33 @@
 set -e
 set -u
 
-# Dependencies version. This number must be updated every time the version
-# numbers below change, or packages are added/removed.
+# This should match the DEPS_VERSION in CMakeLists.txt.
+# This is mostly to ensure the path the files end up at if you build deps yourself
+# are the same as the ones when extracting from the downloaded packages.
 DEPS_VERSION=5
```
necessarily-equal, line 11: Also I think you want to set this to 6.

slipher: That can only be done after the new builds are on the server. I want to merge first before doing the final builds.

necessarily-equal:
Sure, I'm fine with a merge first too

illwieckz:
I'll have to test this for producing both Linux, Windows and macOS deps archive because that was the initial purpose of starting such work: I was never able to do it myself before and if I can't it means I cannot know if someone else can.

That's also why I'm in favor of turning the version to `6` before merging because by doing that, one has to generate the version `6` deps archives and re-run the checks of this PR on them, meaning we would have checked that this code can generate archives our CI can validate.

If we turn to `6` _after_ the merge, we would have merged the code without any checks.

external_deps/build.sh
```diff
@@ -465,6 +464,7 @@ build_wasisdk() {
 		local WASISDK_PLATFORM=linux
 		;;
 	esac
+	local WASISDK_VERSION_MAJOR="$(echo "${WASISDK_VERSION}" | cut -f1 -d'.')"
```
bmorel, line 454: What's the point of having wasm related things in master since there is currently *no* support for it? Does not it just bloat the system?

bmorel (approved):
I have not tested it, as a matter of fact, and I'm wondering why there's stuff related to code we don't have yet, but otherwise it seems ok.

LGTM.

external_deps/build.sh
slipher: It's not included in the prebuilt packages, so it does not bloat the system.

slipher:
All right, I uploaded the new packages and changed the version to 6.

illwieckz:
I get this on macOS:

```
../unvanquished-release-scripts/build-release -j$(nproc) macos-amd64
[…]
make[2]: *** No rule to make target `external_deps/macosx64-5/lib/libgmp.a', needed by `daemonded'.  Stop.
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [daemon_build/CMakeFiles/ttyclient.dir/all] Error 2
make[1]: *** [daemon_build/CMakeFiles/client.dir/all] Error 2
make: *** [all] Error 2
ERROR INTERNAL: macos-amd64 build failed
```

illwieckz:
Ah it's a bug in `build-release`…

PR #634, opened 2022-06-12
Correct relative mouse input handling

necessarily-equal:
Closes https://github.com/DaemonEngine/Daemon/issues/600. Hopefully for good this time.

### DIFF
--- a/src/engine/sys/sdl_input.cpp
+++ b/src/engine/sys/sdl_input.cpp
@@ -1082,15 +1082,6 @@ static void IN_ProcessEvents( bool dropInput )
 					else
 					{
 						Com_QueueEvent( Util::make_unique<Sys::MouseEvent>(e.motion.xrel, e.motion.yrel) );
-#if defined( __linux__ ) || defined( __BSD__ )
-						if ( !in_nograb->integer && ( e.motion.xrel || e.motion.yrel ) )
-						{
-							// work around X window managers and edge-based workspace flipping
-							// - without this, we get LeaveNotify, no mouse button events, EnterNotify;
-							//   we would likely miss some button presses and releases.
-							IN_CenterMouse();
-						}
-#endif
 					}
 				}
 				break;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -615,21 +615,6 @@ endif()
 # SDL, required for all targets on win32 because of iconv and SDL_SetHint(SDL_TIMER_RESOLUTION, 0)
 if (BUILD_CLIENT OR WIN32)
     find_package(SDL2 REQUIRED)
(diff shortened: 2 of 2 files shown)


necessarily-equal:
I've tested this to work on SDL2.0.20, SDL2.0.22 and SDL main branch (as of https://github.com/libsdl-org/SDL/commit/3807e95029d77a480e80adddd7c0f21abc29baa9).

Testing with old SDL versions is left as an exercise for the reader ;-)

slipher:
This patch breaks Mac by making it impossible to move the mouse outside the window. This is true with both SDL 2.0.12 and latest SDL.

src/engine/sys/sdl_input.cpp
```diff
@@ -427,7 +427,7 @@ void IN_SetMouseMode(MouseMode newMode)
 		}
 
 		SDL_ShowCursor( newMode == MouseMode::SystemCursor ? SDL_ENABLE : SDL_DISABLE );
-		auto grab = newMode != MouseMode::Deltas || in_nograb->integer ? SDL_FALSE : SDL_TRUE;
+		auto grab = in_nograb->integer ? SDL_FALSE : SDL_TRUE;
```
necessarily-equal, line 430: @slipher do it work for you if I remove this change?

slipher: Yes, the bug goes away upon reverting that part.

necessarily-equal: Great, reverting that part then.

necessarily-equal:
@bmorel since you seem to use a minimalist X11 setup, you may want to try to see if this PR don't introduce some regression on your setup.

```C
// work around X window managers and edge-based workspace flipping
// - without this, we get LeaveNotify, no mouse button events, EnterNotify;
//   we would likely miss some button presses and releases.
IN_CenterMouse();
```

I'd say you don't really have to try it either, you can also complain afterwards ;)

slipher:
The workspace flipping thing described in the comment seems to be an xfce behavior. https://askubuntu.com/questions/891859/xubuntu-snap-to-window-switching-workspaces

necessarily-equal:
Tested working by @IngarKCT. It also works for me and it fix a problem that totally breaks the game, so I'm merging this.

I'll have a look at the XFCE situation a bit later, to check if it is still a problem.

necessarily-equal:
I've checked XFCE and I've ticked that thing in the parameters. I 
couldn't reproduce the bug with the fixed daemon.

necessarily-equal:
@Viech it would be nice if you could backport this to the AUR

This patch should* apply cleanly [sdl.patch.txt](https://github.com/DaemonEngine/Daemon/files/8935634/sdl.patch.txt) (although it was hand-edited, and is untested)

Issue #632, opened 2022-06-12
`/set common.framerate.maxUnfocused 1` broke on 0.53.0/sync


necessarily-equal:

slipher:
Works for me. Note that just moving the mouse out of the window doesn't necessarily remove the focus. You need to click on some other application.

PR #635, opened 2022-06-13
Use normal max FPS without unfocused/minimized set

slipher:
A regression in 443bf08c would cause the FPS to be unlimited when
unfocused/minimized with default settings.


necessarily-equal (approved):
Looks like this fixes https://github.com/DaemonEngine/Daemon/issues/632 for me

PR #637, opened 2022-06-27
transforms: Implement Push/PopMatrix.

DolceTriade:
This implements transformations required to support RmlUI's transform
code. This implements the cgame plumbing and the renderer code.

### DIFF
--- a/src/engine/renderer/gl_shader.h
+++ b/src/engine/renderer/gl_shader.h
@@ -117,9 +117,9 @@ protected:
 	std::vector< GLUniformBlock * > _uniformBlocks;
 	std::vector< GLCompileMacro * > _compileMacros;
 
-	
 
-	
+
+
 
 	GLShader( const std::string &name, uint32_t vertexAttribsRequired, GLShaderManager *manager ) :
 		_name( name ),
@@ -797,6 +797,7 @@ protected:
 	  USE_SHADOWING,
 	  LIGHT_DIRECTIONAL,
 	  USE_DEPTH_FADE,
+	  USE_DEPTH_FLATTEN,
 	  USE_PHYSICAL_MAPPING,
 	  USE_ALPHA_TESTING
 	};
@@ -1216,6 +1217,31 @@ public:
 	}
 
 	void SetDepthFade( bool enable )
+	{
+		SetMacro( enable );
+	}
+};
+
+class GLCompileMacro_USE_DEPTH_FLATTEN :
+	GLCompileMacro
+{
+public:
+	GLCompileMacro_USE_DEPTH_FLATTEN( GLShader *shader ) :
+		GLCompileMacro( shader )
+	{
+	}
+
+	const char *GetName() const
+	{
+		return "USE_DEPTH_FLATTEN";
+	}
+
+	EGLCompileMacro GetType() const
+	{
+		return USE_DEPTH_FLATTEN;
+	}
+
+	void SetDepthFlatten( bool enable )
 	{
 		SetMacro( enable );
 	}
@@ -2215,6 +2241,7 @@ class GLShader_generic :
 	public GLCompileMacro_USE_TCGEN_ENVIRONMENT,
 	public GLCompileMacro_USE_TCGEN_LIGHTMAP,
 	public GLCompileMacro_USE_DEPTH_FADE,
+	public GLCompileMacro_USE_DEPTH_FLATTEN,
 	public GLCompileMacro_USE_ALPHA_TESTING
 {
 public:
--- a/src/engine/renderer/gl_shader.h
+++ b/src/engine/renderer/gl_shader.h
@@ -797,7 +797,6 @@ protected:
 	  USE_SHADOWING,
 	  LIGHT_DIRECTIONAL,
 	  USE_DEPTH_FADE,
-	  USE_DEPTH_FLATTEN,
 	  USE_PHYSICAL_MAPPING,
 	  USE_ALPHA_TESTING
 	};
@@ -1217,31 +1216,6 @@ public:
 	}
 
 	void SetDepthFade( bool enable )
-	{
-		SetMacro( enable );
-	}
-};
-
-class GLCompileMacro_USE_DEPTH_FLATTEN :
-	GLCompileMacro
-{
-public:
-	GLCompileMacro_USE_DEPTH_FLATTEN( GLShader *shader ) :
-		GLCompileMacro( shader )
-	{
-	}
-
-	const char *GetName() const
-	{
-		return "USE_DEPTH_FLATTEN";
-	}
-
-	EGLCompileMacro GetType() const
-	{
-		return USE_DEPTH_FLATTEN;
-	}
-
-	void SetDepthFlatten( bool enable )
 	{
 		SetMacro( enable );
 	}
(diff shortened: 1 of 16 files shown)


necessarily-equal:
So, huh, I tried using it and it didn't quite do what I expected:

Screenshot from 2022-06-28 19-48-46

Screenshot from 2022-06-28 19-49-06

I think I may be missing a branch in Unvanquished to add support for transforms? At least there is no open PR on that repo.

src/engine/renderer/tr_backend.cpp
```diff
@@ -5103,6 +5103,30 @@ const RenderCommand *Poly2dIndexedCommand::ExecuteSelf( ) const
...
+bool pushed = false;
+const RenderCommand *PushMatrixCommand::ExecuteSelf() const {
+	Tess_End();
+	if (!pushed) {
+		GL_PushMatrix();
+		pushed = true;
+	}
+	MatrixCopy( glState.projectionMatrix[ glState.stackIndex - 1 ],
+	            glState.projectionMatrix[ glState.stackIndex ] );
+	MatrixCopy( this->matrix,
```
slipher, line 5115: I think GL_LoadModelViewMatrix could be used

src/engine/renderer/tr_backend.cpp
```diff
@@ -5103,6 +5103,30 @@ const RenderCommand *Poly2dIndexedCommand::ExecuteSelf( ) const
 	return this + 1;
 }
 
+bool pushed = false;
+const RenderCommand *PushMatrixCommand::ExecuteSelf() const {
+	Tess_End();
+	if (!pushed) {
```
slipher, line 5109: Why is this `pushed` variable needed?

src/shared/client/cg_api.cpp
```diff
@@ -315,6 +315,17 @@ void trap_R_Add2dPolysIndexedToScene( polyVert_t* polys, int numPolys, int* inde
 	cmdBuffer.SendMsg<Render::Add2dPolysIndexedMsg>(mypolys, numPolys, myindices, numIndexes, trans_x, trans_y, shader);
 }
 
+void trap_R_PushMatrix( const matrix_t matrix )
```
slipher, line 318: Can we get some comments about what it's for? Does it only work for 2D rendering?

DolceTriade:
@necessarily-equal You need the ishq/rmluitransforms branch in Unvanquished. I didn't create a PR for Unvanquished yet because it depends on this change first.

src/engine/renderer/tr_backend.cpp
DolceTriade: Done.

bmorel:
Overall, what I see is new functions and code with zero explanations. What will happen when in N years when author will no longer be around? (nobody is immortal, and I want unv's code to be.)

At the very least, I think we need to start make documenting functions and classes a requirement.
In my opinion, documenting a class should be about describing the class' intent. Function should have documented purpose, pre-conditions and post-conditions.
Note that such a documentation can be written in the code and enforced by C++ compilers, at least partially (with things like const-ness) and that I don't think a comment like

```
/*
￼=============
￼RE_PushMatrix
￼=============
￼*/
```

is of any use. Same for paraphrasing.

Currently, total absence of documentation is a great problem, with only active contributor able to make sense of many things being slipher. It's a really bad situation bus-factor wise.

DolceTriade:
SG, I'll refactor the code with more comments and the make the push limitation to be more prominient. I'll rename this as SetTransform and ResetTransform.

DolceTriade:
Ok added comments.

src/shared/client/cg_api.cpp
DolceTriade: I added them in the renderer rather than in the trap calls to make the semantics clear.

src/engine/renderer/tr_backend.cpp
```diff
@@ -5103,6 +5103,36 @@ const RenderCommand *Poly2dIndexedCommand::ExecuteSelf( ) const
 	return this + 1;
 }
 
+bool pushed = false;
```
slipher, line 5106: static

src/engine/renderer/glsl_source/generic_fp.glsl
```diff
@@ -55,4 +56,7 @@ void	main()
 
 	color *= var_Color;
 	outputColor = color;
+#if defined(USE_DEPTH_FLATTEN)
```
slipher, line 59: Could we use a different shader for 2d rendering? I believe adding a new macro to the generic shader will cause 128 additional variants of the shader to have to be compiled. But all of those things are presumably for 3d rendering so there is no need to combine them with this macro.

DolceTriade: I'm not sure what that would look like. I'm totally a newb with the shader system :(

DolceTriade:
@necessarily-equal: pls retest. Thanks @illwieckz for your help!!

I haven't seen any major bugs due to this.

src/engine/renderer/glsl_source/generic2D_fp.glsl
```diff
@@ -0,0 +1,26 @@
...
+You should have received a copy of the GNU General Public License
+along with XreaL source code; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+===========================================================================
+*/
+
+/* generic2D_fp.glsl */
+
+#define GENERIC_2D 1
+
```
slipher, line 26: There's nothing here?

DolceTriade: Apparently it gets merged with generic_fp.glsl

slipher: You're not actually compiling this, only the "generic" one again.
```
GLShader_generic2D::GLShader_generic2D( GLShaderManager *manager ) :
	GLShader( "generic", ATTR_POSITION | ATTR_TEXCOORD | ATTR_QTANGENT, manager ),
```

illwieckz: An internal `generic2D` compiles `generic_fp.glsl` using `generic2D_fp.glsl` as concatenated header:

```c++
GLShader_generic2D::GLShader_generic2D( GLShaderManager *manager ) :
	GLShader( "generic", ATTR_POSITION | ATTR_TEXCOORD | ATTR_QTANGENT, manager ),
	[…]

void GLShader_generic2D::BuildShaderFragmentLibNames( std::string& fragmentInlines )
{
	fragmentInlines += "generic2D";
}
```

Later I want to implement a mechanism to set a define without needing a dedicated file but this should work.

src/engine/renderer/glsl_source/generic2D_fp.glsl
```diff
@@ -0,0 +1,26 @@
+/*
+===========================================================================
+Copyright (C) 2006-2011 Robert Beckebans <trebor_7@users.sourceforge.net>
```
illwieckz, line 3: I forgot to remove this when I stripped it down to a define, but a define can't have a license, the whole header can be deleted. 🤭️

src/engine/renderer/gl_shader.h
```diff
@@ -2194,6 +2194,26 @@ class u_Lights :
 	}
 };
 
+class GLShader_generic2D :
+	public GLShader,
```
slipher, line 2198: Could use a comment/TODO explaining how a lot of this stuff is copy-pasted junk which is not actually used by the shader.

src/engine/renderer/tr_backend.cpp
slipher, line 5106: `static`

src/engine/renderer/tr_backend.cpp
```diff
@@ -5103,6 +5103,36 @@ const RenderCommand *Poly2dIndexedCommand::ExecuteSelf( ) const
...
+	            glState.projectionMatrix[ glState.stackIndex ] );
+	// Load our new transformation matrix.
+	GL_LoadModelViewMatrix( this->matrix );
+	tess.surfaceShader = nullptr;
+	return this + 1;
+}
+
+const RenderCommand *ResetMatrixTransformCommand::ExecuteSelf() const {
+	Tess_End();
+	GL_PopMatrix();
```
slipher, line 5131: It should return if `pushed` is false

src/engine/renderer/tr_backend.cpp
DolceTriade: Done.

src/engine/renderer/glsl_source/generic2D_fp.glsl
DolceTriade: Done.

src/engine/renderer/gl_shader.h
DolceTriade: Done.

src/engine/renderer/tr_backend.cpp
DolceTriade: done.

src/engine/renderer/tr_backend.cpp
DolceTriade: done.

src/engine/renderer/glsl_source/generic2D_fp.glsl
illwieckz: I still see the jumbo header, maybe the change was not added when committing?

DolceTriade: oops

PR #638, opened 2022-06-27
external_deps: add option to print deps version

illwieckz:
See https://github.com/Unvanquished/release-scripts/pull/17


necessarily-equal (commented):
I don't know the setup enough to say if it's a good idea. Otherwise LGTM

PR #639, opened 2022-07-04
renderer: always look for GL_OUT_OF_MEMORY after an upload, ref #1857

illwieckz:
This is not expected to be merged as is for now, but it provides us a testing ground and a chat place dedicated to this topic.


illwieckz:
@bmorel Do you catch the `GL_OUT_OF_MEMORY` error with non-Debug build this way?

bmorel:
Nope, no luck. Only segfault.

PR #643, opened 2022-07-14
Keep error window in front of daemon window on error

necessarily-equal:
You usually want to see the error instead of a broken frozen window.

Note I couldn't test it because my version of SDL isn't compiled with it or something. In theory it shouldn't blow up, but this change needs someone to test it.

### DIFF
--- a/src/engine/client/ClientApplication.cpp
+++ b/src/engine/client/ClientApplication.cpp
@@ -34,6 +34,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #if defined(_WIN32) || defined(BUILD_GRAPHICAL_CLIENT)
 #include <SDL.h>
+#ifdef BUILD_GRAPHICAL_CLIENT
+extern SDL_Window *window;
+#else
+#define window nullptr
+#endif
 #endif
 
 namespace Application {
@@ -95,7 +100,7 @@ class ClientApplication : public Application {
         void Shutdown(bool error, Str::StringRef message) override {
             #if defined(_WIN32) || defined(BUILD_GRAPHICAL_CLIENT)
                 if (error) {
-                    SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, PRODUCT_NAME, message.c_str(), nullptr);
+                    SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, PRODUCT_NAME, message.c_str(), window);
                 }
             #endif
 


src/engine/client/ClientApplication.cpp
```diff
@@ -93,9 +96,9 @@ class ClientApplication : public Application {
         }
 
         void Shutdown(bool error, Str::StringRef message) override {
-            #if defined(_WIN32) || defined(BUILD_GRAPHICAL_CLIENT)
```
DolceTriade, line 96: Was the windows32 define due to windows spawning a terminal automatically?

necessarily-equal: I think we would use something from the platform abstractions SDL provides? I've never tried running daemonded on windows, but I would be surprised if SDL prints something in a terminal.

If you think it can I can find a somewhat different hack

slipher: On Windows daemonded/daemon-tty don't use a real console; PDCurses creates a regular window and draws the curses interface in it. The messagebox is needed because if the program just exits there is no chance to read the error message.

necessarily-equal: Ok I've added the call back.

DolceTriade:
Looks ok to me, but will let slipher check too.

src/engine/client/ClientApplication.cpp
```diff
@@ -95,7 +100,7 @@ class ClientApplication : public Application {
         void Shutdown(bool error, Str::StringRef message) override {
             #if defined(_WIN32) || defined(BUILD_GRAPHICAL_CLIENT)
```
slipher, line 101: Make this ifdef have 3 branches and put everything here instead of having a macro and stuff at the top of the file.

necessarily-equal: I don't think this is possible. For a start `extern SDL_Window *window;` must be defined only if `BUILD_GRAPHICAL_CLIENT` (the type doesn't exist otherwise), and must be defined out of the function because C++.

slipher: Oh I forgot it's inside a namespace. Maybe it would be better to have a ShowError helper function at the top... I'll stop insisting on any specific thing though. LGTM

necessarily-equal:
Did anyone check this actually works though? I haven't got a version a SDL compiled with xmessage.

DolceTriade:
I just tested this.

Without your change, the error box appears behind the fullscreen window, however with your change, the error box appears in front of the window despite it being fullscreen.

in the windowed case, the error box appears in front of the window regardless (but I think that might be window manager dependent behavior. I'm using i3 with daemon being a floating window in this windowed test.)

DolceTriade (approved):
LGTM % slipher's comments

Issue #471, opened 2021-05-21
Some dynamic light rendering issues on Intel UHD 620


illwieckz:
Reported by Dinkelkeks on Discord:

> here is a screenshot of the lighting issues i told you about. Happens when humans use flamethrowers (black round areas appear).
> 
> ```
> Intel Corporation UHD Graphics 620 (rev 07)
>         Subsystem: Lenovo Device 225d
>         Kernel driver in use: i915
>         Kernel modules: i915
>
> OS: Arch Linux, kernel 5.11.10
> ```
> 
> [![screenshot](https://cdn.discordapp.com/attachments/193638031478095872/845410711311614032/unvanquished_2021-05-21_223659_000.jpg)](https://cdn.discordapp.com/attachments/193638031478095872/845410711311614032/unvanquished_2021-05-21_223659_000.jpg)

> here another example: Black area directly above the human player. Seems to be a cone area around where the flamethrower is aimed
> [![screenshot](https://cdn.discordapp.com/attachments/193638031478095872/845412427775344640/unvanquished_2021-05-21_223756_000.jpg)](https://cdn.discordapp.com/attachments/193638031478095872/845412427775344640/unvanquished_2021-05-21_223756_000.jpg)

He says his CPU is i7-8650U.

This looks like the old bugs when using `cg_shadow` greater than 1 (advanced dynamic shadows) on the forward renderer, except it's set to 1 on his side and anyway options greater than 1 would crash the game instead (see #470).

illwieckz:
He also got this, that may be another issue:

> ok funny enough, I cant reproduce it now. Instead I get this new visual bug:
>
> [![screenshot](https://cdn.discordapp.com/attachments/193638031478095872/845416810182213672/unvanquished_2021-05-21_234352_000.jpg)](https://cdn.discordapp.com/attachments/193638031478095872/845416810182213672/unvanquished_2021-05-21_234352_000.jpg)
> 
> its definitively influenced by the amount of fire. When only using a little, I get smooth yellow light

He confirmed disabling dynamic lighting fixed this bug.

zorun:
Same issue on the previous generation HD 620 (core i3-7100U)

Disabling dynamic lighting indeed fixed the bug.

illwieckz:
> Same issue on the previous generation HD 620 (core i3-7100U)
> 
> Disabling dynamic lighting indeed fixed the bug.

[...]

PR #647, opened 2022-07-20
Factorize some code between basic, specular and physical lighting

illwieckz:
This is low priority but I just stumbled upon it while reading the code.

Note that it has a computational change as the `NdotL` compute that was only clamped for physical mapping is now clamped in all cases. But I have read in some place that `NdotL` is expected to be `0` for `0°` angle and `1` for `90°` and maybe this value should never go beyond.

I have seen no regression with both light mapping and dynamic lighting (this code is used by both), I tested lighting on basic materials, specular materials and physical materials.

Since it modifies a code that is known to be affected by a bug (see #471), it may be interesting for affected people to test this code to see if by pure luck that makes a difference, I doubt though.

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -72,6 +72,20 @@ void computeLight( vec3 lightDir, vec3 normal, vec3 viewDir, vec3 lightColor,
   float NdotH = clamp( dot( normal, H ), 0.0, 1.0 );
 #endif // USE_PHYSICAL_MAPPING || r_specularMapping
 
+  // clamp( NdotL, 0.0, 1.0 ) is done below
+  // if no r_halfLambertLighting and no r_wrapAroundLighting.
+  float NdotL = dot( normal, lightDir );
+
+#if defined(r_halfLambertLighting)
+  // http://developer.valvesoftware.com/wiki/Half_Lambert
+  NdotL = NdotL * 0.5 + 0.5;
+  NdotL *= NdotL;
+#elif defined(r_wrapAroundLighting)
+  NdotL = clamp( NdotL + r_wrapAroundLighting, 0.0, 1.0) / clamp(1.0 + r_wrapAroundLighting, 0.0, 1.0);
+#else
+  NdotL = clamp( NdotL, 0.0, 1.0 );
+#endif
+
 #if defined(USE_PHYSICAL_MAPPING)
   // Daemon PBR packing defaults to ORM like glTF 2.0 defines
   // https://www.khronos.org/blog/art-pipeline-for-gltf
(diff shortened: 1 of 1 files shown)


illwieckz:
It also enables `halfLambertLighting` and `wrapAroundLighting` for all algorithm (physical lighting did not have it).

I noticed that `halfLambertLighting` is done for every surface but we may want to do it only for models (this is out of scope of this PR).

bmorel (approved):
LGTM

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -72,6 +72,18 @@ void computeLight( vec3 lightDir, vec3 normal, vec3 viewDir, vec3 lightColor,
   float NdotH = clamp( dot( normal, H ), 0.0, 1.0 );
 #endif // USE_PHYSICAL_MAPPING || r_specularMapping
 
+  float NdotL = clamp( dot( normal, lightDir ), 0.0, 1.0 );
+
+#if defined(r_halfLambertLighting)
+  // http://developer.valvesoftware.com/wiki/Half_Lambert
+  NdotL = NdotL * 0.5 + 0.5;
+  NdotL *= NdotL;
```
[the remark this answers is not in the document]
bmorel: Oh right. My bad. And it seems GLSL only have pow (like C, actually).

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -72,6 +72,18 @@ void computeLight( vec3 lightDir, vec3 normal, vec3 viewDir, vec3 lightColor,
   float NdotH = clamp( dot( normal, H ), 0.0, 1.0 );
 #endif // USE_PHYSICAL_MAPPING || r_specularMapping
 
+  float NdotL = clamp( dot( normal, lightDir ), 0.0, 1.0 );
+
+#if defined(r_halfLambertLighting)
```
slipher, line 77: I think you will get a different result if the USE_PHYSICAL_MAPPING flag is on, and either r_halfLambertLighting or r_wrapAroundLighting is set.

illwieckz: Yes it is expected this code enables halfLamberLighting and wrapAroundLighting for PBR too as PBR is just another way to compute specularity and things like that.

When refactoring code at 0.51 time I failed to see how I can also factorize that and just let it for the future (PBR code was already not using those features yet, so there was no regression anyway).

illwieckz:
I noticed the `clamp(NdotL, 0, 1)` was already done in specular code, but later, once we know neither halfLambertLighting, nor wrapAroundLighting is done (and so those other computes do something else instead), I adjusted accordingly.

So there is no computational change at all for specular mapping, it just enables the other lighting for pbr, and factorize a bit.

PR #648, opened 2022-07-22
Curl: add a thread for http downloads

necessarily-equal:


necessarily-equal:
I'm not sure why, but this can bork the loading screen progress bar.

slipher:
There's really no need to use a thread. I found a perfect place to call CL_WWWDownload - in Com_Frame, if you set a really low framerate then the main thread wakes up every 50 ms and pumps the event loop ([here](https://github.com/DaemonEngine/Daemon/blob/30330ad0677f14c46155ad36296aaa9009020f92/src/engine/qcommon/common.cpp#L971)). So by invoking the download loop there we could avoid any problems caused by setting a low framerate when minimized.

necessarily-equal:
To get rid of the thread-safety problem this version has, I can make DL_DownloadStatus return the progress (return a struct that contains {status, count, max}) and get rid of the cvars. I think it's a bit better to have this in a thread and to get rid of the cvars.

PR #649, opened 2022-07-23
beacons: Register Beacons as a type in Com_EntityTypeName.

DolceTriade:
This fixes a crash in entityList in debugMode because Com_EntityTypeName
returns nullptr if the enum is not found and is less than ET_EVENT.


necessarily-equal (approved):
Untested, LGTM

src/engine/qcommon/q_shared.cpp
```diff
@@ -410,6 +410,7 @@ const char *Com_EntityTypeName(entityType_t entityType)
 	case entityType_t::ET_MODELDOOR:        return "MODELDOOR";
 	case entityType_t::ET_LIGHTFLARE:       return "LIGHTFLARE";
 	case entityType_t::ET_LEV2_ZAP_CHAIN:   return "LEV2_ZAP_CHAIN";
+	case entityType_t::ET_BEACON:			return "BEACON";
```
bmorel, line 413: Tabs to align?

PR #651, opened 2022-07-24
Key binding cleanup

slipher:


src/engine/sys/sdl_input.cpp
```diff
@@ -1032,9 +1032,8 @@ static void IN_ProcessEvents( bool dropInput )
 					Key kKeycode = IN_TranslateSDLToQ3Key( &e.key.keysym, true );
 					bool consoleFound = false;
 					for (Key k: {kScan, kKeycode} ) {
-						if ( IN_IsConsoleKey( k ) && !keys[ Key(K_ALT) ].down) {
+						if ( IN_IsConsoleKey( k ) ) {
```
DolceTriade, line 1035: This is useful, it allows you to type in backticks and tilde into the console for any reason.

src/engine/client/cl_keys.cpp
```diff
@@ -567,16 +573,15 @@ void CL_KeyEvent( const Keyboard::Key& key, bool down, unsigned time )
 	}
 #endif
 
-	// console key is hardcoded, so the user can never unbind it
-	if ( key == Key(K_CONSOLE) || ( keys[ Key(K_SHIFT) ].down && key == Key(K_ESCAPE) ) )
+	// console key combination is hardcoded, so the user can never unbind it
```
DolceTriade, line 576: what is cl_consoleKeys? Is it an additional way to open the console?

DolceTriade (changes requested):

src/engine/client/cl_keys.cpp
DolceTriade: Ah nvm, this is actually for the shift-esc console bind.

src/engine/sys/sdl_input.cpp
slipher: Does it really? I don't know any OS where you can enter text while Alt is held.

DolceTriade: I can with Linux

bmorel: I confirm I can type `²` in console on linux. As for backticks, french layout requires the use of the `alt gr` key anyway, so I can't really tell.

DolceTriade (approved):

src/engine/sys/sdl_input.cpp
slipher: It works for me too. I thought I had tested this on Linux and it didn't work. So I put it back.

PR #652, opened 2022-07-25
Use floats and float functions to avoid casting all the time

necessarily-equal:


bmorel:
I'm expert on the topic, but as the reason to do this is "performance", I'd like to:

1. know how you verified that there is indeed a performance problem
2. have the results of the report

I think it would we weird if casting a constant would have a runtime cost (everything is known at compilation time after all).

bmorel:
> I'm expert on the topic

I obviously meant: "I'm NOT expert"

necessarily-equal:
> know how you verified that there is indeed a performance problem

I don't really think it's worth getting precise numbers because it would take me quite some time (closing everything, rebooting into ubuntu and checking back, reopening things, relogging into websites like github), but last I looked I was surprised to see that the internal libstdc++ function that did the cast was actually quite important in terms of CPU time

> I think it would we weird if casting a constant would have a runtime cost (everything is known at compilation time after all).

While I agree that `float a = 0.1;` has no impact, `float a, b; b = 0.1 * a;` has an impact, because a is casted to double and the result is casted back to float, so you end up casting twice unwillingly.

bmorel:
> has an impact, because a is casted to double and the result is casted back to float, so you end up casting twice unwillingly.

Interestingly, this rings some bells to me, I've seen other reports about that kind of problems, saying that it was worthless to use floats, and just using doubles was better, because conversion cost.
I'm also almost certain that in the past, glm allowed to easily use various types. Maybe it still does, but it seems to be not advertised at all. If I addition one and one...

necessarily-equal:
For the record in Unvanquished gamelogic I think we may have a second reason to use floats because we actually use "32 bits" executables. I don't know however if NaCl assembly can use hardware for double operations.

If we are unsure about the performance implications (I am not), I can always claim this PR is a cleanup :P

bmorel:
If unsure about the performance means, this PR can possibly make things worst.
While I agree that the 1st commit is a good idea (less code duplication) I have doubts about the 2nd.

necessarily-equal:
Well, less casts from double to float and the other way around shouldn't be able to cause a performance regression

slipher:
> For the record in Unvanquished gamelogic I think we may have a second reason to use floats because we actually use "32 bits" executables.

No, we don't run the processor in 32-bit mode (unless of course it is an actual 32-bit Windows machine). The NaCl binaries are built on an ILP32 model (longs and pointers are 32 bits), but this just means we voluntarily don't use the upper 32 bits of the pointer. It's still using the 64-bit instruction set.

slipher:
LGTM

PR #653, opened 2022-07-25
Fix crash with r_vertexLighting

slipher:
Also show a useful error message when crashing from this class of bug (using an "unused" permutation).

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -748,7 +748,7 @@ std::string     GLShaderManager::BuildGPUShaderText( Str::StringRef mainShaderNa
 	return shaderText;
 }
 
-bool GLShaderManager::buildPermutation( GLShader *shader, int macroIndex, int deformIndex )
+void GLShaderManager::buildPermutation( GLShader *shader, int macroIndex, int deformIndex )
 {
 	std::string compileMacros;
 	int  startTime = ri.Milliseconds();
@@ -759,7 +759,7 @@ bool GLShaderManager::buildPermutation( GLShader *shader, int macroIndex, int de
 	if ( i < shader->_shaderPrograms.size() &&
 	     shader->_shaderPrograms[ i ].program )
 	{
-		return false;
+		return;
 	}
 
 	if( shader->GetCompileMacrosString( macroIndex, compileMacros ) )
@@ -782,7 +782,7 @@ bool GLShaderManager::buildPermutation( GLShader *shader, int macroIndex, int de
 
 			if ( baseShader->unusedPermutation )
 			{
-				return true;
+				return;
 			}
 
 			glAttachShader( shaderProgram->program, baseShader->VS );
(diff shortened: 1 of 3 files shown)


src/engine/renderer/gl_shader.cpp
```diff
@@ -1041,11 +1039,6 @@ void GLShaderManager::CompileGPUShaders( GLShader *shader, shaderProgram_t *prog
 			/* FIXME: add this test: ( strcmp( token, "USE_TCGEN_LIGHTMAP" ) == 0 && r_lightMapping->integer == 0 )
 			when lightmaps are never used when lightmapping is disabled
 			see https://github.com/DaemonEngine/Daemon/issues/296 is fixed */
-			if ( strcmp( token, "USE_LIGHT_MAPPING" ) == 0 && r_vertexLighting->integer != 0 )
```
illwieckz, line 1044: The comment preceding this line is now meaningless (also `is fixed` was a leftover).

Maybe rewrite the comment saying we may want to not build it when surface isn't lightmapped or lightmapping is disabled but this currently does not work (with a link to the issue/PR)?

illwieckz:
Do we know when light mapping code is used when light mapping is disabled?

Anyway I consider this a correct fix for now, as it allows us to release a non-crashing engine, but we may want to track the root issue of using the light mapping code when light mapping is disabled.

I now remind that maps using lightstyles (like metro) was previously still drawing lightmaps on lightstyled surfaces even when grid lighting was enabled: https://github.com/DaemonEngine/Daemon/issues/296

Maybe we still have other issues like that.

slipher:
> Do we know when light mapping code is used when light mapping is disabled?

It's a shader with an ST_LIGHTMAP stage. 

In `Render_lightMapping` we have 
```
	bool enableLightMapping = !r_vertexLighting->integer \
		&& tess.bspSurface \
		&& tess.lightmapNum >= 0 && tess.lightmapNum <= tr.lightmaps.currentElements;
```

but then...

```
	if ( noLightMap )
	{
		lightmap = tr.whiteImage;
		enableLightMapping = true;
	}
```

src/engine/renderer/gl_shader.cpp
slipher: I don't understand it well enough to write an intelligent comment myself.

illwieckz:
Oh I see, If I'm right those are surfaces without lightmap stage!!! We lit them with a dummy lightmap that is a pure white image. Maybe we can do the same dummy lightgrid instead.

illwieckz:
Can you add a comment like that in `tr_shade.cpp`?

```diff
+	/* FIXME: This requires light mapping GLSL code to be compiled even when light mapping is disabled.
+	See https://github.com/DaemonEngine/Daemon/issues/500 */
	if ( noLightMap )
	{
		lightmap = tr.whiteImage;
		enableLightMapping = true;
	}
```

slipher:
Added the comment around the noLightMap hack and deleted the one by the unusedPermutation stuff.

necessarily-equal:
The code looks okish to me, but then I don't know anything about that part of the code or that programming realm — so I guess you can count that as weak approval?

DolceTriade (approved):

illwieckz (approved):
LGTM.

PR #655, opened 2022-07-26
Fix curl error message

necessarily-equal:


DolceTriade:
Seems non-critical to push to 0.54

necessarily-equal:
I think I can do without, but on the other hand troubleshooting download issues is quite annoying because you HAVE TO make a nacl build. I think having proper error reporting can help a bit driving things forward if the binaries are not ready yet.

This change shouldn't risk inducing any breakage.

necessarily-equal:
see https://github.com/Unvanquished/Unvanquished/issues/1981 for an example of how this hinders debugging

slipher:
LGTM

DolceTriade (approved):

necessarily-equal:
Thanks, you can merge it

PR #657, opened 2022-07-26
sdl_glimp: also recreate window on resolution change

illwieckz:
Window resolution can also change without calling `vid_restart`: it happens when the window failed to be created and a lower resolution is tried instead.

This replaces:

- https://github.com/DaemonEngine/Daemon/pull/644

This is a fixup for:

- https://github.com/DaemonEngine/Daemon/pull/478

Not recreating the window when the fallback resolution is tried because the context creation failed was a regression.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -785,20 +785,27 @@ static bool GLimp_CreateContext( const glConfiguration &configuration )
 
 /* GLimp_DestroyWindowIfExists checks if window exists before
 destroying it so we can call GLimp_RecreateWindowWhenChange even
-if no window is created yet.
-
-It is assumed width, height and other things like that are unchanged,
-given vid_restart is called when changing those and then destroying
-the window before calling this. */
+if no window is created yet. */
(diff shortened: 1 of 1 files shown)


src/engine/sys/sdl_glimp.cpp
```diff
@@ -785,20 +785,28 @@ static bool GLimp_CreateContext( const glConfiguration &configuration )
...
-
-It is assumed width, height and other things like that are unchanged,
-given vid_restart is called when changing those and then destroying
-the window before calling this. */
+if no window is created yet. */
 static bool GLimp_RecreateWindowWhenChange( const bool fullscreen, const bool bordered, const glConfiguration &configuration )
 {
+	/* Those Values doen't contribute to anything when the window
+	is't created yet, but we need them to be set to silence warnings
+	about comparing undefined variables. */
```
slipher, line 793: static variables are always initialized, so the comment about warnings is not true.

LGTM

illwieckz: Right, I edited the comment.

DolceTriade (approved):

Issue #658, opened 2022-07-27
0.53.0/sync Daemon broke on XWayland


necessarily-equal:
since commit 60d055c19138e, daemon broke on XWayland, and thus on Wayland because the X11 backend is still the default. `SDL_BACKEND=wayland` fixes it.

```
Warn: Couldn't get window display mode: Couldn't find display mode match
Warn: GLimp: Unknown error for mode 3
Warn: Couldn't load a renderer.
```

reverting the commit makes it work anew. (to test, checkout 2abcb3fca on Unv, `git submodule update --init --recursive`, `cd daemon`, `git revert 60d055c19138eff19e16c934107194520017dc89`)

I'm not sure why is that. I think this is because unv tries to do forbidden things on xwayland, as the screen flickers a lot when I open or select the unv window on Xorg.

I think this is blocking for the 0.53.0 release

necessarily-equal:
a "last resort" workaround that we can consider is to force set SDL_VIDEODRIVER=wayland if we detect wayland is used.

It's kinda ugly, but Unvanquished works ok on wayland, modulo the issues tracked in https://github.com/DaemonEngine/Daemon/issues/482. I'm not sure it has been tested enough to be the only and default backend.

illwieckz:
The issue looks more complex, it may even not be related to Wayland at all.

When trying to render on a window larger than what the GPU hardware supports, the engine would crash with a segmentation fault. Unfortunately it looks like the only error we manage to catch before the engine executes the code that will do a segmentation fault is “_Couldn't find display mode match_”.

So _I thought it was a good idea_ (but maybe I was wrong) to catch that error to restart the engine with a safe resolution instead.

But in fact, doing that may prevent to do many valid usages.

If I understand correctly, the “_Couldn't find display mode match_” error happens when we try to open a window that is not the size of a resolution supported by the screen (not the GPU!). So it would not work if we try to open a window that is larger than the screen (as the screen would say this is not a fullscreen native resolution), but it would probably fail as well when trying to open a window that has a custom resolution that is not part of the native screen resolutions, even if smaller than the native screen resolution (like opening a window of size arbitrary `2000×1234` on a `3840×2160` screen). If some windowing system even hide some hardware limit this may even happen as well.

Catching “_Couldn't find display mode match_” to prevent segfault when the GPU hardware didn't supported the resolution was not that a bad idea because to begin with, the display server would never starts with an unsupported resolution so the GPU hardware limit actually matches the display limit.

[...]

PR #659, opened 2022-07-27
Fix Daemon on XWayland

necessarily-equal:
Fixes https://github.com/DaemonEngine/Daemon/issues/658

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -523,7 +523,7 @@ static bool GLimp_DetectAvailableModes()
 	if ( SDL_GetWindowDisplayMode( window, &windowMode ) < 0 )
 	{
 		logger.Warn("Couldn't get window display mode: %s", SDL_GetError() );
-		return false;
+		return true;
 	}
 
 	for ( i = 0; i < SDL_GetNumDisplayModes( display ); i++ )


necessarily-equal:
```patch
diff --git a/src/engine/sys/sdl_glimp.cpp b/src/engine/sys/sdl_glimp.cpp
index 1500e7cd..f14662d0 100644
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -522,8 +522,14 @@ static bool GLimp_DetectAvailableModes()
 
        if ( SDL_GetWindowDisplayMode( window, &windowMode ) < 0 )
        {
+               // we can't use SDL_GetWindowWMInfo yet, so we can't use
+               // that and use getenv instead.
+               if (Q_stricmp(getenv("XDG_SESSION_TYPE"), "wayland") == 0)
+               {
+                       return true; // don't fail as this is normal on XWayland.
+               }
                logger.Warn("Couldn't get window display mode: %s", SDL_GetError() );
-               return true;
+               return false;
        }
 
        for ( i = 0; i < SDL_GetNumDisplayModes( display ); i++ )
```

illwieckz:
If we're gonna make this test return `true`, we better revert the whole function to a `void` one.

The purpose of this function being a `bool` one and this test returning `false` was to catch this precise “_Couldn't find display mode match”_ error and to prevent a segfault (so it was 100% done on purpose to return false there and abort).

@necessarily-equal What happens if you run latest master on Wayland, including the just merged:

- https://github.com/DaemonEngine/Daemon/pull/657

Does it return to a safe resolution?

Note that this is not a fix for the issue you experience, but it would help to understand more what's going on.

illwieckz:
@necessarily-equal can you add this simple comment above the `return true;` line

```
/* FIXME: returning true means the engine will crash if the window size is larger than what the GPU can do,
but we need to not fail to open window that are not using native screen resolutions even if the GPU can handle it. */
```

I assume it's better to give everyone the ability to open a window with an arbitrary size than preventing a crash that is only likely to happens on some obsolete hardware if the user tries to use an higher resolution than his GPU can do (which is expected to be already larger than his screen, so not something he wants).

We can do better things after 0.53, 0.52 didn't catch the said crash so not catching this crash in 0.53 isn't a regression, while preventing the creation of arbitrary window sizes is a regression.

necessarily-equal:
Sounds good. I'll be afk tomorrow, you can push this to the PR and merge it if you want to proceed with the release

Le 27 juillet 2022 19:00:47 GMT+02:00, Thomas Debesse ***@***.***> a écrit :
***@***.*** can you add this simple comment above the `return true;` line

illwieckz:
merged as ba3538993e18bd6cb0966a4cb8395a7a2960a88a

PR #660, opened 2022-07-28
Fix shader compile error on Mac

slipher:
Tested on M1 Mac and Windows+Nvidia.


DolceTriade:
Surprising that this fixes it.

DolceTriade (approved):

PR #661, opened 2022-07-29
shaders: Remove some ET only shaders.

DolceTriade:
Remove sunShader and flareShader. They are only used in ET and not
actually used anywhere in the engine. They just spam us.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -6824,14 +6824,6 @@ void RE_LoadWorldMap( const char *name )
 	// only set tr.world now that we know the entire level has loaded properly
 	tr.world = &s_worldData;
 
-	//----(SA)  set the sun shader if there is one
-	if ( tr.sunShaderName )
-	{
-		tr.sunShader = R_FindShader( tr.sunShaderName, shaderType_t::SHADER_3D_STATIC, RSF_DEFAULT );
-	}
-
-	//----(SA)  end
-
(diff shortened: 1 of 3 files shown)


slipher:
LGTM

illwieckz:
For information if we want to reimplement them the missing images can be found in XreaL (old zip release, not repository).

illwieckz:
Si I looked at the code and even the `RB_DrawSun` function is missing from our tree.

I sometime have the strong feeling OpenWolf was not forked from latest ET:XreaL…

The sun shader is a special shader to display on the skybox like this (here in ET:Legacy, Oasis map):

[](https://dl.illwieckz.net/b/etlegacy/shots/sunshader/20220729-223147-etlegacy-oasis-sunshader.jpg)

Providing the needed assets and setting `r_drawSun 1` in Dæmon does nothing (`RB_DrawSun` function is missing…).

If we want to implement this one day, the data is there (from `XreaL_PreAlpha_20080704.7z`).

https://dl.illwieckz.net/b/daemon/features/xreal-missing-shaders/engine-xreal_0.dpk

For some unknown reason the data is shipped in `XreaL_PreAlpha_20080704.7z` but there is no code for it, and the data is not shipped in `ETXreaL-0.3.0-20111110` but there is code for it)…

illwieckz:
@DolceTriade you may want to also delete the `r_drawSun` cvar, that was to enable/disable `sunShader`.

PR #672, opened 2022-07-31
Remove cvar functions from refimport_t

necessarily-equal:
The null renderer doesn't even define them.

This doesn't fix a bug or anything, but will make it easier to untangle for the next one.

### DIFF
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -2748,11 +2748,6 @@ static bool CL_InitRef()
 	ri.FS_FCloseFile = FS_FCloseFile;
 	ri.FS_FOpenFileRead = FS_FOpenFileRead;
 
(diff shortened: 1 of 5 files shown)


DolceTriade:
To add some history, the reason that these are defined in refImport_t is back in the day, the renderer wasn't directly linked to the binary. It was dlOpen'd. Part of me is conflicted since yes, it is an obvious clean up when we link the renderer to the engine...

However it'l make it difficult in the unlikely situation if we ever get a new renderer.

necessarily-equal:
I would argue that the renderer shouldn't be reasonable for the cvar system

Le 31 juillet 2022 20:30:29 GMT+02:00, DolceTriade ***@***.***> a écrit :

slipher:
LGTM

PR #673, opened 2022-08-01
CI: bump Mac image

slipher:
"This is a scheduled macOS-10.15 brownout. The macOS-10.15 environment is deprecated and will be removed on September 30th, 2022."

### DIFF
--- a/src/engine/framework/VirtualMachine.cpp
+++ b/src/engine/framework/VirtualMachine.cpp
@@ -39,6 +39,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #else
 #include <unistd.h>
 #include <signal.h>
+#include <spawn.h>
 #include <fcntl.h>
 #include <sys/wait.h>
 #ifdef __linux__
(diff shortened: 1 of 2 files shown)


slipher:
Fixed the vfork deprecation warning by rewriting from vfork+exec to posix_spawn. I verified that everything still has the expected behavior when killing the engine or VM process on Mac or Linux.

src/engine/framework/VirtualMachine.cpp
```diff
@@ -176,62 +177,20 @@ static std::pair<Sys::OSHandle, IPC::Socket> InternalLoadModule(std::pair<IPC::S
...
-
-		// If the exec fails, return errno to the parent through the pipe
-
-		ssize_t wrote = write(pipefds[1], &errno, sizeof(int));
-		Q_UNUSED(wrote);
-		_exit(-1);
+	posix_spawn_file_actions_t fileActions;
+	if (0 != posix_spawn_file_actions_init(&fileActions) ||
+	    0 != posix_spawn_file_actions_addclose(&fileActions, pair.first.GetHandle()) ||
+	    0 != posix_spawn_file_actions_addclose(&fileActions, 0) ||
```
DolceTriade, line 183: Should use use the STDIN_FILENO and STDOUT_FILENO, and STDERR_FILENO

src/engine/framework/VirtualMachine.cpp
```diff
@@ -176,62 +177,20 @@ static std::pair<Sys::OSHandle, IPC::Socket> InternalLoadModule(std::pair<IPC::S
...
+	    0 != posix_spawn_file_actions_addclose(&fileActions, 1) ||
+	    (stderrRedirect &&
+	     0 != posix_spawn_file_actions_adddup2(&fileActions, fileno(stderrRedirect.GetHandle()), 2))) {
+		Sys::Error("VM: failed to construct posix_spawn_file_actions_t");
```
DolceTriade, line 187: Do these return errno? Should we report these so users know what went wrong?

src/engine/framework/VirtualMachine.cpp
```diff
@@ -176,62 +177,20 @@ static std::pair<Sys::OSHandle, IPC::Socket> InternalLoadModule(std::pair<IPC::S
...
-			break;
-	}
-	close(pipefds[0]);
-	if (count) {
-		waitpid(pid, nullptr, 0);
-		Sys::Drop("VM: Failed to exec: %s", strerror(err));
+	pid_t pid;
+	int err = posix_spawn(&pid, args[0], &fileActions, nullptr, const_cast<char* const*>(args), nullptr);
+	if (err != 0) {
+		Sys::Drop("VM: Failed to fork process: %s", strerror(err));
```
DolceTriade, line 193: spawn?

DolceTriade (commented):
Nice, I didn't know this was going to be deprecated. Nice fix.

src/engine/framework/VirtualMachine.cpp
slipher: Done.

src/engine/framework/VirtualMachine.cpp
slipher: These are just building up a data structure and not doing anything with the kernel. So it couldn't fail unless it fails to allocate a new linked list element or something. I don't want to write nice error handling for something that will never happen.

DolceTriade (approved):
lgtm

PR #674, opened 2022-08-02
fix for network port > 2 power 15

DaVinciCode2:
The engine can't use network port > 2 power 15 because of the signedness of the endianness functions and some bad cast in the code.

This pull request fix these problems by:
- adding unsigned endianness functions UBigShort/USwap16 and others for coherency (the same code with unsigned type in the function's signatures).
- editing code who takes and converts network port number in the engine to use these new functions and rectify some broken casts and types.

I built and tested the game with this fix and it works.

### DIFF
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -1070,7 +1070,7 @@ void CL_Connect_f()
 
 	if ( clc.serverAddress.port == 0 )
 	{
-		clc.serverAddress.port = BigShort( PORT_SERVER );
+		clc.serverAddress.port = UBigShort( PORT_SERVER );
 	}
 
 	serverString = NET_AdrToStringwPort( clc.serverAddress );
@@ -1238,7 +1238,7 @@ static netadr_t CL_RconDestinationAddress()
 
 	if ( to.port == 0 )
 	{
-		to.port = BigShort( PORT_SERVER );
+		to.port = UBigShort( PORT_SERVER );
 	}
 
 	return to;
@@ -1897,7 +1897,7 @@ CL_ServerLinksResponsePacket
 */
 void CL_ServerLinksResponsePacket( msg_t *msg )
 {
-	int      port;
+	unsigned short      port;
 	byte      *buffptr;
 	byte      *buffend;
 
@@ -1923,13 +1923,13 @@ void CL_ServerLinksResponsePacket( msg_t *msg )
 		// IPv4 address
 		memcpy( cls.serverLinks[ cls.numserverLinks ].ip, buffptr, 4 );
 		port = buffptr[ 4 ] << 8 | buffptr[ 5 ];
-		cls.serverLinks[ cls.numserverLinks ].port4 = BigShort( port );
+		cls.serverLinks[ cls.numserverLinks ].port4 = UBigShort( port );
 		buffptr += 6;
 
 		// IPv4 address
 		memcpy( cls.serverLinks[ cls.numserverLinks ].ip6, buffptr, 16 );
 		port = buffptr[ 16 ] << 8 | buffptr[ 17 ];
-		cls.serverLinks[ cls.numserverLinks ].port6 = BigShort( port );
+		cls.serverLinks[ cls.numserverLinks ].port6 = UBigShort( port );
 		buffptr += 18;
 
 		++cls.numserverLinks;
@@ -2013,7 +2013,7 @@ void CL_ServersResponsePacket( const netadr_t *from, msg_t *msg, bool extended )
 		bool duplicate = false;
 		byte ip6[16];
 		byte ip[4];
-		short port;
+		unsigned short port;
 
 		// IPv4 address
 		if ( *buffptr == '\\' )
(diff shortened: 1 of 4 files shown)


src/engine/client/cl_main.cpp
```diff
@@ -1897,7 +1897,7 @@ CL_ServerLinksResponsePacket
 */
 void CL_ServerLinksResponsePacket( msg_t *msg )
 {
-	int      port;
+	unsigned short      port;
```
DolceTriade, line 1900: Can we use uint16_t here and elsewhere ?

DaVinciCode2: in documentation https://en.cppreference.com/w/cpp/language/types and https://en.cppreference.com/w/c/types/integer . it's always equal on all platforms so yes, you can replace all unsigned short of the codebase with uint16_t. it's a small fix, so i try to change only the needed code.

DolceTriade: No, I don't mean the codebase. Just where you touch it for this change. I find uint16_t more readable and explicit.

src/engine/client/cl_main.cpp
```diff
@@ -3467,7 +3467,7 @@ void CL_LocalServers_f()
 		// can nicely run multiple servers
 		for ( j = 0; j < NUM_SERVER_PORTS; j++ )
 		{
-			to.port = BigShort( ( short )( PORT_SERVER + j ) );
+			to.port = UBigShort( ( short )( PORT_SERVER + j ) );
```
necessarily-equal, line 3470: I think you would want unsigned/uint16_t here too

DaVinciCode2: all changes made

src/engine/client/cl_main.cpp
DaVinciCode2: all changes made

DolceTriade (approved):

necessarily-equal:
Sounds good. I think we'll have to check the other uses of these functions elsewhere another day

DaVinciCode2:
yes, because i can't understand why a function to convert big/little endian to little/big endian must use signed types.

necessarily-equal:
@DolceTriade I think it's good to merge after a test. Can't do that those weeks so if you could do it when you have time I'd be great

DaVinciCode2:
here, my server.cfg to help @DolceTriade https://pastebin.com/Kq9UBY82
note : ipv6 and port change documentation aren't easy to find (this was a lot of work for testing too)

illwieckz:
I assume the patch doesn't break compatibility with the release (except it allows to host servers on ports current release can't reach)?

bmorel:
Anticipating data loss because too old, the aforementioned configuration:

```
// Example Unvanquished dedicated server config
// command to start the server ./daemonded +exec server.cfg with server.cfg in /home/user/.local/share/unvanquished/config
 
// network config
//set net_ip "127.0.0.1"
//set net_ip6 "ipv6_addr"
set net_port "65000"
//set net_port6 "45000"
//set net_enabled 2
//set dedicated 2
net_restart
 
// advertise the server on public server list
// 0 - advertise everything, the server will be added to the public server list and information about it will be displayed to players
// 1 - don't advertise but reply to status queries, this allows players to join and see what map is being played, how many people are playing, etc.
// 2 - don't reply to status queries but accept connections, players can connect but not see server information
// 3 - only accept LAN connections
set server.private 0
 
// server can be joined without password
// set 1 to require password (see g_password below)
set g_needpass 0
 
// password to join server
//set g_password "(server password here)"
 
// rcon password to remotely manage server
//set rcon.server.password "password"
 
// loading screen
set sv_hostname "temporary serv"
set g_motd "^2welcome ^5to play unvanquished"
 
// disallow downloads from the server
set sv_allowdownload 1
 
// direct and http download can be setup this way
set sv_dl_maxRate "1000"
//set sv_wwwBaseURL "example.com/unvanquished/pkg"
set sv_wwwDlDisconnected "0"
set sv_wwwDownload "1"
//set sv_wwwFallbackURL "dl.unvanquished.net/pkg"
 
// a stat web page can be advertized this way
// the link will be displayed in compatible server browsers
//set sv_statsURL "http://example.com/unvanquished/stats"
 
// maximum number of clients
// greater than 24 is not recommended from a gameplay point of view
set sv_maxclients 24
 
// set 25 slot with 1 private, provides 24 public slots
// while admin can still connect and fill the extra slot
// while server is full
//set sv_maxclients 25
//set sv_privateclients 1
//set sv_privatePassword "(admin server password)"
 
// no time limit
set timelimit 0
 
// end the game as a draw after 45 minutes
//set timelimit 45
 
// load next map if no people is playing on server at this time
// it ensures money, evolution and build points are regularly reset
// on fresh map so real player are not welcomed on a stalled bot match
set g_emptyTeamsSkipMapTime 15
 
// force each team to have a similar number of players
set g_teamForceBalance 0
 
// enable warmup for 20 second, the game will do a countdown before starting
//set g_doWarmup 1
//set g_warmup 20
 
// load per map config file from this subdirectory
// it will load map/<mapname>.cfg like map/plat23.cfg
// or fallback on map/default.cfg
// map config files can be used to fill game with bots
set g_mapConfigs "map"
 
// following the first map, start this rotation
// rotation is described in game/maprotation.cfg
set g_initialMapRotation rotation1
 
// start this map first, usually the last of the rotation
// since rotation will start from the first one after that
map antares
// uncomment to have rotation starting immediately, so that rotation's
// specific commands are run
// nextmap
 
//restart
//net_restart
```

bmorel:
I'm going to merge this, because there's no reason not to apparently, and it seems to be a pretty important bugfix!

necessarily-equal:
I kinda wanted to wait until after my vacations (just arrived) to test. I'd still like to.

PR #679, opened 2022-08-11
Rename some symbols consistently with other parts of the code + add some missing switch case

### LINKED ISSUE #677: More surfaces than needed look to be affected by light styles
Light style is a graphical effect where lightmaps are baked at map build time to be rendered with some mathematical functions (exampe: sine), this is useful to produce intermittent or varying lights while only relying on lightmaps.

With maps making use of this like tremor, pulse or gloom2, it appears more surfaces than needed are blinking.

I first considered it may be either a bug in q3map2, either a bug in the engine, but an original build that is rendered properly in tremfusion is misbehaving on dæmon.

illwieckz:
Some things I noticed while reading the code while trying to fix #677 (those patches don't fix it).

### renderer: name SPECULAR and PHYSICAL consistently with the rest of the code

This patch does not change anything to the logic.

- rename PHONG and PBR as SPECULAR and PHYSICAL to be consistent
  with the rest of the code. The PHONG and PBR names were just
  names I used in my very first commits when rewriting the
  material collapse code refactoring the GLSL code, before I
  had enough knowledge to decide to name the variant specular
  and physical. Now that they are named specular and physical
  everywhere else, it's better to unify back those early
  occurrences.

- I also reordered tests for specular before physical because:
  * specular code is the default if there is no specular or
    physical map (a black specular map is used) or if both
    specular mapping and physical mapping are disabled.
    Physical mapping can also be used with a black physical
    map when there is no specular or physical map but the
    specular code is faster that's why it's the default.
  * there is no game running on Dæmon using PBR textures yet.

Some reordering are just bikeshedding to just make the things
consistent. Specular code is the default anyway.

### renderer: add some missing cases for PHYSICAL in some switches

This one may fix some bugs in PBR code. PHYSICAL and SPECULAR variants are just expected to be usable anywhere where the other is usable, they just compute light effects differently with a different texture. It would even be possible to only use a single type for both and add an extra boolean for the variant but using two types just makes the code simpler everywhere.

### renderer: simplify a switch case

Two switch cases were just doing the same things.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1037,8 +1037,8 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	  TB_NORMALMAP,
 	  TB_HEIGHTMAP,
 	  TB_MATERIALMAP,
-	  TB_PHYSICALMAP = TB_MATERIALMAP,
 	  TB_SPECULARMAP = TB_MATERIALMAP,
+	  TB_PHYSICALMAP = TB_MATERIALMAP,
 	  TB_GLOWMAP,
 	  MAX_TEXTURE_BUNDLES
 	};
@@ -1061,8 +1061,8 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	  ST_DIFFUSEMAP,
 	  ST_NORMALMAP,
 	  ST_HEIGHTMAP,
-	  ST_PHYSICALMAP,
 	  ST_SPECULARMAP,
+	  ST_PHYSICALMAP,
 	  ST_REFLECTIONMAP, // cubeMap based reflection
 	  ST_REFRACTIONMAP,
 	  ST_DISPERSIONMAP,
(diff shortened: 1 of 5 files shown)


src/engine/renderer/tr_local.h
```diff
@@ -1072,8 +1072,8 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	  ST_HEATHAZEMAP, // heatHaze post process effect
 	  ST_LIQUIDMAP,
 	  ST_LIGHTMAP,
-	  ST_COLLAPSE_lighting_PBR,   // map|diffusemap + opt:normalmap + opt:glowmap + opt:physicalmap
-	  ST_COLLAPSE_lighting_PHONG, // map|diffusemap + opt:normalmap + opt:glowmap + specularmap
+	  ST_COLLAPSE_lighting_SPECULAR, // map|diffusemap + opt:normalmap + opt:glowmap + opt:specularmap
```
illwieckz, line 1075: Note: the comment is OK, that's the specular map that is optional.

src/engine/renderer/tr_shader.cpp
```diff
@@ -2062,14 +2062,14 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
...
 				// use PBR instead
 			}
-			else if ( stage->collapseType == collapseType_t::COLLAPSE_lighting_PHONG )
+			else if ( stage->collapseType == collapseType_t::COLLAPSE_lighting_SPECULAR )
 			{
 				Log::Warn("Supposedly you shouldn't have a physicalMap after a specularMap (in shader '%s')?", shader.name);
 				// use PBR instead
 			}
 
 			// Daemon PBR packing defaults to ORM like glTF 2.0
```
necessarily-equal, line 2071: Comment still says PBR

illwieckz: I just rewrote those comments accordingly, also reworded some.

src/engine/renderer/tr_local.h
```diff
@@ -1085,8 +1085,8 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 	{
 	  COLLAPSE_none,
 	  COLLAPSE_generic, // used before we know it's another one
-	  COLLAPSE_lighting_PHONG,
-	  COLLAPSE_lighting_PBR,
+	  COLLAPSE_lighting_SPECULAR,
+	  COLLAPSE_lighting_PHYSICAL,
```
slipher, line 1089: I don't see the sense of renaming PBR to PHYSICAL. The name PHYSICAL seems pretty meaningless while PBR might mean something to a person who is familiar with rendering techniques.

illwieckz:
The fixes are there:

#683

I'll redo other PRs for other cosmetic changes.

PR #680, opened 2022-08-13
Fix some warnings found by valgrind

DolceTriade:


src/engine/qcommon/net_chan.cpp
```diff
@@ -117,7 +117,7 @@ Send one fragment of the current message
 void Netchan_TransmitNextFragment( netchan_t *chan )
 {
 	msg_t send;
-	byte  send_buf[ MAX_PACKETLEN ];
+	byte  send_buf[ MAX_PACKETLEN ] = {0};
```
slipher, line 120: Why is this needed? The code is designed to memset the buffer - see Huff_putBit.

src/common/Serialize.h
```diff
@@ -74,6 +74,10 @@ namespace Util {
 	public:
 		void WriteData(const void* p, size_t len)
 		{
+			if (data.capacity() < data.size() + len)
```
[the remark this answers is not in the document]
DolceTriade: fwiw, this warning prevents compilation using -Werror, so we either need to disable this warning or fix the code.

How does it cause O(N^2) for repeated small writes? If you push_back when you go above the capacity(), doesn't that cause a reallocation anyways?

src/engine/qcommon/net_chan.cpp
DolceTriade: Valgrind still complains :shrug:

DolceTriade:
These aren't critical fixes. Let's just close this for now.

PR #684, opened 2022-08-16
tr_shader: do not multiply standalone lightmap by rgbGen unless it's const, fix #677

### LINKED ISSUE #677: More surfaces than needed look to be affected by light styles

illwieckz:
Do not multiply standalone lightmap by `rgbGen` or `alphaGen` unless they're set as `const`.

In `lightMapping_fp.glsl` we do:

```glsl
diffuse *= var_Color;
```

This is needed to be able to blend two multitextured collapsed stage like when doing terrain blending (blending rock and sand).

But we should not do it with standalone lightmap stages, otherwise it would break q3map2 light styles, likely because in such situation we do not blend a lightmap on a diffusemap but we blend a white texture passed as a lightmap on the lightmap passed as a diffusemap and then we would not multiply a diffuse but a lightmap.

To avoid the lightmap-as-a-diffuse being multiplied by some values intended for diffuse maps, we can simply set the color to white as `diffuse * 1.0  == diffuse`.

But we should still pass the color and do the multiplication if the `rgbGen` is `const` as in such situation the mapper explicitly requested the lightmap stage to be multiplied and the lightmap stage is expected to be multiplied by the values the mapper have set when creating the materials.

This fixes this regression:

- #677

The regression was introduced by 3e71c430895b733ca9f5aafa90ab00402827c1f2. The purpose of that commit was to enable terrain blending by allowing multitextured collapsed stages to be blended. Before that terrain stages were `generic` ones and rendered by `generic` GLSL shader, and there was no mechanism to blend `lightMap` stages (which handles the multitextured collapsed stages) in the `lightMap` GLSL code.

By adding this ability to blend those, it enabled multitextured terrain blending . But standalone lightmaps are also rendered as `lightMap` stages and by the `lightMap` GLSL code, and this multiplication should not be done so enabling the blending required for multitextured terrain blending broke q3map2 lightstyles that make heavy usage of standalone lightmaps, often with more than one lightmap on a single surface.

Unfortunately disabling the blending for every standalone lightmap stages would also break other usages, when mappers intentionally chose to multiply a lightmap of a surface, for example to make it less visible while keeping it. This is a more advanced alternative to `noLightmap` and can be useful on transparent glass surfaces. So to take care of all those three use cases we should reset the blend factor of the standalone lightmap stage unless it was explicitly set by the mapper.

Examples of maps and scenes reproducing the three use cases:

- q3map2 light style  
  `/devmap gloom2 ; delay 100f setviewpos 3925 95 70 55 0`
- multiplied lightmap on glass surface  
  `/devmap pulse ; delay 100f setviewpos 3725 -1380 -670 -50 0`
- multitextured terrain blending  
  `/devmap thunder; delay 100f setviewpos 3925 95 70 55 0`
  
This PR was tested against those three uses cases and produce expected results in those three uses cases.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -1021,7 +1021,53 @@ static void Render_lightMapping( int stage )
 	gl_lightMappingShader->SetUniform_ColorModulate( colorGen, alphaGen );
 
 	// u_Color
-	gl_lightMappingShader->SetUniform_Color( tess.svars.color );
+	Color::Color color = tess.svars.color;
+
+	/* Do not multiply standalone lightmap by rgbGen or alphaGen
+	unless they're set as const.
+
+	In lightMapping_fp.glsl we do:
+
+	  diffuse *= var_Color;
+
+	This is needed to be able to blend two multitextured collapsed
+	stage like when doing terrain blending (blending rock and sand).
+
+	But we should not do it with standalone lightmap stages,
+	otherwise it would break q3map2 light styles, likely
+	because in such situation we do not blend a lightmap
+	on a diffusemap but we blend a white texture passed as
+	a lightmap on the lightmap passed as a diffusemap and
+	then we would not multiply a diffuse but a lightmap.
+
+	To avoid the lightmap-as-a-diffuse being multiplied
+	by some values intended for diffuse maps, we can simply
+	set the color to white as diffuse * 1.0  == diffuse.
+
+	But we should still pass the color and do the multiplication
+	if the rgbGen is const as in such situation the mapper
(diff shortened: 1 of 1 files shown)


slipher (commented):
LGTM I guess. I don't understand it at all though.

PR #685, opened 2022-08-16
tr_shade: correct line ending

illwieckz:
It looks like the day I wrote those lines I have done too much bash.

Also indent some other lines.


slipher:
LGTM

Issue #688, opened 2022-08-18
External lightmaps from lightstyle can't coexist with externalized internal lightmaps because of broken fatLightmap


illwieckz:
We should be able to export every internal lightmap from a bsp named `castle.bsp` and to store them in a folder named `castle/` as files named like `castle/lm_0000.tga`.

But when there is both internal and external lightmaps because of internal lightmaps being used for static lightmaps and external lightmaps being used for dynamic lightmaps from q3map2 lightstyle, we whould be able to rename the existing `castle` folder to something else, fix the related path into `castle.bsp` texture lump and in related `q3map2_castle.shader` file and then export all internal lightmap files as `castle/lm_0000.tga` and others.

But this does not work for various reasons related to the fatLightmap code being broken.

This is not a regression, it never worked.

illwieckz:
What we get (see the floor):

[![broken fatlightmap](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_183641_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_183641_000.jpg)

What we expect:

[![broken fatlightmap](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_183716_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_183716_000.jpg)

illwieckz:
On a side note, when compiling a map purposedly with small resolution but as large `1024×1024` external lightmap we just get something like a fatLightmap except it doesn't use the fatLightmap code (hence it is not buggy) and it can be compressed with lossless WebP or other compression tailored for images.

Issue #687, opened 2022-08-18
Images from materials using `tcGen lightmap` are assumed to be square of 128px side in a fatLightmap


illwieckz:
Because of [this code](https://github.com/DaemonEngine/Daemon/blob/ba9ad6d71db2c8f95c017b702eda2c1b1e776c4f/src/engine/renderer/tr_shade.cpp#L2712-L2718):

```c++
		if ( pStage->tcGen_Lightmap && i == TB_COLORMAP )
		{
			MatrixMultiplyScale( matrix,
			                     tr.fatLightmapStep,
			                     tr.fatLightmapStep,
			                     tr.fatLightmapStep );
		}
```

Every images from materials using `tcGen lightmap` are assumed to internal lightmaps and then be square of 128 px side, with related matrices rescaled to be found in a fat lightmap.

A fat lightmap is a large image where many tiny 128px internal lightmaps are concatenated.

The fat lightmap mechanism is broken, unmaintained, and is a cause of many other problems.

Here this bug breaks light styles using external lightmaps with sizes different than 128px, Dæmon and Unvanquished recommendations for lightmaps being 1024px external lightmaps.

This is not a regression, it never worked.

illwieckz:
What we get:

[![broken fatlightmap](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_182903_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_182903_000.jpg)

What is expected:

[![broken fatlightmap](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_182940_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_182940_000.jpg)

Issue #327, opened 2020-04-23
Spilling out shifted internal lightmap when part of a fatLightmap


illwieckz:
Initially reported there: https://github.com/Unvanquished/Unvanquished/issues/782
But the thread reported two issues, this one if for the spilling out lightmap issue.

internal:

![lightmap](https://dl.illwieckz.net/b/unvanquished/bugs/lightmaps/unvanquished_2015-07-30_005640_000.jpg)

external:

![lightmap](https://dl.illwieckz.net/b/unvanquished/bugs/lightmaps/unvanquished_2015-07-30_010131_000.jpg)

The depicted map is the Terminus map, repackaged from Urban Terror metro map.

Released maps usually don't reproduce the bug since we usually use external lightmaps as they compress better. This issue may still bite us in the back later, and we can't assume any Dæmon-based game will repackage every legacy map available on this world, and maybe some may not have the right to.

illwieckz:
This is another bug related to fatLightmap.

By dropping fatLightmap code and loading and using internal lightmap the exact same way we load external lightmaps the bug is expected to disappear in the process.

illwieckz:
What we get (see the floor):

[![broken fatlightmap](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_165432_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_165432_000.jpg)

What we expect:

[![broken fatlightmap](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_182509_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_182509_000.jpg)

illwieckz:
The lightmap spills out because it is shifted:

What we get (see the shadows around the billboard, repackaged Terminus map from Urban Terror):

[![broken fatlightmap](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_211636_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/broken-fatlightmap/unvanquished_2022-08-18_211636_000.jpg)

What we expect:

[...]

PR #690, opened 2022-08-18
tr_bsp: drop fatLightmap support, fix #689, #327, #687, #688

### LINKED ISSUE #689: We can drop fatLightmap, it's riddled with bug and the maintenance/fixing cost is too high
I first reported issues related to internal lightmaps in 2015:

- https://github.com/Unvanquished/Unvanquished/issues/782

Then with time I seen many other bugs and I started to believe they were linked. Especially, some bugs magically disappeared when moved from internal lump of bsp to external files. Light spill out disappeared when externalizing internal lightmaps, but black externalized lightmaps appeared when internal lightmaps were externalized while there was already external lightmaps because of lightstyles.

Then I noticed the engine doesn't load the internal lightmaps and external lightmaps  the same way, so I started to wonder if unifying the way they are loading would not magically fixes all the related issues.

The last issue I've noticed is that some light style code only works when I don't rebuild the map. It turned out the code works if the produced lightmaps are smalls like in 1999.

So, what is fatLightmap?

Internal lightmaps are lightmaps stored within the BSP size. It's an array of uncompressed images (RGB streams) with fixed 128×128 size (changing the internal size would require to fork the engine as it would be another product incompatible with existing assets).

The engine loads and upload external lightmaps to GPU as separated files. But the engine loads internal lightmaps, concatenate them into a large image named fatLightmap and uploads this fatLightmap to the GPU.

It's possible to see some interest in it, you only upload the fatLightmap once.

But, it has many drawbacks.

First, right now the fatLightmap is 2048×2048 large unless there is few lightmaps and then it's 1024×1024.  Some very old hardware will not like 2048 textures (I faced such problems while testing old hardware) so I don't really know if this would display garbage or if the fatLightmap will be scaled down but on second option the internal lightmaps are already using so very low resolution that dividing by 2 would just make things ugly and would display random colors everywhere. In fact while writing it I remember having seen random colors on an old GL 2.1 laptop years ago, maybe that was that? But that is a problem specific to very old hardware, not a big concern.

But the major problem is that this code is unmaintained, no one modified it, it's like it was at Unvanquished alpha 1 release time. It's riddled with bugs, and parts of the code assume every lightmap is a lightmap in a fatLightmap, and then an internal 128×128 lightmap.

For example in https://github.com/DaemonEngine/Daemon/issues/687 I reported [this code](https://github.com/DaemonEngine/Daemon/blob/ba9ad6d71db2c8f95c017b702eda2c1b1e776c4f/src/engine/renderer/tr_shade.cpp#L2712-L2718):

```c++
		if ( pStage->tcGen_Lightmap && i == TB_COLORMAP )
		{
			MatrixMultiplyScale( matrix,
			                     tr.fatLightmapStep,
			                     tr.fatLightmapStep,
			                     tr.fatLightmapStep );
		}
```

This code rescales every matrices related to textures applied with `tcGen lightmap` assuming they're 128×128px lightmaps stored in the fatLightmap, even if they're external 1024×1024px lightmaps.

Some features like q3map2 lightstyles use external lightmaps to produce precomputed dynamic lighting. A single surface can have 2, 3, 4 lightmaps and they would be blended over time with various functions like sine and others. It allows to have wawing lights etc. This is perfect for simulating a broken neon that is blinking.

If we keep this code, lightstyles will only work if their lightmap sizse are 128×128 and they are stored in a fatLightmap.

If we delete this code, lightstyles will only work if their lightmaps are stored externally, hence not stored in a fatLightmap.

In this precise piece of code, where it is executed, it seems very hard to guess if the related file is an external file or an internal lightmap stored in a fatLightmap. In fact it would even be possible that the texture would not be a lightmap (probably very rare, though).

The cost of maintaining fatLightmap is very heavy for multiple reasons:

- no one ever touched that, it means no one has time for it, and itt's just a rotten piece of code.
- the algorithm is complex and just adds extra complexity in many places, since instead of assuming a texture that is the size of the image with offset 0x0, the engine has to compute the scaling and the offset when rendering the image. The compute is probably cheap but this is just extra complexity that can break and no one care about fixing.
- internal lightmaps are not recommended. First, we recommended 1024×1024px lightmaps for quality purposes, and those can only be externals. Second, we even recommend externalizing lightmaps (even legacy 128×128px ones) because it allows to use dedicated aggressive image compression and then it produces smaller dpk and faster downloads. There is no reason to ship maps with internal lightmaps at all, even when porting old maps, a soon as we have tools to automate the externalization.

At the same time, all the maps I ported from BSP had externalized lightmaps. The Esquirel tool was written for that. This means people are playing maps with externalized 128×128 lightmaps since 2015 without any problem. In fact I externalized them not only to provide smaller download but to workaround the light spill bug.

I see no benefit at keeping the fatLightmap mechanism, it does not look to be worth the cost, and I don't want to spend a single time on trying to fix fatLightmap bugs.

I detailed there the issues that would be magically fixed by dropping the fatLightmap code:

- [ ] https://github.com/DaemonEngine/Daemon/issues/327
- [ ] https://github.com/DaemonEngine/Daemon/issues/687
- [ ] https://github.com/DaemonEngine/Daemon/issues/688

None of those bugs are regressions. It never worked properly.

_Edit:_

Also I'm not sure the time saved by uploading the fatLightmap is worth the time spent as concatenating the lightmaps. The only remaining benefit may be that there are less images to bind but in case of light styles with more than one lightmap per material, the engine will still have to bind all of them separately.

I would not be surprised if in most cases, the fatLightmap will use more GPU memory than needed.

No one official maps make use of this feature as internal Lightmaps are not suitable as the resolution is too low, they can't be compressed, and internal deluxe maps aren't supported by the engine (q3map2 may even don't allow to produce them).

illwieckz:
It is broken, is a nest of bugs, and the maintenance cost is too high for
the very minor unverified benefits purposed for unrecommended things.

Fixes:

- https://github.com/DaemonEngine/Daemon/issues/689
  * https://github.com/DaemonEngine/Daemon/issues/327
  * https://github.com/DaemonEngine/Daemon/issues/687
  * https://github.com/DaemonEngine/Daemon/issues/688

Burn fatLightmap with fire, fix #689, #327, #687, #688.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -451,7 +451,6 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 		return;
 	}
 
-	tr.fatLightmapSize = 0;
 	int len = l->filelen;
 	if ( !len )
 	{
@@ -555,18 +554,6 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 			}
 		}
 	} else {
-		byte *buf, *buf_p;
-
-		//int       BIGSIZE=2048;
-		//int       BIGNUM=16;
-
-		byte *fatbuffer;
-		int  xoff, yoff, x, y;
-		//float           scale = 0.9f;
-
-		tr.fatLightmapSize = 2048;
-		tr.fatLightmapStep = 16;
-
 		len = l->filelen;
 
 		if ( !len )
@@ -574,7 +561,7 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 			return;
 		}
 
-		buf = fileBase + l->fileofs;
+		byte *buf = fileBase + l->fileofs;
 
 		// we are about to upload textures
 		R_SyncRenderThread();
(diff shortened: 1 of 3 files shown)


src/engine/renderer/tr_bsp.cpp
```diff
@@ -557,15 +556,7 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 	} else {
 		byte *buf, *buf_p;
```
illwieckz, line 557: I may move them where they are actually used.

src/engine/renderer/tr_local.h
```diff
@@ -2695,9 +2695,7 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
 		growList_t lightmaps;
 		growList_t deluxemaps;
 
-		image_t    *fatLightmap;
-		int        fatLightmapSize;
-		int        fatLightmapStep;
+		image_t *internalLightmaps[ MAX_LIGHTMAPS ];
```
illwieckz, line 2698: It may be possible to use a dynamically allocated list and even drop the max but for now this is not worst than a large allocated image where smaller ones are concatenated. It's already smaller in memory and it is good enough.

illwieckz: I can add this as a comment, though.

illwieckz: Well, since it's only a list of pointers… I may even drop the list entirely, and then all globals for this.

necessarily-equal:
I don't know anything about that part of the code, but sounds good

slipher:
LGTM

DolceTriade:
LGTM

DolceTriade (approved):

illwieckz:
For having a cleaner Git history I prefer to merge https://github.com/DaemonEngine/Daemon/pull/693 first. Can you look at it? Thanks in advance.

Issue #613, opened 2022-03-22
Scripting `+command` requires the game to not have focus


illwieckz:
This only works if game doesn't have focus (on this example, map has to be loaded and human player spawned):

```
/set att "+attack; delay 1f -attack"
/delay 3000 vstr att
```

One can Alt-Tab after typing those commands.

This also doesn't work if in-game console is opened but that is somewhat consistent with tremulous and quake3. On tremulous and quake3, +command is executed while console is closed, even if game has focus. So we have one shared issue plus a regression.

illwieckz:
I assume it doesn't work when game is actually reading input. But I see no cvar to disable input reading.

illwieckz:
I noticed that if scripting `+attack` works when the game is unfocused, it doesn't work when the game is minimized. Also it doesn't work with `daemon-tty`.

slipher:
This is actually a gamelogic issue. From `bg_pmove.cpp`:

```cpp
	// if talk button is down, disallow all other input
	// this is to prevent any possible intercept proxy from
	// adding fake talk balloons
	if ( usercmdButtonPressed( pmove->cmd.buttons, BTN_TALK ) )
	{
		usercmdClearButtons( pmove->cmd.buttons );
		usercmdPressButton( pmove->cmd.buttons, BTN_TALK );
		pmove->cmd.forwardmove = 0;
		pmove->cmd.rightmove = 0;

		if ( pmove->cmd.upmove > 0 )
		{
			pmove->cmd.upmove = 0;
		}
	}
```

`BUTTON_TALK` is set by the engine when the key catcher state does not allow binds to be used.

PR #692, opened 2022-08-22
cl_input: Fix typing +command in console for repeat key press.

DolceTriade:
If someone wants to type a +command like +speed or +attack in console it
doesn't work (even if they add a random arg with it) because this check
will always pass and it returns. However, since +commands in console
never have a key asscoiated with them, check whether the key is valid
first.


illwieckz:
This fixes this issue:

- https://github.com/DaemonEngine/Daemon/issues/613

and make Dæmon behaves like Tremulous and Quake 3 (as documented in that issue).

Fixing #613 is useful to script the production of a demo file that actually builds buildables like a real players, see:

- https://github.com/Unvanquished/Unvanquished/pull/1791

illwieckz (approved):
LGTM.

slipher:
LGTM

PR #693, opened 2022-08-22
tr_shader: update comment about standalone lightmap blending

illwieckz:
The statement saying standalone lightmaps are passed as
diffuse with white lightmap was wrong, standalone lightmaps
are already passed as lightmaps on white diffuse.

This updates the comment and tells the code may disable some
standalone lightmaps in some corner cases and also says this
is better wanted as it would never prevent default lightmaps
to be applied (and then it would looks good) while this code
is known to fix many other glitches.

The comment now contains an explicit FIXME to recommend
further investigations and states that the tests currently
take care of only allowing operations that are actually
known to not glitch.

This rewrites ba9ad6d71db2c8f95c017b702eda2c1b1e776c4f better.


DolceTriade:
LGTM

PR #696, opened 2022-08-23
cg_api: Remove RmlUi defines from the engine

DolceTriade:
These are legacy defines from the days of having libRocket in the
engine. Now that RmlUI is built into the cgame, they don't need to be
here and should be in the cgame.


illwieckz:
Fix https://github.com/Unvanquished/Unvanquished/issues/2052

@DolceTriade there is also this leftover (not rocket, but can be pruned at the same time):

> Another one like that is entityType_t in q_shared.h

DolceTriade:
I'm not sure if we can remove entityType_t from the engine. entityState_t depends on entityType_t. I don't want to block merging this on cleaning up the entityType stuff though.

illwieckz:
Sure, if it does not looks obvious on first sight, we can postpone. This will be good enough.

illwieckz (approved):
LGTM.

PR #698, opened 2022-08-28
convert sv_www.* cvars to Cvar

bmorel:


src/engine/server/sv_client.cpp
```diff
@@ -658,20 +662,20 @@ when the cvar is set to something, the download server will effectively never us
...
 static bool SV_CheckFallbackURL( client_t *cl, const char* pakName, int downloadSize, msg_t *msg )
 {
-	if ( !sv_wwwFallbackURL->string || !sv_wwwFallbackURL->string[0] )
+	if ( sv_wwwFallbackURL.Get().empty() )
 	{
 		return false;
 	}
 
-	Log::Notice( "clientDownload: sending client '%s' to fallback URL '%s'\n", cl->name, sv_wwwFallbackURL->string );
+	Log::Notice( "clientDownload: sending client '%s' to fallback URL '%s'\n", cl->name, sv_wwwFallbackURL.Get().c_str() );
```
necessarily-equal, line 670: I think you don't need the .c_str()
```suggestion
	Log::Notice( "clientDownload: sending client '%s' to fallback URL '%s'\n", cl->name, sv_wwwFallbackURL.Get() );
```

src/engine/server/sv_client.cpp
```diff
@@ -40,6 +40,10 @@ Maryland 20850 USA.
 #include "qcommon/sys.h"
 #include <common/FileSystem.h>
 
+Cvar::Cvar<bool> sv_wwwDownload("sv_wwwDownload", "server does a www dl redirect", Cvar::NONE, "");
+Cvar::Cvar<std::string> sv_wwwBaseURL("sv_wwwBaseURL", "base URL for redirect", Cvar::NONE, "");
+Cvar::Cvar<std::string> sv_wwwFallbackURL("sv_wwwFallbackURL", "URL to send to if an http/ftp fails or is refused client side", Cvar::NONE, "");
```
DolceTriade, line 45: Is FTP really supported?

necessarily-equal (approved):
LGTM

src/common/Defs.h
```diff
@@ -53,8 +53,6 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define MASTER4_SERVER_NAME ""
 #define MASTER5_SERVER_NAME ""
 
-#define WWW_BASEURL         "dl.unvanquished.net/pkg"
```
DolceTriade, line 56: What's wrong with this default? Also if you are going to change the default, you should at least provide a note and rationale in the commit message.

The commit message suggests this change has no functional impact.

src/engine/server/sv_client.cpp
bmorel: No idea, that was specified in a comment. I didn't rephrased those (no time right now to try to understand that sv_www code), only placed them in the cvar description field.

src/common/Defs.h
bmorel: oh, nothing is wrong. I only forgot to put it back in the value field :S

src/engine/server/sv_client.cpp
```diff
@@ -41,7 +41,7 @@ Maryland 20850 USA.
 #include <common/FileSystem.h>
 
 Cvar::Cvar<bool> sv_wwwDownload("sv_wwwDownload", "server does a www dl redirect", Cvar::NONE, "");
-Cvar::Cvar<std::string> sv_wwwBaseURL("sv_wwwBaseURL", "base URL for redirect", Cvar::NONE, "");
+Cvar::Cvar<std::string> sv_wwwBaseURL("sv_wwwBaseURL", "base URL for redirect", Cvar::NONE, "dl.unvanquished.net/pkg");
```
DolceTriade, line 44: So technically, it would be better to keep it with the define in Defs.h. SInce while Unvanquished is the only game that uses daemon, the idea is to appeal to other people to use daemon as well. It's less forking if people just provide their own Defs.h instead of having to touch random stuff all over the engine.

DolceTriade (approved):
LGTM % putting default URL back in Defs.h

src/engine/server/sv_client.cpp
bmorel: good point.

src/engine/server/sv_client.cpp
necessarily-equal: Pretty sure FTP isn't supported. It is only mentioned in a few outdated comments server-side I think.

slipher:
Suggested description strings

sv_wwwDownload - have clients download missing paks via HTTP
sv_wwwBaseURL - where clients download paks (must NOT be HTTPS, must contain PAKSERVER)
sv_wwwFallbackURL - alternative download site to sv_wwwBaseURL

src/engine/server/sv_client.cpp
bmorel: I used wording proposed by slipher , so there's no more mentions of FTP. In any case, I don't think it's a good idea to promote FTP overall, so even if supported by accident, no need to advertise the feature.

src/engine/server/sv_client.cpp
```diff
@@ -40,6 +40,10 @@ Maryland 20850 USA.
 #include "qcommon/sys.h"
 #include <common/FileSystem.h>
 
+Cvar::Cvar<bool> sv_wwwDownload("sv_wwwDownload", "have clients download missing paks via HTTP", Cvar::NONE, "");
+Cvar::Cvar<std::string> sv_wwwBaseURL("sv_wwwBaseURL", "where clients download paks (must NOT be HTTPS, must contain PAKSERVER)", Cvar::NONE, WWW_BASEURL);
+Cvar::Cvar<std::string> sv_wwwFallbackURL("sv_wwwFallbackURL", "alternative download site to sv_wwwBaseURL", Cvar::NONE, "");
```
necessarily-equal, line 45: (Note we could make this static and remove the extern declarations from the .h)

necessarily-equal (approved):
LGTM

PR #703, opened 2022-09-02
[RFC] Warn when server's download URL is https

necessarily-equal:
Fix https://github.com/Unvanquished/Unvanquished/issues/1744


necessarily-equal:
For now I'm requesting some help making this better. First, should we enforce a http:// URL instead? Can we make it impossible to set the value to an unused one? (I tried making `CheckURL` return a `Cvar::OnValueChangedResult` without success)

slipher:
HTTP is the only supported protocol, so how about complaining on anything besides http before a `://`?

PR #705, opened 2022-09-09
cg_api: Expose const Cmd::Args& trap_Args().

DolceTriade:
This doesn't break compat because this function is already compiled in, just not declared.


slipher:
LGTM

PR #711, opened 2022-09-18
client: Drop early server commands.

DolceTriade:
These server commands don't get processed until we have loaded the game. By then, they will be stale and can lead to bad state as in the stuck scoreboard bug.

### DIFF
--- a/src/engine/client/cl_parse.cpp
+++ b/src/engine/client/cl_parse.cpp
@@ -491,6 +491,11 @@ void CL_ParseCommandString( msg_t *msg )
 	clc.serverCommandSequence = seq;
 
 	index = seq & ( MAX_RELIABLE_COMMANDS - 1 );
+	if ( cls.state < connstate_t::CA_LOADING )
+	{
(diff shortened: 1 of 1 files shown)


necessarily-equal (dismissed):
On principle sounds good once https://github.com/Unvanquished/Unvanquished/pull/2206#pullrequestreview-1111567957 is merged.

I don't know this part of the code and haven't checked, because this requires building a full stack with nacl

DolceTriade:
This one is the dicey change as it does change some pretty old behaviors.

slipher:
I looked at the uses of server commands and some of them might be relevant in the loading stage, namely where the client is kicked / the server shuts down.

DolceTriade:
Hm, right. Maybe we need to just work around this in the cgame somehow.

illwieckz:
Is there something preventing to merge this one?

DolceTriade:
Yes the comments slipher said which are that it breaks bans/kicks.

PR #712, opened 2022-09-23
Produce external_deps archive with naclsdk ported to python3

illwieckz:
Produce external_deps archive with naclsdk ported to python3. I only tested with linux.


DolceTriade (approved):

DolceTriade:
Can we fix the CI and update the CI to work with python3 to verify this change?

illwieckz:
We just need to upload the to-be-next archive zip and restart the CI task.

illwieckz:
So what I did is that I extracted current deps archives, applied the patch, and repackaged. It builds on all systems.

external_deps/build.sh
```diff
@@ -521,6 +521,15 @@ build_naclsdk() {
 	cp pepper_*"/tools/irt_core_${NACLSDK_ARCH}.nexe" "${PREFIX}/irt_core-${DAEMON_ARCH}.nexe"
 	cp pepper_*"/toolchain/${NACLSDK_PLATFORM}_x86_newlib/bin/x86_64-nacl-gdb${EXE}" "${PREFIX}/nacl-gdb${EXE}"
 	rm -rf "${PREFIX}/pnacl"
+	
+	(
+		cd pepper_*"/toolchain/${NACLSDK_PLATFORM}_pnacl/bin/pydir"
+		if head -n 1 'pnacl-clang.py' | egrep -q '#!/usr/bin/python'
```
DolceTriade, line 527: Why do you need this? why not just apply unconditonally?

external_deps/build.sh
```diff
@@ -521,6 +521,15 @@ build_naclsdk() {
...
+		then
+			patch -p1 < "${SCRIPT_DIR}/naclsdk-pydir-python3.patch" >/dev/null
+		fi
```
DolceTriade, line 530: || exit -1?

cmake/toolchain-pnacl.cmake
```diff
@@ -51,12 +51,12 @@ set(CMAKE_C_RESPONSE_FILE_LINK_FLAG "@")
...
-    if (NOT Python2_Interpreter_FOUND)
-        message(FATAL_ERROR "Please install python2 (and run in a python2 virtualenv if that's not enough).")
+    find_package(Python3)
+    if (NOT Python3_Interpreter_FOUND)
+        message(FATAL_ERROR "Please install python3 (and run in a python3 virtualenv if that's not enough).")
     endif()
-    set(PNACLPYTHON_PREFIX "env PNACLPYTHON=${Python2_EXECUTABLE} ")
-    set(PNACLPYTHON_PREFIX2 env "PNACLPYTHON=${Python2_EXECUTABLE} ")
+    set(PNACLPYTHON_PREFIX "env PNACLPYTHON=${Python3_EXECUTABLE} ")
+    set(PNACLPYTHON_PREFIX3 env "PNACLPYTHON=${Python3_EXECUTABLE} ")
```
slipher, line 59: The "2" there has nothing to do with the Python version.

external_deps/naclsdk-pydir-python3.patch
```diff
@@ -0,0 +1,718 @@
+diff --git a/artools.py b/artools.py
+index 6fa6785..abc4c4a 100755
+--- a/artools.py
++++ b/artools.py
+@@ -1,4 +1,4 @@
+-#!/usr/bin/python
++#!/usr/bin/env python3
```
slipher, line 7: I'm asking to get rid of all these shebang changes. On *nix we always set PNACLPYTHON forcing a specific interpreter when invoking from CMake, so they have no effect.

CMakeLists.txt
```diff
@@ -177,7 +177,7 @@ include(DaemonPlatform)
 ################################################################################
 
 # Dependencies version, this must match the number in external_deps/build.sh
-set(DEPS_VERSION 6)
```
slipher, line 180: You have another outstanding pull request changing the deps package, so maybe you could merge only the patch file stuff in this PR and move the CMake changes to #713.

external_deps/build.sh
slipher, line 527: Why is there an `if`? The patch must always be applied.

slipher (commented):
I've pushed a few changes to the patch onto the branch, PTAL. I hit some problems with Windows-specific code sections, which I fixed. Also I restored Python 2 compatibility. Since some of the tools were not ported at all, and there may be bugs we need to investigate, I think we should allow either Python 2 or 3 for the interpreter.

We still need to test on Mac.

slipher:
After some additional patching the produced binaries seem to work on my machine. Nice job!

external_deps/build.sh
illwieckz: I was wondering if we need to make sure we haven't patched it yet, but it looks like the script deletes the folder first so no need to test indeed.

external_deps/naclsdk-pydir-python3.patch
illwieckz: Done.

external_deps/build.sh
illwieckz: Removed.

illwieckz:
I imported the patch to have all packages using `tar.xz` format from #713.

I also implemented the `deps-<system>-<arch>-<compiler>_<version>.<ext>` format. Note that I added a `deps-` prefix because it makes easy to manage files this way (for example one can do `rm deps-*.tar.xz`).

I uploaded ready-to-use packages. I haven't built them myself, what I did is that I used version 6 packages, renamed the folders, applied the python3 patch for naclsdk, and packafed as `tar.xz`.

If people can try to build the game by hand that would be appreciated, as it looks like @slipher caught some problems the CI didn't.

CMakeLists.txt
illwieckz: I opened that other PR to have a different thread to talk about things, but now that everything looks ready to merge, I imported all the PRs here to merge all of them in one go so we only have one new package to send per architecture.

slipher (commented):
Tried to build NaCl on Mac and didn't get past the configuration stage. From the CMake sub-invocation for NaCl it tried to download deps from `https://dl.unvanquished.net/deps/deps--i686-default_7.tar.xz`.

The CI only builds NaCl on an Ubuntu host.

CMakeLists.txt
```diff
@@ -206,25 +206,36 @@ if (CMAKE_CFG_INTDIR STREQUAL ".")
...
+
+if (MSVC)
+    set(DEPS_COMPILER msvc)
+elseif (WIN32)
+    set(DEPS_COMPILER mingw)
+else()
+    set(DEPS_COMPILER default)
 endif()
+
+set(DEPS_DIR ${EXTERNAL_DEPS_DIR}/deps-${DEPS_SYSTEM}-${DEPS_ARCH}-${DEPS_COMPILER}_${DEPS_VERSION})
```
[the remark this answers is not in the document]
illwieckz: I didn't proposed it before, that's why I mentioned it in my comment above.

external_deps/build.sh
```diff
@@ -764,7 +782,7 @@ export MAKEFLAGS="-j`nproc 2> /dev/null || sysctl -n hw.ncpu 2> /dev/null || ech
 
 # Setup platform
 PLATFORM="${1}"
-"setup_${PLATFORM}"
+"setup_${PLATFORM//-/_}"
```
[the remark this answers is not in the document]
illwieckz: Ah right, that's only a limitation of variable indeed.

external_deps/build.sh
```diff
@@ -521,6 +521,12 @@ build_naclsdk() {
 	cp pepper_*"/tools/irt_core_${NACLSDK_ARCH}.nexe" "${PREFIX}/irt_core-${DAEMON_ARCH}.nexe"
 	cp pepper_*"/toolchain/${NACLSDK_PLATFORM}_x86_newlib/bin/x86_64-nacl-gdb${EXE}" "${PREFIX}/nacl-gdb${EXE}"
 	rm -rf "${PREFIX}/pnacl"
+	
+	(
+		cd pepper_*"/toolchain/${NACLSDK_PLATFORM}_pnacl/bin/pydir"
```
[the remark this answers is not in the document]
illwieckz: Right.

CMakeLists.txt
```diff
@@ -206,25 +206,36 @@ if (CMAKE_CFG_INTDIR STREQUAL ".")
...
 set(EXTERNAL_DEPS_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external_deps" CACHE STRING "Directory in which to store the downloaded dependencies.")
 set(DEPS_EXT ".tar.xz")
+
+if (ARCH STREQUAL "x86_64")
+    set(DEPS_ARCH amd64)
+else()
+    set(DEPS_ARCH i686)
+endif()
+
 if (MSVC)
```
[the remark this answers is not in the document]
illwieckz: I don't understood what you meant.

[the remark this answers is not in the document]
slipher: Do "if win32 then platform = windows" not "if msvc then platform = windows else if win32 then platform = windows"

external_deps/build.sh
```diff
@@ -716,12 +714,12 @@ setup_mingw64() {
...
 
 # Set up environment for Mac OS X 64-bit
-setup_macosx64() {
+setup_macos-amd64-default() {
 	HOST=x86_64-apple-darwin11
 	CROSS=
 	MSVC_SHARED=(--disable-shared --enable-static)
 	export MACOSX_DEPLOYMENT_TARGET=10.9
-	export NASM="${PWD}/build-${PLATFORM}-${DEPS_VERSION}/prefix/bin/nasm" # A newer version of nasm is required for 64-bit
+	export NASM="${PWD}/${BUILD_BASEDIR}/prefix/bin/nasm" # A newer version of nasm is required for 64-bit
```
slipher, line 722: I hit a bug here - the definition of BUILD_BASEDIR comes from common_setup. You can fix it by moving this line to the bottom of the function.

illwieckz: Done.

CMakeLists.txt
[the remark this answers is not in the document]
illwieckz: AH! Right.

Done.

illwieckz:
> Tried to build NaCl on Mac and didn't get past the configuration stage. From the CMake sub-invocation for NaCl it tried to download deps from https://dl.unvanquished.net/deps/deps--i686-default_7.tar.xz.

I fail to understand how that can happen. 😲️

illwieckz:
> > Tried to build NaCl on Mac and didn't get past the configuration stage. From the CMake sub-invocation for NaCl it tried to download deps from `https://dl.unvanquished.net/deps/deps--i686-default_7.tar.xz`.
> 
> I fail to understand how that can happen. astonished

I also got this on Linux, after multiple successful attempts:

```
Downloading dependencies from 'https://dl.unvanquished.net/deps/deps--i686-default_7.tar.xz'
-- [download 0% complete]
CMake Error at daemon/CMakeLists.txt:458 (message):
  Error downloading
  'https://dl.unvanquished.net/deps/deps--i686-default_7.tar.xz':
```

That doesn't make sense.

necessarily-equal:
Minimal testing and looking suggests that it work (Linux / NixOS / python2 only). LGTM

slipher:
I get the bogus deps URL when starting from an empty build dir, configuring with BUILD_GAME_NACL=1 and immediately attempting to build the `nacl-vms` target.

illwieckz:
> I get the bogus deps URL on Mac when starting from an empty build dir, configuring with BUILD_GAME_NACL=1 and immediately attempting to build the `nacl-vms` target.

When this happen, `NACL` is set.

I did: 

```
if (WIN32)
    set(DEPS_SYSTEM windows)
elseif (APPLE)
    set(DEPS_SYSTEM macos)
elseif (LINUX)
    set(DEPS_SYSTEM linux)
elseif (NACL)
    set(DEPS_SYSTEM nacl)
endif()
```

And I got:

```
  Error downloading
  'https://dl.unvanquished.net/deps/nacl-i686-default_7.tar.xz':
```

slipher:
Looks like the way it works now is that the parent cmake instance sets `-DDEPS_DIR=xxx` and the child instance carefully avoids overwriting this variable in the nacl case.

(Which is not quite right btw; it means that building nacl won't work in a cross-compile build. If you want you could take this opportunity to determine the *host system architecture* in the nacl case and download the deps for that :stuck_out_tongue:)

illwieckz:
> (Which is not quite right btw; it means that building nacl won't work in a cross-compile build. If you want you could take this opportunity to determine the host system architecture in the nacl case and download the deps for that stuck_out_tongue)

I don't really understand that. Anyway I'm surprised the code worked before… How did it worked before?

slipher:
As I said it's the `-DDEPS_DIR=${DEPS_DIR}` flag to the child cmake invocation in DaemonGame.cmake. It stopped working because you started to always write the value of DEPS_DIR, instead of leaving it untouched when NACL is on.

illwieckz:
> As I said it's the `-DDEPS_DIR=${DEPS_DIR}` flag to the child cmake invocation in DaemonGame.cmake. It stopped working because you started to always write the value of DEPS_DIR, instead of leaving it untouched when NACL is on.

Ah right, there was no NACL mention but no `if` for it so since `DEPS_DIR` was unset the download code was ignored.

So I assume the added `if (NACL)` test is the correct way to do it.

CMakeLists.txt
```diff
@@ -427,9 +431,9 @@ if (DEPS_DIR)
     get_cmake_property(cacheVariableNames CACHE_VARIABLES)
     foreach(cacheVariableName ${cacheVariableNames})
         set(cacheVariableValue "${${cacheVariableName}}")
-        if (cacheVariableValue MATCHES "/(linux64|macosx64|mingw32|mingw64|msvc32|msvc64)-([0-9]+)/")
-            if (NOT ${CMAKE_MATCH_2} EQUAL ${DEPS_VERSION})
-                message(WARNING "You seem to be using a dependency from version ${CMAKE_MATCH_2}"
+        if (cacheVariableValue MATCHES "/(linux|windows|macos)-(i686|amd64)-(default,mingw,msvc)_([0-9]+)/")
```
slipher, line 434: You need to detect the old naming scheme too.

illwieckz: That should be fixed.

illwieckz:
@slipher look at my latest commit, it makes separate `<system>-<arch>-pnacl` packages.

The `7a` version is temporary (will be `7` if it works).

illwieckz:
I give up for now for splitting the archives. It looks like it uncovers bugs like looking for system lua when building nexe and things like that.

So I removed the commit that split the archives and such effort would be done later (way later), I don't have time to investigate archive splitting more right now.

I'm considering this ready if it builds.

CMakeLists.txt
```diff
@@ -206,27 +203,42 @@ if (CMAKE_CFG_INTDIR STREQUAL ".")
 else()
     set(FULL_OUTPUT_DIR ${CMAKE_BINARY_DIR}/${CMAKE_CFG_INTDIR})
 endif()
-set(EXTERNAL_DEPS_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external_deps" CACHE STRING "Directory in which to store the downloaded dependencies.")
-set(DEPS_EXT ".tar.xz")
-if (MSVC)
+
+    # Dependencies version, this must match the number in external_deps/build.sh
```
slipher, line 207: Bad spacing here

slipher (commented):
LGTM. I checked that the deps build script still works for Windows targets.

I guess you'll need to squash the commits (other than the .deb one) to make all commits buildable?

CMakeLists.txt
```diff
@@ -426,17 +430,29 @@ if (DEPS_DIR)
...
+        endif()
+
+        if (cacheVariableValue MATCHES "/(linux|windows|macos)-(i686|amd64)-(default|mingw|msvc)_([0-9]+)/")
+            if (NOT ${CMAKE_MATCH_4} EQUAL ${DEPS_VERSION})
+                set(OLD_VERSION ${CMAKE_MATCH_4})
             endif()
         endif()
+
+        if(OLD_VERSION)
+            message(WARNING "You seem to be using a dependency from version ${CMAKE_MATCH_4}"
```
slipher, line 449: This needs to be OLD_VERSION

illwieckz: fixed.

CMakeLists.txt
illwieckz: fixed.

illwieckz:
> I guess you'll need to squash the commits (other than the .deb one) to make all commits buildable?

I don't like squashing but yes, we better want to have commits buildable, so I squashed?

slipher:
The nacl patch could have gone into its own commit if you put the Python cmake changes into the squashed commit, since I made it work with Python 2.

LGTM

illwieckz:
Right, I did that as I prefer that.

I made sure to move the `DEPS_VERSION` incrementing to the commit changing the archive name and format.

slipher:
I mean the stuff that looks for Python 2/3 instead of Python 2 needs to be in (or after) the big commit.

illwieckz:
@slipher, here is what I did:

Except the commit removing the support for the old nacl deb which is just cleaning-up things,

- first commit introduce the code to produce python2/3 compatible deps archive;
- second commit use the new deps archive (with python2/3 compatibility) and other changes (name, format…);
- third commit enable the ability to use python3.

I think it's now ready for merging! 🙂️

slipher:
The commit for enabling Python 3 says that I am the author which is not right.

Otherwise LGTM

illwieckz:
🎉️

PR #715, opened 2022-10-01
CI: fix Windows bitness configuration

slipher:
Both of the Windows builds were just building the same thing (64-bit). One of them is intended to be 32-bit.


DolceTriade (approved):

necessarily-equal (commented):
Looks ok, I'll trust you on that

PR #717, opened 2022-10-13
navedit: less flickering

necessarily-equal:
This makes navmesh flickering less jaggering, even if there are still some weirdness.

The problem this fixes is that the botDebugDrawCommands buffer was cleared after every rendering frame, while the server frame which did set the value (calling RE_SendBotDebugDrawCommands) did run less often, likely at a rate of sv_fps.

A good way to show how this change is useful is to notice how the flickering was much more annoying at /timescale 0.1, but is now not any worse.

@bmorel I think you'd be interested in that one too, since you complained loudly about that flickering before.

### DIFF
--- a/src/engine/renderer/tr_main.cpp
+++ b/src/engine/renderer/tr_main.cpp
@@ -2696,7 +2696,7 @@ static void RunBotDebugDrawCommands()
 	}
 
 	Util::Reader r;
-	std::swap( r.GetData(), botDebugDrawCommands );
+	r.GetData() = botDebugDrawCommands;
 	while ( true )
 	{
 		debugDrawMode_t mode;


DolceTriade:
This does reduce the flickering, but does not eliminate it completely.

DolceTriade:
LGTM

slipher:
Pretty sure the client and server always run at the same rate in a local game. Note the `Com_IsDedicatedServer`:

```
	if ( Com_IsDedicatedServer() && sv.timeResidual < frameMsec )
	{
		// NET_Sleep will give the OS time slices until either get a packet
		// or time enough for a server frame has gone by
		NET_Sleep( frameMsec - sv.timeResidual );
		return;
	}
```

necessarily-equal:
Observation suggests that this isn't always true then, or that there is 
some sort of desync. Or maybe it only happens on low fps (which happens 
because the client has a hard time displaying all these navmesh)?

Anyway this piece of code you mention is on the server side. Which 
suggests the server would wait on the client, but not necessarily the 
other way around. And that's the other way around that matters.

PR #720, opened 2022-10-20
Remove all trap_CM_* defines

necessarily-equal:
See https://github.com/Unvanquished/Unvanquished/pull/2267 for the unvanquished corresponding commit. This should be merged afterwards.


DolceTriade (approved):

necessarily-equal:
Just thought about this for a second, what if not merging this would help porting other games? Would that make it worth keeping?

DolceTriade:
It might, but there is going to be a bunch of menial renaming/adjusting in any effort to port new games regardless of whether we we keep the trap_ prefix or not. I don't think we should accumulate technical debt for this.

PR #721, opened 2022-10-25
Cvar namespaces + unit tests

### LINKED ISSUE #449: Improve support to cvar namespaces

slipher:
1. Add namespace-aware cvar autocompletion (fixes #449). Thanks to @peoro for the original idea.
2. Introduce unit testing using Googletest (a.k.a GTest). To start with there are tests for the cvar namespace collapsing.

Our codebase is not very modular in terms of having independently buildable units; I have worked around this by building the tests directly into the engine binary. A `test` command is used to run the tests. (We could consider a `-test` command line flag instead, but I fear that would run into problems later with things not being initialized.) This approach could be nice for the gamelogic too since it is not easy to make a standalone NaCl binary.

The tests I added are for some `engine/framework` stuff so they are built into all engine binaries including dummyapp. Other tests might be client-specific; they can be built into `daemon` and `daemon-tty`. 

To do before merging:

- [ ] Test external_deps build on Windows and Mac
- [ ] Redirect output to the console functions instead of stdout
- [x] Ensure that configs (autogen, binds) are not read or written

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "libs/crunch"]
 	path = libs/crunch
 	url = https://github.com/DaemonEngine/crunch.git
+[submodule "libs/googletest"]
+	path = libs/googletest
+	url = https://github.com/google/googletest.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,8 @@ if(NOT DAEMON_EXTERNAL_APP)
     option(USE_BREAKPAD "Generate Daemon crash dumps (which require Breakpad tools to read)" OFF)
 endif()
 
+option(BUILD_TESTS "Build unit test applications" OFF)
+
 option(USE_LTO "Use link-time optimization for release builds" OFF)
 cmake_dependent_option(USE_SLIM_LTO "Generate slim LTO objects, improves build times" 1 "USE_LTO AND \"${CMAKE_CXX_COMPILER_ID}\" STREQUAL GNU" 0)
 option(USE_HARDENING "Use stack protection and other hardening flags" OFF)
@@ -491,13 +493,13 @@ if (DEPS_DIR)
     set(CMAKE_FIND_ROOT_PATH ${DEPS_DIR} ${CMAKE_FIND_ROOT_PATH})
     set(CMAKE_INCLUDE_PATH ${DEPS_DIR} ${DEPS_DIR}/include ${CMAKE_INCLUDE_PATH})
     set(CMAKE_FRAMEWORK_PATH ${DEPS_DIR} ${CMAKE_FRAMEWORK_PATH})
-    set(CMAKE_LIBRARY_PATH ${DEPS_DIR}/lib ${CMAKE_LIBRARY_PATH})
+    set(CMAKE_PREFIX_PATH ${DEPS_DIR} ${CMAKE_PREFIX_PATH})
     if (DAEMON_PARENT_SCOPE_DIR)
         # Also set parent scope so the top level CMakeLists can find precompiled deps
         set(CMAKE_FIND_ROOT_PATH ${DEPS_DIR} ${CMAKE_FIND_ROOT_PATH} PARENT_SCOPE)
         set(CMAKE_INCLUDE_PATH ${DEPS_DIR} ${DEPS_DIR}/include ${CMAKE_INCLUDE_PATH} PARENT_SCOPE)
         set(CMAKE_FRAMEWORK_PATH ${DEPS_DIR} ${CMAKE_FRAMEWORK_PATH} PARENT_SCOPE)
-        set(CMAKE_LIBRARY_PATH ${DEPS_DIR}/lib ${CMAKE_LIBRARY_PATH} PARENT_SCOPE)
+        set(CMAKE_PREFIX_PATH ${DEPS_DIR} ${CMAKE_PREFIX_PATH} PARENT_SCOPE)
     endif()
 
     # Force OpenAL Soft on Mac, otherwise the system framework will be selected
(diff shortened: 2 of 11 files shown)


src/engine/framework/TestRunner.cpp
```diff
@@ -0,0 +1,26 @@
+#include <gtest/gtest.h>
```
DolceTriade, line 1: imo, we should just make a Test app. Similar to how we have the "DummyApp" to just run tests?

slipher:
> imo, we should just make a Test app. Similar to how we have the "DummyApp" to just run tests?

I considered and rejected that earlier, but on second thought the `Application` framework is useful here because there is the `Initialize` function which does the per-app initialization procedures. There's an annoying thing though - we'll probably need at least two test apps. Because we'll probably have some tests that only run in the full client, but it would be really annoying if the other tests can't run headlessly.

So now I have hope that the `test` command won't be necessary, but I'm undecided between two other options.
1. Make a `-test` command line option which does `Application::Initialize()` and then runs the tests.
2. Have multiple test Applications which inherit from the corresponding non-test applications and override `Frame` to run the tests.

A drawback of (2) is that it might be hard to avoid building a whole bunch of files twice, once for the test application and once for the normal application.

DolceTriade:
I'm kinda curious what tests would not be able to run headlessly? To me, it sounds like its actually undesirable to build these into the client because it doesn't allow us to run unit tests on Foo_Init() and Foo_Shutdown() if they are already called by the client and may interfere by a non-clean state.

slipher:
> I'm kinda curious what tests would not be able to run headlessly?

I'm not thinking of stuff that actually accesses the screen or keyboard, but rather stuff that is still "pure computation" but only built in the full client. For example q3 shader parsing, as part of the renderer.

> To me, it sounds like its actually undesirable to build these into the client because it doesn't allow us to run unit tests on Foo_Init() and Foo_Shutdown() if they are already called by the client and may interfere by a non-clean state.

Maybe I'm overestimating the problem, but I think many things will be hard to test without performing program initialization. For example functions that use hunk allocation will crash if Hunk_Init was not called.

DolceTriade:
Sure, maybe I'm also underestimating the problem, but keep in mind these are precisely the types of things we would want to validate in tests to catch memory leaks and validate our assumptions. Perhaps it is a little more work to decouple dependencies, but that will lead to a healthier code base.

For example, I could envision a test like:

```
class ShaderTest : ::testing::Test {
  protected:
    void SetUp() override {
      Hunk_Init();
      // blah
    }

    void TearDown() override {
      Hunk_Free();      
    }
};

TEST_F(ShaderTest, TestFoo) {
  // test
}
```

In the interim we could just hack it by calling `R_Init()` or something in `SetUp()` if we want to unblock on writing tests as well.

slipher:
I dislike this example - it's exactly what I want to avoid. Making people do a lot of work to initialize the global state before they can write the test. I'm not going to write a test if I have to spend way more time on fixing the test itself than validating my actual code.

I say, put the onus to restructure the code on people who want to test init functions, rather than everyone who writes tests. If you want to test init functions, move the subsystem state into a class so that you can make a separate instance of it besides the global one. See DNSResolver for an example; there is one global instance but if you put the class in a header, you'd be able to make another one for testing. 

By the way, `Hunk_Free` does not exist. There is no function to deallocate the hunk as it is intended to persist for the lifetime of the program. A perfect illustration of my point :stuck_out_tongue:

DolceTriade:
Fair. I still think that initializing global state will help write better tests and be more resiliant than tests that run based on global state that can change and break assumptions.

slipher:
I changed the approach from a `/test` command to having test applications like `test-daemonded`, `test-daemon-tty` etc. Next steps for this approach:
- Add an application trait for default homepath and make it a temp directory for test applications
- Try to avoid building all files twice when you build both the normal and test version of an application, and generally make the cmake less horrible

slipher:
Did the things in the previous comment. A temporary homepath is used by default (which can be changed by `-homepath`), and all the object files are reused between the normal and test applications.

DolceTriade:
How are the tests run now? I don't see any new main() file. IIRC, gunit_main provides a main file for regular tests. I assume we would need to override it to call our own tests?

slipher:
Made Googletest a submodule instead of an external_deps package. In the Unvanquished CMakeLists.txt we can make the submodule initialization code avoid downloading it when tests are disabled.

I decided there is no need to make gtest's output go through Daemon's logging system. Writing to stdout works fine.

There is one remaining bug which I will not fix for now, which is that on Windows the temporary homepath cannot be deleted because daemon.log is still open. (Deletion of the log file itself succeeds, but not the containing folder). This is ready for review.

DolceTriade:
LGTM. Did you do all the stuff on your checkmark list in the PR header?

slipher:
> * [ ]  Test external_deps build on Windows and Mac

Obsolete because it is using a submodule instead of external_deps.

> * [ ]  Redirect output to the console functions instead of stdout

I decided this is a stupid/pointless idea.

> * [x]  Ensure that configs (autogen, binds) are not read or written

I did that.

DolceTriade (approved):

PR #727, opened 2022-11-04
renderer: fallback on `EXT_framebuffer_object` if `ARB_framebuffer_object` is missing

illwieckz:
Fallback on `EXT_framebuffer_object` if `ARB_framebuffer_object` is missing.

`ARB_framebuffer_object` does more things than `EXT_framebuffer_object` so hardware supporting `EXT_framebuffer_object` may not support `ARB_framebuffer_object` but it looks like our game does not do crazy things enough and `EXT_framebuffer_object` seems to be usable as a fallback for missing `ARB_framebuffer_object`.

This makes the engine run on all Nvidia NV40 (Curie) on Linux with nouveau nv30 driver.

The proprietary nvidia 304 driver emulated `ARB_framebuffer_object` over `EXT_framebuffer_object` but nouveau doesn't, and the proprietary nvidia 304 driver isn't provided on Linux distributions anymore, leaving nouveau the only driver for this range of hardware.

I first tried to enable `ARB_framebuffer_object` on Mesa directly, but it looks like the hardware doesn't support everything required for it and I don't have enough knowledge to emulate it over `EXT_framebuffer_object`:

- https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/19513

This is implemented in the form of a struct of OpenGL functions pointers, setting the ones from `ARB_framebuffer_object` or `EXT_framebuffer_object` given what extension is provided, and calling them through the struct.

Before this PR is merged and the engine rebuilt and released, people can do the same as this PR with the `MESA_EXTENSION_OVERRIDE=GL_ARB_framebuffer_object` environment variable.

Note that running the Dæmon engine over a nv40 hardware will suffer from an nouveau nv30 driver bug related to vbo model skinning. One can workaround this bug by setting `r_vboVertexSkinning 0` to force the CPU code path for this task:

- https://gitlab.freedesktop.org/mesa/mesa/-/issues/7631

### DIFF
--- a/src/engine/renderer/tr_fbo.cpp
+++ b/src/engine/renderer/tr_fbo.cpp
@@ -34,13 +34,13 @@ bool R_CheckFBO( const FBO_t *fbo )
 	int id;
 
 	glGetIntegerv( GL_FRAMEBUFFER_BINDING, &id );
-	glBindFramebuffer( GL_FRAMEBUFFER, fbo->frameBuffer );
+	GL_fboShim.glBindFramebuffer( GL_FRAMEBUFFER, fbo->frameBuffer );
 
-	code = glCheckFramebufferStatus( GL_FRAMEBUFFER );
+	code = GL_fboShim.glCheckFramebufferStatus( GL_FRAMEBUFFER );
 
 	if ( code == GL_FRAMEBUFFER_COMPLETE )
 	{
-		glBindFramebuffer( GL_FRAMEBUFFER, id );
+		GL_fboShim.glBindFramebuffer( GL_FRAMEBUFFER, id );
 		return true;
 	}
 
@@ -72,7 +72,7 @@ bool R_CheckFBO( const FBO_t *fbo )
 			break;
 	}
 
-	glBindFramebuffer( GL_FRAMEBUFFER, id );
+	GL_fboShim.glBindFramebuffer( GL_FRAMEBUFFER, id );
 
 	return false;
 }
@@ -112,7 +112,7 @@ FBO_t          *R_CreateFBO( const char *name, int width, int height )
 	fbo->width = width;
 	fbo->height = height;
 
-	glGenFramebuffers( 1, &fbo->frameBuffer );
+	GL_fboShim.glGenFramebuffers( 1, &fbo->frameBuffer );
 
 	return fbo;
 }
@@ -140,15 +140,15 @@ void R_CreateFBOColorBuffer( FBO_t *fbo, int format, int index )
 
 	if ( absent )
 	{
-		glGenRenderbuffers( 1, &fbo->colorBuffers[ index ] );
+		GL_fboShim.glGenRenderbuffers( 1, &fbo->colorBuffers[ index ] );
 	}
 
-	glBindRenderbuffer( GL_RENDERBUFFER, fbo->colorBuffers[ index ] );
-	glRenderbufferStorage( GL_RENDERBUFFER, format, fbo->width, fbo->height );
+	GL_fboShim.glBindRenderbuffer( GL_RENDERBUFFER, fbo->colorBuffers[ index ] );
+	GL_fboShim.glRenderbufferStorage( GL_RENDERBUFFER, format, fbo->width, fbo->height );
 
 	if ( absent )
 	{
-		glFramebufferRenderbuffer( GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0 + index, GL_RENDERBUFFER,
+		GL_fboShim.glFramebufferRenderbuffer( GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0 + index, GL_RENDERBUFFER,
 					   fbo->colorBuffers[ index ] );
 	}
 
@@ -177,15 +177,15 @@ void R_CreateFBODepthBuffer( FBO_t *fbo, int format )
 
 	if ( absent )
 	{
-		glGenRenderbuffers( 1, &fbo->depthBuffer );
+		GL_fboShim.glGenRenderbuffers( 1, &fbo->depthBuffer );
 	}
 
-	glBindRenderbuffer( GL_RENDERBUFFER, fbo->depthBuffer );
-	glRenderbufferStorage( GL_RENDERBUFFER, fbo->depthFormat, fbo->width, fbo->height );
+	GL_fboShim.glBindRenderbuffer( GL_RENDERBUFFER, fbo->depthBuffer );
+	GL_fboShim.glRenderbufferStorage( GL_RENDERBUFFER, fbo->depthFormat, fbo->width, fbo->height );
 
 	if ( absent )
 	{
-		glFramebufferRenderbuffer( GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_RENDERBUFFER, fbo->depthBuffer );
+		GL_fboShim.glFramebufferRenderbuffer( GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_RENDERBUFFER, fbo->depthBuffer );
 	}
 
 	GL_CheckErrors();
@@ -214,16 +214,16 @@ void R_CreateFBOStencilBuffer( FBO_t *fbo, int format )
 
 	if ( absent )
 	{
-		glGenRenderbuffers( 1, &fbo->stencilBuffer );
+		GL_fboShim.glGenRenderbuffers( 1, &fbo->stencilBuffer );
 	}
 
-	glBindRenderbuffer( GL_RENDERBUFFER, fbo->stencilBuffer );
-	glRenderbufferStorage( GL_RENDERBUFFER, fbo->stencilFormat, fbo->width, fbo->height );
+	GL_fboShim.glBindRenderbuffer( GL_RENDERBUFFER, fbo->stencilBuffer );
+	GL_fboShim.glRenderbufferStorage( GL_RENDERBUFFER, fbo->stencilFormat, fbo->width, fbo->height );
 	GL_CheckErrors();
 
 	if ( absent )
 	{
-		glFramebufferRenderbuffer( GL_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_RENDERBUFFER, fbo->stencilBuffer );
+		GL_fboShim.glFramebufferRenderbuffer( GL_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_RENDERBUFFER, fbo->stencilBuffer );
 	}
 
 	GL_CheckErrors();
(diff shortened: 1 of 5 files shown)


illwieckz:
Before:

```
SDL_Init( SDL_INIT_VIDEO )...  
Using SDL version 2.0.20 
SDL using driver "x11" 
Initializing OpenGL display 
Using GLEW version 2.2.0 
Display aspect: 1.779 
...setting mode -1: 640×480 
Using preferred context - 24-bit OpenGL 2.1 compatibility 
OpenGL Renderer: NV4E 
Using GL3 Renderer in OpenGL 2.x mode... 
Available modes: '1280x720 1366x768 1360x768 1280x768 1280x800 640x480 800x600 1024x768 1280x960 1280x1024 ' 
Detected graphics driver class 'integrated' 
Detected graphics hardware class 'generic' 
Initializing OpenGL extensions 
...ignoring GL_ARB_debug_output 
...found shading language version 120 
...using GL_ARB_half_float_pixel 
...GL_ARB_texture_float not found. 
...GL_EXT_gpu_shader4 not found. 
...GL_EXT_texture_integer not found. 
...GL_ARB_texture_rg not found. 
...GL_ARB_texture_gather not found. 
...using GL_EXT_texture_compression_s3tc 
...GL_ARB_texture_compression_rgtc not found. 
...ignoring GL_EXT_texture_filter_anisotropic 
...using GL_ARB_half_float_vertex 
...GL_ARB_framebuffer_object not found. 
Warn: Required extension GL_ARB_framebuffer_object is missing. 
```

After:

```
SDL_Init( SDL_INIT_VIDEO )...  
Using SDL version 2.0.20 
SDL using driver "x11" 
Initializing OpenGL display 
Using GLEW version 2.2.0 
Display aspect: 1.779 
...setting mode -1: 640×480 
Using preferred context - 24-bit OpenGL 2.1 compatibility 
]Mesa: User error: GL_INVALID_ENUM in glGetIntegerv(pname=GL_CONTEXT_PROFILE_MASK)
OpenGL Renderer: NV4E 
Using GL3 Renderer in OpenGL 2.x mode... 
Available modes: '1280x720 1366x768 1360x768 1280x768 1280x800 640x480 800x600 1024x768 1280x960 1280x1024 ' 
Detected graphics driver class 'integrated' 
Detected graphics hardware class 'generic' 
Initializing OpenGL extensions 
...ignoring GL_ARB_debug_output 
...found shading language version 120 
...using GL_ARB_half_float_pixel 
...GL_ARB_texture_float not found. 
...GL_EXT_gpu_shader4 not found. 
...GL_EXT_texture_integer not found. 
...GL_ARB_texture_rg not found. 
...GL_ARB_texture_gather not found. 
...using GL_EXT_texture_compression_s3tc 
...using GL_ARB_texture_compression_rgtc 
...ignoring GL_EXT_texture_filter_anisotropic 
...using GL_ARB_half_float_vertex 
...GL_ARB_framebuffer_object not found. 
...using GL_EXT_framebuffer_object 
...using GL_ARB_get_program_binary 
...using GL_ARB_buffer_storage 
...GL_ARB_uniform_buffer_object not found. 
...using GL_ARB_map_buffer_range 
...using GL_ARB_sync 
Warn: SSAO not used because GL_ARB_texture_gather is not available. 
Warn: floating point image '_depthtile1Render' cannot be loaded 
Warn: floating point image '_depthtile2Render' cannot be loaded 
Warn: floating point image '_dlightImage' cannot be loaded 
GL_VENDOR: nouveau 
GL_RENDERER: NV4E 
GL_VERSION: 2.1 Mesa 22.4.0-devel (git-dc08875415) 
GL_MAX_TEXTURE_SIZE: 4096 
GL_SHADING_LANGUAGE_VERSION: 1.20 
GL_MAX_VERTEX_UNIFORM_COMPONENTS 1848 
GL_MAX_VERTEX_ATTRIBS 16 
Occlusion query bits: 64 
GL_MAX_DRAW_BUFFERS: 4 
GL_MAX_RENDERBUFFER_SIZE: 4096 
GL_MAX_COLOR_ATTACHMENTS: 4 
PIXELFORMAT: color(24-bits) 
MODE: -1, 640 x 480 windowed hz: N/A 
Using OpenGL version 2.1, requested: 2.1 
Using OpenGL 2.x context. 
Using OpenGL extensions: GL_ARB_half_float_pixel GL_EXT_texture_compression_s3tc GL_ARB_texture_compression_rgtc GL_ARB_half_float_vertex GL_EXT_framebuffer_object GL_ARB_get_program_binary GL_ARB_buffer_storage GL_ARB_map_buffer_range GL_ARB_sync 
Missing OpenGL extensions: GL_ARB_texture_float GL_EXT_gpu_shader4 GL_EXT_texture_integer GL_ARB_texture_rg GL_ARB_texture_gather GL_ARB_framebuffer_object GL_ARB_uniform_buffer_object 
Using S3TC (DXTC) texture compression. 
```

Some demonstration:

[nouveau nv40](https://dl.illwieckz.net/b/unvanquished/wip/nouveau-nv40/20221104-172053-000.unvanquished-nouveau-nv40.png)



[nouveau nv40](https://dl.illwieckz.net/b/unvanquished/wip/nouveau-nv40/20221104-172115-000.unvanquished-nouveau-nv40.png)

[nouveau nv40](https://dl.illwieckz.net/b/unvanquished/wip/nouveau-nv40/20221104-172126-000.unvanquished-nouveau-nv40.png)

[nouveau nv40](https://dl.illwieckz.net/b/unvanquished/wip/nouveau-nv40/20221104-172202-000.unvanquished-nouveau-nv40.png)

[nouveau nv40](https://dl.illwieckz.net/b/unvanquished/wip/nouveau-nv40/20221104-172442-000.unvanquished-nouveau-nv40.png)

[nouveau nv40](https://dl.illwieckz.net/b/unvanquished/wip/nouveau-nv40/20221104-172805-000.unvanquished-nouveau-nv40.png)

[nouveau nv40](https://dl.illwieckz.net/b/unvanquished/wip/nouveau-nv40/20221104-172905-000.unvanquished-nouveau-nv40.png)

DolceTriade:
Do you think this is better done in the mesa driver? Is this only required for the nouveau driver?

illwieckz:
This is only required on nouveau driver with nv30 range of Nvidia hardware. It is believed the hardware doesn't have all the features to fully implement `ARB_framebuffer_object` but is OK with the earlier version of it being named `EXT_framebuffer_object`. `ARB_framebuffer_object` and `EXT_framebuffer_object` are not equal, `EXT_framebuffer_object` is a subset of `ARB_framebuffer_object`.

The proprietary nvidia 304 driver for those GPUs implements `ARB_framebuffer_object` but probably at the cost of some emulation, and this driver is incompatible with in current distros, while being very unstable (much more unstable than nouveau, which isn't stellar on this hardware, meaning nouveau is the only usable option. I doubt someone will invest the time, resources, and effort to emulate  `ARB_framebuffer_object` over the less complete `EXT_framebuffer_object`.

It happens that if a game only sticks to what `EXT_framebuffer_object` can do (which is a subset of `ARB_framebuffer_object`), it is possible to run the game with the same code, like my patch does. It looks like this happens with Unvanquished, I have only tested low presets but anyway such hardware require low presets so that's fine.

The only reasonable implementation in Mesa (other than the very complex task of emulating  `ARB_framebuffer_object` over the less complete `EXT_framebuffer_object`) looks to be the patch I proposed to Mesa: a game that knows it can fake ARB fbo over EXT fbo would tell Mesa to fake the availability of ARB fbo while in fact doing EXT fbo with EXT fbo limitations.

Some comparisons between one solution and another:

- Implementing such knob in Mesa:
  * Pros: game knowing EXT fbo is enough only have to set one environment variable before loading the GL subsystem, done.
  * Cons: every distro on earth has to ship a Mesa update for one game.
- Implementing a function pointer layer to call ARB fbo or EXT fbo pointers
  * Pros: works today on already release nouveau driver on every current distro on earth.
  * Cons: adds a function call indirection for every call of those functions for every other driver and hardware.

Well, there is a third solution, that can work today without function call indirection neither patching mesa: relying on the non-nouveau specific `MESA_EXTENSION_OVERRIDE=GL_ARB_framebuffer_object` environment variable. To implement it properly, one has to first load the GL subsystem, detect this is a nouveau driver with EXT fbo, shutdown the GL subsystem, _add_ the `GL_ARB_framebuffer_object` string to the `MESA_EXTENSION_OVERRIDE`, since it's a list, one may have to parse it to concatenate the added string, set the environment variable, and restart the GL subsystem… This is the most annoying solution, makes the GL initialization much complex, but would work today on already released drivers without adding any function indirect call on every other driver.

Note: the same mechanism may be usable if `OES_framebuffer_object` is available, maybe in the future we want to extend this kind of indirect function call to support more GL standards.

DolceTriade:
In that case, maybe the engine shim might be better. @slipher what do you think?

slipher:
The approach here seems OK to me.

The function pointers can supposedly be dependent on the context, so make sure they are reloaded if the context is recreated.

illwieckz:
> The function pointers can supposedly be dependent on the context, so make sure they are reloaded if the context is recreated.

I guess `GLimp_InitExtensions` is called anytime a context is recreated? I set the function pointers there.

illwieckz:
`GLimp_CreateContext()` is called by `GLimp_ValidateBestContext()` which is called by `GLimp_SetMode()` which is called by `GLimp_StartDriverAndSetMode()` which is then called by `GLimp_Init()`.

The fbo function pointers are set in `GLimp_InitExtensions()` which is called by `GLimp_Init()` _after_ `GLimp_StartDriverAndSetMode()` and then `GLimp_CreateContext()` is called.

`GLimp_CreateContext()`  is called and only called when `GLimp_StartDriverAndSetMode()` succeeds.

So I assume the function pointers are always set after (re)creating a context.

I also fixed the compilation on 64-bit MSVC, errors like that [[1]](https://ci.appveyor.com/project/DolceTriade/daemon/builds/45380218/job/rlyx52vu8f4lt8eu), [[2]](https://ci.appveyor.com/project/DolceTriade/daemon/builds/45380216/job/k9d3b9uhwhpd2d1u):

```
C:\projects\daemon\src\engine\sys\sdl_glimp.cpp(1839,51):
 error C2440: '=': cannot convert from 'PFNGLBINDFRAMEBUFFERPROC'
 to 'void (__cdecl *)(GLenum,GLuint)' [C:\projects\daemon\build\client-objects.vcxproj]
C:\projects\daemon\src\engine\sys\sdl_glimp.cpp(1839,34): message :
 This conversion requires a reinterpret_cast, a C-style cast
 or function-style cast [C:\projects\daemon\build\client-objects.vcxproj]
```

illwieckz:
I turned the code to set function pointers for this or that extension to inline functions named `glFboSetArb()` and `glFboSetArb()` this way the setting code lives the closest possible to the structure declaration.

I also renamed the struct type from `glShim_t` to `glFboShim_t` with an explicit reference to framebuffer objects in the name, same with the object name. I think that if in the future we want to implement a similar shim for something else we would prefer declare another structure instead of putting everything in the same jumbo structure. So this one is dedicated for fbo tweaks.

DolceTriade:
The code lgtm

PR #729, opened 2022-11-05
Silence -Wmaybe-uninitialized warning

necessarily-equal:
The warning isn't warranted, the code seems correct to me.


slipher:
Yeah this warning is usually bogus. We already turned it off completely (`-Wno-maybe-uninitialized`) in the Unvanquished repository; let's do that here also?

DolceTriade:
agree with @slipher. Let's just disable in the CMake file and not using pragma

PR #731, opened 2022-11-07
cmake: detect target CPU architecture using a C compiler and enable some ARM stuff

illwieckz:
This implements a way to detect the target CPU architecture using a C compiler.

The trick is to compile some C file with the target C compiler to select the CPU architecture based on defines set by the target C compiler. To avoid the problem of actually running foreign code on the host when cross-compiling, the C file only contains preprocessor code and prints the CPU architecture using `#error` preprocessor commands.

For convenience it relies on the [`qprocessordetection.h`](https://github.com/qt/qtbase/blob/dev/src/corelib/global/qprocessordetection.h) header from Qt. This way all the knowledge is outsourced to a project that knows better than us.

Both the C source file and the Qt header don't contribute to the engine and game binaries so there is no license inheritance concern.

For demonstration purpose I have added (but left commented out) CPU architecture detection code for other architectures with growing interest that may or may not be supported in the future: ARM, Power, and RISCV. I really hope ARM engine will be a thing in the future.

_Edit:_ I also enabled `wasm` architecture detection because we already have `cmake/FindWasmtime.cmake` and other Wasm related bits in the tree.

_Edit:_ I also enabled some `arm` stuff as we can already build and run a working `daemonded` + exe sgame with those patches.


slipher:
Huh? Why? What problem does this solve?

illwieckz:
We currently assume that every architecture with `8`-sized pointers is x86_64 and that everything else is x86, which is very wrong… I really hope the move to Wasm will enable us to port the engine to other architectures, so that's a first step in that direction.

Also I noticed the naclsdk is full of ARM stuff (though it [seems to be 32-bit](https://developer.chrome.com/docs/native-client/reference/sandbox_internals/arm-32-bit-sandbox/)) so maybe we can start investigate arm?

slipher:
Architecture info can already be found within CMake. You don't need to run the C compiler to figure anything out.

illwieckz:
Every documentation and public discussions I have read said it is unreliable using CMake features, for example [“on Windows, the host may be AMD64 even when using a MSVC cl compiler with a 32-bit target”](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html) (CMake official documentation).

slipher:
> Every documentation and public discussions I have read said it is unreliable using CMake features, for example [“In many cases, this will correspond to the target architecture for the build, but this is not guaranteed. (E.g. on Windows, the host may be AMD64 even when using a MSVC cl compiler with a 32-bit target.)”](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html) (CMake official documentation).

That doesn't say that it is unreliable. It's just a comment that due to Windows's ability to run binaries of multiple bitnesses, it is possible that CMAKE_SYSTEM_PROCESSOR != CMAKE_HOST_SYSTEM_PROCESSOR even for a non-cross-compile build.

slipher:
We would have to extend the NACL_BUILD_ARCH thing with additional cases every time we added a new supported architecture, *regardless* of whether we used CMake's provided system information or a home-brewed solution, since the NaCl client code has its own specific names for each architecture.



> Note: `CMAKE_SYSTEM_PROCESSOR` is empty.

Under what conditions? Are you talking about with the NaCl toolchain? In that case it's sort of logical for it to be empty, since the compiler is building for the architecture-agnostic PNaCl target. If we want it to have a value it's the responsibility of our toolchain file. (The NACL_BUILD_ARCH stuff would not be affected by this since that's used only by the engine code.)

necessarily-equal:
I'm a bit ambivalent about this. On one side, sure it's an improvement. On the other side that's the kind of thing I would expect the build system to get right so we would never have to care about this.

For the record, with meson getting the architecture you are compiling to is just `host_machine.system()` and knowing the size of things is just  `compiler = meson.get_compiler('c')`, then `compiler.sizeof('void *')`.

illwieckz:
@slipher

> Under what conditions? Are you talking about with the NaCl toolchain?

Yes.

> since the NaCl client code has its own specific names for each architecture.

Windows also has its own specific names for each architectures.

`CMAKE_SYSTEM_PROCESSOR` is dependent on the system CMake is running on, for example the same architecture would be named `AMD64` on Windows but `x86_64` on Linux. NaCl doesn't look much more specific.

@necessarily-equal 
> For the record, with meson getting the architecture you are compiling to is just `host_machine.system()`

You confuse `CMAKE_SYSTEM_HOST_PROCESSOR` (basically `uname -m` on Linux, the architecture running the compiler) with `CMAKE_SYSTEM_PROCESSOR` (the architecture the code is compiled for). But if Meson has an unified naming on all systems that would be an improvement over CMake.

> and knowing the size of things is just `compiler = meson.get_compiler('c')`, then `compiler.sizeof('void *')`.

This is `CMAKE_SIZEOF_VOID_P`

illwieckz:
Anyway, the branch is currently broken with NaCl. As I said, it only worked before because of a test always being true, now that I fixed the test it doesn't build. I can workaround that by always setting the NaCl architecture to x86 since we only support that anyway, but that would still be hiding the fact the way we use the NaCl toolchain relies on luck or on the assumption the game to not be cross-platform.

Also I face something is weird in a way it looks like I have to set an architecture to build a pexe (before translating it to arch-specific nexe).

necessarily-equal:
@illwieckz not that it matters much, but I think I didn't do a mistake, meson just calls "build" the computer that builds, and "host" the one that runs it. 

> Since cross compiling is more complicated than native building, let's first go over some nomenclature. The three most important definitions are traditionally called build, host and target. This is confusing because those terms are used for quite many different things. To simplify the issue, we are going to call these the build machine, host machine and target machine. Their definitions are the following:
> * build machine is the computer that is doing the actual compiling.
> * host machine is the machine on which the compiled binary will run.
> * target machine is the machine on which the compiled binary's output will run, only meaningful if the program produces machine-specific output.

That's a convention I've seen elsewhere, for example in the [nixpkgs documentation](https://ryantm.github.io/nixpkgs/stdenv/cross-compilation/#ssec-cross-platform-parameters) « We distinguish between 3 types of platforms when building a derivation: build, host, and target. In summary, build is the platform on which a package is being built, host is the platform on which it will run. […] ». It seems gcc [uses that nomenclature too](https://stackoverflow.com/questions/7088576/whats-the-difference-between-host-and-target-in-gcc) and apparently [so does autoconf](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Hosts-and-Cross_002dCompilation.html).

illwieckz:
> @illwieckz not that it matters much, but I think I didn't do a mistake, meson just calls "build" the computer that builds, and "host" the one that runs it.

illwieckz:
I have done extensive tests and written some message calls to print stuff in CMake and… this code in `daemon/cmake/DaemonNacl.cmake`:

```
if( CMAKE_SIZEOF_VOID_P EQUAL 8 )
  add_definitions( -DNACL_BUILD_SUBARCH=64 )
else()
endif()
```

This code always set `NACL_BUILD_SUBARCH` to `32` because `CMAKE_SIZEOF_VOID_P` is empty when running the pnacl toolchain. That define is required or the pexe doesn't build, but setting it to `32` doesn't prevent to convert to both `x86.nexe` and `x86_64.nexe`, so maybe that's the same for `NACL_BUILD_ARCH` and setting it to `x86` would not prevent to convert to `arm`.

I'm starting to think those defines are meant for native dll builds but for some reasons nacl is buggy and expect them to never be empty in all cases, including pexe builds.

illwieckz:
Yes, `NACL_BUILD` and `NACL_BUILD_SUBARCH` really looks like useless to build `pexe` (as expected), but they are required to be set to something supported by `dll` and that's probably a NaCl design fault. I can actually build a pexe by setting `-DNACL_BUILD_ARCH=arm` and get `x86.nexe` and `x86_64.nexe` being produced and playable. That's stupid!

slipher:
> `CMAKE_SYSTEM_PROCESSOR` is dependent on the system CMake is running on, for example the same architecture would be named `AMD64` on Windows but `x86_64` on Linux.

OK, finally I can somewhat understand the motivation for this PR. Maybe it's worth considering...



> This code always set `NACL_BUILD_SUBARCH` to `32` because `CMAKE_SIZEOF_VOID_P` is empty when running the pnacl toolchain.

Once again, this definition is only relevant for building the engine. It's for the code that communicates over a socket with the sandbox. It should properly be a public compile definition of the srclibs-nacl-native target.

illwieckz:
> Once again, this definition is only relevant for building the engine.

It makes sense, but `pexe` cannot compile if `NACL_BUILD_SUBARCH` define was not set, and by pure luck this define was set because `CMAKE_SIZEOF_VOID_P` was empty (so, not equal to `8`).

It looks like this whole `daemon/cmake/DaemonNacl.cmake` file only worked with luck.

Actually we can use the define macro trick to detect the nacl compiler by

illwieckz:
```
[100%] Stripping  (arm)
[100%] Built target sgame-arm
[100%] Stripping  (arm)
[100%] Built target cgame-arm
```

```
$ file *-arm.nexe
cgame-arm.nexe: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, BuildID[sha1]=2a199c964a5350fca807ea5fa6d563f88cad8e44, with debug_info, not stripped
sgame-arm.nexe: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, BuildID[sha1]=cfdc0a612f35721e924cd1dab9d9cd82c321f5fa, with debug_info, not stripped
```

Too bad it's only 32-bit arm…

slipher:
You are right, there are some poorly structured NaCl headers that include "trusted" headers from "untrusted" ones. The trusted headers assert that the architecture has been defined, although it should not be used for anything in the untrusted (gamelogic) code. The defines presumably just have to be set to anything to get it to compile.

illwieckz:
With this branch I can build arm nexe on an amd64 host (but not on an arm host).

With some small patches above this branch I can also build then engine on an arm host.

I've tested with a Raspberry Pi 4B.

Here is running the main menu with dll game:

[arm raspberry pi 4](https://dl.illwieckz.net/b/unvanquished/wip/arm-raspberry-pi-4/20221108-012604-000.unvanquished-rpi4.png)

Unfortunately I get a mysterious error while loading a map:

[arm raspberry pi 4](https://dl.illwieckz.net/b/unvanquished/wip/arm-raspberry-pi-4/20221108-013047-000.unvanquished-rpi4.png)

Loading a nexe doesn't work yet:

[arm raspberry pi 4](https://dl.illwieckz.net/b/unvanquished/wip/arm-raspberry-pi-4/20221108-013633-000.unvanquished-rpi4.png)

eli-schwartz:
> `CMAKE_SYSTEM_PROCESSOR` is dependent on the system CMake is running on, for example the same architecture would be named `AMD64` on Windows but `x86_64` on Linux. NaCl doesn't look much more specific.

> But if Meson has an unified naming on all systems that would be an improvement over CMake.

Yes. Meson does canonicalization of `cpu_family()` naming.

64-bit processors of the Intel x86 family, will always be canonicalized to "x86_64".

See https://mesonbuild.com/Reference-tables.html#cpu-families

illwieckz:
I enabled some `arm` stuff as we can already build and run a working `daemonded` + exe sgame with those patches.

illwieckz:
I also made the architecture header file reusable for architecture-specific definitions in C++ code, so both CMake and C++ code use the exact same mechanism to detect the architecture.

illwieckz:
Here is how I build and run with this branch a 32-bit arm native `daemonded` and exe sgame binary on Raspberry Pi OS (Debian arch64) on a Raspberry Pi 4B.

I modified my `/etc/apt/source.list` this way:

```
deb [arch=arm64,armhf] http://deb.debian.org/debian bullseye main contrib non-free
deb [arch=arm64,armhf] http://security.debian.org/debian-security bullseye-security main contrib non-free
deb [arch=arm64,armhf] http://deb.debian.org/debian bullseye-updates main contrib non-free
```

Then I did:

```sh
apt-get install multiarch

dpkg --add-architecture armhf

apt-get update

# Note: not all of that is required to just build the server, some deps are to build the client itself, but whatever.

apt-get install libstdc++-10-dev:armhf gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf binutils-arm-linux-gnueabihf libcurl4-gnutls-dev:armhf libglew-dev:armhf libgmp-dev:armhf nettle-dev:armhf zlib1g-dev:armhf libncursesw5-dev:armhf libsdl2-dev:armhf libopenal-dev:armhf libjpeg-dev:armhf libpng-dev:armhf libwebp-dev:armhf libogg-dev:armhf libvorbis-dev:armhf libopusfile-dev:armhf libfreetype6-dev:armhf liblua5.3-dev:armhf python3-yaml python3-jinja2
```

I build the engine and the game this way:

```sh
cd daemon/external_deps
./build.sh linux-arm-default naclsdk naclports install

cd -

mkdir -p build
cd build
cmake .. \
 -D'BUILD_CLIENT'='OFF' -D'BUILD_SERVER'='ON' \
 -D'BUILD_CGAME'='OFF' -D'BUILD_SGAME='ON' \
 -D'BUILD_GAME_NATIVE_EXE'='ON'
make -j$(nproc)
```

Then I run:

```sh
./daemonded -pakpath ~/.local/share/unvanquished/base/pkg/ -set vm.sgame.type 2 +devmap plat23
```

Note: the cgame nexe binary used by clients must exactly match the revision of the server exe, so you may have to build some cgame nexe and ship them in a dpk. The cgame nexe can only be built on an i686/amd64 machine because the SDK itself isn't compiled for arm.

cmake/DaemonGame.cmake
```diff
@@ -84,12 +85,13 @@ function(GAMEMODULE)
...
                     -DDEPS_DIR=${DEPS_DIR}
                     -DBUILD_GAME_NACL_NEXE=${BUILD_GAME_NACL_NEXE}
-                    -DBUILD_GAME_NACL=1
-                    -DBUILD_GAME_NATIVE_DLL=0
-                    -DBUILD_GAME_NATIVE_EXE=0
-                    -DBUILD_CLIENT=0
-                    -DBUILD_TTY_CLIENT=0
-                    -DBUILD_SERVER=0
+                    -DBUILD_GAME_NACL_NEXE_ARM=${BUILD_GAME_NACL_NEXE_ARM}
+                    -DBUILD_GAME_NACL=ON
```
slipher, line 89: Please avoid additional random cosmetic changes in this mega commit.

external_deps/build.sh
```diff
@@ -745,6 +759,17 @@ setup_linux-amd64-default() {
 	common_setup
 }
 
+# Set up environment for arm Linux
+setup_linux-arm-default() {
```
slipher, line 763: Is it ARM 32 or 64?

cmake/DaemonArchitecture.cmake
```diff
@@ -0,0 +1,75 @@
...
+
+if (ARCH STREQUAL "i686")
+	set(ARCH_STRING "x86")
+elseif(ARCH STREQUAL "amd64")
+	set(ARCH_STRING "x86_64")
+else()
+	set(ARCH_STRING "${ARCH}")
+endif()
+
+# Quotes are part of the define.
```
slipher, line 74: It would be good to actually do this. Like `-DARCH_STRING="\"${ARCH_STRING}\""` I think. Then you won't have to add STRING macros everywhere it is used.

src/common/qprocessordetection.h
```diff
@@ -0,0 +1,390 @@
+// Copyright (C) 2016 The Qt Company Ltd.
```
slipher, line 1: I liked the previous layout more where this stuff is in a `tools` directory. And where the defines come through CMake rather than direct inclusion. I don't want to include all these Q_xxx macros which might even conflict with some Quake stuff.

cmake/DaemonNacl.cmake
```diff
@@ -25,21 +25,35 @@
...
-if( APPLE )
-  add_definitions( -DNACL_WINDOWS=0 -DNACL_LINUX=0 -DNACL_ANDROID=0 -DNACL_OSX=1 )
-elseif( LINUX )
-  add_definitions( -DNACL_WINDOWS=0 -DNACL_LINUX=1 -DNACL_ANDROID=0 -DNACL_OSX=0 )
-elseif( WIN32 )
-  add_definitions( -DNACL_WINDOWS=1 -DNACL_LINUX=0 -DNACL_ANDROID=0 -DNACL_OSX=0 )
-endif()
-if( ARCH STREQUAL "x86" OR ARCH STREQUAL "x86_64" )
+
+if( ARCH STREQUAL "pnacl" )
```
slipher, line 29: Usually we write `if (NACL)` for this condition

cmake/DaemonArchitecture.cmake
```diff
@@ -0,0 +1,75 @@
...
+
+# When adding a new architecture, look at all the places ARCH is used
+
+if(DAEMON_DIR)
+	set(TARGET_DETECT_SOURCE_FILE "${DAEMON_DIR}/src/common/Architecture.cpp")
+else()
+	set(TARGET_DETECT_SOURCE_FILE "${CMAKE_CURRENT_SOURCE_DIR}/src/common/Architecture.cpp")
+endif()
+
+try_run(
```
slipher, line 39: `try_compile` seems better

cmake/DaemonArchitecture.cmake
```diff
@@ -0,0 +1,75 @@
...
+	"${TARGET_DETECT_SOURCE_FILE}"
+	CMAKE_FLAGS CMAKE_OSX_ARCHITECTURES=${CMAKE_OSX_ARCHITECTURES}
+	COMPILE_DEFINITIONS -DPREPROCESSOR_ARCH_DETECTION=1
+	COMPILE_OUTPUT_VARIABLE ARCH
+)
+
+string(REGEX MATCH "DAEMON_ARCH ([a-zA-Z0-9_]+)" ARCH "${ARCH}")
+string(REPLACE "DAEMON_ARCH " "" ARCH "${ARCH}")
+
+if (NOT ARCH)
```
slipher, line 52: I think this condition is not going to work because you recycled the ARCH variable for both the compiler output and the detected architecture

CMakeLists.txt
```diff
@@ -92,12 +92,14 @@ if(NOT DAEMON_EXTERNAL_APP)
 
     if (CMAKE_SYSTEM_NAME STREQUAL CMAKE_HOST_SYSTEM_NAME)
         # can be loaded by daemon with vm.[sc]game.type 0 or 1
-	option(BUILD_GAME_NACL "Build the NaCl gamelogic modules, required to host mods." OFF)
-	option(BUILD_GAME_NACL_NEXE "Build the NaCl \"nexe\" files from the \"pexe\" files." ON)
-	mark_as_advanced(BUILD_GAME_NACL_NEXE)
+        option(BUILD_GAME_NACL "Build the NaCl gamelogic modules, required to host mods." OFF)
```
slipher, line 95: Why are these lines diffs? It looks exactly the same.

illwieckz: there was just random tab/space mix that were annoying me

cmake/DaemonArchitecture.cmake
illwieckz: I tried that, the contradictory “Quotes are part of the define.” comment is a relic of that (I should have removed this comment). This doesn't work, sometime it ends as `"<arch>"`, sometime it ends as  `""<arch>"", and both variants ends in the same PCH. Here it's written that quote characters aren't reliable anyway:

- https://cmake.org/cmake/help/latest/prop_dir/COMPILE_DEFINITIONS.html

That'd why I went back to `STRING` macros.

external_deps/build.sh
illwieckz: 32 bit, NaCl only supports 64-bit x86, 32-bit x86, 32-bit arm, 32-bit mips, no more.

src/common/qprocessordetection.h
illwieckz: Yeah, I tried to reuse the header in code itself but I went back from this. I haven't moved them back to `tools/` yet but none of the two added headers are actually included in compiled code.

external_deps/build.sh
slipher: Shouldn't there be some indication of that in the arch string though since there can be more variants in the future?

CMakeLists.txt
slipher: The request to not include superfluous changes in the big commit applies then

cmake/DaemonArchitecture.cmake
slipher: OK. You should use STRING instead of XSTRING; XSTRING is for when the argument is a macro and you want to expand it first before stringizing the expanded contents

external_deps/build.sh
illwieckz: I expect 64bit arm to simply be named `arm64`, that's the way Debian names it. I had named 32-bit just `arm` because we have no reason to differentiates `arm` and `armhf`, as I assume arm-without-hardware-float is completely excluded from the list of platform we may add support for in the future…

slipher: Actually I think calling it `armhf` would be worthwhile just so you can easily look up the name and understand what it is, even having no intention to port to `armel` or whatever.

cmake/DaemonNacl.cmake
illwieckz: Well, the rest of the code compares with `ARCH`, but whatever, both work.

cmake/DaemonGame.cmake
illwieckz: Moved to dedicated bikeshedding commit.

cmake/DaemonArchitecture.cmake
illwieckz: Done.

CMakeLists.txt
illwieckz: Moved to dedicated bikeshedding commit.

external_deps/build.sh
illwieckz: Done.

src/common/qprocessordetection.h
illwieckz: Moved back to `tools/`.

illwieckz:
I moved the patch to make `external_deps/build.sh` builds armhf deps archive to a dedicated commit, those changes don't need to be stored in the cmake patch as long as the deps patch is applied first to make sure archive can be generated if checking-out the cmake patch.

I've moved some bikeshedding (spacing, wording) to dedicated commits, added more bikeshedding.

Especially, the [sc]game build log is better (there was even an undefined variable being used),

Before:

```
[ 97%] Linking CXX executable ../cgame.pexe
[ 97%] Linking CXX executable ../sgame.pexe
[ 97%] Built target sgame-nacl
[ 97%] Translating  (arm)
[ 97%] Translating  (x86-64)
[ 97%] Translating  (i686)
[ 97%] Stripping  (i686)
[ 97%] Built target sgame-x86
[ 98%] Stripping  (x86-64)
[ 98%] Built target sgame-x86_64
[ 99%] Stripping  (arm)
[ 99%] Built target sgame-armhf
[100%] Built target cgame-nacl
[100%] Translating  (x86-64)
[100%] Translating  (i686)
[100%] Translating  (arm)
[100%] Stripping  (i686)
[100%] Built target cgame-x86
[100%] Stripping  (x86-64)
[100%] Built target cgame-x86_64
[100%] Stripping  (arm)
[100%] Built target cgame-armhf
[ 99%] Forcing build step for 'nacl-vms'
[100%] No install step for 'nacl-vms'
[100%] Completed 'nacl-vms'
[100%] Built target nacl-vms
```

After:

```
[ 87%] Linking CXX executable ../cgame.pexe
[ 97%] Linking CXX executable ../sgame.pexe
[ 97%] Built target sgame-nacl
[ 97%] Translating sgame (armhf)
[ 97%] Translating sgame (x86)
[ 97%] Translating sgame (x86_64)
[ 97%] Stripping sgame (x86)
[ 97%] Built target sgame-x86
[ 98%] Stripping sgame (x86_64)
[ 98%] Built target sgame-x86_64
[ 99%] Built target cgame-nacl
[ 99%] Translating cgame (armhf)
[ 99%] Translating cgame (x86_64)
[ 99%] Translating cgame (x86)
[100%] Stripping sgame (armhf)
[100%] Built target sgame-armhf
[100%] Stripping cgame (x86)
[100%] Built target cgame-x86
[100%] Stripping cgame (x86_64)
[100%] Built target cgame-x86_64
[100%] Stripping cgame (armhf)
[100%] Built target cgame-armhf
[ 99%] Forcing build step for 'nacl-vms'
[100%] No install step for 'nacl-vms'
[100%] Completed 'nacl-vms'
[100%] Built target nacl-vms
```

cmake/DaemonArchitecture.cmake
illwieckz: I had to revert to `XSTRING` because otherwise I would get that:

```
Warn: VM module file not found: sgame-NACL_ARCH_STRING.nexe 
Warn: NaCl integrated runtime not found: irt_core-NACL_ARCH_STRING.nexe 
Loading VM module sgame-NACL_ARCH_STRING.nexe... 
Using loader args:  nacl_helper_bootstrap nacl_loader --r_debug=0xXXXXXXXXXXXXXXXX --reserved_at_zero=0xXXXXXXXXXXXXXXXX -v -B irt_core-NACL_ARCH_STRING.nexe -e -i 100:25 -- sgame-NACL_ARCH_STRING.nexe 100 
Warn: IPC: Socket closed by remote end 
```

external_deps/build.sh
```diff
@@ -654,19 +654,19 @@ common_setup() {
 		export CMAKE_GENERATOR="MSYS Makefiles"
 		CROSS= # Doing this because there is no ${CROSS}strip
 	fi
+	MSVC_SHARED=(--enable-shared --disable-static)
```
slipher, line 671: The MSVC_SHARED change reverses the values on a bunch of platforms, please undo

cmake/toolchain-pnacl.cmake
```diff
@@ -102,26 +102,26 @@ set(NACL_TRANSLATE_OPTIONS
     $<$<CONFIG:MinSizeRel>:-O2>
 )
 
-function(pnacl_translate dir module arch suffix)
+function(pnacl_translate dir module nacl_opt arch)
```
slipher, line 105: What does "opt" mean in nacl_opt?

tools/DaemonArchitecture/qprocessordetection.h
```diff
@@ -0,0 +1,390 @@
+// Copyright (C) 2016 The Qt Company Ltd.
```
slipher, line 1: In the commit message, please say which version/revision of Qt the file comes from.

src/engine/framework/System.cpp
```diff
@@ -564,7 +564,7 @@ static void Init(int argc, char** argv)
 #endif
 
 	// Print a banner and a copy of the command-line arguments
-	Log::Notice(Q3_VERSION " " PLATFORM_STRING " " ARCH_STRING " " __DATE__);
+	Log::Notice(Q3_VERSION " " PLATFORM_STRING " " STRING(ARCH_STRING) " " __DATE__);
```
slipher, line 567: I guess you need to change tho two instances of STRING(ARCH_STRING) back to XSTRING  (sorry)

slipher (commented):
LGTM modulo comments

cmake/toolchain-pnacl.cmake
illwieckz: option, this is the command line option

external_deps/build.sh
illwieckz: Oh my bad, my eyes really get tricked…

cmake/DaemonArchitecture.cmake
illwieckz: Done.

external_deps/build.sh
illwieckz: Fixed.

cmake/toolchain-pnacl.cmake
illwieckz: Renamed to `nacl_option` to make it more obvious.

src/engine/framework/System.cpp
illwieckz: Done.

illwieckz:
OK, applied the various requested fixes.

I also added some `TODO(0.54)` remaining that the next time we break compatibility we may start to use the same architecture name convention everywhere to simplify things (right now we have to make sure to not break `nexe` file names).

tools/DaemonArchitecture/qprocessordetection.h
illwieckz: Done.

illwieckz:
So, it looks ready to me.

illwieckz:
Interesting thing: while I fail to run the armhf nexe on arm, the armhf nexe seems to be correct:

```
build-linux-armhf-default_7/naclsdk/pepper_44/tools$ python2 ./sel_ldr.py ~/dev/Unvanquished/build/sgame-armhf.nexe 
DEBUG MODE ENABLED (bypass acl)
PLATFORM QUALIFICATION DISABLED BY -Q - Native Client's sandbox will be unreliable!
This program is not meant to be invoked directly, it must be invoked by the engine's VM loader.
```

The `This program is not meant to be invoked directly, it must be invoked by the engine's VM loader.` message print is actual game code from `daemon/src/shared/VMMain.cpp`.

Note: I ran it on amd64, `sel_ldr.py` calls `qemu-arm` to test it.

illwieckz:
So the option to make Unvanquished run armhf nexe is to use the `-Q` option of `nacl_loader`:

```diff
diff --git a/src/engine/framework/VirtualMachine.cpp b/src/engine/framework/VirtualMachine.cpp
index c84ae0c8..1d56aba1 100644
--- a/src/engine/framework/VirtualMachine.cpp
+++ b/src/engine/framework/VirtualMachine.cpp
@@ -256,6 +256,25 @@ std::pair<Sys::OSHandle, IPC::Socket> CreateNaClVM(std::pair<IPC::Socket, IPC::S
 	args.push_back(nacl_loader.c_str());
 	args.push_back("--r_debug=0xXXXXXXXXXXXXXXXX");
 	args.push_back("--reserved_at_zero=0xXXXXXXXXXXXXXXXX");
+
+	#if defined(DAEMON_ARCH_armhf)
+		/* This is required to run on Raspberry Pi 4,
+		otherwise nexe loading fails with this message:
+
+		  Error while loading "sgame-armhf.nexe": CPU model is not supported
+
+		From nacl_loader --help we can read:
+
+		  -Q disable platform qualification (dangerous!)
+
+		When this option is enabled, nacl_loader will print:
+
+		  PLATFORM QUALIFICATION DISABLED BY -Q - Native Client's sandbox will be unreliable!
+
+		But the nexe will load and run. */
+
+		args.push_back("-Q");
+	#endif
 #else
 	Q_UNUSED(bootstrap);
 	args.push_back(nacl_loader.c_str());
```

illwieckz:
Note to myself, this lists versions of the NaCl SDK more recent (up to 58) than the one we use (44): https://storage.googleapis.com/nativeclient-mirror/nacl/nacl_sdk/naclsdk_manifest2.json

Unfortunately they're old enough to not ship any arm64 stuff… 😥️

illwieckz:
For some unknown reasons when I build an armhf Release build of the engine I get this crash when loading a map:

```
Warn: Crashed with signal 7: Bus error
```

But if I build an armhf Debug build of the engine, it works.

So, here is a screenshot of Unvanquished with native armhf engine and armhf nexe game on Raspberry Pi 4:

[unvanquished on raspberry pi 4](https://dl.illwieckz.net/b/unvanquished/wip/arm-raspberry-pi-4/20221115-015837-000.unvanquished-rpi4.png)

illwieckz:
It is now possible to build and run a 64-bit arm64 engine with a 32-bit armhf nexe using multiarch on Linux.

@slipher I added a commit to run a 64-bit arm64 engine with a 64-bit amd64 nexe using Rosetta2 on macOS, but it isn't tested, and one would have to update `external_deps/build.sh` to build the arm64 macOS deps with amd64 nacl loader first.

illwieckz:
I added a quirk to provide armhf multiarch on arm64 without requiring the user to set it up himself. Setting up multiarch usually requires root permission, update repositories, run some configuration command, and install packages. With that quirk an arm64 user would just run the arm64 engine with armhf nexe without doing anything more than unzipping Unvanquished.

Right now it relies on a small wrapper which is currently a two-line bash script named `nacl_helper_bootstrap-armhf`, if someones wants to turn it into a native binary, feel free to rewrite it.

slipher:
Please put new commits in a new PR instead of adding stuff to an already approved one

illwieckz:
@slipher I simplified the PR to have less commit,basically up to the commit introducing the compiler-based architecture detection.

Remaining commits are in [illwieckz/arm64](https://github.com/DaemonEngine/Daemon/tree/illwieckz/arm64) to be part of dedicated PRs once this one is merged. Some like #739 and #740 are already opened but require this one to be merged first.

slipher:
LGTM

PR #732, opened 2022-11-08
Redesign Color::TokenIterator API

slipher:
Companion: https://github.com/Unvanquished/Unvanquished/pull/2304

The existing API failed to make it easy to do some common tasks, namely converting a color string to plain text and dealing with the possibility of dangling carets at the end of a string.

### DIFF
--- a/src/common/Color.cpp
+++ b/src/common/Color.cpp
@@ -86,16 +86,16 @@ int StrlenNocolor( const char *string )
 
 char *StripColors( char *string )
 {
-	std::string output;
+	char *out = string;
 
 	for ( const auto& token : Parser( string ) )
 	{
 		Str::StringView text = token.PlainText();
-		output.append( text.begin(), text.end() );
+		memmove( out, text.begin(), text.size() );
+		out += text.size();
 	}
 
-	strcpy( string, output.c_str() );
-
+	*out = '\0';
 	return string;
 }
 
@@ -119,6 +119,7 @@ void StripColors( const char *in, char *out, size_t len )
 std::string StripColors( const std::string& input )
 {
 	std::string output;
+	output.reserve( input.size() );
 
 	for ( const auto& token : Parser( input.c_str() ) )
 	{
--- a/src/common/Color.cpp
+++ b/src/common/Color.cpp
@@ -90,14 +90,8 @@ char *StripColors( char *string )
 
 	for ( const auto& token : Parser( string ) )
 	{
-		if ( token.Type() == Token::TokenType::CHARACTER )
-		{
-			output.append( token.Begin(), token.Size() );
-		}
-		else if ( token.Type() == Token::TokenType::ESCAPE )
-		{
-			output.push_back(Constants::ESCAPE);
-		}
+		Str::StringView text = token.PlainText();
+		output.append( text.begin(), text.end() );
 	}
 
 	strcpy( string, output.c_str() );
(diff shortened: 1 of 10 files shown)


src/engine/renderer/tr_decals.cpp
```diff
@@ -131,8 +131,11 @@ static bool MakeTextureMatrix( vec4_t texMat[ 2 ], vec4_t projection, decalVert_
 		}
 	}
 
-	texMat[ 0 ][ 3 ] = a->st[ 0 ] - DotProduct( pa, texMat[ 0 ] );
-	texMat[ 1 ][ 3 ] = a->st[ 1 ] - DotProduct( pa, texMat[ 1 ] );
+	for ( int k : {0, 1} )
```
DolceTriade, line 134: Wonder what the problem is and why this doesn't seem to have anything to do with the commit message?

src/common/Color.h
```diff
@@ -398,27 +398,27 @@ class Token
 	{}
 
 	/*
-	 * Pointer to the first character of this token in the input sequence
+	 * RawToken defines the extent of the token as it was written in the original string. It
```
DolceTriade, line 401: Perhaps you can document when to use each version of the API? RawToken vs NormalizedToken vs PlainText

src/engine/client/cl_console.cpp
```diff
@@ -351,8 +351,8 @@ bool Con_CheckResize()
 					}
 					if ( len > consoleState.textWidthInChars )
 					{
-						consoleState.lines.emplace_back( begin, token.Begin() );
-						begin = token.Begin();
+						consoleState.lines.emplace_back( begin, token.RawToken().begin() );
```
DolceTriade, line 354: I'm surprised that this doesn't want normalizedtoken

src/engine/client/cl_console.cpp
```diff
@@ -500,22 +500,23 @@ bool CL_InternalConsolePrint( const char *text )
 	{
 		if ( token.Type() == Color::Token::TokenType::COLOR )
 		{
-			consoleState.lines.back().append( token.Begin(), token.Size() );
+			Str::StringView colorToken = token.NormalizedToken();
```
DolceTriade, line 503: I think it's confusing when you have to swap between these APIs in the same function. Can we comment why we do this?

src/common/Color.h
slipher: Added more doc

src/engine/client/cl_console.cpp
DolceTriade: is this something to be fixed in grep rather than worked around in your token iterator API?

src/common/Color.cpp
```diff
@@ -141,14 +122,8 @@ std::string StripColors( const std::string& input )
...
 	{
-		if ( token.Type() == Token::TokenType::CHARACTER )
-		{
-			output.append( token.Begin(), token.Size() );
-		}
-		else if ( token.Type() == Token::TokenType::ESCAPE )
-		{
-			output.push_back(Constants::ESCAPE);
-		}
+		Str::StringView text = token.PlainText();
```
DolceTriade, line 125: Given that we are touching this code, can this be an opportunity to do:

```
std::string StripColors( Str::StringRef in ) {
  std::string output;
  output.resize(in.size()); // output basically guaranteed to be <= input since we will never add any chars
  StripColors(in.c_str(), output.c_str(), output.size());
  return output;
}

File: src/common/Color.h [hunk 6]
```diff
@@ -427,6 +430,20 @@ class Token
 		return type;
 	}
 
+	/*
+	 * Token as plain text, for contexts where color codes aren't interpreted.
+	 * A color token gives the empty string.
```
DolceTriade, line 435: For non-empty strings, this will always point inside the input string?

DolceTriade (approved):
lgtm % some doc nits

src/common/Color.cpp
slipher: You also would need a `strlen` there to reduce the output string to its actual size so I think that would be longer/uglier than what we have now.

I added a commit to pre-allocate the string buffer if that's what you're going for.

DolceTriade (approved):

PR #740, opened 2022-11-17
VirtualMachine: disable NaCl platform qualification to run armhf nexe on Raspberry Pi 4

illwieckz:
#731 must be merged first.

### DIFF
--- a/src/engine/framework/VirtualMachine.cpp
+++ b/src/engine/framework/VirtualMachine.cpp
@@ -256,6 +256,25 @@ std::pair<Sys::OSHandle, IPC::Socket> CreateNaClVM(std::pair<IPC::Socket, IPC::S
 	args.push_back(nacl_loader.c_str());
 	args.push_back("--r_debug=0xXXXXXXXXXXXXXXXX");
 	args.push_back("--reserved_at_zero=0xXXXXXXXXXXXXXXXX");
+
+	#if defined(DAEMON_ARCH_armhf)
+		/* This is required to run on Raspberry Pi 4,
+		otherwise nexe loading fails with this message:
+
+		  Error while loading "sgame-armhf.nexe": CPU model is not supported
+
+		From nacl_loader --help we can read:
+
+		  -Q disable platform qualification (dangerous!)
+
+		When this option is enabled, nacl_loader will print:
+
+		  PLATFORM QUALIFICATION DISABLED BY -Q - Native Client's sandbox will be unreliable!
+
+		But the nexe will load and run. */
+
+		args.push_back("-Q");
+	#endif
 #else
 	Q_UNUSED(bootstrap);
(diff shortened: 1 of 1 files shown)


DolceTriade:
lgtm

src/engine/framework/VirtualMachine.cpp
```diff
@@ -256,6 +256,25 @@ std::pair<Sys::OSHandle, IPC::Socket> CreateNaClVM(std::pair<IPC::Socket, IPC::S
...
+
+		  -Q disable platform qualification (dangerous!)
+
+		When this option is enabled, nacl_loader will print:
+
+		  PLATFORM QUALIFICATION DISABLED BY -Q - Native Client's sandbox will be unreliable!
+
+		But the nexe will load and run. */
+
+		args.push_back("-Q");
```
necessarily-equal, line 276: According to https://github.com/mseaborn/nacl-wiki/blob/master/PlatformQualification.wiki#architecture-specific-checks-arm, here's what this checks:

> ## Architecture-specific checks: ARM
> ### ARMv7-A or later
> 
> We require a minimum architecture of ARMv7-A.
> 
> * Rationale:* ARMv7 adds instructions that are important for PNaCl (64-bit load/store exclusive), clearly defines the CPUID mechanism, and includes the operations from ARMv6T2 that we require for correctness.
> * May change:* across a reboot. Unlikely to change under virtualization.
> * Method:* we use an assembly sequence, suggested by ARM Ltd, to verify architecture revision.
> 
> ### DEP present and working
> 
> Data Execution Prevention (DEP) must be enabled and functional.
> 
> * Rationale:* The ARM sandbox allows the code-data boundary to fall at any page. Thus jumps to non-validated data are trivial. We require DEP to prevent this. Even on processors that support XN (all supported ARM processors), this can be disabled by the operating system.
> * May change:* across a reboot on real hardware. Under virtualization this can change even while the process runs.
> * Method:* we generate functions in heap and stack memory and execute them. We verify that we take the expected exception.
> 
> ## OS-specific checks: Linux
>
> ### SysV SHM
> 
> SysV SHM must be supported and pass certain feature tests, including interacting predictably with mmap (see `src/trusted/platform_qualify/linux/sysv_shm_and_mmap.c`).
> 
> * Rationale:* _unclear_
> * May change:* with kernel upgrades. (Note that with Ksplice this does not imply a reboot necessarily, though Ksplice currently isn't capable of this sort of change.)
> * Method:* See source implementation.

I don't know what check fails, but I think it's "not that bad" to disable this for now.

LGTM

illwieckz: Yes, the RPI4 uses armv8-a and I also tested the fact NX works, so I don't know why the qualification fails.

Also even the naclsdk disables qualification when running nexe over emulated qualified arm cpu on qemu in their scripts…

PR #741, opened 2022-11-17
sdl_glimp,tr_fbo: also require EXT_framebuffer_blit if EXT_framebuffer_object, catch EXT errors

illwieckz:
Follow-up of #727. This also the proof our engine can perfectly works with `EXT_framebuffer_object` as it was written against it, then EXT_fbo functions were replaced with ARB_fbo ones.

Before `EXT_framebuffer_object` was replaced by `ARB_framebuffer_object`
in c122d65c626f0b45507f044c4edae73af3709508, `EXT_framebuffer_blit`
was also required (now part of `ARB_framebuffer_object`) and two extra
errors specific to `EXT_framebuffer_object` were caught.
    
This enforces the requirement of `EXT_framebuffer_blit` if `EXT_framebuffer_object`
is available but not `ARB_framebuffer_object`, and restore catching for those
`EXT_framebuffer_object` errors.

An extra commit do some bikeshedding on comment wording.

### DIFF
--- a/src/engine/renderer/tr_fbo.cpp
+++ b/src/engine/renderer/tr_fbo.cpp
@@ -67,6 +67,18 @@ bool R_CheckFBO( const FBO_t *fbo )
 			Log::Warn("R_CheckFBO: (%s) Framebuffer incomplete, missing read buffer", fbo->name );
 			break;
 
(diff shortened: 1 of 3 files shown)


necessarily-equal (approved):
LGTM, not very knowledgeable on the subject

PR #743, opened 2022-11-17
Build linux arm64 engine and make it works with armhf nexe with builtin multiarch support

illwieckz:
Build linux arm64 engine and make it works with armhf nexe with builtin multiarch support

The 32-bit arm multiarch libs are provided by the naclsdk, formerly meant to test arm nexe on amd64 using a provided qemu. We don't need qemu on arm64 as the arm64 cpu can run both 64-bit and 32-bit executables at the same time. Since engine and game are two separate executables talking to each others using an IPC, they don't have to share the same bitness.

By luck, the `nacl_helper_bootstrap` binary is fully static and can run as 32-bit without any 32-bit loader. Unfortunately the `nacl_loader` binary requires a loader. More unfortunately, the `nacl_helper_bootstrap` binary is very simple and only properly rewrites the `nacl_loader` arguments if they're arguments `0` and arguments `1`, so we cannot do `nacl_helper_bootstrap ld-linux-armhf.so.3 nacl_loader <arguments>`.

Fortunately the `nacl_loader` binary can be dirtily patched to look for the ld loader with a relative path. Fortunately, the `nacl_helper_bootstrap` can still run with the `LD_LIBRARY_PATH` environment variable for `nacl_loader` to avoid the need to use `--library-path` option of the ld loader. So I edited the `external_deps/build.sh` script to dirtily patch `nacl_loader` to look for the ld loader with a relative path, then wrote a wrapper named `nacl_helper_bootstrap-armhf` that changes directory to the path where all deps are stored (basically, engine libpath), set `LD_LIBRARY_PATH` to the relative path to multiarch armhf libraries, and runs `nacl_helper_bootstrap` with `nacl_loader` the way the engine expects it. Since the directory is changed before running the nacl binaries and that the multiarch loader and libraries are using relative paths to be loaded, it works.

Also the `libs/nacl/native_client/src/include/portability.h` file was edited to not have to detect the architecture to detect the bitness, the bitness is provided as a define by cmake. That `portability.h` file is now architecture agnostic, there would be no need to edit it again if we have to enable another architecture.


illwieckz:
I rewrote the wrapper in C.

necessarily-equal (approved):
LGTM, haven't tested

tools/nacl_helper_bootstrap-armhf/nacl_helper_bootstrap-armhf.c
```diff
@@ -0,0 +1,67 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h> 
+
+extern char **environ;
+
+int main(int argc, char *argv[]) {
+	char *wrapper = "nacl_helper_bootstrap-armhf";
+	char *directory = argv[0];
+	directory[strlen(directory) - strlen(wrapper)] = '\0';
```
necessarily-equal, line 42: This has no consequence AFAIK, but you are modifying argv here :eyes:

illwieckz: Yes… it's a bit tricky, I extract the dirpath from the fullpath by writing a `\0` in place of the file name… 👀️

Also the code is not C++ compliant (that's why I made it a C code 😜️): I affect `const char*` to `char*`. when setting `wrapper`, `helper` and doing `putenv`. 🤷‍♀️️ I'm not sure we have to over-engineer this tiny code. 😏️

necessarily-equal: huh oh. Yeah, so how about doing `char *directory = strdup(argv[0]);` and make argv a `const char*` instead? you don't even need to care about freeing memory because the process exits afterwards.

necessarily-equal: Oh by the way, if we are going to use the libc, you can also do that:
```C
#include <libgen.h>
[…]
char *directory = dirname(strdup(argv[0]);
```

tools/nacl_helper_bootstrap-armhf/nacl_helper_bootstrap-armhf.c
```diff
@@ -0,0 +1,67 @@
...
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+===========================================================================
+*/
+
+#include <errno.h>
```
slipher, line 31: What's the point of this program?

CMakeLists.txt
```diff
@@ -924,6 +924,15 @@ if (BUILD_CLIENT OR BUILD_SERVER OR BUILD_TTY_CLIENT)
     )
 
     if (ARCH STREQUAL "arm64")
```
slipher, line 926: This block is missing a os==Linux check

external_deps/build.sh
```diff
@@ -563,6 +563,14 @@ build_naclsdk() {
 		chmod 755 "${PREFIX}/nacl_helper_bootstrap" "${PREFIX}/sel_ldr"
 		;;
 	esac
+	case "${PLATFORM}" in
+	linux-arm64-*)
+		mkdir -p "${PREFIX}/lib-armhf"
+		cp -a pepper_*"/tools/lib/arm_trusted/lib/." "${PREFIX}/lib-armhf/."
+		mv "${PREFIX}/lib-armhf/ld-linux-armhf.so.3" "${PREFIX}/lib-armhf/ld-linux-armhf"
```
slipher, line 570: Why does NaCl ship such a thing as `ld-linux-armhf`?

slipher:
I deleted some questions I asked in review, which were already answered in the OP.

tools/nacl_helper_bootstrap-armhf/nacl_helper_bootstrap-armhf.c
slipher: How about like this to avoid the hackiest assumptions about argv[0]:

```
char* slash = strrchr(argv[0], '/');
if (slash) {
    *slash = '/0';
    int err = chdir(argv[0]);
    if (err != 0) { return err; }
}
```

illwieckz: The reason why I made this as simple is that it really has only one usage in the world and the only program expected to run it (the engine) has its name hardcoded anyway. The bash equivalent script was just 2 lines…

If we're going to implement a real `dirname` maybe we should just use `dirname` like @necessarily-equal suggested…

illwieckz: I applied @necessarily-equal's suggestion and turned the whole program into a c++ one, using `strdup` when setting `char*` from `const char*` constants.

external_deps/build.sh
```diff
@@ -517,7 +517,7 @@ build_naclsdk() {
 		# TODO(0.54): Unify all arch strings using i686 and amd64 strings.
 		local DAEMON_ARCH=x86_64
 		;;
-	*-armhf-*)
+	*-armhf-*|*-arm64-*)
```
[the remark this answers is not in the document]
illwieckz: Right. Fixed.

libs/nacl/native_client/src/include/portability.h
```diff
@@ -66,29 +66,6 @@
...
-/**
- * Processor architecture detection. This code was derived from
- * Chromium's build/build_config.h.
- * For more info on what's defined, see:
- * http://msdn.microsoft.com/en-us/library/b0084kay.aspx
- * http://www.agner.org/optimize/calling_conventions.pdf
- * r with gcc, run: "echo | gcc -E -dM -"
- */
-#if defined(_M_X64) || defined(__x86_64__)
-#define NACL_HOST_WORDSIZE 64
```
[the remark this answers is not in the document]
illwieckz: Done.

CMakeLists.txt

illwieckz:
To answer a question that wasn't asked, why I do this:

```bash
sed -e 's|/lib/ld-linux-armhf.so.3|lib-armhf/ld-linux-armhf|' -i "${PREFIX}/sel_ldr"
```

Instead of relying on dedicated tools like `patchelf --set-interpreter` is that unfortunately `patchelf --set-interpreter` modifies the binary in a way not only the interpreter is modified and then the `nacl_helper_bootstrap` programs fails to validate the format of the `nacl_loader`  (`nacl_loader` is just another name for `sel_ldr`). The sed trick makes sure everything else of the binary remains unmodified, but we need to also make sure the loader path isn't larger than the original path, hence the rename of the `ld-linux-armhf` file.

tools/nacl_helper_bootstrap-armhf/nacl_helper_bootstrap-armhf.cpp
```diff
@@ -0,0 +1,66 @@
...
+	if (err != 0) {
+		return err;
+	}
+
+	char *helper = strdup("./nacl_helper_bootstrap");
+	argv[0] = helper;
+
+	err = execv(helper, argv);
+
+	if (err != 0) {
```
slipher, line 60: It's always an error if execv returns

tools/nacl_helper_bootstrap-armhf/nacl_helper_bootstrap-armhf.cpp
```diff
@@ -0,0 +1,66 @@
...
+*/
+
+#include <errno.h>
+#include <libgen.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h> 
+
+extern char **environ;
```
slipher, line 38: unused

tools/nacl_helper_bootstrap-armhf/nacl_helper_bootstrap-armhf.cpp
```diff
@@ -0,0 +1,66 @@
...
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+===========================================================================
+*/
+
+#include <errno.h>
+#include <libgen.h>
```
slipher, line 32: unused?

tools/nacl_helper_bootstrap-armhf/nacl_helper_bootstrap-armhf.cpp
illwieckz: Fixed.

tools/nacl_helper_bootstrap-armhf/nacl_helper_bootstrap-armhf.cpp
illwieckz: Fixed.

illwieckz:
It looks like all requests were fulfilled so this looks ready to me.

PR #744, opened 2022-11-18
Move entityState_t to gamelogic

necessarily-equal:
Goes along with https://github.com/Unvanquished/Unvanquished/pull/2312


src/engine/server/sv_ccmds.cpp
```diff
@@ -114,15 +114,6 @@ class MapCmd: public Cmd::StaticCmd {
 static MapCmd MapCmdRegistration("map", "starts a new map", false);
 static MapCmd DevmapCmdRegistration("devmap", "starts a new map with cheats enabled", true);
 
-void MSG_PrioritiseEntitystateFields();
-void MSG_PrioritisePlayerStateFields();
-
-static void SV_FieldInfo_f()
```
slipher, line 120: The netcode sends the deltas for each field in the table up to the last one that had any changes. The idea of this command is to run the game for a while and then print the fields in order from most to least frequently changed, to find an order which would require transmitting the fewest fields per packet.

src/engine/qcommon/msg.cpp
```diff
@@ -1137,22 +1137,48 @@ void MSG_ReadDeltaEntity( msg_t *msg, const OpaqueEntityState *from, OpaqueEntit
...
+	{
+		Log::Warn("gamelogic netcode table has incorrect size");
 		return false;
+	}
+
+	// all fields should be 32 bits to avoid any compiler packing issues
+	// the "number" field is not part of the field list
+	// if this assert fails, someone added a field to the entityState_t
+	// struct without updating the message fields
+	if (table.size() + 1 != size / 4)
```
slipher, line 1152: Isn't this wrong if there is a STATS_GROUP_FIELD?

necessarily-equal: Possible. For now I just thought "yeah playerState_t doesn't match I'll look at it later", thinking it's just one field that isn't transmitted by mistake, because of this comment:

```C++ 
// any change in playerState_t should be reflected in the table in bg_misc.cpp
// (Gordon: unless it doesn't need transmission over the network, in which case it should probably go into the new pmext struct anyway)
```

necessarily-equal: I'm not sure how to handle `STATS_GROUP_FIELD` really. It's only supported for playerState_t and it may be nice to check it only there, but I'm not sure how to check for that. Do you think I should add a boolean parameter to the `IsValid` function?

slipher: Well, it looks the functions for transmitting player states and entity states are just big copy/pastes of each other. Maybe they could just use one function, and then there would be no differences.

Alternatively you can do the comparsion of table size to entity state size outside the IsValid function.

src/engine/server/sv_snapshot.cpp
```diff
@@ -744,7 +746,7 @@ static void SV_BuildClientSnapshot( client_t *client )
 	{
 		ent = SV_GentityNum( entityNumbers.snapshotEntities[ i ] );
 		state = &svs.snapshotEntities[ svs.nextSnapshotEntities % svs.numSnapshotEntities ];
-		*state = ent->s;
+		memcpy(state, (const char*)&ent->s, entityStateSize);
```
slipher, line 749: Unnecessary cast

src/engine/qcommon/msg.cpp
necessarily-equal: Ok. I'm not going to deduplicate the netcode functions for now. For a start, I'd be happy with it working and this warning will probably just be addressed in another PR.

PR #746, opened 2022-11-19
tr_image: disable tiled renderer if float image are unavailable, fix #745

### LINKED ISSUE #745: images for the dynamic lighting are loaded even if tiled renderer is disabled, also use unsupported format on low-end hardware
A message I get on Nvidia nv30 (old Nvidia) and Broadcom VideoCore (recent Raspberry Pi):

```
Warn: floating point image '_depthtile1Render' cannot be loaded 
Warn: floating point image '_depthtile2Render' cannot be loaded 
Warn: floating point image '_dlightImage' cannot be loaded 
```

It reveals many things we can do better:

- [ ] the tiled renderer is disabled so we should not loaded them to begin with.
- [ ] we may detect that floating point image format is not supported and disable the tiled renderer.

illwieckz:
- do not load tiled renderer images if tiled renderer is disabled
- disable tiled renderer if float image are unavailable

fix #745

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -541,9 +541,9 @@ static std::string GenEngineConstants() {
 			AddDefine( str, "r_showParallelShadowSplits", 1 );
 	}
 
-	if ( r_dynamicLight->integer != 0 )
+	if ( glConfig2.dynamicLight != 0 )
 	{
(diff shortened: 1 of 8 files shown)


illwieckz:
This also avoids to run some light code when the tiled renderer is disabled.

slipher:
LGTM

PR #748, opened 2022-11-22
submodules: fetch googletest from DaemonEngine organization

illwieckz:
Fetch `googletest` repository from `DaemonEngine` organization.

By using `DaemonEngine/googletest` as upstream (even if a 1:1 copy of `google/googletest`) we are allowed to:

- set the exact same name for the default branches of all submodules and then automatically commit and build submodules based on branch names
- tag all submodules with release tags like `unvanquished/0.54.0`

The same is true for every submodule we use.


necessarily-equal (approved):
Sounds good.

Issue #646, opened 2022-07-20
Do not load deluxe maps when deluxe mapping is disabled


illwieckz:
It looks like deluxe maps are loaded and probably even uploaded even when deluxe mapping is disabled.

illwieckz:
Expected to be fixed in #749 once merged.

VReaperV:
Confirmed fixed.

PR #749, opened 2022-11-22
tr_bsp: do not load deluxemap if deluxemapping is disabled

illwieckz:
Note: _the engine already not loads light maps when light mapping is disabled: d2624fe9bb9d729629ef13dfcf5a152f342d82d1 from #358._

This makes the engine not loads deluxe maps when deluxe mapping is disabled. For example thunder has 7 deluxe maps that are 1024×1024 RGB textures. Before when deluxe mapping was disabled we still uploaded 7×1024×1024×3 bytes (25.9MB) to GPU for nothing.

---

Original message:

First purpose of this is to _not_ load deluxe maps when deluxe mapping is disabled, saving loading time from file system, uploading time to GPU and GPU memory  when using low presets. This is expected to divide by two the GPU memory used by light maps on low presets, which can be a big deal since light maps don't use DXT compressed format. Until now those deluxe maps were loaded in all cases but never used when deluxe mapping was disabled.

For example thunder has 7 light maps and 7 deluxe maps, both are 1024×1024 RGB textures. Before when deluxe mapping was disabled we uploaded 14×1024×1024×3 bytes (51.8 MB) to the GPU, now we upload 7×1024×1024×3 bytes (25.9MB)

While I was at it, I also made sure light maps are not loaded when light mapping is disabled. Though, disabling light mapping is mostly a debug mode to check for the grid lighting renderer on every surface (usually only used for models and world surfaces without light maps). Because deluxe maps are sidecars for light maps, deluxe maps are also not loaded if light mapping is disabled. It means disabling light mapping is now expected to upload 0 bytes of light maps to the GPU.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -494,7 +494,7 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 				ri.Free( ldrImage );
 			}
 
-			if (tr.worldDeluxeMapping) {
+			if (tr.worldDeluxeMapping && r_deluxeMapping->integer != 0) {
 				// load deluxemaps
 				lightmapFiles = R_LoadExternalLightmaps(mapName, &numLightmaps);
 				if (!lightmapFiles || !numLightmaps) {
@@ -542,7 +542,9 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 
 					auto image = R_FindImageFile(va("%s/%s", mapName, lightmapFiles[i]), imageParams);
 					Com_AddToGrowList(&tr.lightmaps, image);
-				} else {
+				}
+				else if (r_deluxeMapping->integer != 0)
+				{
 					imageParams_t imageParams = {};
 					imageParams.bits = IF_NOPICMIP | IF_NORMALMAP;
 					imageParams.filterType = filterType_t::FT_LINEAR;
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -446,7 +446,7 @@ R_LoadLightmaps
 static const int LIGHTMAP_SIZE = 128;
 static void R_LoadLightmaps( lump_t *l, const char *bspName )
 {
-	if ( r_vertexLighting->integer )
+	if ( r_vertexLighting->integer != 0 )
 	{
 		return;
 	}


illwieckz:
I still set the lightmap pointer to `tr.blackImage` because I cannot guarantee there is no code that may attempt to load them. 

This `tr.blackImage` image is always loaded anyway, it's a default one meant to be used in many features (for example the in-game console makes use of it) and to be used as fallback by many features.

Setting this `tr.blackImage` as lightmap or deluxemap when the features are disabled makes sure this patch is not that much intrusive, making sure that if some code tries to access them, there is no invalid pointer returned. Lightmaps can be explicitly referenced in some materials for doing some light effects so making sure the engine never tries to access them would require a lot of work.

illwieckz:
Expected to fix #646.

DolceTriade (approved):

DolceTriade:
optional nit: Prefer having an if for the affirmative case first rather than the negative case (ie, if vertex enabled, do this rather than if vertex disabled do this).

illwieckz:
> optional nit: Prefer having an if for the affirmative case first rather than the negative case (ie, if vertex enabled, do this rather than if vertex disabled do this).

For historical reasons the whole engine still uses “vertex lighting is enabled” as a poor implementation of “light mapping is disabled”. Vertex lighting is even not implemented in engine at all, the engine does grid lighting when light mapping is disabled or unavailable (like on models). What we need is to tell the engine light mapping is disabled or unavailable.

So for now `r_vertexLighting->integer != 0` is a temporary and poor substitute for the future `r_lightMapping->integer == 0` implementation. I plan to rewrite `r_vertexLighting != 0` as `r_lightMapping == 0`  one day, we have to.

illwieckz:
So this:

```c++
if ( r_vertexLighting->integer == 0 )
{
	image = R_CreateImage( va( "%s/%s", mapName, lightmapFiles[ i ] ), (const byte **)&ldrImage, width, height, 1, imageParams );
}
else
{
	image = tr.blackImage;
}
```

actually means:

```c++
if ( /* light mapping is enabled */ )
{
	image = R_CreateImage( va( "%s/%s", mapName, lightmapFiles[ i ] ), (const byte **)&ldrImage, width, height, 1, imageParams );
}
else
{
	image = tr.blackImage;
}
```

illwieckz:
@DolceTriade I feel so stupid, I already implemented “do not load light maps when light mapping is disabled” in d2624fe9bb9d729629ef13dfcf5a152f342d82d1 from #358. 🤦‍♀️️😅️

What was missing was to not load deluxe maps when deluxe mapping is disabled (same example with thunder map, it would save 25.9MB of texture upload to GPU).

Also with more investigations (and it looks like I already did it two years ago when implementing “do not load light map when light mapping is disabled”), we don't need to set `blackImage` as a fallback. So the code is now much more simpler.

I added an extra bikeshedding commit that will make easier to automatize the rewrite the obsolete usage of `r_vertexLighting->integer` in the future.

DolceTriade:
lgtm

PR #750, opened 2022-11-24
tr_shader: disable extra lightmap stages when lightmapping is disabled

illwieckz:
_Note:_ It also removes a previous implementation that revealed itself to be broken.

_Note:_ It would be very easy to add a cvar to reuse this code to disable light styles when light mapping is enabled. This can be useful for getting better performances on low presets.

Light styles are only compatible with light mapping as light styles are just extra light map stages with blend effects.

So we disable extra light map stages when light styles are not supported.

The first light map stage can be rendered using grid lighting, the other ones have no way to be rendered.

When light mapping is disabled, this:

```c
gloom2/7C9AD47A87BC98CCA1FF30C4D788DD47
{ // Q3Map2 defaulted
	{
		map $lightmap
		rgbGen identity
	}
	// Q3Map2 custom lightstyle stage(s)
	{
		map $lightmap
		blendFunc GL_SRC_ALPHA GL_ONE
		rgbGen wave noise 1 .75 1.6 4.2 // style 2
		tcGen lightmap
		tcMod transform 1 0 0 1 0.54492 0.10156
	}
	{
		map $lightmap
		blendFunc GL_SRC_ALPHA GL_ONE
		rgbGen wave noise 1 .5 3.7 4.9 // style 3
		tcGen lightmap
		tcMod transform 1 0 0 1 -0.38672 0.19043
	}
	{
		map $lightmap
		blendFunc GL_SRC_ALPHA GL_ONE
		rgbGen wave noise 1 1 2.6 1.3 // style 4
		tcGen lightmap
		tcMod transform 1 0 0 1 0.24023 0.19043
	}
	{
		map textures/gloom2/e8clangfloor05c.tga
		blendFunc GL_DST_COLOR GL_ZERO
		rgbGen identity
	}
}
```

or this:

```c
gloom2/7C9AD47A87BC98CCA1FF30C4D788DD47
{ // Q3Map2 defaulted
	{
		map $lightmap
		rgbGen identity
	}
	// Q3Map2 custom lightstyle stage(s)
	{
		map maps/gloom2/lm_0000.tga
		blendFunc GL_SRC_ALPHA GL_ONE
		rgbGen wave noise 1 .75 1.6 4.2 // style 2
		tcGen lightmap
		tcMod transform 1 0 0 1 0.54492 0.10156
	}
	{
		map maps/gloom2/lm_0000.tga
		blendFunc GL_SRC_ALPHA GL_ONE
		rgbGen wave noise 1 .5 3.7 4.9 // style 3
[...]

PR #751, opened 2022-11-24
tr_shader: add r_lightStyles cvar to disable light styles

illwieckz:
#750 must be merged first, the target of this branch has to be changed to `master` once #750 is merged.

This allows to set `r_lightStyles 0` to disable light styles.

Light styles are special effects that blends multiple light maps to create fake dynamic lights, like lights of blinking neons.

Every extra light map is a different material stage and can't be collapsed, meaning every extra light map requires an extra render pass.

When extra light map tiles are stored in a light map file not shared with any static light map tile, the option would even prevent the loading of those extra light map files from file system and not upload them to GPU, reducing both load time and memory usage.

The purpose of this option is to be enabled by default but disabled in low graphics presets.


DolceTriade:
Seems reasonable. LGTM

PR #752, opened 2022-11-25
Do not require pthread, cursesw, minizip and zlib when only building the game code

illwieckz:
Do not require pthread, cursesw, minizip and zlib when only building the game code.

Note: for some reasons zlib is still needed when building nacl game code.

We may be able to disable more code in `Fileystem.cpp` that may be unused when building game code, but at least the zlib-specific code is disabled and then not linked against.

This is motivated by some work-in-progress changes I'm doing to the release scripts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -602,24 +602,26 @@ if (DAEMON_PARENT_SCOPE_DIR)
 endif()
 
 # Curses, pdcurses on Windows and ncursesw on Unix
-if (USE_CURSES AND NOT NACL)
-    if (WIN32)
-        set(LIBS_ENGINE_BASE ${LIBS_ENGINE_BASE} gdi32 comdlg32)
+if (BUILD_CLIENT OR BUILD_TTY_CLIENT OR BUILD_SERVER OR BUILD_DUMMY_AP)
+    if (USE_CURSES)
+        if (WIN32)
+            set(LIBS_ENGINE_BASE ${LIBS_ENGINE_BASE} gdi32 comdlg32)
 
-        add_definitions(-DPDC_WIDE -DPDC_FORCE_UTF8 -DUSE_CURSES)
-        add_library(srclibs-pdcurses EXCLUDE_FROM_ALL ${PDCURSESLIST})
-        if (MSVC)
-            target_compile_options(srclibs-pdcurses PRIVATE "/W2") # decrease warning level
+            add_definitions(-DPDC_WIDE -DPDC_FORCE_UTF8 -DUSE_CURSES)
+            add_library(srclibs-pdcurses EXCLUDE_FROM_ALL ${PDCURSESLIST})
+            if (MSVC)
+                target_compile_options(srclibs-pdcurses PRIVATE "/W2") # decrease warning level
+            endif()
+            set_target_properties(srclibs-pdcurses PROPERTIES POSITION_INDEPENDENT_CODE 1 FOLDER "libs")
+            set(LIBS_ENGINE_BASE ${LIBS_ENGINE_BASE} srclibs-pdcurses)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -602,24 +602,26 @@ if (DAEMON_PARENT_SCOPE_DIR)
...
 # Curses, pdcurses on Windows and ncursesw on Unix
-if (USE_CURSES AND NOT NACL)
-    if (WIN32)
-        set(LIBS_ENGINE_BASE ${LIBS_ENGINE_BASE} gdi32 comdlg32)
-
-        add_definitions(-DPDC_WIDE -DPDC_FORCE_UTF8 -DUSE_CURSES)
-        add_library(srclibs-pdcurses EXCLUDE_FROM_ALL ${PDCURSESLIST})
-        if (MSVC)
-            target_compile_options(srclibs-pdcurses PRIVATE "/W2") # decrease warning level
+if (BUILD_CLIENT OR BUILD_TTY_CLIENT OR BUILD_SERVER)
```
slipher, line 609: or BUILD_DUMMY_APP

src/common/FileSystem.cpp
```diff
@@ -674,6 +680,7 @@ void File::SetLineBuffered(bool enable, std::error_code& err) const
 		ClearErrorCode(err);
 }
 
+#if !defined(BUILD_VM)
```
slipher, line 683: Is there some difference between `defined(BUILD_ENGINE)` and `!defined(BUILD_VM)` ?

necessarily-equal:
Wait, does that mean that we could use threads in the gamelogic ?

src/common/FileSystem.cpp
illwieckz: Ah no, I first wrote it the `!defined(BUILD_VM)` way then rewrote as `defined(BUILD_ENGINE)` but forgot this occurrence, I will fix that.

illwieckz:
> Wait, does that mean that we could use threads in the gamelogic ?

No, it just meant CMake was aborting if pthread was not found even when not compiling the engine because it just checked for it as a requirement in all cases.

Well, it may be possible to use threads in dll/exe games if I'm right, but not in nacl, so the game code doesn't use threads at all.

CMakeLists.txt

src/common/FileSystem.cpp
illwieckz: Fixed.

illwieckz:
So I fixed the two remarks.

CMakeLists.txt
```diff
@@ -586,40 +588,44 @@ endif()
...
 # Curses, pdcurses on Windows and ncursesw on Unix
-if (USE_CURSES AND NOT NACL)
-    if (WIN32)
-        set(LIBS_ENGINE_BASE ${LIBS_ENGINE_BASE} gdi32 comdlg32)
-
-        add_definitions(-DPDC_WIDE -DPDC_FORCE_UTF8 -DUSE_CURSES)
-        add_library(srclibs-pdcurses EXCLUDE_FROM_ALL ${PDCURSESLIST})
-        if (MSVC)
-            target_compile_options(srclibs-pdcurses PRIVATE "/W2") # decrease warning level
+if (BUILD_CLIENT OR BUILD_TTY_CLIENT OR BUILD_SERVER OR BUILD_DUMMY_AP)
```
slipher, line 609: BUILD_DUMMY_APP is spelled wrong and also there are 2 other cases where it is missing.

illwieckz: oh lol, I'll fix that. Where are the other places? I don't know at all what needs or doesn't need the dummy app.

illwieckz: See:

- https://github.com/DaemonEngine/Daemon/pull/754

PR #755, opened 2022-11-27
external_deps: do not prevent i686, and minor touchup (almost bikeshedding)

illwieckz:
- linux-i686 cost is null

Now that the toolchain is fully cross-compilation compliant, building the engine for i686 is just a matter of setting `CC` to `i686-linux-gnu-gcc`. So the cost of it is null.

There is no change to be done on engine code neither on engine cmake to build a linux i686 engine as we already take care of engine i686 compatibility for Windows.

While the oldest supported amd64 CPU is 21 years old and the oldest supported GPU is 20 years old, Intel introduced new 32-bit-only processors until 2011 _included_, so the more recent 32-bit only Intel CPU is 11 years old. Also Intel sold amd64 compatible CPUs as 32-bit CPUs hiding the fact they were 64-bit compatible. I myself hosted linux network services for companies I worked for on such hardware, given I was _aware_ of it I ran amd64 distros on them, but other people may run i686 distros on them because the hardware is officially advertised as i686 only. Those would be perfect to run unvanquished servers.

The same way [people asked for arm support to host servers](https://github.com/Unvanquished/Unvanquished/issues/1015), there is no reason a low-end intel CPU can't host a server. On Via side, Via C7 seems to provide all features to run a server. Those kind of Atom/Via CPUs are very common in devices like light clients, nettop, NAS, and things like that, which are perfect as low-consumption servers and cost nothing on second hand market.

Even running the more recent amd64-compatible Atom/Via CPUs as 32-bit may be a wise choice for hosting a server, with hardware just costing 20$ to host a home server.

I myself hosted my Unvanquished server on a low-end Via CPU for years. I assume the only reason why we didn't get any request for this is that our user base is small. The same way our user base is known to not have reached yet less developed countries, or we don't know about it.

We know from SourceForge we have lots of downloads from South-American countries without never seeing the players, except once when some Brazilian people came to troubleshoot problems when hosting LAN servers (so we know we have users we never see online).

We also received reports last year from a French non-profit organization refurbishing old computers to play LAN games with children, including reports for Via GPUs from 2005 and Intel GPUs from 2007, which gives a glance on the kind of hardware they refurbish for playing LAN games.

Given the cost of linux-i686 is null, I see no reason to prevent the building of it.

- more wasmtime and wasisdk care

I added more switch/case stuff to correctly report errors on newly supported platforms like arm64. I updated versions so the knowledge baked in the script matches an actual release.

I noticed Wasmtime doesn't provide 32-bit i686 build at all so that would be very problematic when moving to Wasm as this would phase out windows-i686 build, which is believed to still be used a lot (average people don't reinstall their preinstalled Windows). That would also phase out 32-bit arm support which is still the default and recommended download for Raspberry Pi, the first cause of request for arm and rapsberry pi support being servers.

I noticed the Wasi SDK only provides amd64 build, which is similar to the NaCl SDK that only provided amd64 and i686 build if I'm right (no arm NaCl SDK for sure).

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -550,8 +550,8 @@ build_naclsdk() {
 		# TODO(0.54): Unify all arch strings using i686 and amd64 strings.
 		cp pepper_*"/tools/irt_core_x86_64.nexe" "${PREFIX}/irt_core-x86_64.nexe"
 		;;
-	linux-amd64-*)
-		cp pepper_*"/tools/nacl_helper_bootstrap_x86_64" "${PREFIX}/nacl_helper_bootstrap"
+	linux-amd64-*|linux-i686-*)
+		cp pepper_*"/tools/nacl_helper_bootstrap_${NACLSDK_ARCH}" "${PREFIX}/nacl_helper_bootstrap"
 		# Fix permissions on a few files which deny access to non-owner
 		chmod 644 "${PREFIX}/irt_core-${DAEMON_ARCH}.nexe"
 		chmod 755 "${PREFIX}/nacl_helper_bootstrap" "${PREFIX}/sel_ldr"
@@ -757,6 +757,19 @@ setup_macos-amd64-default() {
 	export CMAKE_OSX_ARCHITECTURES="x86_64"
 	common_setup
 	export NASM="${PWD}/${BUILD_BASEDIR}/prefix/bin/nasm" # A newer version of nasm is required for 64-bit
+}
+
+# Set up environment for 32-bit i686 Linux
+setup_linux-i686-default() {
+	HOST=i386-unknown-linux-gnu
+	CROSS=
+	MSVC_SHARED=(--disable-shared --enable-static)
+	export CC='i686-linux-gnu-gcc'
+	export CXX='i686-linux-gnu-g++'
+	export CFLAGS='-fPIC'
+	export CXXFLAGS='-fPIC'
+	export LDFLAGS=''
+	common_setup
 }
 
 # Set up environment for 64-bit amd64 Linux
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -31,8 +31,8 @@ OPUSFILE_VERSION=0.12
 LUA_VERSION=5.4.4
 NACLSDK_VERSION=44.0.2403.155
 NCURSES_VERSION=6.2
-WASISDK_VERSION=12.0
-WASMTIME_VERSION=0.28.0
+WASISDK_VERSION=16.0
+WASMTIME_VERSION=2.0.2
 
 # Extract an archive into the given subdirectory of the build dir and cd to it
 # Usage: extract <filename> <directory>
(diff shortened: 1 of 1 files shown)


illwieckz:
For completeness, the dæmon engine can run on x86 processors implementing the NX bit, on AMD side it starts with AMD64, on Intel side it starts with the E0 stepping of Prescott Pentium 4 processor ([source](http://h10032.www1.hp.com/ctg/Manual/c00387685.pdf#page=5)).

external_deps/build.sh
```diff
@@ -759,6 +772,19 @@ setup_macos-amd64-default() {
 	export NASM="${PWD}/${BUILD_BASEDIR}/prefix/bin/nasm" # A newer version of nasm is required for 64-bit
 }
 
+# Set up environment for 32-bit i686 Linux
+setup_linux-i686-default() {
+	HOST=i386-unknown-linux-gnu
+	CROSS=
+	MSVC_SHARED=(--disable-shared --enable-static)
```
DolceTriade, line 779: Why are these flags required?

necessarily-equal (commented):
For the record, I don't think adding support for a new architecture has zero cost: there is the testing and support cost.

Adding support to x86_32 seem anachronistic today. So I had a look and tried to see how popular i686 might be. It seems Ubuntu, Fedora, OpenSuse, ArchLinux, Manjaro, RedHat, PopOS and Linux Mint have dropped the support for it. Or never had it in the first place.

In fact I had more trouble to find a 32 bits distro than one which had dropped the support entirely. Some example of distros with 32 bits spin would be Debian, some Debian spinoffs like Puppy Linux, CentOS and [a community-maintained fork of ArchLinux](https://archlinux32.org/).

Debian Stable could be the largest holdout. So I tried having a look on that side.

The [Debian popularity contest for Debian Stable](https://popcon.debian.org/stable/index.html) suggests that x86_32 is on its way out, and is about 3% as much package installations as x86_64. And this includes multilib package installations, which may represent the majority of the installations — and we don't need or want to support these since they can run our x86_64 daemon.

This suggests that the linux-i386 user base is really small. I highly doubt this is worth it, especially if that could stand in the way of future improvement like switching to wasm.

necessarily-equal:
(I've modified the title to add that it also builds for a new architecture)

external_deps/build.sh
slipher: MSVC_SHARED is just goofily named. It defines whether static or shared libraries are preferred for the platform. MSVC is the only one where shared is preferred.

slipher (commented):
Personally I don't like "rely more on variables"

Others LGTM

illwieckz:
> Personally I don't like "rely more on variables"

I removed the EXE thing which was just a matter of taste. The other variable change was needed by the following commit so I squashed the other variable change with the following commit.

illwieckz:
> For the record, I don't think adding support for a new architecture has zero cost: there is the testing and support cost.

@necessarily-equal That's why I don't call this support (I renamed the title to say “do not prevent” instead). It's just I don't think it makes sense to artificially prevent to build something that works. There is no support meant to be provided at all.

That said, one hidden interest of i686 may be emulation. For example the very first person who talked with me about his desire to try to run the Unvanquished client (not the server) on an arm device was asking me if we had a 32-bit build of Unvanquished. The reason was he believed he would get more performance emulating i686 with box32 than emulating amd64 with box64. This need is now completely obsolete as the engine now builds and runs both as arm and arm64, but for legacy reason, there are strong motivations to emulate i686 (same for amd64, but this one has less history). None of this is an argument in favor of _supporting_ such platform, but this contributed to me wondering if that platform was still working or not, it looks like there is nothing incompatible in the code.

For example in the past we had a specific cmake toolchain file for this platform. This was very different to the current status, as there is now nothing specific in our cmake files written for the `linux-i686` platform, we don't have to deal with the `linux-i686` platform when editing cmake files or the C++ code.

illwieckz:
I'm impressed by the growth of arm64:

20221128-000 debian-submission

Also arm64 just outperformed armhf this year, which confirms how much armhf is strong.

Given submission is not enabled by default in Debian as far as I know, those numbers for i686 as the second-most popular Debian platform isn't bad! I would love to get those users on Unvanquished!

necessarily-equal:
Note those Debian numbers include multiarch packages installations. I suspect multiarch installations actually represents the majority of i386 usage, and we don't need to support these IMO because they can also just run the 64bits binary.

PR #756, opened 2022-11-28
external_deps: factorize platform definition and other things

illwieckz:
Some various changes:

- factorize platform definition

Right now adding a linux platform would be as easy as adding a one-line function, this way:

```bash
setup_linux-amd64-default() {
	common_setup linux x86_64-unknown-linux-gnu
}
```

- rename `MSVC_SHARED` as `CONFIGURE_SHARED`

It makes the name more relevant.

- remove `-m32` and `-m64` flags

We already hardcode the 32-bit and 64-bit compilers anyway.

- always reset flags from the environment and require compiler names to be hardcoded

We heavily cross-compile and must not inherit any stray flag from environment.

Require the compiler names to be explicitly hardcoded, we should not inherit them from environment as we heavily cross-compile. They are already all hardcoded, make sure to get an error when adding a new platform without writing down in the script the compiler to use for that platform.

- always set -msse2 on i686 whatever the system or compiler

This is actually the only change that should produce different binaries when building deps with that new version of the script, as I forgot to add it to `linux-i686`. There is now no need to set this flag per platform, it's just set to all i686 platforms.

- various rewrites, especially simplifies some variable usage as some being always set, we don't need to clutter the code with syntax to catch undefined variables, and delete some variables that are now useless.

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -696,12 +696,6 @@ common_setup() {
 		export CXXFLAGS="${CXXFLAGS} -msse2"
 		;;
 	esac
-	if [[ "${MSYSTEM:-}" = MINGW* ]]; then
-		# Experimental MSYS2 support. Most packages won't work;
-		# you need to cross-compile from Linux
-		export CMAKE_GENERATOR="MSYS Makefiles"
-		CROSS= # Doing this because there is no ${CROSS}strip
-	fi
 	mkdir -p "${DOWNLOAD_DIR}"
 	mkdir -p "${PREFIX}"
 	mkdir -p "${PREFIX}/bin"
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -34,6 +34,10 @@ NCURSES_VERSION=6.2
 WASISDK_VERSION=16.0
 WASMTIME_VERSION=2.0.2
 
+# Require the compiler names to be explicitly hardcoded, we should not inherit them
+# from environment as we heavily cross-compile.
+export CC='false'
+export CXX='false'
 # Set defaults.
 export LD='ld'
 export AR='ar'
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -144,10 +144,10 @@ build_gmp() {
 	case "${PLATFORM}" in
 	macos-*-*)
 		# The assembler objects are incompatible with PIE
-		CFLAGS="${CFLAGS:-} -O2" ./configure --host="${HOST}" --prefix="${PREFIX}" ${MSVC_SHARED[@]} --disable-assembly
+		CFLAGS="${CFLAGS:-} -O2" ./configure --host="${HOST}" --prefix="${PREFIX}" ${CONFIGURE_SHARED[@]} --disable-assembly
 		;;
 	*)
-		CFLAGS="${CFLAGS:-} -O2" ./configure --host="${HOST}" --prefix="${PREFIX}" ${MSVC_SHARED[@]}
+		CFLAGS="${CFLAGS:-} -O2" ./configure --host="${HOST}" --prefix="${PREFIX}" ${CONFIGURE_SHARED[@]}
 		;;
 	esac
 
@@ -167,7 +167,7 @@ build_nettle() {
 	download "nettle-${NETTLE_VERSION}.tar.gz" "https://ftp.gnu.org/gnu/nettle/nettle-${NETTLE_VERSION}.tar.gz" nettle
 	cd "nettle-${NETTLE_VERSION}"
 	# The default -O2 is dropped when there's user-provided CFLAGS.
-	CFLAGS="${CFLAGS:-} -O2" ./configure --host="${HOST}" --prefix="${PREFIX}" ${MSVC_SHARED[@]}
+	CFLAGS="${CFLAGS:-} -O2" ./configure --host="${HOST}" --prefix="${PREFIX}" ${CONFIGURE_SHARED[@]}
 	make
 	make install
 }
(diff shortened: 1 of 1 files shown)


external_deps/build.sh
```diff
@@ -759,7 +759,6 @@ setup_windows-amd64-mingw() {
 # Set up environment for 64-bit amd64 macOS
 setup_macos-amd64-default() {
 	HOST=x86_64-apple-darwin11
-	CROSS=
```
slipher, line 762: Won't removing CROSS for some platforms cause errors because we have `set -u`? Looks like Lua uses CROSS unconditionally.

slipher (commented):
I stopped reviewing at "factorize common platform setup"

external_deps/build.sh
illwieckz: Right, this one is entirely removed later. I now squashed that commit `it with `entirely remove CROSS variable`, I also extracted some CROSS-related editions from `factorize common platform setup` and squashed them in `entirely remove CROSS variable` as well. So now `entirely remove CROSS variable` does all the CROSS removal job and `factorize common platform setup` is simpler.

external_deps/build.sh
```diff
@@ -674,154 +690,134 @@ build_wipe() {
...
+	CFLAGS+=' -D__USE_MINGW_ANSI_STDIO=0'
+}
+
+common_setup_mingw() {
+	CC="${HOST}-gcc"
+	CXX="${HOST}-g++"
+	LD="${HOST}-ld"
+	AR="${HOST}-ar"
+	RANLIB="${HOST}-ranlib"
+	export STRIP="${HOST}-strip"
```
slipher, line 749: We didn't have STRIP generally exported before. GLEW is the only one using this so it's better not to export it everywhere

external_deps/build.sh
```diff
@@ -674,154 +690,134 @@ build_wipe() {
...
+# point did not implement certain printf specifiers, in particular %lld (long long).
+# Lua does use this one, which results in compiler warnings. But this is OK because
+# the Windows build of Lua is only used in developer gamelogic builds, and Microsoft
+# supports %lld since Visual Studio 2013.
+common_setup_msvc() {
+	CONFIGURE_SHARED=(--enable-shared --disable-static)
+	# Libtool bug prevents -static-libgcc from being set in LDFLAGS
+	CC="${HOST}-gcc -static-libgcc"
+	CXX="${HOST}-g++ -static-libgcc"
+	export DLLTOOL="${HOST}-dlltool"
```
slipher, line 739: Does not need to be exported

external_deps/build.sh
illwieckz: Hmm, for this one this is not easy, as there is another code somewhere else reverting that `STRIP`:

```sh
	if [[ "${MSYSTEM:-}" = MINGW* ]]; then
		# Experimental MSYS2 support. Most packages won't work;
		# you need to cross-compile from Linux
		export CMAKE_GENERATOR="MSYS Makefiles"
		export STRIP='strip' # There is no ${HOST}-strip
	fi
```

illwieckz: Well, the code was even doing that before:

```sh
CROSS= # Doing this because there is no ${CROSS}strip
```

Which was expected to break the all the other things previously relying on CROSS (like ar, ranlib…)

slipher: You can delete the MSYS2 thing if that makes things earlier. It probably wasn't worth merging.

illwieckz: I think the same, and it can be redone if someone want's to pick that task.

illwieckz: I removed that incomplete MSYS2 thing.

external_deps/build.sh
illwieckz: I removed that `DLLTOOL` variable, it is only used once anyway.

external_deps/build.sh
illwieckz: For the `STRIP`, well I doubt it has any effect, it's not like `CC`. Anyway I removed it.

external_deps/build.sh
```diff
@@ -674,6 +674,8 @@ build_wipe() {
 
 # Common setup code
 common_setup() {
+	HOST="${2}"
+	"common_setup_${1}"
```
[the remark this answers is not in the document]
illwieckz: I added a commit that de-factorize that but I really dislike that as it makes it much verbose and require more boiler plate when adding platforms, making it more error prone.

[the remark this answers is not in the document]
illwieckz: This more-verbose-more-boilerplate-more-copypaste-error-prone-implementation was one of my early factorization iteration I factorized over again, and was only meant to be a temporary step in a step-by-step factorization effort, I would consider merging that extra commit as a bad.

This is a build script, this can suffer an extra indirection if that removes boilerplate copypaste. With indirection, one cannot add an extra platdform and forget to write the platform parent without getting an obvious error yelled by the bash, while with the boilerplate code, one can add a platform parent and forget to call `common_setup` and get unobvious errors instead.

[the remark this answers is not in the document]
illwieckz: I removed the `more-verbose-more-boilerplate-more-copypaste-error-prone-implementation` commit because In find this stupid to add more than 10 lines of copypaste just to avoid an extra level of indirection in a build script that is likely used by two people once in a year…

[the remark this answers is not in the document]
slipher: > I removed the `more-verbose-more-boilerplate-more-copypaste-error-prone-implementation` commit

RIP, I never got to see it. But it's not a super important issue so LGTM

PR #760, opened 2022-12-03
Shut down to stop time overflow

### LINKED ISSUE #758: The DNS resolver may uncontrollably flood daemon.log
I have discovered this morning that my daemon.log is some 36G after seemingly failing to resolve the master server overnight.

While I have not been able to analyse the 36G daemon.log because I had to restore a backup of my VPS I surmise that it was caused by this as the stdout was flooded with the following message:

`Failed to resolve hostname`

before daemonded finally died with:

`terminate called after throwing an instance of 'std::system_error' what():  No space left on device` after my VPS's virtual disk was filled completely.

Please can this be investigated as a matter of urgency - either stop the spam (and enabling `logs.suppression.enabled` is not the answer because this breaks other things on the server console) or prevent it from logging this garbage to the disk (not really the answer either, because some server owner may run this as a systemd daemon, which means it would just flood the journal anyway).

slipher:
Fixes #758.

### DIFF
--- a/src/common/Log.cpp
+++ b/src/common/Log.cpp
@@ -106,7 +106,7 @@ namespace Log {
     static const int noticeTargets = (1 << GRAPHICAL_CONSOLE) | (1 << TTY_CONSOLE) | (1 << LOGFILE);
     static const int warnTargets = (1 << GRAPHICAL_CONSOLE) | (1 << TTY_CONSOLE) | (1 << LOGFILE);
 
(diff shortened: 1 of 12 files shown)


necessarily-equal:
I'm not sure about the principle since that means servers go down by themselves in an unexpected way, but also this is better than the current state i guess.

LGTM, let's hope people do restart their services automatically.

DolceTriade (approved):

PR #763, opened 2022-12-05
Drop checksumFeed

necessarily-equal:
From reading, this shouldn't break demo loading for demos recorded before this commit. This changes the netcode however, so this PR is against 0.54.0/sync.


DolceTriade:
lgtm. Seems like it is a remnant of the old FS code. we don't use it anymore so it can be safely deleted.

PR #766, opened 2022-12-11
Read various files from pakpath only (not homepath)

slipher:
A lot of places in the engine are using a file opening function that checks both VFS and homepath for stuff that should really only come from the VFS. For example if q3shaders are read from the homepath then you could cheat by adding a shader that makes walls invisible or something.

### DIFF
--- a/src/engine/renderer/tr_image_jpg.cpp
+++ b/src/engine/renderer/tr_image_jpg.cpp
@@ -93,24 +93,11 @@ void LoadJPG( const char *filename, unsigned char **pic, int *width, int *height
 	unsigned int          pixelcount, memcount;
 	unsigned int          sindex, dindex;
 	byte                  *out;
-	int                   len;
-	union
-	{
-		byte *b;
-		void *v;
-	} fbuffer;
-
 	byte *buf;
 
-	/* In this example we want to open the input file before doing anything else,
-	 * so that the setjmp() error recovery below can assume the file is open.
-	 * VERY IMPORTANT: use "b" option to fopen() if you are on a machine that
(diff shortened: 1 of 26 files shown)


src/engine/renderer/tr_image_jpg.cpp
```diff
@@ -135,10 +122,10 @@ void LoadJPG( const char *filename, unsigned char **pic, int *width, int *height
 	/* Step 2: specify data source (eg, a file) */
 
 #if JPEG_LIB_VERSION < 80
-	jpegfd = fmemopen( fbuffer.b, len, "r" );
+	jpegfd = fmemopen( reinterpret_cast<const unsigned char*>(data.data()), len, "r" );
```
necessarily-equal, line 125: You need this not to break things for some code path:
```suggestion
	jpegfd = fmemopen( reinterpret_cast<const unsigned char*>(data.data()), data.size(), "r" );
```

I have an impression that this `#if JPEG_LIB_VERSION < 80` codepath is dumb though, because the biggest libjpeg implementation today is libjpeg-turbo and while it exposes an ABI compatible with libjpeg 6.2 (JPEG_LIB_VERSION == 62), it implements and exposes `jpeg_mem_src` since at least debian oldoldstable.

I've opened https://github.com/DaemonEngine/Daemon/pull/769 for that, you can cherry-pick it here if you want.

slipher: I fixed it in case we decide not to remove that codepath

DolceTriade:
lgtm

DolceTriade (approved):

PR #769, opened 2022-12-11
Remove old libjpeg 6.2 checks

necessarily-equal:
Current implementations should likely be either more recent than libjpeg6.2 or be libjpeg-turbo which from at least debian-oldoldstable supports the new command, even though it exposes `JPEG_LIB_VERSION` == 62.

### DIFF
--- a/src/engine/renderer/tr_image_jpg.cpp
+++ b/src/engine/renderer/tr_image_jpg.cpp
@@ -101,9 +101,6 @@ void LoadJPG( const char *filename, unsigned char **pic, int *width, int *height
 	} fbuffer;
 
 	byte *buf;
-#if JPEG_LIB_VERSION < 80
-	FILE *jpegfd;
-#endif
 
 	/* In this example we want to open the input file before doing anything else,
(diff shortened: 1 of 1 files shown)


src/engine/renderer/tr_image_jpg.cpp
```diff
@@ -102,7 +102,7 @@ void LoadJPG( const char *filename, unsigned char **pic, int *width, int *height
 
 	byte *buf;
 #if JPEG_LIB_VERSION < 80
```
slipher, line 104: Missed some spots...

necessarily-equal: My bad, forgot to push the latest version.

necessarily-equal:
To answer the

> I guess that code path is not entirely dead...

I agree, it triggers on Linux about everywhere where libjpeg-turbo is around, which should be around everywhere. That's because libjpeg-turbo prefers to expose an API which is an compatible with 6.2, with enhancements like the `jpeg_mem_src` function. This is why I think these if are actually somewhat detrimental.

slipher:
Oh I get it. LGTM

Issue #532, opened 2021-10-12
MSAA antialiasing is not implemented


illwieckz:
It was disabled when moving to OpenGL core profile in 2016, see f3705e57cd6b06aa7ce176be207840ab3270eda4

When I re-enable it I don't get it working anyway, even with compatibility profile.

illwieckz:
Related cvar to enable the feature: `r_ext_multisample` (int, takes value 0, 2, 4, 8, 16), `vid_restart` has to be called. See also Video tab in Graphics options (tweak related option and then click Apply button).

I don't know if that should be working outside of rendering of maps, so better load a map and see at edges of a corner in the map. The aliasing should look altered (I don't know how).

The cvar looks to currently be orphaned.

illwieckz:
There may be documentation to implement it on https://learnopengl.com/Advanced-OpenGL/Anti-Aliasing

and https://learnopengl.com/code_viewer_gh.php?code=src/4.advanced_opengl/11.2.anti_aliasing_offscreen/anti_aliasing_offscreen.cpp

I made a quick try but failed.

Anyway, if we fix FXAA, we may decide to never implement MSAA:

- https://github.com/DaemonEngine/Daemon/issues/533

Currently we don't have any working anti-aliasing method, MSAA isn't implemented and FXAA does nothing.

illwieckz:
If we decide to implement MSAA one day, we may want to name the cvar `r_MSAA` like we have `r_FXAA`, and never mind about the old `r_ext_multisample` cvar.

DolceTriade:
Worst case, we can use GPU based overrides like in nvidia-settings?

PR #773, opened 2023-01-03
renderer: drop unused `r_ext_multisample` and `r_alphabits` cvars

illwieckz:
renderer: drop unused `r_ext_multisample` and `r_alphabits` cvars

- `r_ext_multisample` belongs to MSAA and MSAA was only implemented in legacy renderer and was never reimplemented since.
- `r_alphabits` was only used in the legacy renderer.
____

Initial message:

We set some `r_ext_multisample` values with Unvanquished graphics preset, but the feature is not implemented yet and then the values aren't tested. It means the day we implement the feature, people may have the bad surprise of a potential performance hit because the variable may have been silently set to an high value before.

This patch ensures that any future released build of the engine will reset the cvar on user's configuration to clean-up them from random untested values.

Deleting the cvar would not reset the bad value in user's config, so if we delete the cvar today and re-add it in the future the day we implement the feature, the untested user's cvar would still be there and used.

See also https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/169

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -103,7 +103,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 	cvar_t      *r_depthbits;
 	cvar_t      *r_colorbits;
-	cvar_t      *r_alphabits;
 
 	cvar_t      *r_drawBuffer;
 	cvar_t      *r_shadows;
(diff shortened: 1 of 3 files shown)


slipher:
What about just giving it a different name when it is implemented. The `ext` prefix seems meaningless and incomprehensible anyway

illwieckz:
I assume those cvars are prefixed with `_ext` because at OpenGL 2.1 time they were enabling features that were only available through OpenGL extension, here it was `GL_EXT_framebuffer_multisample`.

We have a bunch of those OpenGL-extension-named cvars:

```c++
	extern cvar_t *r_ext_multisample;  // desired number of MSAA samples
	// …
	extern cvar_t *r_ext_occlusion_query; // these control use of specific extensions
	extern cvar_t *r_ext_draw_buffers;
	extern cvar_t *r_ext_vertex_array_object;
	extern cvar_t *r_ext_half_float_pixel;
	extern cvar_t *r_ext_texture_float;
	extern cvar_t *r_ext_texture_integer;
	extern cvar_t *r_ext_texture_rg;
	extern cvar_t *r_ext_texture_filter_anisotropic;
	extern cvar_t *r_ext_gpu_shader4;
	extern cvar_t *r_arb_buffer_storage;
	extern cvar_t *r_arb_map_buffer_range;
	extern cvar_t *r_arb_sync;
	extern cvar_t *r_arb_uniform_buffer_object;
	extern cvar_t *r_arb_texture_gather;
	extern cvar_t *r_arb_gpu_shader5;
```

I assume some were just booleans to enable features, while some like `r_ext_multisample` expected to be set to a configuration value (like sample numbers, here) as an enablement.

slipher:
Yeah there are some that make sense - `r_ext_texture_float`, `r_ext_gpu_shader4`, etc. - cvars which actually control extension loading. These may be used by a developer to test compatibility with hardware that lacks the extension. But `r_ext_multisample` is just idiotically named. We should just delete it from the engine and next time something is implemented we will give it a non-idiotic name.

illwieckz:
OK, I suggested that if we implement MSAA we name the cvar `r_MSAA` like the `r_FXAA` cvar we already have:

- https://github.com/DaemonEngine/Daemon/issues/532#issuecomment-1369285204

Let's delete the cvar entirely.

illwieckz:
I rewrote the PR to delete `r_ext_multisample`, I also deleted unused `r_alphabits` while I was at it (old cvar from legacy renderer).

slipher:
LGTM

PR #774, opened 2023-01-03
Fix a bug with filesystem syscalls

slipher:


src/engine/framework/CommonVMServices.cpp
```diff
@@ -274,7 +274,6 @@ namespace VM {
                 IPC::HandleMsg<FSReadMsg>(channel, std::move(reader), [this](int handle, int len, std::string& res, int& ret) {
                     FS_CheckOwnership(handle, fileOwnership);
                     std::unique_ptr<char[]> buffer(new char[len]);
-                    buffer[0] = '\0';
```
DolceTriade, line 277: Should we instead check for `len == 0` and skip this instead?

DolceTriade (approved):

src/engine/framework/CommonVMServices.cpp
slipher: It seems to me that it is nonsense which should not be done in any case.

PR #776, opened 2023-01-12
Small cleanups

necessarily-equal:
One change is a functional change, but that shouldn't change anything for Unvanquished because it cannot trigger.


src/engine/server/sv_client.cpp
```diff
@@ -159,23 +159,9 @@ void SV_DirectConnect( const netadr_t& from, const Cmd::Args& args )
...
-				if ( all_bots )
-				{
-					SV_DropClient( &svs.clients[ sv_maxclients->integer - 1 ], "only bots on server" );
-					new_client = &svs.clients[ sv_maxclients->integer - 1 ];
-				}
-				else
-				{
-					Sys::Error( "server is full on local connect" );
-				}
+				Sys::Error( "server is full on local connect" );
```
slipher, line 164: The restriction against a bot using slot 0 comes not from the gamelogic but `SV_BotAllocateClient`. 

I think the ` && sv_privateClients.Get() < sv_maxclients->integer` part of the condition could also be dropped.

illwieckz:
> Avoid calculation of one unused matrix (in `RB_SetupLightForShadowing`)

I wonder if that's part of why those realtime shadows don't work, maybe some other part of the code wrongly uses another matrix instead of this one?

PR #777, opened 2023-01-12
Disable udp download if http download is available

necessarily-equal:
Hey, so we wanted to do something about udp downloads for a while. This is a proposal where UDP downloads will only be proposed if HTTP downloads are not set up. I'm also fine with nuking the feature, though. Downloading things at 15kB/s is just terrible. And according to @cu-kai, if we increase the maximum bandwidth, « but.. if more than one person starts doing this [downloading over udp], the server feels it ».

Note: in the current state this received little to no testing. This can come later once people say if they like the idea or not.

Fixes https://github.com/Unvanquished/Unvanquished/issues/1926.

### DIFF
--- a/src/engine/server/sv_client.cpp
+++ b/src/engine/server/sv_client.cpp
@@ -811,64 +811,71 @@ void SV_WriteDownloadToClient( client_t *cl, msg_t *msg )
 
 				if ( SV_CheckFallbackURL( cl, pakName.c_str(), downloadSize, msg ) )
 				{
-					return;
+					return; // success (?)
 				}
 
-				Log::Warn("Client '%s': falling back to regular downloading for failed file %s", cl->name,
+				Log::Warn("Client \"%s\" downloading file \"%s\" failed", cl->name,
 						 cl->downloadName );
 			}
+			return; // failure
 		}
+		else
+		{
+			/////
+			// Old-style udp downloads, attempted only if sv_wwwDownload is not set
+			/////
 
-		// find file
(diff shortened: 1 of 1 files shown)


slipher:
How about making a `sv_udpDownloads` cvar to enable it or not, instead of having it randomly depend on HTTP cvars

necessarily-equal:
If we have a cvar for that, then we can't disable udp download as soon as www download is setup. That kind of defeats the point of this: making udp downloads obsolete should there be any other option. I'd rather know if my http download setup is borked that having it fallback immediately to something that is almost unusable.

slipher:
> If we have a cvar for that, then we can't disable udp download as soon as www download is setup.

What do you mean? You could set the cvar to disable UDP right next to where you set the one for HTTP url


> I'd rather know if my http download setup is borked

You mean by seeing people disconnect from the server?

illwieckz:
To me the UDP download is simply too slow to be usable, meaning we don't use it and we never use it and every tutorial giving instructions to enable download would document http download. Our downloadable content is too large for UDP download.

If we don't know how to make UDP download faster, we better remove it entirely, as it is simply unused and will never be.

To make autodownload standalone, I would favor embedding a minimalistic http server we would run on the same port than UDP port but as TCP port. This would also be reusable for other things like downloading levelshot for map vote screen or things like that.

DolceTriade:
A minimal HTTP server isn't a bad idea I suppose.

PR #778, opened 2023-01-13
Prevent R_Rotate buffer overrun

slipher:
I have not tested


illwieckz (approved):
I've tested that code with some Xonotic maps that are known to use different skybox format requiring a rotation and then calling `R_Rotate` and I haven't seen any issue. I haven't tried with bad non-square skybox, but valid square ones look to be rendered properly.

PR #779, opened 2023-01-14
allow depth shader name to be longer than MAX_QPATH, fix #2370

illwieckz:
Allow depth shader name to be longer than `MAX_QPATH`, fix Unvanquished/Unvanquished#2370

Also unify some logging.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5396,7 +5396,18 @@ static shader_t *FinishShader()
 		// keep only the first stage
 		stages[1].active = false;
 		shader.numStages = 1;
-		strcat(shader.name, "$depth");
+
+		const char* depthShaderSuffix = "$depth";
+
+		if ( strlen( shader.name ) + strlen( depthShaderSuffix ) >= MAX_QPATH )
+		{
+			Log::Warn( "Shader name %s%s length longer than MAX_QPATH %d", shader.name, depthShaderSuffix, MAX_QPATH );
+			
+			ret->depthShader = nullptr;
+			return ret;
+		}
+
+		strcat( shader.name, depthShaderSuffix );
 
 		if( stages[0].stateBits & GLS_ATEST_BITS ) {
 			// alpha test requires a custom depth shader
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5144,6 +5144,17 @@ static void GeneratePermanentShaderTable( const float *values, int numValues )
 	shaderTableHashTable[ hash ] = newTable;
 }
 
+bool CheckShaderNameLength( const char* func_err, const char* name, const char* suffix )
+{
+	if ( strlen( name ) + strlen( suffix ) >= MAX_QPATH )
+	{
+		Log::Warn("%s Shader name %s%s length longer than MAX_QPATH %d", func_err, name, suffix, MAX_QPATH );
+		return false;
+	}
+
+	return true;
+}
+
 /*
 =========================
 FinishShader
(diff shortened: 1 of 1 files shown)


src/engine/renderer/tr_shader.cpp
```diff
@@ -5396,7 +5396,16 @@ static shader_t *FinishShader()
 		// keep only the first stage
 		stages[1].active = false;
 		shader.numStages = 1;
-		strcat(shader.name, "$depth");
+
+		const char *shaderSuffix = "$depth";
+		if ( strlen(shader.name) + strlen(shaderSuffix) >= MAX_QPATH )
+		{
+			Log::Warn( "Shader name %s%s exceeds MAX_QPATH", shader.name, shaderSuffix );
```
DolceTriade, line 5403: Let's also tell them what MAX_QPATH is.

illwieckz: Do you mean to print `MAX_QPATH` value?

illwieckz:
I redid the patch in a way the shader name is a bit large in struct so we don't abort when making a depth shader.

src/engine/renderer/tr_shader.cpp
DolceTriade: Yes

illwieckz: I did that, I also rewrote the patch to allow shaders to have length of `MAX_QPATH` + "$depth", except it's checked against MAX_QPATH everywhere it's not a depth shader, so we even not invalidate the depth shader if the non-depth shader has valid path name.

src/engine/renderer/tr_local.h
```diff
@@ -1324,9 +1324,18 @@ static inline void glFboSetExt()
 	  SHADER_LIGHT // light material: attenuation
 	};
 
+	// Size of MAX_QPATH + "$depth".
+	const int depthShaderNameLen = MAX_QPATH + 6;
```
DolceTriade, line 1328: constexpr is the new cool. Also do you need +7 for null char?

since its a constant should we call it DEPTH_SHADER_NAME_LEN?

src/engine/renderer/tr_local.h
```diff
@@ -1324,9 +1324,18 @@ static inline void glFboSetExt()
...
+
 	struct shader_t
 	{
-		char         name[ MAX_QPATH ]; // game path, including extension
+		/* Shader path, including optional internal depth suffix and extension.
+		The code should always check for shader name to not be longer than MAX_QPATH
+		except when checking for depthShader names. This string is longer than
+		MAX_QPATH for the only reason we add a suffix to existing shader names when
+		creating a depth shader. */
+		char name[ depthShaderNameLen ];
```
DolceTriade, line 1337: Will this allow any shader to be longer than MAX_QPATH? will this cause any issues if we have other buffers that limit to MAX_QPATH?

slipher:
I don't like this idea to allow some shaders to have a longer name. It's not worth the risk of problems to improve compatibility with one single map. The original version looked good to me.

illwieckz:
I reverted to the simpler check against `MAX_QPATH` and then disabling too-long depth shader we may be able to support, so we don't have to check other parts of the code for now.

src/engine/renderer/tr_local.h
illwieckz: I reverted to the simpler check where this string is always MAX_QPATH long.

src/engine/renderer/tr_local.h
illwieckz: The null terminator was already counted in MAX_QPATH, as a single string requires only one null terminator. Anyway I removed this code.

slipher:
LGTM

PR #780, opened 2023-01-15
tr_model: Fix typo when reading md3 models

DolceTriade:
MD3 models do some extra processing so it copies `name` to `filename`. We were loading `name` which was causing crashes when trying to view md3 models.

Fixes https://github.com/Unvanquished/Unvanquished/issues/2371


slipher:
LGTM

DolceTriade:
To add more color it seems by using `name`, we would load the same model for all the LODs (whereas if we use `filename`, the file won't exist and we would not load any code there), which triggers some mad lod code and causes a crash.

sweet235:
I'm playing this on bunker server right now, with 0.54.0/sync. No crashes anymore.

sweet235 (approved):

PR #781, opened 2023-01-16
PakPath::ListFiles - don't error on empty result

slipher:
FS::PakPath::ListFiles could return a 'no such directory' error. But the pakpath doesn't even have a concept of directories. It was actually just returning an error if the number of files found is zero. So it doesn't make sense to report an error.

Fixes https://github.com/Unvanquished/Unvanquished/issues/2373.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -1713,7 +1713,7 @@ bool DirectoryRange::Advance(std::error_code& err)
 	return InternalAdvance();
 }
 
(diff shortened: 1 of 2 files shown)


illwieckz:
I confirm it fixes:

- https://github.com/Unvanquished/Unvanquished/issues/2373

necessarily-equal (approved):
Sounds good, haven't tested

illwieckz (approved):
LGTM

DolceTriade (approved):

PR #784, opened 2023-01-19
Render all dynamic lights, not a random 25%

slipher:
The dynamic lights are evenly divided into four "layers". Due to the typo of 'return' instead of 'break' (unnoticed since the original tiled renderer commit), only lights assigned to the first layer were being rendered.

This seems to fix https://github.com/Unvanquished/Unvanquished/issues/2351, although it doesn't makes sense. The bug fixed here is that we skip rendering some lights, which should make things dimmer, not brighter. (With the NVidia bug the light is far brighter than what you get from just rendering all lights correctly.) If I change the `return;` to `layer = 9; break;`, which should not change the meaning of the existing code, then the excessive brightness go away. So it seems that we have some sort of compiler bug or optimization quirk with Nvidia drivers.
 

It was not as easy as it sounds to notice 3 out of 4 lights not being rendered because many of our use cases for dynamic lights spawn a lot of small lights rather than a single light. For example, a pulse rifle shot seems to contain 5 lights. So despite the bug you would always see at least some light. 

We probably need to rework some assets before merging this since many more lights may now be displayed. For example, the blaster impact site lights seem excessive now with this fixed.

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -215,13 +215,7 @@ void computeDLights( vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse, vec4 mater
 
       if( idx >= u_numLights )
       {
-#if defined(r_showLightTiles)
-        if (numLights > 0.0)
-        {
-          color = vec4(numLights/(lightsPerLayer*numLayers), numLights/(lightsPerLayer*numLayers), numLights/(lightsPerLayer*numLayers), 1.0);
-        }
-#endif
-        return;
+        break;
       }
 
       computeDLight( idx, P, normal, viewDir, diffuse, material, color );


cu-kai:
I gave this a try on my AMD laptop, and to me, lights seem smoother, more pronounced and I also don't see any of the blockiness reported in https://github.com/Unvanquished/Unvanquished/issues/2343.

I do however now note that some dynamic lights don't appear at all (for example the blaster effect and pulse rifle effects do not always appear).

Here's a screenshot with the client in your branch compiled:
unvanquished_2023-01-19_131634_000

And with an older 0.53 client:
unvanquished_2023-01-19_131643_000


Here is a screenshot showing the blaster effect:
unvanquished_2023-01-19_132026_000

illwieckz:
I also see far less blockiness, but I managed to get some:

[light blockiness](https://dl.illwieckz.net/b/unvanquished/bugs/light-blockiness/unvanquished_2023-01-19_145745_000.jpg)

I guess we see less blockiness because the lights overlap more, and then paint over the blockiness.

cu-kai:
Yes just to correct my previous reply, I saw some blockiness in extreme circumstances (and even saw the black screen bug again) but it is nowhere near as pronounced as previously.

illwieckz:
I edited some particles/trails to adjust the dynamic lights. It looks like dividing by two what I assume to be the intensity does a good job.

- https://github.com/UnvanquishedAssets/res-buildables_src.dpkdir/pull/10
- https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/13

illwieckz:
@slipher what and why is the other code removed? It looks like some debug code but I don't know more about it…

slipher:
> @slipher what and why is the other code removed? It looks like some debug code but I don't know more about it…

The same code is present at the end of the function. There is no longer an early return so it does not need to be duplicated.

Changed the target to 0.54/sync since this drastically changes how some assets are rendered.

illwieckz:
> The same code is present at the end of the function. There is no longer an early return so it does not need to be duplicated.

Oh, so that even means it was well known it was a return… 🤭️

illwieckz (approved):
LGTM PLZ MRG ASAP 🙈️

necessarily-equal:
I confirm it fixes the bug I had with juggernaut lighting.Thanks :smile:

PR #785, opened 2023-01-25
Use pkg-config for curses

necessarily-equal:
This reportedly fixes a compilation error on Gentoo, similar to https://github.com/jonathanklee/ngp/issues/30.


DolceTriade:
Does this have implications on other platforms? Can we do find_package first and then pkg-config  after?

necessarily-equal:
pkg-config is love, pkg-config is life.

I don't know the implications for other platforms, but at least it 
seems CI works.

I know we rely on pkg-config for Unvanquished by the way.

slipher:
It surely breaks Windows. The CI image just has a ton of crap preinstalled on it.

necessarily-equal:
Windows is defined here: https://github.com/DaemonEngine/Daemon/pull/785/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR636

slipher:
You still need to guard `find_package(PkgConfig)` though

PR #790, opened 2023-01-29
Add entity debug field

necessarily-equal:
This will allow for a better implementation of
https://github.com/Unvanquished/Unvanquished/pull/2270

This has [an Unvanquished counterpart too](https://github.com/Unvanquished/Unvanquished/pull/2386), because the intent was to modify entityState_t, and apparently entityState_t has to be kept in sync with playerState_t, and playerState_t lies in Unvanquished. Yep.


DolceTriade:
Companion PR to https://github.com/Unvanquished/Unvanquished/pull/2386, since its unlikely to be merged, let's close as we work to gear up for the 0.55 release

PR #793, opened 2023-02-02
Declare trap_* from CommonProxies.cpp in a single header

slipher:
Declare these functions used in both gamelogics in their rightful header rather than duplicating the definitions in cg_api.h and sg_trapcalls.h.


slipher:
@DolceTriade LGTMed in IRC

PR #799, opened 2023-03-07
cmake: do not abort build on Android

illwieckz:
@cu-kai tried to build `daemon-tty` on an Android phone.

This is a trivial commit to not abort compilation. Maybe more work is needed. This is not tested.

The ANDROID variable is set since CMake 3.7, we require 3.12: https://cmake.org/cmake/help/latest/variable/ANDROID.html

It is assumed the __linux__ define is also set on Android so the code only uses __ANDROID__ for naming the platform.

An additional commit just rewrites some macOS wording following current Apple naming conventions.


illwieckz:
I updated first message.

slipher:
Why should we add a bunch of kludges to build for Android when it has zero prospects of becoming a usable platform?

illwieckz:
> Why should we add a bunch of kludges to build for Android when it has zero prospects of becoming a usable platform?

@cu-kai was hoping to run `daemon-tty` on his Phone, so I tried to make it work! 😅️

Anyway that's interesting because given the very small size of patches to make Unvanquished server run with exe on Android, maybe we can run Unvanquished servers on whatever system that looks like Linux enough on whatever architecture that is not that exotic without much larger change given the server runs exe game code.

Jcmora21 (approved):

PR #800, opened 2023-03-08
common/FileSystem: introduce `FS::DefaultTempPath()`

illwieckz:
Introduce `FS::DefaultTempPath()`
    
Make sure `TMPDIR` is always used when needing a temporary file
outside of Windows.
    
`TMPDIR` is a standard variable, for example it is used by `mktemp`.
Android doesn't give access to `/tmp` but Termux sets `TMPDIR`.

This is needed by:

This fixes this error on systems not allowing to write in `/tmp` but setting `TMPDIR` instead:

> `Warn: Could not bind singleton socket: No such file or directory`

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -2353,6 +2353,18 @@ std::string DefaultHomePath()
 #endif
 }
 
+// Determine path to temporary directory
+#ifndef _WIN32
+std::string DefaultTempPath()
+{
+	const char* tmpDir = getenv("TMPDIR");
+	if (!tmpDir || !tmpDir[0]) {
+		tmpDir = "/tmp";
+	}
+	return tmpDir;
+}
+#endif
+
 #endif // BUILD_VM
 
 #ifdef BUILD_VM
--- a/src/common/FileSystem.h
+++ b/src/common/FileSystem.h
@@ -509,6 +509,9 @@ void Initialize();
 #else
 std::string DefaultBasePath();
 std::string DefaultHomePath();
+#ifndef _WIN32
+std::string DefaultTempPath();
+#endif
 void Initialize(Str::StringRef homePath, Str::StringRef libPath, const std::vector<std::string>& paths);
 #endif
 
--- a/src/engine/framework/ApplicationInternals.h
+++ b/src/engine/framework/ApplicationInternals.h
@@ -88,9 +88,7 @@ public:
         this->traits.defaultHomepath = name;
         free(name);
 #else
-        const char* tmpdir = getenv("TMPDIR");
-        tmpdir = tmpdir ? tmpdir : "/tmp";
-        std::string name = FS::Path::Build(tmpdir, "daemon-test-homepath.XXXXXX");
+        std::string name = FS::Path::Build(FS::DefaultTempPath(), "daemon-test-homepath.XXXXXX");
         // mktempd creates the directory
         if (mkdtemp(&name[0]) == nullptr) {
             Sys::Error("Couldn't create temp dir for test: %s", strerror(errno));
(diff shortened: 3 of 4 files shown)


illwieckz:
I added another commit to remove a C++ reserved `_` prefix in a similar code.

illwieckz:
Reading at the `FileSystem` code, should I use `Path::Build`?

slipher:
Underscore followed by lowercase later is NOT a reserved word

illwieckz:
Well, if underscore followed by lowercase is reserved in global context and underscore followed by uppercase is reserved everywhere, it's probably better to never do any underscore prefix by ourselves, that would reduce mental load.

illwieckz:
I added a commit introducing `FS::DefaultTempPath()` the same way there is already `FS::DefaultBasePath()` and `FS::DefaultHomePath()`.

src/common/FileSystem.cpp
```diff
@@ -2353,6 +2353,18 @@ std::string DefaultHomePath()
 #endif
 }
 
+// Determine path to temporary directory
+std::string DefaultTempPath()
```
slipher, line 2357: This is wrong as a temp directory. You can't make normal files in this path.

You can see one other use of a temp dir in TestApplication. In that place I got a Windows temp directory with the `tempnam` function which figures out the temp dir on its own, so maybe DefaultTempPath is not a suitable abstraction for Windows anyway.

illwieckz: I rewrote the code to not use DefaultTempPath on Windows since it was not using a temp path indeed.

illwieckz: Though maybe I would rename `DefaultTempPath` to `DefaultTempPrefix` or something like that because it doesn't return a directory but a prefix for a file name.

illwieckz:
I actually made a `FS::DefaultTempPath()` that returns the temp directory, it is used in both TestApplication and full featured engine when creating the socket. This function is not defined on Windows for now.

I did more rewrite on `FS::DefaultHomePath()` and moved it to its own PR:

- https://github.com/DaemonEngine/Daemon/pull/803

src/common/FileSystem.cpp
illwieckz: The thread is no longer relevant to the current implementation.

illwieckz:
> Actually I think it's a bad idea to have a function called `DefaultTempPath` in general because it would give the false impression that it's a good idea to use this for temp files.

Well, we have already two places having one exact same need so it's better to factorize that code. Current `master` fails to run on systems not allowing to write to `/tmp` while we already have a code reading `$TMPDIR` but it's not used in the main application because of duplication…

> In fact temp files should almost always go in the homepath;

If something is really a temp file, and not a cache file that can be re-read after restart, there is no reason to write it in home path and there are plenty of reasons to not write it to home file. First one being the risk of leaving garbage in case of crash, second one being that systems may optimize file-system for temp folders : ramdisk, disabled copy-on-write, etc. 

Like the code making use of that `DefaultTempPath` function says, some network-file system may not accept sockets, but this would even be a bad idea to  write on a network file system (even if that worked) a socket that would only makes sense for the host creating it… Though, maybe there are conventions we may look at implementing for writing sockets (what about `/run` ?), but I don't see having a function to get the temp directory is bad, especially when we decide to write a file that should not survive an engine restart.

The current `DefaultTempPath` implementation only returns the current temp directory, and is only implemented for non-Windows.

If we want to implement it for Windows maybe we can just do `dirname(tempnam())` but there is less need for that, and that would just be for the sake of factorizing some code.

slipher:
LGTM

PR #801, opened 2023-03-08
cmake: make mumblelink optional

illwieckz:
This makes possible to build the engine without mumblelink, by setting `USE_MUMBLE` cmake option to `OFF`.

I was too lazy to fix mumblelink on Android, and since it's not required to play the game, why not making possible to disable it on platform that doesn't build that code?

Borrowed from:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,6 +149,11 @@ if(NOT DAEMON_EXTERNAL_APP)
     else()
         option(USE_CURSES "Enable fancy colors in terminal's output" ON)
     endif()
+
+    if (BUILD_CLIENT OR BUILD_TTY_CLIENT)
+        option(USE_MUMBLE "Build Daemon with mumblelink sumpport" ON)
+    endif()
+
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -149,6 +149,11 @@ if(NOT DAEMON_EXTERNAL_APP)
     else()
         option(USE_CURSES "Enable fancy colors in terminal's output" ON)
     endif()
+
+    if (BUILD_CLIENT OR BUILD_TTY_CLIENT)
```
DolceTriade, line 153: Does the TTY client need mumblelink?

DolceTriade (approved):

CMakeLists.txt
illwieckz: I don't know but building TTY currently builds it. Maybe it builds it because we never added any ifdef for it.

But it may make sense that TTY client does mumble link: the same way TTY can be used to do text chat, why not doing voice chat as well… The biggest concern is that the current mumble link is still mostly a stub, whatever the client.

slipher:
Remember to wait at least 24 hours before merging

PR #803, opened 2023-03-09
common/FileSystem: rewrite `FS::DefaultHomePath()`

illwieckz:
Some rewrite:

- Do not check for `HOME` if `XDG_DATA_HOME` is set
- Check for environment variables not only being set but also not empty
- More usage of `Path::Build()`
- Single `return` for all failures
- etc.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -2332,25 +2332,28 @@ std::string DefaultHomePath()
 {
 #ifdef _WIN32
 	wchar_t buffer[MAX_PATH];
-	if (!SUCCEEDED(SHGetFolderPathW(nullptr, CSIDL_PERSONAL, nullptr, SHGFP_TYPE_CURRENT, buffer)))
-		return "";
-	return Str::UTF16To8(buffer) + "\\My Games\\" PRODUCT_NAME;
-#else
+	if (SUCCEEDED(SHGetFolderPathW(nullptr, CSIDL_PERSONAL, nullptr, SHGFP_TYPE_CURRENT, buffer))) {
+		return Path::Build(Path::Build(Str::UTF16To8(buffer), "My Games"), PRODUCT_NAME);
+	}
+#elif defined(__APPLE__)
 	const char* home = getenv("HOME");
-	if (!home)
-		return "";
-#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


src/common/FileSystem.cpp
```diff
@@ -2332,25 +2332,28 @@ std::string DefaultHomePath()
...
+		return Path::Build(Path::Build(Str::UTF16To8(buffer), "My Games"), PRODUCT_NAME);
+	}
+#elif defined(__APPLE__)
 	const char* home = getenv("HOME");
-	if (!home)
-		return "";
-#ifdef __APPLE__
-	return std::string(home) + "/Library/Application Support/" PRODUCT_NAME;
+	if (home && !!home[0]) {
+		return Path::Build(Path::Build(Path::Build(home, "Library"), "Application Support"), PRODUCT_NAME);
```
DolceTriade, line 2341: We should just rewrite Path::Build to use variadic templates. This is hardly easy to read :-/

src/common/FileSystem.cpp
```diff
@@ -2332,25 +2332,28 @@ std::string DefaultHomePath()
...
-
-	return xdgHomePath;
-#endif
+	const char* xdgDataHome_ = getenv("XDG_DATA_HOME");
+	if (xdgDataHome_ && !!xdgDataHome_[0]) {
+		return Path::Build(xdgDataHome_, PRODUCT_NAME_LOWER);
+	}
+	else {
+		const char* home = getenv("HOME");
+		if (home && !!home[0]) {
```
DolceTriade, line 2350: Why do you need the !!. `if (home && home[0])` will work just fine?

src/common/FileSystem.cpp
illwieckz: If you know how to do it, I'll love it! 😀️

src/common/FileSystem.cpp
illwieckz: Well, it's pedantic, `home[0]` is a `char` while `!!home[0]` is a `bool`, but I can remove that.

illwieckz: Done.

DolceTriade:
lgtm

PR #805, opened 2023-03-11
cmake: do not require dependencies on unsupported systems

illwieckz:
Do not require dependencies on unsupported systems.

Display a warning instead.

This allows to build the native binaries, for example to run a server using exe game code, even if there is no provided deps and especially no NativeClient binaries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,11 +270,15 @@ if (NOT NACL)
         set(DEPS_COMPILER default)
     endif()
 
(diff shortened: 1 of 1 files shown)


illwieckz:
This was enough to build and run Android arm engine and server exe game code once some  `ifdef __ANDROID__` had been added for this system:

Maybe this is enough to make possible to build and run the server with exe game code on a load of systems like FreeBSD or Haiku and a load of architectures like Power or riscv64 given the related `ifdef` are added.

DolceTriade:
Seems reasonable. lgtm

PR #806, opened 2023-03-12
cmake: make `-march` and `-mtune` options explicit on GCC-like compilers and other tweaks

illwieckz:
While tweaking `cmake/DaemonFlags.cmake` for Android (see #799), I noticed that part of the cmake code was not very consistent:

- i686 and amd64 were setting `-mtune=generic` but not arm64 and armhf
- arm64 and armhf were setting some `-march` but not amd64 and ai686
- i686 and amd64 were setting `-m64` and `-m32` but that code is actually selected if the compiler already knows to produce 32 or 64 bit build: we now select the architecture based on what the compiler outputs: the existing code would never set `-m64` if the compiler doesn't already produce amd64 binaries without it.

So I thought it would be a good idea to drop `-m64` and `-m32` options.

I also thought it would be a good idea to hardcode options for release builds: predefined `-march` and `-mtune=generic` for every platform.

I also thought that if `-march` and `-mtune` are hardcoded for release builds, we may offer the user to tweak them on his side, so people get the ability to do things like that:

```sh
cmake .. -D'CC_MARCH'='native' -D'CC_MTUNE'='native'
```

This would affect GCC-like compilers (or said otherwise, non-MSVC compilers) like GCC or Clang.

---

cmake: make -march and -mtune options explicit on GCC-like compilers and other tweaks

This affects compilers like GCC, Clang and derivatives.

- remove -m64 and -m32, when this code is selected the compiler already knows how to produce 64-bit or 32-bit code.
- make -march and -mtune options explicit
- add CC_MARCH and CC_MTUNE to let user customize -march and -mtune, for example to set them to "native".


slipher:
Getting rid of -m32 or -m64 is good.

It doesn't seem great to make people find a special cmake option to set the architecture instead of using the generic way to set flags. I would rather have a cmake option that enables all the flags intended to make the binary portable to different machines, and we can set that option in the release script.

illwieckz:
I reworked it:

**cmake: introduce `USE_GENERIC_BUILD` and `USE_EFFICIENT_HARDWARE`**

This affects compilers like GCC, Clang and derivatives.

- `USE_GENERIC_BUILD` enforces generic -march and -mtune options
  to produce portable binaries. This must be disabled if the user wants
  to set custom values like -march=native or other more specific ones.
- `USE_EFFICIENT_HARDWARE` enforces hardware features like SSE, NEON,
  VFP, MCX16, etc.

People wanting to avoid any arch and flag enforcement (to disable them or to override them) would set `-D'USE_GENERIC_BUILD'='OFF'` and `-D'USE_EFFICIENT_HARDWARE'='OFF'`.

For example someone wanting to build a non-portable native build would do:

```sh
cmake .. -D'USE_GENERIC_BUILD'='OFF' -D'CMAKE_C_FLAGS'='-march=native' -D'CMAKE_CXX_FLAGS'='-march=native'
```

Someone wanting to build for an old architecture, for example ARM without VFP or NEON, or an early AMD64 without MCX16:

```sh
cmake .. -D'USE_EFFICIENT_HARDWARE'='OFF'
```

illwieckz:
I added some extra commits:

**cmake: do not require dependencies on unsupported architectures**

Similar to:

- https://github.com/DaemonEngine/Daemon/pull/805

but for architectures

**cmake: use `NACL_ARCH` instead of `ARCH` when setting NaCl options**

**cmake: remove some `TODO` that are actually done.**

illwieckz:
The same way I removed `-m32` or `-m64` which are useless since when we run this cmake code we already know if the compiler produces 32-bit or 64-bit binaries, I also removed `-mfloat-abi=hard` flag since it is expected that when this part of the cmake code is run, the compiler already produces armhf binaries. The architecture is expected to be selected by selecting a compiler.

illwieckz:
I also added a commit to detect armel and armhf compilers. I first put it aside, thinking I would include it in the Android PR (#799), but in fact it's welcome to do it now because it is hardening the cmake code. It makes sure every `if(ARCH STREQUAL armhf)` test in cmake is actually done when using a compiler that produces armhf binaries, and not any other arm ABI.

There is no plan to provide armel binaries in releases, there is no prebuilt `nacl_loader` anyway.

**cmake: add armel architecture for soft float 32-bit ARM ABI**

Mainstream 32-bit ARM Linux distributions use "`hard`" float ABI.
Example: Debian armhf
Some (less used) 32-bit ARM Linux distributions use "`soft`" float ABI.
Example: Debian armel

Android 32-bit ARM distributions use older "`soft`" float ABI,
like Debian armel. This is very likely to maintain compatibility
over a very large range of hardware, especially since VFP was still
optional in Armv7-A which is still widely used.

On Linux, armhf is now preferred over armel on recent hardware,
starting with Armv7-A with VFP. Before that, armel was used on
older hardware up to Armv7-A without VFP, it should still works.
Debian Buster provides both armhf and armel distributions.

By building the software with the compatible "`softfp`" ABI variant it
is allowed to do hardware floats using the VFP while keeping compatibility
with "`soft`" libraries and then armel systems at the cost of some overhead.

Distributions favorising performance will prefer armhf.
Distributions favorising compatibility will prefer armel.

cmake/DaemonFlags.cmake
```diff
@@ -185,26 +185,42 @@ else()
...
     endif()
 
+    option(USE_GENERIC_BUILD "Enforce generic -march and -mtune compiler options" ON)
+    if (USE_GENERIC_BUILD)
+        set_c_cxx_flag("-march=${GENERIC_ARCH}")
+        set_c_cxx_flag("-mtune=generic")
+    endif()
+
+    option(USE_EFFICIENT_HARDWARE "Enforce usage of hardware features like SSE, NEON, VFP, MCX16, etc." ON)
+    if (USE_EFFICIENT_HARDWARE)
```
slipher, line 214: - USE_GENERIC_BUILD=ON USE_EFFICIENT_HARDWARE=OFF is a configuration to build for the absolute minimum specifications of the supported architectures
-  USE_GENERIC_BUILD=ON USE_EFFICIENT_HARDWARE=ON is the release build; support most machines of the supported architectures except some super old ones
- USE_GENERIC_BUILD=OFF USE_EFFICIENT_HARDWARE=ON probably doesn't make sense?

Do I have this right?

cmake/DaemonFlags.cmake
```diff
@@ -185,26 +185,49 @@ else()
...
-        set_c_cxx_flag("-m32")
-        set_c_cxx_flag("-msse2")
-        set_c_cxx_flag("-mtune=generic")
-        try_c_cxx_flag_werror(MFPMATH_SSE "-mfpmath=sse")
-    elseif (ARCH STREQUAL "amd64")
-        set_c_cxx_flag("-m64")
-        set_c_cxx_flag("-mtune=generic")
-        try_c_cxx_flag_werror(MCX16 "-mcx16")
+    if (ARCH STREQUAL "amd64")
+        set(GENERIC_ARCH "x86-64")
```
slipher, line 189: I suggest naming this variable GCC_GENERIC_ARCH, to make it clear that the name is just used for feeding GCC flags

cmake/DaemonFlags.cmake
illwieckz: > - USE_GENERIC_BUILD=ON USE_EFFICIENT_HARDWARE=OFF is a configuration to build for the absolute minimum specifications of the supported architectures
>  - USE_GENERIC_BUILD=ON USE_EFFICIENT_HARDWARE=ON is the release build; support most machines of the supported architectures except some super old ones

Right.

> - USE_GENERIC_BUILD=OFF USE_EFFICIENT_HARDWARE=ON probably doesn't make sense?

I would not say  “doesn't make sense”, but it's very niche yes. I believe it's still good to keep `USE_EFFICIENT_HARDWARE` enabled even if `USE_GENERIC_BUILD` isn't.

We may make `USE_EFFICIENT_HARDWARE` to only be processed (and then listed) if `USE_GENERIC_BUILD` is enabled, but I believe it's a bad idea: disabling `USE_GENERIC_BUILD` may produce some i386 build without SSE or some very old stuff if that's the current compiler default, and the user would have no clue about what's going on because the disabled option would even not be listed as disabled.

I can imagine people disabling `USE_GENERIC_BUILD` by wrongly believing this would magically produce an optimized build, keeping `USE_EFFICIENT_HARDWARE` (which sounds appealing) enabled would make sure such tweaker doesn't disable SSE or things like that without knowing.

cmake/DaemonFlags.cmake
illwieckz: done.

cmake/DaemonFlags.cmake
slipher: Yeah USE_EFFICIENT_HARDWARE sounds kind of like it should do `-march=native`. 

I agree that conditional options are often a bad idea due to hiding information.

Maybe it could be a tri-state option like USE_GENERIC_BUILD=full, USE_GENERIC_BUILD=modern, or USE_GENERIC_BUILD=none

illwieckz: I also thought about defining the state by a word, but I like the fact cmake actually knows what is a bool, and that you can cycle bool values in CMake GUIs.

For a word-based state it would be possible to name the variable `BUILD_PROFILE` and provide the options:

- `release`: lowest hardware + things like SSE
- `lowend`: lowest hardware
- `custom`: enforces nothing

This would make possible other states like:

- `native`:  `-march=native`

But then I'm unhappy with the fact someone selecting `custom` will get SSE not being disabled, this is still hiding information, I prefer to require the user to explicitly disable things like SSE on purpose.

While the word-based state allows to define more states like the `native` one, it would still fail to provide a very basic option to enable advanced features like SSE, something a bool is very efficient at. Technically providing the equivalent of `USE_GENERIC_BUILD=OFF USE_EFFICIENT_HARDWARE=ON` to let the compiler selects the architecture our the user pick a custom architecture while making sure advanced features don't get disabled in the process would require to define yet another state word. In the end people would have to learn our custom states instead of learning about their compiler. We can expect people not doing release builds to learn what their compiler do.

In all cases I like the idea to provide the user the ability to make sure advanced features like SSE don't get disabled without knowing about it, that means without explicitly asking for it, something the bool does very well.

Defining a word-based state system may be reconsidered if there is a need for something like a ready-to-use `native` build and especially if we want it to be portable over systems and definable on other compilers like MSVC. Right now the only use I see would be distros like Gentoo that install packages by building them (and then making binary portability optional) or some service providers that would rent ready-to-use game servers and optimizing them to the maximum for their hardware. Those needs are niche if not existing yet, and I assume those would know how to add a compiler-specific `-march=native` option to their build script.

slipher: OK that's a decent argument for having a bool for those features. We should just have a better name instead of USE_EFFICIENT_HARDWARE. How about DISABLE_MODERN_CPU_FEATURES?

illwieckz: I like the `USE_` prefix, and the fact the release build would require both the features to be ON, which looks less confusing to me. I don't really like the _modern_ word, but mentioning `CPU` directly is a good idea.

Why not:

- `USE_CPU_GENERIC_ARCHITECTURE`
- `USE_CPU_RECOMMENDED_FEATURES`

I also thought about the `ADVANCED` qualifier but then it would be misleading as people may think it would select the latest advanced features like AVX. My concern is exactly that: I don't want to give the user some ability to configure things without letting know he may disable something we recommend.

slipher: RECOMMENDED_FEATURES is better than my idea. LGTM with that

PR #809, opened 2023-03-15
say which file failed when checking socket

bmorel:
Currently when daemon fails to create a socket, it does not says what it tries to do, making it a pain to know which file to clean. This PR gives this information in logs.

I got annoyed by the lack of info about which file to remove after a crash, hence this PR.

### DIFF
--- a/src/engine/framework/System.cpp
+++ b/src/engine/framework/System.cpp
@@ -124,10 +124,10 @@ static void CreateSingletonSocket()
 	addr.sun_family = AF_UNIX;
 	Q_strncpyz(addr.sun_path, singletonSocketPath.c_str(), sizeof(addr.sun_path));
 	if (bind(singletonSocket, reinterpret_cast<struct sockaddr*>(&addr), sizeof(addr)) == -1)
-		Sys::Error("Could not bind singleton socket: %s", strerror(errno));
+		Sys::Error("Could not bind singleton socket at file: \"%s\", error: \"%s\"", singletonSocketPath, strerror(errno) );
 
 	if (listen(singletonSocket, SOMAXCONN) == -1)
-		Sys::Error("Could not listen on singleton socket: %s", strerror(errno));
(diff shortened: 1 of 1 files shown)


src/engine/framework/System.cpp
```diff
@@ -124,7 +124,7 @@ static void CreateSingletonSocket()
 	addr.sun_family = AF_UNIX;
 	Q_strncpyz(addr.sun_path, singletonSocketPath.c_str(), sizeof(addr.sun_path));
 	if (bind(singletonSocket, reinterpret_cast<struct sockaddr*>(&addr), sizeof(addr)) == -1)
-		Sys::Error("Could not bind singleton socket: %s", strerror(errno));
+		Sys::Error("Could not bind singleton socket: %s, %s", strerror(errno), singletonSocketPath.c_str());
```
slipher, line 127: The wording could be better, maybe `Could not bind singleton socket at <file>: <error>`

DolceTriade (approved):

src/engine/framework/System.cpp
```diff
@@ -124,10 +124,10 @@ static void CreateSingletonSocket()
 	addr.sun_family = AF_UNIX;
 	Q_strncpyz(addr.sun_path, singletonSocketPath.c_str(), sizeof(addr.sun_path));
 	if (bind(singletonSocket, reinterpret_cast<struct sockaddr*>(&addr), sizeof(addr)) == -1)
-		Sys::Error("Could not bind singleton socket: %s", strerror(errno));
+		Sys::Error("Could not bind singleton socket at file: \"%s\", error: \"%s\"", singletonSocketPath.c_str(), strerror(errno) );
```
slipher, line 127: Remove the `.c_str()`. Otherwise LGTM

bmorel: Oh, it cab take strings? I'll be happy to oblige!

PR #810, opened 2023-03-21
bump `MAX_IMAGE_ANIMATIONS` to `32`

illwieckz:
Bump `MAX_IMAGE_ANIMATIONS` to `32` (from `24`)

On game side, the particle system has `MAX_PS_SHADER_FRAMES` set to `32`, so let's unify this number.

There may be animations with 25 images around or more, like the new Unvanquished groundfire:

- https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/15

Without this patch, it was expected the engine only loaded `24` images instead of `25`.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1124,7 +1124,7 @@ static inline void glFboSetExt()
 		expression_t rExp;
 	};
 
-#define MAX_IMAGE_ANIMATIONS 24
+#define MAX_IMAGE_ANIMATIONS 32
 
 	enum
 	{


illwieckz:
On its side, the game is allowing `32` images for a single particle which are played one after the other in a similar fashion (see `MAX_PS_SHADER_FRAMES`), so maybe we want to set `animMap` limit to `32` as well.

DolceTriade (approved):

illwieckz:
I finally bumped `MAX_IMAGE_ANIMATIONS` to `32` to unify it with `MAX_PS_SHADER_FRAMES`. For example a single fireball animation can be used both in particles or in materials, so I see no reason to keep one number lower than the other one.

DolceTriade:
lgtm

PR #811, opened 2023-03-22
framework/System: restore socket path

illwieckz:
It looks like that, after many rewrite before merging:

- https://github.com/DaemonEngine/Daemon/pull/800

I failed to notice I actually changed the socket path, changing it from a “`file named X in a directory named Y`” to a “`file named Y`”.

While the merged code works (the socket works), the unwanted socket path modification broke command forwarding between older and newer engines because they don't agree on that socket path.


slipher:
LGTM

PR #812, opened 2023-03-22
implement workarounds for division by zero

illwieckz:
This now looks ready to me. This uses `0` as the result of the division if it would be `0/0`.

See for a similar PR on game side:

- https://github.com/Unvanquished/Unvanquished/pull/2533

---

Original message:

This is a draft and work-in-progress.

I dirtily enabled exceptions for divizion by zeros to catch them, and implemented mindless dirty workarounds for some of them.

This needs more thoughts for better workarounds. For now I divided by `1` any time there was a divizion by zero, but some other cases may require better values, like actually returning `1` for the whole division if we do `0/0`, or returning a specific default value for when it cannot be computed.

That's why this PR is marked as work-in-progress.

### DIFF
--- a/src/engine/qcommon/q_math.cpp
+++ b/src/engine/qcommon/q_math.cpp
@@ -562,8 +562,14 @@ float ProjectPointOntoRectangleOutwards( vec2_t out, const vec2_t point, const v
 	dsign[ 0 ] = ( dir[ 0 ] < 0.0f );
 	dsign[ 1 ] = ( dir[ 1 ] < 0.0f );
 
-	t = ( bounds[ 1 - dsign[ 0 ] ][ 0 ] - point[ 0 ] ) / dir[ 0 ];
-	ty = ( bounds[ 1 - dsign[ 1 ] ][ 1 ] - point[ 1 ] ) / dir[ 1 ];
+	/* Avoids division-by-zero.
+	Valid values are like 1153.3669 or 0.0027. */
+	vec2_t div;
+	div[ 0 ] = dir[ 0 ] ? dir[ 0 ] : 0.0001f;
+	div[ 1 ] = dir[ 1 ] ? dir[ 1 ] : 0.0001f;
+
+	t = ( bounds[ 1 - dsign[ 0 ] ][ 0 ] - point[ 0 ] ) / div[ 0 ];
+	ty = ( bounds[ 1 - dsign[ 1 ] ][ 1 ] - point[ 1 ] ) / div[ 1 ];
 
 	if( ty < t )
 		t = ty;


illwieckz:
Interestingly if I set `minMsec` to `0` when upcaping framerate, I get crazy fps:

[high fps](https://dl.illwieckz.net/b/unvanquished/bugs/high-fps/unvanquished_2023-03-22_161318_000.jpg)

It looks like this code was doing `if 0 < 1; sleep 1ms ` and then waited at least 1ms when framerate was unlimited:

```c++
	msec = com_frameTime - lastTime;

	while ( msec < minMsec )
	{
		//give cycles back to the OS
		Sys::SleepFor(std::chrono::milliseconds(std::min(minMsec - msec, 50)));
```

And with those division by 0 fix patches, things do not glitch:

[high fps](https://dl.illwieckz.net/b/unvanquished/bugs/high-fps/unvanquished_2023-03-22_160748_000.jpg)

slipher:
When I tried this, there were way more places in the renderer with division by zero than just one. I decided against trying to "fix" them as this would just make the code slower and not fix any actual bugs.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -4028,7 +4028,10 @@ void R_LoadLightGrid( lump_t *l )
 		gridPoint1->color[ 0 ] = floatToUnorm8( 0.5f * (ambientColor[ 0 ] + directedColor[ 0 ]) );
 		gridPoint1->color[ 1 ] = floatToUnorm8( 0.5f * (ambientColor[ 1 ] + directedColor[ 1 ]) );
 		gridPoint1->color[ 2 ] = floatToUnorm8( 0.5f * (ambientColor[ 2 ] + directedColor[ 2 ]) );
-		gridPoint1->ambientPart = floatToUnorm8( VectorLength(ambientColor) / (VectorLength(ambientColor) + VectorLength(directedColor)) );
+
+		// avoid division-by-zero
+		float ambientLength = VectorLength(ambientColor);
+		gridPoint1->ambientPart = ambientLength ? floatToUnorm8( ambientLength / ( ambientLength + VectorLength(directedColor) ) ) : 0.0f;
```
bmorel, line 4034: I think using `if` here would be more readable. Something like:

```
float ambientLength = VectorLength(ambientColor);
float colorLength = VectorLength(directedColor);
if ( ambientLength + colorLength )
{
    gridPoint1->ambientPart = floatToUnorm8( ambientLength / ( ambientLength + colorLength ) );
}
else
{
    gridPoint1->ambientPart = 0;
}
```

bmorel (commented):
The modified function is a function that loads a file, so I do not think it would slow down the game much.
I do not know the engine's codebase, so I do not know what the consequence of `gridPoint1->ambientPart = 0;` would be.
If this is safe, then the patch is very likely not hurtful and could be merged.

src/engine/renderer/tr_bsp.cpp
illwieckz: Welp, I made the quick assumption that if the sum of both were `0` (assuming they can't be negative) they would be both `0`. Adding some logger shown me that. Sometime `directedLength` is zero, but anytime I get `ambientColor` as  zero, they are both zero. Though I don't need to make bold assumptions in such code as this code isn't run on every frame and we can bother doing the addition in all case even when the result will be zero:

I rewrote it that way:

```c++
		// avoid division-by-zero
		float ambientLength = VectorLength(ambientColor);
		float vectorLength = VectorLength(directedColor);
		float length = ambientLength + vectorLength;
		gridPoint1->ambientPart = length ? floatToUnorm8( ambientLength / length ) : 0;
```

I'm using `0` as a fallback for the result of the operation because I doubt the value should be set to something else than `0` if all other values are `0`.

For reference this is how looks the log, with `l` the sum of ambient length and directed length, `a` the ambient length and `v` the vector length, `!!!!!!!!!!!!!!!!!!!` was a quick&dirty marker to help me grep the lines:

```
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0 a: 0 v: 0 !!!!!!!!!!!!!!!!!!!
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0271694 a: 0.0271694 v: 0
Warn: l: 0.0543389 a: 0.0271694 v: 0.0271694
Warn: l: 0.0543389 a: 0.0271694 v: 0.0271694
Warn: l: 0.0543389 a: 0.0271694 v: 0.0271694
Warn: l: 0.0543389 a: 0.0271694 v: 0.0271694
Warn: l: 0.0543389 a: 0.0271694 v: 0.0271694
Warn: l: 0.0543389 a: 0.0271694 v: 0.0271694
Warn: l: 0.0815083 a: 0.0271694 v: 0.0543389
Warn: l: 0.0815083 a: 0.0271694 v: 0.0543389
Warn: l: 0.0815083 a: 0.0271694 v: 0.0543389
Warn: l: 0.0815083 a: 0.0271694 v: 0.0543389
Warn: l: 0.108678 a: 0.0271694 v: 0.0815083
Warn: l: 0.108678 a: 0.0271694 v: 0.0815083
Warn: l: 0.108678 a: 0.0271694 v: 0.0815083
Warn: l: 0.118635 a: 0.0271694 v: 0.0914659
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
Warn: l: 0.135847 a: 0.0271694 v: 0.108678
```

illwieckz:
I found another division by zero by reproducing the bug that totally breaks the renderer while going outside of the metro map in some places. By totally breaking the renderer I mean even loading a new map can't fix it, one has to entirely quit the game and restart the engine.

Unfortunately avoiding the division by zero doesn't fix the bug.

```
1900 -1971 100 -40 9
Warn: div: 0.000000
```
```gdb
Thread 1 "daemon" received signal SIGFPE, Arithmetic exception.
0x000055555566dbe4 in SetViewportAndScissor () at Daemon/src/engine/renderer/tr_backend.cpp:758
758			scale = 2.0f / div;

Thread 1 (Thread 0x7ffff59fba40 (LWP 1792738) "daemon"):
#0  0x000055555566dbe4 in SetViewportAndScissor () at Daemon/src/engine/renderer/tr_backend.cpp:758
#1  0x0000555555678cf3 in RB_RenderView (depthPass=false) at Daemon/src/engine/renderer/tr_backend.cpp:4686
#2  DrawViewCommand::ExecuteSelf (this=0x7fffd74ccb94) at Daemon/src/engine/renderer/tr_backend.cpp:5588
#3  0x000055555566da49 in RB_ExecuteRenderCommands (data=<optimized out>) at Daemon/src/engine/renderer/tr_backend.cpp:5832
#4  0x000055555568b752 in RE_EndFrame (frontEndMsec=0x0, backEndMsec=0x0) at Daemon/src/engine/renderer/tr_cmds.cpp:889
#5  0x00005555556161b0 in SCR_UpdateScreen () at Daemon/src/engine/client/cl_scrn.cpp:327
#6  0x00005555556084ed in CL_Frame (msec=msec@entry=1) at Daemon/src/engine/client/cl_main.cpp:2553
#7  0x00005555555a0c5f in Com_Frame () at Daemon/src/engine/qcommon/common.cpp:1031
#8  0x000055555559a37d in main (argc=<optimized out>, argv=0x7fffffffd938) at Daemon/src/engine/framework/System.cpp:762
```
```

**Comment by bmorel** [comment]
@illwieckz  something like [this](https://github.com/Unvanquished/Unvanquished/issues/2447) ?

**Comment by illwieckz** [comment]
I removed a workaround in favor of a proper fix:

- https://github.com/DaemonEngine/Daemon/pull/813

**Comment by illwieckz** [comment]
can I get some approval?

File: src/engine/qcommon/q_math.cpp [hunk 2]
```diff
@@ -549,8 +549,14 @@ float ProjectPointOntoRectangleOutwards( vec2_t out, const vec2_t point, const v
 	dsign[ 0 ] = ( dir[ 0 ] < 0.0f );
 	dsign[ 1 ] = ( dir[ 1 ] < 0.0f );
 
-	t = ( bounds[ 1 - dsign[ 0 ] ][ 0 ] - point[ 0 ] ) / dir[ 0 ];
-	ty = ( bounds[ 1 - dsign[ 1 ] ][ 1 ] - point[ 1 ] ) / dir[ 1 ];
+	/* Avoids division-by-zero.
```
slipher, line 552: I don't like this one. It looks more like sweeping bugs under the rug rather than fixing them

illwieckz: All three callers of this code can feed a legit 0.

slipher:
The beacon ones seem wrong to me. The `clamp_dir` field of cbeacon_t is set only if the calculation of the preferred 2d position of the beacon is off-screen and it needs to be moved to be on the screen (this condition is indicated by `clamped` = true). So it's a bug that this ProjectPointOntoRectangleOutwards is performed for some reason when clamp_dir is not set. If there is no distance from the center of the screen, the operation should have been skipped.

illwieckz:
> The beacon ones seem wrong to me. The `clamp_dir` field of cbeacon_t is set only if the calculation of the preferred 2d position of the beacon is off-screen and it needs to be moved to be on the screen (this condition is indicated by `clamped` = true). So it's a bug that this ProjectPointOntoRectangleOutwards is performed for some reason when clamp_dir is not set. If there is no distance from the center of the screen, the operation should have been skipped.

I caught division by zero with the 2 others occurrence just by being unlucky at looking there or being there. I haven't reproduced with beacon but I haven't tried.

slipher:
2 out of 3 cases are using clamp_dir from the beacon struct. The other one is the minimap. It uses the same kind of logic, where ProjectPointOntoRectangleOutwards is only supposed to be called if the point is outside the rectangle. So it is also a bug. There are no non-buggy cases.

slipher:
I noticed that the usage of ProjectPointOntoRectangleOutwards in cg_beacon.cpp is using the result of CG_WorldToScreen. So your fix for CG_WorldToScreen in the gamelogic zero division PR likely prevents the bug.

illwieckz:
Can I at least merge the first commit?

- `tr_bsp: implement workaround for division by zero`

If I'm rioght it's about a division by zero occurring when loading data from BSP file, by using data from the BSP file.

slipher:
Yeah the BSP lighting one LGTM.

illwieckz:
> Yeah the BSP lighting one LGTM.

I merged this commit.

src/engine/qcommon/q_math.cpp
illwieckz: Maybe the conceptual mistake in such context is to assume a float can be equal to an integer to begin with, not just `0`.

illwieckz:
So, there is one unmerged commit remaining in this branch, this also means the related division by zero is still alive.

Here is the dedicated thread for this specific commit:

- https://github.com/DaemonEngine/Daemon/pull/812/files#r1186484684

src/engine/qcommon/q_math.cpp
illwieckz: The three callers of `ProjectPointOntoRectangleOutwards` are things drawing beacons:

- `src/cgame/cg_beacon.cpp`, `DrawBeacon`:

```c++
static void DrawBeacon( cbeacon_t *b )
{
// …
	// project onto screen
	front = CG_WorldToScreen( b->origin, b->pos, b->pos + 1);

	// CG_WorldToScreen flips the result if behind so correct it
	if( !front )
	{
		b->pos[ 0 ] = 640.0 - b->pos[ 0 ],
		b->pos[ 1 ] = 480.0 - b->pos[ 1 ];
	}

	// virtual 640x480 to real
	b->pos[ 0 ] *= cgs.glconfig.vidWidth / 640.0;
	b->pos[ 1 ] *= cgs.glconfig.vidHeight / 480.0;

	// clamp to edges
	if( !front ||
	    b->pos[ 0 ] < cgs.bc.hudRect[0][0] + b->size/2 ||
	    b->pos[ 0 ] > cgs.bc.hudRect[1][0] - b->size/2 ||
	    b->pos[ 1 ] < cgs.bc.hudRect[0][1] + b->size/2 ||
	    b->pos[ 1 ] > cgs.bc.hudRect[1][1] - b->size/2 )
	{
		vec2_t screen[ 2 ];
		Vector2Set( screen[ 0 ], cgs.bc.hudRect[0][0] + b->size/2,
		                         cgs.bc.hudRect[0][1] + b->size/2 );
		Vector2Set( screen[ 1 ], cgs.bc.hudRect[1][0] - b->size/2,
		                         cgs.bc.hudRect[1][1] - b->size/2 );
		Vector2Subtract( b->pos, cgs.bc.hudCenter, b->clamp_dir );
		ProjectPointOntoRectangleOutwards( b->pos, cgs.bc.hudCenter, b->clamp_dir, (const vec2_t*)screen );
		b->clamped = true;
	}
	else
		b->clamped = false;
}
```

- `src/cgame/cg_draw.cpp`, `CG_DrawBeacon`:

```c++
static void CG_DrawBeacon( cbeacon_t *b )
{
// …
	if( b->type == BCT_TIMER )
	{
		int num;

		num = ( BEACON_TIMER_TIME + b->ctime - cg.time ) / 100;

		if( num > 0 )
		{
			float h, tw;
			const char *p;
			vec2_t pos, dir, rect[ 2 ];
			int i, l, frame;

			h = b->size * 0.4f;
			p = va( "%d", num );
			l = strlen( p );
			tw = h * l;

			if( !b->clamped )
			{
				pos[ 0 ] = b->pos[ 0 ];
				pos[ 1 ] = b->pos[ 1 ] + b->size/2 + h/2;
			}
			else
			{
				rect[ 0 ][ 0 ] = b->pos[ 0 ] - b->size/2 - tw/2;
				rect[ 1 ][ 0 ] = b->pos[ 0 ] + b->size/2 + tw/2;
				rect[ 0 ][ 1 ] = b->pos[ 1 ] - b->size/2 - h/2;
				rect[ 1 ][ 1 ] = b->pos[ 1 ] + b->size/2 + h/2;

				for( i = 0; i < 2; i++ )
					dir[ i ] = - b->clamp_dir[ i ];

				ProjectPointOntoRectangleOutwards( pos, b->pos, dir, (const vec2_t*)rect );
			}

			pos[ 0 ] -= tw/2;
			pos[ 1 ] -= h/2;

			for( i = 0; i < l; i++ )
			{
				if( p[ i ] >= '0' && p[ i ] <= '9' )
					frame = p[ i ] - '0';
				else if( p[ i ] == '-' )
					frame = STAT_MINUS;
				else
					frame = -1;

				if( frame != -1 )
					trap_R_DrawStretchPic( pos[ 0 ], pos[ 1 ], h, h, 0, 0, 1, 1, cgs.media.numberShaders[ frame ] );

				pos[ 0 ] += h;
			}
		}
	}

	trap_R_ClearColor();
}
```

- `src/cgame/cg_minimap.cpp`, `CG_MinimapDrawBeacon`:

```c++
static void CG_MinimapDrawBeacon( const cbeacon_t *b, float size, const vec2_t center, const vec2_t *bounds )
{
	vec2_t offset, pos2d, dir;
	bool clamped;

	size *= b->scale;

	CG_WorldToMinimap( b->origin, offset );
	pos2d[ 0 ] = - size/2 + offset[ 0 ];
	pos2d[ 1 ] = - size/2 + offset[ 1 ];

	if( pos2d[ 0 ] < bounds[ 0 ][ 0 ] ||
	    pos2d[ 0 ] > bounds[ 1 ][ 0 ] ||
	    pos2d[ 1 ] < bounds[ 0 ][ 1 ] ||
	    pos2d[ 1 ] > bounds[ 1 ][ 1 ] )
	{
		clamped = true;

		if( b->type == BCT_TAG )
			return;

		Vector2Subtract( pos2d, center, dir );
		ProjectPointOntoRectangleOutwards( pos2d, center, dir, bounds );
	}
	else
	{
		clamped = false;
	}

	Color::Color color = b->color;
	color.SetAlpha( cgs.bc.minimapAlpha );
	trap_R_SetColor( color );

	trap_R_DrawRotatedPic( pos2d[ 0 ], pos2d[ 1 ], size, size, 0.0f, 0.0f, 1.0f, 1.0f, CG_BeaconIcon( b ), 0.0f );
	if( b->flags & EF_BC_DYING )
		trap_R_DrawStretchPic( pos2d[ 0 ] - size/2 * 0.3f,
		                       pos2d[ 1 ] - size/2 * 0.3f,
		                       size * 1.3, size * 1.3f,
		                       0, 0, 1, 1,
		                       cgs.media.beaconNoTarget );
	if( clamped )
		trap_R_DrawRotatedPic( pos2d[ 0 ] - size * 0.25f,
		                       pos2d[ 1 ] - size * 0.25f,
		                       size * 1.5, size * 1.5f,
		                       0.0f, 0.0f, 1.0f, 1.0f,
		                       cgs.media.beaconIconArrow,
		                       270.0f - atan2f( dir[ 1 ], dir[ 0 ] ) * 180 / M_PI );
}
```

slipher:
Obsoleted by https://github.com/Unvanquished/Unvanquished/pull/3028

PR #813, opened 2023-03-24
A bunch of portal fixes

illwieckz:
### tr_main: do not render broken portal
This may happen if, for example, a portal is seen while being outside the map.

There can be other broken ways to reproduce the bug.

The bug can be reproduced with Spacetracks map 1.1.3 from Unvanquished 0.54 and: `/setviewpos 378 -1906 625 80 1`

This would break the render, even loading another map would not fix the render.

To fix the render, one would have to do: `vid_restart` or move to a place where a portal is properly rendered like this: `/setviewpos 27 -1838 668 147 13`

### tr_main: fix recursive portal

The code already supports recursive portals, the maximum recursion
is configured with  `r_max_portal_levels` cvar (set to `5` by default).

### tr_shader: increase default portalRange

`256` seems too low.

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -179,6 +179,11 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_lockpvs;
 	cvar_t      *r_noportals;
 	cvar_t      *r_portalOnly;
+
+	/* Quake3 used 256 as default portal range. */
+	Cvar::Modified<Cvar::Cvar<int>> r_portalDefaultRange(
+		"r_portalDefaultRange", "Default portal range", Cvar::LATCH, 1024 );
+
 	cvar_t      *r_portalSky;
 	cvar_t      *r_max_portal_levels;
 
--- a/src/engine/renderer/tr_main.cpp
+++ b/src/engine/renderer/tr_main.cpp
@@ -1841,7 +1841,10 @@ static bool R_MirrorViewBySurface(drawSurf_t *drawSurf)
 	if ( tr.viewParms.portalLevel >= r_max_portal_levels->integer &&
 	     tr.viewParms.portalLevel > 0 )
 	{
-		Log::Warn("recursive mirror/portal found");
+		/* Having more than one mirror in a scene is not a bug,
+		we just need to prevent infinite recursion.
+
+		Log::Warn("recursive mirror/portal found"); */
 		return false;
 	}
 
@@ -2071,8 +2074,6 @@ static void R_SortDrawSurfs()
 			{
 				return;
 			}
-
-			break; // only one mirror view at a time
 		}
 	}
 
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -3043,6 +3043,7 @@ static inline void glFboSetExt()
 	extern cvar_t *r_lockpvs;
 	extern cvar_t *r_noportals;
 	extern cvar_t *r_portalOnly;
+	extern Cvar::Modified<Cvar::Cvar<int>> r_portalDefaultRange;
 	extern cvar_t *r_portalSky;
 	extern cvar_t *r_max_portal_levels;
 
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -2796,7 +2796,19 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
 				//stage->type = ST_PORTALMAP;
 				stage->alphaGen = alphaGen_t::AGEN_CONST;
 				stage->constantColor.SetAlpha( 0 );
-				SkipRestOfLine( text );
+
+				token = COM_ParseExt2( text, false );
+
+				/* Until we implement “alphaGen portal” to blend textures on them
+				we are still rendering portals so we need to parse the range. */
+				if ( token[ 0 ] )
+				{
+					shader.portalRange = atof( token );
+				}
+
+				/* Quake3 shaders made the portalRange a requirement, we choose
+				to make it optional and fallback later on the default value or the
+				value that is previously set as “portal <range>” at shader level. */
 			}
 			else
 			{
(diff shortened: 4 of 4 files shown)


illwieckz:
In fact `backEnd.viewParms.portalFrustum[FRUSTUM_NEAR].dist` is set (in `R_SetupFrustum`) _after_ this code is run, so we can't count on it…

illwieckz:
Now the fixes works for real. Portals still work, but broken one doesn't break the render.

I verified with `spacetracks` and `nintendo-land` maps.

illwieckz:
I added a commit: _fix recursive portal_

The code already supports recursive portals, the maximum recursion
is configured with r_max_portal_levels cvar.

This allows for example to have mirror reflecting mirrors. This fixes the bug in spacetracks WC room were only one mirror could be displayed at a time, it also fixes the camera room in pulse or in nova where only one screen was rendered at a time, etc.

illwieckz:
I added a commit: _increase default portalRange_

`256` seems too low.

This fixes the portals that only renders from very close in `mission-control`, `nova`, `pulse`, etc.

For reference, `512` is not enough in `mission-control` to get the portal rendered from the other side of the room.

illwieckz:
Among remaining issues we have about portals:

- portals being rendered above translucent surfaces that exist between portal and player.
- `alphaGen portal` isn't supported, meaning we can't blend textures over portals like the monitors do in `pulse` map.

Those are out of topic of this PR.

illwieckz:
I updated first post with latest additions.

illwieckz:
We now get `Warn: recursive mirror/portal found` when mirrors are reflecting themselves infinitely. In fact this is not a bug, maybe we should silent that?

illwieckz:
nintendo-land, portal from a far:

[portal render](https://dl.illwieckz.net/b/unvanquished/bugs/portal-render/unvanquished_2023-03-24_144638_000.jpg)

mission-control, portal from a far

[portal render](https://dl.illwieckz.net/b/unvanquished/bugs/portal-render/unvanquished_2023-03-24_144709_000.jpg)

pulse, two monitors:

[portal render](https://dl.illwieckz.net/b/unvanquished/bugs/portal-render/unvanquished_2023-03-24_144949_000.jpg)

nova, two monitors from a far:

[portal render](https://dl.illwieckz.net/b/unvanquished/bugs/portal-render/unvanquished_2023-03-24_145205_000.jpg)

spacetracks, two mirrors at the same time:

[portal render](https://dl.illwieckz.net/b/unvanquished/bugs/portal-render/unvanquished_2023-03-24_145346_000.jpg)

bmorel:
Awesome!

> 256 seems too low.

Maybe this can be made a cvar or something? I don't know if this have performance implications... and it would be annoying to need a new release because we find a map for which 1024 qu is too small: it's only 32 meters after all?
Or am I completely mistaken?

illwieckz:
Note to myself: that range bump is likely a workaround. Our engine knows how to read portal range from the portal material keyword, while games like Tremulous were reading that from the `alphaGen portal` material stage keyword, but our engine doesn't support (yet?) `alphaGen portal`. We may still want to read the range from it anyway.

illwieckz:
I implemented the portal range parsing from the `alphaGen portal` material stage keyword (not parsing it was a regression).

I also implemented the default portal range (for when it's not set in the material) as a cvar named `r_portalDefaultRange`. It's a default value: this value will not be used if the material sets an explicit range.

Setting an explicit portal range in a material is welcome because it allows the mapper to prevent the portal from being rendered from too far. It can be used for example if the portal is only seen from inside a room but the inside of the room may still not be excluded from the VIS when being outside of it. Or someone may want to explicitly set a long range portal, larger than the default, for a very big room. Quake 3 actually expected the range to be set in the material or a warning would have been printed.

Because the Dæmon renderer was historically able to read the range at the stage level without warning (as Quake 3 didn't supported that and then materials never did that) and since the Quake3 renderer was historically reading the range at the stage level, there is now no one warning if it's missing there or there: the default value will be used and then the range is now totally optional, including from the `alphaGen portal` stage keyword.

**About the nova map**: Because the code now properly parses the portal range, the monitors on nova map now don't render again unless being very close, this is because those materials actually set a very close range, this is a map bug.

src/engine/renderer/tr_main.cpp
```diff
@@ -1868,10 +1868,7 @@ static bool R_MirrorViewBySurface(drawSurf_t *drawSurf)
 
 	newParms = tr.viewParms;
 
-	if (!R_GetPortalOrientations(drawSurf, &surface, &camera, newParms.pvsOrigin, &newParms.isMirror))
```
slipher, line 1871: I think the "prepare portal command" ought to be skipped here. See #829.

bmorel:
For some reason, I can't reproduce the spacetracks bug with the current unvanquished master, with submodules synced (I think? I'm never entirely sure what git is doing in regard with those...)

illwieckz:
> For some reason, I can't reproduce the spacetracks bug with the current unvanquished master, with submodules synced (I think? I'm never entirely sure what git is doing in regard with those...)

A fix for that specific bug is already merged, here:

- https://github.com/DaemonEngine/Daemon/pull/829

I'll rebase.

src/engine/renderer/tr_init.cpp
```diff
@@ -179,6 +179,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_lockpvs;
 	cvar_t      *r_noportals;
 	cvar_t      *r_portalOnly;
+	cvar_t      *r_portalDefaultRange;
```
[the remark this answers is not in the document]

illwieckz:
So I rebased and turned the cvar into a newstyle one. I also added the `CVAR_LATCH` flags I previously forgot.

src/engine/renderer/tr_init.cpp
```diff
@@ -179,6 +179,11 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_lockpvs;
 	cvar_t      *r_noportals;
 	cvar_t      *r_portalOnly;
+
+	/* Quake3 used 256 as default portal range. */
+	Cvar::Modified<Cvar::Cvar<int>> r_portalDefaultRange(
+		"r_portalDefaultRange", "Default portal range", CVAR_LATCH | CVAR_ARCHIVE, 1024 );
```
slipher, line 185: You have to use `Cvar::LATCH` with new style

Otherwise LGTM

illwieckz: This way?

```c++
	Cvar::Modified<Cvar::Cvar<int>> r_portalDefaultRange(
		"r_portalDefaultRange", "Default portal range", Cvar::LATCH | Cvar::ARCHIVE, 1024 );
```

slipher: Yeah. Although I'm not sure why it needs to be archive :man_shrugging:

illwieckz: I removed the `ARCHIVE` flag.

PR #819, opened 2023-03-30
Make feature disablement more consistent and move some compute at stage configuration

illwieckz:
The PR was split, please review those PR before reviewing this one:

- https://github.com/DaemonEngine/Daemon/pull/820
- https://github.com/DaemonEngine/Daemon/pull/821
- https://github.com/DaemonEngine/Daemon/pull/822
- https://github.com/DaemonEngine/Daemon/pull/823
- https://github.com/DaemonEngine/Daemon/pull/824

Once those other PRs are merged, the remaining commits to review will be:

- `tr_shade: Render_generic() calls Render_generic2D() if 2D`
- `tr_shade,tr_shader: set colorGen and alphaGen only once`

---

_Original message:_

- Make feature disablement more consistent: various combinations of disabled `r_precomputedLight` where wrong.
- Make possible to load deluxe maps even if light mapping is disabled, as deluxe mapping with grid lighting is possible.
- Do not load lightgrid if precomputed lighting is disabled, the same way lightmap loading is already disabled.
- Call `Render_lightMapping()` on every related shader stages because that function knows how to render in all cases.
- `Render_generic()` calls `Render_generic2D()` if 2D, This allows to move the function from `Tess_StageIteratorGeneric()` which means many tests from `Tess_StageIteratorGeneric()` can be moved to shader stage configuration.
- Select shader stage render function only once. Introduce `SetShaderStageRenderers()` and call it at shader stage configuration. Not only factorise the code, but makes sure the Render function as selected once and only once, if possible when finalizing the shader stage, instead of running it on every frame
- Set `colorGen` and `alphaGen` only once. Factorise the code, do it only once per shader stage, and do it when finalizing the shader stage if possible.
- Make `r_precomputedLighting` a cheat cvar. Disabling this cvar disables all precomputed shadows and then makes everything fullbright, which gives an unbalanced advantage by making everything visible.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -446,7 +446,6 @@ static std::vector<std::string> R_LoadExternalLightmaps( const char *mapName )
 R_LoadLightmaps
 ===============
 */
-static const int LIGHTMAP_SIZE = 128;
 static void R_LoadLightmaps( lump_t *l, const char *bspName )
 {
 	tr.worldLightMapping = r_precomputedLighting->integer && !r_vertexLighting->integer;
@@ -591,8 +590,10 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 		// we are about to upload textures
 		R_SyncRenderThread();
 
+		const int internalLightMapSize = 128;
+
 		// create all the lightmaps
-		int numLightmaps = len / ( LIGHTMAP_SIZE * LIGHTMAP_SIZE * 3 );
+		int numLightmaps = len / ( internalLightMapSize * internalLightMapSize * 3 );
 
 		if ( numLightmaps == 1 )
 		{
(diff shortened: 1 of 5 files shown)


illwieckz:
The two biggest points would be summarized that way:

- Disabling `r_precomputedLighting` now works correctly, and it is now a cheat cvar (disables all precomputed lights and then shadows).
- Some code to select the `Render_*` function and sets some `colorGen` and `alphaGen` options are now run only once for the whole game time instead of doing the related computations on every frame. This unifies many `Render_*` function signatures to be the same. It means that on every frame for every shader stage the code now just runs the related function and that's all. The code that is optimized wasn't super heavy on CPU (it was mostly tests) but that design bothered me for a long time, and the rewrite also brings some factorisation and code deduplication, also reducing line count. Less code to read for humans, less code to execute for CPUs.

illwieckz:
I'm splitting the PR.

illwieckz:
Please review those PR before reviewing this one:

- https://github.com/DaemonEngine/Daemon/pull/820
- https://github.com/DaemonEngine/Daemon/pull/821
- https://github.com/DaemonEngine/Daemon/pull/822
- https://github.com/DaemonEngine/Daemon/pull/823
- https://github.com/DaemonEngine/Daemon/pull/824

VReaperV:
It seems that only `tr_shade,tr_shader: set colorGen and alphaGen only once` is relevant for this pr now? Everything else is already merged and `tr_shade: Render_generic() calls Render_generic2D() if 2D` was done elsewhere.

illwieckz:
Yes, and `tr_shade,tr_shader: set colorGen and alphaGen only once` was buggy if my memory is right. I may still investigate that, but we can close that PR.

PR #820, opened 2023-03-31
tr_bsp,tr_shade: rework precomputed light, light map, deluxe map enablement

illwieckz:
Extracted from:

The render is now as expected not only when `r_precomputedLighting` is enabled but also when it is disabled.

Disabling lightmapping while keeping deluxemapping enabled now still loads the deluxemaps to use them with gridlighting.

A variable named `tr.worldLightMapping` is used to know when light mapping is enabled instead of redoing all the tests
with potential mistakes. The is similar to the already existing `tr.worldDeluxeMapping` variable.

`Render_LightMapping()` always knows how to render, there are fallback everywhere.

Rendering with depth fill when there is no lighting is simply wrong, it was probably a very early workaround from before then engine was setting up a fallback lighting.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -448,7 +448,11 @@ R_LoadLightmaps
 */
 static void R_LoadLightmaps( lump_t *l, const char *bspName )
 {
-	if ( r_vertexLighting->integer != 0 )
+	tr.worldLightMapping = r_precomputedLighting->integer && !r_vertexLighting->integer;
+
+	/* All lightmaps will be loaded if either light mapping
+	or deluxe mapping is enabled. */
+	if ( !tr.worldLightMapping && !tr.worldDeluxeMapping )
 	{
 		return;
 	}
@@ -481,6 +485,8 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 			if ( hdrFiles.empty() )
 			{
 				Log::Warn("no lightmap files found");
+				tr.worldLightMapping = false;
+				tr.worldDeluxeMapping = false;
 				return;
 			}
 			std::sort( hdrFiles.begin(), hdrFiles.end(), LightmapNameLess );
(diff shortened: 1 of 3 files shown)


illwieckz:
@DolceTriade did you see this one? 🙂️

illwieckz:
Any comment on this? It actually makes the code better at describing what's enabled or not, and fixes a bug happening when `r_precomputedLighting` is disabled.

Making the code better at describing what's enabled or not is not only making the code looks better, it is also required to actually implement some optimizations based on reliable knowledge of what's enabled or not at a given time.

bmorel:
This does not pass CI apparently.

> fixes a bug happening when r_precomputedLighting is disabled.

Can you tell me how to reproduce the bug? I can't really comment on rendering stuff, but I can at least check if I can reproduce the bug and check the fix.

illwieckz:
> This does not pass CI apparently.

This was just the prior CI failure that affected all PRs, it was unrelated to this branch.

> > fixes a bug happening when r_precomputedLighting is disabled.
> 
> Can you tell me how to reproduce the bug? I can't really comment on rendering stuff, but I can at least check if I can reproduce the bug and check the fix.

The bug can be read in the code itself, the depth data was written as a lightmap:

```diff
 			case stageType_t::ST_COLLAPSE_lighting_PBR:
 				{
-					if ( r_precomputedLighting->integer )
-					{
-						Render_lightMapping( stage );
-					}
-					else
-					{
-						Render_depthFill( stage );
-					}
+					Render_lightMapping( stage );
 					break;
```

You can also set `r_precomputedLighting` to `0` and do `vid_restart`.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -484,7 +484,7 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 			}
 			if ( hdrFiles.empty() )
 			{
-				Log::Warn("no lightmap files found");
+				Log::Warn("no lightmap file found");
```
slipher, line 487: The original is more correct.

illwieckz: Why? Is “none” plural in English?

slipher: `no lightmap file found` suggests that you were only expecting to find one.

illwieckz: Wow, I would have expected that meaning from `lightmap file not found` instead.

slipher: Yes that also would suggest seeking only one

illwieckz: I removed that commit.

PR #821, opened 2023-03-31
tr_bsp: set internalLightMapSize in the block using it

illwieckz:
Extracted from:

Set `internalLightMapSize` in the block using it, a global variable is useless.


DolceTriade (approved):
I guess

PR #822, opened 2023-03-31
tr_bsp: do not load the light grid if precomputed lighting is disabled

illwieckz:
Extracted from:

Do not load the light grid if precomputed lighting is disabled.


DolceTriade (commented):
Is this for vertex lighting?

illwieckz:
> Is this for vertex lighting?

We have three options:

- light mapping, uses lightmap for the world, grid light for the models
- grid lighting, uses gridlight for everything, (for now enabled by `r_vertexLighting` for historical reasons)
- no precomputed light, neither lightmap and lightgrid should be used, that should produce the same result we would get with a map built without lights.

Currently the engine doesn't implement vertex lighting, maybe we could to be visually compatible with some maps that did not computed lightmaps on all surfaces as a trick to save lightmap size… But that's another topic! 😀️

DolceTriade:
lgtm

PR #827, opened 2023-03-31
cl_input: use `Math::Clamp`

illwieckz:
Use `Math::Clamp`.

---
_Original message:_

Use `std::min` and `std::max`.

### DIFF
--- a/src/engine/client/cl_input.cpp
+++ b/src/engine/client/cl_input.cpp
@@ -172,7 +172,6 @@ Returns the fraction of the frame that the key was down
 */
 float CL_KeyState( kbutton_t *key )
 {
-	float val;
 	int   msec;
 
 	msec = key->msec;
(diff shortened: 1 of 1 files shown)


src/engine/client/cl_input.cpp
```diff
@@ -195,17 +195,7 @@ float CL_KeyState( kbutton_t *key )
...
-		val = 0;
-	}
-
-	if ( val > 1 )
-	{
-		val = 1;
-	}
-
-	return val;
+	return std::min( std::max( 0, val ), 1 );
```
bmorel, line 198: why not using Math::Clamp?

illwieckz: Good idea!

illwieckz: Except it's not `static_cast<float>( msec / frame_msec )` but `static_cast<float>( msec ) / frame_msec`! 😉️

```c++
int i=10;
int j=3;
printf("%f\n", (float) i/j);
printf("%f\n", static_cast<float>( i/j ));
printf("%f\n", static_cast<float>( i ) / j );
```
```
3.333333
3.000000
3.333333
```

bmorel (approved):

PR #829, opened 2023-04-02
Fix "bad portal" handling (permanently broke renderer)

slipher:
Alternative version of one of the commits in #813.


illwieckz (approved):
That seems to fix it, thank you!

LGTM.

PR #833, opened 2023-04-06
Fix extra skeleton hands in mirror

slipher:
Fixes https://github.com/Unvanquished/Unvanquished/issues/603.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -3683,8 +3683,8 @@ static void RB_RenderDebugUtils()
 
 		for ( i = 0; i < backEnd.refdef.numEntities; i++, ent++ )
 		{
-			if ( ( ent->e.renderfx & RF_THIRD_PERSON ) &&
-			     backEnd.viewParms.portalLevel == 0 )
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

illwieckz:
Thanks for looking for this!

It removes one skeleton out of three, but maybe that's not the right one? Actually it seems to have removed the one from the first person model, not the one from the wonky “third-person first person model”.

`cg_mirrorgun off`:

[extra skeletons in mirror](https://dl.illwieckz.net/b/daemon/bugs/extra-skeletons-in-mirror/unvanquished_2023-04-07_023156_000.jpg)

`cg_mirrorgun on`:

[extra skeletons in mirror](https://dl.illwieckz.net/b/daemon/bugs/extra-skeletons-in-mirror/unvanquished_2023-04-07_023202_000.jpg)

slipher:
With this patch the skeleton now looks the same viewed through a mirror or directly. I couldn't tell you what a skeleton is supposed to look like, but it looks the same with or without mirrors, so if there is a bug it's not related to mirrors.

slipher:
`cg_mirrorgun` is only referenced from CG_AddViewWeapon, the function that draws the first-person weapon. So it's a cgame bug that it is not applied to the 3rd-person weapon.

PR #837, opened 2023-04-22
Import and rewrite a fix from ioq3

illwieckz:
Original patch by @ineedbots

See:

- https://github.com/ioquake/ioq3/pull/596
- https://github.com/callofduty4x/CoD4x_Server/pull/336

Commit message:

> Fix bad client reliableAcknowledge DOS exploit
>
> Having a reliableAcknowledge of 0x7FFFFFFF causes a massive loop to be executed in SV_UpdateServerCommandsToClient due to the + 1 overflow.

Rewritten from a suggestion by @slipher

### DIFF
--- a/src/engine/server/sv_client.cpp
+++ b/src/engine/server/sv_client.cpp
@@ -1402,7 +1402,7 @@ void SV_ExecuteClientMessage( client_t *cl, msg_t *msg )
 	// NOTE: when the client message is fux0red the acknowledgement numbers
 	// can be out of range, this could cause the server to send thousands of server
 	// commands which the server thinks are not yet acknowledged in SV_UpdateServerCommandsToClient
-	if ( cl->reliableAcknowledge < cl->reliableSequence - MAX_RELIABLE_COMMANDS )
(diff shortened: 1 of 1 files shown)


src/engine/server/sv_client.cpp
```diff
@@ -1402,7 +1402,7 @@ void SV_ExecuteClientMessage( client_t *cl, msg_t *msg )
 	// NOTE: when the client message is fux0red the acknowledgement numbers
 	// can be out of range, this could cause the server to send thousands of server
 	// commands which the server thinks are not yet acknowledged in SV_UpdateServerCommandsToClient
-	if ( cl->reliableAcknowledge < cl->reliableSequence - MAX_RELIABLE_COMMANDS )
```
slipher, line 1405: This patch is kinda poorly written - the subtraction is UB if they send a big negative number. How about `cl->reliableAcknowledge < 0 || cl->reliableAcknowledge < cl->reliableSequence - MAX_RELIABLE_COMMANDS || cl->reliableAcknowledge > cl->reliableSequence` (preventing negative ones as a bonus)

illwieckz: Ok, replaced the patch with one doing what you suggest.

slipher: LGTM

diamante0018:
Glad to see these patches being replicate all across Q3 based engines 🙌

PR #838, opened 2023-04-22
tr_init: r_novis should be latched

illwieckz:
`r_novis` should be latched


slipher:
Why? It doesn't appear to be used during loading.

illwieckz:
Well, I only get tris being drawn from “invisible parts” after a `vid_restart`.

slipher:
Seems it's a little bit persistent between frames as long as you stay in one spot. You have to go to a different room for it to start working.

illwieckz:
Ok then, that was a bit surprising, but it looks like moving around makes the tris updating according to vis and novis, one has to actually quit rooms and re-enters to see any difference.

PR #840, opened 2023-04-26
external_deps: also fix `pnacl-clang --version` on python3

illwieckz:
That's a very minor patch and is not required to build the game.

It makes possible to get pnacl compiler version this way:

```
external_deps/linux-amd64-default_8/pnacl/bin/pnacl-clang --version
```


slipher:
LGTM

PR #841, opened 2023-04-26
 Compiler: introduce `ResetStruct` function, reusable and customizable per compiler

illwieckz:
Introduce `ResetStruct` function, reusable and customizable per compiler.

### DIFF
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -262,4 +262,28 @@ inline int CountTrailingZeroes(unsigned long long x) { int i = 0; while (i < 64 
 	#define UNREACHABLE()
 #endif // UNREACHABLE
 
+/* Use a C++11 braced initializer on MSVC instead of a bracket initializer
+when zeroing a struct. This works around a bug in how MSVC generates implicit
+default constructors. -- Amanieu
+
+The MSVC workaround is known to crash ICC, there is no reason to apply MSVC
+workarounds on non-MSVC compilers. -- illwieckz
+
+I believe GCC, Clang, and MSVC have all been seen to crash during compilation
+on either the () or {} forms of these constructors. -- slipher
+
+The actual occasion for using this function is when you want to clear a struct
+with a very large size. Supposedly with some compiler this could cause a
+temporary copy to be placed on the stack and cause a stack overflow if you did
+it the normal way like foo = {}. -- slipher */
+template<typename T>
+void ResetStruct( T& object ) {
+	object.~T();
+#if defined(_MSC_VER)
+	new( &object ) T{};
+#else
+	new( &object ) T();
+#endif
+}
+
 #endif // COMMON_COMPILER_H_
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -792,8 +792,7 @@ Called before parsing a gamestate
 */
 void CL_ClearState()
 {
-	cl.~clientActive_t();
-	new(&cl) clientActive_t{}; // Using {} instead of () to work around MSVC bug
+	ResetStruct( cl );
 }
 
 /*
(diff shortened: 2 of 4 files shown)


src/common/Compiler.h
```diff
@@ -262,4 +262,19 @@ inline int CountTrailingZeroes(unsigned long long x) { int i = 0; while (i < 64
...
+/* Use a C++11 braced initializer on MSVC instead of a bracket initializer
+when zeroing a struct. This works around a bug in how MSVC generates implicit
+default constructors. -- Amanieu
+
+The MSVC workaround is known to crash ICC, there is no reason to apply MSVC
+workarounds on non-MSVC compilers. -- illwieckz
+
+I believe GCC, Clang, and MSVC have all been seen to crash during compilation
+on either the () or {} forms of these constructors. -- slipher */
+#if defined(_MSC_VER)
```
DolceTriade, line 274: Am I blind, or are these the exact same?

illwieckz: The difference is `T{};` (braces) and `T();` (parenthesis).

DolceTriade: Ah

DolceTriade (approved):

src/common/Compiler.h
```diff
@@ -262,4 +262,19 @@ inline int CountTrailingZeroes(unsigned long long x) { int i = 0; while (i < 64
...
+	#define ZERO_STRUCT( S, T ) do { S.~T(); new( &S ) T{}; } while ( false )
```
slipher, line 275: It should be a template function instead of a macro
```
template<typename T>
void ResetStruct(const T& object) {
... // inside is the same
}
```

illwieckz: I tried that:

```c++
template<typename T>
void ResetStruct(const T& object) {
	object.~T();
#if defined(_MSC_VER)
	new( &object ) T{};
#else
	new( &object ) T();
#endif
}
```

I use it that way:

```c++
void CL_ClearState()
{
	ResetStruct( cl );
}
```

But I get that:

```
In file included from src/engine/client/cl_main.cpp:1:
In file included from src/common/Common.h:35:
In file included from src/engine/qcommon/./q_shared.h:161:
src/common/Compiler.h:282:2: error: no matching 'operator new' function for non-allocating placement new expression; include <new>
        new( &object ) T();
        ^~~
src/engine/client/cl_main.cpp:795:2: note: in instantiation of function template specialization 'ResetStruct<clientActive_t>' requested here
        ResetStruct( cl );
        ^
```

I tried to add `#include <new>` but that fixed nothing.

slipher: Brain fart on my part, there shouldn't be a `const`

illwieckz: Thanks, it seems to work. See my additional commit, if that looks OK to you I'll squash.

src/common/Compiler.h
```diff
@@ -262,4 +262,23 @@ inline int CountTrailingZeroes(unsigned long long x) { int i = 0; while (i < 64
 	#define UNREACHABLE()
 #endif // UNREACHABLE
 
+/* Use a C++11 braced initializer on MSVC instead of a bracket initializer
```
slipher, line 265: I guess if we're going to have a verbose comment we should mention the actual occasion for using this function, which is when you want to clear a struct with a very large size. Supposedly with some compiler this could cause a temporary copy to be placed on the stack and cause a stack overflow if you did it the normal way like `foo = {}`.

slipher (commented):
LGTM

src/common/Compiler.h
illwieckz: Right, this is basically what I got with ICC. I'll add your wording.

illwieckz: Well, except that was ICC doing a stack overflow, not the compiled program…

PR #842, opened 2023-04-26
external_deps: copy `ld-linux-armhf.so.3` instead of renaming it

illwieckz:
Copy `ld-linux-armhf.so.3` instead of renaming it, binaries may still reference `ld-linux-armhf.so.3` as dependency (not only as a loader).

Add a comment explaining why we don't use `patchelf` but do some hacky in-place `sed` rewriting instead.

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -593,7 +593,11 @@ build_naclsdk() {
 	linux-arm64-*)
 		mkdir -p "${PREFIX}/lib-armhf"
 		cp -a pepper_*"/tools/lib/arm_trusted/lib/." "${PREFIX}/lib-armhf/."
-		mv "${PREFIX}/lib-armhf/ld-linux-armhf.so.3" "${PREFIX}/lib-armhf/ld-linux-armhf"
+		# Copy the library loader instead of renaming it because there may still be some
+		# references to ld-linux-armhf.so.3 in binaries.
+		cp "${PREFIX}/lib-armhf/ld-linux-armhf.so.3" "${PREFIX}/lib-armhf/ld-linux-armhf"
+		# We can't use patchelf or 'nacl_helper_bootstrap nacl_loader' will complain:
+		#   bootstrap_helper: nacl_loader: ELF file has unreasonable e_phnum=13
 		sed -e 's|/lib/ld-linux-armhf.so.3|lib-armhf/ld-linux-armhf|' -i "${PREFIX}/nacl_loader"
 		;;
 	esac


external_deps/build.sh
```diff
@@ -593,7 +593,11 @@ build_naclsdk() {
 	linux-arm64-*)
 		mkdir -p "${PREFIX}/lib-armhf"
 		cp -a pepper_*"/tools/lib/arm_trusted/lib/." "${PREFIX}/lib-armhf/."
-		mv "${PREFIX}/lib-armhf/ld-linux-armhf.so.3" "${PREFIX}/lib-armhf/ld-linux-armhf"
+		# Copy the library loader instead of renaming it because there may still be some
+		# references to ld-linux-armhf.so.3 in binaries.
```
slipher, line 597: What do you mean? What kind of references, in which binaries?

illwieckz:
```
linux-arm64-default_8 $ find nacl_loader nacl_helper_bootstrap lib-armhf/*.so* | while read f; do echo "${f}"; readelf -d "${f}" | grep -E 'NEEDED|SONAME'; echo ''; done
nacl_loader
 0x00000001 (NEEDED)                     Shared library: [librt.so.1]
 0x00000001 (NEEDED)                     Shared library: [libpthread.so.0]
 0x00000001 (NEEDED)                     Shared library: [libstdc++.so.6]
 0x00000001 (NEEDED)                     Shared library: [libgcc_s.so.1]
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x00000001 (NEEDED)                     Shared library: [ld-linux-armhf.so.3]

nacl_helper_bootstrap

lib-armhf/ld-linux-armhf.so.3
 0x0000000e (SONAME)                     Library soname: [ld-linux-armhf.so.3]

lib-armhf/libc.so.6
 0x00000001 (NEEDED)                     Shared library: [ld-linux-armhf.so.3]
 0x0000000e (SONAME)                     Library soname: [libc.so.6]

lib-armhf/libgcc_s.so.1
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x0000000e (SONAME)                     Library soname: [libgcc_s.so.1]

lib-armhf/libm.so.6
 0x00000001 (NEEDED)                     Shared library: [ld-linux-armhf.so.3]
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x0000000e (SONAME)                     Library soname: [libm.so.6]

lib-armhf/libpthread.so.0
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x00000001 (NEEDED)                     Shared library: [ld-linux-armhf.so.3]
 0x0000000e (SONAME)                     Library soname: [libpthread.so.0]

lib-armhf/librt.so.1
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x00000001 (NEEDED)                     Shared library: [libpthread.so.0]
 0x0000000e (SONAME)                     Library soname: [librt.so.1]

lib-armhf/libstdc++.so.6
 0x00000001 (NEEDED)                     Shared library: [libm.so.6]
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x00000001 (NEEDED)                     Shared library: [ld-linux-armhf.so.3]
 0x00000001 (NEEDED)                     Shared library: [libgcc_s.so.1]
 0x0000000e (SONAME)                     Library soname: [libstdc++.so.6]
```

I prefer to not assume that doing `./ld-linux-armhf.so.3 <something>` makes sure the library will never been looked for. That was what I assumed first, but some months ago when toying on Android I got some messages of it not being found in some ways, so that gave me the idea we better want to keep the file available with its original name, the same name it is named in the binaries we run. I don't remember precisely the error I met, but that was enough for me to write in my TODO list to keep the file with the original name around.

illwieckz: Also notice how `nacl_loader` still depends on `ld-linux-armhf.so.3` after its loader has been replace by `lib-armhf/ld-linux-armhf`.

```
linux-arm64-default_8 $ file nacl_loader 
nacl_loader: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked,
  interpreter lib-armhf/ld-linux-armhf, for GNU/Linux 2.6.26, 
  BuildID[sha1]=ec9a606fd2f2d2fae9b1a5af9fea8564c3341d15, not stripped
```

```
linux-arm64-default_8 $ readelf -d nacl_loader | grep -E 'NEEDED|SONAME'
 0x00000001 (NEEDED)                     Shared library: [librt.so.1]
 0x00000001 (NEEDED)                     Shared library: [libpthread.so.0]
 0x00000001 (NEEDED)                     Shared library: [libstdc++.so.6]
 0x00000001 (NEEDED)                     Shared library: [libgcc_s.so.1]
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x00000001 (NEEDED)                     Shared library: [ld-linux-armhf.so.3]
```

illwieckz: This is probably useless to run the game, as the loader tells what name it provides, but providing the file with the same name as the one written in binaries and lib is useful, especially by preventing to report false positive errors when checking if all the files have their deps (for example by running `ldd` on them).

So I prefer to merge this even if that looks like putting on both belt and suspenders.

Actually the HACK is to copy the file to another name so the new name can be stored in an existing string we rewrite in `nacl_loader`, the HACK is not to keep the file with its original name.

Issue #846, opened 2023-04-28
Optimizing client side prediction and giving extra 200fps to anyone


illwieckz:
So, I noticed that when I switch a local game with lowest graphics preset from prediction off and on (`cg_nopredict off` on client or `g_synchronousClients 0` on server), performance drops from 1000fps to 500fps.

On an online game using a public server, and ultra graphics preset, disabling prediction gave me 200fps more. I actually reached 500fps on a public game.

It happens that when client-side prediction is enabled, most of the CPU time is spent in `CG_PredictPlayerState`, and in that function, most of the time is spent in `trap_GetUserCmd`.

There is such code in `CG_PredictPlayerState`:

```c++
	for ( cmdNum = current - CMD_BACKUP + 1; cmdNum <= current; cmdNum++ )
	{
		// get the command
		trap_GetUserCmd( cmdNum, &cg_pmove.cmd );
```

This code is calling `trap_GetUserCmd` 63 times per frame… 😱️

In engine, `CMD_BACKUP` is `64`, this value was already `64` at Quake3 source code release time.. Some comment that were already there at the time also said:

> allow a lot of command backups for very fast systems

So, despite the code always using the max, the comment says such max is not for everyone.

My first attempt to save performance without entirely disabling the prediction was to add a cvar that would only fetch a given amount of command backups. And it works. We may still use such cvar in graphics preset to do less prediction on lower one. The good thing with such patch is that it doesn't break engine compatibility.

But, but, but. I assume doing IPC is slow, very slow, compared to just running code directly in CPU cache.

On engine side, the function behind `trap_GetUserCmd` just does that:

```
bool CL_GetUserCmd( int cmdNumber, usercmd_t *ucmd )
{
	// cmds[cmdNumber] is the last properly generated command

	// can't return anything that we haven't created yet
	if ( cmdNumber > cl.cmdNumber )
	{
		Sys::Drop( "CL_GetUserCmd: %i >= %i", cmdNumber, cl.cmdNumber );
	}

	// the usercmd has been overwritten in the wrapping
	// buffer because it is too far out of date
	if ( cmdNumber <= cl.cmdNumber - CMD_BACKUP )
	{
		return false;
	}

	*ucmd = cl.cmds[ cmdNumber & CMD_MASK ];

	return true;
}
```

I see nothing in that function that can eat 200 or 500 fps. But well, IPC is always slower than what can do a code in CPU cache

[...]

PR #847, opened 2023-04-28
WIP: implement `getcmds` to fetch an array of command backups in one call

illwieckz:
See this thread for details about this:

- https://github.com/DaemonEngine/Daemon/issues/846

This branch doesn't build, I don't know how to write the serializer for this.

I believe merging this would break engine compatibility, so we may want to make this branch target a future branch, like `0.55.0/sync`.


illwieckz:
Closing because I'll rename the branch.

PR #848, opened 2023-04-28
cg_api: implement `trap_GetUserCmdArray` to fetch an array of command backups in one call

illwieckz:
Engine sidecar of:

- https://github.com/Unvanquished/Unvanquished/pull/2660

See this thread for details about this:

- https://github.com/DaemonEngine/Daemon/issues/846

This branch doesn't build, I don't know how to write the serializer for this.

I believe merging this would break engine compatibility, so we may want to make this branch target a future branch, like `0.55.0/sync`.

### DIFF
--- a/src/engine/client/cg_api.h
+++ b/src/engine/client/cg_api.h
@@ -36,12 +36,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #define CGAME_API_VERSION 3
 
-#define CMD_BACKUP               64
-#define CMD_MASK                 ( CMD_BACKUP - 1 )
-// allow a lot of command backups for very fast systems
-// multiple commands may be combined into a single packet, so this
-// needs to be larger than PACKET_BACKUP
-
 #define MAX_ENTITIES_IN_SNAPSHOT 512
 
 // snapshots are a view of the server at a given time
@@ -132,6 +126,7 @@ void            trap_GetCurrentSnapshotNumber( int *snapshotNumber, int *serverT
 bool        trap_GetSnapshot( int snapshotNumber, snapshot_t *snapshot );
 int             trap_GetCurrentCmdNumber();
 bool        trap_GetUserCmd( int cmdNumber, usercmd_t *ucmd );
+int trap_GetUserCmdArray( userCmdArray_t& userCmdArray );
 void            trap_SetUserCmdValue( int stateValue, int flags, float sensitivityScale );
 int             trap_Key_GetCatcher();
 void            trap_Key_SetCatcher( int catcher );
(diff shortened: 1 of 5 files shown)


slipher:
Check out #854. It should be even faster by requesting only one usercmd_t per frame, and doesn't require ABI changes.

bmorel:
I believe this issue might need an update since mentioned issue was closed?

illwieckz:
We may still want to implement this even if we already merged a workaround that fixes the performance problem. What this can bring is to make the code simpler by calling a function once at the beginning of the frame processing, instead of calling a function in various places and in many times.

The good thing with the workaround is that it achieves the same performance without break engine compatibility, we may still want to refactor and clean-up the code anyway. The priority is just far more lower.

illwieckz:
This one may be less needed since an alternative code already exists and merged in #854: 

- https://github.com/DaemonEngine/Daemon/pull/854

I like the fact this implementation makes the code less convoluted on game side.

This PR makes only sense if `sizeof( userCmds_t )` is smaller than the kernel buffer, because if the single call to fetch all `usercmd_t` per frame splits the packets, the caching code only fetching one `usrcmd_t` per frame would be faster.

For reference:

```c++
#define CMD_BACKUP 64

#define USERCMD_BUTTONS 16

struct usercmd_t
{
	int         serverTime;
	int         angles[ 3 ];

	signed char forwardmove, rightmove, upmove;
	dtType_t    doubleTap; // Arnout: only 3 bits used

	byte        weapon;
	byte        flags;

	byte        buttons[ USERCMD_BUTTONS / 8 ];
};

using userCmds_t = std::array<usercmd_t, CMD_BACKUP>;
```

illwieckz:
So, `sizeof( userCmds_t )` is `1536 `.

illwieckz:
Actually one advantage of this implementation is that by turning the `array` into a `vector`, one could easily make the amount of cmd backups fetched for prediction very easy to change for the user with a cvar.

illwieckz:
I noticed this:

```c++
void CG_DrawActiveFrame( int serverTime, bool demoPlayback )
{
	// …
	cg.currentCmdNumber = trap_GetCurrentCmdNumber();

	// …

	cg.currentCmd = trap_BatchGetUserCmds( cg.userCmds );
	
	// …
}
```

20240506-082804-000 unvanquished-orbit-trap

I guess that at some point we will be able to merge those two trap calls into a single one.

But really, those trap calls now count for less than 1%, that's very good!

This Orbit screenshot was taken while running a branch featuring all my currently implement batch trap calls.

illwieckz:
In fact `cg.currentCmdNumber` is not used anymore ! We can just drop everything related to it! 🤪️

illwieckz:
Yay!

20240506-084303-000 unvanquished-orbit-trap

PR #851, opened 2023-05-02
Rewrite some iterators in audio code

illwieckz:
- audio: rewrite an iterator in a code requiring performance

That's the reason why I did this branch first, it's to rewrite a loop in a code that has a critical need for performance (it's one of the hot spot of the engine).

- audio: rewrite some iterators for the sake of more c++ism

This commit is just doing similar changes in less or not critical parts of the code and that's mostly bikeshedding, I just took care of them too while I was around.

### DIFF
--- a/src/engine/audio/Audio.cpp
+++ b/src/engine/audio/Audio.cpp
@@ -147,8 +147,8 @@ namespace Audio {
 
         UpdateListenerGain();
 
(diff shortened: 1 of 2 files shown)


slipher:
Copying a shared_ptr is indeed expensive as it incurs thread locking. Good catch.

LGTM

PR #852, opened 2023-05-06
cg_api: implement `trap_R_BatchInPVS` to test if a bunch of entities are in PVS in one call

illwieckz:
Implement `trap_R_inPVSArray` to test array of entities in PVS in one call.

Engine sidecar of:

- https://github.com/Unvanquished/Unvanquished/pull/2679

### DIFF
--- a/src/engine/client/cg_api.h
+++ b/src/engine/client/cg_api.h
@@ -163,6 +163,11 @@ bool        trap_GetEntityToken( char *buffer, int bufferSize );
 std::vector<std::vector<Keyboard::Key>> trap_Key_GetKeysForBinds(int team, const std::vector<std::string>& binds);
 int             trap_Key_GetCharForScancode( int scancode );
 bool        trap_R_inPVS( const vec3_t p1, const vec3_t p2 );
+
+std::vector<bool> trap_R_BatchInPVS(
+	const vec3_t origin,
+	const std::vector<std::array<float, 3>>& posEntities );
+
 bool        trap_R_inPVVS( const vec3_t p1, const vec3_t p2 );
 bool        trap_R_LoadDynamicShader( const char *shadername, const char *shadertext );
 int             trap_R_LightForPoint( vec3_t point, vec3_t ambientLight, vec3_t directedLight, vec3_t lightDir );
--- a/src/engine/client/cl_cgame.cpp
+++ b/src/engine/client/cl_cgame.cpp
@@ -1287,6 +1287,21 @@ void CGameVM::QVMSyscall(int syscallNum, Util::Reader& reader, IPC::Channel& cha
 			});
 			break;
 
+		case CG_R_BATCHINPVS:
+			IPC::HandleMsg<Render::BatchInPVSMsg>(channel, std::move(reader), [this] (
+				const std::array<float, 3>& origin,
+				const std::vector<std::array<float, 3>>& posEntities,
+				std::vector<bool>& inPVS)
+			{
+				inPVS.reserve(posEntities.size());
+
+				for (const auto& posEntity : posEntities)
+				{
+					inPVS.push_back(re.inPVS(origin.data(), posEntity.data()));
+				}
+			});
+			break;
+
 		case CG_R_LIGHTFORPOINT:
 			IPC::HandleMsg<Render::LightForPointMsg>(channel, std::move(reader), [this] (std::array<float, 3> point, std::array<float, 3>& ambient, std::array<float, 3>& directed, std::array<float, 3>& dir, int& res) {
 				res = re.LightForPoint(point.data(), ambient.data(), directed.data(), dir.data());
--- a/src/shared/client/cg_api.cpp
+++ b/src/shared/client/cg_api.cpp
@@ -472,6 +472,17 @@ bool trap_R_inPVS( const vec3_t p1, const vec3_t p2 )
 	return res;
 }
 
+std::vector<bool> trap_R_BatchInPVS(
+	const vec3_t origin,
+	const std::vector<std::array<float, 3>>& posEntities )
+{
+	std::array<float, 3> myOrigin;
+	VectorCopy(origin, myOrigin);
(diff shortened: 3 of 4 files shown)


illwieckz:
I rewrote this using lessons learned while implementing:

- https://github.com/DaemonEngine/Daemon/pull/1114/files

The code now sends and receives the shortest vector possible for valid and active entities only.

src/shared/client/cg_api.cpp
```diff
@@ -460,6 +460,17 @@ bool trap_R_inPVS( const vec3_t p1, const vec3_t p2 )
 	return res;
 }
 
+void trap_R_BatchInPVS(
+	const vec3_t origin,
+	const std::vector<std::array<float, 3>>& posEntities,
+	std::vector<bool>& inPVS )
+{
+	std::array<float, 3> myOrigin;
+	VectorCopy(origin, myOrigin.data());
```
slipher, line 469: The `.data` is not needed.

slipher (commented):
We should ultimately avoid the IPC entirely by implementing inPVS in the cgame, but this is fine for now.

LGTM to the engine side.

src/engine/client/cl_cgame.cpp
```diff
@@ -1248,6 +1248,25 @@ void CGameVM::QVMSyscall(int syscallNum, Util::Reader& reader, IPC::Channel& cha
...
+		case CG_R_BATCHINPVS:
+			IPC::HandleMsg<Render::BatchInPVSMsg>(channel, std::move(reader), [this] (
+				const std::array<float, 3>& origin,
+				const std::vector<std::array<float, 3>>& posEntities,
+				std::vector<bool>& inPVS)
+			{
+
+				size_t numEntities = posEntities.size();
+
+				inPVS.reserve(MAX_LOCATIONS);
```
illwieckz, line 1260: Does it makes sense to also do a `inPVS.reserve(MAX_LOCATIONS);` here as `inPVS` is an argument?

I tried it, the code builds and run, but I don't know if that changes something.

src/shared/client/cg_api.cpp
illwieckz: Removed.

src/engine/client/cl_cgame.cpp
slipher: It doesn't make sense to have both `reserve` and `resize`. You can do either reserve+push_back, or resize+`a[b]=c`.

illwieckz: I rewrote that part, in fact I could iterate the source array directly by using `push_back` for the destination, which is cleaner.

src/engine/client/cg_api.h
```diff
@@ -163,6 +163,12 @@ bool        trap_GetEntityToken( char *buffer, int bufferSize );
 std::vector<std::vector<Keyboard::Key>> trap_Key_GetKeysForBinds(int team, const std::vector<std::string>& binds);
 int             trap_Key_GetCharForScancode( int scancode );
 bool        trap_R_inPVS( const vec3_t p1, const vec3_t p2 );
+
+void trap_R_BatchInPVS(
+	const vec3_t origin,
+	const std::vector<std::array<float, 3>>& posEntities,
+	std::vector<bool>& inPVS );
```
slipher, line 170: How about returning the result?

src/engine/client/cl_cgame.cpp
```diff
@@ -1287,6 +1287,21 @@ void CGameVM::QVMSyscall(int syscallNum, Util::Reader& reader, IPC::Channel& cha
 			});
 			break;
 
+		case CG_R_BATCHINPVS:
+			IPC::HandleMsg<Render::BatchInPVSMsg>(channel, std::move(reader), [this] (
+				const std::array<float, 3>& origin,
+				const std::vector<std::array<float, 3>>& posEntities,
+				std::vector<bool>& inPVS)
+			{
+				inPVS.resize(posEntities.size());
```
slipher, line 1296: Did you mean `reserve`?

illwieckz: Right that's what I meant. Fixed.

What I'm not sure is, how do we are sure the size is always `0` here?

src/engine/client/cg_api.h
illwieckz: What do you mean?

slipher: Return the `vector<bool>` instead of writing into a reference.

illwieckz: Wouldn't this produce an extra copy of the array?

slipher: No. A return value can be automatically moved from a local variable.

src/engine/client/cl_cgame.cpp
slipher: The IPC system will construct an empty vector.

illwieckz: OK! That's good to know.

src/shared/client/cg_api.cpp
```diff
@@ -472,6 +472,17 @@ bool trap_R_inPVS( const vec3_t p1, const vec3_t p2 )
 	return res;
 }
 
+std::vector<bool> trap_R_BatchInPVS(
+	const vec3_t origin,
+	const std::vector<std::array<float, 3>>& posEntities )
+{
+	std::array<float, 3> myOrigin;
+	VectorCopy(origin, myOrigin.data());
```
slipher, line 480: Nit: `.data()` is not needed

slipher (approved):
LGTM

src/shared/client/cg_api.cpp
illwieckz: Right, this was mimicking what other functions do, I'll remove that and maybe we can do further PR to simplify code of other functions.

illwieckz:
> LGTM

Any comment on the game counterpart?

- https://github.com/Unvanquished/Unvanquished/pull/2679

PR #854, opened 2023-05-07
Keep a cache of usercmd_t's in cgame

slipher:
Note that this code lives in the gamelogic, so we can use it immediately without needing a new engine version.

### DIFF
--- a/src/shared/client/cg_api.cpp
+++ b/src/shared/client/cg_api.cpp
@@ -83,10 +83,35 @@ int trap_GetCurrentCmdNumber()
 	return res;
 }
 
+// Use a local cache to reduce IPC traffic. We assume that the user command number never decreases as
+// the command number only resets on an svc_gamestate command, and this command is only sent once
+// per cgame lifetime when the user enters the game (it's not even used for map_restart).
 bool trap_GetUserCmd( int cmdNumber, usercmd_t *ucmd )
 {
+	static usercmd_t cache[ CMD_BACKUP ];
+	static int latestInCache = -1;
+
+	if ( cmdNumber <= latestInCache - CMD_BACKUP )
+	{
+		// Either the cgame is buggy and trying to request really old stuff, or there is some case
+		// of additional calls to CL_ClearState that I didn't know about. Reset the cache in case
+		// it's legit.
+		Log::Warn( "Unexpectedly old command number requested in trap_GetUserCmd" );
+		latestInCache = -1;
+	}
+	else if ( cmdNumber <= latestInCache )
+	{
+		*ucmd = cache[ cmdNumber & CMD_MASK ];
+		return true;
+	}
+
 	bool res;
 	VM::SendMsg<GetUserCmdMsg>(cmdNumber, res, *ucmd);
+	if ( res )
+	{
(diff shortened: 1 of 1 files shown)


illwieckz:
> Check out #854. It should be even faster by requesting only one usercmd_t per frame, and doesn't require ABI changes.

That's an interesting trick. Especially since it can be used right now by mods.

Anyway, something my patch does (it's better seen on [game side](https://github.com/Unvanquished/Unvanquished/pull/2660)), is to entirely remove the call to `trap_GetCurrentCmdNumber`. Would you have a way to cache `trap_GetCurrentCmdNumber` as well?

Notice that even if your patch has the advantage to don't break compatibility, this keeps the crufty design and we may want to clean-up the engine at some point. One good example is that function returns a bool. Actually no one code in game code makes uses of that bool. And if I may image there may have been people wanting that bool return, the need for that bool return is only motivated by the fact the code has to check that your request is properly done. By requesting the whole array once for all, the code is not only simpler, but it totally removes the need for any check, and then totally removes the bool return.

slipher:
OK I made a cache for the number too. https://github.com/Unvanquished/Unvanquished/pull/2684

slipher:
Oh I only read your code just now; I see it uses the same approach for caching the command number.

> Notice that even if your patch has the advantage to don't break compatibility, this keeps the crufty design and we may want to clean-up the engine at some point. One good example is that function returns a bool. Actually no one code in game code makes uses of that bool.

Well if the cgame doesn't do any error checking, so much the worse for it. But then the engine doesn't try too hard to *report* errors, either. Right now we have negative command numbers requested at the beginning of the match, which is an error that goes undetected. But it was convenient to implement the cache the same way as in the engine, so I didn't touch it for the moment.

The current design may be inefficient (when used naively), but I don't see anything crufty about it.

illwieckz:
Yeah, caching the number within a frame is easy to do (and to redo 😁️), but this still means two trapcalls are done per frame: `trap_GetUserCmd` and `trap_GetCurrentCmdNumber`. To only do one trap call I assume we have to add a `trap_GetUserCmdWithNumber` function (and break compat).

illwieckz (approved):
I'm a bit annoyed that this patch not only keeps the convoluted code, but also makes it more convoluted.

But it has the huge benefit of being immediately usable by mods, and the performance boost is expected to be huge, so we better want to merge this as soon as possible.

I think that doesn't close the door for an engine redesign on that part but such redesign can wait once this is merged.

I haven't tested this patch.

LGTM.

bmorel:
this at least would merit an overflow check, since I read "we assume it never decrease" well, yet it can.

slipher:
> this at least would merit an overflow check, since I read "we assume it never decrease" well, yet it can.

The case of the number resetting is already handled correctly UNLESS it resets less than 64 frames from the previous reset. Anyway the worst thing that can happen is wrong prediction.

bmorel:
ok then.

illwieckz:
> the worst thing that can happen is wrong prediction.

And basically a wrong prediction looks like a small lag from my experience at trying to optimizing this (and doing mistakes 😁️), so not a big deal.

Issue #798, opened 2023-02-27
memory leak


ghost:
Since the release, the game no longer frees *lot* of memory (more than 100MB on plat23) when a game ends, ultimately leading to crashes for an allocation failure.

On game's menu screen:

```
  RSS    VSZ COMMAND
 1724  14016 crash_server
209164 1356584 daemon
69724 88153672 nacl_helper_boo
```

Loading thunder:

```
  RSS    VSZ COMMAND
 1724  14016 crash_server
645020 1968716 daemon
```

Re-loading thunder:

```
  RSS    VSZ COMMAND
 1724  14016 crash_server
803548 2125344 daemon
```

One more time:
```
RSS    VSZ COMMAND
 1724  14016 crash_server
943952 2265668 daemon
```

Restarting game:

```
 1724  14016 crash_server
205904 1355984 daemon
69500 88153672 nacl_helper_boo
```

Loading anthill:
```
  RSS    VSZ COMMAND
 1724  14016 crash_server
652864 1918148 daemon
```

Re-loading it:
```
  RSS    VSZ COMMAND
 1724  14016 crash_server
760608 2035192 daemon
```

When trying whith plat23, I had similar RAM increase over time, so I do not think it's related to the map itself. Even test-doors increase RAM usage by similar amount.

@cu-kai found this (I only was wondering why game was more crashy than before, suspecting another driver issue or something along those lines).

When it runs for long enough to saturate memory, here's the error message one gets:
```
/map test-doors  
==== ShutdownGame ==== 
----- Server Initialization ----- 
Map: test-doors 
[FS] Loading pak 'Unvanquished/pkg/unvanquished_0.54.0.dpk'... 
[FS] Loading pak 'Unvanquished/pkg/tex-common_2.3.dpk'... 
[FS] Loading pak 'Unvanquished/pkg/res-players_0.54.dpk'... 
[FS] Loading pak 'Unvanquished/pkg/res-legacy_0.54.dpk'... 
[FS] Loading pak 'Unvanquished/pkg/res-weapons_0.54-dirty.dpk'... 
[FS] Loading pak 'Unvanquished/pkg/res-buildables_0.54.dpk'... 
[FS] Loading pak 'Unvanquished/pkg/res-voices_0.54.dpk'... 
[FS] Loading pak 'Unvanquished/pkg/res-soundtrack_0.54.dpk'... 
[FS] Loading pak 'unvanquished/pkg/map-test-doors_0.1.dpk'... 
[...]

Issue #628, opened 2022-05-26
Mouse capture issue on Mac


slipher:
Steps to reproduce:

1. Launch Unvanquished in a window with a map, like `./daemon -set r_fullscreen 0 -set r_mode 9 +devmap plat23`
2. Click on Unvanquished's icon in the task bar before it finishes loading the map.
3. Leave the cursor outside the window until loading finishes.

On 0.52 with SDL 2.0.12, the bug manifests as following: the cursor is outside the window, but the game responds to mouse movements (changing view direction). The program returns to a normal state upon moving the cursor into the window.

With SDL 2.0.22, the bug is worse. The game does not respond to mouse movements and does not capture the mouse, even if you move the mouse into the window and click on it. To get out of the buggy state you must click on another program and then click on Unvanquished again.

slipher:
It's worse than I thought with 2.0.22. The bug occurs after every cgame restart, even if you keep the mouse in the window. I think I'll hold back SDL to 2.0.12 for the release.

illwieckz:
Good to know… and another argument in favor of recommending our own builds !

slipher:
> The SDL 2.0.22 part would hopefully be solved by #600.

The symptoms of these bugs don't look similar to me :\

illwieckz:
> On 0.52 with SDL 2.0.12, the bug manifests as following: the cursor is outside the window, but the game responds to mouse movements (changing view direction). The program returns to a normal state upon moving the cursor into the window.

For information this bug is also present on Linux.

slipher:
Another way to produce the bug is to start the game like `./daemon -resetconfig -set r_mode 7 -set r_fullscreen 0 +map parpax` with the mouse outside the window when it starts, and then not touch anything until the map loads. I guess the important thing is that when the game loads the app has focus but the mouse is outside of it.

I think the problem is that IN_CenterMouse doesn't work when the cursor is outside the window.

slipher:
[...]

PR #856, opened 2023-05-10
sys: rewrite mouse mode handling

illwieckz:
Rewrite mouse mode handling

- do not grab mouse and respond to mouse move if the window does not have focus
- make `in_mouse == 0` always displays the pointer and never respond to mouse move
- make `in_nograb == 1` disables mouse grabbing and pointer centering
- drop unused `mouseAvailable` and `IN_GetMouseMode`
- do not prevent `r_nograb` in fullscreen

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1518,13 +1518,6 @@ static bool GLimp_StartDriverAndSetMode( int mode, bool fullscreen, bool bordere
 
 	AssertCvarRange( r_displayIndex, 0, numDisplays - 1, true );
 
-	if ( fullscreen && Cvar_VariableIntegerValue( "in_nograb" ) )
-	{
-		logger.Notice("Fullscreen not allowed with in_nograb 1" );
-		r_fullscreen.Set( false );
-		fullscreen = false;
-	}
-
 	rserr_t err = GLimp_SetMode(mode, fullscreen, bordered);
 
 	switch ( err )
@@ -2154,13 +2147,6 @@ void GLimp_HandleCvars()
 		int sdlToggled = false;
 		bool needToToggle = true;
 		bool fullscreen = !!( SDL_GetWindowFlags( window ) & SDL_WINDOW_FULLSCREEN );
-
-		if ( *wantFullscreen && Cvar_VariableIntegerValue( "in_nograb" ) )
-		{
-			logger.Notice("Fullscreen not allowed with in_nograb 1" );
-			*wantFullscreen = false;
-			r_fullscreen.Set( false );
-		}
 
 		// Is the state we want different from the current state?
 		needToToggle = *wantFullscreen != fullscreen;
--- a/src/engine/sys/sdl_input.cpp
+++ b/src/engine/sys/sdl_input.cpp
@@ -48,8 +48,6 @@ static cvar_t       *in_keyboardDebug = nullptr;
 static SDL_Joystick *stick = nullptr;
 static SDL_GameController *gamepad = nullptr;
 
-static bool     mouseAvailable = false;
-
 static cvar_t       *in_mouse = nullptr;
 
 static cvar_t       *in_nograb;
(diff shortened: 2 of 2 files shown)


slipher:
It works OK on Windows with our old SDL (2.0.12).

> The `in_mouse` disablement now works, that's handy when catching crashes in a debugger as the crashed game is kept in memory and the mouse pointer would be kept invisible, making impossible to use mouse in the debugger.
> 
> I don't really know for what `in_nograb` can be useful for, but at least this implementation works.

Well that's what the Quake 3 1.32b release notes claim in_nograb is for too:
> in_nograb - for mod developers. Set to 1 and run windowed, Q3 won’t be grabbing your mouse pointer (useful for debug breaks)

https://discourse.ioquake.org/t/quake-3-changelog-version-history/375

Apparently this is a Linux-specific issue. I've never had any problems with that on Windows and 1.32b is described as an update for Linux clients.

illwieckz:
Actually in this code the difference between `in_mouse` and `in_nograb` is that:

- with `in_nograb 1` the game still responds to mouse move, which is what I remembered from early Unvanquished releases
- with `in_nograb 1` the system cursor still hides while custom cursor is used

While with this `in_mouse 0` implementation, the game doesn't responds to mouse move at all (but menu still responds to mouse hovering), and the system cursor is never hidden.

I would not be against totally removing `in_nograb` anyway.

slipher:
On Mac with SDL 2.0.12 the rewrite makes things more buggy than before. #628 is still there and additionally, during cgame load in windowed mode, the system cursor is always invisible when it is at any location outside the Unvanquished window and outside the taskbar.

illwieckz:
> On Mac with SDL 2.0.12 the rewrite makes things more buggy than before. #628 is still there and additionally, during cgame load in windowed mode, the system cursor is always invisible when it is at any location outside the Unvanquished window and outside the taskbar.

This should be the least of our concern and we should not spend time with that.

Instead, the game should not capture the mouse, not center the cursor, and not hide the cursor when rendering the map loading screen:

- https://github.com/Unvanquished/Unvanquished/pull/2688

slipher:
This definitely causes a regression on Mac. I opened #866 thinking it could happen independently but that was a mistake on my part. This PR causes the system cursor to be invisible outside the window as mentioned above.

illwieckz:
I modified the code to make sure it applies the changes for the default mouse mode on every `IN_Init` call.

In previous iteration of the PR I made sure `IN_Init` always called `IN_SetMouseMode` with the default mouse mode on every `IN_Init` call, but I forgotten that `IN_SetMouseMode` does not applies the changes if the mode doesn't change (and on first `IN_Init` load, the mode obviously did not have previous state and then did not change.

illwieckz:
Now, if it happens that `SDL_WasInit( SDL_INIT_VIDEO )` is not done yet when `IN_SetMouseMode` is called for the first time, we have undefined behavior.

illwieckz:
> Now, if it happens that `SDL_WasInit( SDL_INIT_VIDEO )` is not done yet when `IN_SetMouseMode` is called for the first time, we have undefined behavior.

I now should have taken care of this too.

illwieckz:
Something wrong that is remaining is that the game uses the in-game `Deltas` mode (mouse capture, pointer hidden, move-based input) when connecting to a server and loading a map. I suspect that the game switch from `CustomCursor` to in-game `Deltas` mode when a menu is displayed or not, and server connection screen and map loading screen may not be considered as a menu.

illwieckz:
Now, all the bugs I experienced on Linux look to be fixed.

illwieckz:
Now,

- [ ] does this bring any regression on Windows?
- [ ] does this bring any regression on macOS?

illwieckz:
Once the macOS mouse workaround is merged:

- https://github.com/DaemonEngine/Daemon/pull/871

And once SDL2 deps are updated:

Is there any regression coming from this?

slipher:
This centers the mouse every time the mode changes. It is undesirable to center the mouse when switching between the console and a menu (i.e. between SystemCursor/CustomCursor). The old behavior of centering only when exiting Deltas mode is preferable.

The design w.r.t. unfocused/minimized doesn’t make sense. Currently the system works by having the cgame change the mode when it receives a focus event. There are no bugs with this now AFAIK, but it might make sense to force SystemCursor mode when unfocused to guard against badly-behaved cgames that would try to steal input when minimized or whatever. However, this PR does not accomplish that because a badly-behaved cgame can simply *not send any SetMouseMode messages* on unfocus, and the mode will stay in the wrong state. So if you want to do this, the engine should set the mode to SystemCursor on unfocus and block any attempts from the cgame to change it.

illwieckz:
> This centers the mouse every time the mode changes. It is undesirable to center the mouse when switching between the console and a menu (i.e. between SystemCursor/CustomCursor). The old behavior of centering only when exiting Deltas mode is preferable.

Right, I reverted that.

> The design w.r.t. unfocused/minimized doesn’t make sense […].

I don't get everything of what you say, this code actually fixes the actual bug of the game centering the mouse and reading input while unfocused.

illwieckz:
> the actual bug of the game centering the mouse and reading input while unfocused.

How to reproduce this bug:

- `./daemon +devmap plat23`
- while the map is loading, put a window atop the window game and focus it

At the moment the map loading screen ends, while the game is unfocused, the mouse is centered and the game reads input from mouse movement, moving the camera, this even when you interact with the currently focused window that is not the game. I reproduce this on Linux.

illwieckz:
So what I understand is that when the loading of the map ends and the game requests the engine to grab the mouse, the engine now says _nope_.

illwieckz:
> badly-behaved cgame can simply _not send any SetMouseMode messages_ on unfocus

Our current game manages to request the mouse grab while unfocused, though.

illwieckz:
Not merging this makes debugging the engine memory leak (#798) very painful.

To debug #798, one has to have the game in background, a memory monitor on screen, then load maps after maps, the easiest way to do so being to load maps from the console. But every time the next map is loaded, the mouse is centered by the unfocused background game.

Not merging this also makes profiling the game very painful.

In fact, debugging any bug or profiling anything when keeping the game in background should make it easier to debug just makes it very painful. That's the primary motivation for this PR: remove the pain of debugging or profiling the game when the game is in background.

slipher:
> But every time the next map is loaded, the mouse is centered by the unfocused background game.

Thanks, now I understand. I was able to reproduce on Mac as well.

Check out the [slipher/mouse](https://github.com/DaemonEngine/Daemon/compare/illwieckz/mouse...slipher/mouse?expand=1) branch. I added a couple commits based on yours to make the state handling less convoluted.

illwieckz:
Obsoleted by:

- https://github.com/DaemonEngine/Daemon/pull/922

PR #857, opened 2023-05-10
Updating deps and fixing some bugs

illwieckz:
- bump SDL version,  
  Bugs are likely fixed by https://github.com/DaemonEngine/Daemon/pull/856
- bump OpenAL version, bug is fixed.
- bump PNG, JPEG, WebP, Opus, Nettle, Freetype versions.
- bump Nasm version,  
  this one looks to be pure bikeshedding.
- Bump the external deps archive version number.


illwieckz:
Actually I strongly doubt the MSVC GLEW error is related to this PR because it is one I haven't updated, so that should not be blocking.

illwieckz:
I'm running the build tests on Debian Buster, the same distro used to build the Unvanquished releases.

illwieckz:
I uploaded some preview deps tarball and Windows build fails because of missing `libjpeg.lib` file. That reminds me something, are we able to crosscompile `libjpeg.lib`?

illwieckz:
About the engine failing to build against updated SDL2 on macOS:

- https://github.com/libsdl-org/SDL/issues/7695

illwieckz:
The error when building GLEW for MSVC isn't introduced by this PR:

- https://github.com/DaemonEngine/Daemon/issues/859

illwieckz:
So, SDL2 on MSVC using the "builtin cmake” files do that:

```
MSVCRTD.lib(exe_winmain.obj) : error LNK2019: unresolved external symbol WinMain referenced in function "int __cdecl invoke_main(void)" (?invoke_main@@YAHXZ) [C:\projects\daemon\build\server.vcxproj]
C:\projects\daemon\build\Debug\daemonded.exe : fatal error LNK1120: 1 unresolved externals [C:\projects\daemon\build\server.vcxproj]
```

illwieckz:
It looks like something in my various fixes also fixed the PNG linkage on i686.

illwieckz:
I extended the Azure CI to also build using MingW, as our release builds are done with Mingw, we may want to catch regressions on it too.

illwieckz:
So, it looks like everything builds.

This is actually the first time I produce working MSVC deps. This required no manual task from me. That was the last part I was not able to do before. And as « I », I also mean, « any random people ».

So, it looks like this PR no only updates the libraries, it also makes the deps building and packaging doable by anyone.

I did the macOS packages on macOS Mojave, all the other ones in a Debian Buster chroot.

At some point we may want to generate them on docker like we do with the engine release.

illwieckz:
I split the PR to make easy to review patches.

All the patches that are not updating the libraries, but may be required to update libraries, or fixing the script to begin with, are there:

- https://github.com/DaemonEngine/Daemon/pull/861
- https://github.com/DaemonEngine/Daemon/pull/862
- https://github.com/DaemonEngine/Daemon/pull/863
- https://github.com/DaemonEngine/Daemon/pull/864

There is also that:

- https://github.com/DaemonEngine/Daemon/pull/860

All those patches are also living in this PR, because they are needed by this PR. Once those other PRs are merged, they will disappear from there in a way only the patches updating the libraries will remain.

illwieckz:
I noticed GLEW installs 32-bit libs in `lib64/`. This works but we may want to clean-up that.

slipher:
It seems that nothing is working to fix Mac mouse bugs for the moment, but maybe the SDL cmake stuff could also be split off into another PR so that people can freely experiment with different versions.

slipher:
As described in the latest #628 post, a new SDL actually enables us to fix a bug on Mac. So I'm good with new SDL.

For jpeg I don't think you should need to patch CMakeLists.txt.

> When cross-compiling, a [CMAKE_TOOLCHAIN_FILE](https://cmake.org/cmake/help/latest/variable/CMAKE_TOOLCHAIN_FILE.html#variable:CMAKE_TOOLCHAIN_FILE) should set the CMAKE_SYSTEM_PROCESSOR variable to match target architecture that it specifies (via [CMAKE_\<LANG>_COMPILER](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html#variable:CMAKE_%3CLANG%3E_COMPILER) and perhaps [CMAKE_\<LANG>_COMPILER_TARGET](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_TARGET.html#variable:CMAKE_%3CLANG%3E_COMPILER_TARGET)).

-- https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html

So for targets that don't already have a toolchain file (i.e. non-Windows), you can add `-DCMAKE_SYSTEM_PROCESSOR=<something appropriate>` to the CMake command line.

illwieckz:
> As described in the latest #628 post, a new SDL actually enables us to fix a bug on Mac. So I'm good with new SDL.

Yay! 🎉️

> For jpeg I don't think you should need to patch CMakeLists.txt.
> 
> > When cross-compiling, a [CMAKE_TOOLCHAIN_FILE](https://cmake.org/cmake/help/latest/variable/CMAKE_TOOLCHAIN_FILE.html#variable:CMAKE_TOOLCHAIN_FILE) should set the CMAKE_SYSTEM_PROCESSOR variable to match target architecture that it specifies (via [CMAKE_<LANG>_COMPILER](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html#variable:CMAKE_%3CLANG%3E_COMPILER) and perhaps [CMAKE_<LANG>_COMPILER_TARGET](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_TARGET.html#variable:CMAKE_%3CLANG%3E_COMPILER_TARGET)).
> 
> -- https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html
> 
> So for targets that don't already have a toolchain file (i.e. non-Windows), you can add `-DCMAKE_SYSTEM_PROCESSOR=<something appropriate>` to the CMake command line.

Yes but no. I tried that, with strong hope. It doesn't work. 

It's probably a cmake bug, but it doesn't work at all, neither on cmake from Debian Buster, neither on cmake from Ubuntu 22.10. Cmake actually rewrites `CMAKE_SYSTEM_PROCESSOR` with system host processor in all case, whatever the toolchain file.

slipher:
Apparently there is a [cross-compiling bit](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html) that determines whether CMake tries to detect the target architecture from the host system and it is flipped by setting CMAKE_SYSTEM_NAME. So if you add `set(CMAKE_SYSTEM_NAME foobar)` to toolchain.cmake then your example works.

illwieckz:
Oooh, this seems to work, without a toolchain file, thanks!

```
$ cmake -S. -Bbuild -DCMAKE_SYSTEM_NAME=ON -DCMAKE_SYSTEM_PROCESSOR=aarch64
-- The C compiler identification is GNU 12.2.0
System is unknown to cmake, create:
Platform/ON to use this system, please post your config file on discourse.cmake.org so it can be added to cmake
-- Detecting C compiler ABI info
System is unknown to cmake, create:
Platform/ON to use this system, please post your config file on discourse.cmake.org so it can be added to cmake
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
[CMakeLists.txt] CMAKE_SYSTEM_PROCESSOR: aarch64
-- Configuring done
-- Generating done
-- Build files have been written to: build
```

Well, at least it works with a simple cmake file. The libjpegturbo ones also does many other things and other wrong assumptions in other places, so let's try it!

illwieckz:
For reference I tried to do `cmake -S. -Bbuild -DCMAKE_CROSSCOMPILING=ON -DCMAKE_SYSTEM_PROCESSOR=aarch64` and it didn't worked.

illwieckz:
Actually it also work if the system name is the same as the host, it just has to be set to enable the mechanism:

```
$ cmake -S. -Bbuild -DCMAKE_SYSTEM_PROCESSOR=aarch64
-- The C compiler identification is GNU 12.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
[CMakeLists.txt] CMAKE_SYSTEM_PROCESSOR: x86_64
[CMakeLists.txt] CMAKE_SYSTEM_NAME: Linux
-- Configuring done
-- Generating done
-- Build files have been written to: build

$ rm -r build

$ cmake -S. -Bbuild -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_PROCESSOR=aarch64
-- The C compiler identification is GNU 12.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
[CMakeLists.txt] CMAKE_SYSTEM_PROCESSOR: aarch64
[CMakeLists.txt] CMAKE_SYSTEM_NAME: Linux
-- Configuring done
-- Generating done
-- Build files have been written to: build
```

slipher:
Is this ready? Can we get the JPEG stuff squashed, and all of the version bumps which didn't require any changes squashed?

illwieckz:
Yes it's ready, I'll squash things and rebuild the deps archives and merge this once the checks on rebuilt archives.

PR #860, opened 2023-05-12
azure-pipelines: add Mingw build

illwieckz:
Stolen from:

This adds Mingw build in Azure Pipelines CI.

Since our official release builds for Windows are built on Linux using Mingw, we may also want to make sure such build scenario doesn't break.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,19 +21,33 @@ jobs:
         GCC:
           C_COMPILER: gcc
           CXX_COMPILER: g++
+          EXTRA_PACKAGES:
+          EXTRA_INSTALLS:
+          TOOLCHAIN_FILE:
         Clang:
           C_COMPILER: clang
           CXX_COMPILER: clang++
+          EXTRA_PACKAGES:
+          EXTRA_INSTALLS:
+          TOOLCHAIN_FILE:
+        Mingw:
+          C_COMPILER: x86_64-w64-mingw32-gcc
+          CXX_COMPILER: x86_64-w64-mingw32-g++
+          EXTRA_PACKAGES: gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 mingw-w64-x86-64-dev
+          EXTRA_INSTALLS: sudo update-alternatives --set x86_64-w64-mingw32-gcc /usr/bin/x86_64-w64-mingw32-gcc-posix ; sudo update-alternatives --set x86_64-w64-mingw32-g++ /usr/bin/x86_64-w64-mingw32-g++-posix
+          TOOLCHAIN_FILE: cmake/cross-toolchain-mingw64.cmake
+
     steps:
     - bash: |
         set -e
         sudo apt-get update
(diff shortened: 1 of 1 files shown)


slipher:
We don't have to use `update-alternatives` anymore. See https://wiki.unvanquished.net/wiki/MinGW#Installation_2 (we're using Ubuntu 20.04 which is based on Bullseye)

illwieckz:
> We don't have to use `update-alternatives` anymore. See https://wiki.unvanquished.net/wiki/MinGW#Installation_2 (we're using Ubuntu 20.04 which is based on Bullseye)

If I don't do that I get:

```
mutex in namespace std does not name a type
```

illwieckz:
Ubuntu 22.10:

```
[ 36%] Built target client-objects-pch
In file included from daemon/src/engine/framework/ConsoleField.h:34,
                 from daemon/src/engine/framework/ConsoleField.cpp:32:
daemon/src/engine/framework/ConsoleHistory.h:58:21: error: ‘mutex’ in namespace ‘std’ does not name a type
   58 |         static std::mutex lines_mutex;
      |                     ^~~~~
daemon/src/engine/framework/ConsoleHistory.h:36:1: note: ‘std::mutex’ is defined in header ‘<mutex>’; did you forget to ‘#include <mutex>’?
   35 | #include <vector>
  +++ |+#include <mutex>
   36 | 
```

Well, maybe we should do that include?

illwieckz:
Uh no, this file already includes mutex:

```
#include <mutex>
#include <vector>
```

slipher:
> > We don't have to use `update-alternatives` anymore. See https://wiki.unvanquished.net/wiki/MinGW#Installation_2 (we're using Ubuntu 20.04 which is based on Bullseye)
> 
> If I don't do that I get:
> 
> ```
> mutex in namespace std does not name a type
> ```

I found your [build](https://dev.azure.com/UnvanquishedDevelopment/Daemon/_build/results?buildId=5925&view=results) and you only deleted the update-alternatives part. There's a second thing needed: to use the narrower APT packages as shown at https://wiki.unvanquished.net/wiki/MinGW#Installation_2

illwieckz:
This doesn't work.

slipher:
You put the `-posix` prefix on the compiler but the important place for it to go is the *package name*. (Maybe putting it on the toolchain binaries would work in principle but you'd probably need to modify the toolchain.cmake file.)

> APT (Debian Bullseye or newer):
    sudo apt install gcc-mingw-w64-i686-posix (32-bit, posix, sjlj)
    sudo apt install gcc-mingw-w64-x86-64-posix (64-bit, posix, SEH)

illwieckz:
The package doesn't exist yet in 20.04 focal:

- https://packages.ubuntu.com/search?suite=focal&searchon=names&keywords=gcc-mingw-w64-x86-64-posix

slipher:
I swear I looked at the Ubuntu package list and checked that it was there before my first comment. Maybe I ended up on the wrong Ubuntu version somehow. LGTM to the original version

PR #861, opened 2023-05-12
external_deps: improvements

illwieckz:
Patches stolen from:

Those ones do not update libraries, they improve the script.

Some of those changes look a bit like bikeshedding, but others make the script more robust and some may fixes bugs.


illwieckz:
Oh lol, I wanted to cherry-pick an obvious commit that would make easy to split the PR to ease the rewiew but I used `merge` instead of `cherry-pick`.

Well, all of this is obvious.

PR #862, opened 2023-05-12
external_deps: some help rewording

illwieckz:
Patches stolen from:

This is just about rewording some help message.

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -881,7 +881,7 @@ if [ "${#}" -lt "2" ]; then
 	  package - create a zip/tarball of the dependencies so they can be distributed
 	  wipe - remove products of build process, excepting download cache but INCLUDING installed files. Must be last
 
-	Packages requires for each platform:
+	Packages required for each platform:
 
 	Native Windows compile:
 	  pkgconfig zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis opus opusfile lua naclsdk naclports genlib


external_deps/build.sh
```diff
@@ -846,7 +846,7 @@ if [ "${#}" -lt "2" ]; then
 	  package - create a zip/tarball of the dependencies so they can be distributed
 	  wipe - remove products of build process, excepting download cache but INCLUDING installed files. Must be last
 
-	Packages requires for each platform:
+	Packages required by each platform:
```
slipher, line 849: `required for` is better

external_deps/build.sh
```diff
@@ -863,6 +863,9 @@ if [ "${#}" -lt "2" ]; then
 	Linux arm64 and armhf native compile:
 	  naclsdk (and possibly others depending on what packages your distribution provides)
 
+	Packages also required by Linux platforms when building a release:
```
slipher, line 866: I'd say something like "Packages required by the Unvanquished release script for Linux" and list all of them (without subtracting nacl)

illwieckz: > (without subtracting nacl)

The arm linux packages don't ship naclports and naclports is not used by the Unvanquished release script.

That's why I used the wording `also required by Linux platforms`

external_deps/build.sh
illwieckz: fixed.

external_deps/build.sh
slipher: I mean that naclsdk should be included. The runtime binaries are required.

I think the Unvanquished release script should be mentioned specifically because this is just a random set of packages where using the one from the Debian repositories wasn't good for whatever reason. If someone made a release script for a different platform the list could be different.

illwieckz: This sentence will be totally rewritten the day we are able to build the release for all platforms with the same libraries. I removed it, there is no need to spend time on that.

PR #863, opened 2023-05-12
external_deps: set recommended architecture and build options like the engine

illwieckz:
Patch stolen from:

external_deps: set recommended architecture and build options like the engine


slipher:
LGTM

PR #864, opened 2023-05-12
external_deps: improvements around NASM support

illwieckz:
Patches stolen from:

This makes easier to update libraries requiring NASM.

Actually the `NASM` environment variable used by libjpeg doesn't have the same name in this or that libjpeg (newer versions use another name), and this is specific to libjpeg. Relying on `PATH` mechanism is more reliable, future-proof and library-independent.

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -278,8 +278,12 @@ build_png() {
 build_jpeg() {
 	download "libjpeg-turbo-${JPEG_VERSION}.tar.gz" "https://downloads.sourceforge.net/project/libjpeg-turbo/${JPEG_VERSION}/libjpeg-turbo-${JPEG_VERSION}.tar.gz" jpeg
 
-	# Ensure NASM is available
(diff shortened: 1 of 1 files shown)


slipher:
> PNG also requires Nasm 

Really? The only assembler file I could find is `arm/filter_neon.S` which doesn't match the platforms you tested. The string `nasm` doesn't appear anywhere in the source.

illwieckz:
> > PNG also requires Nasm
> 
> Really? The only assembler file I could find is `arm/filter_neon.S` which doesn't match the platforms you tested. The string `nasm` doesn't appear anywhere in the source.

Hmm, weird, I just built PNG for linux, windows and macos without NASM. That was probably me misleading an issue with JPEG (which is fixed by the prefix).

illwieckz:
I removed that commit.

illwieckz:
This code works with both version 8 and upcoming version 9 of the deps, unlike the original one.

PR #869, opened 2023-05-14
cmake: detect c++14 a better way

### LINKED ISSUE #868: Other ways to detect C++14?
Right now we detect if C++14 is supported by doing this in `DaemonFlags.cmake`:

```cmake
    # C++14 support
    try_cxx_flag(GNUXX14 "-std=gnu++14")
    if (NOT FLAG_GNUXX14)
        try_cxx_flag(GNUXX1Y "-std=gnu++1y")
        if (NOT FLAG_GNUXX1Y)
            message(FATAL_ERROR "C++14 not supported by compiler")
        endif()
    endif()
```

I already have found 3 compilers that fail at this test but are perfectly able to compile the engine and the game if I ignore the error:

- The Android Termux clang, clang 15.0.7, should totally support C++14
- Zig 0.11.0-dev, based on clang 16.0.1, should totally support C++14
- ICC 2021.9.0 20230302

Is there other ways to detect C++14 compatibility?

illwieckz:
Fix #868:

- https://github.com/DaemonEngine/Daemon/issues/868

It works with ICC and Zig, I haven't tried Android Termux Clang.


slipher:
LGTM

PR #870, opened 2023-05-14
Use pkg-config for curses if find_package doesn't work

illwieckz:
@afontain, does this fix the issue too, or is the issue always happen if `find_package` is used?

This is an alternate implementation for:

- https://github.com/DaemonEngine/Daemon/pull/785

---
Original message:

This reportedly fixes a compilation error on Gentoo, similar to https://github.com/jonathanklee/ngp/issues/30.


illwieckz:
I mean @necessarily-equal

CMakeLists.txt
```diff
@@ -644,9 +644,14 @@ if (BUILD_CLIENT OR BUILD_TTY_CLIENT OR BUILD_SERVER OR BUILD_DUMMY_APP)
         else ()
             add_definitions(-DUSE_CURSES)
             set(CURSES_NEED_NCURSES 1) # Tells FindCurses that ncurses is required
-            find_package(CursesW REQUIRED)
-            set(LIBS_ENGINE_BASE ${LIBS_ENGINE_BASE} ${CURSESW_LIBRARIES})
-            include_directories(${CURSESW_INCLUDE_DIR})
+            find_package(CursesW)
+            if (CursesW_FOUND)
+                include_directories(${CURSESW_INCLUDE_DIR})
```
slipher, line 649: The update to LIBS_ENGINE_BASE is missing in the first branch and there's no include_directories in the second branch.

necessarily-equal:
@illwieckz: I unfortunately have no mean of testing if this is fixed, 
as I don't use gentoo.

PR #871, opened 2023-05-15
Fix disappearing cursor during load on Mac

### LINKED ISSUE #866: [Mac] Invisible system cursor during load
I started seeing this bug after updating the Mac OS to Ventura 13.3.1. (I don't remember what it was on before, probably 12.something?) During the loading screen in windowed mode, the system cursor is invisible everywhere except in the Unvanquished window and in the taskbar.

Curiously this is the same bug I saw while testing #856 on the previous OS version. Maybe there's a chance that the bug was already there in the previous OS version and I just didn't notice it... seems unlikely though as the testing involved clicking on stuff during loading and that's noticeably more difficult when the cursor is invisible. Anyway I'm sure it wasn't there on whichever version I had when reporting #628 as I did extensive Mac testing then.

slipher:
When combined with a call to trap_SetMouseMode(MouseMode::SystemCursor) during CG_Init, this fixes #866.

### DIFF
--- a/src/engine/sys/sdl_input.cpp
+++ b/src/engine/sys/sdl_input.cpp
@@ -437,6 +437,10 @@ void IN_SetMouseMode(MouseMode newMode)
 		}
 		mouse_mode = newMode;
 	}
+#ifdef __APPLE__
+	// This prevents the system cursor from disappearing in the loading screen
+	SDL_PumpEvents();
+#endif
 }
 
 static bool in_focus = false;


illwieckz (approved):
LGTM.

slipher:
When combined with updating to a newer SDL, this also fixes #628.

slipher:
The other needed component to fix the disappearing cursor was done in https://github.com/Unvanquished/Unvanquished/pull/2695.

PR #874, opened 2023-05-19
external_deps: some rewrites

illwieckz:
- more error detection
- set some variables early
- whitespace

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 
-# Exit on error
-# Error on undefined variable
-set -e
-set -u
+# Exit on undefined variable and error.
+set -u -e -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+WORK_DIR="${PWD}"
 
 # This should match the DEPS_VERSION in CMakeLists.txt.
 # This is mostly to ensure the path the files end up at if you build deps yourself
@@ -762,8 +763,6 @@ common_setup() {
 	HOST="${2}"
 	"common_setup_${1}"
 	common_setup_arch
-	WORK_DIR="${PWD}"
-	SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
 	DOWNLOAD_DIR="${WORK_DIR}/download_cache"
 	PKG_BASEDIR="${PLATFORM}_${DEPS_VERSION}"
 	BUILD_BASEDIR="build-${PKG_BASEDIR}"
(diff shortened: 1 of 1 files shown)


external_deps/build.sh
```diff
@@ -1,9 +1,10 @@
...
 
-# Exit on error
-# Error on undefined variable
-set -e
-set -u
+# Exit on undefined variable and error.
+set -u -e -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+cd "${SCRIPT_DIR}"
```
slipher, line 7: Please no. I often run this in different directories so that I can test things without clobbering the ones that I usually use.

external_deps/build.sh
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 
-# Exit on error
-# Error on undefined variable
-set -e
-set -u
+# Exit on undefined variable and error.
+set -u -e -o pipefail
```
slipher, line 4: Have you fully tested this? I seem to recall one of our scripts not working with pipefail

external_deps/build.sh
illwieckz: Oh? Well…

external_deps/build.sh
illwieckz: That's the first time I've heard that pipefail can be removed on purpose? What was the problem? If a script fails with pipefail that should be because of a real error…

external_deps/build.sh
illwieckz: I remove the `cd`.

I kept the `SCRIPT_DIR=` at the top and also moved `WORK_DIR=` at the top because I feel more confident this way, expecially with the `WORK_DIR=` which is calling for trouble if done late in execution in a function. I don't want to be bitten by this in the future after possible rewrites.

external_deps/build.sh
slipher: I don't remember but you can imagine scenarios that aren't a real error like a `grep` with 0 matches

illwieckz: Weird, I probably never faced myself a scenario with a grep with zero match…

illwieckz: Basically I usually do:

`if something | grep`

And that's not an error, since it is meant to be caught by `if`.

illwieckz: If an exception is found, I prefer someone to put an explicit `|| true` on the exception than entirely removing the check for the whole script.

slipher: It must not have been this one since there is only a single pipe in the entire script lol. LGTM

illwieckz:
I added a commit making possible to do `./build.sh <platform> base` and `./build.sh <platform> all`, as testing the script without those were too much painful.

I actually rebuilt everything with `-o pipefail` without error. I haven't tried `install`, `package`, and `wipe` though.

illwieckz:
I also edited curl build function to not build curl with brotli and zstd. If the host has brotli and zstd installed for the current architecture, one cannot cross-compile curl for another architecture, as headers would be found and then features will be enabled, but linking will fail.

PR #875, opened 2023-05-19
external_deps: add a tool to build cross-compilers

illwieckz:
Add a tool to build cross-compilers.

This is currently a proof of concept and it is not used yet. The toolchains are built with [crosstool-NG](https://crosstool-ng.github.io/).

It is rebased on:

- https://github.com/DaemonEngine/Daemon/pull/874

The `build.sh` script can optionally use the built toolchain with `--with-toolchain`.

The idea is that someone can do:

```
./build-toolchain.sh linux-arm64
./build.sh --with-toolchain linux-arm64-default jpeg
```

As I was testing the ability to do it, the script currently builds the compiler with the oldest glibc and oldest kernel the [crosstool-NG](https://crosstool-ng.github.io/) because I wanted to verify the ability to build new compilers with old glibc. We may not want to build the oldest glibc possible if there is some advantage to not use a too-old glibc. I haven't seen any way to force any old libc++, so I don't know how the builds behave on that side of things.

### DIFF
--- a/external_deps/build-toolchain.sh
+++ b/external_deps/build-toolchain.sh
@@ -1,0 +1,209 @@
+#! /usr/bin/env bash
+
+# Exit on undefined variable and error.
+set -u -e -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+WORK_DIR="${PWD}"
+
+# This should match the TOOLCHAINS_VERSION in build.sh.
+TOOLCHAINS_VERSION='1'
+
+# Package versions
+TOOLCHAIN_CTNG_VERSION='1.25.0'
+
+setup_linux-amd64() {
+	TOOLCHAIN_SYSTEM='linux'
+	TOOLCHAIN_TARGET='x86_64-unknown-linux-gnu'
+	TOOLCHAIN_GCC_VERSION='11'
+	TOOLCHAIN_GLIBC_VERSION='2.17'
+	TOOLCHAIN_LINUX_VERSION='3.2'
+}
+
+setup_linux-i686() {
+	TOOLCHAIN_SYSTEM='linux'
+	TOOLCHAIN_TARGET='i686-ubuntu14.04-linux-gnu'
+	TOOLCHAIN_GCC_VERSION='11'
+	TOOLCHAIN_GLIBC_VERSION='2.17'
+	TOOLCHAIN_LINUX_VERSION='2.6.32'
+}
+
+setup_linux-arm64() {
+	TOOLCHAIN_SYSTEM='linux'
+	TOOLCHAIN_TARGET='aarch64-unknown-linux-gnu'
+	TOOLCHAIN_GCC_VERSION='11'
+	TOOLCHAIN_GLIBC_VERSION='2.17'
+	TOOLCHAIN_LINUX_VERSION='3.10'
+}
+
+setup_linux-armhf() {
+	TOOLCHAIN_SYSTEM='linux'
+	TOOLCHAIN_TARGET='arm-unknown-linux-gnueabi'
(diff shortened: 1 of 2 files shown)


illwieckz:
This is something I get on Debian Buster:

```
$ ./build.sh linux-arm64-default jpeg
-- The C compiler identification is GNU 8.3.0
…
-- Performing Test HAVE_VLD1_S16_X3
-- Performing Test HAVE_VLD1_S16_X3 - Failed
-- Performing Test HAVE_VLD1_U16_X2
-- Performing Test HAVE_VLD1_U16_X2 - Success
-- Performing Test HAVE_VLD1Q_U8_X4
-- Performing Test HAVE_VLD1Q_U8_X4 - Failed
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/aarch64-linux-gnu-gcc
-- Use partial Neon SIMD intrinsics implementation (NEON_INTRINSICS = OFF)
-- CMAKE_ASM_FLAGS =  -fPIC -march=armv8-a  -O3 -DNDEBUG
```

```
$ ./build.sh --toolchain linux-arm64-default jpeg
-- The C compiler identification is GNU 11.2.0
…
-- Performing Test HAVE_VLD1_S16_X3
-- Performing Test HAVE_VLD1_S16_X3 - Success
-- Performing Test HAVE_VLD1_U16_X2
-- Performing Test HAVE_VLD1_U16_X2 - Success
-- Performing Test HAVE_VLD1Q_U8_X4
-- Performing Test HAVE_VLD1Q_U8_X4 - Success
-- The ASM compiler identification is GNU
-- Found assembler: Unvanquished/daemon/external_deps/toolchains_1/linux-arm64/bin/aarch64-unknown-linux-gnu-gcc
-- Use partial Neon SIMD intrinsics implementation (NEON_INTRINSICS = OFF)
-- CMAKE_ASM_FLAGS =  -fPIC -march=armv8-a  -O3 -DNDEBUG
```

Unfortunately this is not using NEON yet, but we can see the compiler being upgraded and some tests succeeding.

external_deps/build-toolchain.sh
```diff
@@ -0,0 +1,209 @@
+#! /usr/bin/env bash
+
+# Exit on undefined variable and error.
+set -u -e -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+WORK_DIR="${PWD}"
+
+# This should match the TOOLCHAINS_VERSION in build.sh.
+TOOLCHAINS_VERSION='1'
```
slipher, line 10: Why is this needed? Are you going to distribute binary packages?

external_deps/build-toolchain.sh
```diff
@@ -0,0 +1,209 @@
+#! /usr/bin/env bash
```
slipher, line 1: Could this be Python? Compared to build.sh it seems like there are less simple shell commands and more management of temp directories etc. so maybe it would benefit.

slipher (commented):
What's the main purpose of this? Is it mainly the old libc thing or is it actually needed for cross-compiling?

illwieckz:
> What's the main purpose of this? Is it mainly the old libc thing or is it actually needed for cross-compiling?

I'm trying to get a way to build the deps and the game with a new compiler but with an old libc.

With that script I managed to build a GCC11 with an old libc on Debian Buster and build a deps. I have not tested more but that starts to look good.

external_deps/build-toolchain.sh
illwieckz: Yeah, maybe, I also thought about the same after having written this.

illwieckz:
If we do this one day, this would be better to implement it in `unvanquished-release-scripts` instead:

- https://github.com/Unvanquished/release-scripts

PR #876, opened 2023-05-19
Don't archive language cvar

slipher:
The auto-detected language will no longer by archived by default. It will only be archived if you use `seta` or choose a language from a (not yet existing) menu in the UI.

Fixes https://github.com/Unvanquished/Unvanquished/issues/2165.

### DIFF
--- a/src/engine/qcommon/translation.cpp
+++ b/src/engine/qcommon/translation.cpp
@@ -39,7 +39,7 @@ extern "C" {
     #include "findlocale/findlocale.h"
 }
 
-static Cvar::Cvar<std::string> language("language", "language for UI text", Cvar::ARCHIVE, "");
+static Cvar::Cvar<std::string> language("language", "language for UI text", Cvar::NONE, "");
 
 void Trans_LoadDefaultLanguage()
 {


illwieckz:
We need a mechanism to return in 'auto', selectable in drop-down list, i.e. remove the archive flag and maybe resetting the value.

This may be hacked by setting "auto" to the cvar, with code resetting the flag and the value is “auto” is set.

But then such dropdown menu would not display it is set to "auto (Language)".

slipher:
> We need a mechanism to return to 'auto', selectable in drop-down list, i.e. remove the archive flag and maybe resetting the value.
> 
> This may be hacked by setting "auto" to the cvar, with code resetting the flag and the value is “auto” is set.
> 
> But then such dropdown list would not display it is set to "auto (Language)".

We absolutely do NOT need an auto option in the dropdown list. The only people who need that are developers who want to test the language auto detection code and they can use `/reset language`.

illwieckz:
Well, I'm sure I have used this kind of option as an user in other software.

But you remind me we maybe can just add a `reset` button (something we should have for many other things too).

illwieckz:
For reference, the calling code does this:

```c++
Color::StripColors( prompt );
Field_Draw( g_consoleField,
	consoleState.margin.sides + consoleState.padding.sides + SCR_ConsoleFontStringWidth( prompt, strlen( stripped ) ),
```

If I do this:

```c++
prompt = Color::StripColors( prompt );
Field_Draw( g_consoleField,
	consoleState.margin.sides + consoleState.padding.sides + SCR_ConsoleFontStringWidth( prompt, strlen( stripped ) ),
```

I get:

```
Unvanquished/daemon/src/engine/client/cl_console.cpp:645:16: error: incompatible types in assignment of ‘char*’ to ‘char [1024]’
  645 |         prompt = Color::StripColors( prompt );
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

If I do this:



char* stripped = Color::StripColors( prompt );
Field_Draw( g_consoleField,
	consoleState.margin.sides + consoleState.padding.sides + SCR_ConsoleFontStringWidth( prompt, strlen( stripped ) ),

It didn't fixed it

slipher:
@illwieckz You seem to have commented in the wrong issue.

illwieckz:
Oh lol

illwieckz (approved):
LGTM.

Issue #730, opened 2022-11-07
Renderer: broken sprite code path


slipher:
`vboData_t::spriteOrientation` is never set anywhere, and it was that way since the field was introduced in [b10e8c2](https://github.com/DaemonEngine/Daemon/commit/b10e8c278c7817d52b6095f9f961598792a04dcb). I guess the "autosprite" thing (no idea what it is) has probably been broken at least since then.

illwieckz says:
> note that we have some known bugs related to sprites. Some can be seen in metro map, in catacombs: missing fire sprites, distorted chain sprite…

illwieckz:
It's possible that `vboData_t::spriteOrientation` is dead code.

Here are some materials from the metro map:

```
textures/metro/kosad_flammebougie
{
	qer_editorimage textures/metro/kosad_flammebougie1

	surfaceparm nolightmap
	surfaceparm nomarks
	surfaceparm trans

	cull disable
	deformvertexes autosprite2

	{
		animmap 50 textures/metro/kosad_flammebougie1 textures/metro/kosad_flammebougie2 textures/metro/kosad_flammebougie3 textures/metro/kosad_flammebougie4 textures/metro/kosad_flammebougie5 textures/metro/kosad_flammebougie6 textures/metro/kosad_flammebougie7 textures/metro/kosad_flammebougie8
		blendfunc add
	}
}


textures/metro/kosad_chaine
{
	qer_editorimage textures/metro/kosad_chaine

	cull disable
	deformvertexes autosprite2

	{
		map textures/metro/kosad_chaine
		alphafunc ge128
		depthwrite
	}
	{
		map $lightmap
		blendfunc filter
		depthfunc equal
	}
}
```

First one is candle flame, next is a metal string, both use `deformvertexes autosprite2`.

Here is a scene with both of them. The metal string is incorrectly distorted, the lightmap is not applied, but the texture is rendered and is always facing the player, there is a candle on the left, without flame:

![unvanquished_2023-12-15_095616_000](https://github.com/DaemonEngine/Daemon/assets/2311649/47405813-a488-40d4-b979-f554c9fc638f)

Now, if I swap the names:

```
textures/metro/kosad_chaine
//textures/metro/kosad_flammebougie
{
	qer_editorimage textures/metro/kosad_flammebougie1

	surfaceparm nolightmap
	surfaceparm nomarks
	surfaceparm trans

	cull disable
	deformvertexes autosprite2

	{
[...]

PR #882, opened 2023-05-21
tr_model_iqm: fix use of uninitialized data

### LINKED ISSUE #880: Some “Conditional jump or move depends on uninitialised value” in IQM code reported by valgrind
Reported by valgrind:

```
==00:00:02:18.181 1066938== Conditional jump or move depends on uninitialised value(s)
==00:00:02:18.181 1066938==    at 0x002a5074: R_CopyVertexData(VBO_t*, unsigned char*, vboData_t)
==00:00:02:18.181 1066938==    by 0x002a6c60: R_CreateStaticVBO(char const*, vboData_t, vboLayout_t) (tr_vbo.cpp:595)
==00:00:02:18.181 1066938==    by 0x00275b0d: R_LoadIQModel(model_t*, void const*, int, char const*) (tr_model_iqm.cpp:819)
==00:00:02:18.181 1066938==    by 0x00272cb1: RE_RegisterModel(char const*) (tr_model.cpp:155)
==00:00:02:18.181 1066938==    Uninitialised value was created by a stack allocation
==00:00:02:18.181 1066938==    at 0x002745a0: R_LoadIQModel(model_t*, void const*, int, char const*) (tr_model_iqm.cpp:444)
```

illwieckz:
Should fix #880:

- https://github.com/DaemonEngine/Daemon/issues/880

All other similar vboData initialization code already do the same, like:

- `src/engine/renderer/tr_model_skel.cpp` (md5 models)
- `src/engine/renderer/tr_model_md3.cpp`
- `src/engine/renderer/tr_light.cpp`
- `src/engine/renderer/tr_vbo.cpp`
- `src/engine/renderer/tr_bsp.cpp`

### DIFF
--- a/src/engine/renderer/tr_model_iqm.cpp
+++ b/src/engine/renderer/tr_model_iqm.cpp
@@ -805,6 +805,8 @@ bool R_LoadIQModel( model_t *mod, const void *buffer, int filesize,
 					  qtangentbuf[ i ] );
 		}
 
+		memset( &vboData, 0, sizeof( vboData ) );
+
 		vboData.xyz = (vec3_t *)IQModel->positions;
 		vboData.qtangent = qtangentbuf;
 		vboData.numFrames = 0;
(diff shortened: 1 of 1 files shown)


slipher:
That's #730

illwieckz:
Ah, interesting! Anyway an IQM model is not a sprite so that sprite field should not be set in all cases…

And we better zero the struct in all cases like every other code doing vboData does, there may be other fields.

illwieckz:
That's the only vbo code not doing that yet, that can't be wrong.

PR #886, opened 2023-05-21
cm: call `CM_FreeAll()` when doing `CM_ClearMap()`

illwieckz:
I've noticed a lot of stuff allocated with `CM_Alloc` were remaining on shutdown.

Previously it was only done on map load.

So I made sure `CM_FreeAll()` is also called on shutdown.

Also don't call `CM_ClearMap` in `CL_ShutdownAll`.

### DIFF
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -726,9 +726,6 @@ void CL_ShutdownAll()
 	{
 		void SV_ShutdownGameProgs();
 		SV_ShutdownGameProgs();
-
-		// clear collision map data
-		CM_ClearMap();
 	}
 
 	Hunk_Clear();
--- a/src/common/cm/cm_load.cpp
+++ b/src/common/cm/cm_load.cpp
@@ -918,8 +918,7 @@ void CM_LoadMap(Str::StringRef name)
 		Sys::Drop("Could not load %s", mapFile.c_str());
 	}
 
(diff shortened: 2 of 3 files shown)


src/engine/client/cl_main.cpp
```diff
@@ -727,6 +727,8 @@ void CL_ShutdownAll()
 		void SV_ShutdownGameProgs();
 		SV_ShutdownGameProgs();
 
+		// free old stuff
+		CM_FreeAll();
```
slipher, line 731: There is no CM_LoadMap in the client so this probably shouldn't be here (that goes for the existing CM_ClearMap as well)

src/common/cm/cm_public.h
```diff
@@ -38,6 +38,7 @@ Maryland 20850 USA.
 
 void         CM_LoadMap(Str::StringRef name);
 void         CM_ClearMap();
+void CM_FreeAll();
```
slipher, line 41: I think CM_FreeAll should be called by CM_ClearMap. External modules have no need to control these two things independently.

illwieckz: Done.

src/engine/client/cl_main.cpp
illwieckz: I removed both.

slipher:
LGTM

PR #891, opened 2023-05-22
external_deps: do not install Linux libraries to lib64

illwieckz:
Includes one commit from https://github.com/DaemonEngine/Daemon/pull/877 because there would be a merge conflict otherwise.

Do not install Linux external_deps libraries to lib64. Until now some were installed in lib, others in lib64, and when cross-compiling, some 32-bit libraries where installed in lib64.

The GLEW CMake change is probably useless, as all the 32-bit builds already worked before with `lib` but this makes it more consistent.


slipher:
LGTM

PR #894, opened 2023-05-23
don't use the same cache file for the generic and generic2D GLSL shader, etc.

### LINKED ISSUE #889: The generic shader permutations are always rebuilt on every map load
The generic shader permutations are always rebuilt on every map load on my end:

```
Debug: building generic shader permutation with macro: USE_DEPTH_FADE 
Debug: building generic shader permutation with macro: USE_ALPHA_TESTING 
Debug: building generic shader permutation with macro: none
Debug: building generic shader permutation with macro: USE_VERTEX_SKINNING 
Debug: building generic shader permutation with macro: USE_VERTEX_ANIMATION 
```

illwieckz:
- Don't use the same cache file for the generic and generic2D GLSL shader
  Do not rebuild generic GLSL shaders on every map load.
- set `GENERIC_2D` define without relying on an extra file.

Fixes #889:

- https://github.com/DaemonEngine/Daemon/issues/889

### DIFF
--- a/src.cmake
+++ b/src.cmake
@@ -173,7 +173,6 @@ set(GLSLSOURCELIST
     ${ENGINE_DIR}/renderer/glsl_source/fxaa_fp.glsl
     ${ENGINE_DIR}/renderer/glsl_source/fxaa_vp.glsl
     ${ENGINE_DIR}/renderer/glsl_source/fxaa3_11_fp.glsl
-    ${ENGINE_DIR}/renderer/glsl_source/generic2D_fp.glsl
     ${ENGINE_DIR}/renderer/glsl_source/generic_fp.glsl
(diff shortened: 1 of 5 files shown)


illwieckz:
It's a fixup for:

- https://github.com/DaemonEngine/Daemon/pull/637

At the time I advised @DolceTriade to do what is the current implementation,

- I didn't knew there was already a function to give an alternate name for the same GLSL file and didn't thought about the fact the cache would be reused for both variants.  The current implementation is then buggy in the way the GLSL shader is always recompiled (but that should not affect rendering).
- I didn't knew yet we already had a mechanism to set a specific define to reuse the same source file, I found it when trying to implement it. The current implementation is not buggy on that part, just uselessly convoluted.

slipher:
LGTM

Issue #885, opened 2023-05-21
Cvar code leaks small data very often according to valgrind


illwieckz:
Valgrind says the Cvar code leaks very often, it does not leak a lot of data (only some hundreds of bytes each time), but it does that very often. In the end the leaked data is not that big, but it pollutes a lot the valgrind ouput.

If that's not a bug, it would be good to have a pedantic cleaner to make easier to debug with valgrind.

Very small example of valgrind output:

```
==00:00:01:26.545 1430047== 200 bytes in 1 blocks are possibly lost in loss record 177 of 232
==00:00:01:26.545 1430047==    at 0x4845013: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==00:00:01:26.545 1430047==    by 0x2B2D2F: Cvar::Register(Cvar::CvarProxy*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (CvarSystem.cpp:336)
==00:00:01:26.545 1430047==    by 0x12DC4A: Cvar_Get(char const*, char const*, int) (cvar.cpp:108)
==00:00:01:26.545 1430047==    by 0x1F68F6: CL_InitRenderer() (cl_main.cpp:2596)

==00:00:01:26.545 1430047== 200 bytes in 1 blocks are possibly lost in loss record 178 of 232
==00:00:01:26.545 1430047==    at 0x4845013: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==00:00:01:26.545 1430047==    by 0x2B2D2F: Cvar::Register(Cvar::CvarProxy*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (CvarSystem.cpp:336)
==00:00:01:26.545 1430047==    by 0x12DC4A: Cvar_Get(char const*, char const*, int) (cvar.cpp:108)
==00:00:01:26.545 1430047==    by 0x1F691B: CL_InitRenderer() (cl_main.cpp:2597)

[...]

PR #896, opened 2023-05-24
CommandSystem: free all commands on exit

illwieckz:
Partially frees some cvar leaks:

- https://github.com/DaemonEngine/Daemon/issues/885

It looks like all cvars spawn a command, those commands are never freed.

This only _partially_ fixes the problem because:

- Cvar code is still leaking a lot because CvarProxy code is leaking a lot too.

### DIFF
--- a/src/engine/framework/CommandSystem.cpp
+++ b/src/engine/framework/CommandSystem.cpp
@@ -115,8 +115,8 @@ namespace Cmd {
     // can be registered before main. The first time this function is called the command map
     // is initialized so we are sure it is initialized as soon as we need it.
     CommandMap& GetCommandMap() {
-        static CommandMap* commands = new CommandMap();
-        return *commands;
+        static CommandMap commands;
+        return commands;
     }
 
     // Used to emulate the C API


slipher:
This doesn't help. The command map is already freed by global destructors. And still nothing is done to free the commands themselves.

slipher:
Neither cvar commmands nor VM commands, nor old-style engine commands actually allocate a new command object. So few if any commands should be leaked.

illwieckz:
Then why `valgrind` and `heapusage` reports me thousands of leaks for those?

illwieckz:
This patch divides by 2 the size of the `heapusage` log: from 12M it goes down to 6M. File size is a bad metric, but with such obvious difference there should be something going on.

slipher:
Are you setting `common.pedanticShutdown`? Global destructors don't run unless you set that.

illwieckz:
> Are you setting `common.pedanticShutdown`? Global destructors don't run unless you set that.

Yes, looking for leaks without it would be impossible otherwise.

slipher:
Maybe your toolchain has a bug with not destroying function-local static variables.

illwieckz:
My test process is the following:

```
$ ( heapusage ./daemon -set common.pedanticShutdown on 2>&1 ) | tee "heapusage-$(date '+%Y%m%d-%H%M%S').log"
] devmap plat23
] quit
$ vim $( ls -t heapusage-*.log | head -n 1 )
/heapusage
```

Unlike `valgrind`, `heapusage` puts the summary and the highest leak at the top.

I build and run a Debug build with nexe vms (to not be polluted by possible game leaks).

slipher:
Oh it's something stupid. This:
```
static CommandMap* commands = new CommandMap();
```
Ought to be this:
```
static CommandMap commands;
```

slipher:
So in conclusion there is no real memory leak. It's just a pedantic cleanup issue, which you could fix by making the above change.

illwieckz:
I get the same with GCC `12.2.0` and Clang `15.0.7`.

illwieckz:
> `static CommandMap commands;`

This seems to work.

PR #898, opened 2023-05-24
use saner defaults for console colours

bmorel:
2nd half for fixing definitely https://github.com/Unvanquished/Unvanquished/issues/2441

This one is probably more a matter of tastes, so I consider it optional for fixing the issue, but those defaults (proposed by @cu-kai ) are imo much, much more readable than current colours.


bmorel:
ping.

cu-kai:
While we're improving the console, could we set `cl_consoleFontScaling` to 0 by default, and add an option to the GUI for both this and `cl_consoleFontSize`? The scaled option looks ugly and blurry IMO.

bmorel:
Can certainly do.

PR #901, opened 2023-05-25
Compiler: set __SANITIZE_ADDRESS__ on clang if has attribute no_sanitize_address

illwieckz:
This fixes ASAN on clang.

I also modified a bit the code to be less clang-specific and doing [like clang recommends](https://clang.llvm.org/docs/AddressSanitizer.html#id11):

The commits are meant to be squashed, first ones are experiments to do this in other ways, but most of them doesn't work, if you knows how to make `ATTRIBUTE_NO_SANITIZE_ADDRESS` for our `new` and `delete` override, this may be better, but this maybe not possible.

### DIFF
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -111,7 +111,7 @@ int CountTrailingZeroes(unsigned long long x);
 #define CONSTEXPR constexpr
 
 // To mark functions which cause issues with address sanitizer
-#if __clang__
+#if defined(__has_feature)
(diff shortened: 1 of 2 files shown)


slipher:
It's bad to define system macros ourselves so I propose #903

illwieckz:
Yeah, I was going to ask if we should set a custom define instead, I expected comments on the ideas first. 😄️

PR #903, opened 2023-05-25
Fix address sanitizer detection with Clang

slipher:


DolceTriade (approved):

illwieckz (approved):
LGTM.

illwieckz:
Only thing I dislike is the `#[ident]if` syntax, my brain prefers `[indent]#if` one.

slipher:
> Only thing I dislike is the `#[ident]if` syntax, my brain prefers `[indent]#if` one.

It seems there's a good mix of both in the codebase... and even within this file. `grep -rE '^#\s' src` vs. `grep -rE '^\s+#' src`

illwieckz:
Yeah, but if someone rewrites a line, I would prefer `#if` always being `#if` whatever the indentation, otherwise the brain has to parse both `#if` and `#    if` that's annoying.

PR #904, opened 2023-05-25
cmake: fix SDL2 copy on MSVC

illwieckz:
Fix SDL2 copy on MSVC.

I hope this don't break MingW copy.

I have not tested it on Windows with MSVC.


illwieckz:
> Generate a list of files that match the <globbing-expressions> and store it into the <variable>.
> -- https://cmake.org/cmake/help/latest/command/file.html#filesystem

So I guess it shouldn't list a garbage `lib/*/SDL2.dll` path if not found like `bash` does, so it should work with Mingw too (the special MSVC path being ignored).

slipher:
LGTM, works with msvc/mingw

PR #905, opened 2023-05-25
get rid of almost all qcommon.h cvar_t

bmorel:
Does not builds yet because I don't know how to fix:

> daemon/src/engine/qcommon/net_ip.cpp:1708:15: error: no member named 'modified' in 'Cvar::Cvar<int>'
        net_enabled->modified = false;

The only remaining cvar_t in that header file will then be `com_version` because I'm even less sure of how to migrate it. Maybe I'll just avoir migrating net_enabled, though. Dunno yet.

### DIFF
--- a/src/engine/qcommon/common.cpp
+++ b/src/engine/qcommon/common.cpp
@@ -52,12 +52,13 @@ Maryland 20850 USA.
 #include "sys/sys_events.h"
 #include <common/FileSystem.h>
 
+Cvar::Cvar<bool> com_developer = Cvar::Cvar<bool>( "developer", "", Cvar::TEMPORARY, false );
+Cvar::Cvar<int> com_speeds = Cvar::Cvar<int>( "com_speeds", "", Cvar::NONE, 0 );
+Cvar::Cvar<float> com_timescale( "timescale", "", Cvar::CHEAT | Cvar::SYSTEMINFO, 1.0f );
+
 cvar_t *com_pid; // bani - process id
 
-cvar_t *com_speeds;
-cvar_t *com_developer;
-cvar_t *com_timescale;
-cvar_t *com_dropsim; // 0.0 to 1.0, simulated packet drops
+Cvar::Range<Cvar::Cvar<float>> com_dropsim = Cvar::Range<Cvar::Cvar<float>>( "com_dropsim", "simulated packet drop", Cvar::CHEAT, 0.0f, 0.0f, 1.0f );
 
 Cvar::Cvar<bool> cvar_demo_timedemo(
     "demo.timedemo",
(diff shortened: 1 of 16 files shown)


slipher:
FYI for the gamelogic cvar migration I used a [script](https://github.com/slipher/source-tools/blob/master/cvar_migrate.py) to automate the more tedious parts. However the script needs to be modified to work on engine cvars also.

I would recommend skipping net_enabled because of #908. 

Are you going to provide descriptions for the cvars?

bmorel:
> Are you going to provide descriptions for the cvars?

I might do my best for that, but I'm afraid in many cases it will still be a "no" sadly.
I'll look at your script, thanks.

bmorel:
Removed the net_enabled cvar which was incomplete anyway, let's see if things build now.

src/engine/qcommon/common.cpp
```diff
@@ -52,25 +52,27 @@ Maryland 20850 USA.
 #include "sys/sys_events.h"
 #include <common/FileSystem.h>
 
+Cvar::Cvar<bool> com_developer = Cvar::Cvar<bool>( "developer", "", Cvar::TEMPORARY, false );
+Cvar::Cvar<int> com_speeds = Cvar::Cvar<int>( "com_speeds", "", Cvar::NONE, 0 );
```
necessarily-equal, line 56: I think you can remove the ` = Cvar::Cvar<…>` part of these lines

src/engine/qcommon/common.cpp
```diff
@@ -52,25 +52,27 @@ Maryland 20850 USA.
...
+Cvar::Cvar<float> com_timescale( "timescale", "", Cvar::CHEAT | Cvar::SYSTEMINFO, 1.0f );
+
 cvar_t *com_pid; // bani - process id
 
-cvar_t *com_speeds;
-cvar_t *com_developer;
-cvar_t *com_timescale;
-cvar_t *com_dropsim; // 0.0 to 1.0, simulated packet drops
+Cvar::Range<Cvar::Cvar<float>> com_dropsim = Cvar::Range<Cvar::Cvar<float>>( "com_dropsim", "simulated packet drop", Cvar::CHEAT, 0.0f, 0.0f, 1.0f );
```
necessarily-equal, line 61: I think you can remove the ` = Cvar::Cvar<…>` part of this line

src/engine/qcommon/common.cpp
```diff
@@ -52,25 +52,27 @@ Maryland 20850 USA.
...
-cvar_t *com_version;
 
-cvar_t *com_unfocused;
-cvar_t *com_minimized;
+cvar_t *com_version;
+Cvar::Cvar<bool> com_sv_running = Cvar::Cvar<bool>( "sv_running", "", Cvar::ROM, false );
+Cvar::Cvar<bool> com_unfocused = Cvar::Cvar<bool>( "com_unfocused", "", Cvar::ROM, false );
+Cvar::Cvar<bool> com_minimized = Cvar::Cvar<bool>( "com_minimized", "", Cvar::ROM, false );
+//NOTE: this one is never used in that file, this may belong somewhere else
+Cvar::Cvar<bool> com_cl_running = Cvar::Cvar<bool>( "cl_running", "", Cvar::ROM, false );
```
necessarily-equal, line 75: I think you can remove the ` = Cvar::Cvar<bool>` part of these lines

necessarily-equal (commented):
Had a quick early read, thanks for starting this.

src/engine/qcommon/common.cpp
bmorel: nitpick.

src/engine/qcommon/common.cpp
bmorel: nitpick.

src/engine/qcommon/common.cpp
bmorel: nitpick.

PR #907, opened 2023-05-26
make Str::BasicStringRef closer to C++17 string_view()

bmorel:


bmorel:
Please review this quickly, as it prevents the CI to make https://github.com/Unvanquished/Unvanquished/pull/2720 pass, but the real reason is *here* not in the aforementioned proposal.

slipher:
The way I've heard the rule of 5 is that you should define either 0 or 5 of those things. We define 0 so we are already in compliance.

bmorel:
Maybe I'm wrong then. I'm too lazy to check on that. Please could you comment on the PR?

slipher:
> The Rule of Five is a modern extension to the Rule of Three. The Rule of Five states that if a type ever needs one of the following, then it must have all five.
>- copy constructor
>- copy assignment
>- destructor
>- move constructor
>- move assignment

https://www.codementor.io/@sandesh87/the-rule-of-five-in-c-1pdgpzb04f

bmorel:
Does it makes this PRs bad or good though? I can change title if you want. Sorry, I'm doing a barbecue right now with some aperitif, so can't think too much.
I'd just like to know: is this going to generate regressions or not? If not, is it going to fix problems in the present (see unv's pr) and future?

bmorel:
Ok, I changed title and added a bit of content even, so that this implementation is closer to the C++17 std::string_view, which does allow default construction and copy construction. It also allows construction from a `char*` + a size.

The 2 implementations I've checked from C++ STL on my computer (LLVM's libc++11 and GNU's libstdc++) do not provide the move ctors, but I took the freedom to add them [because]https://en.cppreference.com/w/cpp/language/rule_of_three):

> Unlike Rule of Three, failing to provide move constructor and move assignment is usually not an error, but a missed optimization opportunity.

bmorel:
Looks like I forgot to actually commit before pushing. Oops.

slipher:
StringRef is intentionally different from string_view. StringRef is supposed to be a C-string, i.e. a pointer to a null-terminated byte sequence. Except with precomputed size information for performance. string_view is an arbitrary sequence of bytes that need not be null-terminated. We have Str::StringView for the exact equivalent of that.

bmorel:
Ok then.
I think this is all a bit confusing, and don't see what this PR would break or make slower, though. But if you really consider this a bad idea I'll close this, no problem.

slipher:
Yeah I don't think the PR helps anything.

src/common/String.h
```diff
@@ -49,10 +49,19 @@ namespace Str {
...
+        BasicStringRef& operator=( BasicStringRef<T> const& ) noexcept = default;
+        BasicStringRef& operator=( BasicStringRef<T> && ) noexcept = default;
+        ~BasicStringRef() noexcept = default;
+
         BasicStringRef(const std::basic_string<T>& other)
             : ptr(other.c_str()), len(other.size()) {}
         BasicStringRef(const T* other)
             : ptr(other), len(std::char_traits<T>::length(other)) {}
+        BasicStringRef(const T* other, size_t len_)
+            : ptr(other), len(len_) { ASSERT( len_ <= std::char_traits<T>::length(other) ); }
```
necessarily-equal, line 64:
```suggestion
            : ptr(other), len(len_) { ASSERT_EQ( len_, std::char_traits<T>::length(other) ); }
```

necessarily-equal (commented):
I'm also not sure about the point of this PR, and I think it also breaks an invariant that Str::StringRef are never pointing to nullptr, or that len == strlen(str). I would expect some code to rely on that.

I think we can close this PR, or keep only the new non-null constructor as above (with the update the assert above)

bmorel:
> and I think it also breaks an invariant that Str::StringRef are never pointing to nullptr, or that len == strlen(str).

If you ran tests more often, I'd be interested in the reasons why you think that. But I'm not really. This change is not wished for, so it's fine. If only all not wished changes were closed instead of merged, it would be nice.

PR #912, opened 2023-06-02
Fix SDL2.dll copying for MSVC

slipher:
#904 doesn't work... maybe I had already manually fixed it when I tried it then forgot


DolceTriade:
lgtm

Issue #654, opened 2022-07-26
Empty or incomplete server list when using French Free Mobile and Bouygues Telecom Internet providers (LTE mobile broadband)


illwieckz:
It looks like there was no dedicated issue, here is a first attempt to workaround the problem:

- https://github.com/DaemonEngine/Daemon/pull/283

Free Mobile is a French Internet Provider (mobile broadband), it is very used as it is one of the only 4 mobile broadband providers for France.

When an user is connected to Internet through this kind of connection (for example sharing Internet with his phone), the game don't get answers from queried servers. It's possible the servers don't receive the queries to begin with.

It appears that waiting a bit between each server query works around the issue (hence the #283 attempt).

Among people who experience the problem are:

- @illwieckz
- @bmorel
- @necessarily-equal
- @superdupont-fr
- `err404` (chat user)
- `BL0CKY_DUDE` (chat user)

The current workaround is to use https://unvanquished.net/servers/ if `unv://` protocol integration is working.

I tag this issue as `Enhancement` and not as `Bug` because that's likely not a bug in our code, but a problem in provider's infrastructure. Though, it would be useful if our engine was more resilient to providers with network issues.

illwieckz:
On a side note, a new implementation of server querying may also want to take care of that other issue:

- https://github.com/DaemonEngine/Daemon/issues/284

necessarily-equal:
the duplicate https://github.com/DaemonEngine/Daemon/issues/654 has some screenshots

superdupont-fr:
I almost never use shared connexion. But, I can test this case if needed. (since you added me in the list :D )

illwieckz:
@bmorel reports the same issue happens with another French Mobile Internet provider: Bouygues Telecom.

slipher:
This is nothing out of the ordinary... some servers missing during a client server list refresh has always happened since Tremulous. The client just sends a serverinfo request and the server doesn't appear if the packet gets lost or the response takes too long. If you guys have not seen this before now, you must have been blessed with unusually reliable ISPs.

illwieckz:
But when this bugs happen, having at least one sever listed in the list is the exception, otherwise none are listed.

bmorel:
FWIW I think the problem I had yesterday was a temporary problem, as I did not had it before, and it seems it's gone this morning.

PR #915, opened 2023-06-03
Server listing cleanup

### LINKED ISSUE #284: Server list refresh is blocking th UI (all servers are queried in a single frame)
As seen in #283, server list refreshing is done within its own loop. So either we manage to put that network thing in a separate thread, either we have to rewrite that thing in a way the main loop iterates the servers.

slipher:
Some cleanup in client's server listing code, in preparation to rewrite the server list API to fix #284, #654, etc.


necessarily-equal (approved):
Code changes LGTM, haven't tested

PR #919, opened 2023-07-06
Wshadow

bmorel:
Fixes the -Wshadow warning in daemon.


slipher:
We have a trailing underscore convention for class members at https://wiki.unvanquished.net/wiki/Coding_convention#Naming_conventions so it would be better not to use this for variable names.

PR #920, opened 2023-08-06
Implement numbered arguments for format strings

slipher:
Fixes https://github.com/Unvanquished/Unvanquished/issues/2696.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -449,6 +449,9 @@ endfunction()
 include(srclibs.cmake)
 include(src.cmake)
 
+# tinyformat is a header-only library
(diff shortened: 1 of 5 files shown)


DolceTriade (approved):

slipher:
I have discovered that the tests to prevent results like `0inf` fail with Clang. I guess I should actually add the tests to the CI at this point.

slipher:
I dropped the backported commit to prevent outputs like `format("%04.0f", nan)` -> `"0nan"` and associated tests, since it is not compatible with `-ffast-math`.

DolceTriade (approved):

PR #921, opened 2023-08-11
CI: run unit tests

slipher:


DolceTriade (approved):
Do you think we should mandate some form of tests on new PRs?

slipher:
> Do you think we should mandate some form of tests on new PRs?

We can hardly mandate tests on all PRs since many things like rendering are extremely difficult to test. Reviewers might ask for them in specific cases if it seems easy to do and beneficial.

PR #925, opened 2023-08-22
Fix debug drawing (used for navedit)

### LINKED ISSUE #924: Debug draw interface is broken
I inserted a simple test for the bot debug drawing facility at the beginning of `G_RunFrame`:

```
	Cvar::SetValue("r_debugSurface", "1");
	DebugDrawQuake dd;
	dd.init();
	dd.begin(DU_DRAW_TRIS);
	dd.vertex(500, 500, 600, -1);
	dd.vertex(600, 500, 500, -1);
	dd.vertex(500, 600, 500, -1);

	dd.vertex(-500, 500, -600, -1);
	dd.vertex(-600, 500, -500, -1);
	dd.vertex(-500, 600, -500, -1);
	dd.end();
	dd.sendCommands();
```

There should be one triangle around (500, 500, 500) and one around (-500, -500, 500), taking into account the swapped y/z coordinates. However most of the time, the triangles are drawn in some incorrect way, with triangles that use some verts from both sides. Tested on map Pushcannon:
unvanquished_2023-08-21_115331_000

The triangles that are actually drawn changes every now and then, on the order of one minute. A similar behavior can be seen with `navedit`, where triangles frequently appear and disappear.

## Workaround
I have tried this workaround with both Nvidia and Intel graphics. The bug can be avoided by setting these cvars together:
- `r_arb_map_buffer_range 0`
- `r_arb_buffer_storage 0`

slipher:
I've copied a bit of code from Tess_DrawElements without comprehension.

Fixes #924.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -4634,7 +4634,14 @@ void DebugDrawEnd() {
 
 	if ( glState.currentVBO && glState.currentIBO )
 	{
-		glDrawElements( drawMode, tess.numIndexes, GL_INDEX_TYPE, BUFFER_OFFSET( 0 ) );
+		uintptr_t base = 0;
+		if ( glState.currentIBO == tess.ibo )
+		{
+			base = tess.indexBase * sizeof(glIndex_t);
+		}
+
+		glDrawElements( drawMode, tess.numIndexes, GL_INDEX_TYPE, BUFFER_OFFSET( base ) );
+		GL_CheckErrors();
 
 		backEnd.pc.c_drawElements++;
 


illwieckz:
Awesome!

I confirm it fixes the bug on my end, while keeping performances acceptable for a debug draw.

With a scene on chasm, I got:

- `200 fps` with navedit disabled,
- `100 fps` with navedit enabled with default `r_arb_map_buffer_range 1` and `r_arb_buffer_storage 1`,
- but only `35 fps` with navedit enabled and both `r_arb_map_buffer_range 0` and `r_arb_buffer_storage 0`.

With this patch `navedit` works with those cvars set to `1` (default values).

unvanquished_2023-08-22_174850_000

illwieckz (approved):

PR #931, opened 2023-09-15
Tracing documentation, bugfixes and tests

slipher:
- Detailed documentation of trace_t explaining the semantics of the collision detection functions in the "cm" library. Besides the textual documentation there is also a function that verifies that a trace_t has the expected properties.

- Fix wrong values for the startsolid and allsolid trace_t flags in some cases.

- Add some tests, which are done against a plat23 BSP that I checked into the repository.  (Probably a bad choice of map in retrospect, as it lacks some features such as movers which would be nice to test against.) The tests are mostly for the parts that needed bug fixes rather than stuff that was already working.


sweet235:
Do you think we should try to make a test map instead of using plat23?

slipher:
> Do you think we should try to make a test map instead of using plat23?

With what goal? To reduce file size? To include every possible geometrical feature that could be tested in one map?

illwieckz:
That's a nice idea ! More tests is better ! 🎉️

One remark about file layout:

The purpose of `pkg/daemon_src.dpkdir` is to be able to embed its content in Dæmon binary in the future, in a similar way we do with `.glsl` files, in a way those files can always be found by the engine without requiring games to ship them in their own `.dpk` files.

Right now the only usage of that folder is to make easier to third-party game projects to run the engine without requiring Unvanquished `.dpk` files (it was useful to run the Xonotic experiment for example).

The reason why those files are not embedded in engine yet is only because only half the work has been done yet, but the purpose is clearly to automatically embed those files in engine binary at build time in the future.

So I suggest to add a special dpkdir for the test data instead, like `pkg/testdata_src.dpkdir`, to avoid any confusion and to make sure we don't mistakenly embed useless files in release builds the day we finish to implement the embedding of files from `pkg/daemon_src.dpkdir`.

slipher:
> So I suggest to add a special dpkdir for the test data instead, like `pkg/testdata_src.dpkdir`

Done

sweet235:
> To include every possible geometrical feature that could be tested in one map?

Yes.

slipher:
> > To include every possible geometrical feature that could be tested in one map?
> 
> Yes.

Well, it's hard to think of them all in advance. I don't know how to make a map anyway.

illwieckz:
For memories:

```
2023-10-08 00:05:28 +0200 <perturbed`> any objections to https://github.com/DaemonEngine/Daemon/pull/931
                                       https://github.com/Unvanquished/Unvanquished/pull/2793 ?
2023-10-08 00:20:21 +0200 <illwieckz> none
```

PR #943, opened 2023-10-24
Retries and rate limiting for serverinfo queries

slipher:
The important commit:

> The cvars cl_pingSpacing, cl_pingSpacingRetry1, and cl_pingSpacingRetry2
can be used to configure serverinfo queries (used to populate the server
list) to be attempted up to 3 times, possibly with a minimum delay
between outgoing packets which may be different for each round of
attempts. All first tries must be completed/timed out before starting
2nd attempts, etc.
>
> Default behavior is the same as 0.54 - one attempt with no delays.
>
> Fixes https://github.com/DaemonEngine/Daemon/issues/654.

### DIFF
--- a/src/engine/client/client.h
+++ b/src/engine/client/client.h
@@ -248,6 +248,7 @@ struct ping_t
 	netadr_t adr;
 	int      start;
 	int      time;
+	char     challenge[ 9 ]; // 8-character challenge string
 	char     info[ MAX_INFO_STRING ];
 };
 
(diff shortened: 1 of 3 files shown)


src/engine/client/client.h
```diff
@@ -248,6 +248,7 @@ struct ping_t
 	netadr_t adr;
 	int      start;
 	int      time;
+	char     challenge[ 9 ];
```
DolceTriade, line 251: Please leave a comment about the length of this field

DolceTriade (approved):

PR #944, opened 2023-10-26
Log: replace DoDebugCode(lambda) with if statement

slipher:
Instead of having Logger::DoDebugCode, DoVerboseCode etc. that conditionally execute a function based on the log level (to avoid evaluating expensive expressions for disabled messages), provide functions returning a bool that tell whether that level would currently be printed: Logger::ShowDebug(), ShowVerbose() etc. I don't see any benefit to the lambda; if (logger.ShowDebug()) { ... } is easier to type and more flexible.

Draft to remind me to make the gamelogic side before merging

What do you guys thing of the ShowDebug/ShowVerbose/etc. name? Is there anything better?

### DIFF
--- a/src/common/Log.h
+++ b/src/common/Log.h
@@ -64,7 +64,7 @@ namespace Log {
      *
      *   // However functions calls will still be performed.
      *   // To run code depending on the logger state use the following:
-     *   fooLog.DoNoticeCode([&](){
+     *   if (fooLog.ShowNotice()) {
      *       ExpensiveCall();
      *       fooLog.Notice("Printing the expensive expression %s", <the expression>);
      *   });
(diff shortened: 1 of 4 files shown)


DolceTriade:
I might replace this with an API like
```
if ( fooLog.IsLevelEnabled( Level::DEBUG ) ) 
{
  // Expensive debug code
}
```

That way it can be programatic.

DolceTriade:
I don't think function name length is a good way to measure verbosity... I mean none of the approaches are java length...

In the end they're functionally equivalent. `IsLevelEnabled` is just more mainstream :)

slipher:
> don't think function name length is a good way to measure verbosity...

I'm not measuring function length, but rather the length of all the code that must surround a conditionally executed log statement.

PR #948, opened 2023-10-27
deps: SDL2 CMake requires the bin/ dir to exist even when not using anything from it

### LINKED ISSUE #946: Fail to find `external_deps/linux-amd64-default_9/bin`
I tried to make a `linux-amd64` release build from current Unvanquished/Daemon tree:

```
./docker-build --reference 15c4ef159a2106e68e45937e3de987d8119dcec8 --targets linux-amd64
```

And I got:

```
CMake Error at daemon/external_deps/linux-amd64-default_9/lib/cmake/SDL2/sdl2-config.cmake:13 (message):
  File or directory
  /Unvanquished/daemon/external_deps/linux-amd64-default_9/bin referenced by
  variable SDL2_BINDIR does not exist !
  daemon/external_deps/linux-amd64-default_9/lib/cmake/SDL2/sdl2-config.cmake:26 (set_and_check)
  daemon/CMakeLists.txt:694 (find_package)
```

That reminds me something, if I'm right it's a bug in SDL2, where nothing from that folder is used but that folder has to exist because the SDL2 CMake script has to look for this folder being empty to know how to not use it, or something like that.

illwieckz:
SDL2 CMake requires the `bin/` dir to exist even when not using anything from it

Fixes #946:

- https://github.com/DaemonEngine/Daemon/issues/946

This also fixes doing release builds for Linux (yay!).

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -242,6 +242,11 @@ build_sdl2() {
 		CFLAGS="${CFLAGS} -O3" ./configure --host="${HOST}" --prefix="${PREFIX}" --libdir="${PREFIX}/lib" "${CONFIGURE_SHARED[@]}"
 		make
 		make install
+		# Workaround for an SDL2 CMake bug, we need to provide
+		# a bin/ directory even when nothing is used from it.
+		mkdir -p "${PREFIX}/bin"
+		# We don't keep empty folders.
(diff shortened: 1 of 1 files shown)


slipher:
LGTM

PR #952, opened 2023-11-04
Windows: fix manifest embedding and bug with screen scaling

slipher:
Fixes https://github.com/Unvanquished/Unvanquished/issues/2243


illwieckz (approved):
I'll have to trust you on the fact it works.

LGTM.

PR #956, opened 2023-11-08
renderer/tr_surface: use vectorNormalizeFast in MD5/IQM code

illwieckz:
This is CPU fallback code for when GPU can't accelerate model, to test this code one can do `set r_vboVertexSkinning 0`.

Continuation of the talk from there:

- https://github.com/DaemonEngine/Daemon/pull/953#discussion_r1382713656

> @slipher: Hmm do you have an argument for why precision isn't needed here? We had that bug with black patches in models due to the GLSL compiler compiling the same code slightly differently in 2 different shaders, so you can't just say that it's graphics so the precision doesn't matter...

> @illwieckz: Unfortunately I don't have an argument for why precision isn't needed here, I only have an argument for why performance is wanted here. 😅️

> @illwieckz: I don't know if that can help, but as far as I know when this code is used, the GLSL compiler is not used (because it is the fallback when that job can't be done on GPU), so we don't have to care about different GLSL compilers behaving differently.

### DIFF
--- a/src/engine/renderer/tr_surface.cpp
+++ b/src/engine/renderer/tr_surface.cpp
@@ -1183,9 +1183,9 @@ static void Tess_SurfaceMD5( md5Surface_t *srf )
 				VectorMA( binormal, *boneWeight, tmp, binormal );
 			}
 
-			VectorNormalize( normal );
-			VectorNormalize( tangent );
-			VectorNormalize( binormal );
+			VectorNormalizeFast( normal );
+			VectorNormalizeFast( tangent );
+			VectorNormalizeFast( binormal );
 			VectorCopy( position, tessVertex->xyz );
 
 			R_TBNtoQtangents( tangent, binormal, normal, tessVertex->qtangents );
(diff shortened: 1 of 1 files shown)


slipher:
I still don't know how precise normals need to be, but I found some relative error bounds for VectorNormalizeFast paths. We should write this in the code
- For the [SSE instruction](https://www.felixcloutier.com/x86/rsqrtss), 3.7 * 10^-4 (seems surprisingly bad)
- For the [Q3 bit hacking method two Newton iterations](https://en.wikipedia.org/wiki/Fast_inverse_square_root#/media/File:2nd-iter.png), 5 * 10^-6

slipher:
Now I realized that there is a Newton's method iteration to refine the result in the SSE case as well. (Misread it as being inside the preprocessor block for the non-asm case).

LGTM

illwieckz:
Thanks, then this raises my original question: maybe we can use `vectorNormalizeFast` anywhere there is a `vectorNormalizeFast` without return. There are other occurrences in other places of the code.

PR #960, opened 2023-11-11
Fix vistest counting bugs

slipher:


illwieckz:
What are “vis test” counting bugs?

slipher:
To be honest I don't know what a vis test is. Only that there are bugs in counting them :laughing:

illwieckz:
How did you noticed them then? 😄️

slipher:
Reading one of the commits from https://github.com/DaemonEngine/Daemon/pull/957.

illwieckz (approved):
So slipher is reading the matrix undeciphered, it makes sense. 😀️

I'll have to trust you on that, LGTM.

PR #964, opened 2023-11-13
files: implement fs_pakprefixes

illwieckz:
The idea is to allow to provide a complete tree of dpk archives in an subfolder and get them loaded first.

This will be useful for the nightly server, or for any mod.

Basically by running the game like this:

```sh
./daemon -set fs_pakprefixes 'dev'
```

The engine will look for `pkg/dev/unvanquished` dpk before the `pkg/unvanquished` one, and load it if exists.

It means we can host a whole copy of upcoming Unvanquished dpk archives in `pkg/dev` and get clients download it from `pkg/dev`, not breaking default installs on user side.

Like `fs_extrapaks`, the `fs_pakprefixs` cvar can gets a space-separated list of folders, so for example one can set `race dev` and the engine is expected to look first in `pkg/race`, then in `pkg/dev`, then in `pkg`.

First prefix to have a package for the given name gets the priority to load the package, even if an higher version of it exists in the next prefix (or base folder) to look for. I consider this is good enough and it makes sense anyway.

Right now I only tested with dpkdir, but I verified that only the server has to set the prefix, the client gets the list of translated paths (for example `dev/unvanquished`) from the server as expected.

PR #965, opened 2023-11-14
files: use std::string in FS_LoadPak

illwieckz:
A clean-up commit that was first pushed to #964 but now unneeded so I submit it separately.

### DIFF
--- a/src/engine/qcommon/files.cpp
+++ b/src/engine/qcommon/files.cpp
@@ -632,13 +632,15 @@ const char* FS_LoadedPaks()
 	return info;
 }
 
-bool FS_LoadPak(const char* name)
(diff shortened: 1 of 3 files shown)


src/engine/qcommon/files.cpp
```diff
@@ -632,13 +632,15 @@ const char* FS_LoadedPaks()
 	return info;
 }
 
-bool FS_LoadPak(const char* name)
+bool FS_LoadPak(const std::string &name)
```
DolceTriade, line 635: Prefer Str::StringRef right?

illwieckz: Done.

slipher:
LGTM

PR #966, opened 2023-11-16
Bring back videoMap with ROQ decoder

### LINKED ISSUE #936: Reimplementing RoQ video format and make sure video integration in maps work

slipher:
Cleaned-up version of #954. Uses bounds-checked accesses to all arrays to avoid buffer overflow issues. Unlike the old implementation, sizes other than 256x256 and multiple concurrent videos are supported.

Fixes #936.


illwieckz:
Nice! This works very well with the test map!

I would have preferred to have more history and separate commits removing Theora and Cinematic code (like on my branch) though…

illwieckz:
This is perfect:

- https://www.youtube.com/watch?v=4zvrkOruWRU

cu-kai:
I have just tested this with the `mxl-school` map and the RoQ plays once then the following error appears in my console: 
`Warn: error during ROQ video playback of video/guess256.RoQ: not enough data for codebook`

illwieckz:
Right, I reproduce the bug on my end.

For those having concerns about repackaged maps, this is a way to load the original one:

- download http://en.ws.q3df.org/maps/downloads/mxl_school.pk3
- store it in `pkg/`
- run the engine this way: `./daemon -set fs_legacypaks on -set g_neverEnd on +devmap mxl_school`
- go to the projection room: `/setviewpos 715 1227 320 -48 -2`

slipher:
> I have just tested this with the `mxl-school` map and the RoQ plays once then the following error appears in my console:
> `Warn: error during ROQ video playback of video/guess256.RoQ: not enough data for codebook`

Fixed. The ROQ_CODEBOOK command reads in two arrays but apparently the second one is optional, as demonstrated in the middle of the mxl_school video. Naturally, the original implementation deals with this by using random garbage from the read buffer for the second array, rather than detecting that it is not provided.

cu-kai:
Yes, the latest commit fixes it. Thanks.

src/engine/renderer/tr_video.cpp
```diff
@@ -0,0 +1,849 @@
...
+*/
+
+// tr_video.cpp: ROQ video playback on in-game surfaces
+//
+// This is a greatly cut-down version of the ROQ support in Quake 3. Cinematic mode has
+// been removed, along with a lot of dead or broken code. If something with in-game videos
+// works with Q3 but not Daemon, one may consult the ioq3 implementation:
+// https://github.com/ioquake/ioq3/blob/main/code/client/cl_cin.c
+// Variable and function names have been kept close to the original, even when the names
+// don't make much sense, to assist in the comparison.
```
illwieckz, line 42: If you'd like to get better function names, what you can do is to add the ioq3 function name in a comment in the previous line for reference?

illwieckz:
Now that I think about it, reading the `stretch` word in `RE_StretchRaw()` name, I guess some legacy cinematic RoQ files were square (like `512×512`) but displayed as `4:3` or so. I'm sure I've seen such cinematic RoQ files in some q3 game or mod around. Anyway, if that was only used for cinematics and not for videoMap it would be fine to not have this feature since we don't do cinematics.

illwieckz:
Yeah, typically `video/intro.RoQ` from Quake3 `pak0.pk3` has `512×256` resolution but I guess it's not the resolution it's played on screen. This feature is probably cinematic-only. I guess `videoMap` can use usual texture stretching like any other texture?

src/engine/renderer/tr_video.cpp
slipher: Yeah that would work. I'm not going to do further rewriting at this point in time though.

slipher:
> Yeah, typically `video/intro.RoQ` from Quake3 `pak0.pk3` has `512×256` resolution but I guess it's not the resolution it's played on screen. This feature is probably cinematic-only. I guess `videoMap` can use usual texture stretching like any other texture?

Yes, the normal texture stretching will work.

illwieckz:
Thanks, the PR looks good to me then.

illwieckz (approved):

Issue #681, opened 2022-08-14
Looping pak download


illwieckz:
I haven't found an existing issue, sorry if I missed it.

I just got the looping download issue after trying to join the nightly server:

```
Debug: Finished WWW download of 'pkg/dev/unv-nightly_0.53.1-20220811-213024-e5ed0ce~nightly_73da93bd.dpk.tmp', moving it to 'pkg/dev/unv-nightly_0.53.1-20220811-213024-e5ed0ce~nightly_73da' 
Debug: Downloaded something, reload the paks 
Debug:  The paks to load are 'dev/unv-nightly_0.53.1-20220811-213024-e5ed0ce~nightly_73da93bd.dpk unvanquished_0.53.1_e544bc58.dpk unvanquished_0.52.0_48e0bc40.dpk tex-common_2.2_cc893cee.dpk res-players_0.53.0_077fc7d9.dpk res-players_0.52_11bd6726.dpk res-legacy_0.52_a60d9e30.dpk res-weapons_0.53.0_76b152e5.dpk res-weapons_0.52.1_3d3f647f.dpk res-weapons_0.52_e3a09e99.dpk res-buildables_0.53.0_c859f572.dpk res-buildables_0.52_52a5777a.dpk res-voices_0.52_bd83cbc3.dpk res-soundtrack_0.52_e0e79e58.dpk map-chasm_1.2.2_5a090f7d.dpk tex-ex_1.2_89dbbaf5.dpk tex-pk01_1.2_7f032864.dpk tex-pk02_1.2_bff2a11c.dpk' 
[FS] Loading pak 'pkg/dev/unv-nightly_0.53.1-20220811-213024-e5ed0ce~nightly.dpk'... 
[FS] Loading pak 'base/pkg/unvanquished_0.53.1.dpk'... 
[FS] Loading pak 'base/pkg/unvanquished_0.52.0.dpk'... 
[FS] Loading pak 'base/pkg/tex-common_2.2.dpk'... 
[FS] Loading pak 'base/pkg/res-players_0.53.0.dpk'... 
[FS] Loading pak 'base/pkg/res-players_0.52.dpk'... 
[FS] Loading pak 'base/pkg/res-legacy_0.52.dpk'... 
[FS] Loading pak 'base/pkg/res-weapons_0.53.0.dpk'... 
[FS] Loading pak 'base/pkg/res-weapons_0.52.1.dpk'... 
[...]

PR #967, opened 2023-11-19
client: set originalDownloadName size to MAX_OSPATH since we copy there downloadName which is size MAX_OSPATH

illwieckz:
Set `originalDownloadName` size to `MAX_OSPATH` since we copy there `downloadName` which is size `MAX_OSPATH`.

We have this code in `src/engine/client/cl_download.cpp`:

```
  Q_strncpyz( cls.originalDownloadName, cls.downloadName, sizeof( cls.originalDownloadName ) );
```

but `sizeof(cls.downloadName)` is `256` (`MAX_OSPATH`) and `sizeof(cls.originalDownloadName)` is `64` (`MAX_QPATH`), meaning the server can tell the client to download a package whose name is longer than what the client string can carry. Let's just enlarge that string.

Large pak names with checksums would be truncated otherwise, letting the client stuck in an infinite download loop because the download of the pak will never succeed.

See:

- https://github.com/DaemonEngine/Daemon/issues/681

### DIFF
--- a/src/engine/client/client.h
+++ b/src/engine/client/client.h
@@ -326,7 +326,7 @@ struct clientStatic_t
 	// if new stuff gets added, CL_ClearStaticDownload code needs to be updated for clear up
 	char     downloadName[ MAX_OSPATH ];
 	char     downloadTempName[ MAX_OSPATH ]; // in wwwdl mode, this is OS path (it's a qpath otherwise)
-	char     originalDownloadName[ MAX_QPATH ]; // if we get a redirect, keep a copy of the original file path
+	char     originalDownloadName[ MAX_OSPATH ]; // if we get a redirect, keep a copy of the original file path
 	bool downloadRestart; // if true, we need to do another FS_Restart because we downloaded a pak
 };
 


illwieckz:
Here is a quick patch that reported me the error in logs when it happened (the download was still stuck in an infinite loop though):

```diff
diff --git a/src/engine/client/cl_download.cpp b/src/engine/client/cl_download.cpp
index ce9a6c7ba..84f35fb54 100644
--- a/src/engine/client/cl_download.cpp
+++ b/src/engine/client/cl_download.cpp
@@ -399,6 +399,15 @@ void CL_ParseDownload( msg_t *msg )
 			CL_AddReliableCommand( "wwwdl ack" );
 			cls.state = connstate_t::CA_DOWNLOADING;
 
+			// tell the developper something is wrong
+			if ( strnlen( cls.downloadName, sizeof( cls.downloadName ) ) >= sizeof( cls.originalDownloadName ) )
+			{
+				Log::Warn( "pak download name is too long", cls.downloadName );
+				CL_AddReliableCommand( "wwwdl fail" );
+				clc.bWWWDlAborting = true;
+				return;
+			}
+
 			// make sure the server is not trying to redirect us again on a bad checksum
 			if ( strstr( clc.badChecksumList, va( "@%s", cls.originalDownloadName ) ) )
 			{
```

illwieckz:
What I did to test this was to have a server serving `pkg/nightly/res-weapons_0.54-20230502-141543-25a9b80.dpk` as autodownload, downloading it properly by joining the server, then truncating the file this way:

```
echo toto > ~/.local/share/unvanquished/pkg/nightly/res-weapons_0.54-20230502-141543-25a9b80.dpk
```

After that when rejoining the server, it would try to write and load `pkg/nightly/res-weapons_0.54-20230502-141543-25a9b80_3f9ead50.dpk` but then the name is too long and truncated.

slipher:
LGTM

PR #968, opened 2023-11-20
common/String: include cstdint to define uint8_t

illwieckz:
Reported by Intel ICX 2023.2.2:

```
Unvanquished/daemon/src/common/String.h:287:20: error: unknown type name 'uint8_t'
    char HexDigit( uint8_t digit );
                   ^
1 error generated.
```

### DIFF
--- a/src/common/String.h
+++ b/src/common/String.h
@@ -33,6 +33,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <algorithm>
 #include <limits>
+#include <cstdint>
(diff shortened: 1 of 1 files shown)


slipher:
A lot of the headers in common/ are not self-contained. Unless you are going to make a big effort to make them all self-contained, the better fix would be to ensure that `Common.h` is included first in the .cpp file that fails to compile. Common.h needs to be included first in all files for the precompiled headers to work consistently between compilers.

illwieckz:
Closed in favor of:

- https://github.com/DaemonEngine/Daemon/pull/969

PR #970, opened 2023-11-21
cmake/pnacl: minor rewrite

illwieckz:
This is pure bikeshedding.

When toying with the Saigo toolchain I copy/pasted our PNaCl toolchain code and rewrote some parts in a better way.

This backports those “rewriting in a better way” to the PNaCl toolchain.

Basically:

- set variable for `PNACL_TRANSLATE` and `PNACL_STRIP` like we do for the `CC`, `CXX`, `AR`, `RANLIB`, etc. The strip variable is prefixed with `PNACL_` and not with `CMAKE_` because it's not purposed to be used by CMake builtin strip functions, so I thought it would be less confusing this way.

- `pnacl_translate` is renamed `pnacl_finalize` because it does more than translating, it does multiple things (including translating) as part of a finalization process.

- some code is factored.

- the `nacl_option` variable is renamed `pnacl_arch` because that's what it is.


slipher:
LGTM

PR #972, opened 2023-11-25
Filesystem: libpath-related fixup

slipher:


illwieckz:
Thanks, such kind of obsolete terminology can be pretty confusing, especially the “basepath” which was the libpath but could be confused which what is baseq3/baseoa in other games, which is not the libpath but the primary pakpath.

illwieckz (approved):
LGTM.

PR #974, opened 2023-11-27
Cvar improvements: float serialization and color bleed

slipher:


slipher:
Bump

illwieckz:
What if, instead of:

```c++
description = Str::Format("\"%s%s\" - %s", defaultValue, detail::ColorReset(defaultValue), …
```

we would do:

```c++
description = Str::Format("\"%\" - %s", detail::ColorSecure(defaultValue), …
```

So, basically `ColorSecure()` (just a random name that come to my mind, not an actual name recommendation) would return the equivalent of `"%s%s", defaultValue, detail::ColorReset(defaultValue)`

slipher:
On second thought I decided that trying to handle the trailing '^' case is overengineering - no other part of the code does that after all. I changed it to simply use `^*` everywhere.

illwieckz (approved):
LGTM

PR #975, opened 2023-11-30
external_deps: always use system curl, not self-built curl from external_deps custom PATH

illwieckz:
external_deps: always use system curl, not self-built curl from external_deps custom PATH


slipher:
lol I've never gotten that, I wonder what sequence of operations you had to do.

LGTM

illwieckz:
I did:

```
./build.sh linux-amd64-default ogg opus opusfile curl sdl2 glew openal ncurses naclsdk install
```

Then I did:

```
./build.sh linux-amd64-default freetype install
```

🤡️

PR #979, opened 2023-12-02
Actually disable PCH when Ninja generator is used

slipher:
The code that was supposed to turn it off by default didn't work.


illwieckz (approved):
LGTM.

PR #980, opened 2023-12-02
cmake/DaemonGame: make code more reusable for other toolchains

illwieckz:
Very minor.

When toying with Saigo I noticed this code could be factored: the `PLATFORM_EXE_SUFFIX` is set in the toolchain file so there is no need to hardcode the value there. The same code would probably work for `.pexe`, direct `.nexe` or even `.wasm` as long as the toolchain properly sets `PLATFORM_EXE_SUFFIX`.

### DIFF
--- a/cmake/DaemonGame.cmake
+++ b/cmake/DaemonGame.cmake
@@ -129,17 +129,7 @@ function(GAMEMODULE)
 
         # Generate NaCl executables for supported architectures.
         foreach(NACL_TARGET ${NACL_TARGETS})
-            if (NACL_TARGET STREQUAL "i686")
-                set(PNACL_ARCH "i686")
-            elseif (NACL_TARGET STREQUAL "amd64")
-                set(PNACL_ARCH "x86-64")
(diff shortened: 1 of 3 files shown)


slipher:
It's a bit of an obfuscation as .nexe is what we actually use. It's not obvious which thing that variable should contain.

illwieckz:
Then it's better to add a comment saying it's pexe with PNaCl than to hardcode something that would require to duplicate code for other toolchains.

illwieckz:
> Then it's better to add a comment saying it's pexe with PNaCl than to hardcode something that would require to duplicate code for other toolchains.

I did that.

I also added two other commits:

- cmake/toolchain-pnacl: remove unused `pnacl_finalize(target)` function \
  also avoid confusion with `pnacl_finalize(dir, module, pnacl_arch, arch)`

- cmake/DaemonGame: move some code to toolchain-pnacl

illwieckz:
I added a commit:

- cmake/DaemonNacl: fix an error message

slipher:
LGTM

PR #971, opened 2023-11-21
WIP: Saigo toolchain

illwieckz:
**DO NOT MERGE THIS**

This PR only exists because yesterday I toyed with the Saigo toolchain to change my mind, and then I thought I can still share the code, and we may have a dedicated thread if people want to talk about it.

**It doesn't work.**

How to use:

Get the Saigo toolchain, see:

- https://github.com/DaemonEngine/Daemon/issues/795#issuecomment-1488442249

Symlink the `native_client/toolchain/linux_x86/saigo_newlib` folder into `external_deps/linux-amd64-default_9`

Then do:

```sh
mkdir build/vms-saigo
cd build/vms-saigo
cmake ../.. -DUSE_NACL_SAIGO=ON -DBUILD_GAME_NACL=ON -DBUILD_GAME_NACL_TARGETS=native \
	-DBUILD_GAME_NATIVE_DLL=OFF -DBUILD_GAME_NATIVE_EXE=OFF \
	-DBUILD_CLIENT=OFF -DBUILD_SERVER=OFF -DBUILD_TTY_CLIENT=OFF \
&& make -j$(nproc)
```

Issue #795, opened 2023-02-08
Consider updating NaCl as an alternative to WebAssembly


slipher:
Google Native Client, or NaCl, is not as dead as has been previously believed. Although support for NaCl in websites and 3rd-party browser extensions has been removed, it lives on in some form. [This bug tracking issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1106311) details the development in 2020-2021 of a C++17-capable NaCl compiler toolchain, codenamed "Saigo". CI-related commits show that it is actively tested on numerous platforms, so it is not just a ChromeOS thing.

After poking around a bit in [Chromium Code Search](https://source.chromium.org/), I still haven't been able to determine what NaCl is being used for. Chromium has a lot of "sandboxed" processes, but in general these are not NaCl-based. But as of now the code is being shipped with Google Chrome: inspection of a Windows install shows a `nacl_irt_x86_64.nexe` and various NaCl-related strings in `chrome.dll`. So the chances of it being available for any platform Daemon would want to support seem good. (Contrast this with standalone WebAssembly engines which are more popular for Linux servers and have not been ported to 32-bit architectures.)

illwieckz:
Some weeks ago I investigated “chromium builds” and only Linux had nacl stuff. Even macOS amd64 chromium did not have any nacl stuff, and I have seen some commits disabling nacl for macOS.

illwieckz:
Some search about “Saigo” and “Native Client” returned me this:

https://chromium.googlesource.com/native_client/nacl-llvm-project-v10/+/refs/heads/saigo-2023-01-02

necessarily-equal:
For the record, the C++ toolchain that we would be most likely to use for wasm is upstream clang. This might be the lowest maintenance toolchain we could ever get.

DolceTriade:
+1, if Google somehow makes the NaCl thing slightly more use friendly, I think it's better than wasm for now. Wasm doesn't really provide us any features beyond even more modern C++.

For example, in the switch from QVM -> NaCl, if we could have supported C++ on QVMs, we probably would have done that instead of using NaCl.

illwieckz:
To be able to upgrade NaCl we need two things:

- a working SDK able to build nexe for hardware platforms we want to run the game on,
  * at least, the SDK should run on Linux amd64 to be able to build and release the game, but to be on par with current SDK such SDK should run on Linux, Windows and macOS amd64;
[...]

PR #981, opened 2023-12-03
Early support of the Saigo toolchain

illwieckz:
Continuing the work done in:

- https://github.com/DaemonEngine/Daemon/pull/971

## How to use

Use this game branch too:

- https://github.com/Unvanquished/Unvanquished/pull/2862

Get the Saigo toolchain, see:

- https://github.com/DaemonEngine/Daemon/issues/795#issuecomment-1488442249

Symlink the `native_client/toolchain/linux_x86/saigo_newlib` folder into `external_deps/linux-amd64-default_10`.

Then do:

```sh
mkdir build/vms-saigo
cd build/vms-saigo
cmake ../.. -DUSE_NACL_SAIGO=ON -DBUILD_GAME_NACL=ON -DBUILD_GAME_NACL_TARGETS=native \
	-DBUILD_GAME_NATIVE_DLL=OFF -DBUILD_GAME_NATIVE_EXE=OFF \
	-DBUILD_CLIENT=OFF -DBUILD_SERVER=OFF -DBUILD_TTY_CLIENT=OFF \
&& make -j$(nproc)
```

## Current status

- ✅️: success
- ❌️: failure
- ❔️: not tested

sgame|amd64|i686|armhf
---|---|---|--
build|✅️|✅️|✅️ 
link|✅️|✅️|✅️ 
run|✅️|❔️|❔️

cgame|amd64|i686|armhf
---|---|---|--
build|✅️|✅️|✅️
link|✅️|✅️|✅️ 
run|✅️|❔️|❔️


illwieckz:
Currently the sgame builds on all architectures.

I have tested the amd64 sgame and it runs.

The cgame builds on all architectures but the final linkage fails.

To fix this we may build freetype, libpng and lua with Saigo (PNaCl provided pexe `.a` for them we can't use).

It is believed this is the last thing required to acheive cgame building.

slipher:
You can find the old repository with scripts and sometimes patches for building the PNaCl libraries [here](https://chromium.googlesource.com/webports/). Also there is a nice [summary](https://web.archive.org/web/20150802043029/https://code.google.com/p/naclports/wiki/PortList).

illwieckz:
At first I hoped this was for building the NaCl libc. 🥲️

illwieckz:
More should be done later. The cgame still doesn't build because we would also need to build some dependencies like zlib with Saigo, but the current status is good enough to be merged and avoid piling merge conflicts with it in the future.

src/common/System.cpp
```diff
@@ -335,8 +337,10 @@ static void CrashHandler(const void* data, size_t n)
 
 void SetupCrashHandler()
 {
+#if !defined(__saigo__)
```
slipher, line 340: Why is this removed? It's the only debugging mechanism we have for NaCl.

cmake/DaemonGame.cmake
```diff
@@ -75,41 +76,80 @@ function(GAMEMODULE)
...
+                    ALWAYS 1
+                )
+            endif()
         endif()
     else()
         if (FORK EQUAL 2)
-            # Put the .nexe and .pexe files in the same directory as the engine
-            set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/..)
+            if(USE_NACL_SAIGO)
+                set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
```
slipher, line 148: Why is this different? We want them to be in the same directory as the engine so that VM type 1 can be used.

slipher:
I think we should be able to avoid zlib by adjusting the Freetype configuration. zlib is only depended on from functions concerning gzip support. Don't think that's something we need but I could be wrong

src/common/System.cpp
illwieckz: Probably because I didn't make it work.

cmake/DaemonGame.cmake
illwieckz: If my memory is right (I wrote this 1 year ago), with PNaCl we can output `sgame.pexe` to the main build directory because it's the same for all platforms, then `pnacl_finalize` translate it to `sgame-amd64.nexe`, `sgame-i686.nexe`, etc.

But with Saigo we have one subproject per platform, and it compiles to nexe directly so every subproject output an `sgame.nexe` file with the exact same name and then every platform would overwrite the file if we output it to the main build directory. So we output it the `sgame.nexe` file to every subproject build directory and then `saigo_finalize` rename them to `sgame-amd64.nexe`, `sgame-i686.nexe`, etc. into the main build directory.

You can see below that `saigo_finalize` gets `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/..` (which was to `${CMAKE_BINARY_DIR}/..`) while `pnacl_finalize` gets `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}` (which was already set to `${CMAKE_BINARY_DIR}/..`).

illwieckz:
> I think we should be able to avoid zlib by adjusting the Freetype configuration. zlib is only depended on from functions concerning gzip support. Don't think that's something we need but I could be wrong

I can build with Saigo freetype without libpng and without zlib in cgame and get it run properly.

illwieckz:
So, this looks ready to me, it is known to be able to build a full game (cgame+sgame) that runs once other PRs taking care of other topic are merged. So these two PRs centered on the toolchain itself are good.

Improvements, like investigating and possibly re-enabling `SetupCrashHandler()` can be done later in future PRs.

cmake/toolchain-saigo.cmake
```diff
@@ -0,0 +1,75 @@
...
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM BOTH)
+
+set(NACL ON)
+
+set(CMAKE_C_FLAGS "")
+set(CMAKE_CXX_FLAGS "")
+
+function(saigo_finalize dir module saigo_arch arch)
```
slipher, line 46: saigo_arch is unused

cmake/DaemonFlags.cmake
```diff
@@ -238,7 +238,9 @@ else()
 	endif()
 
 	# Optimizations.
+	if (NACL AND USE_NACL_SAIGO)
```
slipher, line 241: What's this change?

libs/zlib/gzguts.h
```diff
@@ -19,6 +19,7 @@
 #endif
 
 #include <stdio.h>
+#include <unistd.h>
```
slipher, line 22: You said we can do without zlib so this is not needed, right?

CMakeLists.txt
```diff
@@ -634,9 +634,8 @@ if ( USE_STATIC_LIBS )
     endif()
 endif()
 
-
 # zlib
-if (BUILD_CLIENT OR BUILD_TTY_CLIENT OR BUILD_SERVER OR BUILD_DUMMY_APP OR NACL)
+if (BUILD_CLIENT OR BUILD_TTY_CLIENT OR BUILD_SERVER OR BUILD_DUMMY_APP OR (NACL AND BUILD_CGAME))
```
slipher, line 640: If we don't need zlib condition could be updated to exclude Saigo builds

cmake/DaemonGame.cmake
```diff
@@ -77,6 +77,8 @@ function(GAMEMODULE)
             endforeach(inherited_option)
```
slipher, line 77: The variable `vm` defined a bit above is PNaCl-specific (and redundant with the newly added NACL_VMS_PROJECT).

cmake/DaemonFlags.cmake
illwieckz: Something I committed by mistake when I tested:

Fixed.

libs/zlib/gzguts.h
illwieckz: PNaCl still needs that file, but we are lucky that PNaCl doesn't require this change, so yes we can probably skip this commit, this whole folder will be removed once we drop PNaCl.

cmake/DaemonGame.cmake
illwieckz: Removed.

cmake/toolchain-saigo.cmake
illwieckz: Removed.

CMakeLists.txt

cmake/DaemonPlatform.cmake
```diff
@@ -40,4 +40,17 @@ else()
   message( FATAL_ERROR "Platform not supported" )
 endif()
 
+if (USE_NACL_SAIGO)
+	# Saigo clang reports weird errors when building some cgame and sgame arm nexe with PIE.
+	# Saigo clang crashes when building amd64 cgame with PIE, sgame builds properly though.
+	if (NACL_TARGET STREQUAL "armhf" OR NACL_TARGET STREQUAL "amd64")
```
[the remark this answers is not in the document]
illwieckz: Done.

slipher:
LGTM

Issue #839, opened 2023-04-25
console has some weird offset when built with clang


illwieckz:
Compiler|Version|Clang|System|Architecture|Build|Status|Linux,C14+,Rel
--|--|--|--|--|--|--|--
GCC|9.5.0| |🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|✔️ fine|
GCC|10.4.0| |🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|✔️ fine|
GCC|11.3.0| |🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|✔️ fine|
GCC|12.2.0| |🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|✔️ fine|
ICC|2021.9.0| |🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|✔️ fine|
Clang|11.1.0|11.1.0|🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|✔️ fine|
Clang|13.0.1|13.0.1|🐧️ Linux, Ubuntu|🟪️ i686|🟢️ Rel|✔️ fine|
Clang|13.0.1|13.0.1|🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|✔️ fine|
Clang|13.0.1|13.0.1|🐧️ Linux, Debian|🟦️ arm64|🟢️ Rel|✔️ fine|
AOCC|4.0.0|14.0.6|🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|❌️ bug|yes
Clang|14.0.6|14.0.6|🐧️ Linux, Ubuntu|🟪️ i686|🟢️ Rel|❌️ bug|yes
Clang|14.0.6|14.0.6|🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|❌️ bug|yes
Clang|14.0.6|14.0.6|🐧️ Linux, Debian|🟦️ arm64|🟢️ Rel|❌️ bug|yes
Clang|15.0.7|15.0.7|🐧️ Linux, Ubuntu|🟪️ i686|🟢️ Rel|❌️ bug|yes
Clang|15.0.7|15.0.7|🐧️ Linux, Ubuntu|🟫️ amd64|🟢️ Rel|❌️ bug|yes
[...]

PR #982, opened 2023-12-03
Report what compiler is used to build the game

illwieckz:
Those are old commits that were sitting on my hard drive for a long time. I believe I wrote them when we were investigating the console bug that only happened on clang and clang-based compilers. I assume that enabling PCH with Saigo would now just require to replace the test for `NACL` with a test for the PNaCl compiler.

The first need I wanted to fulfill was to write in `daemon.log` the compiler that was used to build the game, in order to give useful information to people dealing with bug reports.

The second need I wanted to fulfill was to know at build time by reading the build log what CMake was really using when I set some options.

While I was at it, I added custom code to recognize custom compilers like PNaCl and Saigo (in fact at the time I had added Wasi recognition but I removed that before pushing because we really don't know yet which compiler we will use for Wasm).

I also added recognition and compatibility with some other compilers I usually use to test the game, which usually catches different warnings in ways that can be useful to debug things or spot errors. This includes Intel ICC and ICX, AMD AOCC, and Zig. This is not meant to add “support” in a way we would claim to support this, we actually don't have to advertise the fact those compilers can build the game, and this does not give us any obligation to fix the compatibility if one day it breaks. This is only useful for testing, the same way I test the game on various CPU and GPU, I test the build on many compilers.

While doing this I spotted some shortcomings like some flags not being the same for the engine if built within Unvanquished repository or from Daemon repository.

I also spotted that we configure engine platform, compiler even when we don't build the engine, even when we will use another compiler for building the vms.

For those two shortcomings I only added comments and some print to tell when some CMake logs are meaningless, because fixing those shortcomings that exist in our CMake code since forever is out of topic of this PR.

If the compiler is unknown, it reports `Unknown` as compiler name and `Unknown` as compiler version. It is expected the clang version is appended if the unknown compiler is clang-based (like known ones).

See also:

- https://github.com/Unvanquished/Unvanquished/pull/2864

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,16 @@ if (NOT CMAKE_OSX_ARCHITECTURES)
     set(CMAKE_OSX_ARCHITECTURES x86_64)
 endif()
 
+# FIXME: this reports an unused architecture if only vms are built.
+# We may not be able to avoid this for now without doing a huge rewrite
+# of this file.
+# FIXME: this reports an unused compiler if only vms are built.
+# We cannot avoid this for now without doing a huge rewrite of this files
+# since we also set useless flags using informations provided by this.
+if (NOT (BUILD_CLIENT OR BUILD_TTY_CLIENT OR BUILD_SERVER OR BUILD_DUMMY_APP))
+    message(NOTICE "You can safely ignore the following reported architecture, it is not used.")
+    message(NOTICE "You can safely ignore the following reported compilers, they are unused.")
+endif()
 include(DaemonPlatform)
 
 if (Daemon_OUT)
@@ -288,8 +298,17 @@ endif()
 # Compile and link flags
 ################################################################################
 
+# FIXME: This sets unused flags for an unused compiler if only nacl vms
+# are built and we are not yet running the chainloaded build. We cannot
+# avoid this for now or we would have to do a large rewrite of that file.
+# FIXME: This defines the try_flag function that is used in
+# Unvanquished CMakeLists.txt even if the engine is not built,
+# meaning we have to detect useless engine compiler to set
+# useless engine flags when only building vms.
 include(DaemonFlags)
 
+# FIXME: Once engine flags would be only set when building the engine
+# this would be likely unused by vms as they are built by DaemonGame.
 # Warning options (for Daemon only)
 # Note the different scopes used for warning options:
 # * set_c_cxx_flag(xxx) or try_c_cxx_flag(xxx) sets it for all code including dependencies
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -418,8 +418,13 @@ function(ADD_PRECOMPILED_HEADER Target)
     try_c_cxx_flag(PCH "-pch")
 
     if (PCH)
+        # CMAKE_CXX_COMPILER_ARG1 is used with compilers using subcommands.
+        # For example when doing: cmake -D'CMAKE_CXX_COMPILER'='zig;c++'
+        # CMAKE_CXX_COMPILER will be "zig",
+        # CMAKE_CXX_COMPILER_ARG1 will be "c++".
         add_custom_command(OUTPUT "${OBJ_DIR}/${Target}.h.gch"
-            COMMAND ${PNACLPYTHON_PREFIX2} ${CMAKE_CXX_COMPILER} ${Defs} ${Flags} -x c++-header ${Header} -o "${OBJ_DIR}/${Target}.h.gch"
+            COMMAND ${PNACLPYTHON_PREFIX2} ${CMAKE_CXX_COMPILER} ${CMAKE_CXX_COMPILER_ARG1}
+                ${Defs} ${Flags} -x c++-header ${Header} -o "${OBJ_DIR}/${Target}.h.gch"
             DEPENDS ${Header}
             IMPLICIT_DEPENDS CXX ${Header}
         )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -415,7 +415,9 @@ function(ADD_PRECOMPILED_HEADER Target)
     set(Header ${COMMON_DIR}/Common.h)
 
     # Compiler-specific PCH support
-    if (CMAKE_COMPILER_IS_GNUCXX OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR NACL)
+    try_c_cxx_flag(PCH "-pch")
+
+    if (PCH)
         add_custom_command(OUTPUT "${OBJ_DIR}/${Target}.h.gch"
             COMMAND ${PNACLPYTHON_PREFIX2} ${CMAKE_CXX_COMPILER} ${Defs} ${Flags} -x c++-header ${Header} -o "${OBJ_DIR}/${Target}.h.gch"
             DEPENDS ${Header}
@@ -426,7 +428,7 @@ function(ADD_PRECOMPILED_HEADER Target)
         add_dependencies(${Target} ${Target}-pch)
 
         # PNaCl clang doesn't support -include-pch properly
-        if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND NOT NACL)
+        if (NOT DAEMON_CXX_COMPILER_NAME STREQUAL "PNaCl")
             set_property(TARGET ${Target} APPEND PROPERTY COMPILE_OPTIONS "-include-pch;${OBJ_DIR}/${Target}.h.gch")
         else()
             set_property(TARGET ${Target} APPEND PROPERTY COMPILE_OPTIONS "-include;${OBJ_DIR}/${Target}.h;-Winvalid-pch")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -428,10 +428,10 @@ function(ADD_PRECOMPILED_HEADER Target)
         add_dependencies(${Target} ${Target}-pch)
 
         # PNaCl clang doesn't support -include-pch properly
-        if (NOT DAEMON_CXX_COMPILER_NAME STREQUAL "PNaCl")
-            set_property(TARGET ${Target} APPEND PROPERTY COMPILE_OPTIONS "-include-pch;${OBJ_DIR}/${Target}.h.gch")
-        else()
+        if (DAEMON_CXX_COMPILER_NAME STREQUAL "PNaCl")
             set_property(TARGET ${Target} APPEND PROPERTY COMPILE_OPTIONS "-include;${OBJ_DIR}/${Target}.h;-Winvalid-pch")
+        else()
+            set_property(TARGET ${Target} APPEND PROPERTY COMPILE_OPTIONS "-include-pch;${OBJ_DIR}/${Target}.h.gch")
         endif()
     elseif (MSVC)
         # /Fp sets the PCH path used by either of the /Yc and /Yu options.
--- a/cmake/DaemonCompiler.cmake
+++ b/cmake/DaemonCompiler.cmake
@@ -1,0 +1,240 @@
+# Daemon BSD Source Code
+# Copyright (c) 2024, Daemon Developers
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of the <organization> nor the
+#    names of its contributors may be used to endorse or promote products
+#    derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+################################################################################
+# Determine compiler
+################################################################################
+
+# FIXME: Force -W#pragma-messages and -Wno-error
+# In case there is -Wno-#pragma-messages or -Werror in CFLAGS/CXXFLAGS
+
+function(detect_daemon_compiler lang)
+	set(C_NAME "C")
+	set(CXX_NAME "C++")
+	set(C_EXT ".c")
+	set(CXX_EXT ".cpp")
+
+	try_compile(BUILD_RESULT
+		"${CMAKE_BINARY_DIR}"
+		"${DAEMON_DIR}/cmake/DaemonCompiler/DaemonCompiler${${lang}_EXT}"
+		CMAKE_FLAGS CMAKE_OSX_ARCHITECTURES=${CMAKE_OSX_ARCHITECTURES}
+		OUTPUT_VARIABLE BUILD_LOG
+	)
+
+	get_filename_component(compiler_basename "${CMAKE_${lang}_COMPILER}" NAME)
+
+	if (NOT BUILD_RESULT)
+		message(WARNING "Failed to build DaemonCompiler${${lang}_EXT}, relying on CMake builtin detection.")
+		set(compiler_name "Unknown")
+	else()
+		set(BUILD_LOG "\n${BUILD_LOG}\n")
+		string(REGEX REPLACE "\n[^\n]*<REPORT<" "\n" BUILD_LOG "${BUILD_LOG}")
+		string(REGEX REPLACE ">REPORT>[^\n]*\n" "\n" BUILD_LOG "${BUILD_LOG}")
+
+		string(REGEX REPLACE ".*\nDAEMON_COMPILER_NAME=([^\n]*)\n.*" "\\1"
+			compiler_name "${BUILD_LOG}")
+
+		foreach(name GCC;Clang;generic;${compiler_name})
+			set(compatibility_regex ".*\nDAEMON_COMPILER_${name}_COMPATIBILITY=([^\n]*)\n.*")
+			if ("${BUILD_LOG}" MATCHES ${compatibility_regex})
+				string(REGEX REPLACE ${compatibility_regex} "\\1"
+				compiler_${name}_compatibility "${BUILD_LOG}")
+			endif()
+
+			set(version_regex ".*\nDAEMON_COMPILER_${name}_VERSION=([^\n]*)\n.*")
+			if ("${BUILD_LOG}" MATCHES ${version_regex})
+				string(REGEX REPLACE ${version_regex} "\\1"
+				compiler_${name}_version "${BUILD_LOG}")
+			endif()
+
+			set(version_string_regex ".*\nDAEMON_COMPILER_${name}_VERSION_STRING=([^\n]*)\n.*")
+			if ("${BUILD_LOG}" MATCHES ${version_string_regex})
+				string(REGEX REPLACE ${version_string_regex} "\\1"
+					compiler_${name}_version_string "${BUILD_LOG}")
+			endif()
+
+			set(DAEMON_${lang}_COMPILER_${name}_VERSION
+				"${compiler_${name}_version}"
+				PARENT_SCOPE)
+
+			set(DAEMON_${lang}_COMPILER_${name}_COMPATIBILITY
+				"${compiler_${name}_compatibility}"
+				PARENT_SCOPE)
+		endforeach()
+	endif()
+
+	if (compiler_name STREQUAL "Unknown")
+		if (CMAKE_${lang}_COMPILER_ID)
+			set(compiler_name "${CMAKE_${lang}_COMPILER_ID}")
+			# Compiler version is done below.
+		else()
+			message(WARNING "Unknown ${${lang}_NAME} compiler")
+		endif()
+	endif()
+
+	# AOCC
+	if (compiler_Clang_version_string)
+		set(aocc_version_regex ".*CLANG: AOCC_([^ )]+).*")
+		if (compiler_Clang_version_string MATCHES ${aocc_version_regex})
+			set(compiler_name "AOCC")
+			string(REGEX REPLACE ${aocc_version_regex} "\\1"
+				compiler_AOCC_version "${compiler_Clang_version_string}")
+			string(REGEX REPLACE "(.*)-Build.*" "\\1"
+				compiler_AOCC_version "${compiler_AOCC_version}")
+		endif()
+	endif()
+
+	# Zig
+	if (compiler_Clang_version_string)
+		set(zig_version_regex ".*[(]https://github.com/ziglang/zig-bootstrap .*[)]")
+		if (compiler_Clang_version_string MATCHES ${zig_version_regex})
+			set(compiler_name "Zig")
+		endif()
+
+		# Parse “zig version”
+		execute_process(COMMAND "${CMAKE_${lang}_COMPILER}" version
+			OUTPUT_VARIABLE CUSTOM_${lang}_ZIG_OUTPUT
+			RESULT_VARIABLE CUSTOM_${lang}_ZIG_RETURN_CODE
+			ERROR_QUIET
+			OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+		if (NOT CUSTOM_${lang}_ZIG_RETURN_CODE) # Success
+			set(compiler_Zig_version "${CUSTOM_${lang}_ZIG_OUTPUT}")
+		endif()
+	endif()
+
+	# Compilers that use underlying Clang version as their own version.
+	foreach(name in AppleClang)
+		if (compiler_name STREQUAL "${name}")
+			set(compiler_${name}_version "${compiler_Clang_version}")
+		endif()
+	endforeach()
+
+	# Compilers that write the version number at the beginning of the VERSION string.
+	set(string_version_regex "([^ ]+).*")
+	foreach(name in PNaCl)
+		if (compiler_name STREQUAL "${name}")
+			if (compiler_generic_version_string)
+				if (compiler_generic_version_string MATCHES ${string_version_regex})
+					string(REGEX REPLACE ${string_version_regex} "\\1"
+						compiler_${name}_version "${compiler_generic_version_string}")
+				endif()
+			endif()
+		endif()
+	endforeach()
+
+	if (compiler_ARMClang_version_string)
+		# There is no __armclang_patchlevel__ so we should parse __armclang_version__ to get it.
+		if (compiler_ARMClang_version_string MATCHES ${string_version_regex})
+			string(REGEX REPLACE ${string_version_regex} "\\1"
+				compiler_ARMClang_version "${compiler_ARMClang_version_string}")
+		endif()
+	endif()
+
+	if (compiler_ICX_version)
+		# 20240000 becomes 2024.0.0
+		string(REGEX REPLACE "(....)(..)(..)" "\\1.\\2.\\3"
+			compiler_ICX_version "${compiler_ICX_version}")
+		string(REGEX REPLACE "\\.0" "."
+			compiler_ICX_version "${compiler_ICX_version}")
+	endif()
+
+	if (compiler_${compiler_name}_version)
+		set(compiler_version "${compiler_${compiler_name}_version}")
+	elseif (CMAKE_${lang}_COMPILER_VERSION)
+		set(compiler_version "${CMAKE_${lang}_COMPILER_VERSION}")
+	else()
+		set(compiler_version "Unknown")
+		message(WARNING "Unknown ${${lang}_NAME} compiler version")
+	endif()
+
+	set(DAEMON_${lang}_COMPILER_BASENAME "${compiler_basename}" PARENT_SCOPE)
+	set(DAEMON_${lang}_COMPILER_NAME "${compiler_name}" PARENT_SCOPE)
+	set(DAEMON_${lang}_COMPILER_VERSION "${compiler_version}" PARENT_SCOPE)
+endfunction()
+
+message(STATUS "CMake generator: ${CMAKE_GENERATOR}")
+
+foreach(lang C;CXX)
+	set(C_NAME "C")
+	set(CXX_NAME "C++")
+
+	if (MSVC)
+		# Let CMake do the job, it does it very well,
+		# and there is probably no variant to take care about.
+		set(DAEMON_${lang}_COMPILER_NAME "${CMAKE_${lang}_COMPILER_ID}")
+		set(DAEMON_${lang}_COMPILER_VERSION "${CMAKE_${lang}_COMPILER_VERSION}")
+		get_filename_component(DAEMON_${lang}_COMPILER_BASENAME "${CMAKE_${lang}_COMPILER}" NAME)
+	else()
+		detect_daemon_compiler(${lang})
+
+		if (DAEMON_${lang}_COMPILER_Clang_COMPATIBILITY)
+			if (NOT DAEMON_${lang}_COMPILER_NAME STREQUAL "Clang")
+				set(DAEMON_${lang}_COMPILER_EXTENDED_VERSION
+					"${DAEMON_${lang}_COMPILER_VERSION}/clang-${DAEMON_${lang}_COMPILER_Clang_VERSION}")
+			endif()
+		elseif (DAEMON_${lang}_COMPILER_GCC_COMPATIBILITY)
+			if (NOT DAEMON_${lang}_COMPILER_NAME STREQUAL "GCC")
+				# Almost all compilers on Earth pretend to be GCC compatible.
+				# So we first have to check it's really a GCC variant.
+				# Parse “<compiler> -v”
+				execute_process(COMMAND "${CMAKE_${lang}_COMPILER}" -v
+					ERROR_VARIABLE CUSTOM_${lang}_GCC_OUTPUT
+					RESULT_VARIABLE CUSTOM_${lang}_GCC_RETURN_CODE
+					OUTPUT_QUIET)
+
+				if (NOT CUSTOM_${lang}_GCC_RETURN_CODE) # Success
+					# The existence of this string tells us it's a GCC variant.
+					# The version in this string is the same as __VERSION__,
+					# the version of the GCC variant, not the version of the upstream
+					# GCC we are looking for.
+					if ("${CUSTOM_${lang}_GCC_OUTPUT}" MATCHES "\ngcc version ")
+						set(DAEMON_${lang}_COMPILER_EXTENDED_VERSION
+							"${DAEMON_${lang}_COMPILER_VERSION}/gcc-${DAEMON_${lang}_COMPILER_GCC_VERSION}")
+					endif()
+				endif()
+			endif()
+		endif()
+	endif()
+
+	if (NOT DAEMON_${lang}_COMPILER_EXTENDED_VERSION)
+		set(DAEMON_${lang}_COMPILER_EXTENDED_VERSION "${DAEMON_${lang}_COMPILER_VERSION}")
+	endif()
+
+	set(DAEMON_${lang}_COMPILER_STRING
+		"${DAEMON_${lang}_COMPILER_NAME} ${DAEMON_${lang}_COMPILER_EXTENDED_VERSION} ${DAEMON_${lang}_COMPILER_BASENAME}")
+
+	message(STATUS "Detected ${${lang}_NAME} compiler: ${DAEMON_${lang}_COMPILER_STRING}")
+
+	add_definitions(-DDAEMON_${lang}_COMPILER_${DAEMON_${lang}_COMPILER_NAME}=1)
+
+	# Preprocessor definitions containing '#' may not be passed on the compiler
+	# command line because many compilers do not support it.
+	string(REGEX REPLACE "\#" "~" DAEMON_DEFINE_${lang}_COMPILER_STRING "${DAEMON_${lang}_COMPILER_STRING}")
+
+	# Quotes cannot be part of the define as support for them is not reliable.
+	add_definitions(-DDAEMON_${lang}_COMPILER_STRING=${DAEMON_DEFINE_${lang}_COMPILER_STRING})
+endforeach()
--- a/cmake/DaemonCompiler/DaemonCompiler.c
+++ b/cmake/DaemonCompiler/DaemonCompiler.c
@@ -1,0 +1,165 @@
+/*
+===========================================================================
+Daemon BSD Source Code
+Copyright (c) 2024, Daemon Developers
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of the Daemon developers nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS AS IS AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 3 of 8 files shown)


illwieckz:
MinGW is now detected too:

```
-- Detected architecture: amd64
-- CMake generator: Unix Makefiles
-- Detected C compiler: MinGW64 12-posix/gcc-12.0.0 
-- Detected C++ compiler: MinGW64 12-posix/gcc-12.0.0 
```

illwieckz:
The proposed code parses compiler defines. It only parses things like `-v` output if the defines are not usable, for example:

1. zig doesn't set any zig define, we can find some zig word in some version string but that's because it's part of reported repository url used to build the compiler (so a custom build may report another url): \
  `#define __clang_version__ "16.0.1 (https://github.com/ziglang/zig-bootstrap 710c5d12660235bc4eac103a8c6677c61f0a9ded)"`
2. gcc sets many gcc define but almost all compilers mimic gcc and pretend to be gcc and define all thoes gcc defines…

Otherwise the proposed code relies on macro parsing, it does `g++ -E -dM emptyfile.cpp` which dumps all the compiler-generated defines to standard out, this works with both gcc and clang derivatives and some compilers mimicking gcc like icc. I didn't went the way of what I did for architecture detection because we need to do more parsing out of defines, but the idea is the same.

Unlike CMake architecture detection which is totally unreliable, CMake compiler detection is actually not bad and considered well tested, since CMake has to write millions of compilation commands. The reason why I want extra detection is that most of the time, CMake only reports the underlying compiler, which is totally good enough for its task.

CMake some extra detection for some clang-based compilers like Intel ICX and armclang but even if it would not detect them and considered them all being clang I guess CMake would be fine, like CMake is already fine with AMD AOCC or Saigo by simply considering it's clang.

But I want to know more, like make the difference between GCC and MinGW, between Clang, PNaCl, Saigo, Wasi, etc. I also wan to know which version of underlying compiler is used (which is useful when a bug comes after a specific clang version for example).

So let's look at the code.

Here is code that gets clang version found by CMake, we will use that as “underlying” version if we detect a clang variant. We should be able to count on it. Well, it would not work if someone publishes a new compiler variant based on a variant, like if someone ports Saigo patches over ICX it would be reported as IntelLLVM and not Clang so this code will not work.

```cmake
function(detect_custom_clang_version lang file)
	# If CMake already detected the compiler as Clang but we know
	# it's something else.
	if(CUSTOM_${lang}_COMPILER_ID AND CMAKE_${lang}_COMPILER_ID STREQUAL "Clang")
		set(CUSTOM_${lang}_CLANG_VERSION "${CMAKE_${lang}_COMPILER_VERSION}")
		set(CUSTOM_${lang}_CLANG_VERSION "${CUSTOM_${lang}_CLANG_VERSION}" PARENT_SCOPE)
		return()
	endif()
```

Here is probably what CMake does internally, this is a fallback. This is like doing `printf(__clang_version__);` but with extra parsing because some compilers like to add some data after the number.

```cmake
	# Parse “<compiler> -E -dM <source file>”
	execute_process(COMMAND "${CMAKE_${lang}_COMPILER}" ${CUSTOM_${lang}_COMPILER_SUBCOMMAND} -E -dM "${file}"
		OUTPUT_VARIABLE CUSTOM_${lang}_CLANG_OUTPUT
		RESULT_VARIABLE CUSTOM_${lang}_RETURN_CODE
		ERROR_QUIET)

	if (NOT CUSTOM_${lang}_RETURN_CODE) # Success
		if ("${CUSTOM_${lang}_CLANG_OUTPUT}" MATCHES "\#define __clang_version__ ")
			string(REGEX REPLACE ".*#define __clang_version__ \"([^ ]+)[\" ]*.*" "\\1" CUSTOM_${lang}_CLANG_VERSION "${CUSTOM_${lang}_CLANG_OUTPUT}")
			set(CUSTOM_${lang}_CLANG_VERSION "${CUSTOM_${lang}_CLANG_VERSION}" PARENT_SCOPE)
		endif()
	endif()
endfunction()
```

Here is code that gets GCC version found by CMake, we will use that as “underlying” version if we detect a GCC variant. We should be able to count on it.

```cmake
function(detect_custom_gcc_version lang file)
	# If CMake already detected the compiler as GCC but we know
	# it's something else.
	if(CUSTOM_${lang}_COMPILER_ID AND CMAKE_${lang}_COMPILER_ID STREQUAL "GNU")
		set(CUSTOM_${lang}_GCC_VERSION "${CMAKE_${lang}_COMPILER_VERSION}")
		set(CUSTOM_${lang}_GCC_VERSION "${CUSTOM_${lang}_GCC_VERSION}" PARENT_SCOPE)
		return()
	endif()
```

Here is probably what CMake does internally, this is a fallback. First we parse `gcc -v` output because there is no define known to be unique to GCC, then we do something like `printf("%d.%d.%d", __GNUC__, __GNUC_MINOR__, __GNUC_PATCHLEVEL__);`.

```cmake
	# Almost all compilers on Earth define __GNUC__, __GNUC_MINOR__, and __GNUC_PATCHLEVEL__,
	# So we first have to check it's really a GCC variant.
	# Parse “<compiler> -v”
	execute_process(COMMAND "${CMAKE_${lang}_COMPILER}" -v
		ERROR_VARIABLE CUSTOM_${lang}_GCC_OUTPUT
		RESULT_VARIABLE CUSTOM_${lang}_RETURN_CODE
		OUTPUT_QUIET)

	if (NOT CUSTOM_${lang}_RETURN_CODE) # Success
		# The existence of this string tells us it's a GCC variant.
		# The version in this string is the same as __VERSION__,
		# the version of the GCC variant, not the version of the upstream
		# GCC we are looking for.
		if ("${CUSTOM_${lang}_GCC_OUTPUT}" MATCHES "\ngcc version ")
			# Parse “<compiler> -E -dM <source file>”
			# No subcommand implemented for now, there may be no usage.
			execute_process(COMMAND "${CMAKE_${lang}_COMPILER}" -E -dM "${file}"
			OUTPUT_VARIABLE CUSTOM_${lang}_GCC_OUTPUT
			RESULT_VARIABLE CUSTOM_${lang}_RETURN_CODE
			ERROR_QUIET)

			if (NOT CUSTOM_${lang}_RETURN_CODE) # Success
				string(REGEX REPLACE ".*#define __GNUC__ ([^ \n]+).*" "\\1" CUSTOM_${lang}_GCC_MAJOR "${CUSTOM_${lang}_GCC_OUTPUT}")
				string(REGEX REPLACE ".*#define __GNUC_MINOR__ ([^ \n]+).*" "\\1" CUSTOM_${lang}_GCC_MINOR "${CUSTOM_${lang}_GCC_OUTPUT}")
				string(REGEX REPLACE ".*#define __GNUC_PATCHLEVEL__ ([^ \n]+).*" "\\1" CUSTOM_${lang}_GCC_PATCHLEVEL "${CUSTOM_${lang}_GCC_OUTPUT}")

				set(CUSTOM_${lang}_GCC_VERSION "${CUSTOM_${lang}_GCC_MAJOR}.${CUSTOM_${lang}_GCC_MINOR}.${CUSTOM_${lang}_GCC_PATCHLEVEL}")
				set(CUSTOM_${lang}_GCC_VERSION "${CUSTOM_${lang}_GCC_VERSION}" PARENT_SCOPE)
				return()
			endif()
		endif()
	endif()
endfunction()
```

Here is the compiler detection, it requests the compiler to dump defines and parses it. This is similar to what we do for arch detection. It's like doing `#ifdef __pnacl__`, `printf(__VERSION__);`, etc.

```cmake
function(detect_custom_compiler_id_version lang file)
	# Parse “<compiler> -E -dM <source file>”
	execute_process(COMMAND "${CMAKE_${lang}_COMPILER}" -E -dM "${file}"
		OUTPUT_VARIABLE CUSTOM_${lang}_COMPILER_OUTPUT
		RESULT_VARIABLE CUSTOM_${lang}_RETURN_CODE
		ERROR_QUIET)

	if (NOT CUSTOM_${lang}_RETURN_CODE) # Success
		# PNaCL
		if ("${CUSTOM_${lang}_COMPILER_OUTPUT}" MATCHES "\#define __pnacl__ 1")
			set(CUSTOM_${lang}_COMPILER_ID "PNaCl" PARENT_SCOPE)

			string(REGEX REPLACE ".*#define __VERSION__ \"([^ ]+).*" "\\1" CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_OUTPUT}")
			set(CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_VERSION}" PARENT_SCOPE)

			return()
		endif()

		# Saigo
		if ("${CUSTOM_${lang}_COMPILER_OUTPUT}" MATCHES "\#define __saigo__ 1")
			set(CUSTOM_${lang}_COMPILER_ID "Saigo" PARENT_SCOPE)

			string(REGEX REPLACE ".*#define __VERSION__ \"Clang ([^ \"]+).*" "\\1" CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_OUTPUT}")
			set(CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_VERSION}" PARENT_SCOPE)

			return()
		endif()

		# AOCC
		if ("${CUSTOM_${lang}_COMPILER_OUTPUT}" MATCHES "CLANG: AOCC")
			set(CUSTOM_${lang}_COMPILER_ID "AOCC" PARENT_SCOPE)

			string(REGEX REPLACE ".*CLANG: AOCC_([^ \"]+).*" "\\1" CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_OUTPUT}")
			set(CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_VERSION}" PARENT_SCOPE)

			return()
		endif()

		# MinGW64
		# This must be tested before MinGW32 since MinGW64 also defines "__MINGW32__".
		if ("${CUSTOM_${lang}_COMPILER_OUTPUT}" MATCHES "\#define __MINGW64__ 1")
			set(CUSTOM_${lang}_COMPILER_ID "MinGW64" PARENT_SCOPE)

			string(REGEX REPLACE ".*#define __VERSION__ \"([^ \"]+).*" "\\1" CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_OUTPUT}")
			set(CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_VERSION}" PARENT_SCOPE)

			return()
		endif()

		# MinGW32
		if ("${CUSTOM_${lang}_COMPILER_OUTPUT}" MATCHES "\#define __MINGW32__ 1")
			set(CUSTOM_${lang}_COMPILER_ID "MinGW32" PARENT_SCOPE)

			string(REGEX REPLACE ".*#define __VERSION__ \"([^ \"]+).*" "\\1" CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_OUTPUT}")
			set(CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_VERSION}" PARENT_SCOPE)

			return()
		endif()
	endif()

	# Parse “<compiler> --help”
	# There is a bug in cmake: if we set CMAKE_C_COMPILER to "zig:cc",
	# only "zig" is returned when using the ${CMAKE_C_COMPILER} variable,
	# so here the command will be "zig --help".
	execute_process(COMMAND "${CMAKE_${lang}_COMPILER}" --help
		OUTPUT_VARIABLE CUSTOM_${lang}_COMPILER_OUTPUT
		RESULT_VARIABLE CUSTOM_${lang}_RETURN_CODE
		ERROR_QUIET)

	if (NOT CUSTOM_${lang}_RETURN_CODE) # Success
		# Zig
		if ("${CUSTOM_${lang}_COMPILER_OUTPUT}" MATCHES ": zig ")
			set(CUSTOM_${lang}_COMPILER_ID "Zig" PARENT_SCOPE)

			execute_process(COMMAND "${CMAKE_${lang}_COMPILER}" version OUTPUT_VARIABLE CUSTOM_${lang}_COMPILER_VERSION)
			string(STRIP "${CUSTOM_${lang}_COMPILER_VERSION}" CUSTOM_${lang}_COMPILER_VERSION)
			set(CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_VERSION}" PARENT_SCOPE)

			# There is a bug in CMake: if we set CMAKE_C_COMPILER to "zig;cc",
			# only "zig" is returned when using the ${CMAKE_C_COMPILER} variable.
			set(C_SUBCOMMAND "cc")
			set(CXX_SUBCOMMAND "c++")

			set(CUSTOM_${lang}_COMPILER_SUBCOMMAND "${${lang}_SUBCOMMAND}")
			set(CUSTOM_${lang}_COMPILER_SUBCOMMAND "${CUSTOM_${lang}_COMPILER_SUBCOMMAND}" PARENT_SCOPE)

			return()
		endif()
	endif()
endfunction()
```

Then a code running those functions with both C and C++ compilers:

```cmake
function(detect_custom_compiler lang)
	set(C_EXT ".c")
	set(CXX_EXT ".cpp")

	string(RANDOM RANDOM_SUFFIX)
	set(COMPILER_TEST_FILE "${PROJECT_BINARY_DIR}/test_compiler_${RANDOM_SUFFIX}${${lang}_EXT}")
	file(TOUCH "${COMPILER_TEST_FILE}")

	detect_custom_compiler_id_version("${lang}" "${COMPILER_TEST_FILE}")
	detect_custom_clang_version("${lang}" "${COMPILER_TEST_FILE}")
	detect_custom_gcc_version("${lang}" "${COMPILER_TEST_FILE}")

	file(REMOVE "${COMPILER_TEST_FILE}")

	set(CUSTOM_${lang}_COMPILER_ID "${CUSTOM_${lang}_COMPILER_ID}" PARENT_SCOPE)
	set(CUSTOM_${lang}_COMPILER_VERSION "${CUSTOM_${lang}_COMPILER_VERSION}" PARENT_SCOPE)
	set(CUSTOM_${lang}_COMPILER_SUBCOMMAND "${CUSTOM_${lang}_COMPILER_SUBCOMMAND}" PARENT_SCOPE)

	set(CUSTOM_${lang}_CLANG_VERSION "${CUSTOM_${lang}_CLANG_VERSION}" PARENT_SCOPE)
	set(CUSTOM_${lang}_GCC_VERSION "${CUSTOM_${lang}_GCC_VERSION}" PARENT_SCOPE)
endfunction()
```

illwieckz:
I tried to implement it a bit like I did for architectures:

```c++
#define STRING(s) #s
#define XSTRING(s) STRING(s)

#define REPORT(key, value) \
	"##REPORT## ##[## DAEMON_" key " ##|## " value " ##]##"
#define REPORT_VERSION_3(name, major, minor, patch) \
	REPORT(name "_VERSION", XSTRING(major) "." XSTRING(minor) "." XSTRING(patch))
#define REPORT_VERSION_2(name, major, minor) \
	REPORT(name "_VERSION", XSTRING(major) "." XSTRING(minor))
#define REPORT_VERSION_1(name, major) \
	REPORT(name "_VERSION", XSTRING(major))
#define REPORT_VERSION_STRING(name, value) \
	REPORT(name "_VERSION_STRING", value)
#define REPORT_COMPATIBILITY(name) \
	REPORT(name "_COMPATIBILITY", "YES")
#define REPORT_NAME(name) \
	REPORT("COMPILER_NAME", name)

// GCC

#if defined(__GNUC__)
	#pragma message(REPORT_COMPATIBILITY("GCC"))
#endif

#if defined(__GNUC__) && defined(__GNUC_MINOR__) && defined(__GNUC_PATCHLEVEL__)
	#pragma message(REPORT_VERSION_3("GCC", __GNUC__, __GNUC_MINOR__, __GNUC_PATCHLEVEL__))
#endif

// Clang

#if defined(__clang__)
	#pragma message(REPORT_COMPATIBILITY("CLANG"))
#endif

#if defined(__clang_major__) && defined(__clang_minor__) && defined(__clang_patchlevel__)
	#pragma message(REPORT_VERSION_3("CLANG", __clang_major__, __clang_minor__, __clang_patchlevel__))
#endif

#if defined(__clang_version__)
	#pragma message(REPORT_VERSION_STRING("CLANG", __clang_version__))
#endif

// ICC

#if defined(__INTEL_COMPILER) && defined(__INTEL_COMPILER_UPDATE)
	#pragma message(REPORT_VERSION_2("Intel", __INTEL_COMPILER, __INTEL_COMPILER_UPDATE))
#elif defined(_ICC)
	#pragma message(REPORT_VERSION_1("Intel", __ICC))
#endif

// ICX

#if defined(__INTEL_CLANG_COMPILER)
	// This would require extra parsing since it's the form 20240000 for 2024.0.0
	#pragma message(REPORT_VERSION("ICX", __INTEL_CLANG_COMPILER)
#elif defined(__INTEL_LLVM_COMPILER)
	// This would require extra parsing since it's the form 20240000 for 2024.0.0
	#pragma message(REPORT_VERSION("ICX", __INTEL_LLVM_COMPILER)
#endif

// Generic

#if defined(__VERSION__)
	#pragma message(REPORT_VERSION_STRING("COMPILER", __VERSION__))
#endif

// Selection

// There is no Zig specific define.

// There is no AOCC specific define, we should parse other defines:
// #define __VERSION__ "AMD Clang 14.0.6 (CLANG: AOCC_4.0.0-Build#434 2022_10_28)"
// #define __clang_version__ "14.0.6 (CLANG: AOCC_4.0.0-Build#434 2022_10_28)"

#if defined(__INTEL_COMPILER) || defined(__ICC)
	#pragma message(REPORT_NAME("ICC")) // Intel
#elif defined(__INTEL_CLANG_COMPILER) || defined(__INTEL_LLVM_COMPILER)
	#pragma message(REPORT_NAME("ICX")) // IntelLLVM
#elif defined(__wasi__)
	#pragma message(REPORT_NAME("WASI"))
#elif defined(__saigo__)
	#pragma message(REPORT_NAME("Saigo"))
#elif defined(__pnacl__)
	#pragma message(REPORT_NAME("PNaCl"))
#elif defined(__MINGW64__) || defined(__MINGW32__)
	#pragma message(REPORT_NAME("MingGW"))
#elif defined(__clang__)
	#pragma message(REPORT_NAME("Clang"))
#elif defined(__GNUC__)
	#pragma message(REPORT_NAME("GCC")) // GNU
#else
	#pragma message(REPORT_NAME("Unknown"))
#endif
```

This mostly works:

```
$ icpc -o /dev/null detect.cpp 2>&1 | grep '##REPORT##' | sed -e 's/.*##REPORT## ##\[## //;s/ ##|## /: /;s/ ##]##.*//'
DAEMON_GCC_COMPATIBILITY: YES
DAEMON_GCC_VERSION: 13.2.0
DAEMON_Intel_VERSION: 2021.10
DAEMON_COMPILER_VERSION_STRING: Intel(R) C++ g++ 13.2 mode
DAEMON_COMPILER_NAME: ICC
```

```
$ /opt/intel/oneapi/compiler/latest/bin/icpx 2>&1 | grep '##REPORT##' | sed -e 's/.*##REPORT## ##\[## //;s/ ##|## /: /;s/ ##]##.*//'
DAEMON_GCC_COMPATIBILITY: YES
DAEMON_GCC_VERSION: 4.2.1
DAEMON_CLANG_COMPATIBILITY: YES
DAEMON_CLANG_VERSION: 17.0.0
DAEMON_CLANG_VERSION_STRING: 17.0.0 (icx 2024.0.0.20231017)
DAEMON_COMPILER_VERSION_STRING: Intel(R) oneAPI DPC++/C++ Compiler 2024.0.0 (2024.0.0.20231017)
DAEMON_COMPILER_NAME: ICX
```

```
$ wasi-sdk/bin/clang++ -o /dev/null detect.cpp 2>&1 | grep '##REPORT##' | sed -e 's/.*##REPORT## ##\[## //;s/ ##|## /: /;s/ ##]##.*//'
DAEMON_GCC_COMPATIBILITY: YES
DAEMON_GCC_VERSION: 4.2.1
DAEMON_CLANG_COMPATIBILITY: YES
DAEMON_CLANG_VERSION: 16.0.0
DAEMON_CLANG_VERSION_STRING: 16.0.0 
DAEMON_COMPILER_VERSION_STRING: Clang 16.0.0
DAEMON_COMPILER_NAME: WASI
```

But for some reasons `__pnacl__` and `__saigo__` are no set:

```
$ saigo_newlib/bin/clang++ -o /dev/null detect.cpp 2>&1 | grep '##REPORT##' | sed -e 's/.*##REPORT## ##\[## //;s/ ##|## /: /;s/ ##]##.*//'
DAEMON_GCC_COMPATIBILITY: YES
DAEMON_GCC_VERSION: 4.2.1
DAEMON_CLANG_COMPATIBILITY: YES
DAEMON_CLANG_VERSION: 18.0.0
DAEMON_CLANG_VERSION_STRING: 18.0.0 (https://chromium.googlesource.com/a/native_client/nacl-llvm-project-v10.git 65262d842b48c9888e9e2f7eedd1ee3e2ee1b0a7)
DAEMON_COMPILER_VERSION_STRING: Clang 18.0.0 (https://chromium.googlesource.com/a/native_client/nacl-llvm-project-v10.git 65262d842b48c9888e9e2f7eedd1ee3e2ee1b0a7)
DAEMON_COMPILER_NAME: Clang
```

```
$ pnacl/bin/clang++ -o /dev/null detect.cpp 2>&1 | grep '##REPORT##' | sed -e 's/.*##REPORT## ##\[## //;s/ ##|## /: /;s/ ##]##.*//'
DAEMON_GCC_COMPATIBILITY: YES
DAEMON_GCC_VERSION: 4.2.1
DAEMON_CLANG_COMPATIBILITY: YES
DAEMON_CLANG_VERSION: 3.6.0
DAEMON_CLANG_VERSION_STRING: 3.6.0 (https://chromium.googlesource.com/a/native_client/pnacl-clang.git 96b3da27dcefc9d152e51cf54280989b2206d789) (https://chromium.googlesource.com/a/native_client/pnacl-llvm.git d0089f0b008e03cfd141f05c80e3b628c2df75c1)
DAEMON_COMPILER_VERSION_STRING: 4.2.1 Compatible Clang 3.6.0 (https://chromium.googlesource.com/a/native_client/pnacl-clang.git 96b3da27dcefc9d152e51cf54280989b2206d789) (https://chromium.googlesource.com/a/native_client/pnacl-llvm.git d0089f0b008e03cfd141f05c80e3b628c2df75c1)
DAEMON_COMPILER_NAME: Clang
```

I don't know way, I now fail to get `__saigo__`, `__pnacl__`, etc, even with `<compiler> -E -dM <emptyfile>`. 🤷‍♀️️

illwieckz:
Ah no I feel stupid, I need to use the target compilers, like `pnacl-clang++` or `x86_64-nacl-clang`.

slipher:
> But I want to know more, like make the difference between GCC and MinGW, between Clang, PNaCl, Saigo, Wasi, etc.

CMake already knows about MinGW. We have various `if (MINGW)` in our build. As for NaCl compilers, those can only be used via our custom toolchain file so we should just set needed info in the toolchain file.

> I also wan to know which version of underlying compiler is used (which is useful when a bug comes after a specific clang version for example).

Making build-time decisions based on a version could have some utility, but not enough to be worth it. If you just want to know what version something was built with you can save the version string without parsing it.

illwieckz:
The underlying version is useful, not for build-time decision (I simply concatenate it so it's not usable by CMake), but for bisecting bug that only occurs with specific compiler versions like:

- https://github.com/DaemonEngine/Daemon/issues/839

In this issue I used such compiler detection to make sure the bug started to appear when a compiler was based on clang > 13.

illwieckz:
The primary purpose of such detection is to print compiler versions in `daemon.log` for when we get bug reports.

illwieckz:
One thing that can be used for decision is like “if gcc-based or clang-based (but not pnacl), enable PCH”.

illwieckz:
I pushed a different implementation, similar to the one I did to detect architectures (i.e. relying on preprocesor to execute some `#ifdef`, etc.).

illwieckz:
Note that `DaemonCompiler.c` and `DaemonCompiler.cpp` are the exact same file, CMake is annoying because it requires the file extension, I'll look for a solution without file duplication later.

illwieckz:
I rebased on current `master`, this looks ready to me.

I find it very useful when tracking down issues like:

- https://github.com/DaemonEngine/Daemon/issues/839

This is especially useful when printing the underlying compiler version of a fork (it may not be that easy to get it otherwise).

When building it prints lines like this:

```
-- Detected C compiler: GCC 13.2.0 gcc
-- Detected C++ compiler: GCC 13.2.0 g++
```

```
-- Detected C compiler: Clang 16.0.6 clang
-- Detected C++ compiler: Clang 16.0.6 clang++
```

```
-- Detected C compiler: AOCC 4.2.0/clang-16.0.3 clang
-- Detected C++ compiler: AOCC 4.2.0/clang-16.0.3 clang++
```

```
-- Detected C compiler: ICX 2024.1.0/clang-18.0.0 icx
-- Detected C++ compiler: ICX 2024.1.0/clang-18.0.0 icpx
```

```
-- Detected C compiler: MinGW 12.0.0/gcc-12.0.0 x86_64-w64-mingw32-gcc
-- Detected C++ compiler: MinGW 12.0.0/gcc-12.0.0 x86_64-w64-mingw32-g++
```

```
-- Detected C compiler: PNaCl 4.2.1/clang-3.6.0 pnacl-clang
-- Detected C++ compiler: PNaCl 4.2.1/clang-3.6.0 pnacl-clang++
```

When running the game it prints lines like this:

```
Unvanquished 0.54.1 Linux amd64 (GCC 13.2.0 g++) Apr 30 2024
SGame Linux amd64 (GCC 13.2.0 g++) Apr 30 2024
CGame Linux amd64 (GCC 13.2.0 g++) Apr 30 2024
```

```
Unvanquished 0.54.1 Linux amd64 (Clang 16.0.6 clang++) Apr 30 2024
SGame Linux amd64 (Clang 16.0.6 clang++) Apr 30 2024
CGame Linux amd64 (Clang 16.0.6 clang++) Apr 30 2024
```

```
Unvanquished 0.54.1 Linux amd64 (AOCC 4.2.0/clang-16.0.3 clang++) Apr 30 2024
SGame Linux amd64 (AOCC 4.2.0/clang-16.0.3 clang++) Apr 30 2024
CGame Linux amd64 (AOCC 4.2.0/clang-16.0.3 clang++) Apr 30 2024
```

```
Unvanquished 0.54.1 Linux amd64 (ICX 2024.1.0/clang-18.0.0 icpx) Apr 30 2024
SGame Linux amd64 (ICX 2024.1.0/clang-18.0.0 icpx) Apr 30 2024
CGame Linux amd64 (ICX 2024.1.0/clang-18.0.0 icpx) Apr 30 2024
```

```
Unvanquished 0.54.1 Windows amd64 (MinGW 12.0.0/gcc-12.0.0 x86_64-w64-mingw32-g++) Apr 30 2024
SGame NaCl nacl (PNaCl 4.2.1/clang-3.6.0 pnacl-clang++) Apr 30 2024
CGame NaCl nacl (PNaCl 4.2.1/clang-3.6.0 pnacl-clang++) Apr 30 2024
```

illwieckz:
The `__VERSION__` string is not reliable. For example AppleClang:

```
4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)
```

The `4.2.1` number at the very beginning of the string is just the `__GNUC__` one, the GCC version AppleClang pretends to be. It has nothing to do with the compiler version at all (neither Apple version, neither Clang version), and while `LLVM 10.0.1` is the actual LLVM/Clang version, and the number in parentheses `clang-1001.0.46.4` only means something internal to Apple, it's not an Clang version.

For example AOCC (AMD compiler):

```
16.0.3 (CLANG: AOCC_4.2.0-Build#89 2023_12_13)
```

The `16.0.3` number at the very beginning of the string is the actual LLVM/Clang version this compiler is based on (not the one it claims to be compatible with, and definitely not an emulated GCC version). The actual version string of the compiler is in parentheses.
The `16.0.3`

For example ICC (historical Intel compiler): 

```
Intel(R) C++ g++ 13.2 mode
```

The `13.2` number is just the `__GNUC__` one, the GCC version ICC pretends to be. The actual version number of the compiler (`2021.10`) is not written at all.

For example ICX (latest Intel compiler):

```
18.0.0 (icx 2024.1.0.20240308)
```

This Clang-based compiler `__VERSION__` string format is basically similar to what does AMD with Clang-based AOCC, but not what Apple does with Clang-based AppleClang…

So basically just with 4 compilers we have:

- `__GNUC__ Compatible Apple LLVM ___clang_major__.__clang_minor__.__clang_patchlevel__ (clang-__clang_major____clang_minor____clang_patchlevel__.<custom>)`
- `__clang_major__.__clang_minor__.__clang_patchlevel__ (CLANG: AOCC_<custom> <custom>)`
- `Intel(R) C++ g++ __GNUC__ mode`
- `__clang_major__.__clang_minor__.__clang_patchlevel__ (icx <custom>.<custom>)`

All but one are based on Clang, none are based on GCC.

> just grab the version string, which requires much less code to implement

This whole code is written **because** the `__VERSION__` string doesn't do the job, at all.

Even AppleClang it actually and intentionally lying to the user, and if ICC ctually discloses it is a lie, it doesn't tell the truth.

> can detect compilers we don't know about

- This code fallbacks on CMake own detection if it fails to detect something, this is actually used for MSVC because unlike some other compiler, CMake will always do a better job than me on that part.
- This reports the underlying GCC or Clang versions of compilers.

And if you talk about the `<compiler> -v` output string, while it is less lying than the `__VERSION__` string, it also has it's own problem, for example ICC is printing a warning when calling it. You actually have to know the compiler is ICC beforehand to set a flag to disable the warning, and if you don't disable the warning but print the first line anyway, you actually print the warning, not the compiler version.

slipher:
I've never heard of `__VERSION__`. Indeed I am talking about requesting the version from the command line. I don't think this has a "lying" problem. I tried ICC, ICX, and MSVC in the [comment](https://github.com/DaemonEngine/Daemon/pull/982#issuecomment-1838248692) which I already linked in the previous post. 

I didn't see a warning first with ICX but that could be because Godbolt always sets an input file. (Or it could be because the site interleaves stdout and stderr in a possibly random order as there is a warning further down.) Shouldn't warnings be on stderr and --version on stdout though? Then it wouldn't be a problem. MSVC does put its version on stderr rather than stdout but that should probably be a special case anyway.

illwieckz:
I'm not against printing a line with the output of `<compiler> -v` but this would be in addition to this, really.

Maybe I misnamed the PR. This does much more than printing the compiler version.

It makes us aware, not only in reading game log, but also in writing CMake code, if the compiler is GCC-based or Clang-based, it also makes the difference between GCC and MinGW, and other things like that.

This just brings more compiler awareness within CMake than what CMake provides.

This is also why even if we would only print the output of `<compiler> -v`, I would implement this and keep this.

As an example, I can now do this:

```cmake
if (DAEMON_CXX_COMPILER_NAME STREQUAL "PNaCl")
```

Instead of this:

```cmake
if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND NACL)
```

And this code will not have to change in my Saigo branch, because we know the compiler for real.

Without this code, while transitionning to Saigo, we would have to write this line this way:

```cmake
if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND NACL AND NOT USE_NACL_SAIGO)
```

This makes things much much easier in multiple ways.

Also, the lines printed by this code are actually parseable, even for human it makes it easier to have common patterns to read.

I'm not against printing the ouput of `<compiler> -v`, it can even bring more informations (compiler build number and build date for some), but this would not replace what is implemented here that is meant to bring as less ambiguous information as possible.

slipher:
CMake already knows about GCC, Clang, and MinGW. We use the `MINGW` variable often. 

The toolchain files for NaCl toolchains are written by us ourselves so we can simply write any needed information in there. We can add e.g. `NACL_PNACL` or `NACL_SAIGO` variables. Then we can use them like `if (NACL_SAIGO)` etc. (An undefined variable evaluates to false). 

BTW you should be using substring tests like `CMAKE_CXX_COMPILER_ID MATCHES "Clang"` instead of `CMAKE_CXX_COMPILER_ID STREQUAL "Clang"` because the compiler ID thing distinguishes [several variants of Clang](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html). I think the boolean variable approach is actually better than defining a single compiler type since you can have things like the `clang-cl` driver which is a Clang, but also `MSVC` is true.

illwieckz:
I don't really trust https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html
For example it claims to know about ARMClang, but truth is it does not.

> CMake already knows about MinGW. We use the `MINGW` variable often.

And then what? For printing a compiler string we build the string from that? That would just lead to the third implementation of this PR!

The current implementation in that PR is already the second implementation as the first one was parsing compiler output.

illwieckz:
> BTW you should be using substring tests like `CMAKE_CXX_COMPILER_ID MATCHES "Clang"` instead of `CMAKE_CXX_COMPILER_ID STREQUAL "Clang"`

Actually CMake names ICX `IntelLLVM`, not `IntelClang`. The `*Clang` match is not reliable.

slipher:
> > CMake already knows about MinGW. We use the `MINGW` variable often.
> 
> And then what? For printing a compiler string we build the string from that? That would just lead to the third implementation of this PR!

No. My suggestion was to use the `--version` string for making the string to display at runtime. `if (MINGW)` is used for making build-time decisions and works fine for that.



> > BTW you should be using substring tests like `CMAKE_CXX_COMPILER_ID MATCHES "Clang"` instead of `CMAKE_CXX_COMPILER_ID STREQUAL "Clang"`
> 
> Actually CMake names ICX `IntelLLVM`, not `IntelClang`. The `*Clang` match is not reliable.

Someone claimed that the results you get for this depend on the version you put in `cmake_minimum_required` and will be according to the documentation for that version.

DolceTriade (commented):
At the risk of throwing additional fuel on the fire...

We could use https://github.com/sbellus/json-cmake to output JSON so the parsing could become much simpler....

illwieckz:
> At the risk of throwing additional fuel on the fire...
> 
> We could use https://github.com/sbellus/json-cmake to output JSON so the parsing could become much simpler....

This can't help, all we can do is some compiler's `pragma message`.

We can't produce json with compiler's `pragma message`.

What the compiler outputs looks like this:

```
cmake/DaemonCompiler/DaemonCompiler.c:52:10: warning: <REPORT<DAEMON_COMPILER_GCC_COMPATIBILITY=ON>REPORT> [-W#pragma-messages]
   52 |         #pragma message(REPORT_COMPATIBILITY("GCC"))
      |                 ^
cmake/DaemonCompiler/DaemonCompiler.c:56:10: warning: <REPORT<DAEMON_COMPILER_GCC_VERSION=4.2.1>REPORT> [-W#pragma-messages]
   56 |         #pragma message(REPORT_VERSION_3("GCC", __GNUC__, __GNUC_MINOR__, __GNUC_PATCHLEVEL__))
      |                 ^
cmake/DaemonCompiler/DaemonCompiler.c:62:10: warning: <REPORT<DAEMON_COMPILER_Clang_COMPATIBILITY=ON>REPORT> [-W#pragma-messages]
   62 |         #pragma message(REPORT_COMPATIBILITY("Clang"))
      |                 ^
cmake/DaemonCompiler/DaemonCompiler.c:66:10: warning: <REPORT<DAEMON_COMPILER_Clang_VERSION=19.0.0>REPORT> [-W#pragma-messages]
   66 |         #pragma message(REPORT_VERSION_3("Clang", __clang_major__, __clang_minor__, __clang_patchlevel__))
      |                 ^
cmake/DaemonCompiler/DaemonCompiler.c:70:10: warning: <REPORT<DAEMON_COMPILER_Clang_VERSION_STRING=19.0.0git (https://chromium.googlesource.com/a/native_client/nacl-llvm-project-v10.git e25355fddbdece2ef08747ead05b7f69f3bc6dca)>REPORT> [-W#pragma-messages]
   70 |         #pragma message(REPORT_VERSION_STRING("Clang", __clang_version__))
      |                 ^
cmake/DaemonCompiler/DaemonCompiler.c:109:10: warning: <REPORT<DAEMON_COMPILER_generic_VERSION_STRING=Clang 19.0.0git (https://chromium.googlesource.com/a/native_client/nacl-llvm-project-v10.git e25355fddbdece2ef08747ead05b7f69f3bc6dca)>REPORT> [-W#pragma-messages]
  109 |         #pragma message(REPORT_VERSION_STRING("generic", __VERSION__))
      |                 ^
cmake/DaemonCompiler/DaemonCompiler.c:139:10: warning: <REPORT<DAEMON_COMPILER_NAME=Saigo>REPORT> [-W#pragma-messages]
  139 |         #pragma message(REPORT_NAME("Saigo"))
      |                 ^
7 warnings generated.
```

illwieckz:
> Someone claimed that the results you get for this depend on the version you put in `cmake_minimum_required` and will be according to the documentation for that version.

I never seen it, and even if it was true it would be a motivation to do it ourselves instead.

> No. My suggestion was to use the `--version` string for making the string to display at runtime. `if (MINGW)` is used for making build-time decisions and works fine for that.

I'm not against adding the output of `--version` to some define but that would be to be as an extra to this, not as a replacement to this. Also to properly call `--version` one may have to know the compiler first. For example ICC now requires to do `--diag-disable=10441 --version` otherwise the first line is a warning that is printed before the version line. I don't know what is the command line for MSVC but I would not be surprised if the option is not `--version`. For zig, what do we want, `zig cc --version` (reporting clang version) or `zig version` (reporting zig version)? This code actually reports both.

Also just capturing the `--version` output doesn't help when custom configuration has to be done, the MINGW may have been a bad example, but for example ICC doesn't support `-march=x86-64` (Maybe Intel did not liked that name that was made-up by AMD…) and Zig doesn't support `-mtune`.

An alternative to also print the `--version` output would be to also print things like `__VERSION__` string or the `__clang_version__` one, here is the comparison of them on Saigo:

- `--version`:  
`clang version 19.0.0git (https://chromium.googlesource.com/a/native_client/nacl-llvm-project-v10.git e25355fddbdece2ef08747ead05b7f69f3bc6dca)`
- `__clang_version__`:  
`19.0.0git (https://chromium.googlesource.com/a/native_client/nacl-llvm-project-v10.git e25355fddbdece2ef08747ead05b7f69f3bc6dca)`

This can be useful because of the commit refs my code strips out, but what I output is much more readable.

This PR now brings much more than just generating some macros to report the compiler used to build the game, it also brings more compiler support thanks to the detection, and also improves things like extending PCH support to compilers with sub-commands.

Here is some example of nexe cgame reporting which compiler was used to build it:

- `cgame NaCl nacl (PNaCl 4.2.1/clang-3.6.0 pnacl-clang++) Dec  3 2024`
- `cgame NaCl nacl (Saigo 19.0.0/clang-19.0.0 x86_64-nacl-clang) Dec  3 2024`

Such report is also useful for compiled games as Saigo suffered from the model bug but PNaCl didn't.

Regarding Zig, I have a long term side-project to experiment cross-compiling native binaries with it, that's why I'm doing some efforts to improve support of it and part of this compiler detection contributes to that.

slipher:
Still don't believe it's worth adding +500 lines of code just to print a compiler version string. Not worth the increase in complexity.

I don't know/remember what the motivation for this PR is, but if you're worried about forgetting which compiler a release was built with, there's this: https://unix.stackexchange.com/a/722. Clang does it too.

illwieckz:
My first purpose is to have the information in `daemon.log`.

Let's take the model bug as an example: someone reports distorted models when he builds the game on his side, we ask for the `daemon.log` file, the compiler information can be used as a guess along other things like OpenGL driver version, etc.

The same happen if some people report an engine bug when running the game from AUR on Arch, we can ask for the `daemon.log` file, the compiler information can be used as a guess along other things like OpenGL driver version, etc.

First motivation is to report in `daemon.log` the compiler version used to build the C++ code, the same way we report the GLSL compiler version.

Then the second motivation is to use that information to improve compiler support for various reasons.

What motivated me at first was the console offset on clang:

- https://github.com/DaemonEngine/Daemon/issues/839

The model bug is another example where reporting the compiler can be useful:

The more people will build the game themselves, or the more Linux distro may package the game for their users, the more this kind of bug will be expected to be reported in the future, and from people having less knowledge than the core development team. And since we want to reach a wider audience, it means we want a situation where bugs specific to compiler are expected to be reported more from people with less debugging skills.

illwieckz:
As an example with latest ICX:

```sh
/opt/intel/oneapi/compiler/latest/bin/icpx -v
Intel(R) oneAPI DPC++/C++ Compiler 2025.0.4 (2025.0.4.20241205)
```

This doesn't report the compiler use Clang and which version of Clang it uses.

But this does:

```sh
echo '' | /opt/intel/oneapi/compiler/latest/bin/icpx -E -dM - | grep __clang_version__
#define __clang_version__ "19.0.0git (icx 2025.0.4.20241205)"
```

illwieckz:
With the extra logging, this PR also brings interesting features to CMake: for example knowing the underlying compiler gives the ability to use Clang-specific compile flags by knowing the compiler is Clang-based even if CMake itself reports something else, and other things like that become possible.

So, the only opinion against is that it adds code. But I want those features and this is the best implementation I can do. The proposed alternatives don't match the need. Actually the code si ready for more than a year and it works. After a year, the need isn't gone, and I have plan to improve the logging in other ways and to avoid merge conflicts I blocked those other changes for the same time.

So I will simply merge this once the CI finishes.

illwieckz:
@slipher I answer here because I don't want to pollute that other thread with that topic:

- https://github.com/DaemonEngine/Daemon/issues/1505

> This is a good example of why we should not merge a complex PR like https://github.com/DaemonEngine/Daemon/pull/982 without any code review.

It was reviewed by you.

For example you said [there](https://github.com/DaemonEngine/Daemon/pull/982#issuecomment-1838248692):

> I don't find it worth the hassle to have hundreds of lines of code attempting to parse the compiler's version output. Let's compare this with the architecture determining tool:
>
> - The architecture tool uses compiler macros which are a well-documented and stable interface. Version strings are free-form and might be changed without warning as they are not intended as a machine-readable interface.

So, based on that review of you, I [entirely rewrote](https://github.com/DaemonEngine/Daemon/pull/982#issuecomment-1840016694) that compiler detection to be written the way of the architecture determining tool. In fact it was already using compiler macros, but instead of parsing the output of the `cc -dM -E` (which is intended as a machine-readable interface), I went the warning-emitting-and-parsing way like we already did for the architecture detection, as you said. In fact this made the parsing more convoluted but I didn't mind.

This PR has been reviewed by you more than a year ago and I rewrote accordingly.

So, after more than a year of review by you, once I had addressed all your remarks about the implementation, once you had nothing to review anymore about the implementation, all that was remaining was your opinion ([here](https://github.com/DaemonEngine/Daemon/pull/982#issuecomment-2513623483)) on the idea of having this feature or not.

> Still don't believe it's worth adding +500 lines of code just to print a compiler version string.

Actually you may haven't grasped yet the purposes of that feature as every time you describe what you think is its purpose, you're close but not right. It's like saying “_you want bread_” instead of “_you're hungry_”, it's close but not right.

I believe we need that feature so any code providing that is worth to add. We don't share this opinion, but the review was done, all that was remaining were opinions and I was in position to chose among the opinions to follow.

I also believe it's not useful to discuss more the need of this. So hiding reverts under [pull requests titles](https://github.com/DaemonEngine/Daemon/pull/1497) named “Fix” or other misleading move is not welcome, because it would not be welcome even if it was named “Revert” explicitly to begin with.

PR #991, opened 2023-12-11
Fix the use of lightgrid on BSP surfaces without lightmap

### LINKED ISSUE #990: Some map models are not properly lit
Some map models are not properly lit:

unvanquished_2023-12-11_024128_000

That rings a bell but I don't remember what.

I was playing on Der Bunker (still 0.54.0 based if I'm right) while using 0.54.1 engine.

illwieckz:
Fix the use of lightgrid on bsp surfaces without lightmap.

It looks like a line was mistakenly deleted in 56d0d83ca64d0712a315d6637a3dfb0ccf74a5ab.

Seems to fix #990:

- https://github.com/DaemonEngine/Daemon/issues/990

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -855,7 +855,8 @@ static void Render_lightMapping( shaderStage_t *pStage )
 	GLimp_LogComment( "--- Render_lightMapping ---\n" );
 
 	bool enableLightMapping = tr.worldLightMapping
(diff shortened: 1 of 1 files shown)


slipher:
LGTM

illwieckz:
The perfect map to test this is `rusty`, the default intermission displays a scene with such map model not using lightmaps:

With bug:

unvanquished_2023-12-11_075425_000

Without bug:

unvanquished_2023-12-11_075626_000

PR #992, opened 2023-12-12
Introduce r_lightMode, implement vertex lighting and use it when needed.

illwieckz:
Historically we had the `r_vertexLighting` cvar, but we didn't have vertexLighting, in fact we used it as reverse cvar for enabling light mapping:

-  `r_vertexLighting 0`: enable light mapping (default)
-  `r_vertexLighting 1`: disable light mapping, fallback on grid lighting

This replaces it with `r_lightMode`, with possible values:

- `r_lightMode 0`: fullbright
- `r_lightMode 1`: vertex lighting
- `r_lightMode 2`: grid lighting
- `r_lightMode 3`: light mapping (default)

The only `r_lightMode` values expected to be set by players are `1` (vertex lighting) and `3` (light mapping), especially since others can be a ways to cheat. To use other light modes one should set `r_cheatLightModes` to `on`.

Vertex lighting does not do less computation than light mapping, but skips the uploading of lightmaps and then save on GPU memory size, it is the only reason why it's not a bad idea to provide it to users.

Light grid is not faster than light mapping, there are more GLSL code to execute, it may uses less GPU memory though, but on 20 years old hardware I got less fps last time I compared. Grid lighting is cheaty because the light grid skips the projected shadow.

Fullbright is cheaty.

The cheat light modes are meant for debugging purposes, to give the developer/mapper the ability to see the map with those various lighting techniques.

Currently the engine uses light mapping on world surfaces and grid lighting on game models (players, buildables, weapons…), that's why we have both light mapping and grid lighting implemented. We also used grid lighting on world surfaces when there is no lightmap for the surface. #990 was an example of bug occurring when there is no fallback for world surfaces without lightmap.

But it is expected world surfaces without lightmap don't use grid lighting but vertex lighting. Why? Because the q3map2 map compiler has an option to skip some lightmaps when the color is very similar to vertex colors. So basically, with that option, if the light is consistent, the surface doesn't get a lightmap, and if the light is not consistent (for example it has a projected shadow), it gets a lightmap. A game like Wolf:ET made strong usage of that technique and some maps had like only half of surfaces having a lightmap. On maps using such optimization, the light guessed from the lightgrid is close to the lightmap one but not equal, it makes the map look checkered, since surfaces lit with lightgrid don't match the surfaces lit with lightmap. We should then use vertex lighting on world surfaces without lightmaps.

Since I needed to implement vertex lighting for those surfaces, it was easy to add this light mode to `r_lightMode` options.


illwieckz:
I may turn the cvar into a modern one and use a range too, and the final commits may differ, that's why I currently marked the PR as draft, but it should work.

illwieckz:
Some things:

I noticed that with `fullbright` the game models still use grid lighting. I'll look at it later. I consider this a bug, though totally minor since it only affects a debug mode.

I want to use grid lighting for world movers, if not enforced, at least optional for the mapper. It would prevent the movers to move with their baked shadow, but get the light color and intensity from their position. That may be another PR anyway.

If we want to keep grid lighting and vertex ligthing not a cheat, we may do a `r_fullbright` cvar instead so `r_lighting` can chose between vertex lighting, grid lighting, light mapping.

illwieckz:
I finally did a very deep revamp, to make the code more obvious, and relying less on hacks.

Especially, I had to add a macro to avoid guessing the information out of other macros in GLSL, this means I had to increase by 1 the amount of macros a single shader can received, but since I also did other rewrites to make sure some macro can exclude themselves, the amount of compiled shaders are exactly the same in the end.

All the uses case I wanted to take care about are now working.

I still see one bug with a Wolf:ET map but it is not introduced by this code and may even require to dive deeper in Wolf:ET specificity so it's out of topic twice.

illwieckz:
It is believed vertex lighting should not be faster than light mapping because the implementation is simply to blend vertex light color over a white light map, meaning the light map code runs in all cases. All it would do would be to save GPU memory by not uploading light maps. To get a faster implementation one would need to write more GLSL code and more C++ plumbing, which is likely not worth the effort. In all cases the base line is light map and that's not bad.

illwieckz:
While doing that work I stumbled on many other bugs (I fixed) like:

- a variable for the rim lighting feature meant to be used for models was only set for surfaces of the map.
- while we have both enabled half lambert and wrap around lighting for models, only one of them was used at once.
- lightgrid used to guess light direction to emulate the deluxemap for models was always set when a deluxemap was missing, even if there was no computation to do
- etc.

illwieckz:
In fact with latest iteration it even fixes the rendering of Wolf:ET maps (something I was aware for 8 years!). I'm now unaware of any map lighting bug remaining.

illwieckz:
So, here are some before/after comparisons, with those maps:

game|map|viewpos
--|--|--
tremulous|rusty|-1672 -936 88 -55 3
quake3|school|205 1140 5 36 8
unvanquished|grangermaze|-1705 2600 216 135 51
wolfet|oasis|2219 2861 -31 93 13
wolfet|goldrush|2050 -533 -115 -116 10
smokinguns|elpaso|566 725 99 -125 7
tremulous|karith|2080 4484 -416 120 0
unvanquished|chasm|-2627 1622 272 -38 0
unvanquished|hangar28|-1286 740 -765 -28 23
unvanquished|plat23|2240 2100 230 42 16
unvanquished|forlorn|2346 1124 442 94 30

### rusty

`0.54.1`, missing light fallback on models without lightmap (regression):

unvanquished_2023-12-14_214931_000

`master`, grid light used as fallback (revert of the regression):

unvanquished_2023-12-14_220745_000

this branch, vertex lighting used as fallback (what should be done):

unvanquished_2023-12-14_221427_000

### school

`0.54.1`, missing light fallback on models without lightmap (regression):

unvanquished_2023-12-14_220018_000

`master`, grid light used as fallback (revert of the regression), but with a default lightgrid because this map doesn't have one:
unvanquished_2023-12-14_220818_000

this branch, vertex lighting used as fallback (what should be done):

unvanquished_2023-12-14_221451_000

### grangermaze

`master`, regression:

unvanquished_2023-12-14_220850_000

this branch, what should be done:

unvanquished_2023-12-14_221511_000

illwieckz:
### oasis

`master`, grid light used as fallback (what the engine did before):

unvanquished_2023-12-14_220911_000

this branch, vertex lighting used as fallback (what should be done):

unvanquished_2023-12-14_221546_000

### goldrush

`master`, grid light used as fallback (what the engine did before):

unvanquished_2023-12-14_220933_000

this branch, vertex lighting used as fallback (what should be done):

unvanquished_2023-12-14_221607_000

### elpaso

`master`, as expected:

unvanquished_2023-12-14_222835_000

this branch, no regression:

unvanquished_2023-12-14_222910_000

## karith

`master`, as expected:

unvanquished_2023-12-14_220948_000

this branch, no regression:

unvanquished_2023-12-14_221629_000

illwieckz:
### chasm

`master`, as expected:

unvanquished_2023-12-14_221013_000

this branch, no regression:

unvanquished_2023-12-14_221645_000

### hangar28

`master`, as expected:

unvanquished_2023-12-14_221110_000

this branch, no regression, this scene is known to break when moving ou of the room with `noclip` if we change the surface properties instead of using temporary variables to set vertex lighting fallback:

unvanquished_2023-12-14_221719_000

### plat23

`master`:
unvanquished_2023-12-14_221154_000

this branch, models lighting may have subtle differences since features that were wrongly disabled are now enabled:

unvanquished_2023-12-14_221809_000

### forlorn

`master`:

unvanquished_2023-12-14_221216_000

this branch, models lighting may have subtle differences since features that were wrongly disabled are now enabled:

unvanquished_2023-12-14_221929_000

Actually, the armoury screen is interesting on this one.

illwieckz:
This is now ready for review.

I played all my recent games with that branch, and the only things I spotted were fixes, even unexpected ones.

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -76,15 +84,18 @@ void computeLight( vec3 lightDir, vec3 normal, vec3 viewDir, vec3 lightColor,
...
   float NdotL = dot( normal, lightDir );
 
+#if !defined(USE_BSP_SURFACE)
 #if defined(r_halfLambertLighting)
   // http://developer.valvesoftware.com/wiki/Half_Lambert
   NdotL = NdotL * 0.5 + 0.5;
   NdotL *= NdotL;
-#elif defined(r_wrapAroundLighting)
+#endif
+#if defined(r_wrapAroundLighting)
```
slipher, line 93: I think it's intentional that these are mutually exclusive. It doesn't make sense to use the two formulas manipulating normals together.

Maybe wrapAroundLighting should just be nuked? The cvar is marked cheat so it can't be used anyway. Also the code here seems wrong (using r_wrapAroundLighting as a number - probably it should be u_wrapAroundLighting)

src/engine/renderer/tr_init.cpp
```diff
@@ -1107,7 +1107,14 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
...
 		r_precomputedLighting = Cvar_Get( "r_precomputedLighting", "1", CVAR_CHEAT | CVAR_LATCH );
-		r_vertexLighting = Cvar_Get( "r_vertexLighting", "0", CVAR_LATCH | CVAR_ARCHIVE );
+		/* Default lighting mode, the engine fallbacks on lower ones if missing for the
+		whole map or for the given surface.
+		This is a cheat cvar as changing the lighting mode may give an unfair advantage.
+		0: fullbright
+		1: vertex lighting
+		2: grid lighting
+		3: light mapping */
+		r_lighting = Cvar_Get( "r_lighting", "3", CVAR_LATCH | CVAR_CHEAT );
```
slipher, line 1117: do the Cvar::Range

src/engine/renderer/gl_shader.cpp
```diff
@@ -1369,6 +1373,62 @@ bool GLCompileMacro_USE_DEPTH_FADE::HasConflictingMacros(size_t permutation, con
 	return false;
 }
 
+bool GLCompileMacro_USE_DELUXE_MAPPING::HasConflictingMacros(size_t permutation, const std::vector<GLCompileMacro*> &macros) const
+{
+	for (const GLCompileMacro* macro : macros)
+	{
+		if ((permutation & macro->GetBit()) != 0 && (macro->GetType() == USE_GRID_DELUXE_MAPPING || macro->GetType() == USE_GRID_LIGHTING))
+		{
+			//Log::Notice("conflicting macro! canceling '%s' vs. '%s'", GetName(), macro->GetName());
```
slipher, line 1382: delete commented code

slipher (commented):
Will review more later.

Can you explain the concept of vertex lighting more? I don't understand where the lighting information comes from in that case.

illwieckz:
> Can you explain the concept of vertex lighting more? I don't understand where the lighting information comes from in that case.

In the BSP file, in the struct storing vertexes, there is a RGBA color field:

> ### Vertexes
>
> The vertexes lump stores lists of vertices used to describe faces. There are a total of length / sizeof(vertex) records in the lump, where length is the size of the lump itself, as specified in the lump directory.
>
> vertex|&nbsp;
> ---|---
> float[3] position|Vertex position.
> float[2][2] texcoord|Vertex texture coordinates. 0=surface, 1=lightmap.
> float[3] normal|Vertex normal.
> ubyte[4] color|Vertex color. RGBA. 

Cf. http://www.mralligator.com/q3/#Vertexes

Those vertex colors are computed by q3map2 at the same time the light grid and the light maps are computed. It means

The code modified in that PR doesn't process those colors, they are assumed to be already applied by the engine in a way the lightmapping GLSL code blends over it.

The reason why I assume that is that we already have other features that require that underlying mechanism to be already working, for example terrain blending. This is the feature we can see in outdoor of Thunder map where rock texture blends with sand texture, or in garden in Station15 map, were grass blends with mud. This terrain blending uses vertex colors as well, more precisely the alpha channel of it, the engine is instructed to use an operation that would use such alpha channel to blend a lightmap.

In case of vertex lighting, what I just do is then to set some color field to black and then zero, and tell the engine is has to use an operation to blend over colors. I assume that color field set in lightmapping C++ code do not carry the vertex color and is only used with some operations with vertex colors (here, probably an addition), hence the need for such color filed to be zeroed to get the expected result. The alpha channel appears to also need to be zeroed (transparent) when the vertex lighting has to be applied on a transucent texture.

Those values (what operation to set, what color to set for the operation: black, transparent…) were guessed with intuition and some kind of trial & error process as I miss knowledge and understanding of some parts of the engine, but I knew vertex colors should already be applied by the engine and that all I had to do was to find the proper blend operation.

Vertex color blending was already implemented (otherwise many features would be broken), what I had to do was to tell the engine to blend textures over such vertext color when vertex lighting is enabled, which is like a 5 line patch.

The vertex colors are very likely interpolated somewhere in the engine, this is needed for the terrain blending, and we can see on school map screenshot above some emulation of projected shadow on some models using vertex colors, this can be done by interpolating colors of all we have to emulate a shadow is to darken the surface by interpolate light to dark colors between one point to another (and then, between edges).

Everything else in that PR is some very detailed decision code to be make all various lighting mode usables in a single scene, plus other rewrites that makes writing that decision code easier.

src/engine/renderer/tr_shade.cpp
```diff
@@ -1075,7 +1194,8 @@ static void Render_lightMapping( shaderStage_t *pStage )
 
 	gl_lightMappingShader->SetUniform_Color( color );
 
-	if ( tess.bspSurface && !enableLightMapping )
+	// It looks like WrapAroundLighting is meant for models.
+	if ( !tess.bspSurface && enableGridLighting )
```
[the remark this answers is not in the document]
illwieckz: `!tess.bspSurface` is true when the surface comes from a model (not a bsp surface), if wrapAroundLighting is for models, then testing for `tess.bspSurface` was a bug.

I see no reason why world surfaces would be rendered with tricks.

I find no documentation about what is “wrap around lighting” for a surface renderer. Searching on the internet only gives me some kind of lamp and photography techniques involving lamps. So, in that context, “wrap around lighting” is about lamps, not about surface, and a shader processing surface pixels can do nothing about it.

The only thing I barely found last week was something about emulating the way light navigates within some materials (like skin) , and that is something a shader can help to implement, but if that is what we have, it's for models.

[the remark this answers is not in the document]
illwieckz: Something good to know is that this code was likely written by me in a previous refactoring, same for the GLSL `#ifelse`.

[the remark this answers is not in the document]
slipher: Well similarly to half-Lambert lighting, "wraparound lighting" seems to be a hack for making stuff less dark. With deluxe maps the (really stupid?) assumption is made that all light comes from a single direction. In reality there are often various sources, plus even if the light comes from a point source it can diffract and bounce. Therefore the hacks are needed to avoid stuff not facing towards the light direction being overly dark. "Wraparound lighting" artificially boosts the dot product of the normal and light direction, as if the light turns slightly to hit the surface, hence the "wrap around" name. Like a crude model of diffraction I guess.

Half-Lambert lighting does a different calculation to mess with the dot product of normal and light direction, so the two techniques should not be used at once.

It's true that these techniques make more sense for models than BSP surfaces. I assume that when there are multiple light sources, the map compiler would appropriately adjust the lightmaps and deluxemaps to get the desired final amount of illumination.

I suppose it's likely that wraparound lighting was some one-off Xreal test or something, so no game models using it exist.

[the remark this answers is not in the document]
illwieckz: > It's true that these techniques make more sense for models than BSP surfaces. I assume that when there are multiple light sources, the map compiler would appropriately adjust the lightmaps and deluxemaps to get the desired final amount of illumination.

Especially, q3map2 knows how to bounce the light. This misses other mechanism that exists in real life, but at least in a room, something not facing towards the light direction gets lit by the light bouncing on other surfaces around (something that may be implementable in q3map2 to make it more realistic would be to read the specular map to bounce less when the surface is less shiny, it already bounces according to the color though).

src/engine/renderer/gl_shader.h
```diff
@@ -30,7 +30,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #define USE_UNIFORM_FIREWALL 1
 
 // *INDENT-OFF*
-static const unsigned int MAX_SHADER_MACROS = 9;
+static const unsigned int MAX_SHADER_MACROS = 10;
 static const unsigned int GL_SHADER_VERSION = 4;
```
slipher, line 34: It seems you need to increase GL_SHADER_VERSION as well (which means the version of the shader cache files), because changing MAX_SHADER_MACROS changes the header struct's layout. When I run 0.54.1 after running this branch, my Nvidia driver crashes.

illwieckz: Ah! Good to know! Maybe we should add `MAX_SHADER_MACROS` to the checksum?

src/engine/renderer/tr_shade.cpp
```diff
@@ -903,46 +882,184 @@ static void Render_lightMapping( shaderStage_t *pStage )
...
+		{
+			/* TODO: There is no reason this test fails:
+			( pStage->implicitLightmap || pStage->type == stageType_t::ST_LIGHTMAP )
+			Anyway, this test is useless. */
+			bool hasMap = tr.worldLightMapping
+				&& ( tess.lightmapNum >= 0 && tess.lightmapNum <= tr.lightmaps.currentElements );
+
+			if ( !hasMap )
+			{
+				/* Use vertex light instead of grid light as a fallback on world
```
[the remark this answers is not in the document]
illwieckz: Well, both are true, but the one I wrote lacks context. There are no lightmap, but two remaining source of lights, vertex light or grid light. As a fallback to the lack of lightmap, use vertex light and not grid light. Using vertex light instead of grid light in such case is one of the bugfix this PR brings.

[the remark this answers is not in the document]
illwieckz: I guess I can reword that comment.

src/engine/renderer/gl_shader.h
slipher: Not the checksum exactly, but yes it would be good to add it to the header struct.

illwieckz: Yes, “_ in a way it contributes to the data used to compute the checksum_” or something like that.

slipher: The GLSL source code is checksummed. Other data are stored directly into the header and simply compared for equality.

illwieckz: I see.

src/engine/renderer/tr_shade.cpp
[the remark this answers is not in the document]
illwieckz: I just thought about something, maybe rim lighting can be useful on world surfaces when using dynamic lights?

src/engine/renderer/tr_init.cpp
illwieckz: We need this to make it lateched:

- https://github.com/DaemonEngine/Daemon/pull/999

src/engine/renderer/glsl_source/computeLight_fp.glsl
illwieckz: > The cvar is marked cheat so it can't be used anyway. 

The cvar is marked cheat so it can't be _disabled_.

illwieckz: This is how it is set:

```c++
r_wrapAroundLighting = Cvar_Get( "r_wrapAroundLighting", "0.7", CVAR_CHEAT | CVAR_LATCH );
```

illwieckz: In `tr_shader.cpp`, the `wrapAroundLighting` keyword is expected to be set on material stages.

illwieckz: RBQUAKE3 (latest version of XreaL) use a variable indead (`u_LightWrapAround`):

https://github.com/RobertBeckebans/RBQUAKE-3/blob/9559591679085e5462de80049990a6a302616b94/base/glsl/forwardLighting_fp.glsl#L1037-L1041

illwieckz: We do that, though:

```
	if ( r_wrapAroundLighting->value )
		AddConst( str, "r_wrapAroundLighting", r_wrapAroundLighting->value );
```

RBQUAKE3 does this like us:

```
r_wrapAroundLighting  = ri.Cvar_Get( "r_wrapAroundLighting", "0.7", CVAR_CHEAT | CVAR_SHADER );
```

I assume the value is set in a similar way thanks to the `CVAR_SHADER` flag, but then no code makes use of this value in RBQUAKE3.

illwieckz: I guess the cvar being a float is a relic of a time it was being prototyped, the value is expected to be set by materials.

illwieckz: Our legacy `forwardLighting_fp.glsl` also uses `u_LightWrapAround`:

```c++
float NL = clamp(dot(normal, lightDir) + u_LightWrapAround, 0.0, 1.0) / clamp(1.0 + u_LightWrapAround, 0.0, 1.0);
```

illwieckz: So, I modified the code to use the variable, so the feature is only used on materials setting it, but I disabled the cvar by default. I removed the cheat flag as I assume it was part of a prototype, like the float value (that is now expected to be `0`/`1` to disable/enable).

src/engine/renderer/tr_shade.cpp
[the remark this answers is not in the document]
illwieckz: Reworded.

src/engine/renderer/tr_shade.cpp
```diff
@@ -903,46 +882,184 @@ static void Render_lightMapping( shaderStage_t *pStage )
...
+
+	uint32_t stateBits = pStage->stateBits;
+
+	if ( light == Light::Map )
 	{
-		lightmap = GetLightMap();
+		if ( tess.bspSurface )
+		{
+			/* TODO: There is no reason this test fails:
+			( pStage->implicitLightmap || pStage->type == stageType_t::ST_LIGHTMAP )
```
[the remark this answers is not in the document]
illwieckz: From my mind!

I first thought, “we may need this” but then it broke so I was like “maybe it's not needed”, this comment reflected this state.  😃️

But I was right at first, we need this but something else was missing to make this test work: the materials autogenerated from textures without `.shader` files needed `pStage->implicitLightmap` to be set to `true`.

Now the code does this test, this is to properly avoid to apply lightmap on collapsed stages without lightmaps, like this:

```
{
  map textures/texture_d
  heightMap textures/texture_h
}
```

This is one of the very rare use cases that are possible.

illwieckz:
I added a fix for not loading light style map when light mapping is disabled.

I moved some code to compute what light mode to use in a way it is precomputed once at map load instead of doing it on every surface on every frame. What is now done on every surface on every frame is the computation specific to the surface (like a material having `nolightmap` keyword), no more.

illwieckz:
This now uses `Cvar::Range` with `Cvar::Latch()`.

illwieckz:
I added a commit adding the `r_mapOverBrightBits` cvar.

To quote a lenghty comment I wrote in the code (first paragraph is original):

```
	/* These are the values expected by the rest of the renderer (esp. tr_bsp),
	used for "gamma correction of the map".
	Both were set to 0 if we had neither COMPAT_ET nor COMPAT_Q3,
	it may be interesting to remember.

	Quake 3 and Tremulous values:

	  tr.overbrightBits = 0; // Software implementation.
	  tr.mapOverBrightBits = 2; // Quake 3 default.
	  tr.identityLight = 1.0f / ( 1 << tr.overbrightBits );

	Games like Quake 3 and Tremulous require tr.mapOverBrightBits
	to be set to 2. Because this engine is primarily maintained for
	Unvanquished and needs to keep compatibility with legacy Tremulous
	maps, this value is set to 2.

	Games like True Combat: Elite (Wolf:ET mod) or Urban Terror 4
	(Quake 3 mod) require tr.mapOverBrightBits to be set to 0.

	For some reasons the Wolf:ET engine sets this to 0 by default
	but the game sets it to 2 (and requires it to be 2 for maps
	looking as expected).

	The mapOverBrightBits value will be read as map entity key
	by R_LoadEntities() in tr_bsp, making possible to override
	the default value and properly render a map with another
	value than the default one.

	If this key is missing in map entity lump, there is no way
	to know the required value for mapOverBrightBits when loading
	a BSP, one may rely on arena files to do some guessing when
	loading foreign maps and games ported to the Dæmon engine may
	require to set a different default than what Unvanquished
	requires.

	Using a non-zero value for tr.mapOverBrightBits turns light
	non-linear and makes deluxe mapping buggy though.

	Mappers may port and fix maps by multiplying the lights by 2.5
	and set the mapOverBrightBits key to 0 in map entities lump.

	It will be possible to assume tr.mapOverBrightBits is 0 when
	loading maps compiled with sRGB lightmaps as there is no
	legacy map using sRGB lightmap yet, and then we will be
	able to avoid the need to explicitly set mapOverBrightBits
	to 0 in map entities. It will be required to assume that
	tr.mapOverBrightBits is 0 when loading maps compiled with
	sRGB lightmaps because otherwise the color shift computation
	will break the light computation, not only the deluxe one.

	In legacy engines, tr.overbrightBits was non-zero when
	hardware overbright bits were enabled, zero when disabled.
	This engine do not implement hardware overbright bit, so
	this is always zero, and we can remove it and simplify all
	the computations making use of it.

	Because tr.overbrightBits is always 0, tr.identityLight is
	always 1.0f. We can entirely remove it. */
```

Since `tr.overbrightBits` is always `0`, tr.identityLight is always `1.0f`, so I also removed all their occurrences to simplify the code.

Here are some before/after screenshot on some Urban Terror 4 and True Combat: Elit maps when setting `r_mapOverBrightBit` to `0` as those games expect:

game|map|viewpos
-|-|-
urt4|abbey|1945 -193 379 131 13
urt4|mandolin|553 3956 202 136 10
tce|railhouse|1329 2582 49 138 -1
tce|snow|1242 93 222 -45 14

UrT4 Abbey before/after:

unvanquished_2024-01-01_110350_000

unvanquished_2024-01-01_110409_000

UrT4 Mandoli, before/after:

unvanquished_2024-01-01_110544_000

unvanquished_2024-01-01_110559_000

TCE railhouse before/after:

unvanquished_2024-01-01_111244_000
unvanquished_2024-01-01_111258_000

TCE snow before/after:

unvanquished_2024-01-01_111948_000
unvanquished_2024-01-01_112006_000

src/engine/renderer/glsl_source/computeLight_fp.glsl
slipher: OK but it should still be mutually exclusive with half lambert

src/engine/renderer/tr_local.h
```diff
@@ -292,6 +292,9 @@ static inline void glFboSetExt()
 	GL_fboShim.glRenderbufferStorageMultisample = glRenderbufferStorageMultisampleEXT; */
 }
 
+enum class Light { FullBright, Vertex, Grid, Map };
```
slipher, line 295: The naming on these is pretty far from our conventions. I guess I don't insist on '_t' but something like lighting style would be more descriptive than just "light". The values (e.g. FullBright, Vertex) should be all caps.

src/engine/renderer/glsl_source/computeLight_fp.glsl
illwieckz: I still don't get why this couldn't be cumulative.

Also doing either one or the other is not decidable: we have no way to know in this code if a material uses wrapAround or not.

src/engine/renderer/tr_local.h
illwieckz: Light Style refers to something else in engine, light styles are extra lightmaps dynamically blended in `.shader`.

Light Mode would be better.

illwieckz: I renamed them `lightMode_t` and `deluxeMode_t` and set values all caps.

src/engine/renderer/glsl_source/computeLight_fp.glsl
slipher: Well if you really wanted to do both at once, the only way that would make sense is to do "wrap around" first.



> Also doing either one or the other is not decidable: we have no way to know in this code if a material uses wrapAround or not.

Well the old way was just do disable wraparound if halflambert is enabled.

illwieckz: > The old way was just to disable wraparound if halflambert is enabled.

That is not per-material.

Things like halfLambert and wrapAround are to meant to be used on some surfaces of the game in the same scene.

slipher: I might suggest once again nuking wraparound as it causes us problems for no benefit.

I don't see any way to set half-lambert per material. It seems to be a global cvar only. So that's how it is possible to disable wraparound whenever half-lambert is on.

By the way if half-lambert is disabled for BSP surfaces, then wraparound probably should be too for the same reasons.

illwieckz: I sent an e-mail to Tr3b to know what wrapAround was meant to be in XreaL to begin with.

I noticed something is wrong in that code anyway, as I believe `clamp(1.0 + u_LightWrapAround, 0.0, 1.0)` is always `1`.

illwieckz: In [that page](https://www.jordanstevenstechart.com/lighting-models) the writer also talk about a variant of Half-Lambert that is named `Diffuse Wrap`, which allows to use an arbitrary value instead of 0.5:

```glsl
pow(NdotL * wrapValue + (1-wrapValue),2) 
```

We may allow to customize the `wrapValue` in materials like wrapAround does.

slipher: > I noticed something is wrong in that code anyway, as I believe `clamp(1.0 + u_LightWrapAround, 0.0, 1.0)` is always `1`.

Not if it's negative.

src/engine/renderer/tr_light.cpp
```diff
@@ -318,27 +318,37 @@ void R_SetupEntityLighting( const trRefdef_t *refdef, trRefEntity_t *ent, vec3_t
...
 
 		VectorSet( ent->lightDir, -1, 1, 1.25 );
 		VectorNormalize( ent->lightDir );
 	}
 
 	if ( ( ent->e.renderfx & RF_MINLIGHT ) ) // && VectorLength(ent->ambientLight) <= 0)
 	{
+		/* Historically those values were multiplied by
+		tr.identityLight but tr.identityLight is always 1.0f
+		in Dæmon engine as the as the overbright bit
```
illwieckz, line 345: Note to myself: fix the comment.

illwieckz: Fixed.

src/engine/renderer/glsl_source/computeLight_fp.glsl
illwieckz: wrapAroundLighting has nuked:

- https://github.com/DaemonEngine/Daemon/pull/1015

src/engine/renderer/tr_shade.cpp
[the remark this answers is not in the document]
illwieckz: The code was NUKed:

- https://github.com/DaemonEngine/Daemon/pull/1015

illwieckz:
That looks ready, except I haven't fully decided what to do about `r_lighting` being cheat or not.

@cu-kai in a recent game I read you saying “this is why r_lighting should not be cheat”, can you elaborate?

I believe there is a way to make `r_lighting` not a cheat but preventing `fullbright` to be selected by adding a `r_allowFullbright` cheat cvar that would be only enable-able in devmap, and the engine would reset to lightmap if r_lighting is set to fullbright but that cvar is not enabled. But do we really need that?

Actually the fact that grid lighting produces a more bright render of the map should be considered a bug, not a feature. There is no reason the same light source produces different lighting intensity given the way it is stored.

illwieckz:
Actually the overBrightBits hack used to render quake3/tremulous maps (and then unvanquished) map in a “brighter way” by shifting the light values not only break the linearity of the light computation and not only break the deluxe computation, it emphasises _a lot_ the fact that our computation lighting surfaces from lightgrid has a tendency to gives an higher light intensity to surface.

In our engine mapOverBrightBits is enabled by default for compatibility with legacy assets and we had no option to disable it before this PR.

So, actually, using grid lighting to render a map while mapOverBrightBits is enabled is literally a cheat: it's nothing more than a combinations of various kludges to force the engine to use a broken light computation.

illwieckz:
In the future we will do correct lighting computations by disabling the overBrightBits hack and lighting our maps correctly in source instead of relying on a hack at render time to compensate. This is what games like Urban Terror 4 or True Combat: Elite did maybe more than a decade ago.

The day we do non-broken lighting computation, and we can already do that, this day the hack of using lightgrid for world surfaces to get a brighter map will not work or not in a significant way.

illwieckz:
I can't also not exclude there would be actual hacks in lightgrid code (either in q3map2 map compiler, either in engine), to artificially increase light from lightgrid to workaround the fact models get only one source of light without bouncing and other mechanisms.

illwieckz:
If we want to give users alternate ways of lighting the map, we can give

- light mapping
- vertex lighting

Like Quake 3 and Tremulous did.

Vertex lighting should be no faster than light mapping: we still apply a stub lightmap and we can't avoid that without making the code much more complex, so the computation is as heavy, but it will save GPU memory by not uploading lightmaps.

Grid lighting uploads lightgrid instead of lightmap, this may use less GPU memory, but the lightgrid computation is much more heavy than the lightmap/vertex lighting one. I see no reason to provide a grid option lighting to users for performance purpose (even not counting the fact the compute is broken with overBrightBits).

illwieckz:
I did some tests with some maps, actually the benefit of lightgrid in `overthehill` is not that big.

It looks like the big difference affects newer maps like our official ones, and some others like `rusty`, maybe there is a q3map2 bug…

**overthehill**

viewpos `-1113 298 159 57 5`; lightmap, lightgrid, vertex light:

unvanquished_2024-01-08_114511_000

unvanquished_2024-01-08_114524_000

unvanquished_2024-01-08_114541_000

viewpos `-470 631 56 105 -0`; lightmap, lightgrid, vertex light:

unvanquished_2024-01-08_114627_000

unvanquished_2024-01-08_114638_000

unvanquished_2024-01-08_114650_000

viewpos `521 513 271 -136 15`; lightmap, lightgrid, vertex light:

unvanquished_2024-01-08_114735_000

unvanquished_2024-01-08_114755_000

unvanquished_2024-01-08_114808_000

**rusty**

default view; lightmap, lightgrid, vertex light:

unvanquished_2024-01-08_121147_000

unvanquished_2024-01-08_121159_000

unvanquished_2024-01-08_121212_000

**chasm**

default view; lightmap with deluxemap, lightgrid, vertex light:

unvanquished_2024-01-08_120712_000

unvanquished_2024-01-08_120723_000

unvanquished_2024-01-08_120737_000

**plat23**

default view; lightmap with deluxemap, lightmap, lightgrid, vertex light:

unvanquished_2024-01-08_120848_000

unvanquished_2024-01-08_120904_000

unvanquished_2024-01-08_120919_000

unvanquished_2024-01-08_120940_000

illwieckz:
Enabling deluxemapping makes the map a bit brighter too, but not as strong as using light grid on recent maps.

Actually people can already get brighter maps with `r_gamma` so forbidding light grid is probably pointless anyway.

illwieckz:
The big problem of lightgrid is that it completely ignores projected shadows:

**arachnid2**

viewpos `-697 -1144 586 -74 18`; lightmap with deluxemap, lightmap, lightgrid, vertex light:

unvanquished_2024-01-08_122325_000

unvanquished_2024-01-08_122338_000

unvanquished_2024-01-08_122400_000

unvanquished_2024-01-08_122419_000

viewpos `-548 -1644 337 -74 -2`; lightmap with deluxemap, lightmap, lightgrid, vertex light:

unvanquished_2024-01-08_122508_000

unvanquished_2024-01-08_122523_000

unvanquished_2024-01-08_122544_000

unvanquished_2024-01-08_122558_000

illwieckz:
I added a commit adding an `r_cheatLighting` cvar, once enabled all light modes can be used (fullbright, vertex light, grid light, light map), but when disabled (by default), only vertex light and light map can be used, not selecting light map will fallback on vertex lighting.

illwieckz:
This looks ready to me.

illwieckz:
I know one map that glitches with this branch, it's nova. But nova was also glitching with Tremulous OpenGL 1 renderer, so actually glitching is being compatible and rendering things as expected. I don't know yet why nova glitches, but this may be a bug in the map data.

In 0.54.1 the glitch was not visible _because of a bug_ disabling the features that glitches.

illwieckz:
We may actually name the cvars `r_lightMode` and `r_cheatLightModes`.

illwieckz:
> We may actually name the cvars `r_lightMode` and `r_cheatLightModes`.

This sounds far better, I did that.

slipher:
I don't think this 'cheat lighting' cvar helps anything. Better to do like
```
if ( !Com_AreCheatsAllowed() && r_lightmode.Get() == FULLBRIGHT )
{
    r_lightmode.SetValue(MAP);
    Cvar::Latch(lightmode);
}
```

illwieckz:
OK I used `Com_AreCheatsAllowed`, but I don't think changing the cvar value is a good idea.

I also turned the black list into a white list.

slipher:
> OK I used `Com_AreCheatsAllowed`, but I don't think changing the cvar value is a good idea.

That's how every other cheat cvar works. E.g. if you set cg_thirdperson 1 then connect to a server, it gets changed to 0.

illwieckz:
> That's how every other cheat cvar works. E.g. if you set cg_thirdperson 1 then connect to a server, it gets changed to 0.

I would prefer to return to previous value but I guess it's not decidable.

illwieckz:
I did as you said.

cu-kai:
> @cu-kai in a recent game on “over the hill” map I read you saying “this is why r_lighting should not be cheat”, can you elaborate?

I think this was almost a joke in reference to dretches blending in with the texture used in the hall on that map.

illwieckz:
> > @cu-kai in a recent game on “over the hill” map I read you saying “this is why r_lighting should not be cheat”, can you elaborate?
> 
> I think this was almost a joke in reference to dretches blending in with the texture used in the hall on that map.

OK, thanks for the answer. 👍️

illwieckz:
```
10 23:20 <illwieckz> do someone has anything against https://github.com/DaemonEngine/Daemon/pull/992 ?
10 23:30 <cu-kai> i don't really have anything to say on this, i am not qualified to comment
10 23:36 <illwieckz> no one is qualified to comment :-/
10 23:37 <illwieckz> well, perturbed-mob has some interesting remarks about the cvar interface, and I believe their are now all addressed
10 23:39 <perturbed-mob> Lgtm 
10 23:39 <illwieckz> THX
```

illwieckz:
Today is a great day! 🎉️

PR #993, opened 2023-12-13
tr_image: _cinematic image is not a normal map

illwieckz:
Something I spotted while reading the code.

On line `2884` we have

```
imageParams.bits = IF_NOPICMIP | IF_NORMALMAP;
```

So we must re-set this variable to `IF_NOPICMIP` when creating the cinematic images.

Once such variables are set, there is no need to set them again for following images with same options, so I just moved existing lines at an early point.


slipher:
LGTM

PR #994, opened 2023-12-14
Localization: don't try to detect country

slipher:
Will fix https://github.com/Unvanquished/Unvanquished/issues/2867 when combined with deleting the country suffixes from the selector.

### DIFF
--- a/src/engine/qcommon/translation.cpp
+++ b/src/engine/qcommon/translation.cpp
@@ -41,6 +41,14 @@ extern "C" {
 
 static Cvar::Cvar<std::string> language("language", "language for UI text", Cvar::NONE, "");
 
+// Auto-detect the language, but not any further locale information.
+// If other information such as country is desired in the future, note that FindLocale isn't very
+// accurate with the country or "variant" components. A GNU-compliant locale name [1] could have
+// up to 4 components, e.g. "sr_RS@latin.UTF-8", and all of them except the language are optional.
+// FindLocale naively assumes that the order is always (language, country, "variant") and
+// that any punctuation mark is a separator. So "C.UTF-8" would be language C, country UTF and
+// variant 8.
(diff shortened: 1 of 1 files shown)


src/engine/qcommon/translation.cpp
```diff
@@ -41,6 +41,14 @@ extern "C" {
...
 
+// Auto-detect the language, but not any further locale information.
+// If other information such as country is desired in the future, note that FindLocale isn't very
+// accurate with the country or "variant" components. A GNU-compliant locale name [1] could have
+// up to 4 components, e.g. "sr_RS@latin.UTF-8", and all of them except the language are optional.
+// FindLocale naively assumes that the order is always (language, country, "variant") and
+// that any punctuation mark is a separator. So "C.UTF-8" would be language C, country UTF and
+// variant 8.
+// [1] https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html
 void Trans_LoadDefaultLanguage()
```
DolceTriade, line 52: Wouldn't the right thing to do a fall though approach where if lang_COUNTRY doesn't exist, then fallback to lang then fallback to english?

slipher: I guess if you wanted to have country information, the best thing would be to put it in another cvar instead of making the gamelogic parse a multi-part string.

illwieckz: I guess two cvars would make harder to produce a single language list?

slipher: With a single language+country cvar it is already difficult to have a single language list that works correctly (see https://github.com/Unvanquished/Unvanquished/issues/2867). To correctly handle languages with country variants, extra code will be needed in either case. This seems like a waste of time for now so I want to make it *easy* to have single language-only language list.

PR #995, opened 2023-12-14
CM: cleanup + an optimization

slipher:


DolceTriade (approved):
tested. didn't spot any bugs.

PR #996, opened 2023-12-15
Remove unused git revision info

slipher:
There was only one reference to REVISION, in qcommon.h, but it was never used because the relevant header was not included.

It's not implemented reliably anyway. The revision detection only reruns at random times (when cmake stuff is modified), so it is only reliable for a freshly created build dir, like for a release. But for releases we just hard-code the version with update-version-number.sh. The revision could be made always correct by forcing the revision header to be re-created on every build but that's ugly since it slows down incremental builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,42 +296,6 @@ try_flag(WARNINGS           "-Wstrict-null-sentinel")
 try_flag(WARNINGS           "-W${WARNMODE}sign-compare")
 
 ################################################################################
-# Git revision info
-################################################################################
-
-if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git")
-    find_package(Git)
-    if (GIT_FOUND)
-        execute_process(
-            COMMAND ${GIT_EXECUTABLE} describe --tags --long --match v* --dirty=+dirty
-            OUTPUT_VARIABLE GIT_DESCRIBE_REPORT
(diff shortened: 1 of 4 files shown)


illwieckz:
We may prevent the slow down of incremental build by making Q3_VERSION a variable in a dedicated C++ file that would be the only file to include such version header, right?

slipher:
> We may prevent the slow down of incremental build by making `Q3_VERSION` a variable in a dedicated C++ file that would be the only file to include such version header, right?

I have a project where I correctly implemented revision detection and that's what I do, but it still noticeably slows the build. In particular in the case where you changed nothing but redo the build each time you run (e.g. by starting the program from Visual Studio or `ninja && ./daemon`) since the linker has to reassemble the binary every time.

illwieckz:
Maybe we may add an option to disable such feature for developers doing quick rebuild&run?

I still believe it's a good idea to print the commit that was used to build the binary, by default.

slipher:
I guess it would be OK to have an opt-in feature to embed the commit hash, but I'm not volunteering to develop it. For now I just want to delete the currently unused code which slows down the configuration for zero benefit. Note that this code is NOT usable as part of a proper implementation, which needs to have a custom target or something (I forget the exact term) to force it to rerun every time.

illwieckz (approved):
LGTM

PR #998, opened 2023-12-15
Better detect incompatible versions of game VMs

slipher:
Companion: https://github.com/Unvanquished/Unvanquished/pull/2874

The previous mechanism was that one should update some constants CGAME_API_VERSION and GAME_API_VERSION, but no one ever did it. The counters only got to 3 in 10 years of development.

Instead tie gamelogic ABI versioning to "major" Unvanquished releases. The assumption is that in each major release there will always be a few compatibility-breaking changes, while a point release definitely should not have any such changes or else the previous gamelogic can't run.

This will help users get an informative error message when they try to replay a demo or run a mod that is not compatible with the current Daemon.

Additionally there is a constant defined for whether the gamelogic ABI contains breaking changes intended for an upcoming release. This should be set on the next-release branch, for-X.Y.Z/sync. The detection of discrepancies in the compat break bit will help developers keep things straight when they are switching between the upcoming release branch and other branches, forget to rebuild stuff, set the wrong vm type flags etc. This can also be useful when bisecting the engine as you could try to avoid commits from the compat-breaking branch to avoid having to rebuild gamelogic every time.

Updates to Unvanquished's update-version-number.sh will help to automatically update the ABI version and compat break bit.

### DIFF
--- a/src/common/IPC/Common.h
+++ b/src/common/IPC/Common.h
@@ -70,6 +70,19 @@ namespace IPC {
         void Close() const;
 	};
 
+    // Version of the protocol for detecting what ABI version the VM has upon startup
+    constexpr uint32_t ABI_VERSION_DETECTION_ABI_VERSION = 4;
+
+    // Version for the syscall signatures between engine and VM. This must change if the syscall
+    // IDs, argument types, or return types change, or if serialization procedures change.
+    // Follows Daemon major versions.
+    // This should be updated only by update-version-number.py when a "major" release is indicated
+    constexpr const char* SYSCALL_ABI_VERSION = "0.54.0";
+
+    // This should be manually set to true when starting a 'for-X.Y.Z/sync' branch.
+    // This should be set to false by update-version-number.py when a (major) release is created.
+    constexpr bool DAEMON_HAS_COMPATIBILITY_BREAKING_SYSCALL_CHANGES = true;
+
     /*
      * The messages sent between the VM and the engine are defined by a numerical
      * ID used for dispatch, a list of input types and an optional list of return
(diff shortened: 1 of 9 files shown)


illwieckz:
I wonder if `DAEMON_HAS_COMPATIBILITY_BREAKING_SYSCALL_CHANGES` can be a `int`, it would be bumped on every commit introducing a compat-breaking changed, and zeroed on releases. This way we would even detect mismatches between two builds of the “future” branch.

slipher:
> I wonder if `DAEMON_HAS_COMPATIBILITY_BREAKING_SYSCALL_CHANGES` can be a `int`, it would be bumped on every commit introducing a compat-breaking changed, and zeroed on releases. This way we would even detect mismatches between two builds of the “future” branch.

What you are suggesting is basically the current system, which I want to replace with something else

> The previous mechanism was that one should update some constants CGAME_API_VERSION and GAME_API_VERSION, but no one ever did it. The counters only got to 3 in 10 years of development.

The thing I'm talking about here. Supposedly these version numbers would be bumped with each commit breaking compatibility, but they were only bumped a couple of times ever. Even if people did bump then it would be a pain since if there are 2+ compatibility-breaking PRs, they would either get merge conflicts on the version bump, or worse, there would be no conflict and thus the version wouldn't actually increase for some commits.

DolceTriade:
often our engine changes don't actually break compat tho... Maybe there is a  way to hash the syscalls or something and make that when we bump the ABI version?

slipher:
> often our engine changes don't actually break compat tho... Maybe there is a way to hash the syscalls or something and make that when we bump the ABI version?

Well not every commit breaks compat, but probably every non-point-release engine version does. Maybe the syscall interfaces could be hashed, but those aren't everything, since also the serialization can change, or the semantics of the syscall.

DolceTriade:
I suppose it's better than what we have for now, so might as well merge it until we have something better

DolceTriade (approved):

PR #1000, opened 2023-12-20
Some rework of the architecture detection and processing

illwieckz:
Since Debian Bookworm, the armhf compiler does not use armv7 as a baseline but armv6. The compiler also requires to explicitly set the fpu for hardwre float processing. This means our CMake script was failing on Debian Bookworm 32-bit for two reasons: testing for version 7 would detect an unknown architecture, and not setting the fpu would make the compiler unable to build.

While I was at it, I bundled some other commits for other changes I wanted to see.

In the end, the cmake build script should not prevent to build on an unsupported platform if it can work. Let's raise a warning instead and let the user deal with errors himself.

As a test I tried to build a ppc64el native sgame, something that didn't required any external library, and then easy to test. It built. I have no tested it though (I don't have ppc64 hardware, and I didn't built the engine). I see no reason to prevent people to build native servers on exotic platforms if that works, neither to prevent them to try.

This code should make easier to add more architectures or systems, while making our own code more generic and simpler.

I also fixed an issue where building for macOS arm64 or Windows arm64/armhf would be mistakenly considered as supported: we had a list of supported systems and a list of supported architectures, instead of lists of supported architectures per system.

---

-   cmake: detect arm compiler supporting hard float but ignore the target version
    
    In more recent distribution like Debian 12 Bookworm
    or Ubuntu 23.10 Mantic Minautor, it has been observed
    that arm-linux-gnueabihf-gcc sets:
    
     #define __ARM_ARCH 6
     #define __ARM_ARCH_6__ 1
     #define __ARM_PCS_VFP 1
    
    Previously it was setting __ARM_ARCH to 7.
    
    So it looks like the default target is armhf with armv6 as a baseline
    instead of armhf with armv7 as a baseline. Even if we will build for
    armv7 with neon, we need to detect it's an armhf compiler and we may
    just ignore the default arm version. What we need to know is if
    the target is using hard float ABI, hence __ARM_PCS_VFP.
    
    In fact it even exists some armv5 with hard float, so we better not
    check the default target, all we want is to detect hard float, and
    from that we set our default target anyway.

-   cmake: set a default fpu for armv7-a
    
    An FPU should be explicitly set with armv7-a on recent compilers or this error would be raised:
      cc1: error: ‘-mfloat-abi=hard’: selected architecture lacks an FPU

-   common: raise a warning when BREAKPOINT is not implemented

    No need to hard code every known architecture to
    set the workaround.
    
    Do not raise a warning for NaCl missing BREAKPOINT,
    we gave up.

-   cmake: only use -mtune if arch as a generic -mtune option
    
    Some architectures don't have any generic -mtune option.

-   cmake: armel doesn't have generic -mtune option, fp is optional
    
    It is not planned to distribute armel build, but armel native build works.

-   cmake: make -march flag unavailability an error
    
    If we don't report that the configured -march is an error
    we then get random error elsewhere, like a report that C++ 14
    is not available, not because C++ 14 flag raises an error,
    but because the -march one does.
    
    If the -march flag is unavailable, then we cannot use
    the USE_CPU_GENERIC_ARCHITECTURE option so something
    is broken and the user should know.

-   cmake: make recommended compiler flag unavailability an error
    
    If those flags are faulty, the error will be wrongly reported
    for other flags. If those flags are not available, then we cannot
    use USE_CPU_RECOMMENDED_FEATURES option so something is broken and
    the user should know.

-   make: set supported arch per system

-   cmake,nacl: don't fail on unknown platform
    
    It may be possible to build a server with exe sgame
    on unsupported platforms, we can let the one trying
    to build the engine deal with it but not prevent it.

-   DaemonArchitecture: uncomment some known architecture detection

-   DaemonArchitecture: pnacl is a toolchain, nacl is the architecture

-   common: use a single word for the platform name, it makes easier to parse strings


src/common/Platform.h
```diff
@@ -39,7 +39,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #elif defined(__linux__)
 #define PLATFORM_STRING "Linux"
 #elif defined(__native_client__)
-#define PLATFORM_STRING "Native Client"
+#define PLATFORM_STRING "NativeClient"
```
slipher, line 42: I don't like it. But anyway it's not used anywhere

illwieckz: We can also say “NaCl”. I want a single word because this can end up in version strings, and having one word per token makes parsing by third-party software easier.

cmake/DaemonFlags.cmake
```diff
@@ -217,10 +217,11 @@ else()
         set(GCC_GENERIC_ARCH "armv7-a+fp")
         set(GCC_GENERIC_TUNE "generic-armv7-a")
     elseif (ARCH STREQUAL "armel")
-        # Armv6 minimum, optional VFP: ARM11.
-        # Soft float ABI (mainstream 32-bit ARM Android distributions).
+        # Armv6 minimum with optional VFP: ARM11.
+        # Soft float ABI (previous maintstream 32-bit ARM Linux distributions, mainstream 32-bit ARM Android distributions).
```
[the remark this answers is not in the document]

src/common/Platform.h
illwieckz: I've set `NaCl`.

slipher:
LGTM

PR #1001, opened 2023-12-21
Validate shader cache binary size

slipher:
We shouldn't just believe what the shader cache file says the size of its binary program is. If the file were truncated, believing it would cause an out-of-bounds read.

On my machine the Nvidia driver crashed even when incorrectly told that the length of the shader binary is 0, which would seem fairly innocuous...


illwieckz (approved):
LGTM.

Issue #1007, opened 2023-12-30
`zlib.net/<archive>` only provides the latest download


illwieckz:
Zlib 1.3 is out so the `https://zlib.net/zlib-1.2.13.tar.gz` link is dead.

illwieckz:
Fixed by:

- https://github.com/DaemonEngine/Daemon/pull/1008

PR #1008, opened 2023-12-30
external_deps: allow to fetch archives from various mirrors

### LINKED ISSUE #1006: `openal-soft.org` website is down
I noticed it yesterday, https://openal-soft.org doesn't respond (but ping does).

This prevents to fetch the external deps in the docker script to build releases.

Issue is tracked on OpenAL side there:

- https://github.com/kcat/openal-soft/issues/956

I don't know if there are mirrors, if there are, it may be useful to update our script to iterate known mirrors in case the primary source is not available.

illwieckz:
Allow to fetch archives from various mirrors, cycling until one works.

Add a bunch of known mirrors for some packages.

Do some minor rewriting to unify the wording.

Fixes #1006 and #1007:

- https://github.com/DaemonEngine/Daemon/issues/1006
- https://github.com/DaemonEngine/Daemon/issues/1007

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -56,6 +56,12 @@ CXXFLAGS=''
 CPPFLAGS=''
 LDFLAGS=''
 
+log() {
+	level="${1}"; shift
+	printf '%s: %s\n' "${level^^}" "${@}" >&2
+	[ "${level}" != 'error' ]
+}
+
 # Extract an archive into the given subdirectory of the build dir and cd to it
 # Usage: extract <filename> <directory>
 extract() {
@@ -87,8 +93,7 @@ extract() {
 		rmdir "${2}-dmg"
 		;;
 	*)
-		echo "Unknown archive type for ${1}"
-		exit 1
+		log error "Unknown archive type for ${1}"
 		;;
 	esac
 	cd "${2}"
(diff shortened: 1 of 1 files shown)


external_deps/build.sh
```diff
@@ -59,54 +59,65 @@ LDFLAGS=''
...
-	fi
-	extract "${1}" "${3}"
+	local extract_dir="${BUILD_DIR}/${1}"; shift
+	local tarball_file="${DOWNLOAD_DIR}/${1}"; shift
+	while [ ! -f "${tarball_file}" ]; do
+		[ -n "${1:-}" ]
+		local download_url="${1}"; shift
+		echo "Downloading ${download_url}"
+		"${CURL}" -L --fail -o "${tarball_file}" "${download_url}" \
+		|| rm -f "${tarball_file}"
```
DolceTriade, line 107: you should log on error at least

external_deps/build.sh
```diff
@@ -59,54 +59,65 @@ LDFLAGS=''
...
 # Usage: download <filename> <URL> <dir>
 download() {
-	if [ ! -f "${DOWNLOAD_DIR}/${1}" ]; then
-		"${CURL}" -L --fail -o "${DOWNLOAD_DIR}/${1}" "${2}"
-	fi
-	extract "${1}" "${3}"
+	local extract_dir="${BUILD_DIR}/${1}"; shift
+	local tarball_file="${DOWNLOAD_DIR}/${1}"; shift
+	while [ ! -f "${tarball_file}" ]; do
+		[ -n "${1:-}" ]
```
DolceTriade, line 103: Should we do something like:
`[ -n "${1:-}" ] ||( echo "No more URLs passed in!" >&2 && exit -1)`

DolceTriade (approved):

external_deps/build.sh
illwieckz: Yeah I can add more logging, those two things you stumbled on are mostly code golf with implicit behaviour. 😁️
Even the `rm` is less about removing the fail (but curl says `--fail` may not remove the output on some errors, so we better want to make sure we remove it) than to turn the error into a `true` thanks to the `||`. 😅️

I'll make the code more explicit and more verbose.

illwieckz:
@DolceTriade I added a commit to rework the logging of the script. So this change is larger than just applying your juggestions, can you review it as well? Thanks!

illwieckz:
Well, the extra change is just logging and it works, so let's go.

DolceTriade:
lgtm

PR #1013, opened 2024-01-05
external_deps: add our own mirror, and options to only download or prefer our own mirror

illwieckz:
Fixes Unvanquished/unvanquished-infrastructure#37:

- https://github.com/Unvanquished/unvanquished-infrastructure/issues/37

Also brings other improvements and fixes.

This implements:

-   external_deps: tell curl to set the remote timestamp of the downloaded archive

-   external_deps: add --download-only option

-   external_deps: fix all windows-amd64-mingw package list

-   external_deps: add meta platforms like 'all'
    
    This make possible to do:
    
    ./build.sh --download-only all all
    
    or:
    
    ./build.sh build-linux base

-   external_deps: add our own mirror and add --prefer-ours option (disabled by default)

-   external_deps: make options cumulative

-   external_deps: fix help printing on macOS

-   external_deps: prefer xz archives when available

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -99,11 +99,9 @@ extract() {
 	cd "${2}"
 }
 
-# Download a file if it doesn't exist yet, and extract it into the build dir
-# Usage: download <filename> <URL> <dir>
 download() {
-	local extract_dir="${BUILD_DIR}/${1}"; shift
-	local tarball_file="${DOWNLOAD_DIR}/${1}"; shift
+	local tarball_file="${1}"; shift
+
 	while [ ! -f "${tarball_file}" ]; do
 		if [ -z "${1:-}" ]
 		then
@@ -117,6 +115,21 @@ download() {
 			rm -f "${tarball_file}"
 		fi
 	done
+}
+
+# Download a file if it doesn't exist yet, and extract it into the build dir
+# Usage: download <filename> <URL> <dir>
+download_extract() {
+	local extract_dir="${BUILD_DIR}/${1}"; shift
+	local our_mirror="https://dl.unvanquished.net/deps/original/${1}"
+	local tarball_file="${DOWNLOAD_DIR}/${1}"; shift
+
+	if "${prefer_ours}"
+	then
+		download "${tarball_file}" "${our_mirror}" "${@}"
+	else
+		download "${tarball_file}" "${@}" "${our_mirror}"
+	fi
 
 	"${download_only}" && return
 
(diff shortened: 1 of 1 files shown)


external_deps/build.sh
```diff
@@ -1159,17 +1232,74 @@ if [ "${#}" -lt "2" ]; then
...
+	if [ -z "${platform_list}" ]
+	then
+		errorHelp
+	fi
+;;
+esac
+
+for PLATFORM in ${platform_list}
+do
+	"setup_${PLATFORM}"
```
slipher, line 1298: I have doubts that it is really safe to run more than one setup platform. Some of them might set environment variables that should not be set for others. I suppose it could be worked around by running each platform in a new process.

external_deps/build.sh
```diff
@@ -1111,22 +1175,31 @@ all_linux_arm64_default_packages='zlib gmp nettle curl sdl2 glew png jpeg webp f
...
-	\twindows-i686-msvc windows-amd64-msvc windows-i686-mingw windows-amd64-mingw macos-amd64-default linux-amd64-default linux-i686-default linux-arm64-default linux-armhf-default
+	\t${all_platforms}
 
 	Packages:
 	\tpkgconfig nasm zlib gmp nettle curl sdl2 glew png jpeg webp freetype openal ogg vorbis opus opusfile lua naclsdk naclports wasisdk wasmtime
 
 	Virtual packages:
 	\tbase — build packages for pre-built binaries to be downloaded when building the game
 	\tall — build all supported packages that can possibly be involved in building the game
+	\tbuild-linux — build all packages buildable on linux: ${linux_build_platforms}
```
slipher, line 1201: This is listed as a package, but it's supposed to be a platform?

slipher (commented):
I don't like the `src` URL name much as many of the archives are binaries, not sources. Maybe `mirror`? IDK

Who is responsible for uploading things to the mirror and when?

external_deps/build.sh
illwieckz: Ah, yes, I just have to add some `()`

external_deps/build.sh
illwieckz: Ah yes.

external_deps/build.sh
illwieckz: Fixed.

external_deps/build.sh
illwieckz: Fixed.

illwieckz:
> I don't like the `src` URL name much as many of the archives are binaries, not sources. Maybe `mirror`? IDK

Maybe `original`? Some archive may have renamed file names (like those that don't carry any version string), but the archives themselves are not modified, so that's still “original”.

> Who is responsible for uploading things to the mirror and when?

Uploading things to the mirror are not required and not doing it is not blocking, it's just something we can do to be improve robustness of fix a build that breaks, so I don't think we have to designate someone to be responsible for this.

Outside of that OpenAL story I'm concerned that our NaCl files are from 2014, and if we implement Saigo, we may rely in snapshots I have no idea about the availability in time.

slipher:
> Maybe `original`?

Better than `src` I guess. Another idea is `inputs` (inputs for our deps packages).

illwieckz:
I modified it to use `original`.

slipher:
LGTM

PR #1015, opened 2024-01-05
NUKE wrapAroundLighting

illwieckz:
Robert Beckebans, original XreaL author, said:

> Hi Thomas,
> The wrap around lighting term was something that was discussed in a very old game presentation where the devs provided some cheap outdoor lighting.
> I just experimented with it a bit but you can remove it. I would never use it again in such a way and yeah the code is wrong :)


slipher:
LGTM

PR #1017, opened 2024-01-05
Add FPS counter to curses window and /showfps command

slipher:
Implement an exponential moving average algorithm for counting FPS and make it available for the engine in two ways: a `showfps` printing command, and a counter in the curses console. This can be useful to monitor for server performance issues.

I also have a branch to replace the cgame FPS counter with this algorithm (haven't uploaded it yet). It's better than the current algorithm because if you have a perfectly stable FPS, the FPS counter will also be stable (unless it's right at a half-integer I guess). With the current algorithm the FPS tends to flicker between various numbers even if the frame times are very consistent.

### DIFF
--- a/src.cmake
+++ b/src.cmake
@@ -61,6 +61,7 @@ set(COMMONLIST
     ${COMMON_DIR}/String.h
     ${COMMON_DIR}/System.cpp
     ${COMMON_DIR}/System.h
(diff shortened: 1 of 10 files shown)


illwieckz (approved):
LGTM.

Issue #1005, opened 2023-12-29
Portal rendering seems wonky depending on distance


cu-kai:
When testing grangermaze 0.5, I have been struggling to get all of the new portals to render (the timers in the wall are portals all displaying the same view).

Here is an example of how it should look. You can `/setviewpos -3855 3603 75 -21 2` and you will see that both portals in room 5E render correctly. Here is a screenshot showing this:
![unvanquished_2023-12-29_221523_000](https://github.com/DaemonEngine/Daemon/assets/13281185/baaa4552-619c-4f16-b400-1775f5604f2e)

Being any closer than this will cause the portal on the left to disappear and turn black. Sometimes if you are really close, there will be garbage rendered inside the portal surface too similar to the below.
![image](https://github.com/DaemonEngine/Daemon/assets/13281185/b22e2f66-d2eb-4241-b112-cb70d2a08597)

There are 16 of these portals and I'd say around only 50% of them render properly. It was also noted that when entering the rooms with broken portals, some players would crash as per #1002 so I believe this directly related.

It is also worth noting that the original "map" portals on this map still render correctly.

sweet235:
This is correct.

VReaperV:
There appears to be at least two issues here: meshes are rendered that are not supposed to be visible through the portal (there were drawcalls to render surfaces near me when I was looking at grangermaze, but since they are rendered from the point of view of the portals camera, they are off-screen), and sometimes portals are not rendered at all (PreparePortalCommand and FinalisePortalCommand are never called).

There's also a problem with creep meshes apparently. Below is the wireframe mesh that was rendered with the creep texture:
![image](https://github.com/DaemonEngine/Daemon/assets/10687142/bf944d1b-2ce6-4f61-9c72-6d9021b268a2)
![image](https://github.com/DaemonEngine/Daemon/assets/10687142/6cee48b8-01e4-4f54-88f4-2407d0634780)
This is the stencil buffer used by portal rendering for this draw call, white part is what could actually be rendered:
![image](https://github.com/DaemonEngine/Daemon/assets/10687142/fec27f8e-1ab6-483b-9702-cd91d7270cf1)

I think that's what got rendered in the second screenshot in this issue. I'm not sure if this is solely caused by the portal rendering issue, but if not the nwe have an issue with creep rendering.

VReaperV:
[...]

PR #1019, opened 2024-01-06
Fix portal blending with portal surfaces

### LINKED ISSUE #1011: Portals don't blend
This was described in #935:

https://github.com/DaemonEngine/Daemon/issues/935

But in fact portal blending and alphaGen portal are two different features (while the second one may depend on the former though).

We can't blend rendered cameras with special effects, like in `pulse` map:

Tremulous:

[portal over texture](https://dl.illwieckz.net/b/unvanquished/bugs/portal-over-texture/20230313-183015-000.tremulous-pulse-camera-portal.jpg)

Unvanquished:

[portal over texture](https://dl.illwieckz.net/b/unvanquished/bugs/portal-over-texture/20230313-183122-000.unvanquished-pulse-camera-portal.jpg)

The lack of portal blending is technically not a bug to fix, it's a feature request: the feature _has to be implemented_ or _to be finished_,which is not the same amount of work than fixing something that worked before, one cannot make it work by bisecting or something, _the code is not there yet_.

There may be code for that feature to port from ioquake3.

VReaperV:
Enables writing to color buffer when rendering the portal surface in FinalisePortalCommand::ExecuteSelf(). Changes glStencilOp to avoid stencil test failing in light pass, and clears stencil buffer in case there were multiple portals on the screen. Fixes #1011.

`glStencilOp( GL_KEEP, GL_KEEP, GL_DECR )` was likely intended to prevent incorrect rendering with multiple portals on the screen, but it was preventing the portal surface from rendering.

Here's some screenshots:
unvanquished_2024-01-07_020340_000
unvanquished_2024-01-07_020416_000
unvanquished_2024-01-07_020443_000

~~The colour is slightly off because for some reason the portal surfaces are rendered twice, so they blend with themselves.~~ The colour is off due to light pass, which may or may not be a bug with the lighting system. Using Nsight confirms that the first draw produces the expected result (this is not the final colour that will be in the frame, but it illustrates the issue):
debug_portal_blend1
debug_portal_blend2

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -5532,6 +5532,9 @@ const RenderCommand *PreparePortalCommand::ExecuteSelf( ) const
 
 	GL_LoadModelViewMatrix( backEnd.orientation.modelViewMatrix );
 
+	glClearStencil( 0 ); // Clear stencil buffer to avoid overwritting the color buffer when multiple portals are on the screen
+	glClear( GL_STENCIL_BUFFER_BIT ); // Might need to check if backEnd.viewParms.portalLevel == 0
+
 	if ( backEnd.viewParms.portalLevel == 0 ) {
 		glEnable( GL_STENCIL_TEST );
 		glStencilMask( 0xff );
@@ -5605,10 +5608,10 @@ const RenderCommand *FinalisePortalCommand::ExecuteSelf( ) const
 	GL_LoadModelViewMatrix( backEnd.orientation.modelViewMatrix );
 
 	glStencilFunc( GL_EQUAL, backEnd.viewParms.portalLevel + 1, 0xff );
-	glStencilOp( GL_KEEP, GL_KEEP, GL_DECR );
+	glStencilOp( GL_KEEP, GL_KEEP, GL_KEEP );
 
-	GL_State( GLS_DEPTHMASK_TRUE | GLS_DEPTHFUNC_ALWAYS | GLS_COLORMASK_BITS );
-	glState.glStateBitsMask = GLS_DEPTHMASK_TRUE | GLS_DEPTHFUNC_ALWAYS | GLS_COLORMASK_BITS;
+	GL_State( GLS_DEPTHMASK_TRUE | GLS_DEPTHFUNC_ALWAYS);
+	glState.glStateBitsMask = GLS_DEPTHMASK_TRUE | GLS_DEPTHFUNC_ALWAYS;
 
 	Tess_Begin( Tess_StageIteratorGeneric, nullptr, shader,
 		    nullptr, false, false, -1, -1 );


illwieckz:
Awesome ! 😲️

VReaperV:
> The colour is slightly off because for some reason the portal surfaces are rendered twice, so they blend with themselves. Using Nsight confirms that the first draw produces the expected result (this is not the final colour that will be in the frame, but it illustrates the issue):

Apparently that is the light stage. If the output from it is incorrect, then that issue is likely with lighting. It could also be that tremulous was rendering it wrong.

VReaperV:
> `glStencilOp( GL_KEEP, GL_KEEP, GL_DECR )` was likely intended to prevent incorrect rendering with multiple portals on the screen, but it was preventing the portal surface from rendering.

On second thought, that might be for recursive portals. This pr might break them, but we don't seem to have any maps with those, so I'd need a test case map for that.

illwieckz:
> On second thought, that might be for recursive portals. This pr might break them, but we don't seem to have any maps with those, so I'd need a test case map for that.

spacetracks map, viewpos `128 -1950 602 90 0`, you should see the two mirrors at the same time on the left and on the right.

VReaperV:
> spacetracks map, viewpos `128 -1950 602 90 0`, you should see the two mirrors at the same time on the left and on the right.

They don't seem to be in view of each other though? I've just got a test map from @cu-kai and it looks like recursive portals don't even work in 0.54.1.

illwieckz:
It probably works, but not entirely. The fact we can see the two mirrors of spacetracks at the same time in 0.54.1 is only possible because I enabled portal recursion some time ago. If we disable recursion, only one portal is displayed at a time and a warning is printed.

I also made a test map and recursion seems to only be half working.

VReaperV:
Interesting, I think it might be linked to one of the issues described in https://github.com/DaemonEngine/Daemon/issues/1005#issuecomment-1879504829, since it looks like things that should be drawn through a recursive portal aren't added correctly.

illwieckz:
> Apparently that is the light stage. If the output from it is incorrect, then that issue is likely with lighting. It could also be that tremulous was rendering it wrong.

For now ignore any lightmap issue, there is an ongoing PR to fix many lightmap issues and trying to fix lightmap issues before merging that other PR is likely to introduce more bugs and create strong merge conflicts:

- https://github.com/DaemonEngine/Daemon/pull/992

Also I have some ideas about why there may be lightmap issues with portals, but I will not investigate those ideas before that lightmap PR is merged.

So any lightmap issue remaining when rendering portals should not block the merge of that PR.

VReaperV:
Ah, that's good to hear.

VReaperV:
Closing this in favour of #1031, which fixes this issue among some other things without breaking recursive portals.

PR #1023, opened 2024-01-08
sdl_glimp: workaround texture distorsion bug on Intel GPU, fix #909

### LINKED ISSUE #909: Intel UHD: heavily distorted and flickering textures on world surfaces
I have that kind of flickering on various maps, at extremely precise point.
I know it can be the map itself, but it could be just another case of [this](https://github.com/Unvanquished/Unvanquished/issues/2500) too, which might help finding out the problem.

ATCSZ alpha:

I'll try to think about updating this when I find new spots, I have feeling this happens more often than in the past. Maybe I'm wrong.

illwieckz:
Workaround texture distorsion bug on Intel GPU

This bug seems to only affect gfx9 but detecting gfx9 may not be easy.

See:

- https://github.com/DaemonEngine/Daemon/issues/909
- https://gitlab.freedesktop.org/mesa/mesa/-/issues/10224

This workaround was suggested here:

- https://gitlab.freedesktop.org/mesa/mesa/-/issues/10224#note_2230107

I confirm this workarounds the bug on my end on Intel HD 520 on Linux.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1816,6 +1816,25 @@ static void GLimp_InitExtensions()
 		glConfig2.textureGatherAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, ARB_texture_gather, r_arb_texture_gather->value && !foundNvidia340 );
 	}
 
+	{
+		bool foundIntel = Q_stristr( glConfig.vendor_string, "Intel" );
+
+		if ( foundIntel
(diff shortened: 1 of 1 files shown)


illwieckz:
@slipher does it fix the issue on your side on Windows too?

illwieckz:
bmorel confirmed it fixes the issue on UHD 630 on Linux.

illwieckz:
This seems to be good enough to merge.

It's possible that Intel with ship some workaround in drivers in the future, but at least with that code we don't depend on that.

slipher:
I confirm it fixes the bug with map rush-station on my machine.

PR #1026, opened 2024-01-13
renderer: fix vertex lighting with terrain blending and with non-style lightmap with tcGen lightmap

illwieckz:
I guess I've set that value after some trial and error to workaround what I didn't knew was a bug, but I fixed that bug later.

In the current state of the renderer, I don't see bugs popping when removing that part, and this is needed to fix terrain blending.

I also added a commit fixing some bad spacing that was mistakenly merged.


illwieckz:
In thunder map, terrain blending was already working with vertex lighting, but the two stages are opaque:

unvanquished_2024-01-13_153926_000

```c
{

	q3map_nonplanar
	q3map_shadeAngle 90

	{
		diffuseMap textures/shared_pk02_src/rock01_d
		normalMap textures/shared_pk02_src/rock01_n
		specularMap textures/shared_pk02_src/rock01_s
		rgbGen identity
	}
	{
		diffuseMap textures/shared_pk02_src/sand01_d
		normalMap textures/shared_pk02_src/sand01_n
		blendFunc blend
		alphaGen vertex
	}
}
```

In station15, terrain blending was broken with vertex lighting, as one of the stage is not opaque and the code was resetting the alpha channel:

unvanquished_2024-01-13_154527_000

```c

textures/station15_custom/ter_mudmoss1
{
	qer_editorImage textures/station15_custom_src/ter_mudmoss1_p

	q3map_nonplanar
	q3map_shadeangle 120

	{
		diffuseMap textures/station15_custom_src/ter_moss1_d
		rgbGen identity
	}
	{
		diffuseMap textures/station15_custom_src/ter_mud1_d
		alphaFunc GE128
		rgbGen identity
		alphaGen vertex
	}
}
```

Here is how it now works:

unvanquished_2024-01-13_154914_000

I don't see anymore any problem with other maps, example here with rusty (before/after) that was the map I was used when testing vertex lighting and then when I wrote that original code:

unvanquished_2024-01-13_155651_000

unvanquished_2024-01-13_155831_000

I also tested other maps like wolf:et oasis wich is known to do complex mixed lightmap/vertex ligthing and I see no regression when enabling vertex lighting or light mapping.

Scenes used to debug:

map|viewpos
-|-
thunder|3906 473 142 47 13
station15|532 -1189 21 -180 34
rusty|-1134 -2261 111 31 8

illwieckz:
I added a commit to add a debug cvar named `r_showVertexColors` to render the vertex colors used for vertex lighting:

unvanquished_2024-01-13_172100_000

unvanquished_2024-01-13_172014_000

illwieckz:
I added a commit making a difference between a light style map and a light map with useless “tcGen lightmap”.

This fixes #1027:

- https://github.com/DaemonEngine/Daemon/issues/1027

I verified this don't break light styled maps like gloom2 and pulse.

Before, after:

unvanquished_2024-01-13_162254_000

unvanquished_2024-01-13_180723_000

src/engine/renderer/tr_init.cpp
```diff
@@ -1431,6 +1433,21 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 			}
 		}
 
+		if ( r_showNormalMaps->integer
+			|| r_showMaterialMaps->integer )
+		{
+			tr.lightMode = lightMode_t::MAP;
+		}
+		if ( r_showLightMaps->integer
```
slipher, line 1441: Should this be else if?

src/engine/renderer/tr_init.cpp
```diff
@@ -77,7 +77,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_dynamicLightCastShadows;
 	cvar_t      *r_precomputedLighting;
 	Cvar::Cvar<int> r_mapOverBrightBits("r_mapOverBrightBits", "default map light color shift", Cvar::NONE, 2);
-	Cvar::Range<Cvar::Cvar<int>> r_lightMode("r_lightMode", "lighting mode: 0: fullbright (cheat), 1: vertex light, 2: grid light (cheat), 3: light map", Cvar::NONE, Util::ordinal(lightMode_t::MAP), Util::ordinal(lightMode_t::FULLBRIGHT), Util::ordinal(lightMode_t::MAP));
+	Cvar::Range<Cvar::Cvar<int>> r_lightMode("r_lightMode", "lighting mode: 0: fullbright (cheat), 1: vertex light, 2: grid light (cheat), 3: light map", Cvar::ARCHIVE, Util::ordinal(lightMode_t::MAP), Util::ordinal(lightMode_t::FULLBRIGHT), Util::ordinal(lightMode_t::MAP));
```
slipher, line 80: We shouldn't need to make every cvar configurable by the UI `ARCHIVE` anymore. The UI control adds an archive flag when you use it to change the value.

src/engine/renderer/tr_init.cpp
illwieckz: Fixed.

src/engine/renderer/tr_init.cpp
illwieckz: I removed the commit.

slipher:
LGTM

PR #1029, opened 2024-01-15
Keyboard input handling: refactoring and minor improvements

slipher:
This restructuring is needed for the code to implement input to the UI and binds simultaneously.


illwieckz:
>  Always stop bind command when key is released
>
> Previously the "keyup" command notifying that the key was released was
> only sent if the UI does not have focus.
>
> This makes no difference in practice because the Unvanquished gamelogic
> always calls Key_ClearStates, which stops all bind commands, when
> setting KEYCATCH_UI.

This may fix the bug where scripts can't build buildables unless the game window is unfocused.

illwieckz (approved):
LGTM.

slipher:
> This may fix the bug where scripts can't build buildables unless the game window is unfocused.

I doubt it. Is this bug reported somewhere?

PR #1030, opened 2024-01-15
Allow keyboard input to UI and binds simultaneously

slipher:
Stacked on #1029. Please review that one first.

Gamelogic companion: https://github.com/Unvanquished/Unvanquished/pull/2900

The cgame can now, if it wishes, receive key up and key down events while binds are active. Upon receiving a key down event, the cgame will return a boolean indicating whether the keystroke was consumed by the
cgame. If the keystroke was consumed by the cgame, no binds will be  executed.

Requires breaking ABI changes to the key event and key catcher interfaces.


DolceTriade:
tested it and seemed ok to me.

slipher:
https://github.com/Unvanquished/Unvanquished/pull/2900 was approved so now I just need an LGTM on this one.

DolceTriade:
Lgtm

PR #1031, opened 2024-01-15
Fix portal surface blending, entity selection and recursive mirrors

### LINKED ISSUE #1011: Portals don't blend

VReaperV:
Fixes portal blending with portal surfaces, misc_portal_surface entity selection and GL_FrontFace for mirrors in chains of recursive portals/mirrors. Fixes #1011. Disables the 64qu distance limit for portal entities which wasn't working properly anyway.

Portal entities are now chosen based on their distance to the center of the portals surface (calculated as the average of all vertices). Old code was checking if they are within 64qu of the plane, which was resulting in portals entities that are actually much further away being chosen.

Introduces Tess_StageIteratorPortal(), intended for portals to properly write to stencil buffer.

Screenshots (some blended surfaces look slightly off, this might be an issue with lighting in general, portals themselves blend correctly):
unvanquished_2024-01-09_210516_000
unvanquished_2024-01-09_203209_000
unvanquished_2024-01-09_195110_000
unvanquished_2024-01-09_195048_000
unvanquished_2024-01-07_020443_000
unvanquished_2024-01-07_020416_000
unvanquished_2024-01-07_020340_000

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -386,7 +386,7 @@ void GL_PolygonOffset( float factor, float units )
 
 void GL_Cull( cullType_t cullType )
 {
-	if ( backEnd.viewParms.isMirror )
+	if ( backEnd.viewParms.mirrorLevel % 2 == 1 )
 	{
 		GL_FrontFace( GL_CW );
 	}
@@ -5543,7 +5543,7 @@ const RenderCommand *PreparePortalCommand::ExecuteSelf( ) const
 	GL_State( GLS_COLORMASK_BITS );
 	glState.glStateBitsMask = GLS_COLORMASK_BITS;
 
-	Tess_Begin( Tess_StageIteratorGeneric, nullptr, shader,
+	Tess_Begin( Tess_StageIteratorPortal, nullptr, shader,
 		    nullptr, false, false, -1, -1 );
 	rb_surfaceTable[Util::ordinal(*(surface->surface))](surface->surface );
 	Tess_End();
@@ -5556,10 +5556,10 @@ const RenderCommand *PreparePortalCommand::ExecuteSelf( ) const
 	glStencilFunc( GL_EQUAL, backEnd.viewParms.portalLevel + 1, 0xff );
 	glStencilOp( GL_KEEP, GL_KEEP, GL_KEEP );
 
-	GL_State( GLS_DEPTHMASK_TRUE | GLS_DEPTHTEST_DISABLE | GLS_COLORMASK_BITS );
-	glState.glStateBitsMask = GLS_DEPTHMASK_TRUE | GLS_DEPTHTEST_DISABLE | GLS_COLORMASK_BITS;
+	GL_State( GLS_DEPTHMASK_TRUE | GLS_COLORMASK_BITS | GLS_DEPTHFUNC_ALWAYS);
+	glState.glStateBitsMask = GLS_DEPTHMASK_TRUE | GLS_COLORMASK_BITS | GLS_DEPTHFUNC_ALWAYS;
 
-	Tess_Begin( Tess_StageIteratorGeneric, nullptr, shader,
+	Tess_Begin( Tess_StageIteratorPortal, nullptr, shader,
 		    nullptr, false, false, -1, -1 );
 	rb_surfaceTable[Util::ordinal(*(surface->surface))](surface->surface );
 	Tess_End();
@@ -5604,16 +5604,29 @@ const RenderCommand *FinalisePortalCommand::ExecuteSelf( ) const
 
 	GL_LoadModelViewMatrix( backEnd.orientation.modelViewMatrix );
 
-	glStencilFunc( GL_EQUAL, backEnd.viewParms.portalLevel + 1, 0xff );
-	glStencilOp( GL_KEEP, GL_KEEP, GL_DECR );
+	GL_State( GLS_DEPTHMASK_TRUE | GLS_DEPTHFUNC_ALWAYS );
+	glState.glStateBitsMask = GLS_DEPTHMASK_TRUE | GLS_DEPTHFUNC_ALWAYS;
 
-	GL_State( GLS_DEPTHMASK_TRUE | GLS_DEPTHFUNC_ALWAYS | GLS_COLORMASK_BITS );
-	glState.glStateBitsMask = GLS_DEPTHMASK_TRUE | GLS_DEPTHFUNC_ALWAYS | GLS_COLORMASK_BITS;
+	glStencilFunc( GL_EQUAL, backEnd.viewParms.portalLevel + 1, 0xff );
+	glStencilOp( GL_KEEP, GL_KEEP, GL_KEEP );
 
 	Tess_Begin( Tess_StageIteratorGeneric, nullptr, shader,
+		nullptr, false, false, -1, -1 );
+	rb_surfaceTable[Util::ordinal( *( surface->surface ) )]( surface->surface );
+	Tess_End();
+
+	glStencilOp( GL_KEEP, GL_KEEP, GL_DECR );
+
+	GL_State( GLS_COLORMASK_BITS | GLS_DEPTHFUNC_ALWAYS);
+	glState.glStateBitsMask = GLS_COLORMASK_BITS | GLS_DEPTHFUNC_ALWAYS;
+
+	Tess_Begin( Tess_StageIteratorPortal, nullptr, shader,
 		    nullptr, false, false, -1, -1 );
 	rb_surfaceTable[Util::ordinal(*(surface->surface))](surface->surface );
 	Tess_End();
+
+	glStencilFunc( GL_EQUAL, backEnd.viewParms.portalLevel, 0xff );
+	glStencilOp( GL_KEEP, GL_KEEP, GL_KEEP );
 
 	glState.glStateBitsMask = 0;
 
--- a/src/engine/renderer/tr_main.cpp
+++ b/src/engine/renderer/tr_main.cpp
@@ -1301,28 +1301,39 @@ static bool R_GetPortalOrientations( drawSurf_t *drawSurf, orientation_t *surfac
 	// locate the portal entity closest to this plane.
 	// origin will be the origin of the portal, origin2 will be
 	// the origin of the camera
-	for ( i = 0; i < tr.refdef.numEntities; i++ )
-	{
-		e = &tr.refdef.entities[ i ];
+	int numVertsOld = tess.numVertexes;
+	rb_surfaceTable[Util::ordinal( *( drawSurf->surface ) )]( drawSurf->surface );
+	int numVerts = tess.numVertexes - numVertsOld;
+	vec3_t portalCenter{ 0.0, 0.0, 0.0 };
+	for ( int vertIndex = 0; vertIndex < numVerts; vertIndex++ ) {
+		VectorAdd( portalCenter, tess.verts[vertIndex].xyz, portalCenter );
+	}
+	VectorScale( portalCenter, 1.0 / numVerts, portalCenter );
+	float minDistance = FLT_MAX;
+	trRefEntity_t* currentPortal = nullptr;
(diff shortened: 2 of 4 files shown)


slipher:
Could you give an overview of how portal rendering works? Stuff like, does it render camera view to a separate buffer, or does it do some projection to draw them directly into the current scene?

slipher:
> Portal entities are now chosen based on their distance to the center of the portals surface (calculated as the average of all vertices). Old code was checking if they are within 64qu of the plane, which was resulting in portals entities that are actually much further away being chosen.

If you change how a portal surface's view location, does that mean you would need to update q3map2 as well? I believe it needs to understand this to calculate the PVS.

VReaperV:
> Could you give an overview of how portal rendering works? Stuff like, does it render camera view to a separate buffer, or does it do some projection to draw them directly into the current scene?

Sure. It does in fact just change projection.
1) In each R_SortDrawSurfs, all portal surfaces are rendered first after a depth pass (all surfaces with SS_PORTAL shader).
2) For each portal surface, a new viewParms is created that should represent the correct camera view. All render entities are iterated and the closest entity with type RT_PORTALSURFACE is chosen (old code was choosing the first such entity that was within 64qu of the plane the surface was on). The pvsOrigin and orientation of the new view is calculated, which is currently also done incorrectly (there will be a pr later to fix that).
3) Portal surface is drawn into depth and stencil buffer, stencil is increased by one, depth is set to max.
4) Portal frustums and projection matrix are set up to render the correct portal view onto the portal surface. The view is rendered which can cause another portal to render as above.
5) The portal surface is rendered again (back with the previous viewParms), first into depth and colour buffer to blend the portal view with the portal surface and to set the proper depth, then into stencil to decrease it by 1.
6) The rest of the view is rendered normally.

VReaperV:
> If you change how a portal surface's view location is selected, does that mean you would need to update q3map2 as well? I believe it needs to understand this to calculate the PVS.

No, since it uses the same code as any "normal" view rendering does, any viewpos on the map that has the correct pvs if you just go there or noclip to it will also have the correct pvs when rendering a portal. Even if it's e. g. a room that is "off-map" (not supposed to be able to get there or see it normally during gameplay), it still renders only what's in that room.

slipher:
> > If you change how a portal surface's view location is selected, does that mean you would need to update q3map2 as well? I believe it needs to understand this to calculate the PVS.
> 
> No, since it uses the same code as any "normal" view rendering does, any viewpos on the map that has the correct pvs if you just go there or noclip to it will also have the correct pvs when rendering a portal. Even if it's e. g. a room that is "off-map" (not supposed to be able to get there or see it normally during gameplay), it still renders only what's in that room.

The PVS is not (primarily) for the renderer. It's for the server to know which entities it needs to send to a client because the client might be able to see them. So the map compiler needs to know which areas can be seen through a portal, so it can add those areas to the PVS of the area with the portal texture.

src/engine/renderer/tr_backend.cpp
```diff
@@ -5604,17 +5604,31 @@ const RenderCommand *FinalisePortalCommand::ExecuteSelf( ) const
...
+
+	glStencilFunc( GL_EQUAL, backEnd.viewParms.portalLevel + 1, 0xff );
+	glStencilOp( GL_KEEP, GL_KEEP, GL_KEEP );
+
+	Tess_Begin( Tess_StageIteratorGeneric, nullptr, shader,
+		nullptr, false, false, -1, -1 );
+	rb_surfaceTable[Util::ordinal( *( surface->surface ) )]( surface->surface );
+	Tess_End();
+
 	glStencilFunc( GL_EQUAL, backEnd.viewParms.portalLevel + 1, 0xff );
```
slipher, line 5618: The stencil calls now appear twice. Is that intentional?

src/engine/renderer/tr_main.cpp
```diff
@@ -1301,40 +1301,48 @@ static bool R_GetPortalOrientations( drawSurf_t *drawSurf, orientation_t *surfac
...
 			continue;
 		}
 
 		// get the pvsOrigin from the entity
 		VectorCopy( e->e.oldorigin, pvsOrigin );
 
 		// if the entity is just a mirror, don't use as a camera point
-		if ( e->e.oldorigin[ 0 ] == e->e.origin[ 0 ] && e->e.oldorigin[ 1 ] == e->e.origin[ 1 ] && e->e.oldorigin[ 2 ] == e->e.origin[ 2 ] )
-		{
+		if ( e->e.oldorigin[0] == e->e.origin[0] && e->e.oldorigin[1] == e->e.origin[1] && e->e.oldorigin[2] == e->e.origin[2] ) {
```
slipher, line 1325: It would be preferable to avoid whitespace changes on otherwise unmodified lines

src/engine/renderer/tr_shade.cpp
```diff
@@ -2882,6 +2882,89 @@ void Tess_StageIteratorGeneric()
...
+			case stageType_t::ST_COLLAPSE_reflection_CB:
+			case stageType_t::ST_REFLECTIONMAP:
+			case stageType_t::ST_REFRACTIONMAP:
+			case stageType_t::ST_DISPERSIONMAP:
+			case stageType_t::ST_SKYBOXMAP:
+			case stageType_t::ST_SCREENMAP:
+			case stageType_t::ST_PORTALMAP:
+			case stageType_t::ST_HEATHAZEMAP:
+			case stageType_t::ST_LIQUIDMAP:
+				Render_generic( pStage );
```
slipher, line 2944: What's the motivation of creating a new StageIterator function and why do we want to use Render_generic on things which are usually rendered with a different function?

src/engine/renderer/tr_backend.cpp
VReaperV: You mean the glStencilFunc() call? One of those can be removed, good call

src/engine/renderer/tr_shade.cpp
VReaperV: When writing stencil the portal surface should only be rendered once, otherwise it will be rendered with light stages and the like, which would break stencil. (https://github.com/DaemonEngine/Daemon/pull/1031/commits/d0ecf605d3f69f03dac57ef3acce2e1d4a745f19 updates it slightly to make sure only one stage is rendered)
Render_generic seemed to be the most appropriate there since only depth and stencil output are required at that stage, no extra information.

src/engine/renderer/tr_main.cpp
VReaperV: Fixed.

src/engine/renderer/tr_backend.cpp
VReaperV: Assuming the above, fixed

VReaperV:
> The PVS is not (primarily) for the renderer. It's for the server to know which entities it needs to send to a client because the client might be able to see them. So the map compiler needs to know which areas can be seen through a portal, so it can add those areas to the PVS of the area with the portal texture.

It seems to get the correct entities from my testing.

VReaperV:
Ah, that's good!

src/engine/renderer/tr_shade.cpp
VReaperV: The depth only pass can be changed to Tess_StageIteratorDepthFill() though.

VReaperV: Actually, even that shouldn't be changed since otherwise it might do a wasteful ST_COLLAPSE_lighting_PBR stage render.

src/engine/renderer/tr_shade.cpp
```diff
@@ -2882,6 +2882,89 @@ void Tess_StageIteratorGeneric()
...
+
+			default:
+				break;
+		}
+
+		if ( stageFound ) {
+			break;
+		}
+
+		if ( r_showLightMaps->integer && pStage->type == stageType_t::ST_LIGHTMAP ) {
```
slipher, line 2957: This can be removed since it already always breaks out of the loop

VReaperV: Fixed.

illwieckz:
In the chat you said there would also be a patch on game code side, so I don't know if the bugs I see when running this branch are happening only because I miss that other patch, but on my end some portals are acting as mirrors.

Exemple with the the mirror-mirror-portal-mirror chain that is now a mirror-mirror-mirror chain:

map: `test-portal-recursion`
viewpos: `1397 -315 16 0 0`

unvanquished_2024-01-18_152516_000

unvanquished_2024-01-18_153723_000

Example in the portal+mirror room, the one on the left (the one in front of you when entering the room) should be a portal displaying the camera from the room that is next to it on the right:

map: `test-portal-recursion`
viewpos: `-1287 464 74 45 0`

unvanquished_2024-01-18_153028_000

The room next to it on the left of this one is a mirror+mirror room and it's expected to behave like this, but not this one.

illwieckz:
As a remind of what is a mirror and what is a portal, the portal entities without camera links are mirrors:

20240118-153953-000 test-portal-recursion

illwieckz:
Also I spotted there a missing wall in mirror, in alien base, viewpos `-3010 -65 60 -60 0`:

unvanquished_2024-01-18_155228_000

VReaperV:
> In the chat you said there would also be a patch on game code side, so I don't know if the bugs I see when running this branch are happening only because I miss that other patch, but on my end some portals are acting as mirrors.

Thank you, I removed the portalCenter initialization when I was reverting the other changes to make this pr since I never saw them get initialized anywhere else and thought they would be initialized to {0.0, 0.0, 0.0}. Fixed that now.

Broken mirror on "1397 -315 16 0 0" also fixed, as well as the room with a portal and a mirror.

src/engine/renderer/tr_shade.cpp
```diff
@@ -2882,6 +2882,74 @@ void Tess_StageIteratorGeneric()
...
+			case stageType_t::ST_REFLECTIONMAP:
+			case stageType_t::ST_REFRACTIONMAP:
+			case stageType_t::ST_DISPERSIONMAP:
+			case stageType_t::ST_SKYBOXMAP:
+			case stageType_t::ST_SCREENMAP:
+			case stageType_t::ST_PORTALMAP:
+			case stageType_t::ST_HEATHAZEMAP:
+			case stageType_t::ST_LIQUIDMAP:
+				Render_generic( pStage );
+				stageFound = true;
```
slipher, line 2939: `return` ?

slipher (commented):
LGTM

src/engine/renderer/tr_shade.cpp
VReaperV: Thanks, fixed.

illwieckz:
The blending issue in maps like school is because the surface is not set as `bspSurface` and has `lightmapNum` sets to `-1` (no lightmap), this can be adressed in another PR.

illwieckz (approved):
This PR looks ready to me too as a “big first step”, remaining problems can be adressed in subsequent PRs.

sweet235:
> This PR looks ready to me too as a “big first step”, remaining problems can be adressed in subsequent PRs.

I agree. We should not get stuck in trying to improve one PR until the end of days.

illwieckz:
Once CI has verified everything, you can modify the merge option and select “Rebase and merge”, this will avoid the creation of an useless merge commit.

VReaperV:
Allright, will do.

PR #1032, opened 2024-01-23
renderer: make GLSL code easier to read with computeLight, computeDeluxeLight and computeDynamicLights

illwieckz:
Make GLSL code easier to read with computeLight, computeDirectionalLight and computeDynamicLights.

The real magic is in `lightMapping_fp.glsl`, it makes easier to spot the actual alternative code to directional light computation that was previously disseminated everywhere, we now have a simple if/else condition calling one or the other function.

### DIFF
--- a/src/engine/renderer/glsl_source/lightMapping_fp.glsl
+++ b/src/engine/renderer/glsl_source/lightMapping_fp.glsl
@@ -118,21 +118,13 @@ void main()
 		// Compute light color from world space lightmap.
 		vec3 lightColor = texture2D(u_LightMap, var_TexLight).rgb;
 
-		#if defined(USE_DELUXE_MAPPING) || defined(USE_GRID_DELUXE_MAPPING)
-			color.rgb = vec3(0.0);
-		#else
-			color.rgb = lightColor.rgb * diffuse.rgb;
-		#endif
+		color.rgb = vec3(0.0);
 	#else
 		// Compute light color from lightgrid.
 		vec3 ambientColor, lightColor;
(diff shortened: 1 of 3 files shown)


src/engine/renderer/glsl_source/lightMapping_fp.glsl
```diff
@@ -161,14 +153,16 @@ void main()
...
 	#endif
 
+	// Blend static light.
 	#if defined(USE_DELUXE_MAPPING) || defined(USE_GRID_DELUXE_MAPPING)
-		// Blend static light.
-		computeLight(lightDir, normal, viewDir, lightColor, diffuse, material, color);
+		computeDirectionalLight(lightDir, normal, viewDir, lightColor, diffuse, material, color);
+	#else
+		computeLight(lightColor, diffuse, color);
 	#endif
```
illwieckz, line 161: This is what I wanted to achieve.

slipher:
"Directional light" is a term of art ([see this for example](https://learnopengl.com/Lighting/Light-casters)) meaning something else, so it would be better to avoid this name.

illwieckz:
At first I wrote “DeluxeLight”, I don't know if that means something, but at least in this context we would know what it is about.

slipher:
DeluxeLight works for me.

illwieckz:
What about `computeSpecularLight`?

illwieckz:
Well, computing specular light is only a subset of that compute…

illwieckz:
I've set `computeDeluxeMap`.

slipher:
LGTM

PR #1035, opened 2024-01-24
renderer: implement mapOverBrightBits in GLSL

illwieckz:
Implement mapOverBrightBits in GLSL.

This is required for deluxe mapping since we need to do the computation on lights without blended attenuation.

This will also be required when using lightmaps in non-linear colorspace as the computation would have to be done on lights after they are linearized.

### DIFF
--- a/src/engine/renderer/glsl_source/generic_fp.glsl
+++ b/src/engine/renderer/glsl_source/generic_fp.glsl
@@ -33,7 +33,25 @@ IN(smooth) vec2         var_FadeDepth;
 uniform sampler2D       u_DepthMap;
 #endif
 
+#if !defined(GENERIC_2D) && defined(USE_TCGEN_LIGHTMAP)
+uniform int u_MapOverBrightBitsShift;
+#endif
+
 DECLARE_OUTPUT(vec4)
+
+void mapOverBrightBits(inout vec3 color, in float shift)
+{
+	if (shift > 1)
+	{
+		color *= shift;
+		float maxComponent = 1.0;
+		maxComponent = max(maxComponent, color.r);
+		maxComponent = max(maxComponent, color.g);
+		maxComponent = max(maxComponent, color.b);
+		maxComponent = 1.0 / maxComponent;
+		color *= maxComponent;
+	}
+}
 
 void	main()
 {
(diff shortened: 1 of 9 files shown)


src/engine/renderer/glsl_source/generic_fp.glsl
```diff
@@ -33,8 +33,26 @@ IN(smooth) vec2         var_FadeDepth;
 uniform sampler2D       u_DepthMap;
 #endif
 
+#if !defined(GENERIC_2D) && defined(USE_TCGEN_LIGHTMAP)
+uniform int u_MapOverBrightBitsShift;
```
slipher, line 37: I don't like the name with "shift" as that describes a bit shift e.g. shift by 3 means multiply by 8. How about u_OverbrightMultiplier?

slipher (commented):
Can you say which changes to the GLSL correspond to lightmaps, light grids, and vertex lighting?

illwieckz:
I think I will postpone that work for after 0.55 is released, with other things like sRGB. Such work may uncover some bugs or require more redesign in other place and may be the entrance of a rabbit hole 😄️.

src/engine/renderer/glsl_source/generic_fp.glsl
illwieckz: I can do that.

illwieckz:
> Can you say which changes to the GLSL correspond to lightmaps, light grids, and vertex lighting?

I may have to re-introduce `R_ColorShiftLightingBytes` for vertex lighting as there is no knowledge in GLSL if the color is a vertex color or something else. This branch was based on old branch I wrote years ago, at a time we did not have vertex lighting yet.

All the code in GLSL in this branch is about both lightmap and gridlight. PR #1032 made possible to write a single `mapOverBrightBits` function call for non-ambient light in lightMapping GLSL for both light maps and light grid (light maps don't have ambient light).

- https://github.com/DaemonEngine/Daemon/pull/1032

illwieckz:
This implementation has no future.

The design of the optimal code for us is deeply tied to our linear/sRGB implementation, so it will be part of #1034:

src/engine/renderer/glsl_source/generic_fp.glsl
```diff
@@ -33,8 +33,26 @@ IN(smooth) vec2         var_FadeDepth;
...
+{
+	if (shift > 1)
+	{
+		color *= shift;
+		float maxComponent = 1.0;
+		maxComponent = max(maxComponent, color.r);
+		maxComponent = max(maxComponent, color.g);
+		maxComponent = max(maxComponent, color.b);
+		maxComponent = 1.0 / maxComponent;
+		color *= maxComponent;
```
illwieckz, line 52: This was actually repoducing a bug (light being clamped and then wrong)…

PR #1036, opened 2024-01-24
compiler: avoid breakpoint warning when building nacl gamecode (fix CI)

illwieckz:
The `DAEMON_ARCH_pnacl` define is now `DAEMON_ARCH_nacl` (PNaCl is a toolchain, NaCl is an architecture), without this change, the Game CI raises this warning and aborts.


slipher:
LGTM

Issue #1002, opened 2023-12-28
Engine crash with grangermaze 0.5+1


illwieckz:
Test map:

- http://dl.zittrig.eu/pkg/map-grangermaze_0.5%2B1.dpk

Backtrace:

```
Thread 1 "daemon" received signal SIGSEGV, Segmentation fault.
RB_RenderDrawSurfaces (fromSort=fromSort@entry=shaderSort_t::SS_ENVIRONMENT_FOG, toSort=toSort@entry=shaderSort_t::SS_OPAQUE, drawSurfFilter=drawSurfFilter@entry=DRAWSURFACES_ALL) at Unvanquished/daemon/src/engine/renderer/tr_backend.cpp:870
870			} else if( !( entity->e.renderfx & RF_DEPTHHACK ) ) {

Thread 1 (Thread 0x7ffff5517a80 (LWP 457030) "daemon"):
#0  RB_RenderDrawSurfaces (fromSort=fromSort@entry=shaderSort_t::SS_ENVIRONMENT_FOG, toSort=toSort@entry=shaderSort_t::SS_OPAQUE, drawSurfFilter=drawSurfFilter@entry=DRAWSURFACES_ALL) at Unvanquished/daemon/src/engine/renderer/tr_backend.cpp:870
#1  0x000055555568c510 in RB_RenderView (depthPass=false) at Unvanquished/daemon/src/engine/renderer/tr_backend.cpp:4752
#2  DrawViewCommand::ExecuteSelf (this=0x7fffd6e92b94) at Unvanquished/daemon/src/engine/renderer/tr_backend.cpp:5640
#3  0x0000555555680479 in RB_ExecuteRenderCommands (data=<optimized out>) at Unvanquished/daemon/src/engine/renderer/tr_backend.cpp:5884
#4  0x000055555569e1c2 in RE_EndFrame (frontEndMsec=0x0, backEndMsec=0x0) at Unvanquished/daemon/src/engine/renderer/tr_cmds.cpp:889
#5  0x0000555555620a90 in SCR_UpdateScreen () at Unvanquished/daemon/src/engine/client/cl_scrn.cpp:327
#6  0x0000555555614fed in CL_Frame (msec=msec@entry=15) at Unvanquished/daemon/src/engine/client/cl_main.cpp:2094
#7  0x00005555555a970f in Com_Frame () at Unvanquished/daemon/src/engine/qcommon/common.cpp:1028
[...]

PR #1037, opened 2024-01-26
Fix portal rotation

VReaperV:
Complementary pr to: https://github.com/Unvanquished/Unvanquished/pull/2910

Previously, we had a naive implementation of portal rotations (i. e. the origin and axis of the view rendered in a portal surface), which resulted in incorrect rendering, e. g. on the map grangermaze.

Changes portal calculations to get proper views using local->world and world->local transforms with quats. Quats were chosen for convenience and because I plan to do some things with portals which would involve caching them - likely in cgame or sgame, so storing and potentially sending them from the server as 4 floats instead of e. g. 16 for a matrix would be better. Removed skinNum usage because it was previously used to give roll to cameras. It wasn't working and now it will be parsed in sgame anyway.

The code for "bobbing" portal cameras has been moved and modified a bit, so now the axis of the portal entity itself changes (this only affects rendering).

The transformations for portals that are not mirrors are done as follows:
1. A world->portal camera quat is created using the portal entities angles.
2. A surface->world quat is created using the surfaces normal.
3. The current views forward and right vectors are transformed.
4. First the projection of the vector onto worlds XY plane is transformed with yaw (this is needed to get the correct rotation).
5. The z coordinate is then set again and the vector is rotated by the pitch angle, and then the roll angle (all 3 have to be separate to get the correct results).
6. The resulting vector that was essentially rotated by the inverse of the surfaces rotation, is then rotated with the surface->portal camera quat to get the correct view vector for the portal.
7. The up vector is calculated using cross product, and is flipped for mirrors.

Mirror views are now calculated as `out = in - ( in * normal ) x normal`, where in is the current view vector, out is the new view vector, and normal is the portal surfaces normal (* - dot product, x - cross product).

Fixed some formatting issues in the affected code.

R_MirrorVector() removed since it is now unused.

I've checked old maps with portals that I could find with this change and they work correctly, except gauntlet, which is caused by map bugs (old code and even tremulous were not getting the portal view origin correctly). Thankfully, fixing those issues is easy, and you can find the updated maps [here](https://users.unvanquished.net/~reaper/maps/portal-rotation-fix/). Additionally, grangermaze will require a quick fix since Kai had to use incorrect values to work around this issue, but you can find an edited version which sets all portal timers to use the same portal camera with the correct orientation (uses both "target" and "roll") at the link above.

Screenshots:
unvanquished_2024-01-26_040255_000
unvanquished_2024-01-26_040304_000
unvanquished_2024-01-26_040331_000
unvanquished_2024-01-26_040314_000
unvanquished_2024-01-26_040419_000
unvanquished_2024-01-26_040448_000
unvanquished_2024-01-26_040458_000
unvanquished_2024-01-26_040519_000
unvanquished_2024-01-26_040549_000
unvanquished_2024-01-26_040625_000
unvanquished_2024-01-26_040744_000

### DIFF
--- a/src/engine/renderer/tr_main.cpp
+++ b/src/engine/renderer/tr_main.cpp
@@ -1188,20 +1188,6 @@ void R_MirrorPoint( vec3_t in, orientation_t *surface, orientation_t *camera, ve
 	VectorAdd( transformed, camera->origin, out );
 }
 
-void R_MirrorVector( vec3_t in, orientation_t *surface, orientation_t *camera, vec3_t out )
-{
-	int   i;
-	float d;
-
-	VectorClear( out );
-
-	for ( i = 0; i < 3; i++ )
-	{
-		d = DotProduct( in, surface->axis[ i ] );
-		VectorMA( out, d, camera->axis[ i ], out );
-	}
-}
-
 /*
 =============
 R_PlaneForSurface
@@ -1262,7 +1248,7 @@ Returns true if it should be mirrored
 =================
 */
 static bool R_GetPortalOrientations( drawSurf_t *drawSurf, orientation_t *surface, orientation_t *camera, vec3_t pvsOrigin,
-    bool *mirror )
+    bool *mirror, vec3_t *outOrigin, vec3_t *outAxis )
 {
 	int           i;
 	cplane_t      originalPlane, plane;
@@ -1310,7 +1296,7 @@ static bool R_GetPortalOrientations( drawSurf_t *drawSurf, orientation_t *surfac
 	}
 	VectorScale( portalCenter, 1.0 / numVerts, portalCenter );
 	float minDistance = FLT_MAX;
-	trRefEntity_t* currentPortal = nullptr;
+	trRefEntity_t *currentPortal = nullptr;
 	for ( i = 0; i < tr.refdef.numEntities; i++ ) {
 		e = &tr.refdef.entities[i];
 
@@ -1324,11 +1310,10 @@ static bool R_GetPortalOrientations( drawSurf_t *drawSurf, orientation_t *surfac
 			currentPortal = e;
 		}
 	}
-	if( currentPortal) {
+	if( currentPortal ) {
 		// project the origin onto the surface plane to get
 		// an origin point we can rotate around
 		e = currentPortal;
-		VectorCopy( e->e.oldorigin, pvsOrigin );
 		d = DotProduct( e->e.origin, plane.normal ) - plane.dist;
 		VectorMA( e->e.origin, -d, surface->axis[ 0 ], surface->origin );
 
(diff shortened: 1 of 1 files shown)


illwieckz (approved):
I will have to trust you for the maths, but I confirm this work and I see nothing wrong from my limited point of view.

I don't like the comments being on the same line than code, I prefer if such common is on the immediately preceding line, like this:

```c++
mycode(bla) + something; // comment
```

Turned into:

```c++
// comment
mycode(bla) + something;
```

Not only it prevents to have very long line, but it also means editing a comment don't pollute `git blame` and also make merge/rebase less conflictual.

So it LGTM but I prefer if you can reformat those comments before merging.

VReaperV:
Ah, sure, I'll do that a bit later.

VReaperV:
I can write a more in-depth explanation on the math used, but all the portals I tested (both on test maps and old/existing maps) work properly.

illwieckz:
> I can write a more in-depth explanation on the math used, but all the portals I tested (both on test maps and old/existing maps) work properly.

What I mean is that I probably lack the maths background to review the math computation. For example if it turns out to be uselessly complex and there is a simpler or faster formula that gives the same result, don't expect me to tell you that.

But as soon as the maths produces the expected results, it's fine to me, so I only looked at the code shape.

slipher:
So what is the connection between this and https://github.com/Unvanquished/Unvanquished/issues/2913? Does it change the API contract between gamelogic and engine? If so, things that change the API contract should normally be merged to the branch for the upcoming release (now `for-0.55.0/sync`), rather than `master`.

illwieckz:
I said to @VReaperV  that as long as engine from master can run released game, or that game from master can run on released engine, it's fine to target master, even if we can't build game master without engine master (thing we fix by committing submodules).

VReaperV:
> So what is the connection between this and [Unvanquished/Unvanquished#2913](https://github.com/Unvanquished/Unvanquished/issues/2913)? Does it change the API contract between gamelogic and engine? If so, things that change the API contract should normally be merged to the branch for the upcoming release (now `for-0.55.0/sync`), rather than `master`.

No, the old engine code was just expecting different information from cgame. Right now if you run the engine with this change but without the corresponding cgame+sgame change you will just get incorrect rotations for portals. There's no single way to describe what that information was other than "incorrect".

slipher:
"expecting different information" is what I would call changing the API contract

VReaperV:
I was going by what illwieckz said.

slipher:
> I was going by what illwieckz said.

I assume you mean:

> I said to @VReaperV that as long as engine from master can run released game, or that game from master can run on released engine, it's fine to target master, even if we can't build game master without engine master (thing we fix by committing submodules).

What illwieckz said was true, but not relevant to the question of ABI compatibility that I'm talking about. We want an engine built from master to continue supporting gamelogic binaries from the most recently released version of the game. Also, gamelogic built from master should work with the most recently released engine binary. See https://wiki.unvanquished.net/wiki/Compatibility.

VReaperV:
Rather something that was said on development discord/irc a few days ago:
"we don't care about breaking master compilation with old master submodules, we only care about built engine running built game within a 0.x release cycle
[10:01 PM]
what we do is we commit submodules frequently"
"it's ok if after merging your PR to master, the game repository from today requires the engine repository from today to build, as long as that game build runs on the 0.54.0 engine we built 1 year ago"
"game build runs" I took as meaning it doesn't introduce any crashes etc.

slipher:
Seems to be saying the same thing as the other comment. "The game runs" should be taken to mean that it works correctly (at least to the extent that it did before), not that it merely doesn't crash.

VReaperV:
> I don't like the comments being on the same line than code, I prefer if such common is on the immediately preceding line, like this:

Fixed.

illwieckz:
> "The game runs" should be taken to mean that it works correctly

The game was not working correctly before, so one may say we never seen it running…

As far as I know it was so broken @cu-kai had to do intentional wrong things in his maps to somewhat cancel the wrong things the code does. Also _before_, the released code is crashing: https://github.com/DaemonEngine/Daemon/issues/1002

I have hard time to see what this this code is breaking that is not more broken before.

slipher:
> I've checked old maps with portals that I could find with this change and they work correctly, except pitfall, gauntlet and pulse, which is cause by map bugs (old code and even tremulous were not getting the portal view origin correctly and both were ignoring the roll keyword).

It is not true that the `roll` keyword is ignored. I tested on `map-pulse_1.3.dpk` with a cvar like this:

```
diff --git a/src/cgame/cg_ents.cpp b/src/cgame/cg_ents.cpp
index fed201a91..ce72b32b5 100644
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -618,6 +618,7 @@ static void CG_Mover( centity_t *cent )
 CG_Portal
 ===============
 */
+Cvar::Cvar<float> roll("roll", "", 0, -1);
 static void CG_Portal( centity_t *cent )
 {
        entityState_t *s1;
@@ -640,6 +641,8 @@ static void CG_Portal( centity_t *cent )
        ent.oldframe = s1->misc;
        ent.frame = s1->frame; // rotation speed
        ent.skinNum = s1->clientNum / 256.0 * 360; // roll offset
+       if (*roll >= 0)
+               ent.skinNum = *roll;

        // add to refresh list
        trap_R_AddRefEntityToScene( &ent );
```

Changing this cvar does cause the view to rotate in some way. The portal pointing at the alien base has a roll of 180 and it looks different than if it had no roll.

VReaperV:
Well, I was doing /setviewpos at cameras origin/angle/target as defined in the bsp file, and what I was seeing corresponded to not having any roll. Also, I suggest looking at one of the test maps: https://users.unvanquished.net/~kai/map-test-portals_0.9.dpk, to compare how it worked before and with this change, you can see that old code was very much broken there. You get different views just based on how the portals were set up compared to worlds axes, which is awful behaviour. Maybe roll in those maps was a workaround to that then.

Regardless, there's no reason to keep that old hacky code that breaks because you rotated the portal surface, when instead we can just use the actual angles/target as defined in the map. The few old maps that need some portals adjusted because of that (there aren't that many maps with portals to begin with) I've already uploaded the updated versions of. If there's some maps that I missed fixing those would take very little time.

slipher:
> Well, I was doing /setviewpos at cameras origin/angle/target as defined in the bsp file, and what I was seeing corresponded to not having any roll. 

I guess the 180 in Pulse is a special case since it could also be done by only using pitch. Without the 180 roll, the alien base camera looks like the POV of someone wallwalking on the ceiling.


> The few old maps that need some portals adjusted because of that (there aren't that many maps with portals to begin with) I've already uploaded the updated versions of. If there's some maps that I missed fixing those would take very little time.

There seems to something of a consensus against this approach, preferring to preserve complete backwards compatibility with old assets. [illwieckz gives a rationale.](https://github.com/Unvanquished/Unvanquished/issues/1999#issuecomment-1199964113)



> Also, I suggest looking at one of the test maps: https://users.unvanquished.net/~kai/map-test-portals_0.9.dpk, to compare how it worked before and with this change, you can see that old code was very much broken there. You get different views just based on how the portals were set up relative to worlds axes, which is awful behaviour.

I see in this map all the entities with a "roll" keyword are of class gfx_portal_surface which is incorrect. It should be on gfx_portal_camera. I guess that's why you thought the parsing doesn't work.

Anyway I think I see what you mean, comparing the portal surfaces at (-64 2176 32) and (-64 2176 224). They use the same camera but one shows the skybox and the other a view with a turret. I agree it seems unexpected that they do not look at the same thing.

slipher:
To enable creation of portals with more sensible semantics without breaking existing maps, one might consider adding a new classname of portal entities with different behavior, or adding new spawnflags or keywords to the existing types.

VReaperV:
> To enable creation of portals with more sensible semantics without breaking existing maps, one might consider adding a new classname of portal entities with different behavior, or adding new spawnflags or keywords to the existing types.

`gfx_portal_*` can be used for that perhaps? There would probably be a bunch of old code alongside new one with that approach though.

illwieckz:
To report some things that were discussed in the chat some days ago:

It is believed the `roll` key is meant to set the roll value (third component of `angles`), as a way to set the roll without setting the `angles` key at all.

The reason why there is a need to set the roll without setting the `angles` key at all is that there is two ways to set angles for a camera:

1. Set the `angles` key in the map: `yaw pitch roll` is then hardcoded;
2. Set a position entity as a target of the camera entity,
the game will compute the angles to always face the position entity.

The second method can't set roll, hence the need for a way to still set the roll if using the linked-entity method to set the angles, but one cannot set the roll with the `angles` key without setting `yaw` and `pitch` as `roll` is the third component.

This is the reason why we believe the standalone `roll` exists to begin with.

VReaperV:
We could ignore the pitch and yaw values if there's a target, but we already have maps using "roll".

illwieckz:
The [Quake 3 entity definition](https://gitlab.com/netradiant/gamepacks/quake3-mapeditor-support/-/blob/main/q3.game/baseq3/entities.ent?ref_type=heads#L1485) for `misc_portal_camera` says:

> Both the setting "angle(s)" key or "targeting a target_position" methods can be used to aim the camera. In both cases, the "roll" key must be used to set the roll angle. If either the SLOWROTATE or FASTROTATE spawnflag is set, then the "roll" value is irrelevant.

And:

> angle: Yaw aiming angle of the portal camera (default 0).
> angles: This sets the pitch and yaw aiming angles of the portal camera (default 0 0). Use "roll" key to set roll angle.
> target: Point this to a target_position entity to set the camera's pointing direction.
> targetname: A misc_portal_surface portal surface indicator must point to this.
> roll: Roll angle of camera. For axial directions a value of 0 is upside down and 180 is the same as the player's view. For other sorts of directions values of 90 and -90 may work (inconsistent math).

The [Quake Live entity definition](https://gitlab.com/netradiant/gamepacks/quakelive-mapeditor-support/-/blob/main/quakelive.game/baseq3/entities.def?ref_type=heads#L2840) for `misc_portal_camera` says:

> Portal camera. This camera is used to project its view onto a portal surface in the level through the intermediary of a misc_portal_surface entity. Use the "angles" key or target a target_position or info_notnull entity to set the camera's pointing direction.

> angles: This sets the pitch and yaw aiming angles of the portal camera (default 0 0). Use "roll" key to set roll angle.
target : Point this to a target_position entity to set the camera's pointing direction.
targetname : A misc_portal_surface portal surface indicator must point to this.
roll : Roll angle of camera. A value of 0 is upside down and 180 is the same as the player's view.

This statement is very weird and is there in both files:

> A value of 0 is upside down and 180 is the same as the player's view.

VReaperV:
That stems from the wrong old code I suppose. I think that's what we're seeing with e. g. flipped portals on test maps.

illwieckz:
The original Q3 `entities.def` for GtkRadiant [repository here, but a bug prevents to read it online](http://svn.icculus.org/gtkradiant-gamepacks/Q3Pack/trunk/install/baseq3/scripts/entities.def?view=log) said on `2006-02-09 01:07:34` (commit by SPoG):

> /*QUAKED misc_portal_camera (1 .5 .25) (-8 -8 -8) (8 8 8) SLOWROTATE FASTROTATE

> Portal camera. This camera is used to project its view onto a portal surface in the level through the intermediary of a misc_portal_surface entity. Use the "angles" key or target a target_position or info_notnull entity to set the camera's pointing direction.

> -------- KEYS --------
> angles: this sets the pitch and yaw aiming angles of the portal camera (default 0 0). Use "roll" key to set roll angle.
> target : point this to a target_position entity to set the camera's pointing direction.
> targetname : a misc_portal_surface portal surface indicator must point to this.
> roll: roll angle of camera. A value of 0 is upside down and 180 is the same as the player's view.

illwieckz:
> That stems from the wrong old code I suppose. I think that's what we're seeing with e. g. flipped portals on test maps.

Yeah, it's possible that's something deduced from the observed behavior, not from what it should be.

slipher:
> > To enable creation of portals with more sensible semantics without breaking existing maps, one might consider adding a new classname of portal entities with different behavior, or adding new spawnflags or keywords to the existing types.
> 
> `gfx_portal_*` can be used for that perhaps? Although considering they're supposed to be interchangeable, a keyword or spawnflag might be better.

Yeah that seems fine. There's a [proposal](https://github.com/Unvanquished/Unvanquished/issues/2537) to get rid of all these duplicate `gfx_`xxx entity classes that don't actually function differently from the original ones.

Hopefully we can end with a sensible engine API, a new portal entity scheme with sensible functionality, and legacy portal entities that can translate the old behavior to the current engine API.




> This statement is very weird and is there in both files:
> 
> > A value of 0 is upside down and 180 is the same as the player's view.

On Pulse one camera has this behavior and one camera does not. Maybe this is caused by the use of the `PerpendicularVector` function in CG_Portal. This is a function that has to arbitrarily pick a vector that is perpendicular to one other vector and does not have a mathematically consistent behavior.

VReaperV:
I can also make it use spawnflags instead, e. g. bit 4 would correspond to the new behaviour, whichever approach is more desirable.
> Maybe this is caused by the use of the PerpendicularVector function in CG_Portal.
If it is because of PerpendicularVector, then the compatibility with old maps can even all be handled in sgame, engine would be agnostic and would use the actual angles values passed like one would expect.

src/engine/renderer/tr_main.cpp
```diff
@@ -1340,45 +1325,101 @@ static bool R_GetPortalOrientations( drawSurf_t *drawSurf, orientation_t *surfac
...
-			}
-			else
-			{
+				VectorCopy( e->e.axis[ 1 ], transformed);
+				RotatePointAroundVector( e->e.axis[ 1 ], e->e.axis[ 0 ], transformed, d );
+				CrossProduct( e->e.axis[ 0 ], e->e.axis[ 1 ], e->e.axis[ 2 ] );
+			} else {
 				// bobbing rotate, with skinNum being the rotation offset
 				d = sinf( tr.refdef.time * 0.003f );
 				d = e->e.skinNum + d * 4;
```
slipher, line 1351: This code still refers to skinNum, but the gamelogic-side PR removes the code from CG_Portal that sets that?

VReaperV: Fixed on sgame side.

slipher: You mean you removed the skinNum usage from the renderer?

Having nuked the rotation offset option, it should be removed from the comment as well.
`// bobbing rotate, with skinNum being the rotation offset`

VReaperV: > You mean you removed the skinNum usage from the renderer?

Yes.

> Having nuked the rotation offset option, it should be removed from the comment as well.

Done.

slipher:
LGTM

illwieckz (approved):
LGTM.

PR #1038, opened 2024-01-27
Fix lightmaps for portal surfaces

VReaperV:
Changes the portal surfaces to render as a BSP surface in Tess_Begin() in FinalisePortalCommand. Sets the lightmapNum and fogNum, although this isn't necessary since &lightmap always uses worlds lightmap (-1).

Note: this does not fix dynamic lights for portals.

Screenshots:
Without lightmap:
unvanquished_2024-01-27_173416_000
With lightmap:
unvanquished_2024-01-27_173206_000
Without lightmap:
unvanquished_2024-01-27_173438_000
With lightmap:
unvanquished_2024-01-27_173241_000


slipher:
LGTM

PR #1039, opened 2024-02-01
Unify list command and function naming

illwieckz:
This is some lazy task I did to change my mind and this put an end to something that annoyed me for a long time. This unifies our engine `list` commands and functions.

I was worrided we had `listCvars` and `listCmd` but `imagelist` and `shaderlist`, so we no also have `listImages` and `listShaders`, which also greatly improves command completion when looking for “_what do we can list_”? This also totally removes the extra useless mental load of thinking about “_is the name of the command I need starting or ending with `list`_”?

I also unified the related functions.

For example we had `R_InitVBOs()`, `R_ShutdownVBOs()`, `R_InitFBOs()`, and `R_ShutdownFBOs()` but `R_VBOList_f()` and `R_FBOList_f()`, we now have `R_ListVBOs_f()` and `R_ListFBOs_f()` which is more consistent with other function names.

Same for `BindListCmd()` which is now named `ListBindsCmd()` which is more consistent with existing `BindCmd()` and `EditBindCmd`, basically following the same `verbObject` naming scheme.

An extra commit deletes a duplicate `R_ShutdownVBOs()` declaration.

### DIFF
--- a/src/engine/client/key_binding.cpp
+++ b/src/engine/client/key_binding.cpp
@@ -456,11 +456,11 @@ std::string ExtraInfoKeyString(Key key)
 	return KeyToString(key);
 }
 
-class BindListCmd: public Cmd::StaticCmd
+class ListBindsCmd: public Cmd::StaticCmd
 {
 public:
-	BindListCmd():
-		StaticCmd("bindlist", Cmd::KEY_BINDING, "Lists all key bindings")
+	ListBindsCmd():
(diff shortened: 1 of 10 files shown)


illwieckz:
About unifying, I wonder if we can rename `listCvars` to `listCVars`. This is pure bikeshedding but if we do it, it's time to.

slipher:
LGTM

Issue #103, opened 2018-04-29
random segfault on map change


illwieckz:
we still suffer from random crashes on map change, I got it logged on gdb:

```gdb
Thread 31 "daemon" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff95140700 (LWP 29722)]
0x00007fff9d89fb25 in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.3.so.0

Thread 31 (Thread 0x7fff95140700 (LWP 29722)):
#0  0x00007fff9d89fb25 in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.3.so.0
#1  0x00007fff6a454095 in Rocket::Core::Lua::LuaEventListener::ProcessEvent (this=0x7fff6ce62020, event=...)
    at ./Unvanquished/libs/libRocket/Source/Core/Lua/LuaEventListener.cpp:128
#2  0x00007fff6a35de88 in Rocket::Core::EventDispatcher::TriggerEvents (this=0x7fff6cee2e70, event=0x7fff6f818390)
    at ./Unvanquished/libs/libRocket/Source/Core/EventDispatcher.cpp:182
#3  0x00007fff6a35dc2f in Rocket::Core::EventDispatcher::DispatchEvent (this=0x7fff6cee2e70, target_element=0x7fff6cd0ffa0, name=..., parameters=..., 
    interruptible=false) at ./Unvanquished/libs/libRocket/Source/Core/EventDispatcher.cpp:146
#4  0x00007fff6a329115 in Rocket::Core::Element::DispatchEvent (this=0x7fff6cd0ffa0, event=..., parameters=..., interruptible=false)
    at ./Unvanquished/libs/libRocket/Source/Core/Element.cpp:1053
#5  0x00007fff6a31b7ea in Rocket::Core::RKTEventFunctor::operator() (this=0x7fff9513ea40, element=...)
    at ./Unvanquished/libs/libRocket/Source/Core/EventIterators.h:55
#6  0x00007fff6a31cbbe in std::for_each<__gnu_cxx::__normal_iterator<Rocket::Core::ElementReference*, std::vector<Rocket::Core::ElementReference, std::allocator<Rocket::Core::ElementReference> > >, Rocket::Core::RKTEventFunctor> (__first={element = 0x7fff6cd0ffa0}, __last={element = 0x7fff6fddf3f0}, __f=...)
[...]

PR #1040, opened 2024-02-08
renderer: implement ivec_t and the ability to send integer vectors to GLSL

illwieckz:
Right now we can only send a single integer to GLSL, or float vectors. Meaning that if we need to sends more than one integer, we either have to send multiple variables, or to send integers in floats vectors.

This implements the `ivec_t` type in engine and implements the related classes to send `ivec2`, `ivec3` and `ivec4` to GLSL in a similar way to what we already do for `vec2`, `vec3` and `vec4`.

### DIFF
--- a/src/engine/qcommon/q_shared.h
+++ b/src/engine/qcommon/q_shared.h
@@ -227,7 +227,6 @@ void  Com_Free_Aligned( void *ptr );
 
 	using vec_t = float;
 	using vec2_t = vec_t[2];
-
 	using vec3_t = vec_t[3];
 	using vec4_t = vec_t[4];
 
@@ -235,6 +234,11 @@ void  Com_Free_Aligned( void *ptr );
 	using matrix3x3_t = vec_t[3 * 3];
 	using matrix_t = vec_t[4 * 4];
 	using quat_t = vec_t[4];
+
+	using ivec_t = int;
+	using ivec2_t = ivec_t[2];
+	using ivec3_t = ivec_t[3];
+	using ivec4_t = ivec_t[4];
 
 	// A transform_t represents a product of basic
 	// transformations, which are a rotation about an arbitrary
--- a/src/engine/renderer/gl_shader.h
+++ b/src/engine/renderer/gl_shader.h
@@ -418,6 +418,127 @@ public:
 	}
 };
 
+class GLUniform2i : protected GLUniform
+{
+protected:
+	GLUniform2i( GLShader *shader, const char *name ) :
+	GLUniform( shader, name )
+	{
+	}
+
+	inline void SetValue( const ivec2_t v )
+	{
+		shaderProgram_t *p = _shader->GetProgram();
+
+		ASSERT_EQ(p, glState.currentProgram);
+
+#if defined( LOG_GLSL_UNIFORMS )
+		if ( r_logFile->integer )
+		{
+			GLimp_LogComment( va( "GLSL_SetUniform2i( %s, shader: %s, value: [ %d, %d ] ) ---\n",
+				this->GetName(), _shader->GetName().c_str(), v[ 0 ], v[ 1 ] ) );
+		}
+#endif
+#if defined( USE_UNIFORM_FIREWALL )
+		ivec2_t *firewall = ( ivec2_t * ) &p->uniformFirewall[ _firewallIndex ];
+
+		if ( !memcmp( *firewall, v, sizeof( *firewall ) ) )
+		{
+			return;
+		}
+
+		( *firewall )[ 0 ] = v[ 0 ];
+		( *firewall )[ 1 ] = v[ 1 ];
+#endif
+		glUniform2i( p->uniformLocations[ _locationIndex ], v[ 0 ], v[ 1 ] );
+	}
+
+	size_t GetSize() override
+	{
+		return sizeof( ivec2_t );
+	}
+};
+
+class GLUniform3i : protected GLUniform
+{
+protected:
+	GLUniform3i( GLShader *shader, const char *name ) :
+	GLUniform( shader, name )
+	{
+	}
+
+	inline void SetValue( const ivec3_t v )
+	{
+		shaderProgram_t *p = _shader->GetProgram();
+
+		ASSERT_EQ(p, glState.currentProgram);
+
+#if defined( LOG_GLSL_UNIFORMS )
+		if ( r_logFile->integer )
+		{
+			GLimp_LogComment( va( "GLSL_SetUniform3i( %s, shader: %s, value: [ %d, %d, %d ] ) ---\n",
+				this->GetName(), _shader->GetName().c_str(), v[ 0 ], v[ 1 ], v[ 2 ] ) );
+		}
+#endif
+#if defined( USE_UNIFORM_FIREWALL )
+		ivec3_t *firewall = ( ivec3_t * ) &p->uniformFirewall[ _firewallIndex ];
+
+		if ( !memcmp( *firewall, v, sizeof( *firewall ) ) )
+		{
+			return;
+		}
+
+		VectorCopy( v, *firewall );
+#endif
+		glUniform3i( p->uniformLocations[ _locationIndex ], v[ 0 ], v[ 1 ], v[ 2 ] );
+	}
+public:
+	size_t GetSize() override
+	{
+		return sizeof( ivec3_t );
+	}
+};
+
+class GLUniform4i : protected GLUniform
+{
+protected:
+	GLUniform4i( GLShader *shader, const char *name ) :
+	GLUniform( shader, name )
+	{
+	}
+
+	inline void SetValue( const ivec4_t v )
+	{
+		shaderProgram_t *p = _shader->GetProgram();
+
+		ASSERT_EQ(p, glState.currentProgram);
+
+#if defined( LOG_GLSL_UNIFORMS )
+		if ( r_logFile->integer )
+		{
+			GLimp_LogComment( va( "GLSL_SetUniform4i( %s, shader: %s, value: [ %d, %d, %d, %d ] ) ---\n",
+				this->GetName(), _shader->GetName().c_str(), v[ 0 ], v[ 1 ], v[ 2 ], v[ 3 ] ) );
+		}
+#endif
+#if defined( USE_UNIFORM_FIREWALL )
+		ivec4_t *firewall = ( ivec4_t * ) &p->uniformFirewall[ _firewallIndex ];
+
+		if ( !memcmp( *firewall, v, sizeof( *firewall ) ) )
+		{
+			return;
+		}
+
+		Vector4Copy( v, *firewall );
+#endif
+		glUniform4i( p->uniformLocations[ _locationIndex ], v[ 0 ], v[ 1 ], v[ 2 ], v[ 3 ] );
+	}
+public:
+	size_t GetSize() override
+	{
+		return sizeof( ivec4_t );
+	}
+};
+
 class GLUniform1f : protected GLUniform
 {
 protected:


illwieckz:
Right now this code is not used, but I stumbled upon this need years ago, and I may use it in the future.

My linear light computation code in #1034 currently makes use of it:

- https://github.com/DaemonEngine/Daemon/pull/103

Though, since that code uses a `ivec2` to send two booleans, I may implement a hacky flag system to only send a single integer (GLSL doesn't suport bitwise operations but I only have to track two bits, so the values `0`, `1, `2` and `3`.

In all cases, as I said I found it very annoying that we can only send a single integer and not vector integer, so having this implemented makes the engine easier to deal with when implementing stuff.

slipher:
We should add this if/when it is actually used by something.

That said, please don't add more stuff to `q_shared.h`. The definitions in question are only relevant to renderer so it could go in e.g. `tr_local.h`

illwieckz:
> We should add this if/when it is actually used by something.

I usually agree with this, but here, definitely not.

Those are generic types, it's a library. I faced the need of this more than once in the last years, and I always had to rely on alternative solutions. This is even needed when not merging code that use it, because this is needed to prototype implementations and evaluate solutions. That's why I gave the example that maybe some legit usage of `ivec2` may be one day avoided by using hacky optimizations and hacky flag emulation, but this never removes the need for `ivec2` for the canonical implementation. Even if no one code is merged using this, we need this library code to prototype solutions. I always have to rely on float vectors when I write temporary code to evaluate something, even if each time I did that in the past, after multiple steps of optimizations and collapsing I ended up removing the need for an ivec. I'm fed-up of using floats as booleans when implementing exhaustive implementations prior to simplification, or to copypaste 3 or 4 times all the boiler plate code in 3 or 4 files for every integer, while an `ivec` allows to only do the boilerplate copypaste once in 3 or 4 files.

> That said, please don't add more stuff to `q_shared.h`. The definitions in question are only relevant to renderer so it could go in e.g. `tr_local.h`

This is doable, though this is a very basic generic type system, and the float vectors are defined in `q_shared.h`. In the future we may replace all the `VectorCopy` and other macros with type-safe functions, and then we would need to define all those types in the most global way. Implementing those types and replacing those macros with type-safe functions would even help a migration to something else, one day (like GLM).

This work is not renderer-specific, it's engine architecture and fundamental type implementation. It just happens that I need this in renderer first because I'm focusing on the renderer right now.

We did similar work for float vector types when I worked on the model code. This PR is anothere step on that same road. Last time for models, today for some rendering techniques, but it's generic ground work for the whole engine, not just the renderer.

slipher:
If there are many cases in previously written code where it would have been useful, you should be able to find one of those and migrate it? I'm not saying you need to use every length of vector, but at least *one* of GLUniform2i/GLUniform3i/GLUniform4i/.

Anyway `using foo = bar[3];` typedefs are hardly something we want to encourage greater usage of in the future as they are type-unsafe.
```
void f(vec3_t v);
void g() {
   vec2_t v{};
   f(v);
}
```
The above code compiles without protest because 'f''s declaration is equivalent to `void f(float* v);`

Furthermore if you're going to have a general-purpose integer vector typedef, it should at least be decent enough what size of integer it is using (like `i16vec4_t` or `u16vec2_t` defined in tr_local.h). The generic `ivec` naming only arguably makes sense in the specific context of the OpenGL uniform APIs which only have one integer type. 

You might consider going without a typedef at all and doing it like `SetValue(a, b, c)`.

illwieckz:
> Anyway `using foo = bar[3];` typedefs are hardly something we want to encourage greater usage of in the future as they are type-unsafe.
> 
> ```
> void f(vec3_t v);
> void g() {
>    vec2_t v{};
>    f(v);
> }
> ```
> 
> The above code compiles without protest because 'f''s declaration is equivalent to `void f(float* v);`
> 
> Furthermore if you're going to have a general-purpose integer vector typedef, it should at least be decent enough what size of integer it is using (like `i16vec4_t` or `u16vec2_t` defined in tr_local.h). The generic `ivec` naming only arguably makes sense in the specific context of the OpenGL uniform APIs which only have one integer type.

Right, immediately after having posted my comment I remember that what we had converted before was not `vec2_t` but `f16vec2_t`, so copypasting `vec2_t` for `ivec2_t` could not make it type-safe as our `vec2_t` implementation is not.

So right now I looked how to implement a `vec_t` the same way we did with `f16vec2_t`, I started with this:

```c++
	struct vec_t {
		operator float() const { return bits; };

		float bits;
	};

	using vec2_t = vec_t[2];
	using vec3_t = vec_t[3];
	using vec4_t = vec_t[4];
```

But then I struggle to implement the `float()` cast for `vec_t` castn so we can do `vec_t v; v = 1.0`;

In the end I would like to replace the `Vector2Copy`/`VectorCopy`/`Vector4Copy` macros with something like:

```c++
inline void VectorCopy( const vec2_t in, vec2_t out )
{
	out[ 0 ] = in[ 0 ];
	out[ 1 ] = in[ 1 ];
}

inline void VectorCopy( const vec3_t in, vec3_t out )
{
	out[ 0 ] = in[ 0 ];
	out[ 1 ] = in[ 1 ];
	out[ 2 ] = in[ 2 ];
}

inline void VectorCopy( const vec4_t in, vec4_t out )
{
	out[ 0 ] = in[ 0 ];
	out[ 1 ] = in[ 1 ];
	out[ 2 ] = in[ 2 ];
	out[ 3 ] = in[ 3 ];
}

#define Vector2Copy VectorCopy
#define Vector4Copy VectorCopy
```

And the same with all the other vector macro (subtraction, addition…).

illwieckz:
(I updated my previous comment as some wording was saying the contrary of what I meant).

slipher:
If you must use an array-like type for some reason you can use `std::vector<int32_t, n>`

illwieckz:
> If there are many cases in previously written code where it would have been useful, you should be able to find one of those and migrate it?

Everytime I faced the need myself I went up with other implementations anyway, and as I said, most of the time the ivec disappeared after multiple iteration in development, but this is luck. For example by implementing a compute in GLSL for every pixel, one may read the canonical documentation for the feature to implement and need an ivec3 for three numbers. This is the first iteration. This is a direct translation from a generic math algorithm to GLSL. Then later it is discovered the result of the computation using the ivec3 can be precomputed once for all in the C++ code, instead of doing it it per pixel. So in the next development iteration, the compute on the ivec3 is done in some C++ file and a single integer for the result is sent to be used by GLSL.

The removal of the need for sending the ivec3 then becomes an optimization. It is not guaranteed that any ivec3 computation we may want to do will be exportable to C++ instead in a way only a single integer result is needed. This is just luck if it happened before.

And in all cases, this is very annoying to have to think about the optimization at the same time the naive implementation is done. I want to be able to do naive implementation first, to validate they work, then to think about optimization to save the amount of variable we pass to GLSL and to reduce the GLSL code.

Many work we merged without using any ivec2/ivec3 were optimized code from which the naive implementation in the first iteration were using ivec2/ivec3, so I had to rely on float vectors as booleans or to copy paste many single-integer variable sending code, just to deleted all of that before submitting the PR when everything was optimized after 3 or 4 optimization passes.

One good example I have in mind is some existing code using some `USE_` macro to select this or that compute, my first iterations for them were using many booleans in early stage of development to focus on the actual verification of the expected result, rather than looking for performance. I postponed the implementation of macro exclusion and all the related mental load for _later_, for _after_ the computation was verified to be correct. In those intermediary steps I had to rely on boolean as floats, or copy-paster many useless lines.

illwieckz:
> If you must use an array-like type for some reason you can use `std::array<int32_t, n>`

Ah, I forgot about `std_array`. Is there a reason why we did not used it for `f16vec2_t`?

I'm trying to convert `vec_t` first, then my `ivec_t` implementation would do the same once I'm sure I have an implementation that works for all our `vec_t` use case. My idea is to _not_ implement yet another vector library (that would require to port all our existing code) but to implement a type-safe drop-in implementation that requires to rewrite _none_ of the existing code, by providing the exact same syntax, but in a type-safe way.

I did that:

```c++
using vec_t = float;
using vec2_t = std::array(vec_t, 2);
using vec3_t = std::array(vec_t, 3);
using vec4_t = std::array(vec_t, 4);
```

I'm now getting this:

```
In file included from Unvanquished/daemon/src/common/Common.h:35:
Unvanquished/daemon/src/engine/qcommon/q_shared.h: In function ‘__m128 sseLoadVec3(vec3_t)’:
Unvanquished/daemon/src/engine/qcommon/q_shared.h:1122:38: error: invalid cast from type ‘const vec3_t’ {aka ‘const std::array<float, 3>’} to type ‘__m64*’
 1122 |                 v = _mm_loadl_pi( v, (__m64 *)vec );
      |                                      ^~~~~~~~~~~~
Unvanquished/daemon/src/engine/qcommon/q_shared.h: In function ‘__m128 sseLoadVec3Unsafe(vec3_t)’:
Unvanquished/daemon/src/engine/qcommon/q_shared.h:1127:38: error: cannot convert ‘const vec3_t’ {aka ‘const std::array<float, 3>’} to ‘const float*’
 1127 |                 return _mm_loadu_ps( vec );
      |                                      ^~~
      |                                      |
      |                                      const vec3_t {aka const std::array<float, 3>}
In file included from Unvanquished/daemon/src/common/Platform.h:68,
                 from Unvanquished/daemon/src/engine/qcommon/q_shared.h:156:
/usr/lib/gcc/x86_64-linux-gnu/13/include/xmmintrin.h:938:28: note:   initializing argument 1 of ‘__m128 _mm_loadu_ps(const float*)’
  938 | _mm_loadu_ps (float const *__P)
      |               ~~~~~~~~~~~~~^~~
Unvanquished/daemon/src/engine/qcommon/q_shared.h: In function ‘void sseStoreVec3(__m128, vec3_t)’:
Unvanquished/daemon/src/engine/qcommon/q_shared.h:1130:32: error: invalid cast from type ‘vec3_t’ {aka ‘std::array<float, 3>’} to type ‘__m64*’
 1130 |                 _mm_storel_pi( (__m64 *)out, in );
      |   
```

Any idea to solve those cast issues?

illwieckz:
With help from @DolceTriade on the chat, I did this:

```diff
diff --git a/src/engine/qcommon/q_shared.h b/src/engine/qcommon/q_shared.h
index e27ab1832..a191b3398 100644
--- a/src/engine/qcommon/q_shared.h
+++ b/src/engine/qcommon/q_shared.h
@@ -226,10 +226,9 @@ void  Com_Free_Aligned( void *ptr );
 	*/
 
 	using vec_t = float;
-	using vec2_t = vec_t[2];
-
-	using vec3_t = vec_t[3];
-	using vec4_t = vec_t[4];
+	using vec2_t = std::array<vec_t, 2>;
+	using vec3_t = std::array<vec_t, 3>;
+	using vec4_t = std::array<vec_t, 4>;
 
 	using axis_t = vec3_t[3];
 	using matrix3x3_t = vec_t[3 * 3];
@@ -1120,15 +1119,15 @@ inline float DotProduct( const vec3_t x, const vec3_t y )
 	inline __m128 sseLoadVec3( const vec3_t vec ) {
 		__m128 v = _mm_load_ss( &vec[ 2 ] );
 		v = sseSwizzle( v, YYXY );
-		v = _mm_loadl_pi( v, (__m64 *)vec );
+		v = _mm_loadl_pi( v, (__m64 *) &vec[0] );
 		return v;
 	}
 	ATTRIBUTE_NO_SANITIZE_ADDRESS inline __m128 sseLoadVec3Unsafe( const vec3_t vec ) {
 		// Returns garbage in 4th element
-		return _mm_loadu_ps( vec );
+		return _mm_loadu_ps( &vec[0] );
 	}
 	inline void sseStoreVec3( __m128 in, vec3_t out ) {
-		_mm_storel_pi( (__m64 *)out, in );
+		_mm_storel_pi( (__m64 *) &out[0], in );
 		__m128 v = sseSwizzle( in, ZZZZ );
 		_mm_store_ss( &out[ 2 ], v );
 	}
```

This solved the first errors, but then I face many other errors, I may need to port many stray `float*` to `vec3_t`/`vec4_t`, which is find, but then I have to descend to hell to save the poor souls, like that:

```diff
 bool CM_PlaneFromPoints( vec4_t plane, const vec3_t a, const vec3_t b, const vec3_t c )
 {
-       vec3_t d1, d2;
+       vec3_t d1, d2, d3;
 
        VectorSubtract( b, a, d1 );
        VectorSubtract( c, a, d2 );
-       CrossProduct( d2, d1, plane );
 
-       if ( VectorNormalize( plane ) == 0 )
+       CrossProduct( d2, d1, d3 );
+
+       if ( VectorNormalize( d3 ) == 0 )
        {
                return false;
        }
 
-       plane[ 3 ] = DotProduct( a, plane );
+       plane[ 0 ] = d3[ 0 ];
+       plane[ 1 ] = d3[ 1 ];
+       plane[ 2 ] = d3[ 2 ];
+       plane[ 3 ] = DotProduct( a, d3 );
        return true;
 }
```

It looks like implementing a type-safe `vec_t` implementation would require to port the engine to `vec_t` first. 🤯️

PR #1043, opened 2024-02-13
introduce `plane_t` struct with `normal` and `dist` members

illwieckz:
Previously the plane was stored in a `vec4_t`,
with the three first components being the normal,
and the last one being the distance.

A lot of code was running the `vec3_t` specific
macros on the `vec4_t`, it only worked because
we use unsafe types.

This change is a step on the road of making safe
types like `vec2_t`, `vec3_t`, `vec4_t` and turne
macros like `VectorCopy` into type-safe functions.

This also makes the code more explicit.

Instead of doing:

```c++
    vec3_t normal = { 0, 0, 0 };
    vec4_t plane;
    VectorCopy( normal, plane );
```

We can now do:

```c++
    vec3_t in = { 0, 0, 0 };
    plane_t plane;
    VectorCopy( normal, plane.normal );
```

The day `VectorCopy` is a type-safe function
instead of a macro, it will work.

There already exist a `cplane_t` struct
that carries `normal` and `dist` but also
more information, we may modify it to
use `plane_t` for `normal` and `dist`, this
would allow us to do:

```c++
    PlaneCopy( cplane.plane, plane )
```

instead of:

```c++
    VectorCopy( cplane.normal, plane.normal );
    plane.dist = cplane.dist;
```

But for now this is out of topic.

### DIFF
--- a/src/engine/qcommon/q_math.cpp
+++ b/src/engine/qcommon/q_math.cpp
@@ -225,23 +225,32 @@ void ByteToDir( int b, vec3_t dir )
 	VectorCopy( bytedirs[ b ], dir );
 }
 
-vec_t PlaneNormalize( vec4_t plane )
+void PlaneSet( plane_t &out, const vec_t x, const vec_t y, const vec_t z, const vec_t dist )
 {
-	vec_t length2, ilength;
+	VectorSet( out.normal, x, y, z );
+	out.dist = dist;
+}
 
-	length2 = DotProduct( plane, plane );
+void PlaneSet( plane_t &out, const vec4_t v )
+{
+	PlaneSet( out, v[ 0 ], v[ 1 ], v[ 2 ], v[ 3 ] );
+}
+
+vec_t PlaneNormalize( plane_t &plane )
+{
+	vec_t length2 = DotProduct( plane.normal, plane.normal );
 
 	if ( length2 == 0.0f )
 	{
-		VectorClear( plane );
+		VectorClear( plane.normal );
 		return 0.0f;
 	}
 
-	ilength = Q_rsqrt( length2 );
-	plane[ 0 ] = plane[ 0 ] * ilength;
-	plane[ 1 ] = plane[ 1 ] * ilength;
-	plane[ 2 ] = plane[ 2 ] * ilength;
-	plane[ 3 ] = plane[ 3 ] * ilength;
+	vec_t ilength = Q_rsqrt( length2 );
+	plane.normal[ 0 ] *= ilength;
+	plane.normal[ 1 ] *= ilength;
+	plane.normal[ 2 ] *= ilength;
+	plane.dist *= ilength;
 
 	return length2 * ilength;
 }
@@ -270,20 +279,20 @@ vec_t PlaneNormalize( vec4_t plane )
  * See the function below if you want the normal to be on the other side
  * =====================
  */
-bool PlaneFromPoints( vec4_t plane, const vec3_t a, const vec3_t b, const vec3_t c )
+bool PlaneFromPoints( plane_t &plane, const vec3_t a, const vec3_t b, const vec3_t c )
 {
 	vec3_t d1, d2;
 
 	VectorSubtract( b, a, d1 );
 	VectorSubtract( c, a, d2 );
-	CrossProduct( d2, d1, plane );
+	CrossProduct( d2, d1, plane.normal );
 
-	if ( VectorNormalize( plane ) == 0 )
+	if ( VectorNormalize( plane.normal ) == 0 )
 	{
 		return false;
 	}
 
-	plane[ 3 ] = DotProduct( a, plane );
+	plane.dist = DotProduct( a, plane.normal );
 	return true;
 }
 /*
@@ -293,7 +302,7 @@ bool PlaneFromPoints( vec4_t plane, const vec3_t a, const vec3_t b, const vec3_t
  * Returns false if the triangle is degenerate.
  * =====================
  */
-bool PlaneFromPointsOrder( vec4_t plane, const vec3_t a, const vec3_t b, const vec3_t c, bool cw )
+bool PlaneFromPointsOrder( plane_t &plane, const vec3_t a, const vec3_t b, const vec3_t c, bool cw )
 {
 	vec3_t d1, d2;
 
@@ -302,24 +311,24 @@ bool PlaneFromPointsOrder( vec4_t plane, const vec3_t a, const vec3_t b, const v
 
 	if ( cw )
 	{
-		CrossProduct( d2, d1, plane );
+		CrossProduct( d2, d1, plane.normal );
 	}
 
 	else
 	{
-		CrossProduct( d1, d2, plane );
+		CrossProduct( d1, d2, plane.normal );
 	}
 
-	if ( VectorNormalize( plane ) == 0 )
+	if ( VectorNormalize( plane.normal ) == 0 )
 	{
 		return false;
 	}
 
-	plane[ 3 ] = DotProduct( a, plane );
+	plane.dist = DotProduct( a, plane.normal );
 	return true;
 }
 
-bool PlanesGetIntersectionPoint( const vec4_t plane1, const vec4_t plane2, const vec4_t plane3, vec3_t out )
+bool PlanesGetIntersectionPoint( const plane_t &plane1, const plane_t &plane2, const plane_t &plane3, vec3_t out )
 {
 	// http://www.cgafaq.info/wiki/Intersection_of_three_planes
 
@@ -327,9 +336,9 @@ bool PlanesGetIntersectionPoint( const vec4_t plane1, const vec4_t plane2, const
 	vec3_t n1n2, n2n3, n3n1;
 	vec_t  denom;
 
-	VectorNormalize2( plane1, n1 );
-	VectorNormalize2( plane2, n2 );
-	VectorNormalize2( plane3, n3 );
+	VectorNormalize2( plane1.normal, n1 );
+	VectorNormalize2( plane2.normal, n2 );
+	VectorNormalize2( plane3.normal, n3 );
 
 	CrossProduct( n1, n2, n1n2 );
 	CrossProduct( n2, n3, n2n3 );
@@ -347,23 +356,21 @@ bool PlanesGetIntersectionPoint( const vec4_t plane1, const vec4_t plane2, const
 
 	VectorClear( out );
 
-	VectorMA( out, plane1[ 3 ], n2n3, out );
-	VectorMA( out, plane2[ 3 ], n3n1, out );
-	VectorMA( out, plane3[ 3 ], n1n2, out );
+	VectorMA( out, plane1.dist, n2n3, out );
+	VectorMA( out, plane2.dist, n3n1, out );
+	VectorMA( out, plane3.dist, n1n2, out );
 
 	VectorScale( out, 1.0f / denom, out );
 
 	return true;
 }
 
-void PlaneIntersectRay( const vec3_t rayPos, const vec3_t rayDir, const vec4_t plane, vec3_t res )
+void PlaneIntersectRay( const vec3_t rayPos, const vec3_t rayDir, const plane_t &plane, vec3_t res )
 {
 	vec3_t dir;
-	float  sect;
-
 	VectorNormalize2( rayDir, dir );
 
-	sect = - ( DotProduct( plane, rayPos ) - plane[ 3 ] ) / DotProduct( plane, rayDir );
+	vec_t sect = - ( DotProduct( plane.normal, rayPos ) - plane.dist ) / DotProduct( plane.normal, rayDir );
 	VectorScale( dir, sect, dir );
 	VectorAdd( rayPos, dir, res );
 }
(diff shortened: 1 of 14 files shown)


illwieckz:
Hmm, it looks like there is a mistake in tracing, I verified I can go through the pipe next to human base in station15.

illwieckz:
I fail to spot the bug in the code I wrote.

I wonder if introducing type safety for the plane is uncovering a dormant bug.

DolceTriade:
I worry it might be more harm than good...

illwieckz:
> I worry it might be more harm than good...

Keeping the current code is really bad, it's full of stupid things like that:

```c++
void function1(vec3_t v, float f) {
  /* code */
}

void function2(
  vec4_t v;
  function( v, v[3] );
}
```

Everything is implicit, like this:

```c++
vec4_t v;
VectorCopy( b, v );
```

Which is a shady way to do:

```c++
vec4_t v;
VectorCopy( b, v.xyz );
```

All of this is lying to the reader and asking for troubles.

slipher:
It's broken because you used `plane_t`, a value type, in function arguments, in places where it needs to be a reference/pointer.

illwieckz:
Right. This should now be better (on my end it is).

illwieckz:
On my end the station15 patch collision detection works again.

The unit test seems to be passing again.

This looks ready to me.

illwieckz:
To avoid confusion when reading my comment and my code, I said:

_`There already exist a `cplane_t` struct_ I have not modified to use `plane_t`, but I modified the  `cPlane_t` struct to use `plane_t` (we may look for better names).

src/engine/renderer/tr_local.h
```diff
@@ -3183,15 +3183,15 @@ inline bool checkGLErrors()
 	template<size_t frustumSize>
 	inline
 	typename std::enable_if<(frustumSize >= frustumBits_t::FRUSTUM_NEAR + 1)>::type
-	R_CalcFrustumNearCorners( const vec4_t(&frustum)[frustumSize], vec3_t (&corners)[ 4 ] ) {
-		extern void R_CalcFrustumNearCornersUnsafe( const vec4_t frustum[frustumBits_t::FRUSTUM_NEAR + 1], vec3_t (&corners)[ 4 ] );
+	R_CalcFrustumNearCorners( const plane_t(&frustum)[frustumSize], vec3_t (&corners)[ 4 ] ) {
```
slipher, line 3186: A frustum is not a plane. plane_t is not appropriate here

src/engine/qcommon/q_math.cpp
```diff
@@ -225,23 +225,44 @@ void ByteToDir( int b, vec3_t dir )
...
+}
+
+void PlaneSet( plane_t &out, const vec_t x, const vec_t y, const vec_t z, const vec_t dist )
+{
+	VectorSet( out.normal, x, y, z );
+	out.dist = dist;
+}
 
-	length2 = DotProduct( plane, plane );
+void PlaneMA( const plane_t &v, const float s, const plane_t &b, plane_t &out )
```
slipher, line 240: The notion of multiplying and adding a plane doesn't make sense to me.

slipher (commented):
I suggest reverting the changes to renderer-related files because none of those things actually seem to be planes.

src/engine/qcommon/q_math.cpp
```diff
@@ -225,23 +225,44 @@ void ByteToDir( int b, vec3_t dir )
 	VectorCopy( bytedirs[ b ], dir );
 }
 
-vec_t PlaneNormalize( vec4_t plane )
+void PlaneCopy( const plane_t &in, plane_t &out )
```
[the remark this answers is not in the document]
illwieckz: Right, now done.

src/engine/renderer/tr_local.h
illwieckz: I don't know what is a frustum, but if frustum is not a plane or doesn't have a plane, maybe we need another name for a struct composed of a normal vec3 and a distance float.

This code was already splitting the data into a normal vec3 and a distance float and naming it plane.

Examples:

```c++
VectorCopy(plane, tr.viewParms.frustums[0][i].normal);
tr.viewParms.frustums[0][i].dist = plane[3];
```

```c++
VectorNegate( lightProject[ 3 ], frustum[ FRUSTUM_FAR ] );
frustum[ FRUSTUM_FAR ][ 3 ] = -lightProject[ 3 ][ 3 ] - 1.0f;
```

```c++
using frustum_t = cplane_t[6];
```

```c++
struct cplane_t
{
	vec3_t normal;
	float  dist;
	byte   type; // for fast side tests: 0,1,2 = axial, 3 = nonaxial
	byte   signbits; // signx + (signy<<1) + (signz<<2), used as lookup during collision
	byte   pad[ 2 ];
};
```

All of this is code from _before_ my pull request.

illwieckz: @VReaperV how would you name a generic struct that would be composed like this:

```c++
struct something_t {
	float normal[3];
	float distance;
};
```

illwieckz: Here I read in French:

> On appelle espace métrique tout couple (E, d) o`u E 6 ≠ ∅ est un espace vectoriel et d est une distance.
> -- https://licence-math.univ-lyon1.fr/lib/exe/fetch.php?media=pmi:td3_-_topologie.pdf

Wikipedia says:

> Formally, a metric space is an [ordered pair](https://en.wikipedia.org/wiki/Ordered_pair) (M, d) where M is a set and d is a metric on M, i.e., a [function](https://en.wikipedia.org/wiki/Function_(mathematics))
>  -- https://en.wikipedia.org/wiki/Metric_space

With “a set” just being defined as:

> A set is the [mathematical model](https://en.wikipedia.org/wiki/Mathematical_model) for a collection of different things;
> -- https://en.wikipedia.org/wiki/Set_(mathematics)

Without normal meaning.

slipher:
> using frustum_t = cplane_t[6];

Well that's a tricky typedef. And kinda stupid since cplane_t has a bunch of extra fields that aren't relevant. OK, apparently they're defining the frustum by some bounding planes similarly to a brush. I was mistaken to say that none of the renderer stuff makes sense; the frustum use case is probably fine (though I'd prefer to see a separate commit). I still think a lot of the renderer changes *don't* make sense so I'll try to make some more specific comments.

src/engine/renderer/tr_light.cpp
```diff
@@ -438,7 +438,7 @@ void R_SetupLightLocalBounds( trRefLight_t *light )
 				else
 				{
 					vec3_t top;
-					const vec4_t* frustum = (const vec4_t*)light->localFrustum;
+					const plane_t* frustum = (const plane_t*)light->localFrustum;
```
slipher, line 441: This shouldn't need a cast

src/engine/renderer/tr_light.cpp
```diff
@@ -773,26 +770,28 @@ void R_SetupLightProjection( trRefLight_t *light )
 				VectorScale( right, ( 0.5f * dist ) / rLen, right );
 				VectorScale( up, - ( 0.5f * dist ) / uLen, up );
 
-				Vector4Set( lightProject[ 0 ], right[ 0 ], right[ 1 ], right[ 2 ], 0 );
-				Vector4Set( lightProject[ 1 ], up[ 0 ], up[ 1 ], up[ 2 ], 0 );
-				Vector4Set( lightProject[ 2 ], normal[ 0 ], normal[ 1 ], normal[ 2 ], 0 );
+				plane_t lightProject[ 4 ];
```
slipher, line 773: `lightProject` doesn't look like planes. "PlaneMA" and "DotProductPlane" aren't operations that make sense for planes as far as I can tell.

illwieckz: Maybe, but it is also processed separatly as veC3 and float, and feed some frustum:

```c++
VectorSubtract( lightProject[ 2 ], lightProject[ 1 ], frustum[ FRUSTUM_TOP ] );
frustum[ FRUSTUM_TOP ][ 3 ] = lightProject[ 2 ][ 3 ] - lightProject[ 1 ][ 3 ];
```

If it's not a plane, this would be the pefect candidate for a generic struct name for a `{ vec3, float }` couple which is actually processed like this.

One of the benefit of that change is that the day we use a better vector library that can do swizzles, it may be possible to do `something( lightProject.xyz, lightProject.t )` even if we go back to a vec4.

What I want to kill is to run vec3 things like `VectorSubstract` on a `vec4`.

illwieckz: Maybe “it's not yet a plane but the result of the computation is a plane”?

slipher: > Maybe, but it is also processed separatly as vec3 and float, and feeds some frustum:
> 
> ```c++
> VectorSubtract( lightProject[ 2 ], lightProject[ 1 ], frustum[ FRUSTUM_TOP ] );
> frustum[ FRUSTUM_TOP ][ 3 ] = lightProject[ 2 ][ 3 ] - lightProject[ 1 ][ 3 ];
> ```

This isn't actually treating the 4th component differently. It's just element-wise subtraction on vec4_t. Apparently VectorSubtract was used because there is no Vector4Subtract and the author wanted to code golf by doing it this way instead of writing out all 4 components or using a loop.

src/engine/renderer/tr_local.h
VReaperV: Frustum is a set of planes. More specifically the 6 planes that are the near and far clipping planes, and 4 planes that intersect the view origin, i. e. a pyramid with its top cut off (for the case of frustums used in rendering).

VReaperV: This `const vec4_t(&frustum)[frustumSize]` seems to just be getting the pointer to the first member of a plane in `frustum`, i. e. `cplane_t.normal[0]`. So that change looks correct to me since only normal and dist seem to be used.

src/engine/renderer/tr_light.cpp
illwieckz: Hmm, some operations may not be element-wise, for example:

```c++
VectorNegate( lightProject[ 3 ], frustum[ FRUSTUM_FAR ] );
frustum[ FRUSTUM_FAR ][ 3 ] = -lightProject[ 3 ][ 3 ] - 1.0f;
```
This is basically doing:

```c++
b[ 0 ] = -a[ 0 ];
b[ 1 ] = -a[ 1 ];
b[ 2 ] = -a[ 2 ];
b[ 3 ] = -a[ 3 ] - 1;
```

But this can be rewritten as:

```c++
Vector4Negate( a, b );
b[ 3 ] -= 1;
```

In the end this vec4 is copied to some frustum, so maybe I would need a Vec4ToPlane conversion.

Maybe w can consider this is some vector operation that builds-up a plane?

illwieckz: I reverted this part to some `vec4_t` operation, adding `Vector4Subtract` and `Vector4Negate` macros, and adding a `PlaneSet( plane_t, vec4_t )` function to set the frustum planes using the computed vec4.

VReaperV: > lightProject doesn't look like planes.

Those are, in fact, planes. Two of those are the right and up plane and one is near plane. The other planes are then derived using the near plane and the other 2 planes (since its normal is parallel to the lights direction apparently) and inverted (presumably to make the normals point outside of the frustum).

Plane dot product would presumably just be the dot product of their normals, i. e. represents the angle between planes.

The `ofs` that is calculated there seems to be used for shadowmaps (at least I don't think it can be non-parallel to the light target otherwise).

PlaneMA would then shift projection for the shadowmap.

See https://github.com/DaemonEngine/Daemon/blob/c5f8539fec2b29e7477c924e30c171d3e9fd2f65/src/engine/renderer/tr_main.cpp#L2327

(I haven't checked if DotProductPlane() and PlaneMA are correct to what old code was doing)

illwieckz: > (I haven't checked if DotProductPlane() and PlaneMA are correct to what old code was doing)

This was assumed to be the same as Vector4MA, though maybe multiplying the 4th component was wrong if “Plane dot product would presumably just be the dot product of their normals”. When that light code was “working” (i.e. rendering things), the computes were broken anyway.

I would not be surprised if the plane dot product on 4 elements is wrong, since I expect to find bugs in that code.

@VReaperV I can revert my revert if you prefer.

illwieckz: Well, we can merge this in current state (with things computed as vec4_t), where operations are done as vec4_t and then set to plane_t, with some comments for our future selves to fix and migrare that code later if needed.

Especially I would add a comment raising doubt about the dot product being done on 4 components.

VReaperV: > Hmm, some operations may not be element-wise, for example:

That is the calculation for the far plane derived from near plane. The near and far planes are parallel, so the negation effectively inverses the normal (the normal of far and near plane are opposite). `start` is just the lights position, `falloff` is a normalized vector that determines how far the light effect goes I think. The `-DotProduct( start, falloff )` is then the distance to the near plane if the lights frustum was rotated around the origin (0; 0; 0) (negated to get the normal pointing outside of the frustum).

Note that VectorNegate does not negate the 4th component. The negation is done in `frustum[ FRUSTUM_FAR ][ 3 ] = -lightProject[ 3 ][ 3 ] - 1.0f` to set the the near and far planes to the [0; 1] range.

illwieckz: > Note that VectorNegate does not negate the 4th component. 

Yes, that's exactly the kind of confusion I want to avoid with those patches.

My current code (using `vec4_t`) is doing:

```c++
Vector4Negate( lightProject[ 3 ], v );
v[ 3 ] -= 1.0f;
PlaneSet( frustum[ FRUSTUM_FAR ], v );
```

I added a `Vector4Negate` for this purpose.

VReaperV: > This was assumed to be the same as Vector4MA, though maybe multiplying the 4th component was wrong if “Plane dot product would presumably just be the dot product of their normals”. When that light code was “working” (i.e. rendering things), the computes were broken anyway.

Well, considering we don't know (I assume) what exactly is breaking, we should keep the 4-component MA and dot, and maybe note them.

slipher: Looks like it's doing the dot product of `(x y z 1)` and a plane to get the distance of `(x y z)` from the plane (with the result being positive on the "front" side or negative on the back). Maybe it would be better to use the 3-element dot product and then add the plane `dist` instead of building a vector with a dummy 4th value. That's how it is usually done in the tracing code for exmaple.

illwieckz: Those 4th components were just adding 0 to 0 and mutiplying 0 with 0 anyway…

The first commit do the whole porting of the code of plane_t, but this functions still uses vec4_t as temporary variables before saving them to `plane_t` to keep the changes less intrusive.

The second commit deeply rewrites the `R_SetupLightProjection` function to never use `vec4_t` and to do the usual `VectorMA`, `DotProduct` on vec3. The `plane_t` is then set with this `vec3_t` normal and the related distance, which is zero in three cases out of four (hence the useless Vector4MA and DotProduct4).

The only plane not having distance set to zero was not using `Vector4MA` neither `DotProduct4`.

The now-useless `vec4_t`-related macros are now removed.

src/engine/renderer/tr_light.cpp
```diff
@@ -743,122 +743,190 @@ void R_SetupLightProjection( trRefLight_t *light )
...
+					/* This was previously set as a vec4_t with targetGlobal[ 3 ] being 1
+					but then the fourth component never contributed to any computation
+					and was always zeroed after computation as a multiple of 0. */
+					vec3_t targetGlobal;
+					VectorSet( targetGlobal,
+						light->l.projTarget[ 0 ], light->l.projTarget[ 1 ],
+						light->l.projTarget[ 2 ] );
+
+					{
+						/* This was previously doing a DotProduct4 on a vec4_t
```
slipher, line 788: If you feel unsure about this refactoring, it would be better to avoid merging rather than write a bunch of comments apologizing for it. The comments don't help understand the code and just waste the reader's time.

slipher (commented):
The `vec_t` type is meh since when reading it sounds like a vector and only 1 different character marks it as a scalar. Don't consider this as a request to rewrite anything though.

src/engine/renderer/tr_light.cpp
illwieckz: I'm fine with this refactoring as a separate commit.

My comments are not there to apologize for this, it's just that I don't know what was the code doing, and some stuff didn't make sense. If it was right it was mostly by luck, if it was wrong, this should not make it worse. I just believe the comments may be useful to someone else dealing with this code in the future.

I haven't spotted any regression.

slipher: The point of a refactoring is to make the code nicer to work with before. The lengthy, irrelevant comments make it more difficult to read than it was before, so I will not LGTM it with those.

illwieckz: Well, those comments are useful for me as they are kind of extended line-per-line commit messages, and I want to have those comments in the commit introducing the change if I have to bisect the code later. But I'm fine at removing them right after that. I added a commit to remove those comments.

Also, at the time I wrote this code, the legacy renderer was not re-enabled again. Now that it works again, I could test those changes with the dynamic light renderer and spotted no regression.

illwieckz:
This looks ready to me.

illwieckz:
When I did this:

- https://github.com/DaemonEngine/Daemon/pull/1125

I first implemented it over this branch, and profiled the game in Orbit from time to time to see if I seen a difference in performances, what I have seen.

But then before opening the PR I rebased on master, and I profiled the game again, and I noticed that after rebasing on master, the game went slower. After I rebased back on this branch, the perf gain I was expecting was restored. The screenshots in #1125 were done with the code rebased on this branch, doing an _After_ screenshot while rebased on master would require to also redo the _Before_ one, because the rest of the code on master looks to be slower.

So, it looks like this code does not only clean-up the code and make it more readable to humans, but it may also help the compiler to grasp the code in input and optimize the generated code and make it faster too.

src/engine/renderer/tr_light.cpp
```diff
@@ -743,120 +743,159 @@ void R_SetupLightProjection( trRefLight_t *light )
...
+					PlaneSet( lightProject[ 3 ], falloff, -DotProduct( start, falloff ) );
 				}
 
-				if ( !VectorCompare( light->l.projStart, vec3_origin ) )
+				// we want the planes of s=0, s=q, t=0, and t=q
+				plane_t *frustum = light->localFrustum;
+
 				{
-					VectorCopy( light->l.projStart, start );
+					frustum[ FRUSTUM_LEFT ] = lightProject[ 0 ];
```
slipher, line 845: Remove random braces around single statements?

slipher (commented):
LGTM

src/engine/renderer/tr_light.cpp
illwieckz: Now done. I also found a couple of remaining “noisy comments” I deleted too.

Issue #1021, opened 2024-01-07
Portals don't render the skybox


illwieckz:
Portals don't render the skybox, while it is well known, it looks like we haven't reported it yet with a dedicated ticket.

![unvanquished_2024-01-08_003558_000](https://github.com/DaemonEngine/Daemon/assets/2311649/e4cbb936-014d-41d5-846a-1ac0e73b85b6)

VReaperV:
Fixed by #1044.

PR #1044, opened 2024-02-14
Optimisation: skybox rendering

### LINKED ISSUE #849: Sky performance cost
This isn't new to me, but the skybox is very costly. In fact the single `Tess_ClipSkyPoligons` function is almost as heavy (23.6%) as the whole cgame (26.10%):

[sky performance cost](https://dl.illwieckz.net/b/daemon/bugs/sky-performance-cost/20230429-082650-000.orbit-unvanquished-sky.png)

VReaperV:
Generate a persistent mesh for the skybox consisting of a cube with 8 vertices and 12 triangles.
Old code was creating a mesh consisting of many "blocks" (2 triangles each) on the CPU and uploading it to the GPU *every frame that the sky was rendered*. Not only that, but it performed operations on mapped memory. The mesh itself normally consisted of 1k+ triangles, which in itself is not much for modern hardware, but it's far more than needed and often more than the rest of the view combined. This was horrible for perfomance.

New mesh:
image
Old mesh example:
image
image

Move cloud generation to fragment shader. The fragment shader will now branch on a uniform to sample different textures, which will likely be faster than switching shader programs. The skybox itself will now always be centered on the view origin. For portals, the last projection matrix obtained when portalLevel == 0 is used because portals use a different projection matrix.

NUKE sky code used for clipping (i. e. generating those sky blocks). I've left the Tess_StageIteratorDepthFill() and Render_DepthFill() there even though they're unused, since I noticed that the engine uses different shader programs for depth pre-pass, so it might be possible to repurpose these functions for a more efficient depth pre-pass.

Fixes #849 and #1021.

Reference points for testing from https://github.com/DaemonEngine/Daemon/issues/849#issuecomment-1925448467:

## classic skybox

map|scene
-|-
plat23|default spectator view

new(FPS(ms))|old(FPS(ms))|r_fastsky 1(FPS(ms))
-|-|-
220(4.54ms)|200(5ms)|200(5ms)

## cloudy skybox

map|scene
-|-
chasm|default spectator view

new(FPS(ms))|old(FPS(ms))|r_fastsky 1(FPS(ms))
-|-|-
160(6.25ms)|150(6.66ms)|150(6.66ms)

## sky portal with classic skybox

map|player viewpos|skybox viewpos
-|-|-
station15|270 37 144 121 0|956 -2450 -3763 121 0 

new(FPS(ms))|old(FPS(ms))|r_fastsky 1(FPS(ms))
-|-|-
180(5.55ms)|180(5.55ms)|200(5ms)

## sky portal with cloudy skybox

map|player viewpos|skybox viewpos
-|-|-
[atcs-2015](https://users.unvanquished.net/~sweet/pkg/map-atcs-2015_0%2Bv3.2-final.dpk)|80 -106 457 50 -2|80 -106 -13362 50 0

new(FPS(ms))|old(FPS(ms))|r_fastsky 1(FPS(ms))
-|-|-
180(5.55ms)|170(5.88ms)|190(5.26ms)

## many sky brushes

map|default spectator view
-|-
testing0|default spectator view

new(FPS(ms))|old(FPS(ms))|r_fastsky 1(FPS(ms))
-|-|-
180(5.55ms)|120(8.33ms)|180(5.55ms)

map|default spectator view
-|-
[dretchstorm2](https://users.unvanquished.net/~sweet/pkg/map-dretchstorm2_2.dpk)|default spectator view

new(FPS(ms))|old(FPS(ms))|r_fastsky 1(FPS(ms))
-|-|-
150(6.66ms)|100(10ms)|150(6.66ms)

Portals:
unvanquished_2024-02-14_061302_000
unvanquished_2024-02-14_061337_000

### DIFF
--- a/src/engine/renderer/glsl_source/skybox_fp.glsl
+++ b/src/engine/renderer/glsl_source/skybox_fp.glsl
@@ -22,19 +22,56 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 /* skybox_fp.glsl */
 
+const float radiusWorld = 4096.0; // Value used by quake 3 skybox code
+
 uniform samplerCube	u_ColorMap;
-uniform vec3		u_ViewOrigin;
+uniform sampler2D	u_CloudMap;
+
+uniform bool        u_UseCloudMap;
+uniform float       u_CloudHeight;
+
+uniform mat4		u_TextureMatrix;
+
+uniform float		u_AlphaThreshold;
 
 IN(smooth) vec3		var_Position;
 
 DECLARE_OUTPUT(vec4)
 
+// Parametric cloud function used by quake 3 skybox code
+float ComputeCloudParametric( vec3 skyVec, float radiusWorld, float cloudHeight ) {
+	return (1.0 / 2.0 * dot( skyVec, skyVec ) ) * ( -2.0 * skyVec.z * radiusWorld
+			+ 2.0 * sqrt( skyVec.z * skyVec.z * radiusWorld * radiusWorld
+			+ 2.0 * skyVec.x * skyVec.x * radiusWorld * cloudHeight
+			+ skyVec.x * skyVec.x * cloudHeight * cloudHeight + 2.0 * skyVec.y * skyVec.y * radiusWorld * cloudHeight
+			+ skyVec.y * skyVec.y * cloudHeight * cloudHeight
+			+ 2.0 * skyVec.z * skyVec.z * radiusWorld * cloudHeight + skyVec.z * skyVec.z * cloudHeight * cloudHeight  ) );
+}
+
 void	main()
 {
 	// compute incident ray
-	vec3 incidentRay = normalize(var_Position - u_ViewOrigin);
+	vec3 incidentRay = normalize(var_Position);
+	vec4 color;
 
-	vec4 color = textureCube(u_ColorMap, incidentRay).rgba;
+	if( !u_UseCloudMap ) {
+		color = textureCube(u_ColorMap, incidentRay).rgba;
+	} else {
+		incidentRay *= ComputeCloudParametric( incidentRay, radiusWorld, u_CloudHeight );
+		incidentRay.z += radiusWorld;
+		incidentRay = normalize( incidentRay );
+		vec2 st = vec2( acos(incidentRay.x), acos(incidentRay.y) );
+		st = (u_TextureMatrix * vec4(st, 0.0, 1.0)).xy;
+		color = texture2D( u_CloudMap, st ).rgba;
+	}
+	
+	#if defined(USE_ALPHA_TESTING)
+		if( abs(color.a + u_AlphaThreshold) <= 1.0 )
+			{
+				discard;
+				return;
+			}
+	#endif
 
 	outputColor = color;
 }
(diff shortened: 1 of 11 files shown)


src/engine/renderer/tr_local.h
```diff
@@ -1634,6 +1634,7 @@ enum class deluxeMode_t { NONE, GRID, MAP };
 
 		float          fovX, fovY;
 		matrix_t       projectionMatrix;
+		matrix_t       projectionMatrixNonPortal; // For skybox rendering in portals
```
slipher, line 1637: Why do skyboxes need to work differently here?

VReaperV: Because portals set the near plane to the portal surfaces plane and far plane based on that, which results in the skybox getting clipped.

src/engine/renderer/tr_sky.cpp
```diff
@@ -24,845 +24,173 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
 
-			FillCloudBox( i );
-		}
-	}
-}
+			case stageType_t::ST_LIGHTMAP:
+			case stageType_t::ST_DIFFUSEMAP:
+			case stageType_t::ST_COLLAPSE_lighting_PHONG:
+			case stageType_t::ST_COLLAPSE_lighting_PBR:
+				Tess_DrawElements();
```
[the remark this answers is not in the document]
VReaperV: That was a mistake. Removed the stages that are not allowed in skybox.

src/engine/renderer/glsl_source/skybox_fp.glsl
```diff
@@ -22,19 +22,56 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
 
+// Parametric cloud function used by quake 3 skybox code
+float ComputeCloudParametric( vec3 skyVec, float radiusWorld, float cloudHeight ) {
+	return (1.0 / 2.0 * dot( skyVec, skyVec ) ) * ( -2.0 * skyVec.z * radiusWorld
+			+ 2.0 * sqrt( skyVec.z * skyVec.z * radiusWorld * radiusWorld
+			+ 2.0 * skyVec.x * skyVec.x * radiusWorld * cloudHeight
+			+ skyVec.x * skyVec.x * cloudHeight * cloudHeight + 2.0 * skyVec.y * skyVec.y * radiusWorld * cloudHeight
+			+ skyVec.y * skyVec.y * cloudHeight * cloudHeight
+			+ 2.0 * skyVec.z * skyVec.z * radiusWorld * cloudHeight + skyVec.z * skyVec.z * cloudHeight * cloudHeight  ) );
+}
```
illwieckz, line 49: Such large compute may exceeds the ALU limit of early R300 but anyway running the game on such hardware likely requires to use the `lowest` profile and then, disable the sky entirely. So I don't see this as a problem.

illwieckz (approved):
LGTM.

illwieckz:
Also, I love how the  size of the code is now only 33% of the original one… This is mind blowing! 🤯️

src/engine/renderer/glsl_source/skybox_fp.glsl
VReaperV: Well, https://https://unvanquished.net/about/ says daemon has compatibility for OpenGL 2.1, while R300 doesn't seem to support versions above 2.0, so it's a miracle if it works at all hehe.

PR #1045, opened 2024-02-14
WIP: clean-up the dynamic lighting and shadowing selection code

illwieckz:
This is low priority, but it may proves itself to be useful as a basis for various other tasks.

- First I deeply rewrote the dynamic lighting and shadowing selection code, to make it more readable and less confusing.

The `r_dynamicLight` cvar to enable dynamic lighting or not is now a boolean and is totally independant of the rendering technique. The related `r_staticLight` behaves the same (though, I don't really know what it is about).

The dynamic light renderer is now selected with a dedicated cvar named `r_dynamicLightRenderer`, The valye of `0` is the legacy forward one, `1` is the current tiled one.

As an example of how this clean-up may be useful, a 2 line commit makes the forward renderer working again to render dynamic lights.

I hope this work may prove to be as useful to help making dynamic shadows working again too.

Now that the forward dynamic light renderer is working again, and once dynamic shadowing will be working again, at least on the forward renderer, we may start NUKING things that don't break those features when we remove them.

I noticed that the dynamic lights are too bright when using the legacy forward renderer, which is a bug we had with the tiled renderer.

Even if the forward renderer is slower, it may still be useful. In the future I would like to introduce the ability to make the difference with gameplay-focused dynamic lights (like a lamp a player can carry with him), and purely cosmetic ones.

The fact the tiled renderer doesn't run on the complete hardware range supported by the engine, if we had a way to only render the only one or two dynamic lights a game may require for its gameplay (like a flashlight), having an alternate renderer supported by lower end hardware may prove useful, and its lower performance may not be a problem if, unlike the tiled renderer, it is not meant to render thousands of lights but only a couple of gameplay-related ones.


src/engine/renderer/gl_shader.cpp
```diff
@@ -482,13 +482,13 @@ static std::string GenEngineConstants() {
 	// Engine constants
 	std::string str;
 
-	if ( r_shadows->integer >= Util::ordinal(shadowingMode_t::SHADOWING_ESM16) && glConfig2.textureFloatAvailable )
+	if ( glConfig2.shadowMapping )
 	{
-		if ( r_shadows->integer == Util::ordinal(shadowingMode_t::SHADOWING_ESM16) || r_shadows->integer == Util::ordinal(shadowingMode_t::SHADOWING_ESM32) )
+		if ( r_shadows.Get() == Util::ordinal(shadowingMode_t::SHADOWING_ESM16) || r_shadows.Get() == Util::ordinal(shadowingMode_t::SHADOWING_ESM32) )
```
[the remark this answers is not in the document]
illwieckz: That's totally doable, I already do this in `switch case` blocks. It looks like most of the time the engine is casting enum to int (a quick grep on `Util::ordinal` is very verbose), and I'm accustomed to see them that's why I usually do it this way, unless the other way is more convenient (like in a `switch case`).

VReaperV:
As an aside, forward and tiled are 2 orthogonal things. Forward refers to a renderer where, after an optional depth pre-pass, it draws all the meshes with lighting etc. A deferred renderer outputs a bunch of screen-sized buffers first (diffuse, normal etc), then in another pass it renders a full-screen quad and combines all the buffers output earlier with lighting etc.

"Normal" (non-tiled, non-clustered) just iterates over every light for every fragment and adds up thei rcontributions. Tiled divides the view into tiles and assigns lights to them, then in the fragment shader only computes the lights that are in the same tile as the fragment. Clustered divides the view into clusters (i. e. along all 3 axes) and assigns lights to them, then computes only those lights that are in the same cluster as the fragment.

Both forward and deferred renderers can be "normal", tiled or clustered.

VReaperV:
> I noticed that the dynamic lights are too bright when using the legacy forward renderer, which is a bug we had with the tiled renderer.

This might be a problem with falloff since normally the lights attenuation is inversely proportional to the distance to light source squared, but gamma correction on the monitor effectively makes it squared instead. There might be an error somewhere there.

src/engine/renderer/gl_shader.cpp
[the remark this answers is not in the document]
illwieckz: I've done it in a better way.

illwieckz:
This is low priority, but I appreciate the fact it makes possible to better evaluate our lighting code and some of our fallbacks, and the shadowing code.

The code is also better at allocating or not allocating the related stuff used by the various features (so there can be some memory improvements).

src/engine/renderer/gl_shader.cpp
```diff
@@ -482,39 +482,67 @@ static std::string GenEngineConstants() {
...
-				AddDefine( str,"r_EVSMPostProcess", 1 );
-		}
-		else
-		{
-			AddDefine( str, "VSM", 1 );
 
-			if ( glConfig.hardwareType == glHardwareType_t::GLHW_R300 )
+			if ( r_evsmPostProcess->integer )
 			{
-				AddDefine( str, "VSM_CLAMP", 1 );
```
VReaperV, line 507: Missing whitespace before "r_EVSMPostProcess".

src/engine/renderer/tr_local.h
```diff
@@ -295,6 +295,8 @@ static inline void glFboSetExt()
 enum class lightMode_t { FULLBRIGHT, VERTEX, GRID, MAP };
 enum class deluxeMode_t { NONE, GRID, MAP };
 
+enum class dynamicLightRenderer_t { FORWARD, TILED };
```
VReaperV, line 298: This should be renamed from FORWARD to something else to not create confusion with the forward/deferred terminology.

src/engine/renderer/gl_shader.cpp
```diff
@@ -482,39 +482,67 @@ static std::string GenEngineConstants() {
...
+				AddDefine( str,"r_EVSMPostProcess", 1 );
```
VReaperV, line 512: Missing whitespace before "r_EVSMPostProcess".

src/engine/renderer/tr_local.h
illwieckz: I can use “LEGACY”, what would you recommend that describes best the implementation?

VReaperV: I think LEGACY is fine, there's no widely-accepted name for it.

src/engine/renderer/tr_shade.cpp
```diff
@@ -34,12 +34,28 @@ This file deals with applying shaders to surface data in the tess struct.
...
 	{
-		Log::Warn("Tiled dynamic light renderer not used because GL_ARB_texture_float is not available.");
-		glConfig2.dynamicLight = 0;
+		if ( glConfig2.dynamicLight || glConfig2.staticLight )
+		{
+			Log::Warn("Tiled dynamic light renderer not used because GL_ARB_texture_float is not available.");
+		}
+
+		glConfig2.dynamicLight = false;
+		glConfig2.staticLight = false;
```
slipher, line 49: Why this `staticLight = false`?

src/engine/renderer/tr_image.cpp
```diff
@@ -2496,44 +2513,36 @@ static void R_CreateDepthRenderImage()
...
-	height = glConfig.vidHeight;
-
-	w = (width + TILE_SIZE_STEP1 - 1) >> TILE_SHIFT_STEP1;
-	h = (height + TILE_SIZE_STEP1 - 1) >> TILE_SHIFT_STEP1;
-
-	imageParams_t imageParams = {};
-	imageParams.bits = IF_NOPICMIP | IF_RGBA32F;
-	imageParams.filterType = filterType_t::FT_NEAREST;
-	imageParams.wrapType = wrapTypeEnum_t::WT_ONE_CLAMP;
+	{
```
slipher, line 2516: Useless braces?

src/engine/renderer/tr_main.cpp
```diff
@@ -2353,40 +2353,57 @@ void R_AddLightInteractions()
...
 	{
 		light = tr.currentLight = &tr.refdef.lights[ i ];
 
-		if ( light->isStatic ) {
-			if ( r_staticLight->integer != 1 || ( ( r_precomputedLighting->integer || tr.worldLight != lightMode_t::MAP ) && !light->noRadiosity ) )
+		if ( light->isStatic )
+		{
+			if ( !glConfig2.staticLight )
 			{
-				if( r_staticLight->integer == 2 ) {
```
slipher, line 2364: The old code checks for 1 and 2 as values of r_staticLight but it was changed to bool.

illwieckz: This is on purpose the very first reason of this PR.

When the tiled renderer was made default in d3b1f7fbd6a3cc11dac15718323ff88340140534, it was needed to modify two cvars, `r_staticLight` and `r_dynamicLight`. Each cvar were both selecting a level of feature _AND_ a renderer. When I modified the default value of `r_dynamicLight` I failed to see it was also needed to set `r_staticLight`. Enabling or disabling the tiled renderer required to modify two cvars, but the same cvars were also used to enable or disable the features.

Now the `r_staticLight` and `r_dynamicLight` are only booleans enabling or not those features, and `r_dynamicLightRenderer` is selecting wich renderer we want for both.

src/engine/renderer/tr_image.cpp
illwieckz: It makes things easier to read and keep variables in a local scope. Those blocks could be dedicated functions. Other blocks now just do things the same way, even sharing the same variable names, but in their own dedicated block.

src/engine/renderer/tr_shade.cpp
illwieckz: Those two features use the dynamic light renderer.

They're selecting what kind of lights can be rendered with the dynamic light renderer.

What can be confusing is that we have a “dynamic light renderer”, rendering lights at run time, in comparison with pre-rendered lights like lightmaps.

But the game can feature static lights from the map and dynamic lights from the game, and both can be rendered with the dynamic light renderer in game.

I don't really know what are those static lights, maybe this is an alternative to lightmaps (some code seems to be mutually exclusive with `r_precomputedLighting`).

But the code explicitely was explicitely trying to renders those static lights with the legacy renderer when `r_staticLight` was set to `1` and with the tiled renderer when set to `2`. In fact the code for both kind of lights are almost a copy paste, they just have a different source.

src/engine/renderer/tr_main.cpp
illwieckz: For reference this code is the code to render the static lights with the dynamic renderer, which was and is still mostly a copy-paste of the code rendering dynamic lights with the dynamic renderer (see my other comment on that topic).

If the tiled renderer cannot run because of a missing feature, both those static lights rendered dynamically and the dynamic lights can't be rendered with that renderer.

src/engine/renderer/tr_main.cpp
```diff
@@ -2354,39 +2354,48 @@ void R_AddLightInteractions()
...
-					tr.pc.c_slights++;
-				}
-
-				light->cull = cullResult_t::CULL_OUT;
 				continue;
 			}
-		} else if ( light->l.inverseShadows ) {
-			if( glConfig2.dynamicLight == 0 ) {
+			else if ( ( r_precomputedLighting->integer || tr.worldLight != lightMode_t::MAP ) && !light->noRadiosity )
+			{
```
illwieckz, line 2369: Here is the code skipping the render of static lights as dynamic ones when `r_precomputedLighting` is enabled.

src/engine/renderer/tr_shade.cpp
illwieckz: Some comments in the code suggest the renderer was meant to be able to render the map like Doom3 did.

Though, this feature doesn't work (like shadows), and fixing that is outside of the scope of this PR (like shadows), but this PR makes the code easier to deal with in order to be able to possibly work on fixing the feature later (like shadows).

After this PR, enabling what kind of lights can be rendered dynapically and selecting the default dynamic light renderer will be different options.

slipher:
Were you able to test the "static light" thing? [Here](https://users.unvanquished.net/~slipher/map-entity-directory.txt) it can be seen that only a handful of maps us the `light` entity. Curiously, old versions of Forlorn and Yocto used it, but it was removed years ago.

illwieckz:
> Were you able to test the "static light" thing? [Here](https://users.unvanquished.net/~slipher/map-entity-directory.txt) it can be seen that only a handful of maps us the `light` entity. Curiously, old versions of Forlorn and Yocto used it, but it was removed years ago.

I only tested the cvars, no specific assets. But anyway fixing features is not the purpose of this PR, all I want is to have the selection of them being meaningful, and the code more explicit.

slipher:
LGTM

illwieckz:
I removed the extra commit that was claiming to fix dynamic lights with legacy renderer, it was likely wrong.

I discovered that enabling the legacy dynamic light renderer can't get it working with `vid_restart`, one has to load the next map. I don't know what causes that but it was already that way before this PR and this PR doesn't change anything on that (since it's currently only a rewrite to make the code easier to read).

So, both legacy and tiled dynamic renderers are already working, we just cannot switch from tiled to legacy with `vid_restart` yet because of a bug that is outside the topic of this PR.

PR #1046, opened 2024-02-16
renderer: revamp shader stage collapsing

illwieckz:
This deeply rewrites the stage collapse, I needed this when implementing sRGB as I had to be sure of what I was rendering.

This also fix some bugs and uncovers some. For example our creep material is currently lacking a lightmap but because of a bug it was rendered properly. Once this is merged it will be rendered as buggy as it should be with the mistake.

This also collapse more stages. For legacy materials done this way:

```
color map
lightmap with multiplication
```

```
light map
color map with multiplication
```

They are automatically turned into diffuse stages.

While for legacy materials done this way:

```
color map
lightmap with multiplication
colormap with addition
```

```
light map
color map with multiplication
color map with addition
```

They are automatically turned into diffuse stages with glow map.

The `implicitLightmap` boolean is now removed, all diffuse maps have a light map, collapsed stages without lightmap use another stage type.

So this not only fixes bug, it make things less complex, and improve performance.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -1404,7 +1404,7 @@ static bool ParseMap( const char **text, char *buffer, int bufferSize )
 	return true;
 }
 
-static bool LoadMap( shaderStage_t *stage, const char *buffer, const int bundleIndex = TB_COLORMAP )
+static bool LoadMap( shaderStage_t *stage, const char *buffer, stageType_t type, const int bundleIndex = TB_COLORMAP )
 {
 	char         *token;
 	const char         *buffer_p = &buffer[ 0 ];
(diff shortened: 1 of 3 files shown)


illwieckz:
I need this revamp as a base for multiple changes in multiple branches, so unless someone has a remark on it I will merge it tomorrow.

src/engine/renderer/tr_shader.cpp
```diff
@@ -2060,40 +2081,41 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
...
 				Log::Warn("Supposedly you shouldn't have a physicalMap after a specularMap (in shader '%s')?", shader.name);
-				// use PBR instead
+				// keep PHONG instead
 			}
-
+			else
+			{
 			// Daemon PBR packing defaults to ORM like glTF 2.0
-			stage->collapseType = collapseType_t::COLLAPSE_lighting_PBR;
+				stage->collapseType = collapseType_t::COLLAPSE_PBR;
```
slipher, line 2115: Bad spacing

src/engine/renderer/tr_shader.cpp
```diff
@@ -1404,7 +1404,7 @@ static bool ParseMap( const char **text, char *buffer, int bufferSize )
 	return true;
 }
 
-static bool LoadMap( shaderStage_t *stage, const char *buffer, const int bundleIndex = TB_COLORMAP )
+static bool LoadMap( shaderStage_t *stage, const char *buffer, stageType_t type, const int bundleIndex = TB_COLORMAP )
```
slipher, line 1407: In which case is `type` meaningfully different from `stage->type`? (I saw one case where `stage->type` is ATTENUATIONMAP_Z and `type` is COLORMAP but this doesn't matter as neither receives special treatment.)

slipher (commented):
Well no one understands this besides you, so I can just point out whitespace errors.

src/engine/renderer/tr_shader.cpp
illwieckz: With a legacy material like this:

```
textures/blah
{
    diffuseMap blah_d
    normalMap blah_n
    specularMap blah_s
}
```

or like this (exactly the same but more verbose):

```
textures/blah
{
  {
    map blah_d
    stage diffuseMap
  }
  {
    map blah_n
    stage normalMap
  }
  {
    map blah_s
    stage specularMap
  }
}
```

Every stage is an actual stage and every map (image) is a different stage.

But with a material with a pre-collapsed stage like this

```
textures/blah
{
  {
    diffuseMap blah_d
    normalMap blah_n
    specularMap blah_s
  }
}
```

There is a single stage, for which the type is `diffuseMap`, but three maps (images), for which each type is `diffuseMap`, `normalMap` and `specularMap`.

When loading a map (image), we need to know for what kind of stage it is, being already collapsed or not.

With this pre-collapsed stage, the stage type is always `diffuseMap`, but we tell the image loader to run the specific code for `normalMap` stage when loading the normal maps and the one for `specularMap` stage when loading the specular map.

Passing an explicit `stageType_t` for loading an image that may differ from the stage type itself (because of multitexturing) was the least intrusive way, alternative would be to have an `imageType_t` which would require much more development, and would be prone to mistakes since we would always have to keep it in sync with `stageType_t` anyway.

illwieckz: Actually the reuse of those `stageType_t` for image types is the consequence of the fact we not only need to have map types, but to duplicate map types as stage types to keep compatibility with legacy syntaxes. And since all those image types also have to be separate stage types before collapsing, I just reuse them for image types. But in fact those stage types are the reuse of image types, it's just less verbose, less prone to errors, less messy and less intrusive to have a single enum for everything.

illwieckz:
> Well no one understands this besides you, so I can just point out whitespace errors.

Yeah, that's the problem with this kind of revamp on such part of the code. There can be generic mistakes and typos though, or some C++esque suggestion.

PR #1050, opened 2024-02-20
renderer: implement complete overBright in GLSL and high-precision frame-buffer

illwieckz:
* renderer: implement complete overBright in GLSL
    
We cannot overBright separate stage as stage are clamped within `[0.0, 1.0]` when blending multiple stages together, meaning the result of the multiplication of a separate light stage with a light factor will be clamped before blending it with the color map.
    
This implementation implements overBright in the camera shader, but to only apply this overBright to surfaces having received lights, surfaces that do not receive lights gets divided by the light factor in a way re-multiplying them cancels the division.
    
Stages and surfaces to be blended over other stages or surfaces are also divided to avoid multiplying multiple time the same surface, in order to have the final result being `(m * a) + b` instead of `m * (a + b)` wich would be
`(m * a) + (m * b)` and then sump up the factor, something we don't want to do.

The legacy code for overbrighting (but clamping) the light at BSP load is kept for when the feature is disabled, but rewritten to make the code better.

This feature can be enabled by disabling the `r_mapClampOverBright` cvar.

* renderer: implement high precision rendering framebuffers

This is needed to avoid color-banding with the various division/multiplication round trips of the complete overBright implementation. This will also be needed for sRGB colors in the future too. It is enabled by default if a feature requires it.

This is optional and can be disabled with `r_highPrecisionRendering`, or automatically disabled if the hardware doesn't support the feature.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -1615,6 +1615,7 @@ GLShader_generic::GLShader_generic( GLShaderManager *manager ) :
 	u_ModelMatrix( this ),
 	u_ProjectionMatrixTranspose( this ),
 	u_ModelViewProjectionMatrix( this ),
+	u_InverseLightFactor( this ),
 	u_ColorModulate( this ),
 	u_Color( this ),
 	u_Bones( this ),
@@ -1653,6 +1654,7 @@ GLShader_lightMapping::GLShader_lightMapping( GLShaderManager *manager ) :
 	u_ViewOrigin( this ),
 	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
+	u_InverseLightFactor( this ),
 	u_Bones( this ),
 	u_VertexInterpolation( this ),
 	u_ReliefDepthScale( this ),
@@ -1973,6 +1975,7 @@ GLShader_skybox::GLShader_skybox( GLShaderManager *manager ) :
 	u_AlphaThreshold( this ),
 	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
+	u_InverseLightFactor( this ),
 	u_VertexInterpolation( this ),
 	GLDeformStage( this ),
 	GLCompileMacro_USE_ALPHA_TESTING( this )
@@ -1989,6 +1992,7 @@ GLShader_fogQuake3::GLShader_fogQuake3( GLShaderManager *manager ) :
 	GLShader( "fogQuake3", ATTR_POSITION | ATTR_QTANGENT, manager ),
 	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
+	u_InverseLightFactor( this ),
 	u_Color( this ),
 	u_Bones( this ),
 	u_VertexInterpolation( this ),
@@ -2017,6 +2021,7 @@ GLShader_fogGlobal::GLShader_fogGlobal( GLShaderManager *manager ) :
 	u_ViewMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
 	u_UnprojectMatrix( this ),
+	u_InverseLightFactor( this ),
 	u_Color( this ),
 	u_FogDistanceVector( this ),
 	u_FogDepthVector( this )
@@ -2094,7 +2099,8 @@ void GLShader_portal::SetShaderProgramUniforms( shaderProgram_t *shaderProgram )
 
 GLShader_contrast::GLShader_contrast( GLShaderManager *manager ) :
 	GLShader( "contrast", ATTR_POSITION, manager ),
-	u_ModelViewProjectionMatrix( this )
+	u_ModelViewProjectionMatrix( this ),
+	u_InverseLightFactor( this )
 {
 }
 
@@ -2108,6 +2114,7 @@ GLShader_cameraEffects::GLShader_cameraEffects( GLShaderManager *manager ) :
 	u_ColorModulate( this ),
 	u_TextureMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
+	u_LightFactor( this ),
 	u_DeformMagnitude( this ),
 	u_InverseGamma( this )
 {
--- a/src/engine/renderer/glsl_source/cameraEffects_fp.glsl
+++ b/src/engine/renderer/glsl_source/cameraEffects_fp.glsl
@@ -24,6 +24,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 uniform sampler2D u_CurrentMap;
 uniform sampler3D u_ColorMap;
+
+uniform float u_LightFactor;
 uniform vec4      u_ColorModulate;
 uniform float     u_InverseGamma;
 
@@ -36,9 +38,11 @@ void	main()
 	// calculate the screen texcoord in the 0.0 to 1.0 range
 	vec2 st = gl_FragCoord.st / r_FBufSize;
 
-	vec4 original = clamp(texture2D(u_CurrentMap, st), 0.0, 1.0);
+	vec4 color = texture2D(u_CurrentMap, st);
 
-	vec4 color = original;
+	color.rgb *= u_LightFactor;
+
+	color = clamp(color, 0.0, 1.0);
 
 	// apply color grading
 	vec3 colCoord = color.rgb * 15.0 / 16.0 + 0.5 / 16.0;
--- a/src/engine/renderer/glsl_source/lightMapping_fp.glsl
+++ b/src/engine/renderer/glsl_source/lightMapping_fp.glsl
@@ -27,6 +27,7 @@ uniform sampler2D	u_MaterialMap;
 uniform sampler2D	u_GlowMap;
 
 uniform float		u_AlphaThreshold;
+uniform float u_InverseLightFactor;
 uniform vec3		u_ViewOrigin;
 
 IN(smooth) vec3		var_Position;
@@ -172,9 +173,25 @@ void main()
 		color.rgb += 0.7 * emission;
 	#endif
 
+	/* HACK: use sign to know if there is a light or not, and
+	then if it will receive overbright multiplication or not. */
+	if ( u_InverseLightFactor > 0 )
+	{
+		color.rgb *= u_InverseLightFactor;
+	}
+
 	#if defined(r_glowMapping)
 		// Blend glow map.
-		color.rgb += texture2D(u_GlowMap, texCoords).rgb;
+		vec3 glow = texture2D(u_GlowMap, texCoords).rgb;
+
+		/* HACK: use sign to know if there is a light or not, and
+		then if it will receive overbright multiplication or not. */
+		if ( u_InverseLightFactor < 0 )
+		{
+			glow *= - u_InverseLightFactor;
+		}
+
+		color.rgb += glow;
 	#endif
 
 	outputColor = color;
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -1739,9 +1739,9 @@ image_t *R_FindImageFile( const char *imageName, imageParams_t &imageParams )
 		return nullptr;
 	}
 
-	if ( imageParams.bits & IF_LIGHTMAP )
+	if ( imageParams.bits & IF_LIGHTMAP && tr.forceLegacyMapOverBrightClamping )
 	{
-		R_ProcessLightmap( pic[ 0 ], 4, width, height, imageParams.bits, pic[ 0 ] );
+		R_ProcessLightmap( pic[ 0 ], width, height, imageParams.bits );
 	}
 
 	image = R_CreateImage( ( char * ) buffer, (const byte **)pic, width, height, numMips, imageParams );
@@ -2960,7 +2960,8 @@ void R_InitImages()
 	Because tr.overbrightBits is always 0, tr.identityLight is
 	always 1.0f. We can entirely remove it. */
 
-	tr.mapOverBrightBits =  r_mapOverBrightBits.Get();
+	tr.mapOverBrightBits = r_mapOverBrightBits.Get();
+	tr.forceLegacyMapOverBrightClamping = r_forceLegacyMapOverBrightClamping.Get();
 
 	// create default texture and white texture
 	R_CreateBuiltinImages();
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -2408,13 +2408,22 @@ static void R_CreateCurrentRenderImage()
 
 	imageParams_t imageParams = {};
 	imageParams.bits = IF_NOPICMIP;
+
+	if ( glConfig2.textureFloatAvailable && r_highPrecisionRendering.Get() )
+	{
+		imageParams.bits |= IF_RGBA16;
+	}
+
 	imageParams.filterType = filterType_t::FT_NEAREST;
 	imageParams.wrapType = wrapTypeEnum_t::WT_CLAMP;
 
 	tr.currentRenderImage[0] = R_CreateImage( "_currentRender[0]", nullptr, width, height, 1, imageParams );
 	tr.currentRenderImage[1] = R_CreateImage( "_currentRender[1]", nullptr, width, height, 1, imageParams );
 
-	imageParams.bits |= IF_PACKED_DEPTH24_STENCIL8;
+	imageParams = {};
+	imageParams.bits = IF_NOPICMIP | IF_PACKED_DEPTH24_STENCIL8;
+	imageParams.filterType = filterType_t::FT_NEAREST;
+	imageParams.wrapType = wrapTypeEnum_t::WT_CLAMP;
 
 	tr.currentDepthImage = R_CreateImage( "_currentDepth", nullptr, width, height, 1, imageParams );
 }
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -82,6 +82,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_dynamicLightCastShadows;
 	cvar_t      *r_precomputedLighting;
 	Cvar::Cvar<int> r_mapOverBrightBits("r_mapOverBrightBits", "default map light color shift", Cvar::NONE, 2);
+	Cvar::Cvar<bool> r_forceLegacyMapOverBrightClamping("r_forceLegacyMapOverBrightClamping", "clamp over bright of legacy maps (enable multiplied color clamping and normalization)", Cvar::NONE, false);
 	Cvar::Range<Cvar::Cvar<int>> r_lightMode("r_lightMode", "lighting mode: 0: fullbright (cheat), 1: vertex light, 2: grid light (cheat), 3: light map", Cvar::NONE, Util::ordinal(lightMode_t::MAP), Util::ordinal(lightMode_t::FULLBRIGHT), Util::ordinal(lightMode_t::MAP));
 	cvar_t      *r_lightStyles;
 	cvar_t      *r_exportTextures;
@@ -1115,6 +1116,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_dynamicLightCastShadows = Cvar_Get( "r_dynamicLightCastShadows", "1", 0 );
 		r_precomputedLighting = Cvar_Get( "r_precomputedLighting", "1", CVAR_CHEAT | CVAR_LATCH );
 		Cvar::Latch( r_mapOverBrightBits );
+		Cvar::Latch( r_forceLegacyMapOverBrightClamping );
 		Cvar::Latch( r_lightMode );
 		r_lightStyles = Cvar_Get( "r_lightStyles", "1", CVAR_LATCH | CVAR_ARCHIVE );
 		r_exportTextures = Cvar_Get( "r_exportTextures", "0", 0 );
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -186,6 +186,9 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_halfLambertLighting;
 	cvar_t      *r_rimLighting;
 	cvar_t      *r_rimExponent;
+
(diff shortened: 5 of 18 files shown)


illwieckz:
Complete overBright in GLSL is currently disabled by default, the code is considered good enough to be enabled by users to evaluate it and report issues. This may be improved incrementally in the future but I haven't spot a single error myself yet.

Maps like `terminus` and `area3` are not rendered properly (too bright) but that's not because of a bug of this code. This is  because we render them with mapOverBrightBits feature being enabled but those maps were initially made for a game disabling mapOverBrightBits by default. This can be fixed by setting the `"mapOverBrightBits" "0"` key in the worldspawn entity of these maps, or if the clamped overBrightBits pleases the eye, keep the overBright but clamp it by setting the `"mapClampOverBright" "0"` key in the worldspawn entity.

illwieckz:
Future sRGB code will require this complete overBright implementation and will also need high precision frame buffer for better results.

illwieckz:
This was tested with maps like:

- test-portal-recursion (portals and mirrors).
- atcshd (legacy materials, light known to be broken by clamped overBright)
- atcszalpha (legacy materials, light known to be broken by clamped overBright)
- metro (legacy materials, light styles)
- pulse (legacy materials, light styles)
- plat23 (state of the art materials)
- Wolf:ET's oasis (legacy materials, global fog, mixed light map and vertex light)

No visual regressions were noticed.

src/engine/renderer/tr_init.cpp
```diff
@@ -77,6 +77,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_dynamicLightCastShadows;
 	cvar_t      *r_precomputedLighting;
 	Cvar::Cvar<int> r_mapOverBrightBits("r_mapOverBrightBits", "default map light color shift", Cvar::NONE, 2);
+	Cvar::Cvar<bool> r_mapClampOverBright("r_mapClampOverBright", "clamp over bright of legacy maps (enable multiplied color clamping and normalization)", Cvar::NONE, true);
```
slipher, line 80: Why the word "map" in these cvar names? Seems backwards since these are only used when a value is NOT provided by the map. I suggest `r_defaultOverbrightBits` and `r_defaultClampOverbright` with the word "default" emphasizing that changes to their value might be ignored.

slipher (commented):
Same question as https://github.com/DaemonEngine/Daemon/pull/1035#pullrequestreview-1847103166 - among lightmaps/light grids/vertex lighting which ones is the new algorithm expected to apply to?

illwieckz:
> Same question as [#1035 (review)](https://github.com/DaemonEngine/Daemon/pull/1035#pullrequestreview-1847103166) - among lightmaps/light grids/vertex lighting which ones is the new algorithm expected to apply to?

All of them. There should be overbright anytime there is a precomputed light.

With that reverse implementation there is overbright cancelling anytime there is no precomputed light, in a way multiplying the final result reverts non-lit surface to their original colors, and multiplies lit surfaces to their overbright colors.

That just makes me think we may have to divide the dynamic lights, as those are expected to not be multiplied. The overbright light factor is for precomputed lights only.

src/engine/renderer/tr_init.cpp
illwieckz: There was originally two cvars, `r_overBrightBits` and `r_mapOverBrightBits`. The first one was configuring some hardware feature, and the engine was multiplying the lights with `r_mapOverBrightBits` and `r_overBrightBits`.

Let's imagine we had `r_mapOverBrightBits` set to 2, if there was no hardware overbright bits (`r_overBrightBits 0`), the lights were fully multiplied by 2, if there was 1 hardware overbright bit (`r_overBrightBits 1`), the lights were only multiplied by 1, and if there was 1 hardware overbright bits (`r_overBrightBits 2`), the lights were not multiplied at all.

The related `rgbGen identity` and `rgbGen identityLighting` were also related to this, as one was related to blending something with multiplied light or premultiplied light, meaning one of them had to cancel the hardware overbright multiplication, with `rgbGen identity` and `rgbGen identityLighting`  being equal if there was no hardware overbright bit (nothing to compensate)

But since our engine doesn't implement hardware overbright bit, the only cvar left is `r_mapOverBrightBits` and `rgbGen identity` and `rgbGen identityLighting` are equal (nothing to compensate).

With that being said, one may claim that to be fully compatible with hardware overbrightbits, one may only have to multiply the end result. But the life is more complicated than that. This [official Quake  3 documentation](https://icculus.org/gtkradiant/documentation/Q3AShader_Manual/ch05/pg5_1.htm) said on `6.3.1 RgbGen identityLighting` chapter:

> 6.3.1 RgbGen identityLighting
Colors will be (1.0,1.0,1.0) if running without overbright bits (NT, linux, windowed modes), or (0.5, 0.5, 0.5) if running with overbright. Overbright allows a greater color range at the expense of a loss of precision. Additive and blended stages will get this by default.

According to this documentation, the last platform to support hardware overbright bits was Windows 98 when playing fullscreen. Also, it is said in other places that the feature was not compatible with LCD screens and then required a CRT screen.

It happens that some more recent may have provided the feature more recently, as [mentionned here](https://github.com/DaemonEngine/Daemon/pull/1034#issuecomment-1933790759), but for most people the feature is lost since forever.

It means than since 20 years most mappers and developers were testing their maps and graphical effects without that feature, and may have implemented things without testing them with hardware overbright. And since we don't do hardware overbright, we have to face the ultimate limitation: the precision of the user display, which is 24-bit, 8 bit per color. One good example of thing we cannot multiply is the Wolf:ET global fog, the screen itself is clamping the multiplied global fog, and it's permitted to have very strong doubts developer expected that fog to be multiplied, it looks wrong even on non-clamped values.

illwieckz: The `map` prefix in `r_mapClampOverBright` is mainly cargo cult, but it means that option sets a value that can differ per map.

For `r_mapOverBrightBits`, this is the default per-map value for when the map doesn't set `mapOverBrightBits` key in worldspawn, so I did the same for `r_mapClampOverBright`, the default per-map value for when the map doesn't set it.

illwieckz:
> That just makes me think we may have to divide the dynamic lights, as those are expected to not be multiplied. The overbright light factor is for precomputed lights only.

Dynamic lights are now divided to cancel overBright on them.

src/engine/renderer/tr_init.cpp
illwieckz: We can reword the explanation in the way `mapOverBrightBits` is a BSP worldspawn entity key, and the `r_mapOverBrightBits` cvar carries the engine default value for it when the BSP doesn't set it, hence the same name.

In the same way new `r_mapClampOverBright` engine cvar carries the default value for the new `mapClampOverBright` BSP worldspawn key. I can name those cvar and key as `r_clampOverBright` and `clampOverBright` if people prefer.  It's fine, as the 

But `mapOverBrightBits` and related cvar should be kept that way for legacy compatibility, even if the extra `map` prefix for a map key is superfluous.

illwieckz: > I can name those cvar and key as r_clampOverBright and clampOverBright if people prefer. It's fine.

At least for the engine cvar, the `map` prefix reminds this value can be override by the map itself, which is not bad and may be better. If the map sets the value, it will help people understand why tweaking the engine cvar does nothing.

illwieckz:
So, what people can do (just to exclude bugs from other buggy features):

- check that `r_dynamicLight` is 2
- check that `cg_shadows` is either 0 or 1, not more

Then:

- set `r_mapClampOverBright` to `off`

And play public games this way, and report which map break. This way we can decide to disable it by default or not.

I expect maps from Tremulous being all fine, but some maps built for Unvanquished having bugs. The thing is that mappers have tested against their maps with a broken lighting and then, mappers were not aware they introduced some bugs.

Example of maps I spotted:

map|bug|proposed fix
-|-|-
freeway|building windows are too bright|set `mapClampOverBright 1` in BSP worldspawn
fortification|outdoor surfaces are too bright|set `mapClampOverBright 1` in BSP worldspawn

The two maps ported from Urban Terror are also buggy, but that's expected: we set `mapOverBrightBits 2` on them by default but the original game expected `0`. It happens that `mapOverBrightBits 2` is not bad with the terminus map when overbright is clamped, so we may do that instead.

map|bug|proposed fix
-|-|-
area3|outdoor surfaces are too bright|set `mapOverBrightBits 0` in BSP worldspawn
terminus|many surfaces are too bright|set `mapClampOverBright 1` in BSP worldspawn

illwieckz:
Actually freeway having “too bright windows“ may be due to something wrong: the glow maps are multiplied but we may not want to do this. Problem is, with collapsed stages, we know what is a glow map, but with legacy stages, especially if the engine fails to collapse them, we don't know what is a glow map.

I thought today of an alternate implementation for GLSL not-clamped overBrightBits: the reason why I multiply everything at the end but divide what should not be multiplied to cancel the multiplication is that framebuffers are clamped to 1, so if we multiply a legacy light stage, it will be clamped to 1 before blending the texture over it, so it's as bad as if we clamp the lightmap at load time.

_But_, what I thought today is that since the engine max overBrightBits is 3, we may divide all stages by `2^3`, then multiply the final result by `2^3`, then we would only multiply the lights, only the lights. Because the stage would always be divided before blending, and divided with a value greater or equal to the overBright multiplier, it will never be clamped.

illwieckz:
I cancelled overBright on collapsed glow maps.

It looks like I already cancelled on separate stages one, in fact they are simply detected as other blended surfaces that should not be overBright, without needing to know they are glow stages.

The remaining problem with freeway map seems to be that bloom is multiplied.

illwieckz:
Bloom overbright is now properly cancelled.

illwieckz:
Here is a combination of this branch and the [ATCSHD remaster](https://github.com/InterstellarOasis/map-atcshd_src.dpkdir/pull/3) @RedSkyByte is working on:

unvanquished_2024-03-04_142403_000

unvanquished_2024-03-04_142436_000

unvanquished_2024-03-04_142444_000

unvanquished_2024-03-04_142827_000

DolceTriade:
Been running this for a bit, and I haven't noticed any regression. The only oddness I have seen is in the map Archao, but that oddness was present on master too (flickering textures at the spectator spawn point).

illwieckz:
What do you think about renaming `mapClampOverBright` to `legacyMapClampOverBright`? The clamping is only a backward compatibility purposed for being compatible with _bugs_ of legacy maps.

If you don't mind, I think I will merge this soon if no one object. Not having this is blocking the work on sRGB colorspace:

The clamping will be automatically disabled on maps using the sRGB/linear colorspace in all cases, meaning all newly built maps will use that code.

After this is merged, there can be other improvements done:

- Disable the legacy map overbright clamping by default, I tested many maps, and it's fine.
  In case we stumble upon a map _requiring_ those bugs, we can repackage it with the key to enable the clamping.
- I may evaluate some redesign of the feature in the future with some fresh ideas.
  We don't need to wait for such revamp to merge this, this implementation works.

slipher:
I insist that we should use "default" in the cvar names lest users be surprised when the cvar fails to override the value set in the worldspawn. `r_defaultMapOverbrightBits` and `r_default(Map)ClampOverbright`. (I don't get why the word "map" should be included in the worldspawn key name since there is not any other kind of "clamp overbright" that it needs to be distinguished from.)

illwieckz:
> r_defaultMapOverbrightBits

We better not want to rename `r_mapOverBrightBits`. This one was kind of public API for games to configure the engine.

src/engine/renderer/glsl_source/lightMapping_fp.glsl
```diff
@@ -27,6 +27,7 @@ uniform sampler2D	u_MaterialMap;
 uniform sampler2D	u_GlowMap;
 
 uniform float		u_AlphaThreshold;
+uniform float u_LightFactor;
```
slipher, line 30: Maybe you could optimize it a bit by inverting (x -> 1/x) this uniform so that you can use multiplication instead of division. Also by setting it to 0 instead of 1 when there is no need for multiplications so then the operations can be skipped.

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
```diff
@@ -36,9 +38,11 @@ void	main()
 	// calculate the screen texcoord in the 0.0 to 1.0 range
 	vec2 st = gl_FragCoord.st / r_FBufSize;
 
-	vec4 original = clamp(texture2D(u_CurrentMap, st), 0.0, 1.0);
+	vec4 color = clamp(texture2D(u_CurrentMap, st), 0.0, 1.0);
```
slipher, line 41: This would be a nice place for a comment as the double clamping really makes the foolishness of the legacy algorithm apparent.

slipher:
Have you tested with different values of `r_lightMode`? I tried fullbright mode with clamping off and it looks really bad.

unvanquished-1

src/engine/renderer/tr_local.h
```diff
@@ -1207,6 +1207,8 @@ enum class deluxeMode_t { NONE, GRID, MAP };
 
 		bool            dpMaterial;
 
+		bool shaderHasNoLight;
```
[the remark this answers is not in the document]
illwieckz: This boolean is only used to know there is no light to do only do things when there is no light.
This is used to cancel overbright when there is no light, since overbright only applies to lights.

src/engine/renderer/tr_image.cpp
```diff
@@ -2467,13 +2467,24 @@ static void R_CreateCurrentRenderImage()
 
 	imageParams_t imageParams = {};
 	imageParams.bits = IF_NOPICMIP;
+
+	bool bestWithHighPrecisionFrameBuffer = !r_mapClampOverBright.Get();
```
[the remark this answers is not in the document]
illwieckz: No, the legacy code using clamped overbright doesn't require doesn't do any division/multiplication roundtrip that would lose precision.

On the contrary the non-legacy code (when legacy clamped code is disabled) does some division/multiplication roundtrip and then would lose precision with 8-bit framebuffers.

src/engine/renderer/glsl_source/lightMapping_fp.glsl
illwieckz: Nice catch. I now do the division once at map load time instead, and replaced all the GLSL divisions by a multiplication.

It is assumed most GPUs do not know how to do jumps and then never skip a computation in a test but just discard the result if the test is false, so `something *= a` with `a = 1` is better than `if ( a != 0 ) { something *= a; }` with `a = 0` for when there is nothing to multiply, as we avoid the test instruction in all cases and we cannot avoid the multiplication in all cases.

illwieckz:
> Have you tested with different values of `r_lightMode`? I tried fullbright mode with clamping off and it looks really bad.

Ah right, the fix for it was lost in my multiple rewrites. This is now fixed.

src/engine/renderer/tr_image.cpp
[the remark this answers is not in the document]
slipher: As I understand, the clamp overbright cvar and the clamp overbright worldspawn key are two inputs into the decision of whether clamping is enabled. This final decision is stored in `tr.mapClampOverbright`. And the code should always act according to the final decision, not one or the other of the two inputs.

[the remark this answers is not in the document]
illwieckz: Oh lol I misread your comment!

I thought you were asking to do (without `!`):

```
bool bestWithHighPrecisionFrameBuffer = r_mapClampOverBright.Get();
```

🤣️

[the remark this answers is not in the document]
illwieckz: Now fixed.

src/engine/renderer/glsl_source/lightMapping_fp.glsl
slipher: > It is assumed most GPUs do not know how to do jumps and then never skip a computation in a test but just discard the result if the test is false

My concept is a bit different. I think they can jump, but all threads must execute the same code. With `if (x) { foo(); } else { bar(); }` if `x` is true at least once and false at least once, all threads have to execute both branches. But if `x` is false in all threads it can jump directly to `bar()`.

I agree that `x *= 1;` may very well be faster than `if ( false ) { x *= 1; }`. But my contention is that `if ( false ) { x *= 1; }` is likely faster than `if ( true ) { x *= 1; }`. (Here by `false`, `true` and `1` I mean expressions derived from uniforms that always have the same value, rather than compile-time constants.)

illwieckz: Hmm, maybe, but, that complexifies the code for the reader for a benefit that may be very minor. Maybe @VReaperV has an opinion on it?

illwieckz:
For `r_mapClampOverbright`, I think we may just name it `r_forceLegacyMapClampOverBright` and disable the clamping by default. I can also name the map variable `legacyMapClampOverBright`, as it is not expected to be used by non-legacy maps and will be ignored if used in non-legacy map.

The map variable would tell the renderer to use the legacy code path as a workaround for maps that has may look better with the old buggy code because they were tested with the old buggy code, and the cvar would be usable by the user to test a map with the old buggy code to test the workaround before setting the variable in the map.

Actually, I now don't see good reasons to not disable clamping by default in all maps. Once people start to play with not-clamped overbright, any legacy map with clamped overbright looks broken, much more than the quirks that may be revealed by not clamping the lights.

We were not seeing the bugs because we never seen a map properly rendered before, but once we see a map being properly renderer, the eyes then see the bugs on _almost every surface_ when clamping is enabled. We can't unseen those light bugs once seen, and they are everywhere.

Configuring a map to force the legacy code would require such map to be strongly broken, and for now the only maps that are that broken are maps from another game that used a game using different renderer default values.

Also a `r_forceLegacyMapClampOverBright` cvar would be useful when debugging regressions from Tremulous. For example it looks like our fog is wrong, so if someone wants to debug the fog and have a fair comparison with Tremulous, that one may want to work with a similarly buggy renderer to not be fooled by our fix. For example if a fog color was clamped, we should test the new implementation first with clamping on to test for visual parity, because testing for visual parity with clamping off on our end may lead us to re-implement clamping in fog code itself.

src/engine/renderer/glsl_source/lightMapping_fp.glsl
VReaperV: > Here by false, true and 1 I mean expressions derived from uniforms that always have the same value

If it's based only on uniforms and constant expressions (which it is unless I'm missing something), then it will be statically uniform so there wouldn't be any divergence and the unused code path will likely be eliminated.

Also, keep in mind that a compiler might optimise something like `if( condition > 0.0f ) { value *= condition }` into `value *= condition > 0.0f ? condition : 1` anyway.

At the end of the day though, trying to optimise a single vec4 multiplication that is done once per fragment is a waste of time. (If you really want to know which one will be faster you'd need to profile both and it would be hw/driver specific and you likely won't see any significant difference, and it would just be a giant waste of time)

illwieckz:
So, the cvar to enable old clamped code base is `r_forceLegacyMapOverBrightClamping`, it overrides any map option. It is only meant to be a diagnostic tool to have a fair comparison when comparing renders with engines having old clamped code (be backward compatible on bugs when comparing engines).

The map related option is `forceLegacyMapOverBrightClamping` and can be used as a per-map option to tell the renderer to use that old code path as a workaround. I'm not sure this variable will be kept forever, we may remove it in the future if it turns out it is never needed.

The complete overbright is now enabled by default for all maps. One can't unseen the buggy clamping after having played on maps without buggy clamping.

The `r_mapOverBrightBits` cvar name is kept this way for backward compatibility with legacy game configs when loading third-party game assets when evaluating the engine for running third-party games or loading third-party game data.

I now consider this PR totally ready (after I squash some fixup commits).

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
illwieckz: I don't see that place as a good place for such comment, the double clamping would not be a problem if the first one was not too low in a buggy way.

I added a comment in `tr_bsp.cpp` instead in the `R_ColorShiftLightingBytes` function.

slipher: Oh I was confused and thought that u_LightFactor would be less than 1. Since it is always >= 1 the first clamping has no effect and should be removed.

illwieckz: This one is the one actually multiplying the light (and cancelling the cancelation on non-light).


The first clamp in this source file is now removed.

src/engine/renderer/gl_shader.cpp
```diff
@@ -1615,6 +1615,7 @@ GLShader_generic::GLShader_generic( GLShaderManager *manager ) :
 	u_ModelMatrix( this ),
 	u_ProjectionMatrixTranspose( this ),
 	u_ModelViewProjectionMatrix( this ),
+	u_LightDivisor( this ),
```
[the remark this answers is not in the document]
illwieckz: Done

src/engine/renderer/tr_scene.cpp
```diff
@@ -429,6 +429,9 @@ void RE_AddDynamicLightToSceneET( const vec3_t org, float radius, float intensit
 	light->l.color[ 1 ] = g;
 	light->l.color[ 2 ] = b;
 
+	// Cancel overBright on dynamic lights.
+	VectorScale( light->l.color, 1.0f / tr.mapLightFactor, light->l.color );
```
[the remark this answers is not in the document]
illwieckz: Done

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
```diff
@@ -38,7 +38,7 @@ void	main()
 	// calculate the screen texcoord in the 0.0 to 1.0 range
 	vec2 st = gl_FragCoord.st / r_FBufSize;
 
-	vec4 color = clamp(texture2D(u_CurrentMap, st), 0.0, 1.0);
+	vec4 color = texture2D(u_CurrentMap, st);
 
 	color.rgb *= u_LightFactor;
```
slipher, line 44: I guess the second clamp should now be 4-dimensional instead of only rgb.

slipher (commented):
LGTM

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
illwieckz: Right! 😄️

src/engine/renderer/glsl_source/lightMapping_fp.glsl
illwieckz: Thank you for the explanations! ❤️

sweet235:
This branch is addictive. Since I discovered it a few days ago, I have spent hours just looking at maps.

PR #12, opened 2017-03-21
Remove the need for tr.charsetImage

DefaultUser:
Alternative approach to https://github.com/DaemonEngine/Daemon/pull/11
Difference is that in this approach the console font bitmap is reused for the bone names. Not as clean as https://github.com/DaemonEngine/Daemon/pull/11

PR #1052, opened 2024-03-03
WIP Optimisation: Use all texture units, bindless texture support

VReaperV:
Makes the engine use all texture units and rather than rebinding textures each time sets the correct uniform. Shaders have been modified to return the location for the correct uniform (this is needed because normal and bindless textures use different functions, so making it otherwise would be too convoluted).

Adds support for bindless textures. Bindless textures control texture residency instead of binding them.

Both approaches use texture priority which depends on the textures overall usage and usage during last frame.
Texture priority right now is just calculated as frame bind percentage (amount of times this texture was bound during the last frame / total texture binds last frame) * 0.5 + total bind percentage (amount of times this texture was bound in total / total texture binds) * 1.5.

If there's no available texture units ~~or bindless texture is not made resident~~ then it will replace the lowest priority texture unit ~~or make the lowest priority texture non-resident~~.
Currently disabled for bindless as making whatever texture was last added non-resident has better performance than sorting them every time. Another planned change might get rid of these priorities anyway.

This currently doesn't do anything for bound textures and even slightly lowers their performance when there's a lot of texture bindings. However, using bindless textures gives a slight performance improvement instead. (that is why this pr is draft)
I'll likely have to retain some of the old functionality for this reason. It is possible that using a UBO would still improve performance even without bindless textures, but that requires GLSL 4.00 so still won't be supported by all devices.

### DIFF
--- a/src.cmake
+++ b/src.cmake
@@ -102,6 +102,8 @@ set(RENDERERLIST
     ${ENGINE_DIR}/renderer/tr_font.cpp
     ${ENGINE_DIR}/renderer/InternalImage.cpp
     ${ENGINE_DIR}/renderer/InternalImage.h
+    ${ENGINE_DIR}/renderer/TextureManager.cpp
+    ${ENGINE_DIR}/renderer/TextureManager.h
     ${ENGINE_DIR}/renderer/tr_image.cpp
     ${ENGINE_DIR}/renderer/tr_image.h
     ${ENGINE_DIR}/renderer/tr_image_crn.cpp
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -1,0 +1,1933 @@
+﻿/*
+===========================================================================
+
+Daemon BSD Source Code
+Copyright (c) 2024 Daemon Developers
+All rights reserved.
+
+This file is part of the Daemon BSD Source Code (Daemon Source Code).
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+	* Redistributions of source code must retain the above copyright
+	  notice, this list of conditions and the following disclaimer.
+	* Redistributions in binary form must reproduce the above copyright
+	  notice, this list of conditions and the following disclaimer in the
+	  documentation and/or other materials provided with the distribution.
+	* Neither the name of the Daemon developers nor the
+	  names of its contributors may be used to endorse or promote products
+	  derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL DAEMON DEVELOPERS BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+===========================================================================
+*/
+// Material.cpp
+
+#include "Material.h"
+
+#include "tr_local.h"
+
+GLSSBO materialsSSBO( "materials", 0 );
+GLIndirectBuffer commandBuffer( "drawCommands" );
+MaterialSystem materialSystem;
+
+static void ComputeDynamics( shaderStage_t* pStage ) {
+	// TODO: Move color and texMatrices stuff to a compute shader
+	switch ( pStage->rgbGen ) {
+		case colorGen_t::CGEN_IDENTITY:
+		case colorGen_t::CGEN_ONE_MINUS_VERTEX:
+		default:
+		case colorGen_t::CGEN_IDENTITY_LIGHTING:
+			/* Historically CGEN_IDENTITY_LIGHTING was done this way:
+
+			  tess.svars.color = Color::White * tr.identityLight;
+
+			But tr.identityLight is always 1.0f in Dæmon engine
+			as the as the overbright bit implementation is fully
+			software. */
+		case colorGen_t::CGEN_VERTEX:
+		case colorGen_t::CGEN_CONST:
+		case colorGen_t::CGEN_ENTITY:
+		case colorGen_t::CGEN_ONE_MINUS_ENTITY:
+		{
+			// TODO: Move this to some entity buffer once this is extended past BSP surfaces
+			if ( backEnd.currentEntity ) {
+				//
+			} else {
+				//
+			}
+			pStage->colorDynamic = false;
+
+			break;
+		}
+
+		case colorGen_t::CGEN_WAVEFORM:
+		case colorGen_t::CGEN_CUSTOM_RGB:
+		case colorGen_t::CGEN_CUSTOM_RGBs:
+		{
+			pStage->colorDynamic = true;
+			break;
+		}
+	}
+
+	switch ( pStage->alphaGen ) {
+		default:
+		case alphaGen_t::AGEN_IDENTITY:
+		case alphaGen_t::AGEN_ONE_MINUS_VERTEX:
+		case alphaGen_t::AGEN_VERTEX:
+		case alphaGen_t::AGEN_CONST: {
+		case alphaGen_t::AGEN_ENTITY:
+		case alphaGen_t::AGEN_ONE_MINUS_ENTITY:
+			// TODO: Move this to some entity buffer once this is extended past BSP surfaces
+			/* if ( backEnd.currentEntity ) {
+			} else {
+			} */
+			pStage->colorDynamic = false;
+			break;
+		}
+
+		case alphaGen_t::AGEN_WAVEFORM:
+		case alphaGen_t::AGEN_CUSTOM:
+		{
+			pStage->colorDynamic = true;
+			break;
+		}
+	}
+
+	for ( textureBundle_t& bundle : pStage->bundle ) {
+		for ( int j = 0; j < bundle.numTexMods; j++ ) {
+			switch ( bundle.texMods[j].type ) {
+				case texMod_t::TMOD_NONE:
+				case texMod_t::TMOD_SCALE:
+				case texMod_t::TMOD_TRANSFORM:
+					break;
+
+				case texMod_t::TMOD_TURBULENT:
+				case texMod_t::TMOD_ENTITY_TRANSLATE:
+				case texMod_t::TMOD_SCROLL:
+				{
+					pStage->texMatricesDynamic = true;
+					break;
+				}
+
+				case texMod_t::TMOD_STRETCH:
+				{
+					if( bundle.texMods->wave.func != genFunc_t::GF_NONE ) {
+						pStage->texMatricesDynamic = true;
+					}
+					break;
+				}
+
+				case texMod_t::TMOD_ROTATE:
+				{
+					pStage->texMatricesDynamic = true;
+					break;
+				}
+
+				case texMod_t::TMOD_SCROLL2:
+				case texMod_t::TMOD_SCALE2:
+				case texMod_t::TMOD_CENTERSCALE:
+				case texMod_t::TMOD_SHEAR:
+				{
+					if ( bundle.texMods[j].sExp.active || bundle.texMods[j].tExp.active ) {
+						pStage->texMatricesDynamic = true;
+					}
+					break;
+				}
+
+				case texMod_t::TMOD_ROTATE2:
+				{
+					if( bundle.texMods[j].rExp.active ) {
+						pStage->texMatricesDynamic = true;
+					}
+					break;
+				}
+
+				default:
+					break;
+			}
+		}
+	}
+
+	// TODO: Move this to a different buffer?
+	for ( const textureBundle_t& bundle : pStage->bundle ) {
+		if ( bundle.isVideoMap || bundle.numImages > 1 ) {
+			pStage->texturesDynamic = true;
+			break;
+		}
+	}
+
+	// Can we move this to a compute shader too?
+	// Doesn't seem to be used much if at all, so probably not worth the effort to do that
+	pStage->dynamic = pStage->dynamic || pStage->ifExp.active;
+	pStage->dynamic = pStage->dynamic || pStage->alphaExp.active || pStage->alphaTestExp.active;
+	pStage->dynamic = pStage->dynamic || pStage->rgbExp.active || pStage->redExp.active || pStage->greenExp.active || pStage->blueExp.active;
+	pStage->dynamic = pStage->dynamic || pStage->deformMagnitudeExp.active;
+	pStage->dynamic = pStage->dynamic || pStage->depthScaleExp.active || pStage->etaExp.active || pStage->etaDeltaExp.active
+		|| pStage->fogDensityExp.active || pStage->fresnelBiasExp.active || pStage->fresnelPowerExp.active
+		|| pStage->fresnelScaleExp.active || pStage->normalIntensityExp.active || pStage->refractionIndexExp.active;
+
+	pStage->dynamic = pStage->dynamic || pStage->colorDynamic || pStage->texMatricesDynamic || pStage->texturesDynamic;
+}
+
+static image_t* GetLightMap( drawSurf_t* drawSurf ) {
+	if ( static_cast<size_t>( drawSurf->lightmapNum() ) < tr.lightmaps.size() ) {
+		return tr.lightmaps[drawSurf->lightmapNum()];
+	} else {
+		return tr.whiteImage;
+	}
+}
+
+static image_t* GetDeluxeMap( drawSurf_t* drawSurf ) {
+	if ( static_cast<size_t>( drawSurf->lightmapNum() ) < tr.deluxemaps.size() ) {
+		return tr.deluxemaps[drawSurf->lightmapNum()];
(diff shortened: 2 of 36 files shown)


illwieckz:
I tried to run it but I got:

```
Warn: Compile log:
0:13(8): error: qualifier `bindless_sampler` requires ARB_bindless_texture
Warn: Unhandled exception (15ShaderException): Couldn't compile fragment shader: generic2D 
```

VReaperV:
I assume the extensions supported by your machine include ARB_bindless_texture? Maybe addExtension() is not working properly...

illwieckz:
> I assume the extensions supported by your machine include ARB_bindless_texture? Maybe addExtension() is not working properly...

It looks like `addExtension()` is just a function to set define if the extension is loaded, but you need to load the extension first (in `sys/sdl_climp.cpp`). Also the code was misleading because it was assuming GLSL 4.00 included `ARB_bindless_texture`, which doesn't seem to be true.

Our engine require OpenGL 3.1 Core by default and then we cherry-pick extension, but sometime some drivers still provide OpenGL 4.5 and all extensions, I guess this is why it worked on your end, and by luck it may also have enabled the define because your GLSL version is high enough, but not because the test was correct.

Here is a patch that makes your code run on my end. Notice I added a `r_arb_bindless_texture` cvar to make easy to disable the extension for debugging purpose, and I modified `addExtension()` to receive `-1` when an extension is not part of any GLSL version so we can't assume any version can bring it.

```
diff --git a/src/engine/renderer/gl_shader.cpp b/src/engine/renderer/gl_shader.cpp
index a81c4f049..3234113ab 100644
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -379,7 +379,7 @@ static void addExtension( std::string &str, int enabled, int minGlslVersion,
 			  int supported, const char *name ) {
 	if( !enabled ) {
 		// extension disabled by user
-	} else if( glConfig2.shadingLanguageVersion >= minGlslVersion ) {
+	} else if( minGlslVersion != -1 && glConfig2.shadingLanguageVersion >= minGlslVersion ) {
 		// the extension is available in the core language
 		str += Str::Format( "#define HAVE_%s 1\n", name );
 	} else if( supported ) {
@@ -421,7 +421,7 @@ static std::string GenVersionDeclaration() {
 		      GLEW_ARB_gpu_shader5, "ARB_gpu_shader5" );
 	addExtension( str, r_arb_uniform_buffer_object->integer, 140,
 		      GLEW_ARB_uniform_buffer_object, "ARB_uniform_buffer_object" );
-	addExtension( str, glConfig2.bindlessTexturesAvailable, 400,
+	addExtension( str, glConfig2.bindlessTexturesAvailable, -1,
 		      GLEW_ARB_bindless_texture, "ARB_bindless_texture" );
 
 	return str;
diff --git a/src/engine/renderer/tr_init.cpp b/src/engine/renderer/tr_init.cpp
index 288ad7af4..7ae12c846 100644
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -100,6 +100,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_arb_uniform_buffer_object;
 	cvar_t      *r_arb_texture_gather;
 	cvar_t      *r_arb_gpu_shader5;
+	cvar_t      *r_arb_bindless_texture;
 
 	cvar_t      *r_checkGLErrors;
 	cvar_t      *r_logFile;
@@ -1094,6 +1095,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_arb_uniform_buffer_object = Cvar_Get( "r_arb_uniform_buffer_object", "1", CVAR_CHEAT | CVAR_LATCH );
 		r_arb_texture_gather = Cvar_Get( "r_arb_texture_gather", "1", CVAR_CHEAT | CVAR_LATCH );
 		r_arb_gpu_shader5 = Cvar_Get( "r_arb_gpu_shader5", "1", CVAR_CHEAT | CVAR_LATCH );
+		r_arb_bindless_texture = Cvar_Get( "r_arb_bindless_texture", "1", CVAR_CHEAT | CVAR_LATCH );
 
 		r_picMip = Cvar_Get( "r_picMip", "0",  CVAR_LATCH | CVAR_ARCHIVE );
 		r_imageMaxDimension = Cvar_Get( "r_imageMaxDimension", "0",  CVAR_LATCH | CVAR_ARCHIVE );
index cb5257637..ac269fd85 100644
@@ -2992,6 +2992,7 @@ enum class deluxeMode_t { NONE, GRID, MAP };
 	extern cvar_t *r_arb_uniform_buffer_object;
 	extern cvar_t *r_arb_texture_gather;
 	extern cvar_t *r_arb_gpu_shader5;
+	extern cvar_t *r_arb_bindless_texture;
 
 	extern cvar_t *r_nobind; // turns off binding to appropriate textures
 	extern cvar_t *r_singleShader; // make most world faces use default shader
diff --git a/src/engine/sys/sdl_glimp.cpp b/src/engine/sys/sdl_glimp.cpp
index 7d555c845..c4c322870 100644
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1933,6 +1933,9 @@ static void GLimp_InitExtensions()
 	// made required in OpenGL 3.2
 	glConfig2.syncAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_sync, r_arb_sync->value );
 
+	// not required by any OpenGL version
+	glConfig2.bindlessTexturesAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_bindless_texture, r_arb_bindless_texture->value );
+
 	GL_CheckErrors();
 }
 ```

**Comment by illwieckz** [comment]
> Also, how does it work performance-wise on your end, compared to not using this change?

map|viewpos|legacy|bindless
-|-|-|-
plat23|`0 2688 192 -90 13`|250|380
plat23|`1890 1920 0 0 -10`|160|220

This is an AMD Radeon PRO W7600.

Though, with your code but `r_arb_bindless_texture 0`, many 2D textures are now broken.

**Comment by illwieckz** [comment]
I get the same 2D texture glitches when disabling bindless texture with mesa environment variable (in case you distrust the cvar):

```sh
MESA_EXTENSION_OVERRIDE='-GL_ARB_texture_bindless'
```

VReaperV:
That is a nice FPS boost!

> Though, with your code but r_arb_bindless_texture 0, many 2D textures are now broken.

Can you send some example screenshots (here or on the IRC)? I thought I made all of those work, but maybe I missed something. I'll need to make some changes to resolve conflicts with current master, but other than that it should work(tm). (also, make sure you set the cvar from terminal, I think otherwise it's gonna take an incorrect code path)

> I get the same 2D texture glitches when disabling bindless texture with a Mesa environment variable (in case you distrust the cvar):

Hmm, that's weird...

VReaperV:
I've added your patch and updated this branch to current master.

illwieckz:
Levelshot blended with main menu background, garbage progression bar color:

20240305-180423-000 unvanquished

Missing minimap black background:

20240305-180432-000 unvanquished

Missing console background:

20240305-180434-000 unvanquished

I get the exact same bugs, whatever I disable bindless texture in the Mesa driver directly, or in engine using the cvar.

VReaperV:
Oh, that's new.

VReaperV:
Weirdly enough, it's working on my end.

illwieckz:
Forget about what I said about the Mesa environment variable, I did a mistake writing it, so I only tested the cvar.

Disabling the extension get me a segfault, here is the proper writing of the Mesa variable (I also edited my previous comment):

```sh
MESA_EXTENSION_OVERRIDE='-GL_ARB_bindless_texture'
```

For some unknown reason the game still detect it:

```
...using GL_ARB_bindless_texture
[…]
Using OpenGL extensions: GL_ARB_half_float_pixel GL_ARB_texture_float GL_EXT_gpu_shader4 GL_EXT_texture_integer GL_ARB_texture_rg GL_ARB_bindless_texture GL_ARB_texture_gather GL_EXT_texture_compression_s3tc GL_ARB_texture_compression_rgtc GL_EXT_texture_filter_anisotropic GL_ARB_half_float_vertex GL_ARB_framebuffer_object GL_ARB_get_program_binary GL_ARB_buffer_storage GL_ARB_uniform_buffer_object GL_ARB_map_buffer_range GL_ARB_sync GL_ARB_bindless_texture
```

illwieckz:
But on the other hand:

```sh
$ glxinfo | grep -i bindless
    GL_ARB_arrays_of_arrays, GL_ARB_base_instance, GL_ARB_bindless_texture, 
    GL_ARB_arrays_of_arrays, GL_ARB_base_instance, GL_ARB_bindless_texture, 
```

```sh
$ MESA_EXTENSION_OVERRIDE='-GL_ARB_bindless_texture' glxinfo | grep -i bindless
(nothing)
```

illwieckz:
I also did a mistake with my patch:

```diff
diff --git a/src/engine/sys/sdl_glimp.cpp b/src/engine/sys/sdl_glimp.cpp
index c4c322870..8ee521c39 100644
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1784,7 +1784,7 @@ static void GLimp_InitExtensions()
 	// made required in OpenGL 3.0
 	glConfig2.textureRGAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_texture_rg, r_ext_texture_rg->value );
 
-	glConfig2.bindlessTexturesAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_bindless_texture, r_ext_bindless_textures->value );
+	glConfig2.bindlessTexturesAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, ARB_bindless_texture, r_ext_bindless_textures->value );
 
 	{
 		/* GT218-based GPU with Nvidia 340.108 driver advertising
```

Now I get as expected:

```
Missing OpenGL extensions: GL_ARB_bindless_texture
````

But I still get a segfault.

**Comment by illwieckz** [comment]
Well, this is wrong:

```
Using OpenGL extensions: GL_ARB_half_float_pixel GL_ARB_texture_float GL_EXT_gpu_shader4 GL_EXT_texture_integer GL_ARB_texture_rg GL_ARB_texture_gather GL_EXT_texture_compression_s3tc GL_ARB_texture_compression_rgtc GL_EXT_texture_filter_anisotropic GL_ARB_half_float_vertex GL_ARB_framebuffer_object GL_ARB_get_program_binary GL_ARB_buffer_storage GL_ARB_uniform_buffer_object GL_ARB_map_buffer_range GL_ARB_sync GL_ARB_bindless_texture 
Missing OpenGL extensions: GL_ARB_bindless_texture 
```

It is said to be both used and missing… 🤔️

**Comment by VReaperV** [comment]
I've removed the `r_ext_bindless_textures` since it is, after all, `ARB_bindless_texture`.

**Comment by VReaperV** [comment]
> It is said to be both used and missing… 🤔️

I think it's because there were 2 cvars for it. Can you try the updated version?

**Comment by illwieckz** [comment]
Right, we were loading twice the extension, now you miss that (the extension is not implied by core profile):

```
diff --git a/src/engine/sys/sdl_glimp.cpp b/src/engine/sys/sdl_glimp.cpp
index 03ded78b2..1b4bb29d0 100644
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1932,7 +1932,7 @@ static void GLimp_InitExtensions()
        glConfig2.syncAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_sync, r_arb_sync->value );
 
        // not required by any OpenGL version
-       glConfig2.bindlessTexturesAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_bindless_texture, r_arb_bindless_texture->value );
+       glConfig2.bindlessTexturesAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, ARB_bindless_texture, r_arb_bindless_texture->value );
 
        GL_CheckErrors();
 }
```

**Comment by illwieckz** [comment]
Once the latest patch is applied, the game runs while the extension is disabled in the driver, but reproduce the errors seen in the above screenshots.

**Comment by illwieckz** [comment]
Unfortunately I don't reproduce the bug with Mesa llvmpipe, I reproduce it with Mesa radeonsi…

**Comment by illwieckz** [comment]
We may have a shader cache issue, deleting the `glsl/` folder fixes the bug for me.

**Comment by VReaperV** [comment]
Hmm, that might indeed be the case.

**Comment by illwieckz** [comment]
Not your fault:

- https://github.com/DaemonEngine/Daemon/pull/1054

**Comment by illwieckz** [comment]
Now that I delete my shader cache between every test run to avodi running wrong GLSL code because of shader cache issue, the bindless branch shows me a huge performance drop instead.

Faster is master branch, then this branch with bindless disabled, then this branch with bindless enabled.

preset|resolution|map|viewpos|master|bindless off|bindless on
-|-|-|-|-|-|-
ultra|`2560×1440`|plat23|`0 2688 192 -90 13`|610fps|480fps|380fps
ultra|`2560×1440`|plat23|`1890 1920 0 0 -10`|490fps|350fps|240fps

**Comment by illwieckz** [comment]
It doesn't crash anymore but the motion blur is now garbage:

unvanquished_2024-03-06_190439_000

**Comment by VReaperV** [comment]
Yep, I just checked and on my end it works fine.

**Comment by illwieckz** [comment]
If you can rebase over current `master` to get this:

- https://github.com/DaemonEngine/Daemon/pull/1054

It will make testing your branch easier! 😉️

**Comment by slipher** [comment]
> For some reason this also requires rebuilding cgame, so I might have to rebase this on 0.55/sync (second reason why this is draft).

You can fix that by using `glconfig2_t` instead of `glconfig_t`. 

I guess we should be moving junk from glconfig2 to glconfig when releasing a major version, since the idea of that struct is to make info known to the cgame. No one has been using that stuff though

**Comment by VReaperV** [comment]
That fixed it, thanks!

**Comment by illwieckz** [comment]
I confirm I still get the motion blur garbage, and since #1054, I'm now sure it's not a cache issue. I also managed to reproduce it by deleting the GLSL cache anyway.

**Comment by illwieckz** [comment]
@cu-kai reported on IRC a similar bug with smoke effect:

4BL

**Comment by illwieckz** [comment]
bmorel reported this crash on IRC:

```
#0  0x0000000000000000 in ?? ()
#1  0x000055555568f4d4 in Texture::MakeNonResident (this=0x555557176d00) at daemon/src/engine/renderer/TextureManager.cpp:66
#2  Texture::~Texture (this=0x555557176d00, __in_chrg=<optimized out>) at daemon/src/engine/renderer/TextureManager.cpp:44
#3  Texture::~Texture (this=this@entry=0x555557176d00, __in_chrg=<optimized out>)
    at daemon/src/engine/renderer/TextureManager.cpp:42
#4  0x0000555555694867 in R_ShutdownImages () at daemon/src/engine/renderer/tr_image.cpp:3101
#5  0x00005555556aa4a4 in RE_Shutdown (destroyWindow=<optimized out>) at daemon/src/engine/renderer/tr_init.cpp:1549
#6  0x0000555555601e00 in CL_ShutdownAll () at daemon/src/engine/client/cl_main.cpp:691
#7  0x00005555555faa55 in CL_DownloadsComplete () at daemon/src/engine/client/cl_download.cpp:104
#8  0x000055555560ae36 in CL_ParseGamestate (msg=msg@entry=0x7ffffffedc00) at daemon/src/engine/client/cl_parse.cpp:461
#9  0x000055555560b068 in CL_ParseServerMessage (msg=msg@entry=0x7ffffffedc00) at daemon/src/engine/client/cl_parse.cpp:571
#10 0x0000555555604ccc in CL_PacketEvent (from=..., msg=msg@entry=0x7ffffffedc00) at daemon/src/engine/client/cl_main.cpp:1966
#11 0x00005555555a14a3 in Com_EventLoop () at daemon/src/engine/qcommon/common.cpp:430
#12 0x00005555555a1c9a in Com_Frame () at daemon/src/engine/qcommon/common.cpp:1016
#13 0x00005555556f7309 in Application::ClientApplication::Frame (this=0x555556292100 <Application::GetApp()::app>)
    at daemon/src/engine/client/ClientApplication.cpp:96
#14 0x000055555559c795 in main (argc=<optimized out>, argv=<optimized out>) at daemon/src/engine/framework/System.cpp:755
(gdb) up
#1  0x000055555568f4d4 in Texture::MakeNonResident (this=0x555557176d00) at daemon/src/engine/renderer/TextureManager.cpp:66
66              glMakeTextureHandleNonResidentARB( bindlessTextureHandle );
```

illwieckz:
I also reproduce the performance loss with the ATI OGLP proprietary driver (same GPU) when not doing bindless, and bindless doesn't work with OGLP (black screen).

driver|preset|map|viewpos|master|bindless off|bindless on
-|-|-|-|-|-|-
radeonsi|ultra|plat23|`0 2688 192 -90 13`|610fps|480fps|380fps
oglp|ultra|plat23|`0 2688 192 -90 13`|630fps|560fps|broken
radeonsi|ultra|plat23|`1890 1920 0 0 -10`|490fps|350fps|240fps
oglp|ultra|plat23|`1890 1920 0 0 -10`|630fps|530fps|broken

- GPU: AMD Radeon W7600 PRO
- OS: Ubuntu 23.10
- Mesa radeonsi: 23.2.1
- ATI OGLP: 23.40
- Resolution: `2560×1440`

OGLP is not recommended by AMD anymore but still shipped within latest AMDGPU-PRO driver pack, likely for some corporate need and contractual obligation. So the priority for this is low, but knowing the game cannot render at all on a publicly available driver is still a problem. We can workaround it by detecting the driver to disable bindless, but what I'm afraid of is that we would have to do an instensive testing on various drivers/hardware/OS combination.

VReaperV:
That just sounds like a broken driver.

illwieckz:
Yes, it is very likelya broken driver, but it means there are broken drivers distributed today.

illwieckz:
It looks like bindless textures are not supported on macOS on any GPU. Not only the maximum OpenGL version supported on macOS seems to be OpenGL 4.1, but bindless textures are part of an extension not being part of any core OpenGL version.

This page is a good database of various systems, drivers and gpus combination, allowing for example to search for "MacOS" systems on the “os” column: https://opengl.gpuinfo.org/

Directly looking for drivers/gpu combinations supporting bindless textures and filtering to only keep “MacOS” system in the list leaves the list empty: https://opengl.gpuinfo.org/listreports.php?extension=GL_ARB_bindless_texture

Alongside to the fact our engine has good backward compatibility support down to OpenGL 2.1, MacOS not supporting bindless textures even on latest driver version and latest Apple GPU like M1 and M2 devices shows that if at some point we manage to get some bindless texture working for the lucky people having compatible system, any alternative non-bindless code should not suffer any performance regression.

VReaperV:
Closing this because bindless textures will now be in #1105. I've made enough changes to them there that backporting them here is a pain in the ass and not worth it.

PR #1055, opened 2024-03-07
Fix portal scissor test

VReaperV:
Width and height in glScissor() must not be negative, so if surfRect has lower x2 or y2 than x1 or y1, then flip the scissor parameters. Tested on `test-portals-recursion` on viewpos -2223 639 65 91 3.

Fixes a GL_INVALID_VALUE at https://github.com/DaemonEngine/Daemon/blob/1872b92274fbd9533e622633c56a515a4ff899c0/src/engine/renderer/tr_backend.cpp#L359


DolceTriade (approved):
add comment please?

VReaperV:
Sure

DolceTriade (approved):

PR #1057, opened 2024-03-08
NUKE client-side vertex arrays

VReaperV:
Use VBOs/IBOs everywhere. Avoid unnecessary data transfers to the GPU. The ability to use client-side vertex arrays is also removed in 3.2+ core OpenGL.
`skipVBO` set by portals is never used because portal surfaces are skipped when building the world VBO, so their vbo/ibo is never set.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -907,7 +907,7 @@ static void RB_RenderDrawSurfaces( shaderSort_t fromSort, shaderSort_t toSort,
 				Tess_End();
 			}
 
-			Tess_Begin( Tess_StageIteratorGeneric, nullptr, shader, nullptr, false, false, lightmapNum, fogNum, bspSurface );
+			Tess_Begin( Tess_StageIteratorGeneric, nullptr, shader, nullptr, false, lightmapNum, fogNum, bspSurface );
 
 			oldShader = shader;
 			oldLightmapNum = lightmapNum;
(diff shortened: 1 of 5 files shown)


DolceTriade:
What is the performance impact of this change? Is it dead code? Does everyone support vbos?

slipher:
What do you mean by client-side vertex arrays? It seems that this is only used by `SurfIsOffscreen` to convert a surface into triangles without actually drawing it.

VReaperV:
> What is the performance impact of this change? Is it dead code? Does everyone support vbos?

None, it is pretty much dead code (the one part of the code that was using it shouldn't be doing so), yes.

> Please add to commit message

Seems redundant, but sure I can add it.

VReaperV:
> What do you mean by client-side vertex arrays?

I mean the client-side vertex arrays, i. e. not VBOs.

> It seems that this is only used by SurfIsOffscreen to convert a surface into triangles without actually drawing it.

It works fine without the skipVBO part, and with or without it it should do pretty much the same thing i. e. essentially nothing.

slipher:
LGTM to the code itself. I don't think the commit/pull request messages do a good job of describing why it is right though.

The intention of the removed code is to avoid running some unnecessary code in Tess_SurfaceVBO in the case of processing a surface that we never intend to actually draw. The non-intending-to-draw case is with portal surfaces. But the `if ( !vbo || !ibo )` early return in TessSurfaceVBO always triggers anyway because R_CreateWorldVBO() avoids setting VBOs/IBOs for portal surfaces. So the 2nd early return with `tess.skipVBO` turns out to be dead code.

VReaperV:
I've updated the commit description. Another thing to add here is that core 3.2+ profile removes the ability to use client-side vertex arrays anwyay.

PR #1060, opened 2024-03-11
tr_shader: do not collapse stage with tcGen environment

illwieckz:
- tr_shader: do not collapse stage with tcGen environment

Fixes a bug (regression) where the color map has a different tcGen than the light map: they should not be collapsed

This is usually used to fake reflections, by applying a baked reflection as a texture but using `tcGen environment` the texture moves on the surface while the player moves. The lightmap itself should stick to the surface.

It may be possible to edit the lightMapping code to have two coordinate systems for color map and lightmap, but this extra complexity can be uselessly costly for every surfaces while keeping the fake reflections uncollapsed only affect those surfaces.

For maximum prevention this check is also done on addition map collapse.

With bug (see the train windows):

unvanquished_2024-03-11_161935_000

Without bug:

unvanquished_2024-03-11_162321_000

The regression was introduced by myself in #1046 (I missed a check to detect a stage as not collapsible).

- tr_shader: fix blendFunc detection in collapse code

Some index was wrong, I also deleted dead code.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -4806,19 +4806,13 @@ static void CollapseStages()
 		bool alphaGen_identity =
 			stages[ i ].alphaGen == alphaGen_t::AGEN_IDENTITY;
 
-		/*
-		bool blendFunc_none = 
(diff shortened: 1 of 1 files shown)


slipher:
LGTM

PR #1063, opened 2024-03-16
Detect some operations invalid for a non-main thread

### LINKED ISSUE #1061: Detect non-main thread attempting to use IPC
If we use threads in the gamelogic VMs, we should detect if a non-main thread attempts to use the IPC facility and terminate the program instead of allowing potentially mysterious corruption. It can be done in C++ [like this](https://stackoverflow.com/a/13288025).

The difficult part would be actually communicating any message about what the programmer has done wrong, since the only way to log an error message is using IPC :)

slipher:
Fixes #1061.

### DIFF
--- a/src/common/System.h
+++ b/src/common/System.h
@@ -69,6 +69,9 @@ void SleepFor(SteadyClock::duration time);
 // Results *within a single module* (engine/cgame/sgame) are monotonic.
 int Milliseconds();
 
+// For a DLL this means the thread starting from the VM's entry point, not the real main thread
+bool OnMainThread();
+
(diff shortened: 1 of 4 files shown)


src/common/System.h
```diff
@@ -69,6 +69,9 @@ void SleepFor(SteadyClock::duration time);
 // Results *within a single module* (engine/cgame/sgame) are monotonic.
 int Milliseconds();
 
+// For a DLL this means the thread startinvg from the VM's entry point, not the real main thread
```
illwieckz, line 72: Typo: `startinvg`

illwieckz (approved):
If it works, LGTM.

PR #1066, opened 2024-03-22
use explicit wording for reading cvar value

illwieckz:
hOw DiD wE mIsSeD tHaT fOr So lOnG? 🤔️

### DIFF
--- a/src/common/Cvar.h
+++ b/src/common/Cvar.h
@@ -119,9 +119,8 @@ namespace Cvar {
             Cvar(std::string name, std::string description, int flags, value_type defaultValue);
             Cvar(NoRegisterTag, std::string name, std::string description, int flags, value_type defaultValue);
 
-            //Outside code accesses the Cvar value by doing my_cvar.Get() or *my_cvar
+            //Outside code accesses the Cvar value by doing my_cvar.Get()
             T Get() const;
-            T operator*() const;
 
             //Outside code can also change the value but it won't be seen immediately after with a .Get()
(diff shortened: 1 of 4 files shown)


DolceTriade (commented):
I think this code works fine...

https://github.com/DaemonEngine/Daemon/blob/895cb2c071ca4c0ed69f42e77c147a1d70d3183d/src/common/Cvar.h#L123

Means that `*` is overloaded to call `Get()`

illwieckz:
> I think this code works fine...
> 
> https://github.com/DaemonEngine/Daemon/blob/895cb2c071ca4c0ed69f42e77c147a1d70d3183d/src/common/Cvar.h#L123
> 
> Means that `*` is overloaded to call `Get()`

Ah… Well, I prefer explicit code. 🤷‍♀️️

illwieckz:
I renamed the commit `renderer: use explicit wording for reading cvar value`.

slipher:
Should we nuke `operator*`?

illwieckz:
> Should we nuke `operator*`?

I was going to ask the same. I don't want to have to guess the behavior of `*ptr` knowing it may differ according to the object type. It seems to just add extra mental load only to save on characters and allowing some kind of obfsucated code golf…

So if someone nukes it, I will approve…

illwieckz:
I added a commit to NUKE the confusing `*cvar` syntax, it was used nowhere else.

slipher:
LGTM

PR #1070, opened 2024-04-06
Command system enhancements

slipher:
- Fix args for `exec` scripts
- When setting a cvar via the command with the same name as the cvar, concatenate all arguments instead of taking only the first
- Let command aliases be saved in autogen.cfg


src/engine/framework/BaseCommands.cpp
```diff
@@ -113,13 +113,13 @@ namespace Cmd {
...
                 const std::string& filename = args.Argv(filenameArg);
 
-                SetExecArgs(args, filenameArg + 1);
                 if (not ExecFile(filename, executeSilent)) {
                     if (not failSilent) {
                         Print("couldn't exec '%s'", filename.c_str());
                     }
                     return;
                 }
+                SetExecArgs(args, filenameArg + 1);
```
DolceTriade, line 122: This is super confusing.  At the very least this deserves a comment. Better yet, we should update the implementations of both of things to use `BufferCommandText` so it behaves like we'd expect the execution to behave.

src/engine/framework/BaseCommands.cpp
```diff
@@ -795,12 +795,9 @@ namespace Cmd {
 
     struct aliasRecord_t {
         std::string command;
-        int lastRun;
```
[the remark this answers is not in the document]
slipher: The "intent" was that the author imagined that commands would be executed recursively instead of using a buffer (see my commit message for more info). The command system is Turing-complete in various ways so no avoiding infinite loops really

src/engine/framework/BaseCommands.cpp
```diff
@@ -902,7 +928,8 @@ namespace Cmd {
                 return {};
             }
     };
-    static AliasCmd AliasCmdRegistration;
+    static AliasCmd aliasCmdRegistration("alias", "create or view an alias", false);
+    static AliasCmd aliasaCmdRegistration("aliasa", "create or archive an alias", true);
```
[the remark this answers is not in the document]
slipher: We don't have to prefer one or the other method as they can coexist peacefully. Aliases made with the existing command won't be written to autogen. The advantage of aliasa is that you could save an alias without having to switch to a text editor.

src/engine/framework/BaseCommands.cpp
slipher: Commented. Not sure what you mean by the last sentence. I think just using Cvar::SetValue instead of commands would also work here if that seems better

DolceTriade: I mean using `Cmd::BufferCommandText` directly instead of ExecuteAfter, which is the root of this confusing behavior. If you use the former method, it will queue up commands as expected.

Would be interesting to see that. However, using `Cvar::SetValue` is certainly more clear in terms of intent instead of relying on the command system to set the cvars.

slipher: Switched to Cvar::SetValue.

Using BufferCommandText in the `exec` implementation could mess up the ordering. If you did `exec myscript; echo Done`  then it would say `Done` before executing the commands in the script.

PR #1071, opened 2024-04-06
Remove redundant cullFace

VReaperV:
Remove some redundant code.

`cullFace` is never needed because the check against the current state is already done when comparing `culltype` to `glState.faceCulling` in GL_Cull().


illwieckz (approved):
LGTM.

PR #1072, opened 2024-04-06
Remove vertexSkinning and vertexAnimation from GLShader_generic2D

VReaperV:
UI, which is what this shader is used for, doesn't have bones (vertexSkinning). It also doesn't use .md3/.mdc (vertexAnimation is used to interpolate vertices in those formats).

I'm not entirely sure if vertexSprite is unused, so I've kept it there for now.


illwieckz (approved):
LGTM.

slipher:
There's a lot more stuff that should be removed from the generic2d shader. It was made by copy and pasting one for 3D rendering without deleting any of the 3D-specific stuff

VReaperV:
I'm aware, yeah. I haven't checked which other stuff can be removed there.

PR #1073, opened 2024-04-11
pbr: Don't force ST_PHYSICALMAP when parsing pbr maps

DolceTriade:
When diffuse and pbr maps are used in the same stage, forcing ST_PHYSICALMAP will prevent the lightmapping shader from running causing a black texture to be rendered since lightmapping only runs for diffuse stages.

ST_PHYSICALMAP is only useful when collapsing stages.


illwieckz (approved):
LGTM.

PR #1074, opened 2024-04-11
renderer: rewrite a bit the shader selection

illwieckz:
- use smaller stage struct
- prevent the engine to use the PBR shader when it is disabled (do not crash)
- improve the wording in various places

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -958,13 +958,13 @@ static void Render_lightMapping( shaderStage_t *pStage )
 
 	gl_lightMappingShader->SetGridDeluxeMapping( enableGridDeluxeMapping );
 
-	gl_lightMappingShader->SetHeightMapInNormalMap( pStage->isHeightMapInNormalMap );
+	gl_lightMappingShader->SetHeightMapInNormalMap( pStage->hasHeightMapInNormalMap );
 
 	gl_lightMappingShader->SetReliefMapping( pStage->enableReliefMapping );
 
 	gl_lightMappingShader->SetReflectiveSpecular( pStage->enableNormalMapping && tr.cubeHashTable != nullptr );
 
-	gl_lightMappingShader->SetPhysicalShading( pStage->isMaterialPhysical );
+	gl_lightMappingShader->SetPhysicalShading( pStage->enablePhysicalMapping );
 
 	gl_lightMappingShader->BindProgram( pStage->deformIndex );
 	// end choose right shader program ------------------------------
@@ -1054,7 +1054,7 @@ static void Render_lightMapping( shaderStage_t *pStage )
 		gl_lightMappingShader->SetUniform_ReliefOffsetBias( tess.surfaceShader->reliefOffsetBias );
 
 		// FIXME: if there is both, embedded heightmap in normalmap is used instead of standalone heightmap
-		if ( !pStage->isHeightMapInNormalMap )
+		if ( !pStage->hasHeightMapInNormalMap )
 		{
 			GL_BindToTMU( BIND_HEIGHTMAP, pStage->bundle[ TB_HEIGHTMAP ].image[ 0 ] );
 		}
@@ -1074,7 +1074,7 @@ static void Render_lightMapping( shaderStage_t *pStage )
 	}
 
 	// bind u_NormalMap
-	if ( !!r_normalMapping->integer || pStage->isHeightMapInNormalMap )
+	if ( !!r_normalMapping->integer || pStage->hasHeightMapInNormalMap )
 	{
 		GL_BindToTMU( BIND_NORMALMAP, pStage->bundle[ TB_NORMALMAP ].image[ 0 ] );
 	}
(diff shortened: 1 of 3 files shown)


illwieckz:
This is a commit that was sitting on my hard drive for a while, but it now also fixes a bug revealed when testing:

- https://github.com/DaemonEngine/Daemon/pull/1073

The bug was that the engine was trying to use the PBR shader varient on PBR surfaces when `r_physicalMap` was disabled.

src/engine/renderer/tr_shade.cpp
```diff
@@ -1140,7 +1140,7 @@ static void Render_lightMapping( shaderStage_t *pStage )
 	}
 
 	// bind u_NormalMap
-	if ( !!r_normalMapping->integer || pStage->isHeightMapInNormalMap )
+	if ( !!r_normalMapping->integer || pStage->hasHeightMapInNormalMap )
```
DolceTriade, line 1143: Shouldn't this just be `if ( r_normalMapping->integer )`?

After all, if normal mapping is disable, why bother setting stuff if `hasHeightMapInNormalMap` is true?

src/engine/renderer/tr_shader.cpp
```diff
@@ -5170,20 +5170,31 @@ static void CollapseStages()
...
-		stage->hasMaterialMap = stage->bundle[ TB_MATERIALMAP ].image[ 0 ] != nullptr;
-		stage->hasGlowMap = stage->bundle[ TB_GLOWMAP ].image[ 0 ] != nullptr;
-		stage->isMaterialPhysical = stage->collapseType == collapseType_t::COLLAPSE_PBR;
+		bool hasNormalMap = stage->bundle[ TB_NORMALMAP ].image[ 0 ] != nullptr;
+		bool hasHeightMap = stage->bundle[ TB_HEIGHTMAP ].image[ 0 ] != nullptr;
+		bool hasMaterialMap = stage->bundle[ TB_MATERIALMAP ].image[ 0 ] != nullptr;
+		bool hasGlowMap = stage->bundle[ TB_GLOWMAP ].image[ 0 ] != nullptr;
+
+		// Texture storage variants.
+		stage->hasHeightMapInNormalMap &= hasNormalMap;
```
DolceTriade, line 5179: let's write out the full condition instead of using fancy bit logic on booleans?

src/engine/renderer/tr_shader.cpp
```diff
@@ -5170,20 +5170,31 @@ static void CollapseStages()
...
+
+		stage->enableGlowMapping = hasGlowMap && r_glowMapping->integer;
+
+		if ( stage->collapseType == collapseType_t::COLLAPSE_PBR )
+		{
+			stage->enablePhysicalMapping = hasMaterialMap && r_physicalMapping->integer;
+		}
+		else
+		{
+			stage->enableSpecularMapping = hasMaterialMap && r_specularMapping->integer;
```
DolceTriade, line 5196: Explicitly set the other to false?

ie,

```

stage->enablePhysicalMapping = hasMaterialMap && r_physicalMapping->integer;
stage->enableSpecularMapping = false;
```

src/engine/renderer/tr_shader.cpp
```diff
@@ -5197,16 +5208,17 @@ static void CollapseStages()
 
 		// Finally disable useless heightMapInNormalMap if both normal and relief mapping are disabled.
 		// see https://github.com/DaemonEngine/Daemon/issues/376
-		stage->isHeightMapInNormalMap = stage->isHeightMapInNormalMap && ( stage->enableNormalMapping || stage->enableReliefMapping );
+		stage->hasHeightMapInNormalMap &= ( stage->enableNormalMapping || stage->enableReliefMapping );
```
DolceTriade, line 5211: same thing about writing out the full condition

src/engine/renderer/tr_shader.cpp
```diff
@@ -5197,16 +5208,17 @@ static void CollapseStages()
...
 
 		// Bind fallback textures if required.
-		if ( !stage->enableNormalMapping && !( stage->enableReliefMapping && stage->isHeightMapInNormalMap) )
+		if ( !stage->enableNormalMapping && !( stage->enableReliefMapping && stage->hasHeightMapInNormalMap) )
 		{
 			stage->bundle[ TB_NORMALMAP ].image[ 0 ] = tr.flatImage;
 		}
 
-		if ( !stage->enablePhysicalMapping && !stage->enableSpecularMapping )
+		if ( !stage->enablePhysicalMapping && !stage->enableSpecularMapping && r_specularMapping->integer )
```
DolceTriade, line 5219: also have to check r_physicalMapping?

```
if ( !stage->enablePhysicalMapping && !stage->enableSpecularMapping && ( r_specularMapping->integer || r_physicalMapping->integer ) )
```

src/engine/renderer/tr_shade.cpp
illwieckz: > After all, if normal mapping is disable, why bother setting stuff if `hasHeightMapInNormalMap` is true?

We can load the normalmap+heightmap combined file to process the parallax, even if normal mapping is disabled but height mapping is enabled.

src/engine/renderer/tr_shader.cpp
illwieckz: Well, that's a matter of taste, butI done full condition again.

src/engine/renderer/tr_shader.cpp
illwieckz: Done full condition again.

src/engine/renderer/tr_shader.cpp

DolceTriade (approved):

PR #1075, opened 2024-04-11
renderer: select stage renderer at loading time

illwieckz:
Move more material stage computation and branching previously done at render time to be done at shader parsing time (at map loading time).

Also factorize some boiler plate code.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1215,7 +1215,7 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 
 		bool shaderHasNoLight;
 
-		stageRenderer_t genericRenderer;
+		stageRenderer_t colorRenderer;
 		bool doShadowFill;
 		bool doForwardLighting;
 
@@ -3520,7 +3520,7 @@ inline bool checkGLErrors()
 	void Tess_ComputeColor( shaderStage_t *pStage );
 
 	void Tess_StageIteratorDebug();
-	void Tess_StageIteratorGeneric();
+	void Tess_StageIteratorColor();
 	void Tess_StageIteratorPortal();
 	void Tess_StageIteratorShadowFill();
 	void Tess_StageIteratorLighting();
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1199,6 +1199,10 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 	  COLLAPSE_REFLECTIONMAP,
 	};
 
+	struct shaderStage_t;
+
+	using stageRenderer_t = void(*)(shaderStage_t *);
+
 	struct shaderStage_t
 	{
 		stageType_t     type;
@@ -1210,6 +1214,10 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 		bool            dpMaterial;
 
 		bool shaderHasNoLight;
+
+		stageRenderer_t genericRenderer;
+		bool doShadowFill;
+		bool doForwardLighting;
 
 		textureBundle_t bundle[ MAX_TEXTURE_BUNDLES ];
 
@@ -3540,6 +3548,17 @@ inline bool checkGLErrors()
 	void Tess_UpdateVBOs();
 
 	void RB_ShowImages();
+
+	void Render_NOP( shaderStage_t *pStage );
+	void Render_generic( shaderStage_t *pStage );
+	void Render_generic3D( shaderStage_t *pStage );
+	void Render_lightMapping( shaderStage_t *pStage );
+	void Render_reflection_CB( shaderStage_t *pStage );
+	void Render_skybox( shaderStage_t *pStage );
+	void Render_screen( shaderStage_t *pStage );
+	void Render_portal( shaderStage_t *pStage );
+	void Render_heatHaze( shaderStage_t *pStage );
+	void Render_liquid( shaderStage_t *pStage );
 
 	/*
 	============================================================
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -611,6 +611,10 @@ static void SetRgbaGen( const shaderStage_t *pStage, colorGen_t *rgbGen, alphaGe
 
 // *INDENT-ON*
 
+void Render_NOP( shaderStage_t * )
+{
+}
+
 static void Render_generic2D( shaderStage_t *pStage )
 {
 	GLimp_LogComment( "--- Render_generic2D ---\n" );
@@ -680,9 +684,9 @@ static void Render_generic2D( shaderStage_t *pStage )
 
 static image_t* GetLightMap();
 
-static void Render_generic( shaderStage_t *pStage )
+void Render_generic3D( shaderStage_t *pStage )
 {
-	GLimp_LogComment( "--- Render_generic ---\n" );
+	GLimp_LogComment( "--- Render_generic3D ---\n" );
 
 	GL_State( pStage->stateBits );
 
@@ -783,6 +787,17 @@ static void Render_generic( shaderStage_t *pStage )
 	GL_CheckErrors();
 }
 
+void Render_generic( shaderStage_t *pStage )
+{
+	if ( backEnd.projection2D )
+	{
+		Render_generic2D( pStage );
+		return;
+	}
+
+	Render_generic3D( pStage );
+}
+
 /*
 =================
 GetLightMap
@@ -826,7 +841,7 @@ static image_t* GetDeluxeMap()
 	}
 }
 
-static void Render_lightMapping( shaderStage_t *pStage )
+void Render_lightMapping( shaderStage_t *pStage )
 {
 	GLimp_LogComment( "--- Render_lightMapping ---\n" );
 
@@ -1793,7 +1808,7 @@ static void Render_forwardLighting_DBS_directional( shaderStage_t *pStage, trRef
 	GL_CheckErrors();
 }
 
-static void Render_reflection_CB( shaderStage_t *pStage )
+void Render_reflection_CB( shaderStage_t *pStage )
 {
 	GLimp_LogComment( "--- Render_reflection_CB ---\n" );
 
@@ -1878,7 +1893,7 @@ static void Render_reflection_CB( shaderStage_t *pStage )
 	GL_CheckErrors();
 }
 
-static void Render_skybox( shaderStage_t *pStage )
+void Render_skybox( shaderStage_t *pStage )
 {
 	GLimp_LogComment( "--- Render_skybox ---\n" );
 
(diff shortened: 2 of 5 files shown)


src/engine/renderer/tr_shader.cpp
```diff
@@ -5120,30 +5145,54 @@ static void CollapseStages()
...
+				break;
+			default:
+				break;
+		}
 	}
 
-	shader.numStages = numActiveStages;
-
-	// Do some precomputation.
+	PileUpStages();
```
DolceTriade, line 5191: Add comment here describing what we're doing.

I know you kinda commented this in the function, but let's also comment the flow of the shader parsing code because it can be quite complex.

src/engine/renderer/tr_shader.cpp
```diff
@@ -5120,30 +5145,54 @@ static void CollapseStages()
...
+}
 
+static void PrecomputeStages()
```
DolceTriade, line 5194: Add comment describing what function is doing

src/engine/renderer/tr_shader.cpp
```diff
@@ -5264,7 +5313,147 @@ static void CollapseStages()
 				stage->normalScale[ i ] = 1.0;
 			}
 		}
+
+		// reliefDepthScale
```
DolceTriade, line 5317: What does this comment mean?

DolceTriade (commented):
In general, since you're rewriting a lot of this code, I think this is a good time to comment the code flow of shader parsing and explain what we're doing at each step for future readers (and yourself in 3 months :D )

src/engine/renderer/tr_shader.cpp
```diff
@@ -5265,6 +5314,136 @@ static void CollapseStages()
...
+			{ stageType_t::ST_SKYBOXMAP, &Render_NOP },
+			{ stageType_t::ST_SCREENMAP, &Render_NOP },
+			{ stageType_t::ST_SKYBOXMAP, &Render_skybox },
+			{ stageType_t::ST_SCREENMAP, &Render_screen },
+			{ stageType_t::ST_PORTALMAP, &Render_portal },
+			{ stageType_t::ST_HEATHAZEMAP, &Render_heatHaze },
+			{ stageType_t::ST_LIQUIDMAP, &Render_liquid },
+			{ stageType_t::ST_ATTENUATIONMAP_XY, &Render_NOP },
+			{ stageType_t::ST_ATTENUATIONMAP_Z, &Render_NOP },
+		};
```
illwieckz, line 5358: I have a C++ question for this kind of `unordered_map`. I've done some testing on `godblot.org` and  it looks like when I write such `unordered_map` within a function GCC always rebuild it on every function run, at least the generated code is 2 or3 time longer.

If such `unordered_map` is written outside of the function, the map is done only once.

I wonder if there is a way to tell the compiler “this will never change, please creates it once for all”.

I noticed the same happens with basic arrays, but adding `static` makes GCC defines it globally and once for all, but here `static` doesn't help with maps.

Is there a way to tell the compiler “this will never change, please creates it once for all”?

I like the ability to set this in the block it is used instead of doing it globally, to not have to scroll the code that much in editor.

illwieckz: `Ishq answered on IRC.

src/engine/renderer/tr_shader.cpp
illwieckz: I'm also looking for a better function name.

Basically it turns this:

```c
maxStages = 10;
stages[maxStages] = { active, active, inactive, active, inactive, inactive, active, null, null, null };
numStages = 7;
```

into this:

```c
maxStages = 10;
stages[maxStages] = { active, active, active, active, inactive, inactive, inactive, null, null, null };
numStages = 4;
```

That meaning loops iterating stages don't have to do this:

```c
for ( i = 0; i < 7; i++ )
{
  if ( stages[ i ] == inactive )
    continue;

  render( stage[ i ] );
}
```

but can do this:

```c
for ( i = 0; i < 5; i++ )
{
  render( stage[ i ] );
}
```

For reference our current code in master is doing this overly complicated code:

```c
for ( i = 0; i < 10; i++ )
{
  if ( stages[ i ] == inactive )
    continue;

  if ( stages[ i ] == null )
    break;

  render( stage[ i ] );
}
```

illwieckz: So I look for a function name that makes obvious we are compacting the stage list.

src/engine/renderer/tr_shade.cpp
```diff
@@ -566,34 +566,20 @@ void Tess_Begin( void ( *stageIteratorFunc )(),
...
-	r_normalScale is only applied on Z.
-	This behaviour is inherited.
-	*/
-	normalScale[ 2 ] *= r_normalScale->value;
+	/* The GLSL code disables normal map scaling when normal Z scale
+	is equal to zero. It means normal map scaling is disabled when
+	r_normalScale is set to zero. This is cool enough to be kept as
+	a feature. Normal Z component equal to zero would be wrong anyway.
+	r_normalScale is only applied on Z. */
+	normalScale[ 2 ] = pStage->normalScale[ 2 ] *= r_normalScale->value;
```
slipher, line 582: Either `*=` is a typo, or it should be rewritten not to look weird.

src/engine/renderer/tr_shade.cpp
```diff
@@ -1587,14 +1579,10 @@ static void Render_forwardLighting_DBS_proj( shaderStage_t *pStage,
 	// bind u_HeightMap
 	if ( pStage->enableReliefMapping )
 	{
-		float depthScale;
-		float reliefDepthScale;
+		float depthScale = RB_EvalExpression( &pStage->depthScaleExp, r_reliefDepthScale->value );
```
slipher, line 1587: RB_EvalExpression can have time-dependent values, so there may be reasons not to precompute it.

slipher (commented):
Not a fan of replacing if-else or switch statements with function pointers, especially when there are only 2 or 3 options. It's harder to read. Also function tables are supposed to be slightly slower than `switch`, so optimization is not a justification.

src/engine/renderer/tr_shade.cpp
illwieckz: Right, it should be `*`.

illwieckz: Fixed, the patch is now exported to this PR:

https://github.com/DaemonEngine/Daemon/pull/1081

So it can be merged separately.

illwieckz:
I exported two commits that can be merged separately:

- https://github.com/DaemonEngine/Daemon/pull/1081
- https://github.com/DaemonEngine/Daemon/pull/1082

They are not in this branch anymore.

Those commits are just either rewriting things in less convoluted ways, or factorise code.

illwieckz:
I also exported:

- https://github.com/DaemonEngine/Daemon/pull/1083

src/engine/renderer/tr_shade.cpp
illwieckz: This `RB_EvalExpression()` call do not use precomputed input.

src/engine/renderer/tr_shader.cpp
illwieckz: It meant nothing, I rewrote it. 🤣️

src/engine/renderer/tr_shader.cpp
VReaperV: SortStages(), GroupActiveStages() is what comes to mind for a function name.

illwieckz:
This PR now only ships the commit preselecting the renderers at shader stage.

I appreciate doing this at the same place whatever the kind of renderers, it reduces the risk of forgetting to update the code when adding new stage types.

For example when I added the various light style sages types before, I forgot to update some switch case in some places, The same happened to @VReaperV when updating the portal code where a specific code was not run for all kind of stages. Now the per-stage renderer selection is done in a single place in a way we can't miss to update the various options.

src/engine/renderer/tr_local.h
```diff
@@ -1199,6 +1199,10 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 	  COLLAPSE_REFLECTIONMAP,
 	};
 
+	struct shaderStage_t;
+
+	typedef void ( *stageRenderer_t )( shaderStage_t* );
```
slipher, line 1204: `using stageRenderer_t = void(*)(shaderStage_t *)`

src/engine/renderer/tr_shader.cpp
```diff
@@ -5264,6 +5311,140 @@ static void CollapseStages()
...
+	};
+
+	for ( int s = 0; s < shader.numStages; s++ )
+	{
+		shaderStage_t *stage = &stages[ s ];
+
+		stage->genericRenderer = &Render_NOP;
+		stage->doShadowFill = false;
+		stage->doForwardLighting = false;
+
```
slipher, line 5419: How about ASSERTing the `count` to avoid looking up the key twice in release mode?

src/engine/renderer/tr_shader.cpp
```diff
@@ -5303,20 +5484,16 @@ static void SortNewShader()
...
 }
 
-/*
-====================
-GeneratePermanentShader
-====================
-*/
-static shader_t *GeneratePermanentShader()
+// Copy the current global shader to a newly allocated shader.
+static shader_t *MakeShaderPermanent()
```
slipher, line 5488: Maybe SetStagesRenderers should be called from here since they are always used together? You could name this function FinalizeShader

illwieckz: The function calling this code is already named `FinishShader`, this would be confusing if `FinishShader` calls `FinalizeShader`.

Those functions are basically cutting the large `FinishShader` function in smaller blocks.

src/engine/renderer/tr_shader.cpp
illwieckz: This code is not that critical for performance. You said [here](https://github.com/DaemonEngine/Daemon/pull/1076#issuecomment-2053715689) that calling `.at()` on a `const unordered_map` with a missing key would crash, and if it's true it would even crash in release code, even when `ASSERT` is disabled.

If we care about performance we can do a switch case.

illwieckz: I did a switch case, after all my rewrites, it's now a good solution again.

src/engine/renderer/tr_local.h
illwieckz: Done.

src/engine/renderer/tr_local.h
```diff
@@ -1211,6 +1215,10 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 
 		bool shaderHasNoLight;
 
+		stageRenderer_t genericRenderer;
```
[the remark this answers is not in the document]
illwieckz: I added a commit to rename `Tess_StageIteratorGeneric` as `Tess_StageIteratorColor`, this removes all the confusion with `Render_generic`.

slipher:
LGTM

PR #1076, opened 2024-04-12
renderer: set unordered_map static and const

illwieckz:
This is not critical since it's used in debug command print functions, but why not. Those `unordered_map` are never modified so there is no need to rebuild them on every function call.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -153,14 +153,14 @@ R_ListImages_f
 */
 void R_ListImages_f()
 {
-	std::unordered_map<GLenum, std::string> imageTypeName = {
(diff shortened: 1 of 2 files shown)


DolceTriade (commented):
make them const too?

illwieckz:
Both GCC and Clang error out if I add `const` to an `unordered_map`.

slipher:
To use it with `const` you can do `map.at(key)` instead of `map[key]`. This crashes if the key is missing though, so maybe it's better not to.

illwieckz:
I always check for the key not missing (was not expecting to not crash if the key is missing anyway), so maybe I can try that.

illwieckz:
I now use `const` and `.at()`. It is useful to check for unknown keys and report debug logs if it happens.

DolceTriade (approved):

PR #1077, opened 2024-04-13
Shorten name of light clamping worldspawn key

slipher:
q3map2 finds @illwieckz's names too verbose :stuck_out_tongue: 

When q3map2 sees the key "forceLegacyMapOverBrightClamping" in the worldspawn entity, it protests that it is too long and shuts down. Shorten that to "forceLegacyOverBrightClamping".


illwieckz (approved):
LGTM.

PR #1078, opened 2024-04-13
renderer: missing break in shadowing code and other things

illwieckz:
It was missing in 1872b92274fbd9533e622633c56a515a4ff899c0 from:

- https://github.com/DaemonEngine/Daemon/pull/1045


illwieckz:
I also added a fix for an inverted test.

illwieckz:
I also added a fix for a logic change I mistakenly introduced.

I don't know if the original logic was right or wrong, but my previous patch changed the way it behaved.

illwieckz:
I also added a commit to unify the wording of different functions doing similar things, it makes easier for the mind to spot the patterns.

slipher:
LGTM

PR #1085, opened 2024-04-16
Add variant of -connect command line argument that needn't be last

slipher:
Intended to be used with a protocol handler that routes through the updater. If the updater gets the "URL" first, it can validate it so we don't need to ignore further arguments as a defense against bad input.

This improves on -connect in case the user uses the command line customization feature of the updater like %command% -foo +bar. The -connect2 arg could be inserted in %command% without cutting off the rest.

I'm intentionally leaving the original variant in; it may be useful to have both possibilities.

### DIFF
--- a/src/engine/framework/System.cpp
+++ b/src/engine/framework/System.cpp
@@ -461,9 +461,7 @@ static void ParseCmdline(int argc, char** argv, cmdlineArgs_t& cmdlineArgs)
 			}
 
 			// Make it forwardable.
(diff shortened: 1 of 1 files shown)


illwieckz:
Why not `-connect-validated` or something like that?

illwieckz:
The code look good to me but I would prefer the option to have a more explicit name than just `-connect2`.

slipher:
Thanks, I named it `connect-trusted`.

illwieckz (approved):
LGTM.

PR #1089, opened 2024-04-20
renderer,qmath: avoid useless pointer dereferences and let the compiler vectorize matrix copy

illwieckz:
Avoid useless pointer dereferences and let the compiler vectorize matrix copy.

### DIFF
--- a/src/engine/qcommon/q_math.cpp
+++ b/src/engine/qcommon/q_math.cpp
@@ -1477,62 +1477,17 @@ float VectorDistanceSquared( vec3_t v1, vec3_t v2 )
 // *INDENT-OFF*
 void MatrixIdentity( matrix_t m )
 {
-	m[ 0 ] = 1;
-	m[ 4 ] = 0;
-	m[ 8 ] = 0;
-	m[ 12 ] = 0;
-	m[ 1 ] = 0;
-	m[ 5 ] = 1;
-	m[ 9 ] = 0;
-	m[ 13 ] = 0;
-	m[ 2 ] = 0;
-	m[ 6 ] = 0;
-	m[ 10 ] = 1;
-	m[ 14 ] = 0;
-	m[ 3 ] = 0;
-	m[ 7 ] = 0;
-	m[ 11 ] = 0;
-	m[ 15 ] = 1;
+	memcpy( m, matrixIdentity, sizeof( matrix_t ) );
 }
 
 void MatrixClear( matrix_t m )
 {
-	m[ 0 ] = 0;
-	m[ 4 ] = 0;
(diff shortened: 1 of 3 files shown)


illwieckz:
Before:

20240420-192723-000 daemon-optimize

After:

20240420-192857-000 daemon-optimize

---

Before:

20240420-192725-000 daemon-optimize

After:

20240420-192902-000 daemon-optimize

DolceTriade (approved):

illwieckz:
I added a commit doing similar things on other functions.

From left to right, original code, calling `MatrixIdentity` with LTO and luck, calling `memcpy` directly.

20240420-222759-002 daemon-optimize

The second one is reusing `MatrixIdentity` instead of `memcpy` directly, when with some luck LTO is inlining it, when luck doesn't happen, it still does an explicit call (even when adding `inline`);

20240420-225208-000 daemon-optimize

That's why I call `memcpy` directly instead of reusing `MatrixIdentity`.

Those functions are only used by the legacy lighting renderer though.

slipher:
- `renderer: make RB_CalcTexMatrix iterate pointers `  LGTM
- `renderer: make Tess_ComputeTexMatrices iterate pointers ` LGTM
- `q_math: let the compiler vectorize matrix copy ` LGTM
- `q_math: let the compiler vectorize more matrix operations`  I'm skeptical. It's ugly to write one value and then overwrite it with another one. How did you get such bad disassembly for MatrixSetupScale in the left and right columns of https://github.com/DaemonEngine/Daemon/pull/1089#issuecomment-2067780810? (The middle one isn't the same function.) The right-hand one kind of reminds me of GCC's unoptimized mode where it pretends that registers can't be used for storage. With GCC -O2 I got a perfectly good result for the current version of MatrixSetupScale on Godbolt:
 ```
        mov     QWORD PTR [rdi+4], 0
        mov     QWORD PTR [rdi+12], 0
        mov     QWORD PTR [rdi+24], 0
        mov     QWORD PTR [rdi+32], 0
        mov     QWORD PTR [rdi+44], 0
        mov     QWORD PTR [rdi+52], 0
        mov     DWORD PTR [rdi+60], 0x3f800000
        movss   DWORD PTR [rdi], xmm0
        movss   DWORD PTR [rdi+20], xmm1
        movss   DWORD PTR [rdi+40], xmm2
        ret
```

illwieckz:
Right, it looks like I mistakenly had `USE_DEBUG_OPTIMIZE` set to `OFF`, once I set it set to `ON` I got (my code on the left, the one from `master` on the right):

20240421-011814-001 daemon-optimize

So yes, the extra commit seems to be useless, let's drop it.

PR #1091, opened 2024-04-21
tr_shader: forget lightmap if not collapsed, so it can be rediscovered

illwieckz:
Prevent this kind of error when we detected a lightmap but we cannot collapse it with the next colormap:

```
Warn: more than one lightmap stage in shader 'textures/usstremor_custom_src/redshade_01'
```


illwieckz:
> Should this just be an `else` rather than an `else if` case? Since no matter what the condition evaluates to `lightStage` will be set to -1 if that code is reached.

Right, that would do the same.

illwieckz:
> Should this just be an `else` rather than an `else if` case? Since no matter what the condition evaluates to `lightStage` will be set to -1 if that code is reached.

Now done.

I also added a commit to remove an unused `lightMapCount` variable.

VReaperV (approved):
LGTM

PR #1093, opened 2024-04-21
Make the legacy dynamic light renderer work again

illwieckz:
Something is wrong as the light is too dim, but it works.

About the wrongness, here with both map dynamic light, and weapon projectile dynamic light:

unvanquished_2024-04-21_202022_000

It should look like this:

unvanquished_2024-04-21_204455_000

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -611,6 +611,11 @@ static void SetRgbaGen( const shaderStage_t *pStage, colorGen_t *rgbGen, alphaGe
 
 // *INDENT-ON*
 
+void Render_NONE( shaderStage_t * )
+{
+	ASSERT_UNREACHABLE();
+}
+
 void Render_NOP( shaderStage_t * )
 {
 }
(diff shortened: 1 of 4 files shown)


src/engine/renderer/tr_shade.cpp
```diff
@@ -1580,8 +1594,17 @@ static void Render_forwardLighting_DBS_proj( shaderStage_t *pStage,
...
+	{
+		// standalone lightmap stage: paint shadows over a white texture
+		GL_BindToTMU( 0, tr.whiteImage );
+	}
+	else
+	{
+		GL_BindToTMU( 0, pStage->bundle[ TB_DIFFUSEMAP ].image[ 0 ] );
+
+		gl_forwardLightingShader_projXYZ->SetUniform_TextureMatrix( tess.svars.texMatrices[ TB_DIFFUSEMAP ] );
+	}
```
DolceTriade, line 1607: Should this be a helper function? Sounds like this pattern might need to be repeated for other types of situations?

illwieckz: Right, it can be a helper function. We may even do `pStage->bundle[ TB_DIFFUSEMAP ].image[ 0 ] = tr.whiteImage` in `tr_shader.cpp` at shader processing time to simplify the code here.

illwieckz: I've set the image in `tr_shader.cpp`.

Now making a helper function for a `if` would probably be overkill. It's also possible that `if` is just to skip useless uniform setting as a fullwhite image renders the same whatever the coordinates, and the code may now work without any `if` (I have not tried).

DolceTriade (approved):

PR #1095, opened 2024-04-23
Don't check overflow for data already in VBOs

VReaperV:
Overflow only makes sense when uploading new vertices, models that already have their vertices in vram will not overflow anything.


slipher:
LGTM

DolceTriade:
Tested with my chaingun model that crashes without this patch. Works fine.

PR #1097, opened 2024-04-23
Add sky surfaces to world VBO

### LINKED ISSUE #1096: Surfaces that will never be used for drawing still get added in RB_RenderDrawSurfaces()
See https://users.unvanquished.net/~sweet/reaper/map-example_1.dpk for an example.

The sky and portal surfaces are skipped in R_CreateWorldVBO(), but RB_RenderDrawSurfaces() still adds the sky surfaces, which it ends up uselessly copying into tess.verts and tess.indexes and, in this case, overflowing SHADER_MAX_VERTEXES. Those surfaces are never used for drawing.

VReaperV:
Stops tess.vertexes from overflowing by adding all the sky surfaces to world VBO instead of doing Tess_SurfaceVertsAndTris() on them.

This essentialy moves them from RAM to VRAM. It might be better to just always draw the skybox rather than have those sky surfaces in memory.

Fixes #1096.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -2975,7 +2975,7 @@ static void R_CreateWorldVBO()
 	{
 		surface = &s_worldData.surfaces[ k ];
 
-		if ( surface->shader->isSky || surface->shader->isPortal )
+		if ( surface->shader->isPortal )
 		{
 			continue;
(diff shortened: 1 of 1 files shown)


illwieckz (approved):
I don't know well that part of the code but I see no rendering difference and no performance loss.

LGTM.

PR #1098, opened 2024-04-23
renderer: disable heat haze stage at load time

illwieckz:
Disable heat haze stage at load time, instead of iterating it and skipping it at render time.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -1990,11 +1990,6 @@ void Render_heatHaze( shaderStage_t *pStage )
 
 	GLimp_LogComment( "--- Render_heatHaze ---\n" );
 
-	if ( r_heatHaze->integer == 0 )
-	{
-		return;
-	}
-
 	// remove alpha test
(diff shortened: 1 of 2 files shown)


illwieckz:
I added an extra commit to do `b = condition` instead of `if (!condition) b = false` when disabling stages.

illwieckz:
It's just extending a PR already merged:
- https://github.com/DaemonEngine/Daemon/pull/1075/files#diff-17e274af0ffef15a9550978320ee0f4ad5a61dfa7f3af3b350b456f8083e64faL2631-L2637
- https://github.com/DaemonEngine/Daemon/pull/1075/files#diff-9bbcfb69e22103b6c19b7c09f034a0c69882809de7e9a48de71b499c01a63b59R5182-R5188

Instead of testing if the feature is disabled at render time while iterating active stages, just test if the feature is disabled at shader parsing time and remove it from the list that will be processed at render time:

- https://github.com/DaemonEngine/Daemon/pull/1074

DolceTriade (approved):

PR #1099, opened 2024-04-23
Iterate shader stages directly, avoid array of pointers

illwieckz:
Iterate shader stages directly, avoid array of pointers.

### DIFF
--- a/src/engine/renderer/tr_cmds.cpp
+++ b/src/engine/renderer/tr_cmds.cpp
@@ -418,12 +418,12 @@ void RE_SetColorGrading( int slot, qhandle_t hShader )
 		return;
 	}
 
-	if ( shader->defaultShader || !shader->stages[ 0 ] )
+	if ( shader->defaultShader || shader->stages == shader->lastStage )
 	{
 		return;
 	}
 
-	image = shader->stages[ 0 ]->bundle[ 0 ].image[ 0 ];
+	image = shader->stages[ 0 ].bundle[ 0 ].image[ 0 ];
 
 	if ( !image )
 	{
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -440,8 +440,8 @@ static void DrawTris()
 	gl_genericShader->SetDepthFade( false );
 	gl_genericShader->SetAlphaTesting( false );
 
-	if( tess.surfaceShader->stages[0] ) {
-		deform = tess.surfaceShader->stages[0]->deformIndex;
+	if( tess.surfaceStages != tess.surfaceLastStage ) {
+		deform = tess.surfaceStages[ 0 ].deformIndex;
 	}
 
 	gl_genericShader->BindProgram( deform );
@@ -524,7 +524,7 @@ void Tess_Begin( void ( *stageIteratorFunc )(),
 
 		tess.surfaceShader = state;
 		tess.surfaceStages = state->stages;
-		tess.numSurfaceStages = state->numStages;
+		tess.surfaceLastStage = state->lastStage;
 
 		Tess_MapVBOs( false );
 	}
@@ -532,9 +532,9 @@ void Tess_Begin( void ( *stageIteratorFunc )(),
 	{
 		state = nullptr;
 
-		tess.numSurfaceStages = 0;
 		tess.surfaceShader = nullptr;
 		tess.surfaceStages = nullptr;
+		tess.surfaceLastStage = nullptr;
 		Tess_MapVBOs( false );
 	}
 
@@ -848,7 +848,8 @@ void Render_lightMapping( shaderStage_t *pStage )
 	specific use case. Without knowing which use case this takes care about,
 	any change in the following code may break it. Or it may be a hack we
 	should drop if it is for a bug we don't have anymore. */
-	bool hack = tess.numSurfaceStages > 0 && tess.surfaceStages[0]->rgbGen == colorGen_t::CGEN_VERTEX;
+	bool hack = tess.surfaceLastStage != tess.surfaceStages
+		&& tess.surfaceStages[ 0 ].rgbGen == colorGen_t::CGEN_VERTEX;
 
 	if ( ( tess.surfaceShader->surfaceFlags & SURF_NOLIGHTMAP ) && !hack )
 	{
@@ -2586,10 +2587,8 @@ void Tess_StageIteratorColor()
 	}
 
 	// call shader function
-	for ( int stage = 0; stage < tess.numSurfaceStages; stage++ )
+	for ( shaderStage_t *pStage = tess.surfaceStages; pStage < tess.surfaceLastStage; pStage++ )
 	{
-		shaderStage_t *pStage = tess.surfaceStages[ stage ];
-
 		if ( !RB_EvalExpression( &pStage->ifExp, 1.0 ) )
 		{
 			continue;
@@ -2636,9 +2635,8 @@ void Tess_StageIteratorPortal() {
 		GL_Cull( tess.surfaceShader->cullType );
 
 	// call shader function
-	for ( int stage = 0; stage < tess.numSurfaceStages; stage++ ) {
-		shaderStage_t* pStage = tess.surfaceStages[stage];
-
+	for ( shaderStage_t *pStage = tess.surfaceStages; pStage < tess.surfaceLastStage; pStage++ )
+	{
 		if ( !RB_EvalExpression( &pStage->ifExp, 1.0 ) ) {
 			continue;
 		}
@@ -2680,10 +2678,8 @@ void Tess_StageIteratorShadowFill()
 	}
 
 	// call shader function
-	for ( int stage = 0; stage < tess.numSurfaceStages; stage++ )
+	for ( shaderStage_t *pStage = tess.surfaceStages; pStage < tess.surfaceLastStage; pStage++ )
 	{
-		shaderStage_t *pStage = tess.surfaceStages[ stage ];
-
 		if ( !RB_EvalExpression( &pStage->ifExp, 1.0 ) )
 		{
 			continue;
@@ -2704,8 +2700,6 @@ void Tess_StageIteratorShadowFill()
 void Tess_StageIteratorLighting()
 {
 	trRefLight_t  *light;
-	shaderStage_t *attenuationXYStage;
-	shaderStage_t *attenuationZStage;
 
 	// log this call
 	if ( r_logFile->integer )
@@ -2757,12 +2751,10 @@ void Tess_StageIteratorLighting()
 	}
 
 	// call shader function
-	attenuationZStage = tess.lightShader->stages[ 0 ];
+	shaderStage_t *attenuationZStage = &tess.lightShader->stages[ 0 ];
 
-	for ( int i = 0; i < tess.numSurfaceStages; i++ )
+	for ( shaderStage_t *pStage = tess.surfaceStages; pStage < tess.surfaceLastStage; pStage++ )
 	{
-		shaderStage_t *pStage = tess.surfaceStages[ i ];
-
 		if ( !RB_EvalExpression( &pStage->ifExp, 1.0 ) )
 		{
 			continue;
@@ -2770,10 +2762,10 @@ void Tess_StageIteratorLighting()
 
 		Tess_ComputeTexMatrices( pStage );
 
-		for ( int j = 1; j < tess.lightShader->numStages; j++ )
+		for ( shaderStage_t *attenuationXYStage = tess.lightShader->stages;
+			attenuationXYStage < tess.lightShader->lastStage;
+			attenuationXYStage++ )
 		{
-			attenuationXYStage = tess.lightShader->stages[ j ];
-
 			if ( attenuationXYStage->type != stageType_t::ST_ATTENUATIONMAP_XY )
 			{
 				continue;
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -41,6 +41,7 @@ static char          *s_shaderText;
 // dynamically allocated memory if it is valid.
 static shaderTable_t table;
 static std::array<shaderStage_t, MAX_SHADER_STAGES> stages;
+static size_t numStages;
 static shader_t      shader;
 static texModInfo_t  texMods[ MAX_SHADER_STAGES ][ TR_MAX_TEXMODS ];
 
@@ -4738,9 +4739,9 @@ SHADER OPTIMIZATION AND FOGGING
 // Group all active stages to the begining of the array.
 static void GroupActiveStages()
 {
-	int numActiveStages = 0;
+	size_t numActiveStages = 0;
 
-	for ( int i = 0; i < shader.numStages; i++ )
+	for ( size_t i = 0; i < numStages; i++ )
 	{
 		if ( !stages[ i ].active )
 		{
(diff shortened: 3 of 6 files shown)


illwieckz:
The code iterate an array of `MAX_SHADER_STAGES` stages with `numStages` only when parsing shaders, as long as we can disable some and group them.

When making the shader permanent and ready to feed the renderer, we now only allocate the exact amount of active stages just before copying them to the permanent shader.

Then, at render time, the code now iterate pointers directly, and we avoid dereferencing the pointer of a pointer on every stage.

src/engine/renderer/tr_cmds.cpp
```diff
@@ -418,12 +418,12 @@ void RE_SetColorGrading( int slot, qhandle_t hShader )
 		return;
 	}
 
-	if ( shader->defaultShader || !shader->stages[ 0 ] )
+	if ( shader->defaultShader || !shader->stages )
```
slipher, line 421: Testing whether the begin pointer is null doesn't look good because it really depends on the details of how the array was allocated (e.g. `malloc(0)` may or may not be null)

src/engine/renderer/tr_shader.cpp
```diff
@@ -5486,22 +5486,31 @@ static shader_t *MakeShaderPermanent()
...
+	newShader->stages = nullptr;
+	newShader->lastStage = nullptr;
+
+	DAEMON_ASSERT( numStages < MAX_SHADER_STAGES );
+	DAEMON_ASSERT( MAX_SHADER_STAGES < std::numeric_limits<uint8_t>::max() );
+	DAEMON_ASSERT( MAX_TEXTURE_BUNDLES < std::numeric_limits<uint8_t>::max() );
 
-		newShader->stages[ i ] = (shaderStage_t*) ri.Hunk_Alloc( sizeof( stages[ i ] ), ha_pref::h_low );
-		*newShader->stages[ i ] = stages[ i ];
+	if ( numStages )
```
slipher, line 5496: No need for an if. The functioning should be consistent zero or nonzero

src/engine/renderer/tr_sky.cpp
```diff
@@ -141,10 +141,6 @@ void Tess_StageIteratorSky()
 	for ( int stage = 0; stage < MAX_SHADER_STAGES; stage++ ) {
 		shaderStage_t* pStage = tess.surfaceShader->stages[stage];
 
-		if ( !pStage || !pStage->bundle[0].image[0] ) {
```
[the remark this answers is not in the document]
illwieckz: Right, I modified this commit to iterate to `tess.numSurfaceStages` (before the next commit rewrites that).

src/engine/renderer/tr_shader.cpp
```diff
@@ -41,6 +41,7 @@ static char          *s_shaderText;
 // dynamically allocated memory if it is valid.
 static shaderTable_t table;
 static std::array<shaderStage_t, MAX_SHADER_STAGES> stages;
+static uint8_t numStages;
```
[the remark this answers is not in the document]
illwieckz: This data is only used when processing the global temporary `stages` array.

I don't see why that global temporary `numStage` variable would have more or less less risk to be out of sync with the global temporary `stages` array, if stored or not in the global temporary `shader` struct. It was global, it is still global. The risks are the same. The code behaviour of having thousands of lines modifying global variables before calling a function to copy their content to a newly-allocated “permanent” shader is left as is.

Why not making a global struct for all the temporary things? Like this:

```c
struct tmpShaderData_t {
	shaderTable_t table;
	std::array<shaderStage_t, MAX_SHADER_STAGES> stages;
	shader_t shader;
	texModInfo_t texMods[ MAX_SHADER_STAGES ][ TR_MAX_TEXMODS ];
};

static tmpShaderData_t tmp;
```

It also means we could do:

```c
static void ClearGlobalShader()
{
	ResetStruct( tmp );
}
```

It would also makes more obvious this kind of code:

```c
for ( uint8_t s = 0; s < numStages; s++ )
{
	shaderStage_t *stage = &stages[ s ];

	if ( shader.numDeforms > 0 ) {
		stage->deformIndex = gl_shaderManager.getDeformShaderIndex( shader.deforms, shader.numDeforms );
	}
}
```

by turning it into:

```c
for ( uint8_t s = 0; s < tmp.numStages; s++ )
{
	shaderStage_t *stage = &tmp.stages[ s ];

	if ( tmp.shader.numDeforms > 0 ) {
		tmp.stage->deformIndex = gl_shaderManager.getDeformShaderIndex( tmp.shader.deforms, tmp.shader.numDeforms );
	}
}
```

I'm personnally annoyed by all the global temporary variables being named `shader` and `stages` so I would even love that. I see no reason to use the shader and stage structs for storing data that is only used at parsing them and never used at rendering time, we have some others of them, for example shader variables that are only used to configure stages at load time.

This change would be probably too intrusive so I may postpone it to another PR, but that is something I consider.

src/engine/renderer/tr_cmds.cpp
illwieckz: If there is no stage at all the code explicitly sets this pointer to `nullptr` and does no allocation at all.

src/engine/renderer/tr_shader.cpp
illwieckz: This `if` is the solution for your other remark:

> Testing whether the begin pointer is null doesn't look good because it really depends on the details of how the array was allocated (e.g. `malloc(0)` may or may not be null)

For which I wrote:

> If there is no stage at all the code explicitly sets this pointer to `nullptr` and does no allocation at all.

src/engine/renderer/tr_shader.cpp
```diff
@@ -6687,18 +6698,19 @@ void R_ListShaders_f()
 			continue;
 		}
 
-		stageCount += shader->numStages;
-		highestShaderStageCount = std::max( highestShaderStageCount, shader->numStages );
+		const uint8_t stageCount = shader->lastStage - shader->stages;
```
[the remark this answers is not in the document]
illwieckz: I replaced `uint8_t` by `size_t`.

illwieckz:
Why do I get that?

> AppVeyor was unable to build non-mergeable pull request

DolceTriade:
Rebase maybe?

src/engine/renderer/tr_cmds.cpp
slipher: I know that it works with exactly how it is written now. I'm saying that it looks suspicious; provokes a "WTF" from the reader. In another part of the code you use `tess.surfaceLastStage != tess.surfaceStages` to test for emptiness. I suggest always doing it that way.

src/engine/renderer/tr_shader.cpp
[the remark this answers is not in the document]
slipher: > This accounting variable already existed

How do you mean?

I think `table` is not temporary, but other than that the `tmp` idea seems decent.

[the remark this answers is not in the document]
illwieckz: What I mean is that before, we had the `shader.numStage` global accounting variable for the `stages` global array. Now we have the `numStage`global accounting variable  for the `stages` global array, I see no difference about risks of not being in sync (no less, no more).

The `numStage` variable was never part of the `stages` array.

And `numStage| was and is still global for the “currently processed shader” with the “currently processed stages list”, and there cannot be more than one global currently processed shader and more than one global current processed stage list, so that `numStage` global variable cant be storing data from something else than the `stages` global array.

src/engine/renderer/tr_cmds.cpp
illwieckz: I now test for `stage == lastStage` to test for emptiness everywhere.

Of course it works if both `stage` and `lastStage` are the same, even if those pointers are not `nullptr`.

Though, I wounder if it would be good to make sure things like `Hunk_Alloc` always return `nullptr` when `size` is zero. It may helps debugging: not only returning `nullptr` makes sure any access will crash, but also the developper could make the fair assumption nothing was allocated, helping to track down the possible problems he may face. But that can be done in another PR, and not be specific to that change.

I'm just afraid of seeing a non-`nullptr` in a debugger for something that was never allocated, and don't know what to do with it.

But comparing `stage` and `lastStage` is fine for the purpose of this PR.

src/engine/renderer/tr_shader.cpp
illwieckz: The `if` is now removed.

illwieckz:
> Rebase maybe?

At the time it was rebased on current `master`, I wonder if AppVeyor donesn't build if there are unsolved blocking discussions.

src/engine/renderer/tr_cmds.cpp
illwieckz: Here we have an example of how an allocation of size `0` can corrupt memory and lead to a crash, sometime in some other random places, such crash being hard to track down:

- https://github.com/Unvanquished/Unvanquished/pull/2976

I don't know if it's a compiler bug though… But I wonder if it's a good idea to allocate a memory of size `0`, and what can be returned from this operation.

DolceTriade:
```
C:\projects\daemon\src\engine\renderer\tr_shader.cpp(4751,10): error C2220: the following warning is treated as an error [C:\projects\daemon\build\client-objects.vcxproj]
C:\projects\daemon\src\engine\renderer\tr_shader.cpp(4751,10): warning C4389: '!=': signed/unsigned mismatch [C:\projects\daemon\build\client-objects.vcxproj]
```

illwieckz:
> ```
> C:\projects\daemon\src\engine\renderer\tr_shader.cpp(4751,10): error C2220: the following warning is treated as an error [C:\projects\daemon\build\client-objects.vcxproj]
> C:\projects\daemon\src\engine\renderer\tr_shader.cpp(4751,10): warning C4389: '!=': signed/unsigned mismatch [C:\projects\daemon\build\client-objects.vcxproj]
> ```

This one is new!!! 😀️

I now fixed it.

illwieckz:
I added a commit (to be squashed) that copies the stage data in one go. That's something I tried before but it didn't worked, I probably made a stupid mistake at the time because right now it works.

illwieckz:
I added some fixup commits to be squashed to simplify the code.

illwieckz:
> `if ( shader->defaultShader || !shader->stages )`
> I'm saying that it looks suspicious; provokes a "WTF" from the reader.

> `if ( numStages )`
> No need for an if. The functioning should be consistent zero or nonzero

About setting `stages` to `nullptr` to know if the stage list is empty, the code was already doing that for both the stage list and the shader, and is still doing it for the shader:

```c
if ( surfaceShader != nullptr )
{
	state = ( surfaceShader->remappedShader ) ? surfaceShader->remappedShader : surfaceShader;

	tess.surfaceShader = state;
	tess.surfaceStages = state->stages;
	tess.surfaceLastStage = state->lastStage;

	Tess_MapVBOs( false );
}
else
{
	state = nullptr;

	tess.surfaceShader = nullptr;
	tess.surfaceStages = nullptr;
	tess.surfaceLastStage = nullptr;
	Tess_MapVBOs( false );
}

// …

if ( state != nullptr && state->isSky )
{
	tess.stageIteratorFunc = &Tess_StageIteratorSky;
}
```

So I would not be against setting `stages` to `nullptr` when `empty`, and just testing for `stages` not being null instead of comparing `stages` with `lastStage`, really.

Even my latest changes simplifying the code would carry the `nullptr`. And we can skip calling the allocation and copy functions of the stage list with an `if`, even if the code works without.

illwieckz:
Any last word?

src/engine/renderer/tr_shade.cpp
```diff
@@ -510,54 +510,55 @@ void Tess_Begin( void ( *stageIteratorFunc )(),
...
+		tess.surfaceLastStage = tess.surfaceShader->lastStage;
 	}
 	else
 	{
-		state = nullptr;
+		/* No code is using this on purpose, this is kept because of the
+		scary comment above saying:
 
-		tess.surfaceShader = nullptr;
+		> A surface may be forced to perform a Tess_End due to overflow. */
```
illwieckz, line 550: I have no idea about what to do with this.

slipher:
> Any last word?

I don't see any LGTMs 🤪

illwieckz:
> > Any last word?
> 
> I don't see any LGTMs 🤪

That can be a last word.

illwieckz:
I would like to know if there is something blocking there. I'm planning other changes in the shader code and I want to base them on this.

Among the things remaining in the discussions I see the disliking of the global variable `numStages` not being global while not being part of a struct. First thing a disliking is not blocking. Second thing, it is now only global to `tr_shader.cpp` and not shared with all the code and especially not used anymore in `tr_shade.cpp` and maybe others. It is also now only used for the temporary shader parsing data before making the permanent shader to be used for rendering things. It's actually a step forward to make possible in the future to transform the stage list into a vector or other structure if we prefer that to a “start,end” tuple, most of the code is now properly shaped to make other things possible. As a second step after this PR I also want to make this `numStages` to be part or a temporary struct in `tr_shader.cpp`, only used when parsing.

After this is done I have some refactor of the `FinishShader()` code to make it more readable and better understandable, along with moving some shader checking code to a dedicated function and reduce boilerplate. Doing that before this is merged would just create useless merge conflicts.

illwieckz:
Well, a vector is probably a bad idea because we would drop the Hunk allocator, so this is probably already the closest to a vector we can do with an Hunk allocator.

src/engine/renderer/tr_shader.cpp
```diff
@@ -5483,27 +5480,29 @@ static shader_t *MakeShaderPermanent()
 
 	tr.numShaders++;
 
-	for ( i = 0; i < newShader->numStages; i++ )
-	{
-		if ( !stages[ i ].active )
-		{
-			break;
-		}
+	ASSERT( numStages < MAX_SHADER_STAGES );
```
slipher, line 5483: Shouldn't that be `<=`?

src/engine/renderer/tr_shader.cpp
```diff
@@ -5483,27 +5480,29 @@ static shader_t *MakeShaderPermanent()
...
+	ASSERT( MAX_SHADER_STAGES < std::numeric_limits<size_t>::max() );
```
slipher, line 5484: These last 2 assertions are silly

src/engine/renderer/tr_shader.cpp
```diff
@@ -6711,9 +6709,18 @@ void R_ListShaders_f()
 		shaderName = shader->name;
 		shaderName += shader->defaultShader ? " (DEFAULTED)" : "";
 
-		for ( int j = 0; j < shader->numStages; j++ )
+		if ( shader->stages == shader->lastStage )
+		{
+			continue;
```
slipher, line 6714: Why?

src/engine/renderer/tr_shader.cpp

src/engine/renderer/tr_shader.cpp
illwieckz: I added code to that condition block, printing the shader information line just before continuing, but with `-` as stage number in the list.

Previously shaders without stages were not listed. This can happen when all stages from a shader are disabled (for example a shader with a single heatHaze stage while heatHaze feature is disabled), or with specific volumetric shaders like fog.

slipher:
LGTM to `renderer: iterate shader stage array directly, avoid array of pointers` and `tr_shader: also list shaders without stages`

I have not reviewed the other one.

illwieckz:
Remaining unmerged commit is now to be reviewed there:

- https://github.com/DaemonEngine/Daemon/pull/1165

PR #1100, opened 2024-04-23
gl_shader: more constness

illwieckz:
gl_shader: more constness


slipher:
Adding `const` to pass-by-value parameters is just noise to me.

PR #1101, opened 2024-04-25
tr_shade: do not use reflective specular code on PBR material

illwieckz:
It is not implemented yet and the shader permutation is disabled.

It now makes possible to run `buildcubemaps` command even when there are PBR materials in the scene, to not crash while attempting to render those PBR materials with a shader permutation that doesn't exist.

To test the reflective specular code, one should:

- load a map with specular maps
- have `r_smp` disabled (there are some crashes when generating cubemaps otherwise)
- type `/buildcubemaps` in console

The rendering is assumed to be wrong (it blinks to black), but the `buildcubemaps` command runs.


VReaperV (commented):
LGTM

PR #1102, opened 2024-04-26
Never sleep with common.framerate.max -2

illwieckz:
Never sleep with `common.framerate.max -2`.

This is useful when benchmarking for performance, to avoid benchmarking the sleep function.

unvanquished_2024-04-22_034803_000


slipher:
How can a busy loop where you just poll for events a million times per second possibly be better than sleeping?

illwieckz:
> How can a busy loop where you just poll for events a million times per second possibly be better than sleeping?

Hmm… right I should better skip the loop entirely. 🤣️

Now done.

slipher:
How is the new version supposed to be different than putting `minMsec = 0`?

illwieckz:
I was just going to say the same. 🤣️

I wrote that yesterday late in the night, it looks like my brain wasn't fully functionnal. 🤡️

slipher:
I don't like how there is only 1 value that enables the reasonable version of unlocked FPS and 2147483647 values that enable the version that clearly violates the system's design assumptions. It should be harder to do it by accident.

BTW I'd be skeptical of any "performance measurements" made with 0 ms frame lengths. Some code could be skipped if it assumes something is already up to date since the time hasn't changed...

illwieckz:
> It should be harder to do it by accident.

Why not make sure this one never does anything unless cheat is enabled?

illwieckz:
> BTW I'd be skeptical of any "performance measurements" made with 0 ms frame lengths.

Well… we may benchmark the performance of the game, or the performance of the environment.

This also helps me notice how hardware pieces, hardware performance profiles, or drivers, or compiler level of optimizations, etc. have impact on how fast the computer go to compute the game (even if doing some part uselessy and redondantly). When reaching some level of fastness of the hardware and software environnent, once I reach the 1000fps cap I can't know if this or that option makes the engine run faster.

I started thinking seriously about implementing this option in our code and just not rely on throwaway patches (like I did for years) when I delivered a computer some weeks ago and out of curiosity I ran the game, set `-1` to `common.framerate.max` and with default configuration (so `medium` profile) I reached the 1000fps cap on 4K… I was like : « It looks like I will not be able to compare with newer hardware anymore »…

slipher:
> > It should be harder to do it by accident.
> 
> Why not make sure this one never does anything unless cheat is enabled?

How do you mean? Max FPS can't be a cheat cvar.

slipher:
I guess having another CHEAT cvar to disable all FPS limitations would work.

illwieckz:
I added a commit that implements my idea.

VReaperV:
> BTW I'd be skeptical of any "performance measurements" made with 0 ms frame lengths. Some code could be skipped if it assumes something is already up to date since the time hasn't changed...

This seems to only really be used for some shader time-based effects.

slipher:
How about making a separate cvar to control `minMsec`?  Get rid of magic values with `common.framerate.max`; 0 or less always means unlocked within the min msec constraints. So the new cvar would have a default value of 3 but you can make it smaller to unlock profiling and bugs.

BTW another interesting way to unlock FPS for profiling is with `common.fixedFrameTime` which makes the engine consider that every frame has the same length for the purposes of its calculations, but allows them to complete arbitrarily fast. This is what I use for generating render test screenshots when I want the screenshot to be made at a specific time but I want it to complete as fast as possible. A difference with setting minMsec to 0 is that the ratio of cgame frames per sgame frame will be the same regardless of how fast your machine is so it will profile a "balanced" mix of computing tasks. Which you may or may not want depending on the goal.

illwieckz:
> How about making a separate cvar to control `minMsec`? Get rid of magic values with `common.framerate.max`;

Yes, that's what I now want to do.

Right now I abuse of `demo.timedemon on` to achieve the same, but I still want a proper way to do it.

slipher:
> BTW another interesting way to unlock FPS for profiling is with `common.fixedFrameTime` which makes the engine consider that every frame has the same length for the purposes of its calculations, but allows them to complete arbitrarily fast.

Seems I was wrong about that. Just tried this to test a new GPU and was stuck at 1000 FPS...

PR #1103, opened 2024-04-26
tr_sky: delete now useless tests and use pStage

illwieckz:
Two commits extracted from #1099:

- https://github.com/DaemonEngine/Daemon/pull/1099

They are in fact the continuation of (already merged)  #1075:

- https://github.com/DaemonEngine/Daemon/pull/1075


slipher:
LGTM

PR #1104, opened 2024-04-26
tr_image: correct bundle image pointer

illwieckz:
I noticed it while working on #1099:

- https://github.com/DaemonEngine/Daemon/pull/1099

On other parts of the code like `tr_cmds.cpp` we already have:

```c++
image = shader->stages[ 0 ]->bundle[ 0 ].image[ 0 ];
```

And every image access are done everywhere else as `bundle[ i ].image`.

So doing `shader->stages[ 0 ]->bundle->image[ 0 ]` looks wrong and likely worked by luck.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -3025,7 +3025,7 @@ void RE_GetTextureSize( int textureID, int *width, int *height )
 		return;
 	}
 
-	baseImage = shader->stages[ 0 ]->bundle->image[ 0 ];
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

VReaperV:
It probably worked because *foo and foo[0] is the same thing, but I agree it's better to make it explicit.

illwieckz:
> It probably worked because `*bundle` and `bundle[0]` is the same thing, but I agree it's better to make it explicit.

Yeah, someone probably got a warning, turned `.` into `->` and got it compiling, by luck it meant doing `*bundle` so it was the same as `bundle[0]`, and voilà.

PR #1106, opened 2024-04-29
cg_api: do nothing in trap_R_AddPolysToScene if array size is zero

illwieckz:
May be a better fix for:

than:

- https://github.com/Unvanquished/Unvanquished/pull/2976

Being implemented in engine shared code means the crash would be prevented for any custom game.

I now wonder if we should not add such check for any other trap calls allocating a `vector` of `polyVert_t`.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -3049,6 +3049,13 @@ int numTextures = 0;
 
 qhandle_t RE_GenerateTexture( const byte *pic, int width, int height )
 {
+	size_t size = width * height;
+
+	if ( !size ) {
+		Log::Warn("RE_GenerateTexture: image size %dx%d is 0", width, height );
+		return 0;
+	}
+
 	const char *name = va( "rocket%d", numTextures++ );
 	R_SyncRenderThread();
 
--- a/src/shared/client/cg_api.cpp
+++ b/src/shared/client/cg_api.cpp
@@ -319,6 +319,11 @@ void trap_R_AddRefEntityToScene( const refEntity_t *re )
 
 void trap_R_AddPolyToScene( qhandle_t hShader, int numVerts, const polyVert_t *verts )
 {
+	if (!numVerts)
+	{
+		return;
+	}
+
 	std::vector<polyVert_t> myverts(numVerts);
 	memcpy(myverts.data(), verts, numVerts * sizeof(polyVert_t));
 	cmdBuffer.SendMsg<Render::AddPolyToSceneMsg>(hShader, myverts);
@@ -343,6 +348,11 @@ void trap_R_AddPolysToScene( qhandle_t hShader, int numVerts, const polyVert_t *
 
 void trap_R_Add2dPolysIndexedToScene( polyVert_t* polys, int numPolys, int* indexes, int numIndexes, int trans_x, int trans_y, qhandle_t shader )
 {
+	if (!numIndexes)
+	{
+		return;
+	}
+
 	std::vector<polyVert_t> mypolys(numPolys);
 	std::vector<int> myindices(numIndexes);
 	memcpy(mypolys.data(), polys, numPolys * sizeof( polyVert_t ) );
--- a/src/shared/client/cg_api.cpp
+++ b/src/shared/client/cg_api.cpp
@@ -326,8 +326,18 @@ void trap_R_AddPolyToScene( qhandle_t hShader, int numVerts, const polyVert_t *v
 
 void trap_R_AddPolysToScene( qhandle_t hShader, int numVerts, const polyVert_t *verts, int numPolys )
 {
-	std::vector<polyVert_t> myverts(numVerts * numPolys);
-	memcpy(myverts.data(), verts, numVerts * numPolys * sizeof(polyVert_t));
+	size_t size = numVerts * numPolys;
+
+	if (!size)
+	{
+		return;
+	}
+
+	std::vector<polyVert_t> myverts(size);
+	memcpy(myverts.data(), verts, size * sizeof(polyVert_t));
+
+	/* Known to crash on Clang ≥ 14 in non-Debug native build if size is 0:
+	https://github.com/Unvanquished/Unvanquished/issues/2682 */
 	cmdBuffer.SendMsg<Render::AddPolysToSceneMsg>(hShader, myverts, numVerts, numPolys);
 }
 
--- a/src/shared/client/cg_api.cpp
+++ b/src/shared/client/cg_api.cpp
@@ -568,6 +568,7 @@ void trap_R_GetTextureSize( qhandle_t handle, int *x, int *y )
 
 qhandle_t trap_R_GenerateTexture( const byte *data, int x, int y )
 {
+	ASSERT( x * y );
 	qhandle_t handle;
 	std::vector<byte> mydata(x * y * 4);
 	memcpy(mydata.data(), data, x * y * 4 * sizeof( byte ) );


illwieckz:
Is this a slow (IPC) or fast (shared mem) trap call?

src/shared/client/cg_api.cpp
```diff
@@ -326,8 +326,14 @@ void trap_R_AddPolyToScene( qhandle_t hShader, int numVerts, const polyVert_t *v
 
 void trap_R_AddPolysToScene( qhandle_t hShader, int numVerts, const polyVert_t *verts, int numPolys )
 {
-	std::vector<polyVert_t> myverts(numVerts * numPolys);
-	memcpy(myverts.data(), verts, numVerts * numPolys * sizeof(polyVert_t));
+	size_t size = (numVerts * numPolys);
+	if (!size) { return; }
+
+	/* Known to crash on Clang ≥ 14 in non-Debug native build:
```
slipher, line 332: The comment needs more context to make sense--it doesn't say anything about the condition that something is 0. Also the spacing is arranged so as to imply it crashes in the std::vector constructor before getting to the IPC code?

DolceTriade:
> Is this a slow (IPC) or fast (shared mem) trap call?

shmem

src/shared/client/cg_api.cpp
illwieckz: > The comment needs more context to make sense--it doesn't say anything about the condition that something is 0. 

Right, I will edit that.

> Also the spacing is arranged so as to imply it crashes in the std::vector constructor before getting to the IPC code?

According to this comment:

https://github.com/Unvanquished/Unvanquished/issues/2682#issuecomment-2080398702

It doesn't crash before calling `cmdBuffer.SendMsg()`. I can move the comment too.

illwieckz: Done for both.

illwieckz:
Should I add similar tests to other trap calls sending `std::vector`?

DolceTriade:
> Should I add similar tests to other trap calls sending std::vector?

Only if it is valid to be able to send 0 sized vectors.

slipher:
LGTM

illwieckz:
I verified that if in the game I also pass a `0` size to `trap_R_AddPolyToScene` (another similar function), I get a crash as well. Since this code is engine shared code, meant to be included in games we don't necessarily have control over, it's probably better to add some checks to the other functions too.

As an example I modified the `DrawTracer` in Unvanquished.

I also verified that I can reproduce the crash with a `vector` of `int`, not only a vector of `polyVert_t`.

Most of those functions are “add something to scene”, so if the amount of things i 0, the thing should just not be added to the scene, so it looks a good idea to return early in such situation.

So I also added a check to those other functions.

illwieckz:
I have not modified the `trap_R_GenerateTexture` function because this one is a bit special. I noticed that a nasty game implementing its own `trap_R_GenerateTexture` to avoid any check and make sure to pass a `0` image size to the engine can intentionnally crash the engine (even from a nexe). So, unless we add a check in the non-shared engine code itself to prevent from intentionnal harm, protecting against a compiler bug is the least of our concerns.

@slipher Does it make sense to prevent such intentionnal harm? Or is there other ways to do this we can't fix and so fixing this one would be useless anyway?

illwieckz:
@slipher I added more significant commits since your LGTM.

src/engine/renderer/tr_image.cpp
```diff
@@ -3049,6 +3049,13 @@ int numTextures = 0;
 
 qhandle_t RE_GenerateTexture( const byte *pic, int width, int height )
 {
+	size_t size = width * height;
+
+	ASSERT( !size );
```
DolceTriade, line 3054: I thought ASSERT fails if the condition is false?

src/engine/renderer/tr_image.cpp
```diff
@@ -3049,6 +3049,13 @@ int numTextures = 0;
...
+	if ( !size ) {
```
DolceTriade, line 3055: Let's be consistent about style.

let's just do
```
if ( !size )
{
    return;
}
```

or

```
if ( !size ) return;
```

I think for single line if statements, we should avoid brackets.

src/shared/client/cg_api.cpp
```diff
@@ -319,20 +319,30 @@ void trap_R_AddRefEntityToScene( const refEntity_t *re )
...
 	std::vector<polyVert_t> myverts(numVerts);
 	memcpy(myverts.data(), verts, numVerts * sizeof(polyVert_t));
 	cmdBuffer.SendMsg<Render::AddPolyToSceneMsg>(hShader, myverts);
 }
 
 void trap_R_AddPolysToScene( qhandle_t hShader, int numVerts, const polyVert_t *verts, int numPolys )
 {
-	std::vector<polyVert_t> myverts(numVerts * numPolys);
-	memcpy(myverts.data(), verts, numVerts * numPolys * sizeof(polyVert_t));
+	size_t size = (numVerts * numPolys);
```
DolceTriade, line 331: nit: no ()

src/shared/client/cg_api.cpp
```diff
@@ -548,9 +558,12 @@ void trap_R_GetTextureSize( qhandle_t handle, int *x, int *y )
 
 qhandle_t trap_R_GenerateTexture( const byte *data, int x, int y )
 {
+	size_t size = x * y * 4;
```
DolceTriade, line 561: I think we should assert on generating a zero l/w texture. Seems like something we shouldn't want?

src/shared/client/cg_api.cpp
```diff
@@ -554,9 +558,12 @@ void trap_R_GetTextureSize( qhandle_t handle, int *x, int *y )
...
+	if (!size) { return 0; }
+
 	qhandle_t handle;
-	std::vector<byte> mydata(x * y * 4);
-	memcpy(mydata.data(), data, x * y * 4 * sizeof( byte ) );
+	std::vector<byte> mydata(size);
```
slipher, line 565: The size is now wrong.

src/shared/client/cg_api.cpp
```diff
@@ -339,6 +341,8 @@ void trap_R_AddPolysToScene( qhandle_t hShader, int numVerts, const polyVert_t *
 
 void trap_R_Add2dPolysIndexedToScene( polyVert_t* polys, int numPolys, int* indexes, int numIndexes, int trans_x, int trans_y, qhandle_t shader )
 {
+	if (!numPolys || !numIndexes) { return; }
```
slipher, line 344: This could be just `!numIndexes`. If there are indexes but no polies, it's a mistake rather than a no-op (the indices are necessarily out of bounds).

src/engine/renderer/tr_image.cpp
illwieckz: Done.

src/engine/renderer/tr_image.cpp
illwieckz: Fixed.

src/shared/client/cg_api.cpp
illwieckz: Done.

src/shared/client/cg_api.cpp
illwieckz: Done.

src/shared/client/cg_api.cpp
illwieckz: Done.

src/shared/client/cg_api.cpp
illwieckz: This code is not modified anymore, I just added an ASSERT before running it instead.

src/engine/renderer/tr_image.cpp
slipher: If you write `ASSERT(size); if (!size) {...}`, with our ASSERT implementation, the compiler is allowed to optimize out the test. If you want to ensure the condition is always caught, better to replace the assert+if with a Log::Warn or Sys::Drop.

illwieckz: The primary purpose of this patch is that a third-party individual cannot host a server providing an autodownloaded package that would either crash or abort the user's own game.

A `Log::Warn` would do the job if we want to report the error, but game execution should continue.

illwieckz: Modified with a `Log::Warn`.

slipher: > The primary purpose of this patch is that a third-party individual cannot host a server providing an autodownloaded package that would either crash or abort the user's own game.

If this is about catching mistakes made by the gamelogic, it should not use ASSERT at all. An ASSERT in the engine should be used when the engine detects violations of the assumptions of its own behavior.

illwieckz: Then it's another good reason to use a `Log::Warn`.

slipher:
LGTM

PR #1107, opened 2024-04-29
renderer: expression_t is always allocated as part of structs, ext->numOps always 0 if ext is invalid

illwieckz:
As far as I know,

- `expression_t` is always allocated as part of structs (`texModInfo_t`, `shaderStage_t`),
- `ext->numOps` is always `0` if `ext` is invalid.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -941,8 +941,6 @@ enum class realtimeLightingRenderer_t { LEGACY, TILED };
 	{
 		expOperation_t ops[ MAX_EXPRESSION_OPS ];
 		uint8_t        numOps;
-
-		bool       active; // no parsing problems
 	};
 
 	struct waveForm_t
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -940,7 +940,7 @@ enum class realtimeLightingRenderer_t { LEGACY, TILED };
 	struct expression_t
 	{
 		expOperation_t ops[ MAX_EXPRESSION_OPS ];
-		uint8_t numOps;
+		size_t numOps;
 	};
 
 	struct waveForm_t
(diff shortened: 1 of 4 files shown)


src/engine/renderer/tr_shader.cpp
```diff
@@ -686,31 +686,29 @@ ParseExpression
...
 
 	// convert stack
 	expOperation_t tmpOps[ MAX_EXPRESSION_OPS ];
-	int            numTmpOps;
-
-	numInFixOps = 0;
-	numTmpOps = 0;
+	int numTmpOps = 0;
 
+	// A ext->numOps equals to 0 means empty or invalid expression.
```
slipher, line 698: Suggesting there is such a thing as a valid empty expression, but I don't suppose that's the case?

illwieckz: Maybe an empty expression is invalid, right.

src/engine/renderer/tr_local.h
```diff
@@ -1049,9 +1049,7 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 	struct expression_t
 	{
 		expOperation_t ops[ MAX_EXPRESSION_OPS ];
-		uint8_t        numOps;
-
-		bool       active; // no parsing problems
+		uint8_t numOps;
```
illwieckz, line 1052: Should we use `int`?

src/engine/renderer/tr_shader.cpp
slipher: I took a closer look and in fact there are "valid" empty expressions, because the parser is buggy, e.g. `((()))`. Apparently they evaluate to 0. It's not something that should have been accepted though.

src/engine/renderer/tr_local.h
slipher: Yes

illwieckz: I now use `size_t` for `numOps` and related things everywhere.

slipher:
LGTM

PR #1108, opened 2024-04-29
Fix dynamic light blockiness and strength

### LINKED ISSUE #791: Dynamic light blockiness (tile edges visible)
Using the tiled renderer, the area covered by a dynamic light can have visible, blocky edges. The issue can be observed by firing a single pulse rifle shot. Here's a screenshot by illwieckz: [light blockiness](https://dl.illwieckz.net/b/unvanquished/bugs/light-blockiness/unvanquished_2023-01-19_145745_000.jpg)

The tiled lighting works in two phases:
1. In the `lighttile` shader, compute the set of lights which is believed to affect each tile. The assumption is that tiles sufficiently far away should be close enough to completely dark not to matter. 
2. When rendering a surface, apply each light in the list for the relevant tile.

It seems that the assumption in (1) is wrong and some of the tiles considered to be not affected actually would have had a significant amount of light. Two obvious approaches to fix this:
(a) Change the calculation of which tiles are affected to rule in more tiles
(b) Change the falloff curve of brightness vs. distance so that the brightness is actually close enough to 0 at the edges

VReaperV:
Sets a minimum radius for light culling to avoid blocky artifacts. Still sort of a hack, but it works better than what we had.
Also made attenuation quadratic rather than linear for dynamic lights since that is more correct without broken gamma (previous incorrect lighting was essentially doing distance ^ 4.2 due to gamma correction). In the same pr because without the correct attenuation the blockiness is still there.

Before and after for light culling:
unvanquished_2024-04-30_002814_000
unvanquished_2024-04-30_002707_000

I've chosen the 2.0 * 32.0 value for the minimum radius because it fixes the blockiness issues with the smallest scale lights that we have. Simply always increasing radius results in too many tiles getting the light source.

This also fixes the same issue with https://users.unvanquished.net/~modi/res-psaw_dev%2Bishq1.dpk, however it's more difficult to capture a screenshot of that.

Fixes #791.

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -186,13 +186,19 @@ void computeDynamicLight( int idx, vec3 P, vec3 normal, vec3 viewDir, vec4 diffu
   if( color_type.w == 0.0 ) {
     // point light
     L = center_radius.xyz - P;
-    attenuation = 1.0 / (1.0 + 8.0 * length(L) / center_radius.w);
+    // 2.57 ~= 8.0 ^ ( 1.0 / 2.2 ), adjusted after overbright changes
+    float t = 1.0 + 2.57 * length(L) / center_radius.w;
+    // Quadratic attenuation function instead of linear because of overbright changes
+    attenuation = 1.0 / ( t * t );
     L = normalize(L);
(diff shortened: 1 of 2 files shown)


DolceTriade:
This fixes the issues for me too!!

DolceTriade (approved):

PR #1110, opened 2024-05-02
common: also do pedantic shutdown when running Clang sanitizers

illwieckz:
Also detect MemorySanitizer and ThreadSanitizer.

Without this, the linker complains about `new` and `delete` being redefined when linking with `-fsanitize=leak`, `-fsanitize=memory` or `-fsanitize=thread`.

Previously we were only caring about linking with `-fsanitize=address`.

### DIFF
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -103,11 +103,21 @@ int CountTrailingZeroes(unsigned long long x);
 // Work around lack of constexpr
 #define CONSTEXPR constexpr
 
-#ifdef __SANITIZE_ADDRESS__ // Detects GCC asan
+#if defined(__SANITIZE_ADDRESS__) // Detects GCC and MSVC AddressSanitizer
 #   define USING_ADDRESS_SANITIZER
+#   define USING_SANITIZER
+#elif defined(__SANITIZE_THREAD__) // Detects GCC ThreadSanitizer
+#   define USING_SANITIZER
 #elif defined(__has_feature)
-#   if __has_feature(address_sanitizer) // Detects Clang asan
+#   if __has_feature(address_sanitizer) // Detects Clang AddressSanitizer
 #       define USING_ADDRESS_SANITIZER
+#       define USING_SANITIZER
+#   elif __has_feature(leak_sanitizer) // Detects Clang LeakSanitizer
+#       define USING_SANITIZER
+#   elif __has_feature(memory_sanitizer) // Detects Clang MemorySanitizer
+#       define USING_SANITIZER
+#   elif __has_feature(thread_sanitizer) // Detects Clang ThreadSanitizer
+#       define USING_SANITIZER
 #   endif
 #endif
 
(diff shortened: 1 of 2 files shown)


src/common/Compiler.h
```diff
@@ -106,7 +106,11 @@ int CountTrailingZeroes(unsigned long long x);
 #ifdef __SANITIZE_ADDRESS__ // Detects GCC asan
 #   define USING_ADDRESS_SANITIZER
 #elif defined(__has_feature)
-#   if __has_feature(address_sanitizer) // Detects Clang asan
+#   if __has_feature(address_sanitizer) // Detects Clang AddressSanitizer
+#       define USING_ADDRESS_SANITIZER
+#   elif __has_feature(memory_sanitizer) // Detects Clang MemorySanitizer
+#       define USING_ADDRESS_SANITIZER
```
DolceTriade, line 112: nit: does this need to be renamed?

DolceTriade (approved):

illwieckz:
I also added detection for `-fsanitize=leak`. This one is part of `-fsanitize=address` and can be called separately (it is a subset).

src/common/Compiler.h
illwieckz: For now it works, duplicating all those definitions would be very noisy.

The primary usage is to avoid redefining `new` and `delete`, so it is to always do the same on that side and having 4 defininitions to prevent redefining `new` and `delete` looks overkill.

The secondary usage is for to add `ATTRIBUTE_NO_SANITIZE_ADDRESS` to some function, I have not tested if such function would work on other sanitizers without tht.

DolceTriade (approved):

slipher:
It seems a little wrong to enable the no-sanitize-address attribute because a different, unaffected sanitizer is running. I propose an alternative #1112 to get rid of the `operator new` overrides.

illwieckz:
> It seems a little wrong to enable the no-sanitize-address attribute because a different, unaffected sanitizer is running. I propose an alternative #1112 to get rid of the `operator new` overrides.

Right, this looks good.

Anyway we may still have to set a common define when running sanitizers, because we also need to enable pendantic shutdown when running the Clang sanitizers. See the extra commit I pushed.

slipher:
> Anyway we may still have to set a common define when running sanitizers, because we also need to enable pendantic shutdown when running the Clang sanitizers.

Good point. This can be a separate macro instead of hijacking USING_ADDRESS_SANITIZER which has a specific purpose.

illwieckz:
I defined `USING_MEMORY_SANITIZER` instead.

slipher:
Seems confusing given that clang has one named MemorySanitizer. How about USING_ANY_SANITIZER?

DolceTriade:
I would remove the _ANY_, but don't feel strongly enough to demand it either way.

illwieckz:
> Seems confusing given that clang has one named MemorySanitizer. How about USING_ANY_SANITIZER?

> I would remove the _ANY_, but don't feel strongly enough to demand it either way.

I've done both.

DolceTriade (approved):

PR #1112, opened 2024-05-03
Use std::set_new_handler to send OOM message

slipher:
Instead of defining custom new and delete to ensure that an allocation failure results in a fatal error with the "Out of memory", use std::set_new_handler to accomplish the same. This std lib facility allows setting a custom behavior (instead of throwing std::bad_alloc) on failure of operator new.

This is preferable since we don't have to worry about proliferation of `new` variants or messing up the sanitizers.

I tested on MSVC and Linux. The former works, while the latter gets hit by the OOM killer instead of throwing an exception as before.

### DIFF
--- a/src/common/System.cpp
+++ b/src/common/System.cpp
@@ -425,26 +425,15 @@ void GenRandomBytes(void* dest, size_t size)
 #endif
 }
 
-} // namespace Sys
-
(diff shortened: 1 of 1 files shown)


illwieckz (approved):
LGTM.

PR #1113, opened 2024-05-04
NUKE growList_t

slipher:
This was an equivalent of `std::vector<void*>` in C.


illwieckz (approved):
LGTM. ❤️

illwieckz:
I have seen some instances of:

```c++
for ( size_t i = 0; i < tr.images.size(); i++ )
{
	image = tr.images[ i ];
```

And I wondered if this would have been doable:

```c++
for( image_t *image : tr.images )
```

But maybe `i` is also used in some places?

But that's not blocking.

illwieckz:
Some sign comparison warning caught by CI, but otherwise that looks very good to me:

```
src/engine/renderer/tr_bsp.cpp:472:22: warning: comparison of integers of different signs: 'int' and 'unsigned long' [-Wsign-compare]
                for ( int i = 0; i < ARRAY_LEN( extensions ); i++ )
                                 ~ ^ ~~~~~~~~~~~~~~~~~~~~~~~
src/engine/renderer/tr_bsp.cpp:6838:57: error: format specifies type 'int' but the argument has type 'size_t' (aka 'unsigned long') [-Werror,-Wformat]
                cubeProbe->cubemap = R_AllocImage( va( "_autoCube%d", j ), false );
                                                                 ~~   ^
                                                                 %zu
src/engine/renderer/tr_bsp.cpp:6615:18: warning: comparison of integers of different signs: 'unsigned long' and 'int' [-Wsign-compare]
                if ( ( j + 1 ) >= nextTicCount )
                       ~~~~~   ^  ~~~~~~~~~~~~
```

slipher:
I did use for-each in all cases that didn't use the index for something.

PR #1114, opened 2024-05-04
cg_api: implement `trap_CM_BatchMarkFragments` to process all impact marks in one call

illwieckz:
Engine counterpart of:

- https://github.com/Unvanquished/Unvanquished/pull/2981

### DIFF
--- a/src/engine/client/cg_api.h
+++ b/src/engine/client/cg_api.h
@@ -86,9 +86,26 @@ enum class MouseMode
 	SystemCursor, // The input is sent as positions, the cursor should be rendered by the system
 };
 
+using markMsgInput_t = std::pair<
+	std::vector<std::array<float, 3>>, // points
+	std::array<float, 3> // projection
+>;
+
+using markMsgOutput_t = std::pair<
+	std::vector<std::array<float, 3>>, // points
+	std::vector<markFragment_t>
+>;
+
 void            trap_SendClientCommand( const char *s );
 void            trap_UpdateScreen();
 int             trap_CM_MarkFragments( int numPoints, const vec3_t *points, const vec3_t projection, int maxPoints, vec3_t pointBuffer, int maxFragments, markFragment_t *fragmentBuffer );
+
+void trap_CM_BatchMarkFragments(
+	unsigned maxPoints,
+	unsigned maxFragments,
+	const std::vector<markMsgInput_t> &markMsgInput,
+	std::vector<markMsgOutput_t> &markMsgOutput );
+
 void            trap_S_StartSound( vec3_t origin, int entityNum, soundChannel_t entchannel, sfxHandle_t sfx );
 void            trap_S_StartLocalSound( sfxHandle_t sfx, soundChannel_t channelNum );
 void            trap_S_ClearLoopingSounds( bool killall );
--- a/src/engine/client/cl_cgame.cpp
+++ b/src/engine/client/cl_cgame.cpp
@@ -1102,6 +1102,45 @@ void CGameVM::QVMSyscall(int syscallNum, Util::Reader& reader, IPC::Channel& cha
 			});
 			break;
 
+		case CG_CM_BATCHMARKFRAGMENTS:
+			IPC::HandleMsg<CMBatchMarkFragments>(channel, std::move(reader), [this] (
+				unsigned maxPoints,
+				unsigned maxFragments,
+				const std::vector<markMsgInput_t>& inputs,
+				std::vector<markMsgOutput_t>& outputs)
+			{
+				outputs.reserve(inputs.size());
+				std::vector<std::array<float, 3>> pointBuf(maxPoints);
+				std::vector<markFragment_t> fragmentBuf(maxFragments);
+
+				for (const markMsgInput_t& input : inputs)
+				{
+					auto& inputPoints = input.first;
+					auto& projection = input.second;
+					size_t numFragments = re.MarkFragments(
+						inputPoints.size(),
+						reinterpret_cast<const vec3_t*>(inputPoints.data()),
+						projection.data(),
+						maxPoints,
+						reinterpret_cast<float*>(pointBuf.data()),
+						maxFragments,
+						fragmentBuf.data());
+					size_t numPoints;
+					if (numFragments == 0) {
+						numPoints = 0;
+					} else {
+						// HACK: assume last fragment is last
+						const markFragment_t& lastFragment = fragmentBuf[numFragments - 1];
+						numPoints = lastFragment.firstPoint + lastFragment.numPoints;
+					}
+					outputs.emplace_back(
+						std::vector<std::array<float, 3>>(pointBuf.data(), pointBuf.data() + numPoints),
+						std::vector<markFragment_t>(fragmentBuf.data(), fragmentBuf.data() + numFragments)
+					);
+				}
+			});
+			break;
+
 		case CG_GETCURRENTSNAPSHOTNUMBER:
 			IPC::HandleMsg<GetCurrentSnapshotNumberMsg>(channel, std::move(reader), [this] (int& number, int& serverTime) {
 				number = cl.snap.messageNum;
--- a/src/engine/client/cg_msgdef.h
+++ b/src/engine/client/cg_msgdef.h
@@ -132,6 +132,7 @@ enum cgameImport_t
   CG_SENDCLIENTCOMMAND,
   CG_UPDATESCREEN,
   CG_CM_MARKFRAGMENTS,
+  CG_CM_BATCHMARKFRAGMENTS,
   CG_GETCURRENTSNAPSHOTNUMBER,
   CG_GETSNAPSHOT,
   CG_GETCURRENTCMDNUMBER,
@@ -245,6 +246,15 @@ using UpdateScreenMsg = IPC::SyncMessage<
 using CMMarkFragmentsMsg = IPC::SyncMessage<
 	IPC::Message<IPC::Id<VM::QVM, CG_CM_MARKFRAGMENTS>, std::vector<std::array<float, 3>>, std::array<float, 3>, int, int>,
 	IPC::Reply<std::vector<std::array<float, 3>>, std::vector<markFragment_t>>
+>;
+using CMBatchMarkFragments = IPC::SyncMessage<
+	IPC::Message<
+		IPC::Id<VM::QVM, CG_CM_BATCHMARKFRAGMENTS>,
+		unsigned,
+		unsigned,
+		std::vector<markMsgInput_t>>,
+	IPC::Reply<
+		std::vector<markMsgOutput_t>>
 >;
 // TODO send all snapshots at the beginning of the frame
 using GetCurrentSnapshotNumberMsg = IPC::SyncMessage<
--- a/src/shared/client/cg_api.cpp
+++ b/src/shared/client/cg_api.cpp
@@ -64,6 +64,18 @@ int trap_CM_MarkFragments( int numPoints, const vec3_t *points, const vec3_t pro
 	return myfragmentBuffer.size();
 }
 
+void trap_CM_BatchMarkFragments(
+	unsigned maxPoints, //per mark
+	unsigned maxFragments, //per mark
+	const std::vector<markMsgInput_t> &markMsgInput,
+	std::vector<markMsgOutput_t> &markMsgOutput )
+{
+	if (!markMsgInput.empty())
+	{
+		VM::SendMsg<CMBatchMarkFragments>(maxPoints, maxFragments, markMsgInput, markMsgOutput);
+	}
+}
+
 void trap_GetCurrentSnapshotNumber( int *snapshotNumber, int *serverTime )
 {
 	VM::SendMsg<GetCurrentSnapshotNumberMsg>(*snapshotNumber, *serverTime);


illwieckz:
Hmm, something very annoying is happening, the code is much slower with a single trap call than with many.

Is performance strongly tied to size of IPC messages?

I currently send an `array<MAX>`, should I send `vector(num`) instead?

illwieckz:
The PR is ready, commits will be squashed, no need to review them one by one.

src/engine/client/cg_api.h
```diff
@@ -86,9 +86,30 @@ enum class MouseMode
...
+struct fragmentPacket_t
+{
+	std::array<std::array<float, 3>, 4> originalPoints;
+	std::array<float, 3> projection;
+};
+
+struct fragmentBuffer_t
+{
+	std::array<std::array<float, 3>, MAX_MARK_POINTS> markPoints;
+	std::array<markFragment_t, MAX_MARK_POINTS> markFragments;
```
slipher, line 101: Is this supposed to be MAX_MARK_FRAGMENTS?

src/engine/client/cg_api.h
```diff
@@ -86,9 +86,30 @@ enum class MouseMode
 	SystemCursor, // The input is sent as positions, the cursor should be rendered by the system
 };
 
+#define MAX_MARK_FRAGMENTS 128
+#define MAX_MARK_POINTS 384
+
+struct fragmentPacket_t
```
slipher, line 92: The structs are related to the "mark" feature and so ought to have "mark" in their names. This sounds like something about the network.

src/engine/client/cl_cgame.cpp
```diff
@@ -1102,6 +1102,32 @@ void CGameVM::QVMSyscall(int syscallNum, Util::Reader& reader, IPC::Channel& cha
...
+				{
+					const fragmentPacket_t *fragmentPacket = &fragmentPacketsArray[ m ];
+					fragmentBuffer_t *fragmentBuffer = &fragmentBuffersArray[ m ];
+
+					fragmentBuffer->numFragments = re.MarkFragments(
+						4,
+						(const vec3_t*) fragmentPacket->originalPoints.data(),
+						fragmentPacket->projection.data(),
+						MAX_MARK_POINTS,
+						fragmentBuffer->markPoints.data()[ 0 ].data(),
```
slipher, line 1124: `data()[ 0 ]` is just `[ 0 ]`?

slipher (commented):
Is having a 4KB large fragmentBuffer_t struct really faster than using variable-size arrays? Allocation can be slow but having to split up the socket reads/writes (since the amount of data is bigger than the kernel buffer) could be slower.

illwieckz:
> Is having a 4KB large fragmentBuffer_t struct really faster than using variable-size arrays? Allocation can be slow but having to split up the socket reads/writes (since the amount of data is bigger than the kernel buffer) could be slower.

There can definitely be further optimizations done here. I have verified that on my end using fixed-size arrays for input and output of the single trap call  would make the engine go from 500 fps to 300 fps when rendering ATCHS default alien spectator scene in 2K on my system, and then being even more slower than doing single trap calls for every mark individually. So I measured in terme of slowness: `single trap call with large array > may trap calls with small data > single trap call with small array`.

So yes IPC message size is critical, so anything reducing message size is welcome.

src/engine/client/cl_cgame.cpp
illwieckz:
```
src/engine/client/cl_cgame.cpp:1124:80: error: invalid types ‘<unresolved overloaded function type>[int]’ for array subscript
 1124 |                                                 fragmentBuffer->markPoints.data[ 0 ].data(),
      |                                                                                ^
```

illwieckz: From to-be-squashed commits in:

- https://github.com/Unvanquished/Unvanquished/pull/2981

Here are some of the micro-rewriting-steps leading to this code:

```diff
diff --git a/src/cgame/cg_marks.cpp b/src/cgame/cg_marks.cpp
index 507fabc63..dba1eaef0 100644
--- a/src/cgame/cg_marks.cpp
+++ b/src/cgame/cg_marks.cpp
@@ -183,7 +183,7 @@ void CG_ProcessImpactMarks()
 		byte           colors[ 4 ];
 		int            numFragments;
 		std::array<markFragment_t, MAX_MARK_FRAGMENTS> markFragments;
-		vec3_t         markPoints[ MAX_MARK_POINTS ];
+		std::array<vec3_t, MAX_MARK_POINTS> markPoints;
 		vec3_t         projection;
 
 		//if ( markTotal >= MAX_MARK_POLYS ) {
@@ -210,7 +210,7 @@ void CG_ProcessImpactMarks()
 		// get the fragments
 		VectorScale( dir, -20, projection );
 		numFragments = trap_CM_MarkFragments( 4, ( const vec3_t * ) originalPoints,
-			projection, MAX_MARK_POINTS, markPoints[ 0 ],
+			projection, MAX_MARK_POINTS, markPoints.data()[ 0 ],
 			MAX_MARK_FRAGMENTS, markFragments.data() );
 
 		colors[ 0 ] = red * 255;
```

```diff
diff --git a/src/cgame/cg_marks.cpp b/src/cgame/cg_marks.cpp
index dba1eaef0..10eca5630 100644
--- a/src/cgame/cg_marks.cpp
+++ b/src/cgame/cg_marks.cpp
@@ -183,7 +183,7 @@ void CG_ProcessImpactMarks()
 		byte           colors[ 4 ];
 		int            numFragments;
 		std::array<markFragment_t, MAX_MARK_FRAGMENTS> markFragments;
-		std::array<vec3_t, MAX_MARK_POINTS> markPoints;
+		std::array<std::array<float, 3>, MAX_MARK_POINTS> markPoints;
 		vec3_t         projection;
 
 		//if ( markTotal >= MAX_MARK_POLYS ) {
@@ -210,7 +210,7 @@ void CG_ProcessImpactMarks()
 		// get the fragments
 		VectorScale( dir, -20, projection );
 		numFragments = trap_CM_MarkFragments( 4, ( const vec3_t * ) originalPoints,
-			projection, MAX_MARK_POINTS, markPoints.data()[ 0 ],
+			projection, MAX_MARK_POINTS, markPoints.data()[ 0 ].data(),
 			MAX_MARK_FRAGMENTS, markFragments.data() );
 
 		colors[ 0 ] = red * 255;
```

(I modified the spacing in the diff so it fits in comment width)

src/engine/client/cg_api.h
illwieckz: Right.

src/engine/client/cl_cgame.cpp
```diff
@@ -1102,6 +1102,32 @@ void CGameVM::QVMSyscall(int syscallNum, Util::Reader& reader, IPC::Channel& cha
...
+
+				fragmentBuffersArray.resize(numMarks);
+
+				for ( size_t m = 0; m < numMarks; m++ )
+				{
+					const fragmentPacket_t *fragmentPacket = &fragmentPacketsArray[ m ];
+					fragmentBuffer_t *fragmentBuffer = &fragmentBuffersArray[ m ];
+
+					fragmentBuffer->numFragments = re.MarkFragments(
+						4,
```
illwieckz, line 1120: Note: The rewriting of `re.MarkFragments` to not uselessly expects global constants as input like `4`, `MAX_MARK_POINTS`, and `MAX_MARK_FRAGMENTS` will be done in a future PR after NUKING the legacy `trap_CM_MarkFragments` function where the constants were written in game and sent to engine instead.

illwieckz:
Note: The rewriting of `re.MarkFragments` to not uselessly expects global constants as input like `4`, `MAX_MARK_POINTS`, and `MAX_MARK_FRAGMENTS` will be done in a future PR after NUKING the legacy `trap_CM_MarkFragments` function where the constants were written in game and sent to engine instead.

illwieckz:
> Is having a 4KB large fragmentBuffer_t struct really faster than using variable-size arrays? Allocation can be slow but having to split up the socket reads/writes (since the amount of data is bigger than the kernel buffer) could be slower.

The content of `fragmentBuffer_t` is the return from `re.MarkFragments`, optimizing the size of this return will also be done after rewriting of `re.MarkFragments`, which should be done after NUKING the legacy `trap_CM_MarkFragments` function, and I expect such NUKING PR to come later.

illwieckz:
We can also rewrite `re.MarkFragments` to take `markPoints.data()` as input instead of `markPoints.data()[ 0 ].data()`, but yet again this would happen only after legacy `trap_CM_MarkFragments` is NUKED.

src/engine/client/cg_api.h
illwieckz: I renamed `fragmentPacket_t` and `fragmentBuffer_t` to `markMsgInput_t` and `markMsgOutput_t`.

VReaperV:
This works well for me, there's no longer a significant performance difference between `/cg_marks on` and `/cg_marks off` (still depends on the amount of buildings of course, but that's expected).

illwieckz:
Yeah, this is still dependent on the amount of marks because as soon as the array is too large to fit into the kernel buffer the message will be split, but the split will only be happening when such packet size is outsized. But this means usual scene will likely only do 1 or 2 message instead of doing 1 message per mark.

Every player blob shadow is a mark, every alien creep is a mark, every bullet impact is a mark, every grenade and flamer burnt areas are marks. The number of marks increase very quickly in a game, and this reduce a lot the amount of IPC messages done for them! 

It would be better to entirely avoid the engine-client discussion there by fully moving the related code into the cgame, but doing this kind of vectorization is already a big win while not being too much intrusive, and we still rely on well-tested code.

DolceTriade:
This sounds like it could use the shmem command queue too in the future so maybe everything could fit in one area?

slipher:
> > Is having a 4KB large fragmentBuffer_t struct really faster than using variable-size arrays? Allocation can be slow but having to split up the socket reads/writes (since the amount of data is bigger than the kernel buffer) could be slower.
> 
> There can definitely be further optimizations done here. I have verified that on my end using fixed-size arrays for input and output of the single trap call would make the engine go from 500 fps to 300 fps when rendering ATCHS default alien spectator scene in 2K on my system, and then being even more slower than doing single trap calls for every mark individually. So I measured in terme of slowness: `single trap call with large array > may trap calls with small data > single trap call with small array`.
> 
> So yes IPC message size is critical, so anything reducing message size is welcome.

The obvious thing to do would be 
```
struct markMsgOutput_t
{
	std::vector<std::array<float, 3>> markPoints;
	std::vector<markFragment_t> markFragments;
};
```

To be lazy and avoid writing a serializer you could write it as an `std::pair` instead of a struct.

src/engine/client/cl_cgame.cpp
illwieckz: Done.

illwieckz:
> This sounds like it could use the shmem command queue too in the future so maybe everything could fit in one area?

There are two forward paths:

- Implement `trap_CM_BatchMarkFragments` in game, never queries the engine.
  It may be possible that the cgame has all knowledge required to do this.
- Implement this and other code called after that in engine, send a single async message.
  It looks like the return of `trap_CM_BatchMarkFragments` is used in a computation whose result is then sent again to engine with `trap_R_AddPolyToScene`.

For now this batched call looks good enough as it is less intrusive and less prone to error to achieve, while providing an already large speed up. This is a first good step that is more reachable and it brings almost all the benefit for a smaller development cost (time, effort…).

illwieckz:
I squashed the commits. This looks ready to me.

illwieckz:
As mentionned in previous comments, there are further improvements doable, including some reducing the message size, but this requires to rewrite parts of `R_MarkFragments` which is also used by the legacy `trap_CM_MarkFragments` and to rewrite parts of the legacy `trap_CM_MarkFragments` function. So those further improvement are expected to be done after `trap_CM_MarkFragments` is NUKED and I plan to NUKE `trap_CM_MarkFragments` after this is merged and submodules being synced.

DolceTriade:
Absolutely, we should merge this if @slipher doesn't have any more comments. it seems ok to me.

illwieckz:
Since last review, along the implementation of `std::pair` I did things like this to simplify the readibility of the code:

```c++
const markMsgInput_t& input = markMsgInput[ m ];
const markOriginalPoints_t& originalPoints = input.first;
const markProjection_t& projection = input.second;
```

To avoid writing meaningless things like `markMsgInput[ m ].first` but `originalPoints` and things like that, I assume the compiler knows how to optimize everything out of this and to avoid any useless pointer copy.

slipher:
> > To be lazy and avoid writing a serializer you could write it as an `std::pair` instead of a struct.
> 
> Done.

That's not what I meant. If you have a struct of POD types, it serializes automatically and replacing with a pair just makes it less readable for no reason. The point is that you can use vectors instead of arrays of length 300 or whatever. (And it the struct contains vectors it won't serialize automatically.) Probably there's only a couple triangles most of the time so these large arrays are really wasteful in terms of I/O. This is important if the goal is to reduce the number of socket writes.

illwieckz:
Then I don't know what you meant about using pairs. I don't know what POD means.

> The point is that you can use vectors instead of arrays of length 300 or whatever.

I don't want to change this data layout before NUKING the legacy code, and this PR is already large enough. We will have plenty of time to improve over this later. Using vectors where there are remaining arrays is something I want to do but doing that right now would overcomplexifies the PR. I'm well aware we can use vectors instead of arrays, it's just too early to do it.

illwieckz:
I reverted the pairs (to be squashed).

slipher:
I tweaked it to use varable-size vectors on the `slipher/batch-marks` branches in this repo and Unvanquished.

illwieckz:
I rebased and squashed the vector change.

slipher:
LGTM

PR #1116, opened 2024-05-04
WIP: Increase max network packet sizes

VReaperV:
Increase the maximum length of a network message from 32768 bytes to 262144. Also increased max packet length from 1400 bytes to 265000 bytes, because packet fragmenting is not working correctly with larger message sizes.

This allows for more entities on the map, and brings the soft limit on the amount of buildings from ~250 to ~2k.

Tested with https://users.unvanquished.net/~sweet/layouts/spacetracks/humanpve1.dat (layout for map `spacetracks`): doesn't crash anymore, the game loads fine.

I'm getting a stack overflow when trying to connect to a server for some reason though.

### DIFF
--- a/src/engine/qcommon/net_chan.cpp
+++ b/src/engine/qcommon/net_chan.cpp
@@ -59,7 +59,7 @@ to the new value before sending out any replies.
 
 */
 
-static const int MAX_PACKETLEN = 1400; // max size of a network packet
+static const int MAX_PACKETLEN = 265000; // max size of a network packet
 
 static const int FRAGMENT_SIZE = ( MAX_PACKETLEN - 100 );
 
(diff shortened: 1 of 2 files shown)


src/engine/qcommon/net_chan.cpp
```diff
@@ -59,7 +59,7 @@ to the new value before sending out any replies.
 
 */
 
-static const int MAX_PACKETLEN = 1400; // max size of a network packet
+static const int MAX_PACKETLEN = 265000; // max size of a network packet
```
slipher, line 62: You can't just make the size of the real packet arbitrarily large. Some people suggest the current value is good for [Ethernet](https://stackoverflow.com/a/3074427). And supposedly a packet this large would be [immediately rejected](https://stackoverflow.com/a/62726938) by any of our supported operating systems. Finally [Cloudflare says](https://blog.cloudflare.com/ip-fragmentation-is-broken/) that sending any packets larger than 1500 is likely to be broken.

VReaperV:
Proper fix is in #1427.

PR #1122, opened 2024-05-09
Use the correct amount of texture units

VReaperV:
Query the actual max amount of texture units with GL_MAX_COMBINED_TEXTURE_UNITS instead of a magic number, and use an std::vector with this capacity.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -148,7 +148,7 @@ void GL_SelectTexture( int unit )
 		return;
 	}
 
-	if ( unit >= 0 && unit <= 31 )
+	if ( unit >= 0 && unit < glConfig2.maxTextureUnits )
 	{
 		glActiveTexture( GL_TEXTURE0 + unit );
 
@@ -175,7 +175,7 @@ void GL_BindToTMU( int unit, image_t *image )
 
 	int texnum = image->texnum;
 
-	if ( unit < 0 || unit > 31 )
+	if ( unit < 0 || unit >= glConfig2.maxTextureUnits )
 	{
 		Sys::Drop( "GL_BindToTMU: unit %i is out of range\n", unit );
 	}
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -57,10 +57,10 @@ void GL_Bind( image_t *image )
 		texnum = tr.blackImage->texnum;
 	}
 
-	if ( glState.currenttextures[ glState.currenttmu ] != texnum )
+	if ( tr.currenttextures[ glState.currenttmu ] != texnum )
 	{
 		image->frameUsed = tr.frameCount;
-		glState.currenttextures[ glState.currenttmu ] = texnum;
+		tr.currenttextures[ glState.currenttmu ] = texnum;
 		glBindTexture( image->type, texnum );
 	}
 }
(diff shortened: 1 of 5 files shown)


illwieckz (approved):
LGTM.

src/engine/renderer/tr_init.cpp
```diff
@@ -875,7 +878,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 
 		glState.stackIndex = 0;
 
-		for ( i = 0; i < MAX_GLSTACK; i++ )
+		for ( int i = 0; i < MAX_GLSTACK; i++ )
```
illwieckz, line 881: Bikeshedding, but you can use `size_t` here.

I also wonder if `glConfig2.maxTextureUnits` can be but doing that may be too much intrusive as it is compared with `int` from legacy parts of the code. This can be done later, I'm just wondering.

VReaperV: Probably, yeah. That part was just me moving the variable declaration into the loop, but it doesn't look like those numbers will get anywhere close to overflowing: maximum texture units have been at 192 for years and MAX_GLSTACK is just defined as 5, and it doesn't look like we need it to be higher.

illwieckz: Yeah, it's not blocking, just a nitpick remark without making anything required. The PR is fine as is.

src/engine/renderer/tr_backend.cpp
```diff
@@ -148,7 +148,7 @@ void GL_SelectTexture( int unit )
 		return;
 	}
 
-	if ( unit >= 0 && unit <= 31 )
+	if ( unit >= 0 && unit <= glConfig2.maxTextureUnits )
```
slipher, line 151: Shouldn't it be `<`?

src/engine/renderer/tr_backend.cpp
```diff
@@ -175,7 +175,7 @@ void GL_BindToTMU( int unit, image_t *image )
 
 	int texnum = image->texnum;
 
-	if ( unit < 0 || unit > 31 )
+	if ( unit < 0 || unit > glConfig2.maxTextureUnits )
```
slipher, line 178: Shouldn't it be `>=`?

src/engine/renderer/tr_local.h
```diff
@@ -24,6 +24,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #ifndef TR_LOCAL_H
 #define TR_LOCAL_H
 
+#include <vector>
```
slipher, line 27: This is included (indirectly) from `Common.h`, the precompiled header that should be included from all files, so you don't have to add it here.

src/engine/renderer/tr_backend.cpp
VReaperV: Fixed.

src/engine/renderer/tr_backend.cpp
VReaperV: Also fixed.

src/engine/renderer/tr_local.h
VReaperV: Ah, I see, thanks, fixed this too.

slipher:
LGTM

VReaperV:
Squashed the commit with minor fixes from reviews, no code changes here.

PR #1125, opened 2024-05-10
renderer: micro-optimize CPU culling

illwieckz:
This code is a hot spot, we better avoid computing useless things if we can return early and do bitwise operations instead of relying on the branch prediction being right.

Commits are meant to be squashed, they are just small steps I did one by one for testing I was not breaking anything.


illwieckz:
Before:

20240510-123313-000 unvanquished-orbit

After:

20240510-124016-000 unvanquished-orbit

The engine now spend `38%` of the time in `SCR_UpdateScreen` instead of `46%`.

`BoxOnPlaneSide` is also used in other functions (outside of the screenshot).

The second screenshot is shifted by one line down because now the time spent in that part of the code is smaller than the time spend in the graphics driver so the graphics driver thread was listed first in the second screenshot.

illwieckz:
OK, something is becoming crazy:

20240510-140308-000 unvanquished-orbit

The engine now spends `20%` of the time in `SCR_UpdateScreen`, and now `RE_RenderScene` is `4%`.

illwieckz:
I tested this branch rebased over `for-0.55.0/sync` branch and I also see a speed bump.

Before:

20240512-191032-000 unvanquished-orbit

After:

20240512-191513-000 unvanquished-orbit

So, I'll start to clean-up things and submit the patches for merging in a near future.

src/engine/renderer/tr_world.cpp
```diff
@@ -1061,10 +1036,8 @@ void R_AddPrecachedWorldInteractions( trRefLight_t *light )
 				{
 					continue;
 				}
-				else
-				{
```
slipher, line 1065: A lot of stuff here, for example this, or changing `if (p == nullptr)` to `if (!p)`,  is just cosmetic changes which don't change the meaning of the code.

PR #1129, opened 2024-05-11
pbr: Maybe fix weird dark artifacts in PBR

DolceTriade:
Looking at
https://github.com/Nadrin/PBR/blob/master/data/shaders/glsl/pbr_fs.glsl, this repo does PBR slightly differently. In general, the calculations are similar until they come to adding it to the color. This change ported from this renderer appears to help remove black artifacts.

### Before
unvanquished_2024-05-10_231751_000

### After
unvanquished_2024-05-10_231807_000


illwieckz (approved):
The overall look still looks wrong but this fixes some of the wrongness and doesn't seem to add new wrongness.

LGTM.

DolceTriade:
I'm going to merge this for 0.55. If we find a better solution, we should do that

PR #1130, opened 2024-05-11
tr_shader: properly flag shader based lightmaps with IF_NOPICMIP and IF_LIGHTMAP

illwieckz:
Properly flag shader based lightmaps with IF_NOPICMIP and IF_LIGHTMAP.

This is a cherry pick from that old branch that have many commits unmerged yet:

- https://github.com/DaemonEngine/Daemon/pull/394

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -2147,6 +2147,8 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
 			else
 			{
 				loadMap = true;
+				imageBits |= IF_LIGHTMAP;
+				imageBits |= IF_NOPICMIP;
 			}
 		}
 		// clampmap <name>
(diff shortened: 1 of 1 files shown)


DolceTriade:
Could you explain the implications of this for people less familiar with the renderer code.

illwieckz:
Setting `IF_NOPICMIP` to an image will prevent the image to be downscaled when image downscaling is enabled.

Lightmaps are precisely aligned with geometry so rescaling a light map makes light bleed, so it is only wanted to resize lightmaps if it's too large for the system, not because the user asked to downscale the images in general.

The external lightmaps used in light styles (blending animated lights like blinking neons) where missing that flag, while the default internal lightmaps were already using it.

The `IF_LIGHTMAP` flag is, as the name says, telling the image is a lightmap, which makes easier to determine what to do with it. For example the legacy overbright code relies on this flag to multiply the light or not.

That said I have to check that part. That patch had ben sleeping for 4 years I also have to double-check this. 🤪️

DolceTriade:
Got it, thanks

PR #1131, opened 2024-05-11
WIP: tr_image: downscale image against hardware limits using GL_PROXY feature

illwieckz:
Cherry picked from that old branch:

- https://github.com/DaemonEngine/Daemon/pull/394

I haven't touched this code in 4 years so I'm flagging at draft because I need to refresh my mind on this to begin with.

### DIFF
--- a/src/engine/renderer/InternalImage.cpp
+++ b/src/engine/renderer/InternalImage.cpp
@@ -36,6 +36,163 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // Comments may include short quotes from other authors.
 
+int R_GetBlockSize( const image_t *image )
+{
+	switch ( image->internalFormat )
+	{
+		// S3TC formats
+
+		case GL_COMPRESSED_RGB_S3TC_DXT1_EXT:
+		case GL_COMPRESSED_SRGB_S3TC_DXT1_EXT:
+			return 8;
+
+		case GL_COMPRESSED_RGBA_S3TC_DXT3_EXT:
+		case GL_COMPRESSED_SRGB_ALPHA_S3TC_DXT3_EXT:
+			return 16;
+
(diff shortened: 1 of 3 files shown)


DolceTriade:
Once you remember what this is for, it would be nice to explain it in the commit message :smile:

illwieckz:
```c++
	/* From https://www.khronos.org/opengl/wiki/GLAPI/glTexImage3D

	> If target is GL_PROXY_TEXTURE_3D, no data is read from data,
	> but all of the texture image state is recalculated, checked
	> for consistency, and checked against the implementation's
	> capabilities. If the implementation cannot handle a texture
	> of the requested texture size, it sets all of the image state
	> to 0, but does not generate an error (see glGetError).
	> To query for an entire mipmap array, use an image array level
	> greater than or equal to 1.

	From https://www.khronos.org/opengl/wiki/GLAPI/glTexImage3D
	> To then query this state, call glGetTexLevelParameter. */
```

The idea is to attempt a fake upload of the image with the size expected to be uploaded, and get the exact answer from the driver itself if an image with such format and size  is accepted or not, before doing the image upload for real.

If the driver answers an image of this size for this format is not accepted, reduce the size and retry until the driver accepts. Once an acceptable size is found, upload the  image with that smaller size.

slipher:
My recollection is that it turned out that vendors didn't bother to implement GL_PROXY correctly, so using it was a bad idea.

illwieckz:
> My recollection is that it turned out that vendors didn't bother to implement GL_PROXY correctly, so using it was a bad idea.

The doubt about vendor implementation was about `GL_MAX_CUBE_MAP_TEXTURE_SIZE_ARB` (`glConfig2.maxCubeMapTextureSize`) and it is what motivated this alternate implementation.

That patch was considered by me as ready to merge at the time but this just never happened and it just fall behind other priorities like doing some releases. I never forgot about it and it always had been on my TODO list. Some patches from that branch were already merged at the time, merging the next ones on the line always had ben planned.

PR #1132, opened 2024-05-11
CI: skip on files not contributing to the binary

illwieckz:
Skip on files not contributing to the binary.


.appveyor.yml
```diff
@@ -1,5 +1,17 @@
 # Documentation: https://wiki.unvanquished.net/wiki/Continuous_integration
 
+only_commits:
+  files:
+    - .appveyor.yml
+    - appveyor/
```
slipher, line 6: There is no `appveyor/` in the Daemon repo.

.appveyor.yml
```diff
@@ -1,5 +1,17 @@
...
+    # The rest of this list should stay in sync with azure-pipelines.yml
+    - src/
+    - libs/
+    - cmake/
+    - CMakeLists.txt
+    - '*.cmake'
+    - .gitmodules
```
slipher, line 13: Some stuff in `tools/` is used by the build.

illwieckz: I have not added it on purpose because I planned to do that:

- https://github.com/DaemonEngine/Daemon/pull/1135

The remaining things in `tools/` is not built by the CI (it's arm64 specific).

.appveyor.yml
illwieckz: Oops, removed.

slipher:
LGTM

PR #1133, opened 2024-05-11
VM-side cache for cvars registered there

slipher:
For improving the performance of Cvar::GetValue.

Also I uncovered another funny performance bug where setting a VM cvar was O(N) in the number of cvars.

### DIFF
--- a/src/common/Cvar.cpp
+++ b/src/common/Cvar.cpp
@@ -41,7 +41,7 @@ namespace Cvar {
     }
 
     bool CvarProxy::Register(std::string description) {
-        return ::Cvar::Register(this, name, std::move(description), flags, std::move(defaultValue));
+        return ::Cvar::Register(this, name, std::move(description), flags, defaultValue);
     }
 
     bool ParseCvarValue(Str::StringRef value, bool& result) {
--- a/src/shared/CommonProxies.cpp
+++ b/src/shared/CommonProxies.cpp
@@ -272,11 +272,16 @@ namespace Cmd {
 
 namespace Cvar{
 
-    // Commands can be statically initialized so we must store the registration parameters
-    // so that we can register them after main
-
     struct CvarRecord {
         CvarProxy* cvar;
+
+        // Cache values of registered cvars to reduce trap calls.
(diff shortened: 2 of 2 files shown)


illwieckz:
Following screenshots were all done on a 10 seconds recording.

Here is before (left)/after (right) on Plat23 default scene with spectator HUD:

20240512-065829-001 unvanquished-orbit

We notice a clear difference. It's a big win for the HUD!

For unknown reasons the fourth highlighted `Cvar::GetValue` on the bottom of the list still did an IPC call.

Now when I open the main menu (that processes most of the preferences windows even if not displayed), I still get a lot of IPC calls, I have no idea why:

20240512-071843-001 unvanquished-orbit

For some unknown reasons, it even seems to be a bit slower, I reproduced the before/after comparison twice and got the same difference:

20240512-072507-001 unvanquished-orbit

illwieckz:
Well, for the main menu I guess it's because it's only caching Cvars registered on cgame side, and main menu preferences windows are full of engine cvars, so they go through IPC every time. Main menu being slow isn't the biggest problem.

illwieckz:
The remaining cgame one that is still not cached is `ui_winner`, it is set with `trap_Cvar_Set`:

```
src/cgame/cg_servercmds.cpp
362:		trap_Cvar_Set( "ui_winner", str );

src/cgame/cg_main.cpp
1365:	trap_Cvar_Set( "ui_winner", "" ); // Clear the previous round's winner.

pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml
155:	<inlinecvar cvar="ui_winner" class="winner" />
```

illwieckz:
This problem may not be solvable on the cache side, because the cgame cannot know if the cvar is not managed on engine side when setting it.

I guess the better is just to create the cvar in cgame C++ code.

illwieckz:
There is no single IPC call done anymore by the HUD once this is merged on cgame side:

- https://github.com/Unvanquished/Unvanquished/pull/2997

Here an update before (left)/after (right):

20240512-065829-002 unvanquished-orbit

illwieckz (approved):
LGTM.

PR #1137, opened 2024-05-12
WIP: Material system: Implement GPU frustum culling

VReaperV:
Some steps to implement culling in compute shaders for the material system. Currently there's some weird bug with culling since it's culling away the wrong surfaces. Also all the surfaces look glitched, likely due to concurrent memory reads/writes, although I put all the correct memory barriers in place, so hopefully just triple-buffering them or something will fix it.

Also might crash because I limited it to 4 stages (including depth) per surface right now, without any checks.


VReaperV:
I've fixed the culling now. Just need to deal with the memory issue.

VReaperV:
Also made it work with r_lockPVS so it can be tested now by just doing /r_lockPVS 1 and looking around.

illwieckz:
Ho do I enable GPU culling? I tried the branch but a look in Orbit shows me that it still uses CPU culling on my end.

VReaperV:
> How do I enable GPU culling? I tried the branch but a look in Orbit shows me that it still uses CPU culling on my end.
> 
> I enabled `r_useMaterialSystem` but this seems to not be enough.

Do you have /r_arb_bindless_textures 1? I've disabled it by default because of the perf problems on AMD that we observed earlier.
Also, entities still use CPU culling, but you should no longer see R_RecursiveWorldNode() being called if this is enabled.

VReaperV:
Also, it might crash or something if there are surfaces with more than 3 stages; this can be patched by changing MAX_SURFACE_COMMANDS (both in `cull_cp.glsl` and `Material.h`) to a higher number, but a proper fix will require different shader permutations I think.

illwieckz:
```
]/r_arb_bindless_textures 1
Unknown command r_arb_bindless_textures 
```

If I do this it sets the cvar:

```
/set r_arb_bindless_textures 1
```

But I see no difference after both this and `r_useMaterialSystem` are enabled and I did `vid_restart`, the code still does a lot of `R_RecursiveWorldNode` calls.

VReaperV:
Sorry, got the wrong cvar name. `r_arb_bindless_texture` is the correct one

illwieckz:
Thanks.

I now get some flickering and lines like this (atcshd):

```
Warn: 0 15 
Warn: 15 15 
Warn: 30 2 
Warn: 32 2 
Warn: 34 1 
Warn: 35 1 
Warn: 36 1 
```

And no `R_RecursiveWorldNode` !

So I guess I'm getting this branch running ! 👍️

VReaperV:
Fixed most of the flickering issues now and made it double buffered. Some of the flickering issues still remain, but it's probbaly an off-by-1 error somewhere.

VReaperV:
Made this properly double-buffered now: R_RenderView() will queue views for surface culling, then in RB_RenderPostProcess() it will dipsatch computes for the queued views. I haven't tested that it works with multiple views yet though, but I need to cleanup and fix some things first before that.

VReaperV:
I'm not getting any flickering now. Also fixed a crash and hopefully the incorrect memory accesses.

illwieckz:
Excluding the missing fog, I see no visual glitch anymore in ATCSHD, and I don't experience GPU pagefaults anymore.

This is coming to be in good shape! 😃️

Framerate is around 370 fps on medium preset @4K with a Radeon PRO W7600 On Mesa 23.2.1 radeonsi. I usually get around 570 fps on master.

illwieckz:
I get 400 fps on Mesa 24.0.7.

VReaperV:
Hmm, I wonder if it's because of bindless textures still... Well, it's also still culling less surfaces right now because the far plane is ignored (we also have it as (0, 0, 0)), so if you're looking e. g. towards one of the sides on atcs it will render all of the surfaces behind walls etc.

illwieckz:
Yes I test with the default spectator scene, so in ATCSHD it means the whole outdoor and the whole alien base is in line of sight.

VReaperV:
Yea... I'll make a separate pr later for occlusion culling, that should fix that part :)

VReaperV:
This works slightly faster for me now after removing some unnecessary branching.

VReaperV:
Here's a more concise graph of how this culling works:
Culling

VReaperV:
I've also made this work with multiple different views (i. e. portals) and moved defines to GLHeaders.

Surface commands will now use the minimal array size for the maximum amount of stages used on any compatible surface on the map (padded out to be a multiple of 4 for alignment). This required making the `cull` shader load after the map is loaded and there doesn't seem to be a better solution than that.

VReaperV:
Frustum culling can now be toggled with `r_gpuFrustumCulling`.

VReaperV:
Rebased on material branch now since I'm just waiting for someone to approve the changes in #1105 so I can merge them.

VReaperV:
I've added a comment about `MaterialPack` here.

VReaperV:
Added some comments.

VReaperV:
This pr also fixes the issue seen in #1105 where some textures were selected incorrectly.

illwieckz (approved):
As said here:

https://github.com/DaemonEngine/Daemon/pull/1180#pullrequestreview-2208577967

I'm not sure to understand everything, but the overall looks good to me and I confirm it fixes many bugs.

PR #1141, opened 2024-05-13
sdl_input: rewrite `IN_SetMouseMode` to make it more readable and fix mouse escape on fullscreen toggling

illwieckz:
- sdl_input: rewrite IN_SetMouseMode to make it easier to read
  I cannot grasp the code in the current state so I cannot debug it.

- sdl_input: keep the mouse in current state when toggling fullscreen
  This fixes the mouse capture being lost on Alt-Tab.

- sdl_input: also centers the mouse pointer when entering delta mode
  My experience with NetRadiant taught me that not centering a mouse pointer when entering delta mode may help the mouse to escape the capture. It may be possible that this is useless with SDL, but it can't do harm to do it so I prefer to be extra-precocious.


slipher:
> * sdl_input: keep the mouse in current state when toggling fullscreen
>       This fixes the mouse capture being lost on Alt-Tab.

Did you mean Alt-Enter?

slipher:
Just tried it and `sdl_input: keep the mouse in current state when toggling fullscreen` has the consequence (on Windows at least) that if you vid_restart while in CustomCursor mode (I did it by using a delay command in the maim menu) and move the cursor outside the window, you get two cursors displayed (custom and system) afterward. I think this is a worse bug than the cursor potentially not being captured after Alt-Enter. The existing behavior of setting the mode to SystemCursor on a restart makes the most sense to me because (a) the game may not respond to inputs for a long time at this point; and (b) both of the other modes depend on a specific cgame state to make sense, which may not exist after the restart.

slipher:
Fullscreen toggling normally does `in_restart` but not `vid_restart`. `in_restart` is not used by anything else, so we could make it have a different behavior wrt mouse mode since there should not be a long delay and the cgame state doesn't change.

slipher:
Try #1146.

slipher:
I prefer the old version. But of course I'm biased, as the last person to rewrite that function.

PR #1143, opened 2024-05-13
renderer: fix legacy dynamic light intensity

illwieckz:
When I fixed the legacy dynamic light renderer in #1093, I noticed the dynamic light was too faint.

Since then I noticed the light was brighter with such renderer when legacy map overbright was enforced, meaning it was the overbright cancellation that was running where it should not.

This is now fixed.

Legacy dynamic light renderer with legacy overbright (reference):

unvanquished_2024-05-13_131200_000

Legacy dynamic light renderer with new overbright implementation:

unvanquished_2024-05-13_133738_000

For comparison, this is how it looks with the non-legacy tiled dynamic light renderer:

unvanquished_2024-05-13_131510_000

The light is believed to be different because the lighting algorithm is totally different to begin with.

The legacy light renderer not saturating to white is believed to be because the legacy dynamic light blends the dynamic light as a separate stage and the framebuffers are clamped to `1`, meaning `1 * n` where `n` is next stage color is always `n`. We can't fix that easily, it's a legacy renderer anyway.


slipher:
LGTM (didn't really review or test)

src/engine/renderer/glsl_source/forwardLighting_fp.glsl
```diff
@@ -1005,7 +1006,8 @@ void	main()
 	color.rgb *= attenuationXY;
 	color.rgb *= attenuationZ;
 #endif
-	color.rgb *= abs(u_LightScale);
+	color.rgb *= abs(u_LightScale)
+	color.rgb *= u_InverseLightFactor;
```
illwieckz, line 1010: There was a missing `;` there:

```glsl
color.rgb *= abs(u_LightScale)
color.rgb *= u_InverseLightFactor;
```

Now fixed.

PR #1149, opened 2024-05-15
cmake: add `USE_ARCH_INTRINSICS` and `USE_COMPILER_INTRINSICS` CMake options

illwieckz:
This introduces the `USE_ARCH_INTRINSICS` CMake option. It is enabled by default.

Disabling it is meant to disable custom asm code and usage of intrinsincs functions for the target platform in the Dæmon code base, it may also be used by games built with the Dæmon common code base.

It is not meant to disable asm or intrinsincs usage in third-party libraries.

It is not meant to prevent the compiler to use such intrinsics in its optimization passes.

It is not meant to disable the compiler flags we set to tell the compiler to try to use such intrinsics in its optimization passes. For this, one should disable `USE_CPU_RECOMMENDED_FEATURES` instead.

For obvious reason the asm code in the `BREAKPOINT()` implementation is not meant to be disabled by `USE_ARCH_INTRINSICS`, the alternate code is to do nothing anyway.

The macro syntax is: `DAEMON_ARCH_INTRINSICS_(architecture)[_extension]`

Examples:

- `DAEMON_ARCH_INTRINSICS_i686`: i686 specific code, including asm code.
- `DAEMON_ARCH_INTRINSICS_i686_sse`: i686 SSE specific code.
- `DAEMON_ARCH_INTRINSICS_i686_sse2`: i686 SSE2 specific code.

If a platform inherits feature from an parent platform, the parent platform name is used. For example on `amd64`, the definition enabling SSE code is `DAEMON_ARCH_INTRINSICS_i686_sse`, enabling SSE code on both i686 with SSE and amd64 platforms, and both `DAEMON_ARCH_INTRINSICS_amd64` and `DAEMON_ARCH_INTRINSICS_i686` are available.

This also introduces `USE_COMPILER_BUILTINS` CMake option. It is enabled by default.

Disabling it is meant to test the unknown compiler code.

### DIFF
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -250,15 +250,14 @@ See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0627r0.pdf */
 
 // Keywords specific to C++ versions
 
-/* The noexcept keyword should be used on all move constructors and move
+/* TODO: Rewrite all NOEXCEPT usages from the whole code base.
+
+The noexcept keyword should be used on all move constructors and move
 assignments so that containers move objects instead of copying them.
-The noexcept keyword was added in C++11 but the __cpp_noexcept_function_type
-definition to detect its implementation was only added in C++17. */
-#if defined(__cpp_noexcept_function_type) || __cplusplus >= 201103 
-	#define NOEXCEPT noexcept
-	#define NOEXCEPT_IF(x) noexcept(x)
-	#define NOEXCEPT_EXPR(x) noexcept(x)
-#endif
+That keyword was added in C++11, all compilers should now support it. */
+#define NOEXCEPT noexcept
+#define NOEXCEPT_IF(x) noexcept(x)
+#define NOEXCEPT_EXPR(x) noexcept(x)
 
 // Uses SD-6 Feature Test Recommendations
 #if defined(__cpp_constexpr)
@@ -270,15 +269,6 @@ definition to detect its implementation was only added in C++17. */
 #endif
 
 // Work around lack of language keywords.
-#if !defined(NOEXCEPT)
-	#define NOEXCEPT
-#endif
-#if !defined(NOEXCEPT_IF)
-	#define NOEXCEPT_IF(x)
-#endif
-#if !defined(NOEXCEPT_EXPR)
-	#define NOEXCEPT_EXPR(x) x
-#endif
 #if !defined(CONSTEXPR)
 	#define CONSTEXPR
 #endif
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -148,6 +148,7 @@ inline int CountTrailingZeroes(unsigned long long x)
 
 // Raise an exception and break in the debugger
 #if defined(DAEMON_ARCH_i686) || defined(DAEMON_ARCH_amd64)
+	// Always run this asm code even if DAEMON_USE_ARCH_INTRINSICS is not defined.
 	#define BREAKPOINT() __asm__ __volatile__("int $3\n\t")
 #elif defined(DAEMON_ARCH_nacl)
 	// TODO: find how to implement breakpoint on NaCl
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -91,15 +91,6 @@ int CountTrailingZeroes(unsigned long long x);
 	#define BREAKPOINT()
 #endif
 
-// noexcept keyword, this should be used on all move constructors and move
-// assignments so that containers move objects instead of copying them.
-#define NOEXCEPT noexcept
-#define NOEXCEPT_IF(x) noexcept(x)
-#define NOEXCEPT_EXPR(x) noexcept(x)
-
-// Work around lack of constexpr
-#define CONSTEXPR constexpr
-
 #if defined(__SANITIZE_ADDRESS__) // Detects GCC and MSVC AddressSanitizer
 #   define USING_ADDRESS_SANITIZER
 #   define USING_SANITIZER
@@ -150,10 +141,6 @@ inline int CountTrailingZeroes(unsigned long long x) { return __builtin_ctzll(x)
 #define DLLEXPORT __declspec(dllexport)
 #define DLLIMPORT __declspec(dllimport)
 #define BREAKPOINT() __debugbreak()
-#define NOEXCEPT noexcept
-#define NOEXCEPT_IF(x) noexcept(x)
-#define NOEXCEPT_EXPR(x) noexcept(x)
-#define CONSTEXPR constexpr
 #define ATTRIBUTE_NO_SANITIZE_ADDRESS
 
 inline int CountTrailingZeroes(unsigned int x) { unsigned long ans; _BitScanForward(&ans, x); return ans; }
@@ -199,6 +186,26 @@ inline int CountTrailingZeroes(unsigned long long x) { int i = 0; while (i < 64 
 #   define ALIGN_STACK_FOR_MINGW __attribute__((force_align_arg_pointer))
 #else
 #   define ALIGN_STACK_FOR_MINGW
+#endif
+
+/* The noexcept keyword should be used on all move constructors and move
+assignments so that containers move objects instead of copying them.
+The noexcept keyword was added in C++11 but the __cpp_noexcept_function_type
+definition to detect its implementation was only added in C++17. */
+#if defined(__cpp_noexcept_function_type) || __cplusplus >= 201103 
+	#define NOEXCEPT noexcept
+	#define NOEXCEPT_IF(x) noexcept(x)
+	#define NOEXCEPT_EXPR(x) noexcept(x)
+#else
+	#define NOEXCEPT
+	#define NOEXCEPT_IF(x)
+	#define NOEXCEPT_EXPR(x) x
+#endif
+
+#if defined(__cpp_constexpr)
+	#define CONSTEXPR constexpr
+#else
+	#define CONSTEXPR
 #endif
 
 // Uses SD-6 Feature Test Recommendations
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -244,30 +244,35 @@ definition to detect its implementation was only added in C++17. */
 	#define NOEXCEPT noexcept
 	#define NOEXCEPT_IF(x) noexcept(x)
 	#define NOEXCEPT_EXPR(x) noexcept(x)
-#else
-	#define NOEXCEPT
-	#define NOEXCEPT_IF(x)
-	#define NOEXCEPT_EXPR(x) x
-#endif
-
-#if defined(__cpp_constexpr)
-	#define CONSTEXPR constexpr
-#else
-	#define CONSTEXPR
 #endif
 
 // Uses SD-6 Feature Test Recommendations
-#ifdef __cpp_constexpr
-#   if __cpp_constexpr >= 201304
-#       define CONSTEXPR_FUNCTION_RELAXED constexpr
-#   else
-#       define CONSTEXPR_FUNCTION_RELAXED
-#   endif
-#   define CONSTEXPR_FUNCTION constexpr
-#else
-// Work around lack of constexpr
-#   define CONSTEXPR_FUNCTION
-#   define CONSTEXPR_FUNCTION_RELAXED
+#if defined(__cpp_constexpr)
+	#define CONSTEXPR constexpr
+	#define CONSTEXPR_FUNCTION constexpr
+	#if __cpp_constexpr >= 201304
+		#define CONSTEXPR_FUNCTION_RELAXED constexpr
+	#endif
+#endif
+
+// Work around lack of language keywords.
+#if !defined(NOEXCEPT)
+	#define NOEXCEPT
+#endif
+#if !defined(NOEXCEPT_IF)
+	#define NOEXCEPT_IF(x)
+#endif
+#if !defined(NOEXCEPT_EXPR)
+	#define NOEXCEPT_EXPR(x) x
+#endif
+#if !defined(CONSTEXPR)
+	#define CONSTEXPR
+#endif
+#if !defined(CONSTEXPR_FUNCTION)
+	#define CONSTEXPR_FUNCTION
+#endif
+#if !defined(CONSTEXPR_FUNCTION_RELAXED)
+	#define CONSTEXPR_FUNCTION_RELAXED
 #endif
 
 // Compiler specificities we can't disable.
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -86,6 +86,26 @@ inline int CountTrailingZeroes(unsigned long long x)
 }
 #endif
 
+// Sanitizer detection
+
+#if defined(__SANITIZE_ADDRESS__) // Detects GCC and MSVC AddressSanitizer
+	#define USING_ADDRESS_SANITIZER
+	#define USING_SANITIZER
+#elif defined(__SANITIZE_THREAD__) // Detects GCC ThreadSanitizer
+	#define USING_SANITIZER
+#elif defined(__has_feature)
+	#if __has_feature(address_sanitizer) // Detects Clang AddressSanitizer
+		#define USING_ADDRESS_SANITIZER
+		#define USING_SANITIZER
+	#elif __has_feature(leak_sanitizer) // Detects Clang LeakSanitizer
+		#define USING_SANITIZER
+	#elif __has_feature(memory_sanitizer) // Detects Clang MemorySanitizer
+		#define USING_SANITIZER
+	#elif __has_feature(thread_sanitizer) // Detects Clang ThreadSanitizer
+		#define USING_SANITIZER
+	#endif
+#endif
+
 // GCC and Clang
 #if defined(__GNUC__)
 
@@ -150,24 +170,6 @@ would lack too much features to compile Daemon anyway).
 
 See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0627r0.pdf */
 #define UNREACHABLE() __builtin_unreachable()
-
-#if defined(__SANITIZE_ADDRESS__) // Detects GCC and MSVC AddressSanitizer
-#   define USING_ADDRESS_SANITIZER
-#   define USING_SANITIZER
-#elif defined(__SANITIZE_THREAD__) // Detects GCC ThreadSanitizer
-#   define USING_SANITIZER
-#elif defined(__has_feature)
-#   if __has_feature(address_sanitizer) // Detects Clang AddressSanitizer
-#       define USING_ADDRESS_SANITIZER
-#       define USING_SANITIZER
-#   elif __has_feature(leak_sanitizer) // Detects Clang LeakSanitizer
-#       define USING_SANITIZER
-#   elif __has_feature(memory_sanitizer) // Detects Clang MemorySanitizer
-#       define USING_SANITIZER
-#   elif __has_feature(thread_sanitizer) // Detects Clang ThreadSanitizer
-#       define USING_SANITIZER
-#   endif
-#endif
 
 // To mark functions which cause issues with address sanitizer
 #ifdef USING_ADDRESS_SANITIZER
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -33,14 +33,61 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef COMMON_COMPILER_H_
 #define COMMON_COMPILER_H_
 
-// CountTrailingZeroes returns the number of trailing zeroes of the argument in binary.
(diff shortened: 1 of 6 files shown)


illwieckz:
Fixes Unvanquished/Unvanquished#3001:

- https://github.com/Unvanquished/Unvanquished/issues/3001

illwieckz:
This will make possible to test the non-SSE code and things like that in CI or in a local build.

src/common/Platform.h
```diff
@@ -63,15 +63,38 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+The definitions for the CPU itself is automatically is set by CMake. */
+
+#if defined(DAEMON_USE_CPU_INTRINSICS)
+	// Set CPU extension definitions.
+	#if defined(__SSE2__) || _M_IX86_FP >= 2
+		#define DAEMON_CPU_INTRINSICS_i686_sse2
+	#endif
+
+	#if defined(__SSE__) || _M_IX86_FP >= 1
+		#define DAEMON_CPU_INTRINSICS_i686_sse
```
slipher, line 88: Maybe we should call it sse1 for clarity? It might be hard to tell since we don't have any SSE2 code for now.

slipher (commented):
What about compiler-provided functions that are designed to get intrinsics to be used, but are provided for architectures without them as well? The functions used to implement CountTrailingZerores, `__builtin_ctz*` and `BitScanForward*` are like this. Also a lot of other things beginning with `__builtin_`.

illwieckz:
> What about compiler-provided functions that are designed to get intrinsics to be used, but are provided for architectures without them as well? The functions used to implement CountTrailingZerores, `__builtin_ctz*` and `BitScanForward*` are like this. Also a lot of other things beginning with `__builtin_`.

I don't really know about them. Is the question about do we want to disable them or not?

The purpose of this PR is to be able to test all our code base, I consider a compiler-provided function not within scope of our test bed, the same way it's not our responsibility to check all alternate code GLM may implement for a single function. So this option is not to configure how compiler-provided functions behave. We should trust those compiler-provided functions.

What I want is to prevent code dying or rotting because we always compile the optimized code we wrote.

slipher:
Disabling such functions would help to test all our code. I provided a platform-independent version of CountTrailingZeroes, but I don't think it's possible to actually use it on any supported compiler. So disabling those functions would help to test it. On the other hand, one might argue that things not used on any supported compiler should be nuked.

illwieckz:
About compiler-provided functions, the same need will surface if we do:

```c++
#ifdeg __GNUC__
compilerFunction();
#else
genericCode();
#endif
```

But maybe we would prefer another CMake option and compiler define.

illwieckz:
So reading the source, yes it looks like we could benefit from some `USE_COMPILER_BUILTIN` option to disable.

On the other hand, exotic compilers are more rare than exotic platforms, almost everything can be built using a GCC or Clang-based compiler today. Clang is becoming the chromium of compilers…

illwieckz:
We can do that:

```diff
diff --git a/src/common/Compiler.h b/src/common/Compiler.h
index c3649c5a7..592ee8076 100644
--- a/src/common/Compiler.h
+++ b/src/common/Compiler.h
@@ -40,7 +40,7 @@ int CountTrailingZeroes(unsigned long x);
 int CountTrailingZeroes(unsigned long long x);
 
 // GCC and Clang
-#ifdef __GNUC__
+#if defined(DAEMON_USE_COMPILER_BUILTINS) && defined( __GNUC__ )
 
 // Emit a nice warning when a function is used
 #define DEPRECATED __attribute__((__deprecated__))
@@ -133,7 +133,7 @@ inline int CountTrailingZeroes(unsigned long x) { return __builtin_ctzl(x); }
 inline int CountTrailingZeroes(unsigned long long x) { return __builtin_ctzll(x); }
 
 // Microsoft Visual C++
-#elif defined( _MSC_VER )
+#elif defined(DAEMON_USE_COMPILER_BUILTINS) && defined( _MSC_VER )
 
 // Disable some warnings
 #pragma warning(disable : 4100) // unreferenced formal parameter
```

illwieckz:
I can do a commit for that.

slipher:
> So this option is not to configure how compiler-provided functions behave. We should trust those compiler-provided functions.

The two cases seem pretty similar to me. Yes, we trust the intrinsic/compiler-provided function. We want to be able to test both the version of our code with the intrinsic/compiler-provided function and the version without the intrinsic/compiler-provided function. 

> About compiler-provided functions, the same need will surface if we do something like:
> 
> ```c++
> #ifdef __GNUC__
> compilerFunction();
> #else
> genericCode();
> #endif
> ```
> 
> But maybe we would prefer another CMake option and compiler define.

The problem with a "generic compiler" option would be that the program is likely to not actually build or work, since some of the compiler-specific things are required. For example, you may get an error about a function not returning a value when NORETURN or UNREACHABLE is not implemented. At runtime things may break if ALIGNED is not implemented.

A "generic compiler" option may end up being a near superset of disabling intrinsics. SSE-family intrinsics are a rare case where there is some sort of standard syntax across compilers. I think most other things will be platform-specific, e.g. inline assembly.

There is not really a bright line between "true" intrinsics, and functions designed to be implemented with intrinsics. An optimizing compiler may remove an intrinsic function and implement it in a different way. For example your compiler replaced the `_mm_store_ps` followed by scalar addition with an all-SSE implementation. Maybe some compilers would implement `_mm_*` functions even on a non-SSE platform. I read of a library that implements them using ARM Neon instructions. An intrinsic is just a compiler-provided function that is not implemented on some platforms.

For these reasons I favor including `__builtin_` functions, etc., under the umbrella of USE_CPU_INTRINSICS.

illwieckz:
> For these reasons I favor including `__builtin_ functions`, etc., under the umbrella of USE_CPU_INTRINSICS.

OK that makes sense.

> I read of a library that implements them using ARM Neon instructions.

Yes there is at least [sse2neon](https://github.com/DLTcollab/sse2neon) and [SIMD Everywhere](https://github.com/simd-everywhere/simde) (that does more than ARM, it can even be used for WebAssembly).

I used sse2neon [with an old version of Embree used by LuxMark3](https://github.com/illwieckz/flathub/blob/luxmark3/patches/luxmark3/luxmark3-embree-build-aarch64-with-sse2neon.patch) and I used SIMDe to [with LuxMark3](https://github.com/illwieckz/flathub/blob/luxmark3/patches/luxmark3/luxmark3-luxmark-use-simde-on-non-i686.patch) to [port LuxMark3 on ARM](https://twitter.com/illwieckz/status/1583873855541870592).

I may evaluate SIMDe for Dæmon on ARM64 at some point (we already require NEON).

illwieckz:
Interesting, disabling arch builtins break that unit test:

```
/home/vsts/work/1/s/src/common/cm/unittest.cpp:168: Failure
Expected equality of these values:
  tr.fraction
    Which is: 0.19213901
  0.1921389
```

Log [here](https://dev.azure.com/UnvanquishedDevelopment/Daemon/_build/results?buildId=8532&view=logs&j=e4954a55-bc0b-5fbe-d8f6-05775b1a9559&t=6fedcdd4-7b5b-58c7-109c-77b7f76a6cd5).

illwieckz:
So, there is now:

- `USE_ARCH_INTRINSICS` which is meant to enable asm code, SSE and things like `CountTrailingZeroes`.
  This make possible to test alternate code for when the architecture is unknown.
- `USE_COMPILER_BUILTINS` which is meant to disable specific keywords using compiler-specific builtins like `UNREACHABLE` and things like that.
  This make possible to test the alternate code for when the compiler is unknown.

I rewrote a bit the part of `Compiler.h`, for example defining C++ features according to C++ version definitions instead of compiler definitions.

In the end, while not being the purpose of this PR, the compilation of the server and the TTY client with C++11 is working again. The graphical client makes use of C++14 features and there is plan to change that (it's fine). But all of this means the code is better written and is more robust, being more compatible with older or unknown compilers.

illwieckz:
I don't know why but the appveyor build fails on some curious errors…

illwieckz:
Well, there was a more meaningful error after that:

```
common\FileSystem.cpp(309,22):
  error C2694: 'const char *FS::minizip_category_impl::name(void) const':
  overriding virtual function has less restrictive exception specification than base class virtual member function 'const char *std:
```

MSVC doesn't properly report `__cplusplus` by default to keep compatibility with broken legacy code:

- https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

illwieckz:
I've fixed the problem related to `__cplusplus`, but the padding error still remains.

illwieckz:
What seems to break it is `cmake: add USE_ARCH_INTRINSICS CMake option` commit. That makes no sense.

illwieckz:
I may have found the error… MSVC doesn't set `__SSE*__` neither `_M_IX86_FP` on amd64, meaning that without testing `_M_AMD64` or `_M_X64` we didn't include `xmmintrin.h` and `emmintrin.h`.

Maybe MSVC changes some of its internal behavior when those headers are not included…

slipher:
> engine/renderer/tr_types.h(142,15): warning C4324:
>   'refBone_t': structure was padded due to alignment specifier

I also saw this warning when I tried implementing `USE_RECOMMENDED_CPU_FEATURES` for MSVC so I could test with SSE2 off. It seems that you get this warning when:
- You have a struct A that contains a member of type B
- B has a user-requested alignment (via `__declspec(align(n))` or `alignas`) which is greater than its natural alignment
- The alignment of B is greater than the natural alignment of any member of A
- A does not have a user-requested alignment at least as large as B's

So this warning happens when our SSE code is disabled because the structure's natural alignment is no longer 16, since it doesn't contain an `__m128`. I implemented a fix (see the branch `slipher/align-stuff`), but it is a pretty silly warning so I'd probably just disable it if it annoyed us again.

slipher:
> I implemented a fix (see the branch `slipher/align-stuff`), but it is a pretty silly warning so I'd probably just disable it if it annoyed us again.

Oops, apparently the fix doesn't work. Still getting the warning.

illwieckz:
I noticed we get the warning when:

- `/arch:SSE2` is set
- we don't use any function from `xmmintrin.h` 
- we don't include `xmmintrin.h`

I have not tried when `/arch:SSE2` is not set.

slipher:
New warning theory:
- You have a struct A
- A's alignment is greater than its natural aligment (considered by recursively ignoring alignment specifiers), due to a user-requested alignment (via __declspec(align(n)) or alignas) on A or one of its recursive members
- A has holes in its layout

This is definitely bogus - I'm not interested in writing `char pad[8]` or whatever everywhere. Will disable it in my branch.

illwieckz:
So, even not doing `/arch:SSE2` on x86 doesn't fix the warning, but the warning disappears as soon as `xmmintrin.h` is included…

illwieckz:
Can it be possible that `xmmintrin.h` just do `#pragma warning(disable : 4324)` ?

slipher:
> Can it be possible that `xmmintrin.h` just do `#pragma warning(disable : 4324)` ?

It does not.

It's not about whether SSE headers are included; it's whether we use `__m128` in our structs. Removing `/arch:SSE2` has no effect as that is the default. You have to explicitly request a lower architecture; see #1152.

illwieckz:
> it's whether we use `__m128` in our structs.

Can it be possible that `__m128` would be redefined in a different way if `xmmintrin.h` is included?

I don't get why including this header file makes the warning go away…

src/common/Compiler.h
```diff
@@ -223,6 +210,26 @@ inline int CountTrailingZeroes(unsigned long long x) { int i = 0; while (i < 64
 #   define ALIGN_STACK_FOR_MINGW
 #endif
 
+/* The noexcept keyword should be used on all move constructors and move
+assignments so that containers move objects instead of copying them.
+The noexcept keyword was added in C++11 but the __cpp_noexcept_function_type
+definition to detect its implementation was only added in C++17. */
+#if defined(__cpp_noexcept_function_type) || __cplusplus >= 201103
```
slipher, line 252: We should nuke all this NOEXCEPT and CONSTEXPR stuff. It was used because some compilers took many years to finish implementing C++11, but all the supported compilers should be fine now.

illwieckz:
I removed the HACK including `xmmintrin.h` in non-SSE MSVC contexts now that the build doesn't fail thanks to:

- https://github.com/DaemonEngine/Daemon/pull/1152

src/common/Platform.h
illwieckz: I believe SSE is OK. I don't think it's a good idea to suppose tha “something” means “something and later”, in the same way talking about AVX doesn't assume AVX-512 is available.

src/common/Compiler.h
illwieckz: We better do this in a later PR, especially if we want to rewrite all the occurrences making use of them.

slipher: Sure, that can be done later, and for now you can remove the commit putting lipstick on the stuff to be nuked :stuck_out_tongue:

src/common/Platform.h
```diff
@@ -63,15 +63,73 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+>   that target x64 processors or ARM64EC. Otherwise, undefined.
+> - _M_ARM64EC Defined as 1 for compilations that target ARM64EC.
+>   Otherwise, undefined.
+> -- https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170
+
+It is unclear if xmmintrin.h is available on ARM64EC. */
+#if defined(_MSC_VER)
+	#if defined(_M_IX86_FP)
+		#if _M_IX86_FP >= 2
+			#define MSVC_SSE
```
slipher, line 85: It seems better to define `DAEMON_USE_ARCH_INTRINSICS_i686_sse` etc. directly here, rather than polluting the namepace with another layer of `MSVC_SSE` macros.

src/common/Compiler.h
```diff
@@ -95,7 +95,7 @@ int CountTrailingZeroes(unsigned long long x);
 #endif
 
 // GCC and Clang
-#if defined(__GNUC__)
+#if defined(DAEMON_USE_COMPILER_INTRINSICS) && defined(__GNUC__)
```
slipher, line 98: I tried compiling with GCC with an empty definition of NORETURN and as predicted, compilation fails due to errors about functions not returning a value.

This build option is hardly useful if the code can't compile, and IMO disabling the `__attribute__` type of stuff has scant testing value anyway. The scope of `DAEMON_USE_COMPILER_INTRINSICS` would be better restricted to just the BitScanForward type of stuff (that is, if you insist on maintaining a distinction between two types of intrinsics; merging the two under one option would also be good).

src/common/Compiler.h
illwieckz: That code is not only putting lipstick on it, it is moving it to another place in a way it is not defined per compiler. All the compilers run this unique code, which was not true before, it's now deduplicated. I needed this to test other parts of the code with various compiler options.

Basically instead of doing:

```c++
#if GCC
// GCC compiler thing
// C++ language thing
#elif MSVC
// MSVC compiler thing
// C++ language thing
#else
// other compiler thing
// old C++ language thing
#endif
```

We now do:

```c++
#if GCC
// GCC compiler thing
#elif MSVC
// MSVC compiler thing
#else
// other compiler thing
#endif

#if CPLUSPLUSVERSION
// C++ language thing
#else
// old C++ language thing
#endif
```

src/common/Platform.h
illwieckz: Done.

I could also set `__SSE__` and `__SSE2__` instead of custom `MSVC_SSE` and `MSVC_SSE2` so the rest of the code would be generic.

illwieckz: > I could also set `__SSE__` and `__SSE2__` instead of custom `MSVC_SSE` and `MSVC_SSE2` so the rest of the code would be generic.

In the same way we define `__x86_64__` in all cases, right above.

illwieckz: I added a commit with that other idea, tell me what you think about it.

This _may_ not be wanted if it interfers badly with third-party libraries.

src/common/Compiler.h
illwieckz: Weird, on my end it works with both GCC and Clang.

I want a way to test the “unknown compiler” case.

illwieckz:
So there is now:

- `USE_ARCH_INTRINSICS`, for enabling custom code making use of SSE and asm and things like that.
- `USE_COMPILER_INTRINSICS`, for enabling custom `CountTrailingZeroes` and things like that.
- `USE_COMPILER_CUSTOMIZATION`, for enabling custom compiler attributes and operators.

They are all enabled by default and marked as advanced options (not displayed by default to the user in `ccmake`).

One can disable them either to test non-SSE code, to test code not making use of compiler builtins like `__builtin_ctz`, or the unknown compiler preprocessor code path.

src/common/Compiler.h
slipher: > That code is not only putting lipstick on it, it is moving it to another place in a way it is not defined per compiler.

There's nothing there to cause problems with this. All the definitions of NOEXCEPT* are identical and can simply be deduplicated. All the CONSTEXPR* stuff is already outside of compiler-specific blocks.

src/common/Platform.h
slipher: It's not good to define those macros which are supposed to be defined by the compiler. Technically it's undefined behavior to define anything beginning with double underscore or underscore, capital letter.

illwieckz: OK,  I went back to the `ifdef` tree.

illwieckz:
I reworded a bit the way fallback definitions are done: when there is a known implementation the define is set, and at the end the fallback is set if nothing defined it before. It avoids copy-pasting the fallback definitions for every condition with the risk to miss some.

I also reworded this:

```c++
int CountTrailingZeroes(unsigned int x);
int CountTrailingZeroes(unsigned long x);
int CountTrailingZeroes(unsigned long long x);
inline int CountTrailingZeroes(unsigned int x) { return __builtin_ctz(x); }
inline int CountTrailingZeroes(unsigned long x) { return __builtin_ctzl(x); }
inline int CountTrailingZeroes(unsigned long long x) { return __builtin_ctzll(x); }
```

Into this:

```c++
inline int CountTrailingZeroes(unsigned int x) { return __builtin_ctz(x); }
inline int CountTrailingZeroes(unsigned long x) { return __builtin_ctzl(x); }
inline int CountTrailingZeroes(unsigned long long x) { return __builtin_ctzll(x); }
```

I also removed the useless warning on missing `BREAKPOINT` implementation, we never did anything with it and we silence it for NaCl since forever.

I believe I have nothing to add within the scope of that PR.

Alongside the addition of CMake options, the way things are done are now much more meaningful.

There can be other things done later like evaluating if things are still needed, but I think it's better this way as a first step, and future clean-up commits may even be simpler.

I noticed the server and tty still build with c++11 (and run properly after that) so for now I see no reason to drop the related fallbacks.

This looks ready to me.

src/common/Compiler.h
illwieckz: Deduplicated.

src/common/Compiler.h
illwieckz: The code now always define `NORETURN` and other things if any previous condition failed to define them.

src/common/Compiler.h
```diff
@@ -33,14 +33,63 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-int CountTrailingZeroes(unsigned int x);
-int CountTrailingZeroes(unsigned long x);
-int CountTrailingZeroes(unsigned long long x);
+// Code making use of compiler intrinsics.
+
+/* CountTrailingZeroes returns the number of
+trailing zeroes of the argument in binary.
+The result is unspecified if the input is 0. */
+#if defined(DAEMON_USE_COMPILER_INTRINSICS) && defined(__GNUC__)
+	inline int CountTrailingZeroes(unsigned int x)
```
slipher, line 42: The formatting seems messed up on all the CountTrailingZeroes definitions

illwieckz: Well, I mostly just turned:

```c++
inline int CountTrailingZeroes(unsigned int x) { unsigned long ans; _BitScanForward(&ans, x); return ans; }
```

into:

```c++
inline int CountTrailingZeroes(unsigned int x)
	{ unsigned long ans; _BitScanForward(&ans, x); return ans; }
```

It's just splitting a long line. It's a matter of taste, but otherwise we may just reformat them the non-golfy way.

illwieckz: Or is the problem the extra indentation within the `ifdef`?

slipher: Well if we define a function on multiple lines, the opening brace is normally at the same alignment level as the beginning of the function declaration (`inline int`...). Also being inside a preprocessor block normally isn't a reason to indent non-preprocessor code.

illwieckz: Ok

PR #1154, opened 2024-05-18
Implement compute shaders support, make fragment shader non-mandatory

VReaperV:
Requires #1124.

Cherry-picked from #1137 to split a large pr up a bit.

Adds some code to allow using compute shaders. Shaders can now have either of vertex, fragment, or compute shaders enabled/disabled.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -433,6 +433,34 @@ static std::string GenVersionDeclaration() {
 	return str;
 }
 
+static std::string GenComputeVersionDeclaration() {
+	// Compute version declaration, this has to be separate from other shader stages,
+	// because some things are unique to vertex/fragment or compute shaders
+	std::string str = Str::Format( "#version %d %s\n",
+		glConfig2.shadingLanguageVersion,
+		glConfig2.shadingLanguageVersion >= 150 ? ( glConfig2.glCoreProfile ? "core" : "compatibility" ) : "" );
+
+	// add supported GLSL extensions
+	struct extension_t {
+		bool available;
+		int minGlslVersion;
+		std::string name;
+	};
+
+	const std::vector<extension_t> extensions = {
+		{ glConfig2.computeShaderAvailable, 430, "ARB_compute_shader" },
+		{ glConfig2.gpuShader4Available, 130, "EXT_gpu_shader4" },
+		{ glConfig2.gpuShader5Available, 400, "ARB_gpu_shader5" },
+		{ glConfig2.uniformBufferObjectAvailable, 140, "ARB_uniform_buffer_object" },
+	};
+
+	for ( const auto& extension : extensions ) {
+		addExtension( str, extension.available, extension.minGlslVersion, extension.name );
+	}
+
+	return str;
+}
+
 static std::string GenCompatHeader() {
 	std::string str;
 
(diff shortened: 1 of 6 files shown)


src/engine/renderer/gl_shader.cpp
```diff
@@ -770,9 +809,46 @@ std::string     GLShaderManager::BuildGPUShaderText( Str::StringRef mainShaderNa
...
+	std::string line;
+
+	while ( std::getline( shaderTextStream, line, '\n' ) ) {
+		const std::string::size_type position = line.find( "#insert" );
+		if ( position == std::string::npos ) {
+			shaderMain += line + "\n";
+			continue;
+		}
+
+		const std::string::iterator beginIt = std::find_if( line.begin(), line.end(),
```
slipher, line 826: Could use `find_first_not_of(" \t)`, then you get an unsigned index like `position`.

slipher (commented):
It seems like compute shaders are a totally separate workflow from the existing draw shaders, which would ideally be a separate class. Like instead of just GLShader you have one class with fragment shader, vertex shader, and vertex attribs; and another with just compute shader.

VReaperV:
> It seems like compute shaders are a totally separate workflow from the existing draw shaders, which would ideally be a separate class. Like instead of just GLShader you have one class with fragment shader, vertex shader, and vertex attribs; and another with just compute shader.

I considered that, the reason I made it part of GLShader is that they can still make use of the way uniforms are set, vertex/fragment includes, same paths for loading/compiling shaders including shader binaries, as well as GLHeaders. Maybe both classes should be derived from GLShader which would just provide the afornentioned functionality.

slipher:
A really easy way to make things a bit clearer could be to have two constructors for GLShader, one for each use case. Instead of having a constructor that allows all 3 shader types at once.

VReaperV:
So something like `GLShader( const std::string &name, const std::string &mainShaderName, uint32_t vertexAttribsRequired, GLShaderManager *manager, const bool hasVertexShader, const bool hasFragmentShader)` and `GLShader( const std::string &name, const std::string &mainShaderName, uint32_t vertexAttribsRequired, GLShaderManager *manager, const bool hasComputeShader`?

slipher:
I guess there's nothing that distinguishes the constructors very well so maybe it would be easier to go ahead and define the subclasses even if all the functionality is in the base class. Like

```
class GLShaderBase {
    // everything that's in GLShader right now
};
class GLShader : public GLShaderBase { //draw shader
   // constructor does _hasFragment = true, _haveVertex = true?
};
class ComputeShader : public GLShaderBase {
    // constructor does _hasCompute = true
};

VReaperV:
Yeah, I ran into some issues with GLShader constructors earlier too, since the wrong one was being selected with some types being converted.

src/engine/renderer/gl_shader.cpp
VReaperV: Done now.

VReaperV:
> I guess there's nothing that distinguishes the constructors very well so maybe it would be easier to go ahead and define the subclasses even if all the functionality is in the base class. Like
> 
> ```
> class GLShaderBase {
>     // everything that's in GLShader right now
> };
> class GLShader : public GLShaderBase { //draw shader
>    // constructor does _hasFragment = true, _haveVertex = true?
> };
> class ComputeShader : public GLShaderBase {
>     // constructor does _hasCompute = true
> };
> ```

This should be evaluated later because #1105 and some further PRs add more GLShader constructors.

illwieckz (approved):
LGTM.

illwieckz:
Don't forget to squash the `fix` commit. ;-)

VReaperV:
Yep, done.

PR #1156, opened 2024-05-27
Function to detect NaNs despite -ffast-math mode

slipher:


src/common/Endian.h
```diff
@@ -119,6 +119,12 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 # error The file boost/detail/endian.hpp needs to be set up for your CPU type.
 #endif
 
+// forward declaration
+namespace Util {
+	template<typename ToT, typename FromT>
+	ToT bit_cast(FromT from);
+}
+
```
DolceTriade, line 127: document

src/common/MathTest.cpp
```diff
@@ -0,0 +1,66 @@
...
+*/
+
+#include <gtest/gtest.h>
+
+#include "common/Common.h"
+
+namespace Math {
+namespace {
+
+// Uncomment this and the tests should fail with GCC or Clang in release mode :-)
```
DolceTriade, line 38: nit: is it worth it to add tests for bitcast too?

basically the example given in the docs?

```
constexpr u64v2 = 0x3fe9000000000000ull;
constexpr auto f64v2 = Util::bit_cast<double>(u64v2);
ASSERT_EQ(std::bit_cast<uint64_t]>(f64v2), u64v2); // round-trip
```

DolceTriade (approved):

src/common/MathTest.cpp
slipher: Added bit_cast tests

src/common/Endian.h
slipher: This is just a forward declaration to make DaemonArchitecture compile. The documentation is in Util.h

DolceTriade (approved):

PR #1157, opened 2024-05-27
cmake: empty string is not missing token

illwieckz:
When `CMAKE_CXX_COMPILER_ID` isn't set, this line errors out because `if(STREQUAL GNU)` is a syntax error.


illwieckz:
I noticed this when refactoring this code. For example the PNaCl compiler triggers this error. In the current code state, this code is never run with PNaCl, but once we start refactoring code in to be mutualized for usage with other NaCl compilers like Saigo, such error may surface.

slipher:
LGTM

Issue #1158, opened 2024-05-27
Why do we build engine and non-nexe game with `-fno-strict-aliasing`?


illwieckz:
In `DaemonFlags.cmake` we set `-fno-strict-aliasing` on non-NACL, non-MSVC builds:

https://github.com/DaemonEngine/Daemon/blob/7b5abf5ab45886a553bf2913a56baaeb8acd2906/cmake/DaemonFlags.cmake#L192

It means this flag is used when building released engine for all platforms, and used when building dll and exe games. But this flag is not used when building nexe game. 

This flag is there since the very beginning of the Unvanquished repository, commit `98b3a14` “Upload the codebase”:

```diff
+set( CMAKE_C_FLAGS_RELEASE "-DNDEBUG -O3 -ffast-math -fomit-frame-pointer -fno-strict-aliasing" CACHE STRING "" )
+set( CMAKE_C_FLAGS_RELWITHDEBINFO "-DNDEBUG -O3 -ffast-math -fomit-frame-pointer -fno-strict-aliasing -g" CACHE STRING "" )
```

It has probably been carried over for all those years accross all rewrites.

But when doing some refactoring on `DaemonFlags.cmake` I thought that after 12 years I may carry it over like other people did before me, or question it.

What do you think about this flag?

illwieckz:
Some type-punning issue:

So now we know:

```
engine/qcommon/msg.cpp: In function ‘void MSG_WriteDeltaFloat(msg_t*, float, float)’:
engine/qcommon/msg.cpp:626:24: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
  626 |  MSG_WriteBits( msg, * ( int * ) &newV, 32 );
      |                        ^~~~~~~~~~~~~~~
engine/qcommon/msg.cpp: In function ‘float MSG_ReadDeltaFloat(msg_t*, float)’:
engine/qcommon/msg.cpp:635:5: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
  635 |   * ( int * ) &newV = MSG_ReadBits( msg, 32 );
      |     ^~~~~~~~~~~~~~~
```

illwieckz:
We also have this:

```
engine/renderer/tr_bsp.cpp: In function ‘unsigned int VertexCoordGenerateHash(const vec_t*)’:
[...]

PR #1159, opened 2024-05-27
cmake: only use `-fno-strict-aliasing` on files requiring it

illwieckz:
This flag is there since first commit and it is likely no one remembers why it has been set to begin with.

I experience no difference when removing this.

This was only used on engine build and dll/exe game. This was never used on nexe game.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -879,6 +879,15 @@ function(AddApplication)
         target_link_libraries(test-${A_Target} GTest::gmock)
     endif()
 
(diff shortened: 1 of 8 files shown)


illwieckz:
See #1158: 

- https://github.com/DaemonEngine/Daemon/issues/1158

illwieckz:
This seems to be ready.

illwieckz:
Closed in favor of doing the opposite (always using that flag):

- https://github.com/DaemonEngine/Daemon/pull/1160

PR #1160, opened 2024-05-27
cmake: never use strict aliasing

illwieckz:
Never use strict aliasing, do the same with nexe than with exe and dll.

Alternative to:

- https://github.com/DaemonEngine/Daemon/pull/1159

Instead of trying to always do strict aliasing, never do strict aliasing.

What matters is to have consistent behaviour, #1159 will require more testing than this one.

See:

- https://github.com/DaemonEngine/Daemon/issues/1158


slipher:
LGTM

PR #1161, opened 2024-05-28
Make more info available to cgame in glconfig

slipher:


illwieckz:
I wonder if glConfig ends up in the demo file and then, having too much information makes demo files unreadable on other systems. I noticed in the past that demos recorded on GL3 systems were unplayable on GL2 systems, but I never found time to investigate it.

Also, typo:

> Move almost all of glConfig2 to glConfig***2*** ←

(Lol, it created a label I don't know how)

slipher:
> I wonder if glConfig ends up in the demo file and then, having too much information in it makes demo files unreadable on other systems. I noticed in the past that demos recorded on GL3 systems were unplayable on GL2 systems, but I never found time to investigate it.

Wat? It should only have some stuff that comes from the server: configstrings, an entity baseline, and then packets.

illwieckz:
I cannot reproduce while Mesa env variables, I'll have to wake up an old machine to try to reproduce if the bug is still there. You can ignore my comment for now.

slipher:
Dropped the config moving glconfig2 to glconfig en masse, will do that later. For now I want to get the resolution stuff merged

DolceTriade (approved):

PR #1162, opened 2024-05-28
tr_shader: do not cancel overbright on decals (impact marks)

### LINKED ISSUE #1126: Bullet marks are too faint
The bullet hole marks made by, e.g., the rifle appear much fainter than they did as of the last release. While working on the mark IPC batching commit, I thought the mark batching code was broken for a while because the marks were so faint I didn't notice them.

The marks appear as before if you set `r_forceLegacyOverbrightClamping 1`.

illwieckz:
Fix #1126:

- https://github.com/DaemonEngine/Daemon/issues/1126

While we may assume decals should not cancel overbright (the color is added to a surface that is already lit), one may investigate if other surfaces using `blendFunc GL_ZERO GL_ONE_MINUS_SRC_COLOR` while not being decals should also not cancel overbright in a more generic way.

But this is good enough for now.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5223,6 +5223,13 @@ static void FinishStages()
 		// We should cancel overBrightBits if there is no light stage.
 		stage->shaderHasNoLight = shaderHasNoLight;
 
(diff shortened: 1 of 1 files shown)


slipher:
LGTM, it works.

PR #1164, opened 2024-05-29
renderer: rework the enablement of high-precision/float texture formats

illwieckz:
Rework the enablement of high-precision/float texture formats according to available features and enabled options.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -570,6 +570,11 @@ static std::string GenEngineConstants() {
 			AddDefine( str, "r_showParallelShadowSplits", 1 );
 	}
 
+	if ( r_highPrecisionRendering.Get() )
+	{
+		AddDefine( str, "r_highPrecisionRendering", 1 );
+	}
+
 	if ( glConfig2.dynamicLight )
 	{
 		AddDefine( str, "r_dynamicLight", 1 );
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -146,7 +146,7 @@ void computeDeluxeLight( vec3 lightDir, vec3 normal, vec3 viewDir, vec3 lightCol
 #endif // !USE_PHYSICAL_MAPPING
 }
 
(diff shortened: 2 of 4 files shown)


slipher:
LGTM

illwieckz:
In fact the commit `tr_image: also check if high precision is enabled for doing RGBA32UI` was wrong. We probably need some GLSL change if float are not available and not used. Problem is… there is no check to not run this code if floats are not available. This is an issue to be adressed in another PR anyway. I removed this commit.

The rest works. I added a missing check for the `_lighttileRender` fallback and it works as expected.

illwieckz:
What we can do anyway is to use `GL_RGBA16F` instead of `GL_RGBA32F` when `r_highPrecisionRendering` is disabled. This doesnt break the rendering and can be useful to save on VRAM on low end hardware:

With `r_highPrecisionRendering` enabled:

```
]/listImages _*tile*Render
------------------------------------------------------------------
num   width heigth layers mm  type format   wrap.t   wrap.s   name
------------------------------------------------------------------
50      640    360      1 no  2D   RGBA32F  t.1clmp  s.1clmp  _depthtile1Render
51      160     90      1 no  2D   RGBA32F  t.clmp   s.clmp   _depthtile2Render
52      160     90      4 no  3D   RGBA32UI t.clmp   s.clmp   _lighttileRender
------------------------------------------------------------------
302400 total texels (not including mipmaps)
4.61 MB total image memory (estimated)
823 total images
------------------------------------------------------------------
```

With `r_highPrecisionRendering` disabled:

```
]/listImages _*tile*Render
------------------------------------------------------------------
num   width heigth layers mm  type format   wrap.t   wrap.s   name
------------------------------------------------------------------
50      640    360      1 no  2D   RGBA16F  t.1clmp  s.1clmp  _depthtile1Render
51      160     90      1 no  2D   RGBA16F  t.clmp   s.clmp   _depthtile2Render
52      160     90      4 no  3D   RGBA8    t.clmp   s.clmp   _lighttileRender
------------------------------------------------------------------
302400 total texels (not including mipmaps)
2.08 MB total image memory (estimated)
823 total images
------------------------------------------------------------------
```

The current state of the PR pleases me, I think I have nothing to add.

PR #1173, opened 2024-06-01
renderer: use `GL_DEPTH_CLAMP` only if `GL_ARB_depth_clamp` is available

illwieckz:
Use `GL_DEPTH_CLAMP` only if `GL_ARB_depth_clamp` is available.

The `GL_ARB_depth_clamp` extension was only made required in OpenGL 3.2.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1943,6 +1943,9 @@ static void GLimp_InitExtensions()
 	// made required in OpenGL 3.2
 	glConfig2.syncAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_sync, r_arb_sync->value );
 
+	// made required in OpenGL 3.2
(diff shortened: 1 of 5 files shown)


src/engine/sys/sdl_glimp.cpp
```diff
@@ -1943,6 +1943,9 @@ static void GLimp_InitExtensions()
 	// made required in OpenGL 3.2
 	glConfig2.syncAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_sync, r_arb_sync->value );
 
+	// made required in OpenGL 3.2
+	glConfig2.depthClampAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, ARB_depth_clamp, r_arb_depth_clamp.Get() );
```
slipher, line 1947: It seems you should put ExtFlag_CORE

illwieckz: Right, I forgot that core starts with 3.2.

slipher:
LGTM

PR #1177, opened 2024-06-03
renderer: remove the `USE_ALPHA_TESTING` macro and build less shader permutations

illwieckz:
There is no need to set the `USE_ALPHA_TESTING` macro if `GLS_ATEST_NONE` is set when there is no alpha testing to do, the alpha testing computation will just know there is nothing to do.

This reduces:

- GLSL shader permutation to build.
- GLSL shader changes between textures.

The cost of doing that addition and that test in all case in GLSL should be faster than switching back and forth to a GLSL shader permutation.

This also obviously simplifies code.

I noticed this when investigating this (it will _NOT_ fix it):

- https://github.com/DaemonEngine/Daemon/issues/1172

I noticed there was a difference between `generic` and `lightMapping` where `lightMapping` did not used on that `USE_ALPHA_TESTING` macro already.


slipher:
LGTM

PR #1178, opened 2024-06-03
Move sgame trap call declarations to engine repository

slipher:


slipher:
Added another commit which fixes the cgame trap call declarations too. Those also should be in another header but at least they aren't in the gamelogic repo.

DolceTriade (approved):

PR #1145, opened 2024-05-13
renderer: introduce `IF_FITSCREEN` and `RSF_2D`

illwieckz:
Introduce `IF_FITSCREEN` to rescale image to the smaller size larger than screen.

The idea behind this feature is that we have images that should not be rescaled smaller than screen to look good:

- minimaps,
- UI background,
- levelshot,
- etc.

So for them we use `IF_PICMIP`, either directly in code, either with material keyword `nopicmip`.

But then this stupid thing happen, let's imagine someone playing on a low end device with not a lot of GPU memory and a 800×600 screen:

```
]/listImages
------------------------------------------------------------------
num   width heigth layers mm  type format   wrap.t   wrap.s   name
------------------------------------------------------------------
63      100   2230      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/bg
64       20   1200      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/black
65     2082   1371      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/duel
66       20   1200      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/black2
67     2500   1667      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/circles1
68     2500   1667      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/circles2
69     3500    886      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/fog
121    3840   2160      1 no  2D   DXT1     t.clmp   s.clmp   meta/chasm/chasm
```

The 4K levelshot is fully loaded in GPU memory just to display it on a `800×600` screen!

Here is what happens once this feature is implemented AND (not done yet) the related images are configured to use `IF_FITSCREEN`:

```
]/listImages
------------------------------------------------------------------
num   width heigth layers mm  type format   wrap.t   wrap.s   name
------------------------------------------------------------------
63       50   1115      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/bg
64       20   1200      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/black
65     1041    685      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/duel
66       20   1200      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/black2
67     1250    833      1 no  2D   DXT5     t.clmp   s.clmp   ui/assets/background/circles1
[...]

PR #1179, opened 2024-06-04
Support graphics cards without `ARB_half_float_vertex`

illwieckz:
This is to address those two issues:

1. Some graphics devices don't supports `ARB_half_float_vertex` like the Intel GMA Gen3 architecture with the Mesa driver.
2. Some graphics devices don't supports `ARB_half_float_vertex` but drivers may provide an unplayable slow emulation, like the ATi r300 architecture with the Mesa driver.
   - https://github.com/DaemonEngine/Daemon/issues/1172

Those changes purposes to provide an alternative code path when `ARB_half_float_vertex` is not available or when the `r_arb_half_float_vertex` cvar is disabled on purpose to avoid a slow emulation provided by the driver.

It is the next step over preparatory work that was already merged:

- Some improvements of our type system, helping the compiler to report more type errors and then making the conversion easier with the less risk of introducing silent errors:
  https://github.com/DaemonEngine/Daemon/pull/728
- Some enablement of GL 2.1 on the said Intel graphics chip so we can know GL 2.1 is supported but catch the lack of the half float vertex extension, meaning that if we implement a workaround one day (that is the day today), we would already have everything else already in place for it:
  https://github.com/DaemonEngine/Daemon/pull/478

This branch doesn't break compatibility with the game but I implemented it over the `illwieckz/image-fistcreen/sync` branch anyway:

- https://github.com/DaemonEngine/Daemon/pull/1145

I did it that way because this other branch also helps running the game on this kind of hardware (to not upload textures too big for such hardware) and we better test this hardware over `for-0.55.0/sync` because of all compatibility-breaking changes purposed for performance that were merged there. The `illwieckz/image-fistcreen/sync` branch itself is already implemented over `for-0.55.0/sync` because of compatibility breaking.

This branch also includes the `GL_DEPTH_CLAMP` disablement when `GL_ARB_depth_clamp` to avoid spamming the console with warnings when running the game on ATi r300 graphics card, to make debugging easier:

- https://github.com/DaemonEngine/Daemon/pull/1173

~~⚠️🚧️ **This branch is in very early state!** 🚧️⚠️~~
~~⚠️🚧️ **It's even not ready to review!** 🚧️⚠️~~
~~⚠️🚧️ **This may be full of useless halffloat-to-float-to-halffloat conversions!** 🚧️⚠️~~
~~⚠️🚧️ **I even haven't tested if it doesn't break the usual code path!** 🚧️⚠️~~

~~I push this branch now because it starts to work and this is so precious I don't want the team to lose that work is something happens to me!~~

It's now ready to be reviewed, it works for me with both code paths, and is rebased on `master`.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -4615,10 +4615,18 @@ void DebugDrawVertex(const vec3_t pos, unsigned int color, const vec2_t uv) {
 	tess.verts[ tess.numVertexes ].xyz[ 1 ] = pos[ 1 ];
 	tess.verts[ tess.numVertexes ].xyz[ 2 ] = pos[ 2 ];
 	tess.verts[ tess.numVertexes ].color = colors;
+
 	if( uv ) {
-		tess.verts[ tess.numVertexes ].texCoords[ 0 ] = floatToHalf( uv[ 0 ] );
-		tess.verts[ tess.numVertexes ].texCoords[ 1 ] = floatToHalf( uv[ 1 ] );
+		if ( glConfig2.halfFloatVertexAvailable )
+		{
+			floatToHalf2( uv, tess.verts[ tess.numVertexes ].f16TexCoords );
+		}
+		else
+		{
+			Vector2Copy( uv, tess.verts[ tess.numVertexes ].texCoords );
+		}
 	}
+
 	tess.indexes[ tess.numIndexes ] = tess.numVertexes;
 	tess.numVertexes++;
 	tess.numIndexes++;
@@ -5006,32 +5014,42 @@ const RenderCommand *StretchPicCommand::ExecuteSelf( ) const
 	tess.verts[ numVerts ].xyz[ 2 ] = 0.0f;
 	tess.verts[ numVerts + 0 ].color = backEnd.color2D;
 
-	tess.verts[ numVerts ].texCoords[ 0 ] = floatToHalf( s1 );
-	tess.verts[ numVerts ].texCoords[ 1 ] = floatToHalf( t1 );
-
 	tess.verts[ numVerts + 1 ].xyz[ 0 ] = x + w;
 	tess.verts[ numVerts + 1 ].xyz[ 1 ] = y;
 	tess.verts[ numVerts + 1 ].xyz[ 2 ] = 0.0f;
 	tess.verts[ numVerts + 1 ].color = backEnd.color2D;
-
-	tess.verts[ numVerts + 1 ].texCoords[ 0 ] = floatToHalf( s2 );
-	tess.verts[ numVerts + 1 ].texCoords[ 1 ] = floatToHalf( t1 );
 
 	tess.verts[ numVerts + 2 ].xyz[ 0 ] = x + w;
 	tess.verts[ numVerts + 2 ].xyz[ 1 ] = y + h;
 	tess.verts[ numVerts + 2 ].xyz[ 2 ] = 0.0f;
 	tess.verts[ numVerts + 2 ].color = backEnd.color2D;
 
-	tess.verts[ numVerts + 2 ].texCoords[ 0 ] = floatToHalf( s2 );
-	tess.verts[ numVerts + 2 ].texCoords[ 1 ] = floatToHalf( t2 );
-
 	tess.verts[ numVerts + 3 ].xyz[ 0 ] = x;
 	tess.verts[ numVerts + 3 ].xyz[ 1 ] = y + h;
 	tess.verts[ numVerts + 3 ].xyz[ 2 ] = 0.0f;
 	tess.verts[ numVerts + 3 ].color = backEnd.color2D;
 
-	tess.verts[ numVerts + 3 ].texCoords[ 0 ] = floatToHalf( s1 );
-	tess.verts[ numVerts + 3 ].texCoords[ 1 ] = floatToHalf( t2 );
+	if ( glConfig2.halfFloatVertexAvailable )
+	{
+		tess.verts[ numVerts ].f16TexCoords[ 0 ] = floatToHalf( s1 );
+		tess.verts[ numVerts ].f16TexCoords[ 1 ] = floatToHalf( t1 );
+
+		tess.verts[ numVerts + 1 ].f16TexCoords[ 0 ] = floatToHalf( s2 );
+		tess.verts[ numVerts + 1 ].f16TexCoords[ 1 ] = floatToHalf( t1 );
+
+		tess.verts[ numVerts + 2 ].f16TexCoords[ 0 ] = floatToHalf( s2 );
+		tess.verts[ numVerts + 2 ].f16TexCoords[ 1 ] = floatToHalf( t2 );
+
+		tess.verts[ numVerts + 3 ].f16TexCoords[ 0 ] = floatToHalf( s1 );
+		tess.verts[ numVerts + 3 ].f16TexCoords[ 1 ] = floatToHalf( t2 );
+	}
+	else
+	{
+		Vector2Set( tess.verts[ numVerts ].texCoords, s1, t1 );
+		Vector2Set( tess.verts[ numVerts + 1 ].texCoords, s2, t1 );
+		Vector2Set( tess.verts[ numVerts + 2 ].texCoords, s2, t2 );
+		Vector2Set( tess.verts[ numVerts + 3 ].texCoords, s1, t2 );
+	}
 
 	return this + 1;
 }
@@ -5085,8 +5103,14 @@ const RenderCommand *Poly2dCommand::ExecuteSelf( ) const
 		tess.verts[ tess.numVertexes ].xyz[ 1 ] = verts[ i ].xyz[ 1 ];
 		tess.verts[ tess.numVertexes ].xyz[ 2 ] = 0.0f;
 
-		tess.verts[ tess.numVertexes ].texCoords[ 0 ] = floatToHalf( verts[ i ].st[ 0 ] );
-		tess.verts[ tess.numVertexes ].texCoords[ 1 ] = floatToHalf( verts[ i ].st[ 1 ] );
+		if ( glConfig2.halfFloatVertexAvailable )
+		{
+			floatToHalf2( verts[ i ].st, tess.verts[ tess.numVertexes ].f16TexCoords );
+		}
+		else
+		{
+			Vector2Copy( verts[ i ].st, tess.verts[ tess.numVertexes ].texCoords );
+		}
 
 		tess.verts[ tess.numVertexes ].color = Color::Adapt( verts[ i ].modulate );
 		tess.numVertexes++;
@@ -5142,8 +5166,14 @@ const RenderCommand *Poly2dIndexedCommand::ExecuteSelf( ) const
 		tess.verts[ tess.numVertexes ].xyz[ 1 ] = verts[ i ].xyz[ 1 ] + translation[ 1 ];
 		tess.verts[ tess.numVertexes ].xyz[ 2 ] = 0.0f;
 
-		tess.verts[ tess.numVertexes ].texCoords[ 0 ] = floatToHalf( verts[ i ].st[ 0 ] );
-		tess.verts[ tess.numVertexes ].texCoords[ 1 ] = floatToHalf( verts[ i ].st[ 1 ] );
+		if ( glConfig2.halfFloatVertexAvailable )
+		{
+			floatToHalf2( verts[ i ].st, tess.verts[ tess.numVertexes ].f16TexCoords );
+		}
+		else
+		{
+			Vector2Copy( verts[ i ].st, tess.verts[ tess.numVertexes ].texCoords );
+		}
 
 		tess.verts[ tess.numVertexes ].color = Color::Adapt( verts[ i ].modulate );
 		tess.numVertexes++;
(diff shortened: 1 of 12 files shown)


illwieckz:
For unknown reasons, this is broken:

- mouse pointer
- minimap
- console text

illwieckz:
> For unknown reasons, this is broken:
> 
>     * mouse pointer
> 
>     * minimap
> 
>     * console text

It was just some stupid copy-past mistake.

Now it looks like everything is working.

illwieckz:
Now that I verified the branch works, the branch is now targeting `master`.

It is ready to review.

src/engine/renderer/tr_backend.cpp
```diff
@@ -5396,29 +5436,39 @@ const RenderCommand *GradientPicCommand::ExecuteSelf( ) const
...
 	tess.verts[ numVerts + 3 ].xyz[ 0 ] = x;
 	tess.verts[ numVerts + 3 ].xyz[ 1 ] = y + h;
 	tess.verts[ numVerts + 3 ].xyz[ 2 ] = 0.0f;
 
-	tess.verts[ numVerts + 3 ].texCoords[ 0 ] = floatToHalf( s1 );
-	tess.verts[ numVerts + 3 ].texCoords[ 1 ] = floatToHalf( t2 );
+	if ( glConfig2.halfFloatVertexAvailable )
+	{
+		tess.verts[ numVerts ].f16TexCoords[ 0 ] = floatToHalf( s1 );
+		tess.verts[ numVerts ].f16TexCoords[ 1 ] = floatToHalf( t2 );
```
VReaperV, line 5454: This should be `tess.verts[ numVerts ].f16TexCoords[ 1 ] = floatToHalf( t1 );`.

src/engine/renderer/tr_surface.cpp
```diff
@@ -1113,7 +1200,14 @@ static void Tess_SurfaceMD5( md5Surface_t *srf )
...
 
-			Vector2Copy( surfaceVertex->texCoords, tessVertex->texCoords );
+			if ( glConfig2.halfFloatVertexAvailable )
+			{
+				Vector2Copy( surfaceVertex->f16TexCoords, tessVertex->f16TexCoords );
+			}
+			else
+			{
+				halfToFloat2( surfaceVertex->f16TexCoords, tessVertex->texCoords );
+			}
```
VReaperV, line 1210: The original code doesn't seem to be doing any half-float conversions here? If it is, then these 2 branches should be switched.

src/engine/renderer/tr_vbo.cpp
```diff
@@ -353,7 +377,15 @@ static void R_CopyVertexData( VBO_t *vbo, byte *outData, vboData_t inData )
...
-				Vector2Copy( inData.st[ v ], ptr[ v ].texcoord );
+				if ( glConfig2.halfFloatVertexAvailable )
+				{
+					Vector2Copy( inData.f16st[ v ], ptr[ v ].f16TexCoords );
+				}
+				else
+				{
+					// Model only supports half float for now.
+					halfToFloat2( inData.f16st[ v ], ptr[ v ].texCoords );
+				}
```
VReaperV, line 388: Is this also switched around?

src/engine/renderer/tr_local.h
```diff
@@ -200,6 +215,23 @@ static inline void halfToFloat( const f16vec4_t in, vec4_t out )
...
+extern glVertexShim_t GL_vertexShim;
+
+static inline void glVertexSetHalfFloat()
+{
+	GL_vertexShim.floatFormat = GL_HALF_FLOAT;
+}
+
+static inline void glVertexSetFloat()
+{
+	GL_vertexShim.floatFormat = GL_FLOAT;
```
VReaperV, line 232: These are GLenum, which is an unsigned int, as opposed to GLint, which is just int. I guess it doesn't matter here, since all values happen to be positive, but better to use the correct type.

VReaperV (commented):
Same question w. r. t. `vector2Copy()` vs `floatToHalf()` usage where the original code only has the former and the branches in this pr seem switched up.

src/engine/client/hunk_allocator.cpp
```diff
@@ -325,6 +325,7 @@ void Hunk_FreeTempMemory( void *buf )
 
 	if ( hdr->magic != (int) HUNK_MAGIC )
 	{
+		ASSERT( false );
```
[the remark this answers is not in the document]
illwieckz: Moved to:

- https://github.com/DaemonEngine/Daemon/pull/1187

src/engine/renderer/tr_backend.cpp
illwieckz: Right.

src/engine/renderer/tr_local.h
illwieckz: Right.

src/engine/renderer/tr_surface.cpp
illwieckz: The model structs (md5, iqm…) were not modified, they all store half-floats, so we need to copy them as half floats in the final struct if half-float is supported, or convert them back to float if not.

Note that I renamed half half-float related stuff by prefixing it by `f16`, so the original `Vector2Copy( surfaceVertex->texCoords, tessVertex->texCoords );` code was copying half floats. We are now converting them to float if the hardware doesn't support half float vertexes.

This code is correct.

In the future I may submit a patch to do the reverse: to always store them as float in those model structs, and convert them to half float if supported in the last step, because the fallback CPU code seems to do a lot of useless half-to-float conversions for computing the animations, which is stupid since the fallback CPU code is run on the lowest end hardware… But such design change is not in topic of that PR.

src/engine/renderer/tr_vbo.cpp
illwieckz: It is not. As said in the other comment, the output (second argument) is what will be stored for rendering, so it should be in the format supported by the graphics card/driver, while the input (first argument) is currently always half float.

src/engine/renderer/tr_shade_calc.cpp
```diff
@@ -667,15 +720,37 @@ static void Autosprite2Deform( int firstVertex, int numVertexes, int numIndexes
...
+			int16_t factor;
+
+			if ( glConfig2.halfFloatVertexAvailable )
+			{
+				// I guess this works, since the sign bit is the MSB for both floating point and integers
+				factor = static_cast<int16_t>(v1->f16TexCoords[ 3 ].bits);
+			}
+			else
+			{
+				factor = (int16_t) v1->texCoords[ 3 ];
```
illwieckz, line 734: @slipher may have a comment on those lines.

illwieckz:
The changes are a bit verbose because I renamed all related half float variables by prefixing them with `f16` (for example, half-float `texCoords` becomes `f16TexCoords`, while float `texCoords` is named `texCoords`). Not only it helps to avoid confusion when reading, but it makes far easier to me to implement this code, because I just had to rename `texCoords` to `f16TexCoords` to have my compiler reporting me all the lines requiring an alternate code.

slipher:
As an experiment, I implemented a proof of concept of automatic vertex translation at VBO upload time in the branch `slipher/vertex-translation`. I'm not sure if that approach is better, but I was curious if it would work.

It doesn't translate *every* VBO because so far it only acts on "static" VBOs (ones fixed at load time). But I imagine it would be enough to get good performance on the cards with poor implementation quality of the half-float extension.

If we were to productionize this, the vertex translation process should probably be combined into R_CopyVertexData which already appears to do some similar work.

illwieckz:
The `slipher/vertex-translation` branch is enough to recover performance on ATI r300:

unvanquished_2024-06-12_022513_000

As a remind, I would like to not only accelerate r300 (that have a slow half-float driver emulation) but also enable some Intel cards (that also support GL 2.1 but don't provide half-float, even through emulation).

A good way to emulate with Mesa a card only having GL 2.1 and not having half float vertexes is to set some environment variable this way:

```sh
MESA_GL_VERSION_OVERRIDE='2.1' MESA_EXTENSION_OVERRIDE='-GL_ARB_half_float_vertex' ./daemon +devmap vega
```

This should work with any Mesa driver, even llvmpipe on Windows.

illwieckz:
I don't mind that much the implementation, as long as it works.

What I like in my branch is that I renamed the half float variables with an explicit prefix, and I find it better for the code to read and debug, this may be something I would redo even if we merge another implementation.

We may even decide to merge some current half float code to float, to avoid useless round trips between float and half float. I especially think about the model code: Some parts of the CPU fallback code may processes half floats, but I want this code to be as fast as possible since it's already a fallback for the GPU not being powerful enough…

illwieckz:
What is currently broken in `slipher/vertex-translation` if I make the extension not required and it is missing, is some 2D things like text and images that are either garbage or distorted in wrong ways. I'm fine with the idea of translating the data when needed, especially if it prevents us to duplicate code everywhere, and if it even allows us to add new stuff using half float and relying on a generic converter when needed.

illwieckz:
@slipher do you have some update on your `slipher/vertex-translation` branch? I like your approach, but it was not complete as far as I know.

I will have access to the machine with the Intel card without half-float vertex for two weeks starting with the next week.

If you're not available to work on this for now, I may merge my implementation and we can revert it in the future to merge your implementation instead. I would prefer to have yours, but we can take time to improve it.

slipher:
> @slipher do you have some update on your `slipher/vertex-translation` branch? I like your approach, but it was not complete as far as I know.

I made some progress in implementing a more complete version of the idea, which eliminates some of the unnecessary layers in preparing vertex data (it's often copied about 5 times before reaching the VBO). IIRC I finished 2 of the 3 model formats. Got sidetracked by all the bugs I was finding in the renderer. I'll try to dig that out that code in the next few days.

illwieckz:
> Got sidetracked by all the bugs I was finding in the renderer.

I know what it is. 🤣️

Thanks for the update! 🙂️

illwieckz:
This branch makes the Intel GMA 3 working:

unvanquished_2024-08-03_195448_000

I needed some other minor patches to not compile some shaders for disabled features like bloom, but nothing more related to vertex float was needed.

The output of `glxinfo -B`:

```
name of display: :0
display: :0  screen: 0
direct rendering: Yes
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: Mesa Project (0x8086)
    Device: i915 (chipset: G33) (0x29c2)
    Version: 24.0.9
    Accelerated: yes
    Video memory: 384MB
    Unified memory: yes
    Preferred profile: compat (0x2)
    Max core profile version: 0.0
    Max compat profile version: 2.1
    Max GLES1 profile version: 1.1
    Max GLES[23] profile version: 2.0
OpenGL vendor string: Mesa Project
OpenGL renderer string: i915 (chipset: G33)
OpenGL version string: 2.1 Mesa 24.0.9-0ubuntu0.1
OpenGL shading language version string: 1.20

OpenGL ES profile version string: OpenGL ES 2.0 Mesa 24.0.9-0ubuntu0.1
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 1.0.16
```

slipher:
You can see my current progress on the branch [slipher/runtime-vbo-layout](https://github.com/DaemonEngine/Daemon/compare/slipher/runtime-vbo-layout). The dynamically generated vertex attribute layout, which may be configured to use half float or not, is used for IQM, MD3, and MD5 models. It's not ready to test as other VBOs are not ported yet. Remaining steps:

- Investigate the vertex attribute-related warning spam on station15. IIRC this is caused by some code which creates some verts with the qtangent attribute, then later overwrites that data with the incompatible orientation attribute. I should be careful to avoid breaking this.
- Use the new vertex attribute layout method for the other static VBOs.
- Change the dynamic VBO's vertex struct to (always) use float instead of f16_t. This is probably better anyway. Data in the dynamic VBO is only used once and (I believe; will double check) is usually small. Trying to compress it seems pointless. Also tess.verts is sometimes used for data which is not actually sent to the GPU, so avoiding useless conversions should be a win there.

slipher:
I spelled out my [thoughts](https://github.com/DaemonEngine/Daemon/issues/1236) on the extension cvar naming. To me it looks like a case of cargo culting gone out of control. Maybe some of the first extensions were `EXT` ones, which makes some sense as part of the cvar name. But ARB really doesn't.

illwieckz:
I rebased on latest `master`, which reduced the diff size as some code as some was just NUKED since.

I added a detection of early RV300 hardware were we should prefer non-half float, and added a workaround cvar that can be disabled to test the r300 driver half-float performance when not using such workaround.

This looks ready to me.

Regarding to the [slipher/runtime-vbo-layout](https://github.com/DaemonEngine/Daemon/compare/slipher/runtime-vbo-layout) branch, I do believe it is the future, but I prefer to see that branch rebased on this one. This one also brings some optimizations and some renaming that makes easier to follow the code and I would like to merge that in all cases anyway. 

We can already merge that as it already works, without waiting for `slipher/runtime-vbo-layout`. While I do believe such deeper redesign is the future, it can wait a bit.

Basically a rebase of  `slipher/runtime-vbo-layout` would have to add an extra parent commit to unsplit the float/half-float if/else code paths. I can help at rebasing it.

slipher:
I just looked at my vertex layout branch again and it's actually 95% done. The big thing it was waiting on was the autosprite rework (so that we no longer have overlapping attributes). I can have the PR ready in the next day or two.

illwieckz:
Great!

illwieckz:
Obsoleted by #1344:

- https://github.com/DaemonEngine/Daemon/pull/1344

slipher:
We still need the commit to add a *workaround* flipping off half float vertex for the driver with poor quality of implementation.

illwieckz:
Yes, I'm on it. :-)

illwieckz:
See:

- https://github.com/DaemonEngine/Daemon/pull/1347

PR #1180, opened 2024-06-04
Material system: Implement GPU occlusion culling

VReaperV:
Implement occlusion culling in compute shaders using a Hierarchical Depth-Buffer.

The depth buffer is processed before the post-processing pass, reducing it to the next mip level until a 1x1 resolution is reached.
Then it's used to find out if a surface that passed frustum culling is behind some opaque surfaces. This is done by getting the screen-space AABB from the bounding sphere of the surface, and then comparing its depth to the depth in the appropriate level of the depth-reduction image. The level of the depth image is chosen based on how much of the screen space it's taking up.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -42,6 +42,10 @@ GLSSBO surfaceCommandsSSBO( "surfaceCommands", 2, GL_MAP_WRITE_BIT | GL_MAP_PERS
 GLBuffer culledCommandsBuffer( "culledCommands", 3, GL_MAP_WRITE_BIT | GL_MAP_PERSISTENT_BIT, GL_MAP_FLUSH_EXPLICIT_BIT );
 GLUBO surfaceBatchesUBO( "surfaceBatches", 0, GL_MAP_WRITE_BIT, GL_MAP_INVALIDATE_RANGE_BIT );
 GLBuffer atomicCommandCountersBuffer( "atomicCommandCounters", 4, GL_MAP_WRITE_BIT | GL_MAP_PERSISTENT_BIT, GL_MAP_FLUSH_EXPLICIT_BIT );
+GLSSBO portalSurfacesSSBO( "portalSurfaces", 5, GL_MAP_READ_BIT | GL_MAP_PERSISTENT_BIT, 0 );
+
+PortalView portalStack[MAX_VIEWS];
+
 MaterialSystem materialSystem;
 
 static void ComputeDynamics( shaderStage_t* pStage ) {
@@ -245,7 +249,7 @@ static void UpdateSurfaceDataGeneric( uint32_t* materials, Material& material, d
 	// u_DeformGen
 	gl_genericShaderMaterial->SetUniform_Time( backEnd.refdef.floatTime - backEnd.currentEntity->e.shaderTime );
 
-	// bind u_ColorMap=
+	// bind u_ColorMap
 	if ( pStage->type == stageType_t::ST_STYLELIGHTMAP ) {
 		gl_genericShaderMaterial->SetUniform_ColorMapBindless(
 			GL_BindToTMU( 0, GetLightMap( drawSurf ) )
@@ -1005,7 +1009,7 @@ void MaterialSystem::GenerateWorldCommandBuffer() {
 
 			const uint32_t cmdCount = material.drawCommands.size();
 			const uint32_t batchCount = cmdCount % SURFACE_COMMANDS_PER_BATCH == 0 ? cmdCount / SURFACE_COMMANDS_PER_BATCH
-																			   : cmdCount / SURFACE_COMMANDS_PER_BATCH + 1;
+				: cmdCount / SURFACE_COMMANDS_PER_BATCH + 1;
 
 			material.surfaceCommandBatchOffset = batchOffset;
 			material.surfaceCommandBatchCount = batchCount;
@@ -1145,6 +1149,28 @@ void MaterialSystem::GenerateWorldCommandBuffer() {
 	surfaceBatchesUBO.UnmapBuffer();
 
 	GL_CheckErrors();
+}
+
+void MaterialSystem::GenerateDepthImages( const int width, const int height, imageParams_t imageParms ) {
+	int size = std::max( width, height );
+	imageParms.bits ^= ( IF_NOPICMIP | IF_PACKED_DEPTH24_STENCIL8 );
+	imageParms.bits |= IF_ONECOMP32F;
+
+	depthImageLevels = 0;
+	while ( size > 0 ) {
+		depthImageLevels++;
+		size >>= 1; // mipmaps round down
+	}
+
+	depthImage = R_CreateImage( "_depthImage", nullptr, width, height, depthImageLevels, imageParms );
+	GL_Bind( depthImage );
+	int mipmapWidth = width;
+	int mipmapHeight = height;
+	for ( int j = 0; j < depthImageLevels; j++ ) {
+		glTexImage2D( GL_TEXTURE_2D, j, GL_R32F, mipmapWidth, mipmapHeight, 0, GL_RED, GL_FLOAT, nullptr );
+		mipmapWidth >>= 1;
+		mipmapHeight >>= 1;
+	}
 }
 
 static void BindShaderGeneric( Material* material ) {
@@ -1386,6 +1412,8 @@ void MaterialSystem::GenerateWorldMaterials() {
 	generatingWorldCommandBuffer = true;
 
 	Log::Debug( "Generating world materials" );
+
+	R_SyncRenderThread();
 
 	R_AddWorldSurfaces();
 
@@ -1756,14 +1784,6 @@ void MaterialSystem::UpdateDynamicSurfaces() {
 }
 
 void MaterialSystem::UpdateFrameData() {
-	/* atomicCommandCountersBuffer.AreaIncr();
-
-	atomicCommandCountersBuffer.BindBuffer( GL_ATOMIC_COUNTER_BUFFER );
-	uint32_t* atomicCommandCounters = atomicCommandCountersBuffer.GetCurrentAreaData();
-	memset( atomicCommandCounters, 0, MAX_COMMAND_COUNTERS * sizeof(uint32_t));
-	atomicCommandCountersBuffer.FlushCurrentArea( GL_ATOMIC_COUNTER_BUFFER );
-	atomicCommandCountersBuffer.UnBindBuffer( GL_ATOMIC_COUNTER_BUFFER ); */
-	
 	atomicCommandCountersBuffer.BindBufferBase( GL_SHADER_STORAGE_BUFFER );
 	gl_clearSurfacesShader->BindProgram( 0 );
 	gl_clearSurfacesShader->SetUniform_Frame( nextFrame );
@@ -1773,19 +1793,75 @@ void MaterialSystem::UpdateFrameData() {
 	GL_CheckErrors();
 }
 
-void MaterialSystem::QueueSurfaceCull( const uint32_t viewID, const frustum_t* frustum ) {
+void MaterialSystem::QueueSurfaceCull( const uint32_t viewID, const vec3_t origin, const frustum_t* frustum ) {
+	VectorCopy( origin, frames[nextFrame].viewFrames[viewID].origin );
 	memcpy( frames[nextFrame].viewFrames[viewID].frustum, frustum, sizeof( frustum_t ) );
 	frames[nextFrame].viewCount++;
 }
 
+void MaterialSystem::DepthReduction() {
+	if ( r_lockpvs->integer ) {
+		if ( !PVSLocked ) {
+			lockedDepthImage = depthImage;
+		}
+
+		return;
+	}
+
+	int width = depthImage->width;
+	int height = depthImage->height;
+
+	gl_depthReductionShader->BindProgram( 0 );
+
+	uint32_t globalWorkgroupX = ( width + 7 ) / 8;
+	uint32_t globalWorkgroupY = ( height + 7 ) / 8;
+
+	GL_Bind( tr.currentDepthImage );
+	glBindImageTexture( 2, depthImage->texnum, 0, GL_FALSE, 0, GL_WRITE_ONLY, GL_R32F );
+
+	gl_depthReductionShader->SetUniform_InitialDepthLevel( true );
+	gl_depthReductionShader->SetUniform_ViewWidth( width );
+	gl_depthReductionShader->SetUniform_ViewHeight( height );
+	gl_depthReductionShader->DispatchCompute( globalWorkgroupX, globalWorkgroupY, 1 );
+
+	for ( int i = 0; i < depthImageLevels; i++ ) {
+		width = width > 1 ? width >> 1 : 1;
+		height = height > 1 ? height >> 1 : 1;
+
+		globalWorkgroupX = ( width + 7 ) / 8;
+		globalWorkgroupY = ( height + 7 ) / 8;
+
+		glBindImageTexture( 1, depthImage->texnum, i, GL_FALSE, 0, GL_READ_ONLY, GL_R32F );
+		glBindImageTexture( 2, depthImage->texnum, i + 1, GL_FALSE, 0, GL_WRITE_ONLY, GL_R32F );
+
+		gl_depthReductionShader->SetUniform_InitialDepthLevel( false );
+		gl_depthReductionShader->SetUniform_ViewWidth( width );
+		gl_depthReductionShader->SetUniform_ViewHeight( height );
+		gl_depthReductionShader->DispatchCompute( globalWorkgroupX, globalWorkgroupY, 1 );
+
+		glMemoryBarrier( GL_SHADER_IMAGE_ACCESS_BARRIER_BIT );
+	}
+}
+
 void MaterialSystem::CullSurfaces() {
+	if ( r_gpuOcclusionCulling.Get() ) {
+		DepthReduction();
+	}
+
 	surfaceDescriptorsSSBO.BindBufferBase();
 	surfaceCommandsSSBO.BindBufferBase();
 	culledCommandsBuffer.BindBufferBase( GL_SHADER_STORAGE_BUFFER );
 	surfaceBatchesUBO.BindBufferBase();
 	atomicCommandCountersBuffer.BindBufferBase( GL_ATOMIC_COUNTER_BUFFER );
 
+	if ( totalPortals > 0 ) {
+		portalSurfacesSSBO.BindBufferBase();
+	}
+
+	GL_CheckErrors();
+
 	for ( uint32_t view = 0; view < frames[nextFrame].viewCount; view++ ) {
+		vec3_t origin;
 		frustum_t* frustum = &frames[nextFrame].viewFrames[view].frustum;
 
 		vec4_t frustumPlanes[6];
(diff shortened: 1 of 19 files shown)


VReaperV:
The sphere projection function is giving incorrect results apparently, since I'm pretty sure the world->view space conversion is correct. I'm also not seeing much of any difference in frame time between this change and frustum culling-only: the frame time that I win in having less draws is nullified by the hi-z pass. It could be improved a bit by making the depth buffer downsampling be done in one compute dispatch, but it might just not really be worth it. On the other hand the hi-z buffer could be used for some other effects too.

VReaperV:
Well, I'm seeing some improvements in certain cases, especially with high graphics settings. The whole thing also isn't taking too much frame time in itself anyway (the indirect draws + cull compute).

VReaperV:
So the lower framerate that illwieckz was getting in #1137 might have to do with something else. Ideally we could also use better synchronization, but I'm not sure that would be possible.

VReaperV:
Possibly rasterizing the bounding boxes with colour/depth writes disabled to determine object visibility would be faster. The depth pre-pass is super fast wherever I looked at it here, so this should be fast too. It does, however, mean that object visibility would be lagging behind by a frame, but that's already happening with this hi-z algorithm.

It would also require a bunch of writes to a buffer from every rasterized fragment that passed the depth test, which is not ideal. Perhaps a custom rasterizer in a compute shader that would instead write the visible object ids once per triangle or a batch of triangles would be better.

VReaperV:
There are a few more things that could be improved independent of occlusion culling that might make this work faster in every case anyway:
1. Use batches of e. g. 64 triangles instead of specifying surfaces, which would allow dropping the `processSurfaces_cp` shader and doing writes in the `cull_cp` shader instead.
2. That would also allow creating an index buffer instead of an indirect buffer, for the most part. There would still be one draw per material, but that draw would have all the triangles in it (that weren't culled).
3. Potentially with triangle batches culling at triangle batch level might speed things up even more: e. g. with cone culling.

VReaperV:
Another potential advantage of compute rasterizer is assigning light sources to a tile/cluster buffer, though with low light counts it might not make much of a difference.

VReaperV:
Doing hardware rasterization of BBoxes and writing to a buffer from the fragment shader with an early discard doesn't work well for some reason.

VReaperV:
I believe this fixes the incorrect projection. However there is still some incorrect culling, which might be due to incorrect depth downsampling with even/odd dimensions, haven't checked yet.

VReaperV:
This mostly works now, however there's still an error with AABB depth projection, and some things need to be un-hardcoded.

src/engine/renderer/Material.cpp
```diff
@@ -1887,10 +2239,47 @@
...
+		switch ( material.stageType ) {
+			case stageType_t::ST_LIGHTMAP:
+			case stageType_t::ST_DIFFUSEMAP:
+			case stageType_t::ST_COLLAPSE_COLORMAP:
+			case stageType_t::ST_COLLAPSE_DIFFUSEMAP:
+				gl_lightMappingShaderMaterial->SetUniform_ShowTris( 1 );
+				GL_State( GLS_DEPTHTEST_DISABLE );
+				glMultiDrawElementsIndirectCountARB( GL_LINES, GL_UNSIGNED_INT,
+					BUFFER_OFFSET( material.surfaceCommandBatchOffset * SURFACE_COMMANDS_PER_BATCH * sizeof( GLIndirectBuffer::GLIndirectCommand )
+						+ ( culledCommandsCount * ( MAX_VIEWS * currentFrame + viewID )
```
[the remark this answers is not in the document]
VReaperV: We'd need millions of surfaces to exceed this, so it's fine.

VReaperV:
Added some comments and made `r_lockPVS 1` work with occlusion culling.

VReaperV:
Rebased + optimised the depth reduction a bit.

VReaperV:
Cleaned up a bit, and added a toggle through `r_gpuOcclusionCulling`.

src/engine/renderer/Material.cpp
```diff
@@ -989,40 +995,190 @@ void MaterialSystem::GenerateWorldMaterialsBuffer() {
...
+
+	uint32_t batchOffset = 0;
+	uint32_t globalID = 0;
+	for ( MaterialPack& pack : materialPacks ) {
+		for ( Material& material : pack.materials ) {
+			material.surfaceCommandBatchOffset = batchOffset;
+
+			const uint32_t cmdCount = material.drawCommands.size();
+			const uint32_t batchCount = cmdCount % SURFACE_COMMANDS_PER_BATCH == 0 ? cmdCount / SURFACE_COMMANDS_PER_BATCH
+																			   : cmdCount / SURFACE_COMMANDS_PER_BATCH + 1;
```
illwieckz, line 1008: That's a lot of space, you can park a whole train there! 😅️

src/engine/renderer/Material.cpp
```diff
@@ -1887,10 +2232,47 @@ void MaterialSystem::RenderMaterial( Material& material ) {
...
+	culledCommandsBuffer.BindBuffer( GL_DRAW_INDIRECT_BUFFER );
+
+	atomicCommandCountersBuffer.BindBuffer( GL_PARAMETER_BUFFER_ARB );
+
+	glMultiDrawElementsIndirectCountARB( GL_TRIANGLES, GL_UNSIGNED_INT,
+		BUFFER_OFFSET( material.surfaceCommandBatchOffset * SURFACE_COMMANDS_PER_BATCH * sizeof( GLIndirectBuffer::GLIndirectCommand )
+					   + ( culledCommandsCount * ( MAX_VIEWS * currentFrame + viewID )
+					   * sizeof( GLIndirectBuffer::GLIndirectCommand ) ) ),
+					   //+ ( culledCommandsCount * ( MAX_VIEWS * currentFrame + currentView )
+						   //* sizeof( GLIndirectBuffer::GLIndirectCommand ) ),
```
illwieckz, line 2244: Some space inconsistency there.

src/engine/renderer/Material.cpp
```diff
@@ -1887,10 +2232,47 @@ void MaterialSystem::RenderMaterial( Material& material ) {
...
+			case stageType_t::ST_COLLAPSE_COLORMAP:
+			case stageType_t::ST_COLLAPSE_DIFFUSEMAP:
+				gl_lightMappingShaderMaterial->SetUniform_ShowTris( 1 );
+				GL_State( GLS_DEPTHTEST_DISABLE );
+				glMultiDrawElementsIndirectCountARB( GL_LINES, GL_UNSIGNED_INT,
+					BUFFER_OFFSET( material.surfaceCommandBatchOffset * SURFACE_COMMANDS_PER_BATCH * sizeof( GLIndirectBuffer::GLIndirectCommand )
+						+ ( culledCommandsCount * ( MAX_VIEWS * currentFrame + viewID )
+							* sizeof( GLIndirectBuffer::GLIndirectCommand ) ) ),
+					//+ ( culledCommandsCount * ( MAX_VIEWS * currentFrame + currentView )
+						//* sizeof( GLIndirectBuffer::GLIndirectCommand ) ),
```
illwieckz, line 2262: Some space inconsistency there.

illwieckz (dismissed):
I'm not sure to understand everything, but the overall looks good to me and I confirm it fixes many bugs.

I would appreciate some better spacing:

20240730-215050-000 spaces

I also noticed some new GLSL files use spaces or seem to mix tabs and spaces, I would prefer tabs.

slipher:
> This works by creating a depth-reduction image from the depth buffer, then using it on the next frame.

How can data generated on the previous frame be valid? Everything could be in a different position or viewed from a different angle.

VReaperV:
> How can data generated on the previous frame be valid? Everything could be in a different position or viewed from a different angle.

That's a mistake in pr description, what is meant is that the results of the culling are used on the next frame, i. e. double-buffered (the depth buffer is used from the current frame).

VReaperV:
Fixed pr description.

src/engine/renderer/gl_shader.h
```diff
@@ -3190,6 +3250,54 @@ class u_TotalDrawSurfs :
...
+	u_ShowTris( GLShader* shader ) :
+		GLUniform1Bool( shader, "u_ShowTris" ) {
+	}
+
+	void SetUniform_ShowTris( const int showTris ) {
+		this->SetValue( showTris );
+	}
+};
+
+class u_CameraPosition :
```
slipher, line 3329: Unused?

src/engine/renderer/glsl_source/cull_cp.glsl
```diff
@@ -73,21 +75,97 @@ struct Plane {
...
+
+    // Frustum culling
     // Skip far plane because we always make it encompass the whole map in the current direction
     // This might need to be changed later for shadowmaps since lights could have some far plane set
     for( int i = 0; i < 5; i++ ) {
         const float distance = dot( u_Frustum[i].xyz, boundingSphere.center ) - u_Frustum[i].w;
 
         if( distance < -boundingSphere.radius ) {
-            return true;
+            culled = u_UseFrustumCulling;
```
slipher, line 134: Since u_UseFrustumCulling is a uniform value couldn't we put the for loop inside an `if`?

src/engine/renderer/glsl_source/cull_cp.glsl
```diff
@@ -73,21 +75,97 @@ struct Plane {
...
+
+        // Sample the depth-reduction image at the mip-level where the surface covers 4 pixels
+        const int level = int( floor( log2( max( width, height ) ) ) );
+
+        // Coords for the 4 pixels covered by the AABB, adjusted for partially off-screen surfaces as needed
+        vec2 surfaceCoords = vec2( u_ViewWidth >> level, u_ViewHeight >> level );
+        surfaceCoords *= ( boundingBox.xy + boundingBox.zw ) * 0.5;
+        const ivec2 surfaceCoordsFloor = ivec2( clamp( surfaceCoords.x, 0.0, float( u_ViewWidth >> level ) ), clamp( surfaceCoords.y, 0.0, float( u_ViewHeight >> level ) ) );
+        ivec4 depthCoords = ivec4( surfaceCoordsFloor.xy,
+                                   surfaceCoordsFloor.x + ( surfaceCoords.x - surfaceCoordsFloor.x >= 0.5 ? 1 : -1 ),
```
slipher, line 160: The object won't always occupy 2 pixels in both dimensions, right? It could have some oblong shape and be 1 pixel wide in the x axis and 2 in the y axis for example. I understand this is not a bug as the worst thing that could happen here is a surface not being culled, but I wonder if there isn't another way to do the calculation.

src/engine/renderer/Material.h
```diff
@@ -224,6 +226,7 @@ class MaterialSystem {
 	};
 
 	bool frameStart = false;
+	uint32_t testTex[2];
```
slipher, line 237: Unused?

src/engine/renderer/Material.cpp
```diff
@@ -1778,7 +1802,56 @@ void MaterialSystem::QueueSurfaceCull( const uint32_t viewID, const frustum_t* f
...
+	gl_depthReductionShader->SetUniform_InitialDepthLevel( true );
+	gl_depthReductionShader->SetUniform_ViewWidth( width );
+	gl_depthReductionShader->SetUniform_ViewHeight( height );
+	gl_depthReductionShader->DispatchCompute( globalWorkgroupX, globalWorkgroupY, 1 );
+
+	for ( int i = 0; i < depthImageLevels; i++ ) {
+		width = width > 1 ? width >> 1 : 1;
+		height = height > 1 ? height >> 1 : 1;
+
+		globalWorkgroupX = width % 8 == 0 ? width / 8 : width / 8 + 1;
```
slipher, line 1836: `(width + 7) / 8`?

src/engine/renderer/Material.cpp
```diff
@@ -1778,7 +1802,56 @@ void MaterialSystem::QueueSurfaceCull( const uint32_t viewID, const frustum_t* f
...
+
+		glBindImageTexture( 1, frames[nextFrame].depthImage->texnum, i, GL_FALSE, 0, GL_READ_ONLY, GL_R32F );
+		glBindImageTexture( 2, frames[nextFrame].depthImage->texnum,i + 1, GL_FALSE, 0, GL_WRITE_ONLY, GL_R32F );
+
+		gl_depthReductionShader->SetUniform_InitialDepthLevel( false );
+		gl_depthReductionShader->SetUniform_ViewWidth( width );
+		gl_depthReductionShader->SetUniform_ViewHeight( height );
+		gl_depthReductionShader->DispatchCompute( globalWorkgroupX, globalWorkgroupY, 1 );
+
+		glMemoryBarrier( GL_SHADER_IMAGE_ACCESS_BARRIER_BIT );
```
slipher, line 1847: Is this needed before the first iteration as well?

src/engine/renderer/Material.cpp
```diff
@@ -1147,6 +1147,38 @@ void MaterialSystem::GenerateWorldCommandBuffer() {
...
+	imageParms.bits ^= ( IF_NOPICMIP | IF_PACKED_DEPTH24_STENCIL8 );
+	imageParms.bits |= IF_ONECOMP32F;
+
+	depthImageLevels = 0;
+	while ( size > 0 ) {
+		depthImageLevels++;
+		size >>= 1; // mipmaps round down
+	}
+
+	byte* data = new byte[width * height * 4];
```
slipher, line 1163: `std::unique_ptr<byte> data( new byte[ width * height * 4 ] )`

src/engine/renderer/Material.cpp
VReaperV: No.

src/engine/renderer/Material.cpp
VReaperV: That can likely be removed with nullptr passed to the function instead, will need to check that later. Either way I don't see the point of creating a wrapper for an allocation that only exists within this function and is freed 15 lines later.

src/engine/renderer/glsl_source/cull_cp.glsl
VReaperV: Right, done now.

src/engine/renderer/gl_shader.h
VReaperV: Used now to get distance to portals.

src/engine/renderer/tr_image.cpp
```diff
@@ -2475,7 +2476,9 @@ static void R_CreateCurrentRenderImage()
 	imageParams.filterType = filterType_t::FT_NEAREST;
 	imageParams.wrapType = wrapTypeEnum_t::WT_CLAMP;
 
-	tr.currentDepthImage = R_CreateImage( "_currentDepth", nullptr, width, height, 1, imageParams );
+	// tr.currentDepthImage = R_CreateImage( "_currentDepth", nullptr, width, height, 1, imageParams );
+
+	materialSystem.GenerateDepthImages( width, height, imageParams );
```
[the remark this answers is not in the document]
VReaperV: I've changed this a bit now, only one R32F image is now created, and it's only done if material system is available.

src/engine/renderer/Material.h
VReaperV: Yep, removed.

src/engine/renderer/Material.cpp
VReaperV: Done.

src/engine/renderer/Material.cpp
VReaperV: Removed `data`, changed this to just pass nullptr instead.

src/engine/renderer/glsl_source/cull_cp.glsl
VReaperV: That can happen sometimes, yes. I've tried wrapping the texel fetches in if statements so only the first one is guaranteed and the next ones are only done if the new coordinates don't match any of the previous ones, but some surfaces became invisible when they shouldn't be. Not sure why as I couldn't spot what the issue was. This can be improved later though.

VReaperV: There's also an extension that would allow doing this with just one regular texture sample, but the availability of that extension is not enough to be the only option.

VReaperV: Ah wait, it wasn't working properly because I accidentally used `min` instead of `max`. Anyway though, with that change it seems to be a tiny bit slower (<0.1ms). Any consecutive texel fetches with the same coordinates will likely just come from the same texture cache anyway.

src/engine/renderer/Material.cpp
VReaperV: Fixed.

src/engine/renderer/Material.cpp
VReaperV: Fixed, also removed the commented out code.

src/engine/renderer/Material.cpp
VReaperV: Heh, xhanged the spacing now.

VReaperV:
> I also noticed some new GLSL files use spaces or seem to mix tabs and spaces, I would prefer tabs.

Ah, yep, that should be fixed now.

VReaperV:
I've added a portal readback as well: now instead of just calling `R_MirrorViewBySurface()` for every portal surface after sorting them by distance each frame, the `cull_cp` shader will cull portal surfaces in addition to regular ones, then the CPU will read the results back next frame and use that to determine which portals might be visible.

src/engine/renderer/Material.cpp
```diff
@@ -2150,12 +2343,31 @@
...
+		switch ( material.stageType ) {
+			case stageType_t::ST_LIGHTMAP:
+			case stageType_t::ST_DIFFUSEMAP:
+			case stageType_t::ST_COLLAPSE_COLORMAP:
+			case stageType_t::ST_COLLAPSE_DIFFUSEMAP:
+				gl_lightMappingShaderMaterial->SetUniform_ShowTris( 1 );
+				GL_State( GLS_DEPTHTEST_DISABLE );
+				glMultiDrawElementsIndirectCountARB( GL_LINES, GL_UNSIGNED_INT,
+					BUFFER_OFFSET( material.surfaceCommandBatchOffset * SURFACE_COMMANDS_PER_BATCH * sizeof( GLIndirectBuffer::GLIndirectCommand )
+					+ ( culledCommandsCount * ( MAX_VIEWS * currentFrame + viewID )
```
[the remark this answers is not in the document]
VReaperV: We've been over this, not gonna happen until we get maps with millions of surfaces.

VReaperV:
Adding `R_SyncRenderThread()` to `GenerateWorldMaterials()` also seems to have fixed crashes with `r_smp 1`, so I have disabled that restriction.

src/engine/renderer/Material.h
```diff
@@ -130,13 +130,23 @@ struct Material {
...
 
 	uint32_t drawSurfID;
+	float distance;
+	vec2_t padding;
 };
 
+struct PortalView {
+	uint32_t count;
+	drawSurf_t* drawSurf;
+	uint32_t views[MAX_VIEWS];
```
slipher, line 145: What do these represent? Portals directly visible from this one?

src/engine/renderer/glsl_source/cull_cp.glsl
```diff
@@ -56,59 +58,164 @@ struct GLIndirectCommand {
...
+								  + gl_GlobalInvocationID.x;
+
+	// Portals
+	const uint portalID = globalInvocationID - u_FirstPortalGroup * 64;
+	if( globalGroupID >= u_FirstPortalGroup && ( portalID < u_TotalPortals ) ) {
+		const uint portalSurfaceID = portalID + ( u_Frame * MAX_VIEWS + u_ViewID ) * u_TotalPortals;
+		PortalSurface surface = portalSurfaces[portalSurfaceID];
+		bool culled = CullSurface( surface.boundingSphere );
+
+		portalSurfaces[portalSurfaceID].distance = !culled ? distance( u_CameraPosition, surface.boundingSphere.center ) : -1.0;
```
slipher, line 209: If there are nested portals wouldn't the camera position need to be different?

src/engine/renderer/tr_main.cpp
```diff
@@ -2130,7 +2132,22 @@ static void R_SortDrawSurfs()
...
+			currentView++;
+		}
+
+		for ( uint32_t i = 0; i < portalStack[currentView].count; i++ ) {
+			uint32_t viewID = portalStack[currentView].views[i];
+			if ( viewID == 0 ) {
+				break;
+			}
+
+			R_MirrorViewBySurface( portalStack[portalStack[currentView].views[i]].drawSurf );
```
slipher, line 2148: R_MirrorViewBySurface was already called a view lines above, and it does the entire portal/RenderView mutual recursion, so calling it again seems odd.

slipher (commented):
Could we get a comment somewhere with an overview of how the double buffering works and what is the benefit of it? Like is there an async API for the compute shaders? If world surfaces are double-buffered but models are not, how do we keep them in sync when rendering?

src/engine/renderer/glsl_source/cull_cp.glsl
VReaperV: It would and it is, the correct position is passed to `QueueSurfaceCull()`.

VReaperV:
> Could we get a comment somewhere with an overview of how the double buffering works and what is the benefit of it? Like is there an async API for the compute shaders? If world surfaces are double-buffered but models are not, how do we keep them in sync when rendering?

There's this graph from #1137 which holds true here as well:
image
The benefit is not introducing stalls with extra synchronisation, so they can keep executing alongside the regular pipeline. Compute shaders are async by nature, so nothing extra needs to be done for that.
Double-buffering only applies to which surfaces are rendered, not how they are rendered. And since world and model surfaces are culled in completely different ways synchronisation between them does not make sense.

slipher:
LGTM, but I would still like to understand the double buffering thing. I guess I never really understood the answer to this previous conversation:

> > How can data generated on the previous frame be valid? Everything could be in a different position or viewed from a different angle.
> 
> That's a mistake in pr description, what is meant is that the results of the culling are used on the next frame, i. e. double-buffered (the depth buffer is used from the current frame).

I don't get how surfaces can be double buffered without also buffering everything else that may be rendered (particle effects, 2D drawing, etc.). If you used the surface visibility from an earlier time then the player could have moved to a point that makes some of them newly visible.

VReaperV:
(rebased)

VReaperV:
> If you used the surface visibility from an earlier time then the player could have moved to a point that makes some of them newly visible.

Technically yes, but practically the fact that it's only one frame of difference makes it more or less impossible, even when moving around fast. So far I've not seen any discrepancies when using the engine with this change.

It's also aided by the fact that the cull takes bounding spheres of the surfaces (either directly or projecting an AABB from them), so the culling somewhat coarse.

PR #1181, opened 2024-06-07
Convert vector macros to template functions

slipher:
The point of this is to prevent the arguments from being evaluated multiple times. In the commit message I give an example of this happening in cgame.

Also it can be useful for assigning a temporary GLM vector to a vec3_t. This is one more GLM/vec3_t adaptor we need in addition to [GLM4READ and GLM4RW](https://github.com/Unvanquished/Unvanquished/pull/3021). I was motivated to make this branch upon seeing, in a GLM conversion branch, VectorCopy used to assign a function's GLM return value to a vec3_t (heretofore a bad idea as the function is called 3 times).

### DIFF
--- a/src/engine/qcommon/q_math.cpp
+++ b/src/engine/qcommon/q_math.cpp
@@ -36,6 +36,9 @@
 
 #include "q_shared.h"
 
+#define DotProduct4(x, y)            (( x )[ 0 ] * ( y )[ 0 ] + ( x )[ 1 ] * ( y )[ 1 ] + ( x )[ 2 ] * ( y )[ 2 ] + ( x )[ 3 ] * ( y )[ 3 ] )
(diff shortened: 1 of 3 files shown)


src/engine/qcommon/q_math.cpp
```diff
@@ -36,6 +36,9 @@
 
 #include "q_shared.h"
 
+#define DotProduct4(x, y)            (( x )[ 0 ] * ( y )[ 0 ] + ( x )[ 1 ] * ( y )[ 1 ] + ( x )[ 2 ] * ( y )[ 2 ] + ( x )[ 3 ] * ( y )[ 3 ] )
+#define VectorMultiply( a,b,c )      ( ( c )[ 0 ] = ( a )[ 0 ] * ( b )[ 0 ],( c )[ 1 ] = ( a )[ 1 ] * ( b )[ 1 ],( c )[ 2 ] = ( a )[ 2 ] * ( b )[ 2 ] )
+
```
DolceTriade, line 41: why not these two?

DolceTriade (approved):

src/engine/qcommon/q_math.cpp
slipher: Since they were only used in one file, it seemed to me they are not suitable for inclusion in the header included by all files. And with only 1-2 uses in 1 file they should probably be normal functions not templates, but I didn't want to spend time on them.

PR #1182, opened 2024-06-08
renderer: rewrite the GLSL extension enablement

illwieckz:
- trust `GLimp_InitExtensions` and deduplicate the code
- unify the way to tell an extension is available


VReaperV:
Should all of the changed `#ifdef`s follow one convention with regards to `#ifdef` vs `#if defined`?

illwieckz:
> Should all of the changed `#ifdef`s follow one convention with regards to `#ifdef` vs `#if defined`?

Right, now unified.

VReaperV (approved):
LGTM

PR #1184, opened 2024-06-11
fix bad merge

illwieckz:
Fix bad merge that happened when merging `master` over `for-0.55.0/sync`.


illwieckz:
This is a fixup for a bad merge that breaks build and then breaks CI, let's merge it quick.

PR #1186, opened 2024-06-11
tr_shader: always iterate grouped stages

### LINKED ISSUE #1185: White floor in map Tremor
On master much of the floor and other textures in `map-tremor_1.2.1.dpk` is replaced with all white. I have a build from April 12th that doesn't have this problem.

unvanquished_2024-06-10_214738_000

illwieckz:
Fixes #1185:

- https://github.com/DaemonEngine/Daemon/issues/1185

Alongside fixing #1185, this is something I wanted to do anyway, in order to move the next `for` loop after the first `GroupActiveStages` call to a dedicated function (this is a code validating shader stages and disabling invalid ones).

The bug is that we are now more agressive at disabling unsupported stages as soon as possible, and the stage validation code was stoping iterating the stages on first disabled stage found.

The material looks like this:

```
tremor/54ECD65C96E1201DEABD3DDB5B656605
{
	{
		map $lightmap
		rgbGen identity
	}
	
	// Q3Map2 custom lightstyle stage(s)
	{
		map $lightmap
		blendFunc GL_SRC_ALPHA GL_ONE
		rgbGen wave sawtooth 0 1 0 0.5 // style 1
		tcGen lightmap
		tcMod transform 1 0 0 1 0.45410 0.18457
	}

	{
		map textures/tremor/grate3.tga
		rgbGen identity
	}
}
```

Light styles are not supported on light modes ≠ 3 (light mapping), so it is expected the material is turned into this:

```
tremor/54ECD65C96E1201DEABD3DDB5B656605
{
	{
		map $lightmap
		rgbGen identity
	}

	{
		// disabled stage
	}

	{
		map textures/tremor/grate3.tga
		rgbGen identity
	}
}
```

But the loop doing an early break on the disabled stage, it turned the material into this:

```
tremor/54ECD65C96E1201DEABD3DDB5B656605
{
	{
		map $lightmap
		rgbGen identity
	}
}
```

Only painting the light with no texture blended over.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5647,17 +5647,13 @@ static shader_t *FinishShader()
 		}
 	}
 
-	int stage;
+	numStages = MAX_SHADER_STAGES;
+	GroupActiveStages();
 
 	// set appropriate stage information
-	for ( stage = 0; stage < MAX_SHADER_STAGES; stage++ )
+	for ( size_t stage = 0; stage < numStages; stage++ )
 	{
 		shaderStage_t *pStage = &stages[ stage ];
-
-		if ( !pStage->active )
-		{
-			break;
-		}
 
 		if ( !shader.isSky )
 		{
@@ -5863,7 +5859,7 @@ static shader_t *FinishShader()
 		}
 	}
 
-	numStages = stage;
+	GroupActiveStages();
 
 	// there are times when you will need to manually apply a sort to
 	// opaque alpha tested shaders that have later blend passes


illwieckz:
Once fixed, not only the engine properly removes the unsupported light style stage as expected while keeping the texture stage, producing something like this:

```
tremor/54ECD65C96E1201DEABD3DDB5B656605
{
	{
		map $lightmap
		rgbGen identity
	}
	
	{
		map textures/tremor/grate3.tga
		rgbGen identity
	}
}
```

But the collapsing code called after that even collapses the material into this:

```
tremor/54ECD65C96E1201DEABD3DDB5B656605
{
	{
		diffuseMap textures/tremor/grate3.tga
	}
}
```

DolceTriade:
```
C:\projects\daemon\src\engine\renderer\tr_shader.cpp(5654,29): error C2220: the following warning is treated as an error [C:\projects\daemon\build\client-objects.vcxproj]
C:\projects\daemon\src\engine\renderer\tr_shader.cpp(5654,29): warning C4018: '<': signed/unsigned mismatch [C:\projects\daemon\build\client-objects.vcxproj]
```

slipher:
LGTM I guess.

It kind of sucks that we are doing GroupActiveStages() as many as 5 times. The stage struct is over 1k in size so it would be better not to copy it so many times.

illwieckz:
Yeah but doing it that way makes such very convenient, and that time is only spent at load time… This is a very good trade-of given the huge simplification of code that provides. Maybe in the future we would be able to migrate to some more clever struct, but this is fine for now.

PR #1190, opened 2024-06-15
q_shared: delete unused IS_NAN

illwieckz:
Delete unused `IS_NAN`.

One should prefer using `!Math::IsFinite` instead.


slipher:
LGTM

PR #1191, opened 2024-06-16
Don't CRC all the .crn files in release mode

slipher:
Companion: https://github.com/DaemonEngine/crunch/pull/45

This should decrease the load time of Unvanquished by at least 5%.


DolceTriade:
It also shaved ~500ms from my map load time.

DolceTriade (approved):

PR #1193, opened 2024-06-16
Add #line directive for glsl #insert

VReaperV:
Shader code added with #insert will use lines 10000, 10001, 10002; 20000, 20001 etc, each inserted shader adding 10000, as suggested by @illwieckz in https://github.com/DaemonEngine/Daemon/pull/1124#issuecomment-2122679583.


src/engine/renderer/gl_shader.cpp
```diff
@@ -781,11 +781,14 @@ std::string     GLShaderManager::BuildGPUShaderText( Str::StringRef mainShaderNa
 	std::string shaderMain;
 
 	std::string line;
+	uint insertCount = 0;
+	uint mainLine = 0;
```
slipher, line 785: mainLine is never read

VReaperV: Right, fixed now.

src/engine/renderer/gl_shader.cpp
```diff
@@ -795,15 +798,21 @@ std::string     GLShaderManager::BuildGPUShaderText( Str::StringRef mainShaderNa
...
+		shaderMain += "#line " + std::to_string( ( insertCount + 1 ) * 10000 ) + "\n";
+		insertCount++;
+
 		std::string shaderInsertPath = line.substr( position + 8, std::string::npos );
 		if ( shaderType == GL_VERTEX_SHADER ) {
 			shaderMain += GetShaderText( "glsl/" + shaderInsertPath + ".glsl" );
 		} else {
 			shaderMain += GetShaderText( "glsl/" + shaderInsertPath + ".glsl" );
 		}
+		shaderMain += "#line " + std::to_string( mainLine ) + "\n";
```
slipher, line 815: It appears that mainLine is never incremented when there is a line with an insert directive. mainLine should be incremented on every iteration of the outer loop to keep the line numbers in sync with the source.

VReaperV:
I changed a bit how this works because #line 0 before #insert was making the line count wrong.

Now instead #line -1 is treated as a marker, where `PrintShaderSource()` will keep the correct line count with it.

src/engine/renderer/gl_shader.cpp
VReaperV: This now works properly, eg:

```
// line 10
#line -1
#line 10000
// .. a bunch of code here
#line -1
// This line will have line number 11
```

slipher:
I tested and the line numbering for the main file was off. After using 3 `#insert`s, the line numbers were too high by 3. For example the real line 48 was labeled as 51.

Just my personal taste and not a required change, but I would rather not have lines saying `#LINE_MARKER_END` and `#LINE_MARKER_START`. Showing the filename there could be useful though.

slipher:
> > Showing the filename there could be useful though.
> 
> I like that idea. It would be useful to have these for vertex/fragment libs too.
Hopefully we can nuke those and replace them with `#insert`

VReaperV:
> Hopefully we can nuke those and replace them with `#insert`

I thought of that too, should be pretty easily doable. It would also allow dropping more of the forced `vp`/`fp`/`cp` suffixes.

VReaperV:
> I tested and the line numbering for the main file was off. After using 3 `#insert`s, the line numbers were too high by 3. For example the real line 48 was labeled as 51.

Fixed.



> Just my personal taste and not a required change, but I would rather not have lines saying `#LINE_MARKER_END` and `#LINE_MARKER_START`. Showing the filename there could be useful though.

This now shows the file name/path.

illwieckz (approved):
If it works for you, LGTM.

slipher:
LGTM

PR #1194, opened 2024-06-17
TEST ci/appveyor: skip branch with pr

illwieckz:
Test if `skip_branch_with_pr: true` appveyor option also skip branch with PR targetting a branch with PR (we don't want to).


illwieckz:
Test was OK

Issue #507, opened 2021-07-12
CMAKE Release build defaults to -O3 but RelWithDebInfo build defaults to -O2


illwieckz:
It looks like CMAKE Release build defaults to -O3 but RelWithDebInfo build defaults to -O2:

```
CMakeCache.txt
79:CMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING=-O2 -g -DNDEBUG
103:CMAKE_C_FLAGS_RELWITHDEBINFO:STRING=-O2 -g -DNDEBUG
```

```
CMakeCache.txt
76:CMAKE_CXX_FLAGS_RELEASE:STRING=-O3 -DNDEBUG
100:CMAKE_C_FLAGS_RELEASE:STRING=-O3 -DNDEBUG
```

When I'm working on Unvanquished/Daemon, I either do a Debug build to run on a debugger, or do a Release build to test for performance, meaning I never test the optimization level of our release builds,. On the other hand, that means I'm testing O3 since years without faciing specific issues.

There are maybe strong reason to use O2 when producing debug symbols, but on the other hand, maybe we would want to default to O3 in our released builds?

Note: doing O3 release builds may make Debian unhappy: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=640214

But I don't care that much as we're not integrated in their repo, and they are free to optimize less if they want to.

My main concern would be: how much O3 is problematic for debugging?

slipher:
These default flag values come from CMake, not from us. It seems like people often question whether -O3 is actually better than -O2, so I don't see a strong reason to override it.

illwieckz:
Yes those are CMake default, though maybe we may investigate the pros/cons to distribute `-O3` release builds. Most of the numbers from the [GPU compatibility matrix](https://wiki.unvanquished.net/wiki/GPU_compatibility_matrix) are then done with `-O3` because most of them were done with my pre 0.52 release local builds using `Release` CMake profile and then, `-O3`. I never use `RelWithDebInfo` CMake profile outside of release builds…

PR #1197, opened 2024-06-20
cmake: unify GCC, Clang and PNaCl compiler configuration

illwieckz:
- unify GCC, Clang and PNaCl compiler configuration
- pass more build options to the sub-cmake game build
- always set C standard, it will be required for building zlib with Saigo or Wasi
- add options to not enforce C/C++ standards
- move the target specific build options to another if/else block


illwieckz:
Now PNaCl uses the same configuration as GCC and Clang, avoids discrepancies like this to be introduced:

- https://github.com/DaemonEngine/Daemon/pull/1160

This also make possible to pass build options like LTO to the sub-cmake game call. I'm not sure it really does something with PNaCl as the produced files are the same size (and LTO usually makes our binaries much smaller), but the compiler accepts the option so, it's not bad to pass it. It will be ready for other compilers like Saigo or Wasi.

The C standard is now always enforced, it was required for NaCl because NaCl uses some in-tree zlib source. It makes possible to use an in-tree zlib for things other than PNaCl, something we would want for Saigo or Wasi. And if we ship an own zlib, there is no reason to not build the engine against it. The usage or the in-tree zlib for every binary is not part of that PR, but this PR makes the CMake code ready for it.

Also, for debugging purpose, one can disable the C/C++ standard enforcement with advanced `USE_RECOMMENDED_C_STANDARD` and `USE_RECOMMENDED_CXX_STANDARD` variables.

illwieckz:
I forgot to mention it, but this also makes explicit the optimization level used for the selected build profile. It was done for MSVC but not others that relied on CMake's default. In the future I plan to make `RelWithDebInfo` use `-O3` so everything will be there to just flip the number. For releases we currently use `RelWithDebInfo` with default `-O2`:

- https://github.com/DaemonEngine/Daemon/issues/507

illwieckz:
> I forgot to mention it, but this also makes explicit the optimization level used for the selected build profile.

Well, this is probably a bad idea, even if someday we override them.

The reason why I went this way was because the NaCl code path as doing it, so when I unified it with other compilers, I thought “why not?”. It happens the NaCl code path doing it is a workaround for CMake failing to detect that PNaCl is a clang-compatible compiler. We better want to rely on extensive compiler detection as implemented in #982 to set the fallbacks for every clang-compatible compiler we detect:

- https://github.com/DaemonEngine/Daemon/pull/982

So for now I'll revert the unconditional setting of compiler optimizations and keep the `if (NACL)` workaround until we can rely on something more generic thanks to #982.

slipher:
Why should there be an option not to use the appropriate language standard? I don't get it. Maybe it was useful one time, but not everything needs to be a checkbox option, if someone needs to do something really obscure they can edit the build system.

illwieckz:
> Why should there be an option not to use the appropriate language standard?

Generic answer: _I like knobs._

Specific answers:

- It makes easier for testing other standards without the default ones getting in the way, so it could be useful as a debug tool the day we raise the standards.
- It also makes easy to test the build without those flags, I remember that when I toyed with Android at first it was failing to build with the flags but the without it built. Finding the root cause of why it didn't built with the flag was a specific task that was a different task than testing the rest of the code.
- Not all the code base require those standards. For example C++14 isn't required to build a working game server, C++11 is enough.
- I really like to be able to disable stuff for testing and debugging purpose in general, and I had to comment out those lines at least three times in the recent years for testing purpose, and editing code breaks rebasing and other things.

cmake/DaemonFlags.cmake
```diff
@@ -201,256 +184,350 @@ if (MSVC)
...
+		#   || _ISOC99_SOURCE || _BSD_SOURCE
+		# fseeko, ftello:
+		#   _POSIX_C_SOURCE >= 200112L
+		# strdup:
+		#   _POSIX_C_SOURCE >= 200809L || _XOPEN_SOURCE >= 500
+		#   || (_XOPEN_SOURCE && _XOPEN_SOURCE_EXTENDED)
+		#   || _BSD_SOURCE || _SVID_SOURCE || _GNU_SOURCE
+		# isascii:
+		#   _XOPEN_SOURCE || _DEFAULT_SOURCE || _SVID_SOURCE
+		add_definitions(-D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=500)
```
slipher, line 257: What platform is this stuff needed on?

cmake/DaemonFlags.cmake
```diff
@@ -201,256 +184,350 @@ if (MSVC)
...
+		set_c_cxx_flag("-D_FORTIFY_SOURCE=2" RELEASE)
+		set_c_cxx_flag("-D_FORTIFY_SOURCE=2" RELWITHDEBINFO)
+		set_c_cxx_flag("-D_FORTIFY_SOURCE=2" MINSIZEREL)
+		# Don't set _FORTIFY_SOURCE in debug builds.
+	endif()
+
+	if (USE_HARDENING)
+		message(STATUS "Enabling hardening")
+
+		# PNaCl accepts the flags bug do not define __stack_chk_guard and __stack_chk_fail.
```
slipher, line 370: Typos, `but does not`

cmake/DaemonFlags.cmake
```diff
@@ -201,256 +184,350 @@ if (MSVC)
...
+				set_linker_flag("-Wl,-e,mainCRTStartup")
+			else()
+				message(FATAL_ERROR "Unsupported architecture ${ARCH}")
+			endif()
+		endif()
+	endif()
+
+	# Link-time optimizations.
+	if (USE_LTO)
+		message(STATUS "Enabling link time optimizations")
```
slipher, line 398: The combination of adding these verbose messages, and passing unimplemented-in-NaCl build options such as USE_ARCH_INTRINSICS or USE_LTO to the NaCl build is unfortunate. It just causes the sub-build to lie with false messages such "Enabling link time optimizations"

cmake/DaemonFlags.cmake
```diff
@@ -201,256 +184,350 @@ if (MSVC)
...
+				message(FATAL_ERROR "GNU89 or C99 not supported by compiler")
+			endif()
+		endif()
+	endif()
+
+	option(USE_RECOMMENDED_CXX_STANDARD "Use recommended C++ standard" ON)
+	mark_as_advanced(USE_RECOMMENDED_CXX_STANDARD)
+
+	if (USE_RECOMMENDED_CXX_STANDARD)
+		# PNaCl only defines isascii if __STRICT_ANSI__
```
slipher, line 225: Who needs "isascii"? I don't find any references to this in Daemon or Unvanquished `src` or `libs`.

slipher:
Having a knob that makes the program fail to build kinda sucks. But on the other hand, it sucks that there is often no way to disable all the flags added by our build system, like you could with something like `CXXGLAGS=''` in a simpler build system. So I can see the utility of having a way to knock out some flags. :+1:

cmake/DaemonFlags.cmake
illwieckz: With `-std=c++14` instead of `-std=gnu++14`.

cmake/DaemonFlags.cmake
illwieckz: It's in the libc++ provided by PNaCl:

```
In file included from Unvanquished/daemon/src/common/Common.h:35:
In file included from Unvanquished/daemon/src/engine/qcommon/q_shared.h:112:
In file included from build/deps/linux-amd64-default_10/pnacl/le32-nacl/include/c++/v1/random:1645:
In file included from build/deps/linux-amd64-default_10/pnacl/le32-nacl/include/c++/v1/istream:163:
In file included from build/deps/linux-amd64-default_10/pnacl/le32-nacl/include/c++/v1/ostream:138:
In file included from build/deps/linux-amd64-default_10/pnacl/le32-nacl/include/c++/v1/ios:216:
build/deps/linux-amd64-default_10/pnacl/le32-nacl/include/c++/v1/__locale:553:16: error: 
      use of undeclared identifier 'isascii'
        return isascii(__c) ? (__tab_[static_cast<int>(__c)] & __m) !=0 : false;
               ^
build/deps/linux-amd64-default_10/pnacl/le32-nacl/include/c++/v1/__locale:577:19: error: 
      use of undeclared identifier 'isascii'
            if (!(isascii(*__low) && (__tab_[static_cast<int>(*__low)] & __m)))
                  ^
build/deps/linux-amd64-default_10/pnacl/le32-nacl/include/c++/v1/__locale:560:22: error: 
      use of undeclared identifier 'isascii'
            *__vec = isascii(*__low) ? __tab_[static_cast<int>(*__low)] : 0;
                     ^
build/deps/linux-amd64-default_10/pnacl/le32-nacl/include/c++/v1/__locale:568:17: error: 
      use of undeclared identifier 'isascii'
            if (isascii(*__low) && (__tab_[static_cast<int>(*__low)] & __m))
                ^
build/deps/linux-amd64-default_10/pnacl/le32-nacl/include/c++/v1/__locale:577:19: error: 
      use of undeclared identifier 'isascii'
            if (!(isascii(*__low) && (__tab_[static_cast<int>(*__low)] & __m)))
                  ^
```

cmake/DaemonFlags.cmake
illwieckz: I removed the “Enabling link time optimizations” message, it was not saying the feature was enabled but the user asked for it so it was wrong anyway.
 
 Passing the `USE_LTO` option doesn't sound problematic. If the compiler lies about supporting LTO, it's not our fault. I also turned the `flto` setting as a “try”, so even setting `USE_LTO` on compilers that would reject the flag wouldn't break the build, and CMake will report the status. PNaCl will lie but that's not our problem.

The `USE_ARCH_INTRINSICS` would make sense the day we use Wasm, as Wasm support some kind of SIMD. Right now it's a bit useless for PNaCL but PNaCl not doing anything with it doesn't mean it's wrong to pass the option, in case the toolchain can do something with it.

illwieckz:
I moved the setting of PNaCl default build flag to the PNaCl toolchain itself. It makes much sense like that. I haven't set the `-O0` for debug as it is not what CMake does for other clang-based compilers:

- https://github.com/Unvanquished/Unvanquished/pull/3043

cmake/DaemonFlags.cmake
slipher: I would understand -DUSE_RECOMMENDED_C_STANDARD=0 as telling the build system to get out of the way and not set *any* C standard flags. I oppose having it mean keep setting language standard flags, but it a different way.

cmake/DaemonFlags.cmake

cmake/DaemonFlags.cmake
illwieckz: The code now only sets GNU dialects.

cmake/DaemonFlags.cmake
illwieckz: USE_LTO is not passed to NACL sub-build, also the option is ignored if for some reason the option is passed (like doing some manual stuff). After investigation, it happens that neither PNaCl underlying clang nor Saigo clang support `-flto`. It happens that the `pnacl-clang` wrapper accepts the option but I don't see how it would do something with it, and the produced binaries are the sames anyway.

illwieckz:
This now looks ready to me.

slipher:
LGTM

PR #1203, opened 2024-06-22
tr_shader: factorize stage checking and move the code to a dedicated function

illwieckz:
Yes another step in my effort to make that shader code less messy and easier to follow, here moving more things out of the lengthy `FinishShader` function and factorizing the code more to avoid copy-paste errors when modifying it.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5636,6 +5636,84 @@ bool CheckShaderNameLength( const char* func_err, const char* name, const char* 
 	return true;
 }
 
+static void ValidateStage( shaderStage_t *pStage )
+{
+	struct stageCheck_t {
+		bool expected;
+		bool check;
+		bool fallback;
+		const std::string name;
+	};
+
+	static const std::unordered_map<stageType_t, stageCheck_t> stageTypeCheck = {
+		{ stageType_t::ST_COLORMAP, { true, true, true, "color map" } },
+		{ stageType_t::ST_GLOWMAP, { true, true, false, "glowMap" } },
+		{ stageType_t::ST_DIFFUSEMAP, { true, true, true, "diffuseMap" } },
+		{ stageType_t::ST_NORMALMAP, { true, true, false, "normalMap" } },
+		{ stageType_t::ST_HEIGHTMAP, { true, true, false, "heightMap" } },
+		// There is no standalone physicalMap stage.
+		{ stageType_t::ST_PHYSICALMAP, { false, false, false, "physicalMap" } },
+		{ stageType_t::ST_SPECULARMAP, { true, true, false, "specularMap" } },
+		{ stageType_t::ST_HEATHAZEMAP, { true, true, false, "heatHazeMap" } },
+		{ stageType_t::ST_LIQUIDMAP, { true, true, false, "heatHazeMap" } },
+		// The lightmap is fetched at render time.
+		{ stageType_t::ST_LIGHTMAP, { true, false, false, "light map" } },
+		// The lightmap is fetched at render time.
+		{ stageType_t::ST_STYLELIGHTMAP, { true, false, false, "style light map" } },
+		{ stageType_t::ST_STYLECOLORMAP, { true, true, false, "style color map" } },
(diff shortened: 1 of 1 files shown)


src/engine/renderer/tr_shader.cpp
```diff
@@ -5572,6 +5572,88 @@ bool CheckShaderNameLength( const char* func_err, const char* name, const char*
...
+static bool CheckStage( shaderStage_t *pStage )
+{
+	struct stageCheck_t {
+		bool expected;
+		bool check;
+		bool fallback;
+		std::string name;
+	};
+
+	static const std::unordered_map<stageType_t, stageCheck_t> stageTypeCheck = {
```
slipher, line 5584: The table of boolean flags seems kinda overkill. Why not something like
```
switch( stageType ) {
  case stageType_t::ST_GLOWMAP:
  case stageType_t::ST_DIFFUSEMAP:
  case stageType_t::ST_COLORMAP:
    if (!pStage->bundle[0].image[0]) {
     Log::Warn("%s stage missing image", stageTypeNames[stageType]);
```

That way you can deduplicate the code but you only have to put the stage names in a table.

illwieckz: I prefer the table and it's on purpose: I want that when a line is added for a stage type to be documented in the table, the line cannot be added without documenting every bits about that stage type. It's written this way on purpose of relying on the C++ strong typing to yell at the developer “you haven't told me everything about it!”…

src/engine/renderer/tr_shader.cpp
```diff
@@ -5572,6 +5572,88 @@ bool CheckShaderNameLength( const char* func_err, const char* name, const char*
 	return true;
 }
 
+static bool CheckStage( shaderStage_t *pStage )
```
slipher, line 5575: This could take a const stage. Caller should mark it not active

src/engine/renderer/tr_shader.cpp
```diff
@@ -5572,6 +5572,88 @@ bool CheckShaderNameLength( const char* func_err, const char* name, const char*
...
+{
+	struct stageCheck_t {
+		bool expected;
+		bool check;
+		bool fallback;
+		std::string name;
```
slipher, line 5581: Could be a `const char *` (to avoid allocations)

src/engine/renderer/tr_shader.cpp
```diff
@@ -5572,6 +5572,88 @@ bool CheckShaderNameLength( const char* func_err, const char* name, const char*
...
+		{ stageType_t::ST_STYLELIGHTMAP, { true, false, false, "style light map" } },
+		{ stageType_t::ST_STYLECOLORMAP, { true, true, false, "style color map" } },
+		{ stageType_t::ST_COLLAPSE_COLORMAP, { true, true, true, "collapsed color map" } },
+		{ stageType_t::ST_COLLAPSE_DIFFUSEMAP, { true, true, true, "collapsed diffuseMap" } },
+		{ stageType_t::ST_ATTENUATIONMAP_XY, { true, true, false, "attenuationMapXY" } },
+		{ stageType_t::ST_ATTENUATIONMAP_Z, { true, true, false, "attenuationMapZ" } },
+		// TODO: Document the remaining stage types.
+	};
+
+	if ( stageTypeCheck.count( pStage->type ) )
```
slipher, line 5607: How to use a map:
```
auto it = stageTypeCheck.find(pStage->type);
if (it != stageTypeCheck.end()) {
   stageCheck_t stagecheck = it->second;
```

src/engine/renderer/tr_shader.cpp
illwieckz: I can make the caller to mark it inactive but this is remaining:

```
tr_shader.cpp:5624:39: error: cannot assign to variable 'pStage'
with const-qualified type 'const shaderStage_t *'
 5624 |                             pStage->bundle[ 0 ].image[ 0 ]  = tr.defaultImage;
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ^
```

src/engine/renderer/tr_shader.cpp

src/engine/renderer/tr_shader.cpp

slipher (approved):
LGTM

PR #1204, opened 2024-06-22
Material System: Make depth pre-pass surfaces into stages of the main surface

VReaperV:
Builds on #1180

Avoid culling the same surface twice, which is especially important for occlusion culling since it reduces the amount of texture accesses by half. Also changed material generation a bit to make it easier to add non-world surfaces later.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5901,18 +5901,6 @@ static shader_t *FinishShader()
 	// Copy the current global shader to a newly allocated shader.
 	shader_t *ret = MakeShaderPermanent();
 
-	if ( glConfig2.materialSystemAvailable && !tr.worldLoaded ) {
-		uint8_t maxStages = 0;
-		for ( shaderStage_t* pStage = ret->stages; pStage < ret->lastStage; pStage++ ) {
-			maxStages++;
-		}
-
-		if ( maxStages % 4 != 0 ) { // Aligned to 4 components
-			maxStages = ( maxStages / 4 + 1 ) * 4;
-		}
-		materialSystem.maxStages = maxStages > materialSystem.maxStages ? maxStages : materialSystem.maxStages;
-	}
-
 	// generate depth-only shader if necessary
 	if( !shader.isSky &&
 	    numStages > 0 &&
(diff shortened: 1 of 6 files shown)


VReaperV:
Rebased.

VReaperV:
Actually, it might be better to make this a distinct stage for the renderer both with and without material system, since that would allow using a vertex-only depth shader, unless there's alpha-testing required.

slipher:
> Actually, it might be better to make this a distinct stage for the renderer both with and without material system, since that would allow using a vertex-only depth shader, unless there's alpha-testing required.

Should we wait to review this then?

VReaperV:
> > Actually, it might be better to make this a distinct stage for the renderer both with and without material system, since that would allow using a vertex-only depth shader, unless there's alpha-testing required.
> 
> Should we wait to review this then?

Yeah, I'll convert this to draft for now.

VReaperV:
> Should we wait to review this then?

I've changed my mind. I tried making it a sperate stage and it ended up being convoluted and bug-prone without any real benefit over this pr. So I'm marking this pr ready again.

VReaperV:
Also fixed some uniforms not being set to global, which apparently never resulted in incorrect rendering before this pr.

src/engine/renderer/tr_shader.cpp
```diff
@@ -5979,6 +5980,18 @@ static shader_t *FinishShader()
 		ret->depthShader = nullptr;
 	}
 
+	if ( glConfig2.materialSystemAvailable && !tr.worldLoaded ) {
+		uint8_t maxStages = 0;
+		for ( shaderStage_t* pStage = ret->stages; pStage < ret->lastStage; pStage++ ) {
+			maxStages++;
+		}
+
+		if ( maxStages % 4 != 0 ) { // Aligned to 4 components
```
slipher, line 5989: You could use the PAD macro

src/engine/renderer/tr_shader.cpp
```diff
@@ -5979,6 +5980,18 @@ static shader_t *FinishShader()
...
+			maxStages = ( maxStages / 4 + 1 ) * 4;
+		}
+		materialSystem.maxStages = maxStages > materialSystem.maxStages ? maxStages : materialSystem.maxStages;
```
slipher, line 5992: `std::max`

slipher (commented):
What's the advantage of doing this?

src/engine/renderer/tr_shader.cpp
VReaperV: Interesting, didn't know we had that macro. Done now.

src/engine/renderer/tr_shader.cpp
VReaperV: Sure.

VReaperV:
> What's the advantage of doing this?

Only ~half as many invocations of `cull_cp` would run on average, less memory usage/better caching, less texture accesses.

slipher:
LGTM

PR #1205, opened 2024-06-22
Only do RB_RenderPostDepthLightTile() for the main view

VReaperV:
This will ensure that only the main view computes dynamic lights. Fixes dynamic light disappearing near portals.

This does mean that dynamic lights won't render in portals, but that was already the case anyway, and fixing that would require more complex changes, which would be unlikely to be noticeable.


illwieckz (approved):
LGTM.

PR #1206, opened 2024-06-23
renderer: port GL extension cvars to new style format

illwieckz:
Port GL extension cvars to new style format.

@slipher I have no idea is the `Cvar::Unlatch` implementation is correct.

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -87,23 +87,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_noMarksOnTrisurfs;
 	cvar_t      *r_lazyShaders;
 
-	cvar_t      *r_ext_occlusion_query;
-	cvar_t      *r_ext_draw_buffers;
-	cvar_t      *r_ext_half_float_pixel;
-	cvar_t      *r_ext_texture_float;
-	cvar_t      *r_ext_texture_integer;
-	cvar_t      *r_ext_texture_rg;
-	cvar_t      *r_ext_texture_filter_anisotropic;
-	cvar_t      *r_ext_gpu_shader4;
-	cvar_t      *r_arb_buffer_storage;
-	cvar_t      *r_arb_map_buffer_range;
-	cvar_t      *r_arb_sync;
-	cvar_t      *r_arb_uniform_buffer_object;
-	cvar_t      *r_arb_texture_gather;
-	cvar_t      *r_arb_gpu_shader5;
-	Cvar::Cvar<bool> r_arb_depth_clamp("r_arb_depth_clamp", "Use GL_ARB_depth_clamp if available", Cvar::NONE, true);
-	cvar_t      *r_arb_compute_shader;
-
 	cvar_t      *r_checkGLErrors;
 	cvar_t      *r_logFile;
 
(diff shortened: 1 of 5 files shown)


src/engine/renderer/tr_init.cpp
```diff
@@ -87,22 +87,23 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
-	cvar_t      *r_ext_gpu_shader4;
-	cvar_t      *r_arb_buffer_storage;
-	cvar_t      *r_arb_map_buffer_range;
-	cvar_t      *r_arb_sync;
-	cvar_t      *r_arb_uniform_buffer_object;
-	cvar_t      *r_arb_texture_gather;
-	cvar_t      *r_arb_gpu_shader5;
-	Cvar::Cvar<bool> r_arb_depth_clamp("r_arb_depth_clamp", "Use GL_ARB_depth_clamp if available", Cvar::NONE, true);
-	cvar_t      *r_arb_compute_shader;
+	Cvar::Cvar<bool> r_ext_occlusion_query( "r_ext_occlusion_query", "Use GL_EXT_occlusion_query if available", Cvar::NONE, true );
```
slipher, line 90: It might be nice to make these all `static` in sdl_glimp.cpp since using them elsewhere is basically always a mistake.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1864,10 +1865,21 @@ static void GLimp_InitExtensions()
...
-	if ( LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, EXT_texture_filter_anisotropic, r_ext_texture_filter_anisotropic->value ) )
+	if ( LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, EXT_texture_filter_anisotropic, r_ext_texture_filter_anisotropic.Get() > 0 ) )
 	{
 		glGetFloatv( GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT, &glConfig2.maxTextureAnisotropy );
 		glConfig2.textureAnisotropyAvailable = true;
+
+		// Bound texture anisotropy.
+		if ( r_ext_texture_filter_anisotropic.Get() > glConfig2.maxTextureAnisotropy )
+		{
+			// HACK: We want to set the current value, not the latched value
```
slipher, line 1876: I don't think we should be setting the value anyway (I see that the existing code does this). If the driver lacks an extension, we don't set the cvar to 0 but rather record which value should be used in glConfig(2). Analogously, we could store the effective max anisotropy there instead of the cvar.

illwieckz: Right, we don't need to se the cvar, it was a hack for reading the value in `tr_image`, we could just use some glconfig `int` instead.

but we may still have to implement the `Cvar::Unlatch` function the day we port that other code:

```c++
// HACK: We want to set the current value, not the latched value
Cvar::ClearFlags("r_customwidth", CVAR_LATCH);
Cvar::ClearFlags("r_customheight", CVAR_LATCH);
Cvar_Set( "r_customwidth", va("%d", glConfig.vidWidth ) );
Cvar_Set( "r_customheight", va("%d", glConfig.vidHeight ) );
Cvar::AddFlags("r_customwidth", CVAR_LATCH);
Cvar::AddFlags("r_customheight", CVAR_LATCH);
```

src/engine/renderer/tr_init.cpp

src/engine/sys/sdl_glimp.cpp
illwieckz: Done.

slipher:
LGTM

PR #1208, opened 2024-06-24
tr_shader: colormap stage with 2D image bit is generic2D

illwieckz:
Color map stage with 2D image bit is generic2D.

We can skip the render-time test for them.

Also make a variable local to the block using it.


illwieckz:
Because `RSF_2D` is not in `master`, it was added in:

- https://github.com/DaemonEngine/Daemon/pull/1145

Providing `RSF_FITSCREEN` and `RSF_2D` broke compatibility with game.

slipher:
Is that really a good way to decide whether we are doing 2d rendering? What if the same image were used for both something in the UI and something in the map and it got registered with the 2D flag. The `if ( backEnd.projection2D )` that we have now seems more reliable

illwieckz:
That's something that crossed my mind, but images with `RSF_2D` flags are created by the game for its own UI. There are even no legacy image having `RSF_2D` flag to take care about.

Those images are usually using 2D-specific options like disabling mipmapping and preventing the downscale of them like with `noPicMip` or `fitScreen`. Using them on a 3D surface may work but would be hacky and it would be recommended to create a special shader for them.

Also those images using the `RSF_2D` flag usually live in folder like `ui/` which are not usable at all on map surfaces: only images in `textures/` can be applied on map surface. Well it's possible to edit a bsp to set non-textures-prefixed images, but the map source format make it impossible to do in a map editor or even by hand in the map format itself..

Those 2D images may still be usable by the engine anyway, like on a model, or in a particle system. For example I know some maps have implemented fireworks using a granger emoticon as particle. As far as I know we don't set `RSF_2D` on them. It would be then considered wrong to set `RSF_2D` on them.

If one day we implement UI on some 3D surfaces (like some DOOM3 in-game screens), we would either duplicate the shader, either remove the `RSF_2D` flag.

The purpose of the `RSF_2D` flag is to tell the engine it's an UI image. In fact we may even rename it to `RSF_UI` to avoid confusion with `SHADER_2D` (which I have no idea what it is about, but is unrelated to the 2D UI images).

slipher:
Fine, you can say the use cases for that are doing it wrong, but the benefits from this seem microscopically small. So I disapprove of the proposed change.

VReaperV:
I suppose this isn't really relevant anymore, since `Render_generic2D()` does not exist now.

PR #1211, opened 2024-06-28
Fix bad (v)snprintf on MinGW

slipher:
MinGW was using _vsnprintf and _snprintf, which are broken in all versions of the Windows standard library (even ones that have a fixed snprintf and vsprintf).


illwieckz (approved):
LGTM.

PR #1212, opened 2024-06-30
Revert "Only do RB_RenderPostDepthLightTile() for the main view"

slipher:
This reverts commit c3d538d0990dd45b5a5b688ea7c3b38d1b515cce.

It breaks dynamic lighting. The light tile rectangles are visible.

unvanquished_2024-06-30_152600_000

PR #1213, opened 2024-06-30
Fix post-deph lighttile

VReaperV:
Do `RenderPostDepthTile()` once after depth pre-pass of the first view that doesn't have any nested views. Fixes the issue in #1212.


DolceTriade:
I can't comment on the blockiness, but without this (or the revert) I see a significant increase in dynamic light intensity and both this and the revert both return it to the same level.

I suppose LGTM?

DolceTriade:
Do you think it's odd that all three of us have 3 kinda separate issues?

slipher:
Works for me, LGTM.

> Do you think it's odd that all three of us have 3 kinda separate issues?

Not really, I suppose OpenGL buffers can have undefined contents.

VReaperV:
> Not really, I suppose OpenGL buffers can have undefined contents.

Yes, `glBufferData()` with a nullptr makes the contents of the buffer undefined.

PR #1217, opened 2024-07-06
Fix generic quad verts being added to other surfaces

### LINKED ISSUE #1196: 'Generic quad' is mistakenly added into other draw calls
I have found that the `generic_VBO` vertices are sometimes rendered with other shaders than the ones it is intended to be used with.
Out of curiosity, I checked to see if this would have any visible consequences. I tested with the following patch:

```
diff --git a/src/engine/renderer/tr_backend.cpp b/src/engine/renderer/tr_backend.cpp
index 92ede239f..0d774768b 100644
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -720,6 +720,11 @@ void GL_VertexAttribPointers( uint32_t attribBits )
 				frame = glState.vertexAttribsOldFrame;
 			}
 
+			if ( !(glState.currentVBO->attribBits & bit ) )
+			{
+				Log::Notice("%s lacking in %s", attributeNames[ i ], glState.currentVBO->name);
+			}
+
 			glVertexAttribPointer( i, layout->numComponents, layout->componentType, layout->normalize, layout->stride, BUFFER_OFFSET( layout->ofs + ( frame * layout->frameOffset + base ) ) );
 			glState.vertexAttribPointersSet |= bit;
 		}
diff --git a/src/engine/renderer/tr_vbo.cpp b/src/engine/renderer/tr_vbo.cpp
index accdc89d1..4cfeef779 100644
--- a/src/engine/renderer/tr_vbo.cpp
+++ b/src/engine/renderer/tr_vbo.cpp
@@ -906,9 +906,12 @@ static void R_InitGenericVBOs() {
 	VectorCopy( v3, verts[3].xyz );
 	
 	for ( int i = 0; i < 4; i++ ) {
-		verts[i].color = Color::White;
+		verts[i].color = Color::Magenta;
 		verts[i].texCoords[0] = floatToHalf( i < 2 ? 0.0f : 1.0f );
 		verts[i].texCoords[1] = floatToHalf( i > 0 && i < 3 ? 1.0f : 0.0f );
+
+		vec4_t orent{1, 0, 0, 20};
+		floatToHalf(orent, verts[i].spriteOrientation);
 	}
 	surface->vbo = R_CreateStaticVBO2( "generic_VBO", surface->numVerts, verts, ATTR_POSITION | ATTR_TEXCOORD | ATTR_COLOR );
 
```

- The color is changed to magenta to make it identifiable as this and not some other bug
- Something is set for qtangents/spriteOrientation to avoid randomness from uninitialized memory
- The log message shows whether the bug is active

Also I used the commands `/cg_motionblurMinSpeed -1; cg_flyspeed 30;  setviewpos 0.890782 -0.764073 1.764857 101.173088 57.002563`. Depending on the map, `r_drawWorld 0` or `noclip` may be needed. If you don't see the message `attr_QTangent lacking in generic_VBO`, the bug is not active and you may have to restart or change map a few times.

unvanquished_2024-06-19_160809_000

Voila, a little font sprite sheet is drawn at the origin. Sometimes I also got a base beacon or a solid rectangle.

I think I can fix this by adding back some of the cleanup code to Tess_InstantQuad which was removed when it was changed to use a VBO. Like `tess.multiDrawPrimitives = 0; tess.numVertexes = 0;` etc.

slipher:
Fixes #1196.

Also adds a check to detect requesting vertex attributes which are not provided by a VBO.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -714,6 +714,12 @@ void GL_VertexAttribPointers( uint32_t attribBits )
 				frame = glState.vertexAttribsOldFrame;
 			}
 
+			if ( !( glState.currentVBO->attribBits & bit ) )
+			{
+				Log::Warn( "GL_VertexAttribPointers: %s does not have %s",
+				           glState.currentVBO->name, attributeNames[ i ] );
+			}
+
 			glVertexAttribPointer( i, layout->numComponents, layout->componentType, layout->normalize, layout->stride, BUFFER_OFFSET( layout->ofs + ( frame * layout->frameOffset + base ) ) );
 			glState.vertexAttribPointersSet |= bit;
 		}
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -1879,11 +1879,7 @@ static void RB_SetupLightForLighting( trRefLight_t *light )
 						x = 205 * frustumIndex;
 						y = 70;
 
-						Tess_InstantQuad( x, y, w, h );
-
-						gl_debugShadowMapShader->SetUniform_ModelViewProjectionMatrix( glState.modelViewProjectionMatrix[glState.stackIndex] );
-						
-						Tess_DrawElements();
+						Tess_InstantQuad( *gl_debugShadowMapShader, x, y, w, h );
 
 						{
 							int    j;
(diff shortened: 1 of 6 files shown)


illwieckz (approved):
LGTM.

PR #1219, opened 2024-07-15
Fix GLSL #insert to not use negative #line numbers

slipher:


illwieckz:
Cherry-picking this fixes the bug I reproduce with:

illwieckz (approved):
LGTM.

PR #1220, opened 2024-07-15
Simplify some GLSL shader loading code

VReaperV:
Make somewhat messy code I noticed while working on #1105 and subsequent prs into something more readable.

Merge the two branches in `GLShaderManager::buildPermutation()` into one since they were doing essentially the same thing. Change some variable names to avoid confusion between shader and shader programs, fixed some formatting like whitespaces in [] and missing {} in if statements. Removed a redundant namespace in function call.

There is technically a slight difference in that compute shaders may be loaded even if `deformIndex` is > 0, but I don't think that really matters.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -897,8 +897,7 @@ static bool IsUnusedPermutation( const char *compileMacros )
 void GLShaderManager::buildPermutation( GLShader *shader, int macroIndex, int deformIndex )
 {
 	std::string compileMacros;
-	int  startTime = ri.Milliseconds();
-	int  endTime;
+	int startTime = ri.Milliseconds();
 	size_t i = macroIndex + ( deformIndex << shader->_compileMacros.size() );
 
 	// program already exists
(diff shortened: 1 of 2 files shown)


src/engine/renderer/gl_shader.cpp
```diff
@@ -912,37 +911,25 @@ void GLShaderManager::buildPermutation( GLShader *shader, int macroIndex, int de
 	{
 		shader->BuildShaderCompileMacros( compileMacros );
 
-		if ( IsUnusedPermutation( compileMacros.c_str() ) )
+		if ( IsUnusedPermutation( compileMacros.c_str() ) ) {
```
slipher, line 914: In these files the curly braces should go on a separate line. See https://wiki.unvanquished.net/wiki/Coding_convention#Quake_style

src/engine/renderer/gl_shader.cpp
```diff
@@ -1248,21 +1244,24 @@ void GLShaderManager::CompileGPUShaders( GLShader *shader, shaderProgram_t *prog
 	}
 }
 
-void GLShaderManager::CompileAndLinkGPUShaderProgram( GLShader *shader, shaderProgram_t *program,
+void GLShaderManager::CompileAndLinkGPUShaderProgram( GLShader *shader, shaderProgram_t *program, shaderProgram_t* baseProgram,
 						      Str::StringRef compileMacros, int deformIndex )
 {
-	GLShaderManager::CompileGPUShaders( shader, program, compileMacros );
+	if ( ( !baseProgram->VS && shader->_hasVertexShader ) || ( !baseProgram->FS && shader->_hasFragmentShader )
```
slipher, line 1250: It's kinda confusing that program and baseProgram can be the same. I'm not convinced that this refactoring is really simpler.

src/engine/renderer/gl_shader.cpp
VReaperV: Fixed.

src/engine/renderer/gl_shader.cpp
VReaperV: Alternatively, code from `CompileAndLinkGPUShaderProgram()` can be moved to `buildPermutation()` and `baseProgram` set to `program` explicitly for `deformIndex == 0`.

VReaperV:
This code was removed by #1586.

PR #1221, opened 2024-07-15
Minor fixes and improvements to ssao

VReaperV:
Change unprojectionParams to be a uniform, take map inverse light factor into account when setting the buffer clear colour for `r_ssao < 0`, minor formatting fixes. Removed unused variables and the preprocessor check for GLSL version > 1.20, since there's already a preceding check for `GL_ARB_texture_gather`, which requires version 1.30.

### DIFF
--- a/src/engine/renderer/glsl_source/ssao_fp.glsl
+++ b/src/engine/renderer/glsl_source/ssao_fp.glsl
@@ -23,7 +23,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 /* ssao_fp.glsl */
 
 uniform sampler2D u_DepthMap;
-IN(flat) vec3 unprojectionParams;
+uniform vec3 u_UnprojectionParams;
 
 DECLARE_OUTPUT(vec4)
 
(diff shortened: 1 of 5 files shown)


src/engine/renderer/glsl_source/ssao_fp.glsl
```diff
@@ -23,7 +23,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 /* ssao_fp.glsl */
 
 uniform sampler2D u_DepthMap;
-IN(flat) vec3 unprojectionParams;
```
slipher, line 26: I guess the code in ssao_vp that sets this should be deleted?

src/engine/renderer/tr_backend.cpp
```diff
@@ -3216,18 +3214,26 @@ void RB_RenderSSAO()
...
 		glClear( GL_COLOR_BUFFER_BIT );
 	}
 
 	gl_ssaoShader->BindProgram( 0 );
 
-	zParams[ 0 ] = 2.0f * tanf( DEG2RAD( backEnd.refdef.fov_x * 0.5f) ) / glConfig.vidWidth;
-	zParams[ 1 ] = 2.0f * tanf( DEG2RAD( backEnd.refdef.fov_y * 0.5f) ) / glConfig.vidHeight;
-	zParams[ 2 ] = backEnd.viewParms.zFar;
+	vec3_t zParams;
+	zParams[0] = 2.0f * tanf( DEG2RAD( backEnd.refdef.fov_x * 0.5f ) ) / glConfig.vidWidth;
```
slipher, line 3224: We do want the spaces inside square brackets. If the code has spaces insides parentheses it should have spaces inside brackets too

src/engine/renderer/glsl_source/ssao_fp.glsl
VReaperV: Yeah, must've forgoten to add it to the commit.

VReaperV: Fixed.

src/engine/renderer/tr_backend.cpp
VReaperV: Fixed.

slipher:
LGTM

PR #1222, opened 2024-07-15
Properly reset glState

VReaperV:
Fixes an issue that was preventing `r_ssao < 0` from working correctly when switching maps.

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -784,14 +784,8 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 	{
 		GLimp_LogComment( "--- GL_SetDefaultState ---\n" );
 
-		GL_ClearDepth( 1.0f );
-		GL_ClearStencil( 0 );
-
-		GL_FrontFace( GL_CCW );
 		glCullFace( GL_FRONT );
-
-		glState.faceCulling = CT_TWO_SIDED;
-		glDisable( GL_CULL_FACE );
+		GL_Cull( CT_TWO_SIDED );
 
 		GL_CheckErrors();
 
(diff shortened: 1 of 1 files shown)


DolceTriade:
As someone who doesn't understand this code, why is this proper vs the other way?

VReaperV:
The idea with `glState` is to avoid unnecessarily calling `gl*` functions if the values haven't changed, however if you call the `gl*` functions instead of the wrappers we have, then:
1. You might be calling the `gl*` functions again with the same values as the last call, which incurs an overhead. 
2. Whenever the wrappers are called the engine will compare the values that would be passed in with values in `glState`, which could now be incorrect.

DolceTriade:
LGTM % put the explanation above into the commit message.

VReaperV:
> When do `r_ssao` can be smaller than `0`?

Debug option, just shows where SSAO was applied?

slipher:
GL_SetDefaultState may be called right after the GL context is (re)created. At that point I believe `glState` should have been zeroed by ResetStruct. It's a little sketchy to use the wrapper functions at that point because the `glState` data they rely on for change detection is not accurate. However a bug would only occur if Daemon's default value for something were zero while OpenGL's default value for it were nonzero.

VReaperV:
`ResetStruct( glState );` seems to only be done in `GLimp_Shutdown()`.

slipher:
> `GLimp_Shutdown()`.

Yeah. If the renderer is restarted, for example by the `vid_restart` command, then GLimp_Shutdown will be called, and the window and context created anew.

VReaperV:
> GL_SetDefaultState may be called right after the GL context is (re)created. At that point I believe `glState` should have been zeroed by ResetStruct. It's a little sketchy to use the wrapper functions at that point because the `glState` data they rely on for change detection is not accurate. However a bug would only occur if Daemon's default value for something were zero while OpenGL's default value for it were nonzero.

I've checked with the default API values and they're either the same defaults as we use or we end up setting things correctly due to using enums rather than direct GL values.

slipher:
> I've checked with the default API values and they're either the same defaults as we use or we end up setting things correctly due to using enums rather than direct GL values.

To clarify, when saying "if Daemon's default value for something were zero while OpenGL's default value for it were nonzero", I was referring to whichever representation Daemon uses in glState for both sides of the equation. Yeah I don't believe there are any bugs currently, but the algorithm depends on luck to keep working since the state in `glState` is wrong at context initialization. This is a non-blocking comment.

VReaperV:
Yeah. Luckily adding more things to `glState` seems unlikely.

PR #1223, opened 2024-07-17
Fix imageFormatNameSize for remaining texture formats

VReaperV:
Fix some incorrect multipliers and add the computation for compressed formats.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -145,13 +145,19 @@ void R_ListImages_f()
 
 	typedef std::pair<const std::string, int> nameSizePair;
 	static const std::unordered_map<uint32_t, nameSizePair> imageFormatNameSize = {
-		// TODO: find imageDataSize multiplier
+		/* Generic format that does not have any specified actual format
+		and implementations can choose one of a few different ones
+		GL4.6 spec (https://registry.khronos.org/OpenGL/specs/gl/glspec46.core.pdf):
+		8.5. TEXTURE IMAGE SPECIFICATION
+		"If internalformat is specified as a base internal format, the GL stores the resulting texture with
+		 internal component resolutions of its own choosing, referred to as the effective internal format."
+		 Use 4 bytes as an estimate: */
 		{ GL_RGBA, { "RGBA", 4 } },
 
 		{ GL_RGB8, { "RGB8", 3 } },
 		{ GL_RGBA8, { "RGBA8", 4 } },
 		{ GL_RGB16, { "RGB16", 6 } },
-		{ GL_RGBA16, { "RGBA16", 6 } },
+		{ GL_RGBA16, { "RGBA16", 8 } },
 		{ GL_RGB16F, { "RGB16F", 6 } },
 		{ GL_RGB32F, { "RGB32F", 12 } },
 		{ GL_RGBA16F, { "RGBA16F", 8 } },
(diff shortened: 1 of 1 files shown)


src/engine/renderer/tr_image.cpp
```diff
@@ -315,8 +357,26 @@ void R_ListImages_f()
...
+				case GL_COMPRESSED_SIGNED_RED_RGTC1:
+				case GL_COMPRESSED_RG_RGTC2:
+				case GL_COMPRESSED_SIGNED_RG_RGTC2:
+				case GL_COMPRESSED_RGB_S3TC_DXT1_EXT:
+				case GL_COMPRESSED_RGBA_S3TC_DXT1_EXT:
+				case GL_COMPRESSED_RGBA_S3TC_DXT3_EXT:
+				case GL_COMPRESSED_RGBA_S3TC_DXT5_EXT:
+					format = imageFormatNameSize.at( image->internalFormat ).first;
+					imageDataSize = ceil( image->uploadWidth / 4.0 ) * ceil( image->uploadHeight / 4.0 ) * image->numLayers
+									* imageFormatNameSize.at( image->internalFormat ).second;
```
illwieckz, line 373: Better do a single tab character for a single indentation, ascii art is not for indentation.

illwieckz (approved):
LGTM % whitespace.

src/engine/renderer/tr_image.cpp
```diff
@@ -145,13 +145,19 @@ void R_ListImages_f()
 
 	typedef std::pair<const std::string, int> nameSizePair;
 	static const std::unordered_map<uint32_t, nameSizePair> imageFormatNameSize = {
-		// TODO: find imageDataSize multiplier
+		/* Generic format that does not have any specified actual format
+		   and implementations can choose one of a few different ones
```
illwieckz, line 149: Better avoid any space between  the indentation and the comment text.

The only places where I have seen somewhat meaningful sub-comment indentation was for code quotes or things like that (a bit like a kind of in-comment-markdown).

src/engine/renderer/tr_image.cpp
VReaperV: You mean for line 373? I do that to have formulas left-aligned, but I can change it to a single tab.

src/engine/renderer/tr_image.cpp
```diff
@@ -166,23 +172,59 @@ void R_ListImages_f()
...
+		   blocksize (8 or 16 bytes) is decoded, the corresponding image size (in
+		   bytes) is: ceil( w / 4 ) * ceil( h / 4 ) * blocksize."
+
+		   All of the following have blocksize of 8 bytes: */
+		{ GL_COMPRESSED_RGB_S3TC_DXT1_EXT, { "DXT1", 8 } },
+		{ GL_COMPRESSED_RGBA_S3TC_DXT1_EXT, { "DXT1a", 8 } },
+		{ GL_COMPRESSED_RGBA_S3TC_DXT3_EXT, { "DXT3", 8 } },
+		{ GL_COMPRESSED_RGBA_S3TC_DXT5_EXT, { "DXT5", 8 } },
+
+		/*
```
illwieckz, line 207: Comment open/end token better be on same line (unless it is ambiguous with the meaning of the line, which is very rare.

VReaperV: Yep, missed that one when I was going through the formatting lol.

src/engine/renderer/tr_image.cpp
VReaperV: I prefer it left-aligned, but sure I'll make it aligned to the /* itself.

src/engine/renderer/tr_image.cpp
VReaperV: Fixed.

src/engine/renderer/tr_image.cpp
VReaperV: Changed it to align to /*.

src/engine/renderer/tr_image.cpp
illwieckz: Yes, the line 373. We better avoid to abuse ascii art and reserve it for some schematics in comments or so.

Continuing a line on the next line is better considered an indentation as convention.

In case it would be very wanted (I don't think of any use case anyway), in-block alignment should be done with space as tab size is variable.

Basically:

- indentation = tabs only
- ascii art = spaces only

Doing that makes the spacing being semantic for the human reader, and avoids all the spacing mismatches.

VReaperV: That's true w. r. t. tabs yeah.

VReaperV: Changed this to 1 tab now.

Issue #360, opened 2020-08-04
Looking for a name for a cvar to make system-specific workarounds disableable


illwieckz:
We already have one workaround for buggy nvidia driver (see #63), but it would be useful to be able to disable the workaround to check if the bug still occurs or on which hardware/software combination it occurs.

I would like to see other workaround implemented, like:

- forcing `r_picmip 1` for systems having 2GB or less of RAM. I verified Dæmon running on a system with 2GB of ram (even running the lightweight IceWM desktop on Linux) cannot load `plat23` map with full size textures without being killed by the OS to recover from an _Out Of Memory_ error;

- forcing `r_picmip 1` for systems having less than 512MB of VRAM, I verified systems having 256MB or 384MB of VRAM cannot load `plat23` map with full size textures without hanging the whole computer (on Linux).

There may bee other workarounds to implement to make sure players do not end with a game being silently killed by the OS or the hardware going mad in situation we now well know.

But I'm not really happy with the idea of not being able to disable those workarounds, I want to be able to check in the future if a given workaround is still legit or not. A cvar would be very good for such purpose. The idea would be to have a boolean cvar being enabled by default (disabling the cvar would disable the workarounds).

I'm lacking imagination right now, so if you have an idea, please make suggestions!

illwieckz:
I say « hardware specific » at first, but the same cvar would cover system specific bugs in general, like the nvidia driver one.

slipher:
How about `useHardwareWorkarounds`. Yeah OK an Nvidia driver technically isn't hardware, but close enough.

illwieckz:
What prefix to use? and why some cvars use `dot.naming.scheme` and others use `a_prefixedName`?

slipher:
The dot scheme is a new one introduced in Daemon, but of course most of the old ones were never migrated... not convinced it's better as the dot naming is usually longer to type.

Use `r_` prefix I guess since all hardware workarounds are related to renderer AFAIK.

hypov8:
something like r_ignore...
r_ignoreMemLimitation
r_ignoreNvidiaFix

VReaperV:
Done in #1224.

Issue #343, opened 2020-06-29
Rendering bugs with Radeon HD 3650 (RV635) on Linux with Mesa


illwieckz:
Given we recommend OpenGL 3.2 GPU (and the game may work with OpenGL 2.1) I may attempt to plug some older GPU to my computer to drive some tests.

I just did it with a Radeon HD 3650 (RV635) on Linux with Mesa 20.0.8 and `glxinfo` tells me OpenGL 3.3 is supported.

Don't expect a stellar experience (expect a 30~40 fps range) but it works, and except a warning telling me SSAO is not used because `GL_ARB_texture_gather` is not available, all features seems to work as far as I've seen (even relief mapping works, but you have to disable it because of performance).

The bug is that some far surfaces that are hidden by close geometry are rendered atop the close geometry hiding them. I can get that bug in plat23 or yocto maps for example. On chasm map when being outside the skybox is rendered above everything.

It may also be a driver bug, I don't know.

When outside is not rendered:

[![ati radeon hd3650 rv635](https://dl.illwieckz.net/b/daemon/bugs/ati-radeon-hd3650-rv635/unvanquished_2020-06-29_104709_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/ati-radeon-hd3650-rv635/unvanquished_2020-06-29_104709_000.jpg)

When outside is rendered (walked a bit to the left):

[![ati radeon hd3650 rv635](https://dl.illwieckz.net/b/daemon/bugs/ati-radeon-hd3650-rv635/unvanquished_2020-06-29_104727_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/ati-radeon-hd3650-rv635/unvanquished_2020-06-29_104727_000.jpg)

I'm close to a wall entirely hiding my viewport but I still see through:

[![ati radeon hd3650 rv635](https://dl.illwieckz.net/b/daemon/bugs/ati-radeon-hd3650-rv635/unvanquished_2020-06-29_104817_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/ati-radeon-hd3650-rv635/unvanquished_2020-06-29_104817_000.jpg)

The plat23 map really makes this bug very obvious:

[...]

PR #1224, opened 2024-07-27
sys: add `workaround.*` cvars to disable workarounds

illwieckz:
So, I'm thinking about it for years now, probably since the current implementation of the nvidia 340 workaround:

- https://github.com/DaemonEngine/Daemon/pull/370

I even discussed it there:

- https://github.com/DaemonEngine/Daemon/issues/360

The idea is to selectively disable a workaround we implement, to be able to test the code without it.

This is especially useful when bisecting a bug in some underlying tech like a driver, to make sure we actually test the feature in the way we expect it to fail if the third-party bug is still there.

For now I go with a dot-based hierarchy (there is no legacy cvar to mimic), and I currently use the `renderer.workaround.<whatIsDone>.<whenItIsRequired>` naming scheme.

For example `renderer.workaround.noTexturGather.nvidia340` disables `ARB_texture_gather` on Nvidia driver version 340, while `renderer.workaround.firstProvokingVertex.intel` makes the engine use first provoking vertex on Intel hardware supporting `ARB_provoking_vertex`.

The implementation is trivial, I'm mostly looking for comments about the idea itself, the naming scheme…

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -130,6 +130,7 @@ static Cvar::Cvar<bool> workaround_noBindlessTexture_mesa241( "workaround.noBind
 static Cvar::Cvar<bool> workaround_noBindlessTexture_amdOglp( "workaround.noBindlessTexture.amdOglp", "Disable ARB_bindless_texture on AMD OGLP driver", Cvar::NONE, true );
 static Cvar::Cvar<bool> workaround_noHyperZ_mesaRv600( "workaround.noHyperZ.mesaRv600", "Disable Hyper-Z on Mesa driver on RV600 hardware", Cvar::NONE, true );
 static Cvar::Cvar<bool> workaround_noTextureGather_nvidia340( "workaround.noTextureGather.nvidia340", "Disable ARB_texture_gather on Nvidia 340 driver", Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_s3tc_mesa( "workaround.s3tc.mesa", "Enable S3TC on Mesa even when libtxc-dxtn is not available", Cvar::NONE, true );
 
 SDL_Window *window = nullptr;
 static SDL_GLContext glContext = nullptr;
@@ -2444,8 +2445,20 @@ bool GLimp_Init()
 	Cvar::Latch( workaround_noBindlessTexture_amdOglp );
 	Cvar::Latch( workaround_noHyperZ_mesaRv600 );
 	Cvar::Latch( workaround_noTextureGather_nvidia340 );
+	Cvar::Latch( workaround_s3tc_mesa );
 
 	ri.Cmd_AddCommand( "minimize", GLimp_Minimize );
+
+	/* Enable S3TC on Mesa even if libtxc-dxtn is not available
+	The environment variables is currently always set,
+	it should do nothing with other systems and drivers.
+
+	It should also be set on Win32 when running on Wine
+	on Linux anyway. */
+	if ( workaround_s3tc_mesa.Get() )
+	{
+		Sys::SetEnv( "force_s3tc_enable", "true" );
+	}
 
 	/* Enable 2.1 GL on Intel GMA Gen 3 on Linux Mesa driver.
 
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1345,10 +1345,23 @@ static void GLimp_RegisterConfiguration( const glConfiguration& highestConfigura
 		glConfig2.glMinor = GLminor;
 	}
 
-	/* FIXME: a duplicate of this is done in GLimp_Init, also storing it
-	for gfxinfo command. */
-	const char *glstring = ( char * ) glGetString( GL_RENDERER );
-	logger.Notice("OpenGL Renderer: %s", glstring );
+	// Get our config strings.
+	Q_strncpyz( glConfig.vendor_string, ( char * ) glGetString( GL_VENDOR ), sizeof( glConfig.vendor_string ) );
+	Q_strncpyz( glConfig.renderer_string, ( char * ) glGetString( GL_RENDERER ), sizeof( glConfig.renderer_string ) );
+	Q_strncpyz( glConfig.version_string, ( char * ) glGetString( GL_VERSION ), sizeof( glConfig.version_string ) );
+
+	if ( *glConfig.renderer_string )
+	{
+		int last = strlen( glConfig.renderer_string ) - 1;
+		if ( glConfig.renderer_string[ last ] == '\n' )
+		{
+			glConfig.renderer_string[ last ] = '\0';
+		}
+	}
+
+	logger.Notice("OpenGL vendor: %s", glConfig.vendor_string );
+	logger.Notice("OpenGL renderer: %s", glConfig.renderer_string );
+	logger.Notice("OpenGL version: %s", glConfig.version_string );
 }
 
 static void GLimp_DrawWindowContent()
@@ -2395,17 +2408,6 @@ bool GLimp_Init()
 success:
 	// These values force the UI to disable driver selection
 	glConfig.hardwareType = glHardwareType_t::GLHW_GENERIC;
-
-	// get our config strings
-	Q_strncpyz( glConfig.vendor_string, ( char * ) glGetString( GL_VENDOR ), sizeof( glConfig.vendor_string ) );
-	Q_strncpyz( glConfig.renderer_string, ( char * ) glGetString( GL_RENDERER ), sizeof( glConfig.renderer_string ) );
-
-	if ( *glConfig.renderer_string && glConfig.renderer_string[ strlen( glConfig.renderer_string ) - 1 ] == '\n' )
-	{
-		glConfig.renderer_string[ strlen( glConfig.renderer_string ) - 1 ] = 0;
-	}
-
-	Q_strncpyz( glConfig.version_string, ( char * ) glGetString( GL_VERSION ), sizeof( glConfig.version_string ) );
 
 	DetectGLVendors(
 		glConfig.vendor_string,
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -126,6 +126,7 @@ static Cvar::Cvar<bool> r_khr_shader_subgroup( "r_khr_shader_subgroup",
 static Cvar::Cvar<bool> workaround_extFbo_missingArbFbo( "workaround.extFbo.missingArbFbo", "Use EXT_framebuffer_object and EXT_framebuffer_blit when ARB_framebuffer_object is not available", Cvar::NONE, true );
 static Cvar::Cvar<bool> workaround_firstProvokingVertex_intel( "workaround.firstProvokingVertex.intel", "Use first provoking vertex on Intel hardware supporting ARB_provoking_vertex", Cvar::NONE, true );
 static Cvar::Cvar<bool> workaround_noBindlessTexture_mesa241( "workaround.noBindlessTexture.mesa241", "Disable ARB_bindless_texture on Mesa 24.1 driver", Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_noBindlessTexture_amdOglp( "workaround.noBindlessTexture.amdOglp", "Disable ARB_bindless_texture on AMD OGLP driver", Cvar::NONE, true );
 static Cvar::Cvar<bool> workaround_noTextureGather_nvidia340( "workaround.noTextureGather.nvidia340", "Disable ARB_texture_gather on Nvidia 340 driver", Cvar::NONE, true );
 
 SDL_Window *window = nullptr;
@@ -2202,8 +2203,39 @@ static void GLimp_InitExtensions()
 			}
 		}
 
+		bool foundOglp = false;
+
+		if ( workaround_noBindlessTexture_amdOglp.Get() )
+		{
+			/* AMD OGLP driver for Linux shares the same vendor string than AMD Adrenalin driver
+			for Windows and AMD ATI driver for macOS. When running the Windows engine binary on
+			Wine we must check we're not running Windows or macOS to detect Linux OGLP. */
+			if ( Q_stristr( glConfig.vendor_string, "ATI Technologies Inc." ) )
+			{
+				#if defined(_WIN32)
+					// Detect Wine being used.
+					if ( Sys::isRunningOnWine() )
+					{
+						// Detect AMD ATI driver on macOS not being used.
+						if ( !Q_stristr( glConfig.version_string, "ATI-" ) )
+						{
+							foundOglp = true;
+						}
+					}
+				#elif defined(__APPLE__)
+					// AMD ATI driver for macOS is believed to not implement bindless texture.
+				#else
+					foundOglp = true;
+				#endif
+			}
+
+			if ( foundOglp ) {
+				logger.Warn( "...found buggy AMD OGLP driver, ARB_bindless_texture disabled" );
+			}
+		}
+
 		// not required by any OpenGL version
-		glConfig2.bindlessTexturesAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, ARB_bindless_texture, r_arb_bindless_texture.Get() && !foundMesa241 );
+		glConfig2.bindlessTexturesAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, ARB_bindless_texture, r_arb_bindless_texture.Get() && !foundMesa241 && !foundOglp );
 	}
 
 	// made required in OpenGL 4.6
@@ -2334,6 +2366,7 @@ bool GLimp_Init()
 	Cvar::Latch( workaround_extFbo_missingArbFbo );
 	Cvar::Latch( workaround_firstProvokingVertex_intel );
 	Cvar::Latch( workaround_noBindlessTexture_mesa241 );
+	Cvar::Latch( workaround_noBindlessTexture_amdOglp );
 	Cvar::Latch( workaround_noTextureGather_nvidia340 );
 
 	ri.Cmd_AddCommand( "minimize", GLimp_Minimize );
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -123,6 +123,11 @@ static Cvar::Cvar<bool> r_khr_debug( "r_khr_debug",
 static Cvar::Cvar<bool> r_khr_shader_subgroup( "r_khr_shader_subgroup",
 	"Use GL_KHR_shader_subgroup if available", Cvar::NONE, true );
 
+static Cvar::Cvar<bool> workaround_extFbo_missingArbFbo( "workaround.extFbo.missingArbFbo", "Use EXT_framebuffer_object and EXT_framebuffer_blit when ARB_framebuffer_object is not available", Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_firstProvokingVertex_intel( "workaround.firstProvokingVertex.intel", "Use first provoking vertex on Intel hardware supporting ARB_provoking_vertex", Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_noBindlessTexture_mesa241( "workaround.noBindlessTexture.mesa241", "Disable ARB_bindless_texture on Mesa 24.1 driver", Cvar::NONE, true );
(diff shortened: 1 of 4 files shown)


src/engine/sys/sdl_glimp.cpp
```diff
@@ -101,6 +101,12 @@ static Cvar::Cvar<bool> r_ext_texture_integer( "r_ext_texture_integer",
 static Cvar::Cvar<bool> r_ext_texture_rg( "r_ext_texture_rg",
 	"Use GL_EXT_texture_rg if available", Cvar::NONE, true );
 
+static Cvar::Cvar<bool> workaround_noTexturGather_nvidia340( "workaround.noTexturGather.nvidia340", "Disable ARB_texture_gather on Nvidia 340 driver", Cvar::NONE, true );
```
VReaperV, line 104: `workaround_noTextur**e**Gather_nvidia340`

src/engine/sys/sdl_glimp.cpp
```diff
@@ -101,6 +101,12 @@ static Cvar::Cvar<bool> r_ext_texture_integer( "r_ext_texture_integer",
...
+static Cvar::Cvar<bool> workaround_firstProvokingVertex_intel( "workaround.firstProvokingVertex.intel",
+"Use first provoking vertex on Intel hardware supporting ARB_provoking_vertex", Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_noBindlessTexture_mesa241( "workaround.noBindlessTexture.mesa241", "Disable EXT_bindless_texture on Mesa 24.1 driver", Cvar::NONE, true );
```
VReaperV, line 107: It's ARB  rather than EXT.

src/engine/sys/sdl_glimp.cpp
illwieckz: Fixed.

src/engine/sys/sdl_glimp.cpp
illwieckz: Fixed.

illwieckz:
I added a code to disable bindless texture on AMD OGLP on Linux, and the related `workaround` cvar.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -101,6 +101,13 @@ static Cvar::Cvar<bool> r_ext_texture_integer( "r_ext_texture_integer",
 static Cvar::Cvar<bool> r_ext_texture_rg( "r_ext_texture_rg",
 	"Use GL_EXT_texture_rg if available", Cvar::NONE, true );
 
+static Cvar::Cvar<bool> workaround_noTextureGather_nvidia340( "workaround.noTextureGather.nvidia340", "Disable ARB_texture_gather on Nvidia 340 driver", Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_firstProvokingVertex_intel( "workaround.firstProvokingVertex.intel",
+"Use first provoking vertex on Intel hardware supporting ARB_provoking_vertex", Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_noBindlessTexture_mesa241( "workaround.noBindlessTexture.mesa241", "Disable ARB_bindless_texture on Mesa 24.1 driver", Cvar::NONE, true );
```
slipher, line 107: Why do we need 2 different cvars about enabling the same bindless texture feature?

src/engine/sys/sdl_glimp.cpp
```diff
@@ -101,6 +101,13 @@ static Cvar::Cvar<bool> r_ext_texture_integer( "r_ext_texture_integer",
 static Cvar::Cvar<bool> r_ext_texture_rg( "r_ext_texture_rg",
 	"Use GL_EXT_texture_rg if available", Cvar::NONE, true );
 
+static Cvar::Cvar<bool> workaround_noTextureGather_nvidia340( "workaround.noTextureGather.nvidia340", "Disable ARB_texture_gather on Nvidia 340 driver", Cvar::NONE, true );
```
slipher, line 104: ISTM they should have `r_` prefix

src/engine/sys/sdl_glimp.cpp
illwieckz: This is not about enabling a bindless texture feature, it's is about disabling a workaround preventing bindless texture usage in a special case. The idea is to have a `workaround` cvar for disabling each workaround use case. Those are debugging cvars and having fine granularity is better.

src/engine/sys/sdl_glimp.cpp
illwieckz: This is part of the discussion I want to have indeed.

I assume the hierarchical dot-spaced cvar names is “the new naming”, while `<prefix>_camelCase is “the legacy naming”. We still use the legacy naming a lot, mainly for consistency, like within the renderer (`r_*`), but it doesn't mean we cannot continue using new naming. I even myself suggested in the past we may implement cvar name aliases to be able to port all cvars to the dotted-hierarchical naming while retaining compatibility with old names in scripts (or in user's memory).

This workaround namespace is totally new, so there is no cargo cult attached with it, so it may sounds counter-productive to do it the legacy way if we consider the dotted-hierarchical way is the future. That `workaround.*` namespace is meant to live alongside existing `logs.*`, `common.*`, `vm.*`, `audio.*`, etc.

Now we can discuss the naming, is `workaround` the right top level name? Should it be named otherwise? Should it be child of another name?

Then comes the question about all those workarounds being about the renderer, so we may suggest `renderer.workaround.noTextureGather.nvidia340`.

But we may have other workarounds in the future for other things, like audio, input, network or filesystem and want all of them under the same tree for easy completion, so we may suggest `workaround.renderer.noTextureGather.nvidia340`.

In the end the extra parent (here `.renderer`) between the top level and the workaround name seems to be very noisy, I strongly doubt a possible audio, input, network or filesystem workaround name would collide with a renderer one, so I went with listing all workarounds under the `workaround` top level.

Something I want to remind is that we better not mix the naming schemes, likes using dots in a `r_` cvar, using underscores in a `common.` cvar, or using underscore as tree separator in place of the dot (i.e. no more than one underscore), I now some bot cvars did the latter but we better not want to do more and limit the amount of conventions to one legacy and one new and not mix them.

illwieckz:
I implemented a fix for:

- https://github.com/DaemonEngine/Daemon/issues/343

With `workaround.noHyperZ.mesaRv600 off`:

unvanquished_2024-07-30_140425_000

With `workaround.noHyperZ.mesaRv600 on`:

unvanquished_2024-07-30_140532_000

The fix requires to restart SDL _after_ having read `GL_RENDERER`, so a mechanism to restart SDL on some conditions was implemented.

It was tested on Radeon HD 3650 (RV635) on Mesa 24.0.9.

src/engine/sys/sdl_glimp.cpp
slipher: I don't think it's a good idea to make workaround a top-level namespace. I think all renderer cvars is a more useful category than all workaround

illwieckz: I think it's better to not pollute an existing namespace like `r_` and I actually like having a dedicated namespace for this.

One usage of those cvars will be to use them in bug reports, giving driver developers a way to reproduce a bug we have a workaround for. For example we would be able to tell a Mesa developer to do this to reproduce the R600 Hyper-Z bug:

```
./daemon -set workaround.noHyperZ.mesaRv600 off +devmap plat23
```

I like the way it looks, and it is easy to autocomplete as well.

illwieckz:
Outside of the naming topic, I've implemented some new workarounds, they work as expected on my end.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1422,6 +1425,82 @@ static void GLimp_CheckGLEW( const glConfiguration &requestedConfiguration )
...
+{
+	if ( !firstSdlVideoAttempt )
+	{
+		return false;
+	}
+
+	/* We call RV600 the first generation of R600 cards, to make a difference
+	with RV700 and RV800 cards that are also supported by the Mesa R600 driver.
+
+	The Mesa R600 driver has broken Hyper-Z wth RV600, not RV700 neither RV800. */
```
slipher, line 1443: s/neither/nor/

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1422,6 +1425,82 @@ static void GLimp_CheckGLEW( const glConfiguration &requestedConfiguration )
 	}
 }
 
+static bool IsSdlVideoRestartNeeded()
+{
+	if ( !firstSdlVideoAttempt )
```
slipher, line 1435: I don't think this variable needs to exist.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1422,6 +1425,82 @@ static void GLimp_CheckGLEW( const glConfiguration &requestedConfiguration )
...
+	{
+		return false;
+	}
+
+	/* We call RV600 the first generation of R600 cards, to make a difference
+	with RV700 and RV800 cards that are also supported by the Mesa R600 driver.
+
+	The Mesa R600 driver has broken Hyper-Z wth RV600, not RV700 neither RV800. */
+	static bool r600_hyperz_modified = false;
+	static const char* r600_hyperz = nullptr;
```
slipher, line 1445: This could use a comment or more specific naming. It took me a long time to figure out that the point is to restore a possible user-specified value of the environment variable.

But actually I might suggest that if there is any value already set for this, the workaround should be disabled.

src/engine/sys/sdl_glimp.cpp
illwieckz: I removed it, though I liked the explicitness of it instead of relying on per-workaround restart check.

src/engine/sys/sdl_glimp.cpp
illwieckz: > But actually I might suggest that if there is any value already set for this, the workaround should be disabled.

I implemented that, which simplifies the code.  It also makes possible to remove `firstSdlVideoAttempt` as the presence of the cvar being set by the workaround is enough to know the restart already happened.

I have to test the code again on the hardware to be sure I have not introduced any regression from my original code.

src/engine/sys/sdl_glimp.cpp
illwieckz: Fixed.

illwieckz:
Those being developer cvars, they are not expected to be used by users and it's fine to can rename them later if we find a better naming later. I don't like to have those cvars in  `r_` namespace with annoying name, and the dotted hierarchy actually look better for them. I prefer `workaround.noBindlessTexture.mesa241` to `r_workaroundNoBindlessTextureMesa241`…

The code looks fine to me and I don't want to postpone the feature because of naming concerns.

src/engine/sys/sdl_glimp.cpp
illwieckz: > I have to test the code again on the hardware to be sure I have not introduced any regression from my original code.

It still works, I haven't introduced any regression when simplifying the code.

slipher:
Maybe some 3rd person can give an opinion on whether to stick with `r_` prefix or not?

src/engine/sys/sdl_glimp.cpp
VReaperV: I'll reiterate the point I made on dev channel, here: workaround cvars would be out of place under `r_*`, because most of the time you won't use them. Either `workaround.renderer.*` or `renderer.workaround.*` would be better. The latter especially so, since it's very unlikely we'd have such workarounds for non-renderer things: any sort of OS bug or difference in handling something would most likely be handled at the compiler-level, and we wouldn't be able to anything about it at runtime, unlike avoiding graphics driver bugs.

slipher: > I'll reiterate the point I made on dev channel, here: workaround cvars would be out of place under `r_*`, because most of the time you won't use them.

This point doesn't make sense to me as half of the cvars under `r_` are obscure debugging things already. The workaround stuff is at least something a player might want to use in principle (if the workaround misclassifies the hardware/driver as broken when it actually works fine).

Anyway, I agree that non-renderer workarounds will be extremely rare and so a renderer-based namespace is better. So let's go with `renderer.workaround`.

illwieckz: So I renamed to `renderer.workaround`. 

I don't really see those cvars as ones a player may use, as the provided fallbacks are assumed to have more chance to work than what they are working around.

It's more thought as a developer things, both for engine developers and drivers developers: being able to test code path we intentionally don't provide to users by default on a given hardware or driver.

illwieckz:
I guess I can go forward and merge it?

DolceTriade:
If `renderer.*` is the future, then I'm fine with that. However, if we have no intention of doing a full migration to r_ -> renderer, then I don't think we should use the renderer prefix.

Not a strong opinion either way.

illwieckz:
@DolceTriade here is a thread on the topic:

- https://github.com/DaemonEngine/Daemon/pull/1224#discussion_r1696175164

Basically my own opinion is:

- I don't want to add more `r_` cvars and especially not pollute them with this ones.
- I do think `renderer.*` is the future, I prefer such hierarchical scheme (like we have with `common.framerate.max` or `vm.cgame.type`) than the old messy `x_*` in general, even if more verbose. The only reason I'm still creating `r_*` renderer cvars is for consistency (we better to convert them both at once if we do, and would need an alias mechanism for backward compatibility of scripts).
- I do believe having a `workaround.` prefix is a good idea, like we have.
- I don't see the need to write in the name some `renderer` string, even if maybe all those workarounds will be for the renderer, I'm not excluding the fact we may have to implement a workaround for audio or something else.
- Those cvars are meant to be used by developers (either engine developers, either driver developers, either operating system developers) or people doing bug reporting and testing while being in touch with a developer, not by users.

Actually I do prefer a generic `workaround.*` and I don't see the need to add any `renderer` string somewhere, the name part `noBindless` or `gl21` is enough for a developer. Adding a `renderer` component in the name just looks verbose for no reason.

Even if we go the `renderer.` way for `r_` cvars one day, I don't see the need for those workaround cvars to be part of that.

illwieckz:
In fact we actually do have non-renderer workarounds I want to implement disablement cvars for, related to the nacl loader.

So I believe the better thng is to do `workaround.topic.name`. And the topic for the ones in this PR may even not be the renderer. We do not workaround the renderer, we workaround opengl software or hardware.

So it appears to me better names would be like that:

- `workaround.opengl.*`
- `workaround.nacl.*`

and possibly if needed:

- `workaround.glew.*`
- `workaround.sdl2.*`
- `workaround.openal.*`

etc.

VReaperV:
If we're being pedantic, it's `workaround.driver.*` since the issue is not within the API spec, but with the specific implementation.

illwieckz:
I did that:

```
workaround.glDriver.amd.oglp.disableBindlessTexture
workaround.glDriver.mesa.ati.rv300.disableRgba16Blend
workaround.glDriver.mesa.ati.rv600.disableHyperZ
workaround.glDriver.mesa.forceS3tc
workaround.glDriver.mesa.intel.gma3.forceFragmentShader
workaround.glDriver.mesa.intel.gma3.stubOcclusionQuery
workaround.glDriver.mesa.v241.disableBindlessTexture
workaround.glDriver.nvidia.v340.disableTextureGather
workaround.glExt.missingArbFbo.useExtFbo
workaround.glHardware.intel.useFirstProvokinVertex
```

I have another `glHardware` workaround to implement but it can be done later and we don't need to wait for it. This is just to say even the tree branches whith currently only one leaf may get more leaves in the future.

DolceTriade:
ok sgtm

slipher:
1. Should namespaces be hierarchical? AFAIK no one really opposes that. We can make hierarchies equally well with dot or underscore separators, for example `g_bot_`. Though not all the workaround cvars given above actually seem hierarchical; sometimes it's just a generic word/phrase separator:
  - `workaround.glExtension.missingArbFbo.useExtFbo`: `missingArbFbo` doesn't make sense as a hierarchical element.
  - `workaround.glDriver.mesa.v241.disableBindlessTexture`: version 24.1 shouldn't be a hierarchical element because the work around is not specific to *exactly* 24.1, but rather a version range starting from there. I think some of the ones with hardware version numbers have this property too.
2. Should namespaces be like `r_`, `cl_`, or `renderer.`, `client.`? I prefer the former:
  - The convention of naming cvar C++ objects the same as the cvar name is nice, so that all references can be searched for with a single string.
  - The dotted ones are too verbose. It's annoying every type I don't have my maxfps alias and have to type `common.framerate.max`.
  - Status quo bias - Everyone's already used to the current thing; we don't need to migrate scripts (or create multiple names of one cvar which would also be unfortunate); same as other engines, etc.
3. Should workarounds have a top-level namespace? Heretofore all cvar namespaces were based on subsystems of the program: `g_` for sgame, `cg_` for cgame, `r_` for renderer, `sv_` for server... Only `com_`/`common.` is a bit unclear to me; I think it's supposed to be a part of the engine that's included in both client and server builds (but some FPS cvars which only affect the client must be there by mistake). I'm not sure if illwieckz considers that prefixing cvars by subsystem is a bad idea in general and wants to reorganize top-level namespaces around different principles? Or that "workarounds" have some very special reason to be an exception?

illwieckz:
> The convention of naming cvar C++ objects the same as the cvar name is nice, so that all references can be searched for with a single string.

There may be a bias from my vim-centered mind but a regex searching for `workaround.glDriver.amd.oglp.disableBindlessTexture` also matches `workaround_glDriver_amd_oglp_disableBindlessTexture`.

For renaming the cvars I just did `%s/one\([_.\]\)two.three.four/un\1deux\1trois\1quatre/`.

Of course an universal `_` separator would be easier on that side. I don't really know why the “new cvar naming style” used dots, on that part I'm just sticking to what is already done (like `vm.cgame.type` etc.). I'm not disliking the dot that much though.

> Should workarounds have a top-level namespace? […] Or that "workarounds" have some very special reason to be an exception?

I consider `workaround` to be an exception. The purpose of them is to be used in test scripts, CI, bug reports giving instructions to third-party software developers (like drivers) to test a specific code path, giving some copy-pastable test commands to a tester, etc.  I expect that most usages of such cvars is to be copy pasted commands. The location of them in the source code doesn't matter to me, they targets specific hardware, libraries or operating system components.

For example I would not exclude the possibility of some `workaround.alDriver.alsa.` or `workaround.alHardware.` (audio) or a `workaround.fs.vfat.` (filesystem) cvars. I  doubt they will exist in engine are we may not be low-level enough on those topics, but for having written some radio-broadcast-purposed software in my life, I know that even some sound cards may require similar workarounds as the ones we implemented for graphics card, for example some sound cards may claim to support 48KHz sampling but only produce garbage if used, and an audio software would have to detect such card and force 44KHz and ignore the advertised capability if the default sampling of the software is 48KHz. This is very similar to the Nvidia workaround we have with a driver for specific cards falsely claiming to support `ARB_texture_gather` and failing if attempting to make use of it. Even for filesystems, people doing portable installations of the game on USB sticks may require some workarounds. For example: what if we attempt to create the engine IPC pipe on FAT32? We may also have a `workaround.` cvar to disable the recently implemented case-independent search if we want to test direct access to the file system for whatever debugging purpose.

Also having a `workaround` namespace makes it very easy to get a list of those workaround this way:

```
/listCvars workaround
```

This makes me think it would be nice if `listCvars` would print things in alphabetical order…

illwieckz:
Also I want to say such cvars may be renamed in the future. For example if the bindless texture bug in OGLP is fixed one day, we may rename the related workaround cvar to mention the version. I don't consider those names to be written in stone. All we need is to have something to copy-paste for testing, and I prefer such cvar names to be verbose. For workaround cvars I consider verbosity a feature.

illwieckz:
> * `workaround.glExtension.missingArbFbo.useExtFbo`: `missingArbFbo` doesn't make sense as a hierarchical element.

Well, I mainly named the cvars this way: `workaround.topic.whenTheWorkaroundIsApplicable.enableWorkaround`.

We can continue the discussion later and rename the cvars later, especially since they're not meant to be used by users and to be set in presets. I prefer to not wait more for merging this to not delay the implementation of other patches above this code. It looks like the logic itself is ready and has no comment to receive anymore.

So let's do it.

slipher:
> Of course an universal `_` separator would be easier on that side. I don't really know why the “new cvar naming style” used dots, on that part I'm just sticking to what is already done (like `vm.cgame.type` etc.). I'm not disliking the dot that much though.

I guess we can split that up into question (2a) `r` vs. `renderer` and (2b) `_` vs. `.`.



> I consider `workaround` to be an exception. The purpose of them is to be used in test scripts, CI, bug reports giving instructions to third-party software developers (like drivers) to test a specific code path, giving some copy-pastable test commands to a tester, etc. I expect that most usages of such cvars is to be copy pasted commands.

What does any of that have to do with whether `workaround` is a top-level namespace?

> they targets specific hardware, libraries or operating system components.

I can see this argument somewhat. Wanting to organize by an external subsystem rather than one of the game's subsystems.


> Also having a `workaround` namespace makes it very easy to get a list of those workaround

If "workaround" is somewhere in the middle of the names, you can do `listCvars *workaround*`.

PR #1226, opened 2024-07-30
bikeshedding: make syntax coloring happier

illwieckz:
Purely useless change, but the parenthesis pair being broken by some ifdef make unhappy the syntax coloring of vim and that annoys me every time I open this file. 😏️

I guess the compiler can optimize out the extra variable.


illwieckz:
The exact same change as been discussed in Unvanquished/Unvanquished#3327 and accepted:

- https://github.com/Unvanquished/Unvanquished/pull/3327

It is already merged on game side, so we better merge it on engine side for consistency.

PR #1227, opened 2024-07-31
renderer: our usage of GLEW doesn't require GLU

illwieckz:
This `glew.h` include without the `GLEW_NO_GLU` define broke the docker-based release script:

```
In file included from /Unvanquished/daemon/src/engine/renderer/TextureManager.h:40,
                 from /Unvanquished/daemon/src/engine/renderer/tr_local.h:34,
                 from /Unvanquished/daemon/src/engine/renderer/gl_shader.h:26,
                 from /Unvanquished/daemon/src/engine/renderer/gl_shader.cpp:25:
/Unvanquished/daemon/external_deps/linux-amd64-default_10/include/GL/glew.h:1219:14: fatal error: GL/glu.h: No such file or directory
 #    include <GL/glu.h>
              ^~~~~~~~~~
compilation terminated.
```

### DIFF
--- a/src/engine/renderer/TextureManager.cpp
+++ b/src/engine/renderer/TextureManager.cpp
@@ -33,7 +33,6 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 // TextureManager.cpp
 
-#include "TextureManager.h"
 #include "tr_local.h"
 
 Texture::Texture() {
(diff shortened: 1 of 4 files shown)


slipher:
tr_local.h should be included first to handle that stuff without having to repeat it. (We don't need to repeat `vector` either.)

illwieckz:
Rewrote.

illwieckz:
Alternative way would be to not include `TextureManager.h` in `tr_local.h` and only in files requiring it, but including `TextureManager.h` in requiring files right after `tr_local.h` (would reduce the risk of rebuilding everything on file update), tell me what you prefer.

illwieckz:
There is this line in `tr_local.h` in `trGlobals_t`:

```
		TextureManager textureManager;
```

 so `tr_local.h` should include `TextureManager.h` anyway.

slipher:
LGTM

PR #1228, opened 2024-08-03
tr_shade: do not compile unused GLSL shaders

illwieckz:
Skip compiling GLSL shaders that are unused, it not only speeds-up game loading, but also enables the engine to run on lower device not supporting some GLSL shaders when the related functions are disabled (bloom, etc.).

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -1069,6 +1069,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		Cvar::Latch( r_mapOverBrightBits );
 		Cvar::Latch( r_forceLegacyOverBrightClamping );
 		Cvar::Latch( r_lightMode );
+		Cvar::Latch( r_fastsky );
 		r_lightStyles = Cvar_Get( "r_lightStyles", "1", CVAR_LATCH | CVAR_ARCHIVE );
 		r_exportTextures = Cvar_Get( "r_exportTextures", "0", 0 );
 		r_heatHaze = Cvar_Get( "r_heatHaze", "1", CVAR_LATCH | CVAR_ARCHIVE );
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -3258,7 +3258,7 @@ void RB_RenderSSAO()
 {
 	GLimp_LogComment( "--- RB_RenderSSAO ---\n" );
 
-	if ( !GLEW_ARB_texture_gather ) {
+	if ( !glConfig2.textureGatherAvailable ) {
 		return;
 	}
 
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -96,20 +96,19 @@ static void GLSL_InitGPUShadersOrError()
 	gl_shaderManager.load( gl_generic2DShader );
 	gl_shaderManager.load( gl_genericShader );
 
+	// standard light mapping
+	gl_shaderManager.load( gl_lightMappingShader );
+
 	// Material system shaders that are always loaded if material system is available
-	if ( glConfig2.materialSystemAvailable ) {
+	if ( glConfig2.materialSystemAvailable )
+	{
 		gl_shaderManager.load( gl_genericShaderMaterial );
 		gl_shaderManager.load( gl_lightMappingShaderMaterial );
-		gl_shaderManager.load( gl_skyboxShaderMaterial );
-		gl_shaderManager.load( gl_fogQuake3ShaderMaterial );
-		gl_shaderManager.load( gl_heatHazeShaderMaterial );
+
 		gl_shaderManager.load( gl_cullShader );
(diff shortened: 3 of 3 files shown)


illwieckz:
There is `gl_motionblurShader` that is remaining, but `cg_motionblur` is a client cvar…

In fact we can define it in engine, the same way `cg_shadows` is an engine cvar meant to be used in cgame too, but I assume doing that would break compatibility.

VReaperV:
`gl_genericShader` still needs to be loaded in both cases (it's used for entities), same with `gl_heatHazeShader` (used for flames and on `map-usstremor`... but there it's also an entity), and `gl_screenShader` - it's used for bloom. I made the material version for `Render_screen()` functionality, but bloom uses `Tess_InstantQuad()` stuff.

I'm not entirely certain about `gl_skyboxShader`... I initially made the *Material version because we have `Render_skybox()`, which is different from `Tess_StageIteratorSky()`. Right now I don't know of any map that would use the former, or if it's even used in any map. I also suspect that it might not work correctly since `Render_skybox()` seems to expect some generic surface, while the skybox shader that we have uses a cube mesh. Right now either `gl_skyboxShader` still needs to be loaded or `Tess_StageIteratorSky()` should be modified... or just get rid of `Render_skybox()` if it's not something we use anyway.

illwieckz:
OK, for now the code assumes it is too early to do:

```
if ( glConfig2.materialSystemAvailable )
  …
else
  …
```

VReaperV (approved):
LGTM

src/engine/renderer/tr_init.cpp
```diff
@@ -1069,6 +1069,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		Cvar::Latch( r_mapOverBrightBits );
 		Cvar::Latch( r_forceLegacyOverBrightClamping );
 		Cvar::Latch( r_lightMode );
+		Cvar::Latch( r_fastsky );
```
slipher, line 1072: Is it possible to make this non-latch and just let the shader be lazily loaded if the user turns it off during the game?

illwieckz: When `r_lazyShader` is 1 (current default value), we call `GLSL_FinishGPUShaders()` at the end of map load, which calls `gl_shaderManager.buildAll()` to build all remaining loaded shaders that are not built yet.

With `r_lazyShader` set to 2, loaded shaders only get built when used, which is the best to not build never-used permutations, but is the worst in the way it can freeze the game at any time. 

Doing like 1 for all shaders but doing like 2 only for skybox would require some rework in other parts I guess, likely not impossible to do, but I'm not sure the effort is worth some low-end graphics preset like this.

_Actually_ I noticed some 2D shaders still get built in game (like when opening the circle menu or something like that I don't remember exactly) even with `r_lazyShader` being 1, which is a bug. I would prefer to work on fixing that and get the expected behavior than punching another hole in the fishing net.

But actually enabling this cvar means to stop using a shader, do you mean to just keep the shader loaded in the backround for nothing when enabling this cvar while the shader was loaded and used before?

Keeping that shader loaded when it is unused probably only requires to remove that line.

But building it lazily once the map is already loaded and the shader is not used yet and it was not built at map load it not that trivial, and if we remove that line, I expect a crash when a map was loaded with fastsky enabled then fastsky is disabled after that. For this we still need a restart.

slipher: > _Actually_ I noticed some 2D shaders still get built in game (like when opening the circle menu or something like that I don't remember exactly) even with `r_lazyShader` being 1

Right. That's why I assumed that in reality lazy shader loading is always enabled, even when we try to build up-front. But I'm not sure.



> But actually enabling this cvar means to stop using a shader,

That's why I said "if the user turns it off". Disabling it loads a shader.

> But building it lazily once the map is already loaded and the shader is not used yet and it was not built at map load it not that trivial,

My hypothesis was that this would happen automatically if we change nothing at all besides removing the latch. Is it wrong?

illwieckz: What I understand is that once `gl_shaderManager.load( gl_skyboxShader )` is called, `gl_shaderManager.buildAll()` will compile the GLSL shader at map load even if the GLSL shader is not used.

And if `gl_shaderManager.load( gl_skyboxShader )` is not called, then `gl_skyboxShader` is nullptr and the engine will crash as soon as `r_fastsky` will be disabled, and lazy loading would not build something that was never loaded. In-game lazy loading still requires the shader to be loaded first.

I believe in the past there was a proof of concept of CVAR_SHADER that was meant to be a kind of CVAR_LATCH but for GLSL shaders and meant to basically reload the shader without a complete `vid_restart`, but it was probably incomplete and removed for being just a clone of CVAR_LATCH in its state at the time.

VReaperV: I don't see the benefit of deferring skybox shader loading.

illwieckz: I'll keep it latched for now, it makes things easier for other things to come to unify that design. It is better from an user point of view as users don't randomly switch options mid-game and having less shader building is faster, while the cost for a developer to do `vid_restart` is fairly acceptable.

PR #1229, opened 2024-08-05
renderer: query tex indirections and ALU instructions when possible

illwieckz:
Query some hardware limits like tex indirections and tex ALU instructions when possible.

The mechanism is only available on non-Core OpenGL profile, and hardware having low limits and requiring those checks don't support Core profile so it's fine if we don't check the limits on hardware with Core profile.

For reference an ATI R300 prints:

```
GL_MAX_PROGRAM_NATIVE_TEX_INDIRECTIONS_ARB: 16
GL_MAX_PROGRAM_NATIVE_ALU_INSTRUCTIONS_ARB: 64
```

An Intel GMA 3 prints:

```
GL_MAX_PROGRAM_NATIVE_TEX_INDIRECTIONS_ARB: 4
GL_MAX_PROGRAM_NATIVE_ALU_INSTRUCTIONS_ARB: 64
```

And an AMD RDNA3  (Radeon PRO W7600) prints (once GL 3 compatible context is forced):

```
GL_MAX_PROGRAM_NATIVE_TEX_INDIRECTIONS_ARB: 16384 
GL_MAX_PROGRAM_NATIVE_ALU_INSTRUCTIONS_ARB: 16384 
```

The idea is to automatically disable some GLSL shader we know the hardware cannot support, even if the related cvar is enabled.

This is useful for various reasons:

- The default preset is `medium` so many of those incompatible shaders are enabled. Before this patch, the game would load the main menu then quit with an error on map load, user had to restart the game, change the preset in the main menu, and load the map again. After this patch, the user can load the map, and switch to a lower preset in-game without being kicked out of the game.
- The motion blur shader cannot be disabled in engine yet as it is configured by a cgame cvar, this makes the engine unable to run on some low-end hardware on which the user would disable motion blur anyway. We can migrate the `cg_motionblur` cvar to the engine (like it was done for `cg_shadows`), but that may break client compatibility.

We may guard more GLSL shaders in the future by enabling them on low end hardware and catching the errors to know their limit requirements, but this is enough to make the engine run without error on default medium graphics preset on Intel GMA 3. Switching to a lower preset is now only a matter of performances, not something required to load a map.

Running the game on Intel GMA 3 requires:

- this branch to skip the motionblur GLSL shaders because of hardware limits,
  as there is no engine cvar to put in a preset to disable it.
- float vertex support:
  https://github.com/DaemonEngine/Daemon/pull/1179

In the future I may use that mechanism to make the engine automatically fallback the dynamic light renderer from tiled to legacy when tiled is not supported.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -38,25 +38,99 @@ static void EnableAvailableFeatures()
 	glConfig2.dynamicLight = r_dynamicLight.Get();
 	glConfig2.staticLight = r_staticLight.Get();
 
-	if ( r_dynamicLightRenderer.Get() == Util::ordinal( dynamicLightRenderer_t::TILED )
-		&& !glConfig2.textureFloatAvailable )
+	if ( glConfig2.dynamicLight || glConfig2.staticLight )
 	{
-		if ( glConfig2.dynamicLight || glConfig2.staticLight )
+		if ( r_dynamicLightRenderer.Get() == Util::ordinal( dynamicLightRenderer_t::TILED ) )
 		{
-			Log::Warn("Tiled dynamic light renderer not used because GL_ARB_texture_float is not available.");
-		}
+			if ( !glConfig2.textureFloatAvailable )
+			{
+				Log::Warn( "Tiled dynamic light renderer disabled because GL_ARB_texture_float is not available.");
+				glConfig2.dynamicLight = false;
+				glConfig2.staticLight = false;
+			}
 
-		glConfig2.dynamicLight = false;
-		glConfig2.staticLight = false;
+			// See below about ALU instructions on ATI R300 and Intel GMA 3.
+			if ( !glConfig2.glCoreProfile && glConfig2.maxAluInstructions < 128 )
+			{
+				Log::Warn( "Tiled dynamic light rendered disabled because GL_MAX_PROGRAM_ALU_INSTRUCTIONS_ARB is too small: %d", glConfig2.maxAluInstructions );
+				glConfig2.dynamicLight = false;
+				glConfig2.staticLight = false;
+			}
+		}
 	}
 
 	glConfig2.shadowingMode = shadowingMode_t( r_shadows.Get() );
-
 	glConfig2.shadowMapping = glConfig2.shadowingMode >= shadowingMode_t::SHADOWING_ESM16;
 
-	if ( !glConfig2.textureFloatAvailable )
+	if ( glConfig2.shadowMapping )
 	{
-		glConfig2.shadowMapping = false;
+		if ( !glConfig2.textureFloatAvailable )
+		{
+			Log::Warn( "Shadow mapping disabled because ARB_texture_float is not available" );
+			glConfig2.shadowMapping = false;
+		}
+	}
+
+	glConfig2.deluxeMapping = r_deluxeMapping->integer;
+	glConfig2.normalMapping = r_normalMapping->integer;
+	glConfig2.specularMapping = r_specularMapping->integer;
+	glConfig2.physicalMapping = r_physicalMapping->integer;
+	glConfig2.reliefMapping = r_reliefMapping->integer;
+
+	/* ATI R300 and Intel GMA 3 only have 64 ALU instructions, which is not enough for some shader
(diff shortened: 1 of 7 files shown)


src/engine/renderer/tr_shade.cpp
```diff
@@ -38,25 +38,96 @@ static void EnableAvailableFeatures()
...
+	glConfig2.physicalMapping = r_physicalMapping->integer;
+	glConfig2.reliefMapping = r_reliefMapping->integer;
+
+	/* ATI R300 and Intel GMA 3 only have 64 tex ALU instructions, which is not enough for some shader
+	variants. For example the lightMapping shader permutation with macros USE_GRID_LIGHTING and
+	USE_GRID_DELUXE_MAPPING from the medium graphics preset requires 67 ALU.
+	For comparison, ATI R400 and R500 have 512 of them. */
+	if ( !glConfig2.glCoreProfile && glConfig2.maxAluInstructions < 128 )
+	{
+		static const std::vector<std::pair<bool*, std::string>> aluFeatures = {
```
slipher, line 84: Can be array rather than vector

src/engine/renderer/tr_shade.cpp
```diff
@@ -38,25 +38,96 @@ static void EnableAvailableFeatures()
...
+			/* Normal mapping, specular mapping and physical mapping does nothing when deluxe mapping
+			is disabled. Hardware that can't do deluxe mapping or normal mapping is not powerful
+			enoough to do relief mapping. */
+			{ &glConfig2.deluxeMapping, "deluxeMapping" },
```
slipher, line 88: These strings are only used in a log message so shouldn't they be plain English? E.g. "deluxe mapping" not "deluxeMapping"

illwieckz: Yes, they should. 🙂️

src/engine/renderer/tr_shade.cpp
illwieckz: Array is annoying because it requires a length. 😜️

slipher: Are you thinking of std::array? I meant a C array, which doesn't require a length.

illwieckz: Yeah, I was thinking of shiny C++ `std:array`. I added a commit, good old C to the rescue!

src/engine/renderer/tr_shade.cpp
```diff
@@ -117,18 +118,19 @@ static void EnableAvailableFeatures()
...
-		static const std::vector<std::pair<bool*, std::string>> indirectFeatures = {
+		static const std::pair<bool*, std::string> indirectFeatures[] = {
 			{ &glConfig2.shadowMapping, "Shadow mapping" },
 			{ &glConfig2.bloom, "Bloom" },
 			{ &glConfig2.motionBlur, "Motion blur" },
+			{},
 		};
 
-		for ( auto& f : indirectFeatures )
+		for ( auto* f = indirectFeatures; f->first; f++ )
```
slipher, line 128: You can use for-each with C arrays too.

src/engine/renderer/tr_shade.cpp
```diff
@@ -38,25 +38,101 @@ static void EnableAvailableFeatures()
...
+		}
+	}
+
+	glConfig2.deluxeMapping = r_deluxeMapping->integer;
+	glConfig2.normalMapping = r_normalMapping->integer;
+	glConfig2.specularMapping = r_specularMapping->integer;
+	glConfig2.physicalMapping = r_physicalMapping->integer;
+	glConfig2.reliefMapping = r_reliefMapping->integer;
+
+	/* ATI R300 and Intel GMA 3 only have 64 tex ALU instructions, which is not enough for some shader
```
slipher, line 80: Is "tex" a typo here?

src/engine/renderer/tr_shade.cpp
illwieckz: Oh?! Done.

src/engine/renderer/tr_shade.cpp
illwieckz: Fixed.

slipher:
LGTM

PR #1231, opened 2024-08-07
Show model names in /listVBOs command

slipher:
Also don't quit if the model name is bigger than the buffer. It doesn't really matter.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -736,7 +736,7 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 
 	struct VBO_t
 	{
-		char     name[ MAX_QPATH ];
+		char     name[ 96 ]; // only for debugging with /listVBOs
 
 		uint32_t vertexesVBO;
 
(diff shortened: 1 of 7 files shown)


src/engine/renderer/tr_local.h
```diff
@@ -736,7 +736,7 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 
 	struct VBO_t
 	{
-		char     name[ MAX_QPATH ];
+		char     name[ 96 ];
```
DolceTriade, line 739: Why magic number. is there a significance to it? Since it's shared, maybe define a constant?

DolceTriade (approved):

src/engine/renderer/tr_local.h
slipher: I made it bigger because 64 was too small for some names.

> Since it's shared, maybe define a constant?

In principle, I suppose they're not necessarily the same; I could make a VBO with very long names that doesn't correspond to any IBO. Anyway the field is only used for `/listvbos` so it is not that important. I added a comment explaining how the name is only for debugging.

DolceTriade (approved):

PR #1233, opened 2024-08-09
Only generate some images/FBO when they are used

illwieckz:
- Delete some images/FBO that are never used.
- Only generate some images/FBO when they are used.
- Also latch `r_noFog` since the GLSL shader since shader is built conditionally after https://github.com/DaemonEngine/Daemon/pull/1228

### DIFF
--- a/src/engine/renderer/tr_fbo.cpp
+++ b/src/engine/renderer/tr_fbo.cpp
@@ -520,6 +520,7 @@ void R_InitFBOs()
 		R_CheckFBO( tr.portalRenderFBO );
 	}
 
+	if ( glConfig2.bloom )
 	{
 		width = glConfig.vidWidth * 0.25f;
 		height = glConfig.vidHeight * 0.25f;
--- a/src/engine/renderer/tr_fbo.cpp
+++ b/src/engine/renderer/tr_fbo.cpp
@@ -524,21 +524,6 @@ void R_InitFBOs()
 		width = glConfig.vidWidth * 0.25f;
 		height = glConfig.vidHeight * 0.25f;
 
-		tr.downScaleFBO_quarter = R_CreateFBO( "_downScale_quarter", width, height );
-		R_BindFBO( tr.downScaleFBO_quarter );
-
-		R_AttachFBOTexture2D( GL_TEXTURE_2D, tr.downScaleFBOImage_quarter->texnum, 0 );
-		R_CheckFBO( tr.downScaleFBO_quarter );
-
-		tr.downScaleFBO_64x64 = R_CreateFBO( "_downScale_64x64", 64, 64 );
-		R_BindFBO( tr.downScaleFBO_64x64 );
-
-		R_AttachFBOTexture2D( GL_TEXTURE_2D, tr.downScaleFBOImage_64x64->texnum, 0 );
-		R_CheckFBO( tr.downScaleFBO_64x64 );
-
-		width = glConfig.vidWidth * 0.25f;
-		height = glConfig.vidHeight * 0.25f;
-
 		tr.contrastRenderFBO = R_CreateFBO( "_contrastRender", width, height );
 		R_BindFBO( tr.contrastRenderFBO );
 
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -2249,6 +2249,8 @@ R_CreateFogImage
 */
 static void R_CreateFogImage()
 {
+	// Fog image is always created because disabling fog is cheat.
+
 	int   x, y;
 	byte  *data, *ptr;
 	float d;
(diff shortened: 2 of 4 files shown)


slipher:
I'm not really a fan of the push to make every single renderer setting latched. Having to restart all the time is annoying both as a player who wants to experiment with settings, and when debugging. This should only be done if there are significant savings to be had (like not building 100 shader variants, or not building one that causes the shader compiler to error out). Does sky or fog support really impose a significant cost when not used?

illwieckz:
Actually skipping fog GLSL / image to be built / generated when disabled was dumb because disabling fog is a cheat, so no regular player can win anything at disabling it, because they should not be able to disable it. So I reverted that.

For the sky, I really don't see the reason to build code that will not be used  Disabling a feature that is not a cheat to do so has multiple benefits: better performance in game, faster GLSL compilation at game load, and less risk to hit a GLSL compilation failure because of something being unsupported and seeing the engine abort entirely.

I don't find the need to restart on latched cvars change a big deal, and we can make it even less a deal by either implementing `set` alternative command that would restart after setting, or even a special cvar that would automatically restart if at least one latched cvar was changed in a frame. But anyway, this still looks fine to me to restart.

I still have one thought though, those images I disable like `_whiteCube` or `_blackCube` have names, technically they may be usablein a `.shader` file. I don't know if that would make sense to do so and if we should consider that public API or not. If it is considered public API third-party content can rely on, we should always generate them. I found no occurrence of them from 817 `.shader` files from betaserv listing, so it looks like such names were never used in third-party content, meaning it was either never needed to be used, either it was never a public API, either both. So that sounds safe to me to not generate them when unneeded.

src/engine/renderer/tr_image.cpp
```diff
@@ -2390,8 +2395,13 @@ static void R_CreateContrastRenderFBOImage()
...
 }
 
-static void R_CreateBloomRenderFBOImage()
+static void R_CreateBloomRenderFBOImages()
 {
+	if ( !glConfig2.bloom)
+	{
+		return;
+	}
+
```
slipher, line 2404: R_InitFBOs does something with these images unconditionally. Does that not cause errors?

src/engine/renderer/tr_image.cpp
```diff
@@ -2692,8 +2692,13 @@ static void R_CreateShadowCubeFBOImage()
 // *INDENT-ON*
 
 // *INDENT-OFF*
-static void R_CreateBlackCubeImage()
+static void R_CreateReflectionCubeImages()
 {
+	if ( !r_reflectionMapping->integer || !glConfig2.specularMapping )
```
slipher, line 2697: None of the other code checks specularMapping when deciding whether to use reflection mapping.

src/engine/renderer/tr_fbo.cpp
```diff
@@ -507,6 +507,7 @@ void R_InitFBOs()
 		}
 	}
 
+	if ( !r_noportals->integer || r_liquidMapping->integer )
```
slipher, line 510: r_noportals is not latched

src/engine/renderer/tr_image.cpp
```diff
@@ -2731,6 +2731,11 @@ static void R_CreateReflectionCubeImages()
 // *INDENT-OFF*
 static void R_CreateWhiteCubeImage()
 {
+	if ( r_fastsky.Get() )
```
slipher, line 2734: r_fastsky is not latched. And shouldn't be in my opinion... it is not an expensive thing to build

src/engine/renderer/tr_fbo.cpp
slipher: It's also CHEAT, although that doesn't make any sense

src/engine/renderer/tr_image.cpp
illwieckz: In this branch, `R_InitFBOs` only does something with it if `glConfig2.bloom` is true (it adds such test).

src/engine/renderer/tr_image.cpp
illwieckz: This commit was removed, and this function now only generates blackCube which is now expected to be used by the sky code anyway.

src/engine/renderer/tr_image.cpp
illwieckz: `r_fastsky` is currently latched, but this images is like blackCubeImage, we may still want to generate it in all cases. For example it may make possible to use `_whiteCube` in shaders to compose skyboxes. For example one may add some clouds in front of some white or black cube.

So I removed this commit as well.

src/engine/renderer/tr_fbo.cpp
illwieckz: I removed that.

slipher:
LGTM. There's some bad spacing in the bloom commit

illwieckz:
> There's some bad spacing in the bloom commit

GitHub probably messed-up the display, as I can find none.

slipher:
For example `if ( !glConfig2.bloom)`

PR #1234, opened 2024-08-09
renderer: make r_dynamicLight only enable the related renderer, not a kind of light

illwieckz:
renderer: make `r_dynamicLight` only enable a kind of light and `r_realtimeLighting` enable the renderer

With that change,

- `r_realtimeLighting` enables the rendering of dynamic lights
- `r_realtimeLightingRenderer` selects the dynamic light renderer
- `r_dynamicLight` enables the non-static kind of dynamic lights (weapons, particles…)
- `r_staticLight` enables the static kind of dynamic lights (static map lights)

---

_Original post (obsolete):_

Make r_dynamicLight only enable the related renderer, not a kind of light.

With that change,

- r_dynamicLight enables the dynamic light renderer
- r_animatedLight enables the “animated” kind of dynamic lights (weapons, particles…)
- r_staticLight enables the “static” kind of dynamic lights (static map lights)

The animated/static light naming scheme is inspired from the misc_anim_model and misc_model map entities (the later being described as “static” in documentation).

The purpose of that PR is that by disabling r_dynamicLight, the dynamic light renderer is effectively disabled and related GLSL code is not compiled.

Previously, disabling r_dynamicLight but keeping r_staticLight enable was still compiling the GLSL related code, depite no one map uses the feature yet.

The idea is that we would always keep r_animatedLight and r_staticLight enabled, but only enable/disable r_dynamicLight in presets.

This makes possible to not have to add entries for r_staticLight in presets or in game options.

Basically, r_animatedLight and r_staticLight becomes debugging cvars, while r_dynamicLight is kept as a feature cvar.

In the glConfig2 struct,

- dynamicLight is enabled if the engine should run the dynamic light renderer and then allocate related images and compile the related GLSL code. It is enabled if r_dynamicLight cvar enabled and either r_animatedLight or r_staticLight is enabled.
- animatedLight is enabled if the engine should process animated lights. It is enabled if r_dynamicLight and r_animatedLight is enabled.
- staticLight is enabled if the engine should process static lights. It is enabled if r_dynamicLight and r_staticLight is enabled.

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -73,14 +73,15 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_lightScissors;
 	cvar_t      *r_noLightVisCull;
 	cvar_t      *r_noInteractionSort;
-	Cvar::Range<Cvar::Cvar<int>> r_dynamicLightRenderer( "r_dynamicLightRenderer",
-		"renderer for dynamic lights: 0: legacy, 1: tiled", Cvar::NONE,
-		Util::ordinal(dynamicLightRenderer_t::TILED),
-		Util::ordinal(dynamicLightRenderer_t::LEGACY),
-		Util::ordinal(dynamicLightRenderer_t::TILED) );
-	Cvar::Cvar<bool> r_dynamicLight( "r_dynamicLight", "enable dynamic lights", Cvar::NONE, true );
-	Cvar::Cvar<bool> r_staticLight( "r_staticLight", "enable static lights", Cvar::NONE, true );
-	cvar_t      *r_dynamicLightCastShadows;
+	Cvar::Range<Cvar::Cvar<int>> r_realtimeLightingRenderer( "r_realtimeLightingRenderer",
+		"renderer for real time lights: 0: legacy, 1: tiled", Cvar::NONE,
+		Util::ordinal(realtimeLightingRenderer_t::TILED),
+		Util::ordinal(realtimeLightingRenderer_t::LEGACY),
+		Util::ordinal(realtimeLightingRenderer_t::TILED) );
+	Cvar::Cvar<bool> r_realtimeLighting( "r_realtimeLighting", "enable realtime light rendering", Cvar::NONE, true );
+	Cvar::Cvar<bool> r_dynamicLight( "r_dynamicLight", "enable dynamic lights (if realtime lighting is enabled)", Cvar::NONE, true );
+	Cvar::Cvar<bool> r_staticLight( "r_staticLight", "enable BSP entity lights (if realtime lighting is enabled)", Cvar::NONE, true );
+	cvar_t      *r_realtimeLightingCastShadows;
 	cvar_t      *r_precomputedLighting;
 	Cvar::Cvar<int> r_mapOverBrightBits("r_mapOverBrightBits", "default map light color shift", Cvar::NONE, 2);
 	Cvar::Cvar<bool> r_forceLegacyOverBrightClamping("r_forceLegacyOverBrightClamping", "clamp over bright of legacy maps (enable multiplied color clamping and normalization)", Cvar::NONE, false);
@@ -1075,7 +1076,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		r_customwidth = Cvar_Get( "r_customwidth", "1600", CVAR_LATCH | CVAR_ARCHIVE );
 		r_customheight = Cvar_Get( "r_customheight", "1024", CVAR_LATCH | CVAR_ARCHIVE );
 		r_subdivisions = Cvar_Get( "r_subdivisions", "4", CVAR_LATCH );
-		r_dynamicLightCastShadows = Cvar_Get( "r_dynamicLightCastShadows", "1", 0 );
+		r_realtimeLightingCastShadows = Cvar_Get( "r_realtimeLightingCastShadows", "1", 0 );
 		r_precomputedLighting = Cvar_Get( "r_precomputedLighting", "1", CVAR_CHEAT | CVAR_LATCH );
 		Cvar::Latch( r_mapOverBrightBits );
 		Cvar::Latch( r_forceLegacyOverBrightClamping );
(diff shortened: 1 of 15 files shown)


src/engine/renderer/tr_shade.cpp
```diff
@@ -36,29 +36,33 @@ This file deals with applying shaders to surface data in the tess struct.
...
 			}
 		}
 	}
 
+	// Enable different kinds of dynamic lights.
+	glConfig2.animatedLight = glConfig2.dynamicLight && r_animatedLight.Get();
+	glConfig2.staticLight = glConfig2.dynamicLight && r_staticLight.Get();
+
+	// Finally tell the engine the dynamic light code is needed, whatever the kind of light.
+	glConfig2.dynamicLight = glConfig2.animatedLight || glConfig2.staticLight;
```
slipher, line 64: I would prefer taking this value directly from r_dynamicLights without checking other cvars. Running the dynamic light code path with 0 lights is a valid debugging use case. Also this would let the debugging cvar for dropping dynamic lights be non-latch which really improves the velocity of debugging.

src/engine/renderer/tr_init.cpp
```diff
@@ -74,7 +74,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 		Util::ordinal(dynamicLightRenderer_t::LEGACY),
 		Util::ordinal(dynamicLightRenderer_t::TILED) );
 	Cvar::Cvar<bool> r_dynamicLight( "r_dynamicLight", "enable dynamic lights", Cvar::NONE, true );
-	Cvar::Cvar<bool> r_staticLight( "r_staticLight", "enable static lights", Cvar::NONE, true );
+	Cvar::Cvar<bool> r_animatedLight( "r_animatedLight", "enable animated lights (if dynamic light is enabled)", Cvar::NONE, true );
```
slipher, line 77: I find the term "animated light" too random as it does not make any mention of "dynamic" lights. How about making them negative cvars: `r_ignoreDynamicLights` and `r_ignoreLightEntities`? "ignore dynamic lights (even if dynamic light renderer is enabled", "ignore map light entities (even if dynamic light renderer is enabled)"? (Also taking the opportunity to give a better name than 'static light').

illwieckz: The “animated” word was simply chosen for consistency with “misc_anim_model” (which is about a non-static model).

I agree the lack of mention of  “dynamic” is confusing, for now I did that to not have to rewrite all the `r_staticLight` stuff, but I can do it. I personally would prefer `r_dynamicAnimatedLights` and `r_dynamicStaticLights`, and if we find a better word than “animated”, I prefer the generic `r_dynamic<Kind>Lights` naming scheme anyway.

illwieckz: I meant there are two kind of models: _static_ models to be baked in bsp, and _anim_ models to be loaded in game, so since bsp lights are named _static_, I thought it would not be bad to also name non-bsp lights _anim_, just for consistency of the wording.

The `misc_anim_model` entity kind is not necessarily for using actual animations, it's for any model that is not baked in bsp. If your model needs to be animated, it should be a `misc_anim_model`, if your model needs to move, it better be a `misc_anim_model` (unless you statically bake it into a bsp model that is itself moving), if you want your model to be updated when the game update the model or based on some in-game options (like providing a more detailed model), you want to use `misc_anim_model` even if it is not animated or does not move (a good example are the outdoor decorative turrets in parpax).

On the contrary the static `misc_model` entity is just a kind of prefab that contributes to the bsp (the reason why people may prefer it over `misc_anim_model` is that it then allows to compute lightmaps on them).

Instead of “animatedLights” I was also thinking about “entityLights”, but the BSP is the world entity so it's confusing as well.

slipher: With that knowledge, if I had to draw an analogy between types of models and lights, I would say the models of players and buildables correspond to dynamic lights, `misc_anim_model` to "static" lights, and `misc_model` to lights baked in lightmaps.

> Instead of “animatedLights” I was also thinking about “entityLights”,

That's funny since I suggested `r_ignoreLightEntities` to replace `r_staticlight`. The static lights can only come from BSP entities, while I believe the cgame can insert dynamic lights any way it wants, independently of any BSP or gamelogic entities.

illwieckz: > That's funny since I suggested `r_ignoreLightEntities` to replace `r_staticlight`.

Yeah, that's because you said that I shared my other thought about “entityLights”, I wanted to share the fun! 🤣️

Confusion, confusion everywhere

illwieckz: Alternative naming can be:

- `cvar bool r_dynamicLight`:  moving lights like particles and weapon projectiles,
- `cvar bool r_staticLight`: map lights (we don't use yet),
- `cvar bool r_realtimeLighting`: enables the rendering of them (currently named `r_dynamicLight`)
- `cvar int r_realtimeLightingRenderer`: select the renderer (currently named `r_dynamicLightRenderer`)

slipher: I like it

illwieckz: I did that.

src/engine/renderer/tr_init.cpp
```diff
@@ -73,14 +73,15 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
-	Cvar::Cvar<bool> r_staticLight( "r_staticLight", "enable static lights", Cvar::NONE, true );
-	cvar_t      *r_dynamicLightCastShadows;
+	Cvar::Range<Cvar::Cvar<int>> r_realtimeLightingRenderer( "r_realtimeLightingRenderer",
+		"renderer for real time lights: 0: legacy, 1: tiled", Cvar::NONE,
+		Util::ordinal(realtimeLightingRenderer_t::TILED),
+		Util::ordinal(realtimeLightingRenderer_t::LEGACY),
+		Util::ordinal(realtimeLightingRenderer_t::TILED) );
+	Cvar::Cvar<bool> r_realtimeLighting( "r_realtimeLighting", "enable realtime light rendering", Cvar::NONE, true );
+	Cvar::Cvar<bool> r_dynamicLight( "r_dynamicLight", "enable dynamic lights (if realtime lighting is enabled)", Cvar::NONE, true );
+	Cvar::Cvar<bool> r_staticLight( "r_staticLight", "enable static lights (if realtime lighting is enabled)", Cvar::NONE, true );
```
slipher, line 83: Could use a more descriptive description like "enable BSP entity lights"

slipher (commented):
LGTM with 1 comment

src/engine/renderer/tr_init.cpp

PR #1235, opened 2024-08-09
renderer: detect if hardware cannot blend RGBA16 framebuffers

### LINKED ISSUE #1232: Broken GL_RGBA16 image format on ATI R300
With `r_highPrecisionRendering off`:

unvanquished_2024-08-09_003040_000

With `r_highPrecisionRendering on` (default):

unvanquished_2024-08-09_003028_000

unvanquished_2024-08-08_231920_000

unvanquished_2024-08-08_231940_000

unvanquished_2024-08-07_175313_000

This was added in d2b9a688bc7bc7d238281a4ae6630207007772df from #1050 and slightly modified in #1164:

- https://github.com/DaemonEngine/Daemon/pull/1050
- https://github.com/DaemonEngine/Daemon/pull/1164

The usage of `GL_RGBA16` is meant to reduce color banding when doing full overbright bit computation, and is fully optional. Disabling the `r_highPrecisionRendering` cvar is only expected to increase color banding and then reduce quality, but doesn't disable any feature.

I find it very hard to know more about `GL_RGBA16` support, it looks to be very ancient and probably expected to be ubiquitous. For example someone in that old forum thread is saying that it exists since OpenGL 1.1:

https://gamedev.net/forums/topic/517023-gl_rgba16f_arb-vs-gl_rgba16/

I don't know if there is a way to query the support of `GL_RGBA16`. And if there would be, maybe the `r300` Mesa driver is expected to support it, as I don't get any OpenGL error being reported when I run the game on such hardware and driver.

So maybe it's a bug in the Mesa r300 driver.

On our side we can simply disable `r_highPrecisionRendering` on `low` and `lowest` presets, as those cards would better run with such preset, and we better disable that feature for performance on such hardware anyway.

But we also better want to avoid visual glitches if possible, especially since this feature is enabled on default `medium` preset, and we may implement a workaround to disable the feature on R300.

illwieckz:
Use `GL_ARB_internalformat_query` to check if the hardware cannot blend `GL_RGBA16` framebuffers.

Hardware like ATI R300 support RGBA16 framebuffers but can't blend them. The check should detect them.

Unfortunately the Mesa r300 driver has a bug and wrongly reports such blending to be possible despite the hardware not supporting it, but since we already detect R300 hardware we can also workaround that so a workaround was also implemented.

A fix is on the way on Mesa side.

See this thread for details:

- https://gitlab.freedesktop.org/mesa/mesa/-/issues/11669

See also this is work-in-progress merge-request on Mesa side:

- https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/30612

The extension to check for the feature availability is much more recent than the feature itself:

- https://registry.khronos.org/OpenGL/extensions/ARB/ARB_internalformat_query.txt

So if it is not possible to check for the feature availability, the engine assumes the feature is available.

The purpose is to not use the feature if the driver explicitly says the feature is not supported or if we know the feature is not supported. In case the assumption is wrong and `r_highPrecisionRendering` is enabled, the symptom is translucent textures being rendered opaque, as seen in #1232:

- https://github.com/DaemonEngine/Daemon/issues/1232

Fixes #1232.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -875,7 +875,19 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips, image_t 
 		format = GL_DEPTH_STENCIL;
 		internalFormat = GL_DEPTH24_STENCIL8;
 	}
-	else if ( image->bits & ( IF_RGBA16F | IF_RGBA32F | IF_RGBA16 | IF_TWOCOMP16F | IF_TWOCOMP32F | IF_ONECOMP16F | IF_ONECOMP32F ) )
+	else if ( image->bits & IF_RGBA16 )
+	{
+		if ( !glConfig2.textureRGBA16BlendAvailable )
+		{
+			Log::Warn("RGBA16 image '%s' cannot be blended", image->name );
+			internalFormat = GL_RGBA8;
+		}
+		else
+		{
+			internalFormat = GL_RGBA16;
+		}
+	}
+	else if ( image->bits & ( IF_RGBA16F | IF_RGBA32F | IF_TWOCOMP16F | IF_TWOCOMP32F | IF_ONECOMP16F | IF_ONECOMP32F ) )
 	{
 		if( !glConfig2.textureFloatAvailable ) {
 			Log::Warn("floating point image '%s' cannot be loaded", image->name );
@@ -898,10 +910,6 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips, image_t 
 		{
 			internalFormat = glConfig2.textureRGAvailable ?
 			  GL_RG32F : GL_LUMINANCE_ALPHA32F_ARB;
-		}
-		else if ( image->bits & IF_RGBA16 )
-		{
-			internalFormat = GL_RGBA16;
 		}
 		else if ( image->bits & IF_ONECOMP16F )
 		{
@@ -2462,7 +2470,14 @@ static void R_CreateCurrentRenderImage()
 
 	if ( r_highPrecisionRendering.Get() )
 	{
-		imageParams.bits |= IF_RGBA16;
+		if ( !glConfig2.textureRGBA16BlendAvailable )
+		{
+			Log::Warn( "High-precision current render disabled because RGBA16 framebuffer blending is not available" );
+		}
+		else
+		{
+			imageParams.bits |= IF_RGBA16;
+		}
 	}
 
 	imageParams.filterType = filterType_t::FT_NEAREST;
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -1022,6 +1022,20 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 			Log::Notice("%sMissing GPU vertex skinning, models are not hardware-accelerated.", Color::ToString( Color::Red ) );
 		}
 
+		switch ( glConfig2.textureRGBA16BlendAvailable )
+		{
+			case 1:
+				Log::Notice( "%sUsing GL_RGBA16 with GL_FRAMEBUFFER_BLEND.", Color::ToString( Color::Green ) );
+				break;
+			case -1:
+				Log::Notice( "%sUsing GL_RGBA16 with GL_FRAMEBUFFER_BLEND (assumed to be available).", Color::ToString( Color::Yellow ) );
+				break;
+			default:
+			case 0:
+				Log::Notice( "%sMissing GL_RGBA16 with GL_FRAMEBUFFER_BLEND.", Color::ToString( Color::Red ) );
+				break;
+		}
+
 		if ( glConfig.smpActive )
 		{
 			Log::Notice("Using dual processor acceleration." );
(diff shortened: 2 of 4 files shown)


illwieckz:
This also fixes a bug where `GL_RGBA16` textures were wrongly tested in float-related code.

src/engine/renderer/tr_image.cpp
```diff
@@ -874,7 +874,22 @@ void R_UploadImage( const byte **dataArray, int numLayers, int numMips, image_t
 		format = GL_DEPTH_STENCIL;
 		internalFormat = GL_DEPTH24_STENCIL8;
 	}
-	else if ( image->bits & ( IF_RGBA16F | IF_RGBA32F | IF_RGBA16 | IF_TWOCOMP16F | IF_TWOCOMP32F | IF_ONECOMP16F | IF_ONECOMP32F ) )
+	else if ( image->bits & IF_RGBA16 )
+	{
+		/* Assume this is an old driver without the extension but the RGBA16 blending is available,
```
slipher, line 879: The comment seems to be accidentally copy-pasted

src/engine/renderer/tr_init.cpp
```diff
@@ -1015,6 +1015,27 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
...
+				{
+					Log::Notice( "%sUsing GL_RGBA16 with GL_FRAMEBUFFER_BLEND.", Color::ToString( Color::Green ) );
+				}
+				else
+				{
+					Log::Notice( "%sHaving GL_RGBA16 with GL_FRAMEBUFFER_BLEND.", Color::ToString( Color::Green ) );
+				}
+				break;
+			case -1:
+				Log::Notice( "%sAssuming GL_RGBA16 with GL_FRAMEBUFFER_BLEND.", Color::ToString( Color::Yellow ) );
```
slipher, line 1031: This will say "assuming GL_RGBA16" even if the feature is disabled, so nothing will be done with this assumption. Maybe it's better to have a message like 'Explicitly disabling RGBA16' in all cases where the cvar is disabled? Or alternatively, don't make any of the messages depend on the cvar.

ondracka:
BTW might be better to actually require GL_ARB_internalformat_query2. There the GL_FRAMEBUFFER_BLEND is actually defined, while in GL_ARB_internalformat_query this is not required and only mentioned in the QA section...

src/engine/renderer/tr_init.cpp
illwieckz: This GfxInfo command prints a report about what's available to the engine, it is always right that the engine assumes the feature is available or not, even if unused.

The rest of the code already says “Using that extension”, even if related extension is only used in a GLSL code that is disabled. I actually went too verbose with the Using/Having distinction, but I still want to report the extension is assumed.

src/engine/renderer/tr_image.cpp
illwieckz: Removed.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1882,6 +1885,62 @@ static void GLimp_InitExtensions()
...
+		switch( param )
+		{
+			case GL_TRUE:
+				glConfig2.textureRGBA16BlendAvailable = 1;
+				break;
+			case GL_FALSE:
+				glConfig2.textureRGBA16BlendAvailable = 0;
+				break;
+			default:
+				/* Mesa returning GL_FULL_SUPPORT looks to be a bug, it should return GL_TRUE, returning
```
slipher, line 1903: Maybe the change to use internalFormatQuery2 is not worth it if we have to work around another bug?

slipher (commented):
LGTM (with an optional comment)

illwieckz:
A discrepancy in OpenGL documentation was discovered:

> If _pname_ is `GL_FRAMEBUFFER_BLEND`, _params_ is set to `GL_TRUE` to indicate that the internal format is supported for blending operations when attached to a framebuffer, and to `GL_FALSE` otherwise.  
> -- https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGetInternalformat.xhtml

> `FRAMEBUFFER_BLEND`: The support for rendering to the resource via framebuffer attachment when blending is enabled is returned in _params_. Possible values returned are `FULL_SUPPORT`, `CAVEAT_SUPPORT`, or `NONE`. If the resource is unsupported, `NONE` is returned.  
> -- https://registry.khronos.org/OpenGL/extensions/ARB/ARB_internalformat_query2.txt

And and unfortunately `GL_FULL_SUPPORT` (`33463`) is not `GL_TRUE` (`1`).

illwieckz:
The code now tests for `GL_FULL_SUPPORT` or `GL_TRUE` for enablement (in case different drivers based their implementation on this or that document), but only what is not `GL_FULL_SUPPORT` or `GL_TRUE` is considered to be trusted as a feature disablement.

I still believe it's a good idea to use internalFormatQuery2 because it means that once Mesa is fixed the workaround to detect the missing feature will not be used on R300 (the hardware known to not support the feature). With the help of the upcoming workaround branch providing toggles to disable workarounds, the Dæmon engine/Unvanquished game would be more usable as a testbed for OpenGL compatibility and for reporting issues against upstream drivers, something I find very good: it's always good when our software is considered as a reference for benchmarking/testing and used as such.

src/engine/sys/sdl_glimp.cpp
illwieckz: Answered there about the worthness of it:

- https://github.com/DaemonEngine/Daemon/pull/1235#issuecomment-2286471659

PR #1237, opened 2024-08-11
Switch most vertex/fragment libs to use #insert instead

VReaperV:
Switch ones that can be easily changed. Other shaders might require some extra effort to switch, before vertex/fragment libs can be removed completely.

This is part of a change to move all shaders to use #insert instead of vertex/fragment libs, since that would allow to simplify code and because #insert is a more powerful tool.


slipher:
LGTM

PR #1238, opened 2024-08-11
Fix legacy renderer forward shader

VReaperV:
Fixed missing function argument. Also fixed the incorrect handling of standalone height map vs height map-in-normal map in lightMapping shader.


illwieckz (approved):
LGTM.

PR #1239, opened 2024-08-12
Implement GL timer queries

VReaperV:
Add GL timers for major parts of the engine (frame, depth pre-pass, opaque, transparent, post-process, cull and depth reduction)

### DIFF
--- a/src/engine/client/cg_msgdef.h
+++ b/src/engine/client/cg_msgdef.h
@@ -163,6 +163,7 @@ enum cgameImport_t
   // Renderer
   CG_R_SETALTSHADERTOKENS,
   CG_R_GETSHADERNAMEFROMHANDLE,
+  CG_R_GETENGINETIMERS,
   CG_R_SCISSOR_ENABLE,
   CG_R_SCISSOR_SET,
   CG_R_INPVVS,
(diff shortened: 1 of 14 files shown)


src/engine/client/cg_msgdef.h
```diff
@@ -313,6 +314,10 @@ namespace Render {
 		IPC::Message<IPC::Id<VM::QVM, CG_R_GETSHADERNAMEFROMHANDLE>, int>,
 		IPC::Reply<std::string>
 	>;
+	using GetEngineTimersMsg = IPC::SyncMessage<
+		IPC::Message<IPC::Id<VM::QVM, CG_R_GETENGINETIMERS>>,
+		IPC::Reply<EngineTimers*>
```
slipher, line 319: Sending a pointer through inter-process communication doesn't make sense.

src/engine/renderer/TimerQuery.cpp
```diff
@@ -0,0 +1,161 @@
...
+		return;
+	}
+
+	glQueryCounter( id2, GL_TIMESTAMP );
+	inProgress = false;
+	useBeginEnd = true;
+}
+
+// This will retrieve the time in nanoseconds if this query was issued before, otherwise 0
+// This will block the CPU until the query becomes available
```
slipher, line 93: Are there any cases where it can really block? If so I guess we should avoid updating the timers when they are not being used.

slipher (commented):
Cool, I didn't know there is a way to time these things.

What's the plan for how to use the cgame API? If you add a console command for printing the timers, the timer code could be merged to master and only any API parts need to go in for-0.55.

src/engine/client/cg_msgdef.h
VReaperV: Yeah, I think it can just be made shared instead.

src/engine/renderer/TimerQuery.cpp
VReaperV: Yea, the plan is to have a cvar to enable/disable timers completely. Hence why the pr is draft, here just so it doesn't get lost.

PR #1240, opened 2024-08-13
NUKE vertex/fragment shader libs

VReaperV:
Change the remaining vertex/fragment lib usages to use #insert instead and nuke the libs themselves.


illwieckz (approved):
LGTM.

PR #1247, opened 2024-08-18
Suppress a warning with -connect when no argument are provided

necessarily-equal:
Some context: the flatpack uses a patch to suppress the warning, this commit upstreams the patch.

Slipher seems to be ok with this patch: https://github.com/Unvanquished/updater/issues/95#issuecomment-2223449484.

This patch is currently [added manually in the flatpack](https://github.com/flathub/net.unvanquished.Unvanquished/blob/0217843/0001-Suppress-a-warning.patch), and can be dropped after this is merged.

### DIFF
--- a/src/engine/framework/System.cpp
+++ b/src/engine/framework/System.cpp
@@ -456,7 +456,7 @@ static void ParseCmdline(int argc, char** argv, cmdlineArgs_t& cmdlineArgs)
 		// Any arguments after the URI are discarded.
 		if (Str::IsIEqual("-connect", argv[i])) {
 			if (argc < i + 2) {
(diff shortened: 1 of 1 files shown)


DolceTriade:
I don't understand how this suppresses the warning? Is the default level here warn?

necessarily-equal:
Well, it doesn't make the message disappear, but at least it doesn't make it look like a bug

DolceTriade:
I see. I would rephrase the commit message to say "demote -connect with no server to notice instead of warning" or something. LGTM

DolceTriade (approved):

slipher:
LGTM

PR #1251, opened 2024-08-19
Autosprite-related cleanup

### LINKED ISSUE #730: Renderer: broken sprite code path

slipher:
Closes #730. This doesn't actually fix any bugs (except warning spam), but there are other, better issues which track the bugs.


VReaperV (approved):
LGTM

PR #1252, opened 2024-08-19
Add r_glslCache to toggle GLSL binary caching

slipher:


src/engine/renderer/gl_shader.cpp
```diff
@@ -34,6 +34,12 @@ static_assert(std::is_pod<GLBinaryHeader>::value, "Value must be a pod while cod
 static Cvar::Cvar<std::string> shaderpath(
 	"shaderpath", "path to load GLSL source files at runtime", Cvar::INIT | Cvar::TEMPORARY, "");
 
+// Note that many vendors implement system-level GLSL caches, so it may well be more efficient to
+// turn this off. As of 2024 I believe AMD and Nvidia proprietary drivers usually do this, while
+// Intel does not.
```
illwieckz, line 39: I guess most if not all Mesa drivers implement such cache too.

illwieckz (approved):
LGTM.

src/engine/renderer/gl_shader.cpp
DolceTriade: Sounds like it's better disabled then? When we added this feature a decade ago, this came with a significant decrease in loading time. We should check (independently of this PR) if that's still the case.

illwieckz: Having it enabled by default doesn't look bad to me, in worst case our cache may be a bit slower than driver's one because of our implementation being possibly more naïve (while driver developer will have more internal knowledge about the driver and do dark magic we can't do), but our cache will always be as fast as it used to be, which is not slow.

The various caches are usually pilling over each others, if one deletes his `~/.local/share/unvanquished/glsl` folder, next compilation may still get a cached file from `~/.cache/mesa_shader_cache` and fill `~/.local/share/unvanquished/glsl`  without compiling anything. The driver's cache is especially useful when switching branches, this may invalidate our own cache (we only keep one copy of every shader), but the driver's cache may still provide some already know files (the driver's cache have no idea about what is a copy of a shader from engine point of view, then caches everything).

The risk with disabling the cache is that if by lack of luck or by mistake there is no driver cache, things would be very slow. I believe that our cache looks fast enough to be enabled by default.

slipher: I just tested (instead of relying on vague memories) on one of my machines and actually, neither Nvidia nor Intel drivers show any sign of caching. So I'm going to NUKE this comment, unless someone else wants to produce numbers for their driver demonstrating presence of caching.

PR #1253, opened 2024-08-19
Fix 0-size mip-levels in depth-reduction image

### LINKED ISSUE #1248: Material system: GPU occlusion culling doesn't work if `log2( r_customWidth )` is not equal to `log2( r_customHeight )`
If I have material system and `r_gpuOcclusionCulling` enabled, and set the resolution to anything significantly less than my screen resolution, I get something like this. Many or most surfaces are not drawn that should be.
unvanquished_2024-08-18_202617_000

VReaperV:
Fixes #1248.


slipher:
LGTM

PR #1256, opened 2024-08-23
Debug drawing fixes

slipher:


src/engine/renderer/tr_bsp.cpp
```diff
@@ -6285,7 +6280,7 @@ static const int HASHTABLE_SIZE = 7919; // 32749 // 2039    /* prime, use % */
 #define HASH_USE_EPSILON
 
 #ifdef HASH_USE_EPSILON
-#define HASH_XYZ_EPSILON                 0.01f
+#define HASH_XYZ_EPSILON 100.0f
```
VReaperV, line 6283: What is this change for?

VReaperV (commented):
What exactly do decals have to with debug drawing?

src/engine/renderer/tr_bsp.cpp
slipher: The cubemap search function can only find probes that are in the same epsilon X epsilon X epsilon-sized voxel. 0.01 is unreasonably small so I assume it was inverted by mistake. 100 is still too small for it to find one the majority of the time though. Overall the code really doesn't make any sense, since the probes are generated in an irregularly spaced way according to BSP nodes, but the lookup code seemingly expects them to be evenly spaced like the light grid, or at least placed with a minimal density.

slipher:
> What exactly do decals have to do with debug drawing?

I'm refactoring/fixing debug draw stuff and one of the debug draw options was "show decal projectors". It turns out to be dead code.

src/engine/renderer/tr_bsp.cpp
VReaperV: Gonna need to look at how it places the probes at some point, but since reflections don't fully work right now this change should be fine then.

VReaperV: I've created an issue to track the above here: https://github.com/DaemonEngine/Daemon/issues/1257.

VReaperV (approved):
LGTM

PR #1262, opened 2024-08-25
Fix videoMap regression

slipher:
Make the in-game videos show up again. Regression in f960177aba3c68e17e8ebfc865e8e8682268667a related to bindless texture changes.

Videos still don't work if bindless textures are enabled.


VReaperV (approved):
LGTM

slipher:
Added another commit to fix videos with r_smp.

VReaperV:
Tested the above commit, it fixes the issue with `r_smp 1` for me.

PR #1265, opened 2024-08-25
Some improvements to r_glDebugProfile/r_glDebugMode

VReaperV:
Use the core extension, which is more widely supported, switch the cvars to new-style. Also removed the cheat flag as it made no sense there.


slipher:
LGTM

PR #1267, opened 2024-08-27
Fix stage sync

VReaperV:
Fixes a regression introduced in #1204.


VReaperV:
> I don't understand it but LGTM.

`previousMaterialID` is required to ensure proper order of drawcalls w.r.t. consecutive surface stages. And it has to not be reset after processing a stage for that.

PR #1268, opened 2024-08-28
Material system: load GLSL extension needed for Intel

slipher:
This gets the material system running on Intel drivers. In a cursory test, I didn't see any glaring bugs.

I searched for "intel glsl wrong line number" and by chance found a [thread](https://github.com/RPCS3/rpcs3/issues/12285) about this extension.


slipher:
Actually it doesn't help. I must have made some mistake when testing.

PR #1269, opened 2024-08-28
Deduplicate Material renderer function selection

illwieckz:
That part of the Material code seems to be stable enough to start mutualizing some logic with the non-Material code.

The various `ProcessMaterial*`, `BindShader*` and `UpdateSurfaceData*` functions are selected at the same time the `Render_*` functions are, at q3shader parsing time.

This deduplicates many `switch` boilerplate and reduce risks to introduce discrepancies in the future when modifying some parts of the code and forgetting to report the same change everywhere else.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1096,6 +1096,7 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 	struct drawSurf_t;
 
 	using stageRenderer_t = void(*)(shaderStage_t *);
+	using stageSurfaceDataUpdater_t = void(*)(uint32_t*, Material&, drawSurf_t*, const uint32_t);
 	using stageShaderBinder_t = void(*)(Material*);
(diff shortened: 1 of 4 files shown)


src/engine/renderer/tr_local.h
```diff
@@ -1108,6 +1113,10 @@ enum class dynamicLightRenderer_t { LEGACY, TILED };
 		bool shaderHasNoLight;
 
 		stageRenderer_t colorRenderer;
+		stageSurfaceDataUpdater_t surfaceDataUpdater;
```
slipher, line 1116: The material renderer / non-material renderer/ legacy dynamic light rendering bits are all jumbled together here. Could use some comments to clarify which is which

illwieckz: Yes, I wanted to know how to name the non-material renderer before adding some comments, now done. 😉️

VReaperV (approved):
LGTM.

PR #1270, opened 2024-08-28
Mutualize SetLightDeluxeMode, SetLightMap, and SetDeluxeMap

illwieckz:
Another effort to mutualize code between the non-Material renderer and the Material one.

- Mutualize `SetLightDeluxeMode`, `SetLightMap`, and `SetDeluxeMap`.
- Rewrite `SetRgbaGen` as `SetRgbGen` and `SetAlphaGen`.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -24,6 +24,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "tr_local.h"
 #include "gl_shader.h"
 #include "Material.h"
+#include "ShadeCommon.h"
 
 /*
 =================================================================================
@@ -859,8 +860,6 @@ static void Render_generic2D( shaderStage_t *pStage )
 	GL_CheckErrors();
 }
 
-static image_t* GetLightMap();
-
 void Render_generic3D( shaderStage_t *pStage )
 {
 	GLimp_LogComment( "--- Render_generic3D ---\n" );
@@ -935,7 +934,7 @@ void Render_generic3D( shaderStage_t *pStage )
 	if ( pStage->type == stageType_t::ST_STYLELIGHTMAP )
 	{
 		gl_genericShader->SetUniform_ColorMapBindless(
-			GL_BindToTMU( 0, GetLightMap() )
+			GL_BindToTMU( 0, GetLightMap( &tess ) )
 		);
 	}
 	else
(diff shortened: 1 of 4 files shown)


VReaperV:
Yeah, merging the code between material/non-material would be great. I've thought about merging the `UpdateSurfaceData*` functions with `Render_*` in `tr_shade`, though it would be a bit cumbersome with a bunch of type casts and if checks scattered around.

Why split rgbgen and alphagen though? Neither of those seems to be used individually.

illwieckz:
> Why split rgbgen and alphagen though? Neither of those seems to be used individually.

It just made it easier to do `something = Func()`, less lines of code, and no confusion about things being used while not being set yet. This part is just bikeshedding 

> I've thought about merging the `UpdateSurfaceData*` functions with `Render_*` in `tr_shade`

Yeah, I start with easier things, but in the end I want to make them share more code. Next step would be to make them share the Shader permutation selection and binding.

src/engine/renderer/MaterialCommon.h
```diff
@@ -0,0 +1,242 @@
...
+
+		case deluxeMode_t::GRID:
+			return tr.lightGrid2Image;
+			break;
+
+		default:
+			break;
+	}
+
+	return tr.blackImage;
```
VReaperV, line 197: This defaulted to `tr.whiteImage` in the original code, and might not work correctly otherwise.

slipher:
I don't like the file name. I expect MaterialBlahBlah.h to be something only used for the material system

illwieckz:
> I don't like the file name. I expect MaterialBlahBlah.h to be something only used for the material system

`ShadeCommon`, with `shade` like `tr_shade.cpp`?

(and not `ShaderCommon` which would be confusing with `tr_shader.cpp` which is about q3shaders)

src/engine/renderer/MaterialCommon.h
VReaperV: It seems the difference between 0 and 1 would be that lightDir would be either (1.0, 1.0, 1.0) or (-1.0, -1.0, -1.0), but I haven't checked further if it makes any difference.

slipher: If this is supposed to be a direction then there is this:
```image_t    *flatImage; // use this as default normalmap```

Is this even used though? I would imagine if we have `deluxeMode_t::NONE` then the deluxe-using shaders should just be disabled.

**Comment by slipher** [comment]
> `ShadeCommon`, with `shade` like `tr_shade.cpp`?

All right

File: src/engine/renderer/MaterialCommon.h (hunk 1 above)
Comment (illwieckz) [reply src/engine/renderer/MaterialCommon.h]: The normal map flat image is probably no better at is is `{0.5, 0.5, 1.0}` and here what we may look for may be `{0.5, 0.5, 0.5}`. Anyway, the image is likely not used at all when the feature is not available, it is likely set to avoid handling a null pointer.

**Comment by illwieckz** [comment]
> > `ShadeCommon`, with `shade` like `tr_shade.cpp`?
> 
> All right

Done.

File: src/engine/renderer/tr_shade.cpp [hunk 2]
```diff
@@ -975,40 +947,6 @@ void Render_generic( shaderStage_t *pStage )
...
-*/
-static image_t* GetDeluxeMap()
-{
-	if ( static_cast<size_t>( tess.lightmapNum ) < tr.deluxemaps.size() )
-	{
-		return tr.deluxemaps[ tess.lightmapNum ];
-	}
-	else
-	{
-		return tr.blackImage;
```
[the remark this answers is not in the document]
illwieckz: @VReaperV It means the default white was never used before, as the white was always overwritten to black when there was no deluxe mapping.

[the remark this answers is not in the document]
VReaperV: `tr.blackImage` is fine then.

VReaperV (approved):
LGTM.

PR #1272, opened 2024-08-29
Add sv_novis to disable server PVS checks

slipher:
Disables PVS checking when transmitting entities to clients.

sv_novis was documented at
https://wiki.unvanquished.net/wiki/Testing/Maps, but it turns out such a cvar has never existed. I wanted to use it, so I made it real.


illwieckz (approved):
LGTM.

PR #1274, opened 2024-08-30
Renderer debugging: fix r_debugSort; filter r_glDebugProfile messages

slipher:
- make r_debugSort work instead of crashing
- severity-level filtering for GL debug messages

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -3143,26 +3143,26 @@ void Tess_End()
 	}
 
 	// for debugging of sort order issues, stop rendering after a given sort value
-	if ( r_debugSort->integer && r_debugSort->integer < tess.surfaceShader->sort )
+	bool skip = r_debugSort->integer && tess.surfaceShader != nullptr
+		&& r_debugSort->integer < tess.surfaceShader->sort;
+	if ( !skip )
 	{
-		return;
-	}
+		// update performance counter
+		backEnd.pc.c_batches++;
 
-	// update performance counter
-	backEnd.pc.c_batches++;
+		GL_CheckErrors();
 
-	GL_CheckErrors();
+		// call off to shader specific tess end function
+		tess.stageIteratorFunc();
 
-	// call off to shader specific tess end function
-	tess.stageIteratorFunc();
-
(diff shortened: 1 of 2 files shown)


src/engine/renderer/tr_shade.cpp
```diff
@@ -3143,26 +3143,26 @@ void Tess_End()
 	}
```
DolceTriade, line 3143: why is it safe to return early here but now below?

src/engine/renderer/tr_shade.cpp
```diff
@@ -3143,26 +3143,26 @@ void Tess_End()
...
 
 	// for debugging of sort order issues, stop rendering after a given sort value
-	if ( r_debugSort->integer && r_debugSort->integer < tess.surfaceShader->sort )
+	bool skip = r_debugSort->integer && tess.surfaceShader != nullptr
```
DolceTriade, line 3146: this nested indentation is really ugly. I'm almost wondering if it's worth it to split out the cleanup stuff below into its own function :-/

src/engine/renderer/tr_shade.cpp
slipher: The most important thing is to zero out the counters: numIndexes etc.

src/engine/renderer/tr_shade.cpp
slipher: In an upcoming commit I move the last part to a function like this:
```
void Tess_Clear()
{
	tess.vboVertexSkinning = false;
	tess.vboVertexAnimation = false;

	// clear shader so we can tell we don't have any unclosed surfaces
	tess.multiDrawPrimitives = 0;
	tess.numIndexes = 0;
	tess.numVertexes = 0;

	// This is important after doing CPU-only tessellation with Tess_MapVBOs( true ).
	// A lot of code relies on a behavior of Tess_Begin: automatically map the
	// default VBO *if tess.verts is null*.
	tess.verts = nullptr;
	tess.indexes = nullptr;
}
```

Is that what you mean?

DolceTriade: Yes.

DolceTriade (approved):

PR #1277, opened 2024-09-01
Clear portalStack on MaterialSystem::Free()

### LINKED ISSUE #1271: Material system: leftover portal data after level change
If I load the map `test-portals` and then load `parpax` (which doesn't have any portals) after, I get a log of `Bad surface tesselated.` warnings. It comes from here https://github.com/DaemonEngine/Daemon/blob/7ba945cb1938967f1c4aba77d02638c2eb5364d4/src/engine/renderer/tr_main.cpp#L2134

VReaperV:
Fixes #1271


slipher:
LGTM

Issue #1259, opened 2024-08-24
Material system bugs on map "defenxe"


slipher:
This screenshot shows the two bugs described below:

![unvanquished_2024-08-24_074127_000](https://github.com/user-attachments/assets/aef0188b-7ad4-47ce-9a02-fd7b5a3c410e)

This one shows things working correctly (I deconstructed the OM):
![unvanquished_2024-08-24_074102_000](https://github.com/user-attachments/assets/094d2341-e9a7-47ef-81cf-215bbf527d38)

### 1. Transparent surface sometimes not rendered

The Overmind is supposed to be completely surrounded by glass. The rendering is not always wrong; whether it is broken seems to be determined when the renderer loads the map. This sounds silly, but it appears to depend on whether the OM is there during renderer load. I can decon the OM and vid_restart and the bug goes away, then I reload the initial layout and vid_restart and it comes back.

The glass uses this texture from `scripts/nexus6.shader` in `res-tremulous`:
```
textures/nexus6/glass_02_s
{
	qer_editorimage textures/niveus/scratchglass
	qer_trans 0.5
	surfaceparm nomarks
	surfaceparm trans
	{
		map textures/niveus/scratchglass
		blendfunc add
		rgbGen const ( 0.1 0.1 0.1 )
	}
	{
		map $lightmap
		blendfunc gl_dst_color gl_src_alpha
		rgbGen identity
		tcGen lightmap
	}
}
```

### 2. Surface under the reactor is never rendered when GPU occlusion culling is enabled

If `r_gpuOcclusionCulling` is enabled, you can see through the floor under the reactor to the human default base below. I was using the resolution 1920x1080 in case that matters.

### Other comments
Also the trees, which use `deformvertexes autosprite2`, aren't rendered correctly, but I am planning to fix that soon myself.

Something I have noticed with this map is that the number of materials or surfaces generated is not consistent. Sometimes it changes after a `vid_restart` even though I didn't change any settings, e.g.
```
Generated 12 materials from 973 surfaces
Generated 12 materials from 966 surfaces
```

Maybe some shaders are not requested until after the game starts and that is messing things up somehow?

VReaperV:
Surfaces missing might because `R_AddWorldSurfaces()` is not adding a few surfaces for some reason... I thought I fixed all causes of that, but I might've missed some. It would also explain the missing glass.

Do you have a link to a dpk with this map?

slipher:
[...]

PR #1279, opened 2024-09-06
Fix material system missing surfaces bug

slipher:
Fixes bug (1) described on https://github.com/DaemonEngine/Daemon/issues/1259: in many maps some of the surfaces are at times absent from the material system's buffer of world surfaces.

Plus a couple of small cleanup commits.


VReaperV (approved):
LGTM.

PR #1280, opened 2024-09-06
system: use getrandom() and fallback on urandom if syscall doesn't work

illwieckz:
Sometime with less common compilers I get an error about `getrandom` syscall failures, but we already have an alternate implementation. It happens that the checks is currently done at build time, but then even if the headers are there, it may fail at run time. So we technically can try the syscall then the urandom read if the syscall fails.

What I don't know though, is that if doing that is reducing security or not.

### DIFF
--- a/src/common/System.cpp
+++ b/src/common/System.cpp
@@ -36,12 +36,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <fcntl.h>
 #include <signal.h>
 #ifdef __linux__
-#include <linux/version.h>
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3,17,0)
-#include <sys/syscall.h>
-#include <linux/random.h>
-#define HAS_GETRANDOM_SYSCALL 1
-#endif
+#include <sys/random.h>
 #endif
 #ifdef __native_client__
 #include <nacl/nacl_exception.h>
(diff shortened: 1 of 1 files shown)


slipher:
> Sometime with less common compilers I get an error about `getrandom` syscall failures

Why would it change depending on the compiler?

illwieckz:
I have no idea, I get this kind of error for example when:

- I build with Zig
- I build with ICC on Ubuntu Noble and run on Ubuntu Noble

Even more funny: building with ICC from Ubuntu 24.04 Noble (bind mount on the compiler folder) in Ubuntu 22.10 Kinetic chroot over Ubuntu Noble and running on Ubuntu Noble doesn't reproduce the error…

I guess some libc mismatch or so may happen…

illwieckz:
I noticed that if I do `getrandom(dest, size, GRND_NONBLOCK)` instead of `syscall(SYS_getrandom, dest, size, GRND_NONBLOCK)`, it works on ICC and Zig too. And `getrandom` is expected to be the same syscall. So I guess the error was due to a weird library issue.

Anyway, I investigated more and the `getrandom` syscall also uses `/dev/urandom`, the only difference with reading `/dev/urandom` directly is that the `getrandom` syscall with `GRND_NONBLOCK` is not blocking. So there should not be any security issue at reading `/dev/urandom` as a fallback.

I also don't know what this randomness is for, but such fallback can't harm.

For reference here is the PR that added the feature:

- https://github.com/DaemonEngine/Daemon/pull/4

Discussion happened there:

- https://github.com/Unvanquished/Unvanquished/pull/963

Actually, the only expected situation where the syscall would not work is that if the kernel is 10 years old, which is unlikely, but in case we face again a weird bug like the one I faced, it looks harmless to keep the fallback.

src/common/System.cpp
```diff
@@ -410,19 +405,20 @@ void GenRandomBytes(void* dest, size_t size)
...
-	if (syscall(SYS_getrandom, dest, size, GRND_NONBLOCK) == -1)
-		Sys::Error("Failed getrandom syscall: %s", strerror(errno));
 #elif defined(__linux__)
-	int fd = open("/dev/urandom", O_RDONLY);
-	if (fd == -1)
-		Sys::Error("Failed to open /dev/urandom: %s", strerror(errno));
-	if (read(fd, dest, size) != (ssize_t) size)
-		Sys::Error("Failed to read from /dev/urandom: %s", strerror(errno));
-	close(fd);
+	if (getrandom(dest, size, GRND_NONBLOCK) == -1)
```
slipher, line 409: It's nothing related to your change, but this `== -1` should be changed to `== size`. According to the man page it may return some number of bytes less than requested

illwieckz: Right, this is fixed.

Also I noticed my code was stupid because if the random generator isn't ready yet, we would block on the fallback reading `urandom`, which completely defeats the purpose of using `getrandom` with `GRND_NONBLOCK`. So I rewrote the code to only use the fallback if `errno` is equal to `ENOSYS` (syscall not implemented).

illwieckz:
Just for the knowledge, the error I was getting before with `syscall(SYS_getrandom, …)` was “Invalid argument”, so `EINVAL`.

src/common/System.cpp
illwieckz: I also added a warning for when the syscall is not available, just before doing the fallback. All other errors are calling `Sys::Error`.

slipher: Using nonblock never made any sense because we can't proceed without it. The flag can be removed

slipher: Well maybe it makes sense if the idea is that we don't block indefinitely with no sign of what's happening (and instead print a warning before blocking). But if we keep nonblock I say we keep the fallback.

src/common/System.cpp
```diff
@@ -410,19 +405,32 @@ void GenRandomBytes(void* dest, size_t size)
...
+			close(fd);
+		}
+		else
+		{
+			Sys::Error("Failed getrandom syscall: %s", strerror(errno));
+		}
+	}
+	else if (ret != size)
+	{
+		Sys::Error("Failed getrandom syscall: %s", strerror(EAGAIN));
```
slipher, line 430: Why `strerror(EAGAIN)`?

illwieckz: If I understand the man page correctly, when `getrandom()` returns less bytes than requested, when `GRND_NONBLOCK` is not used, it fails with `EAGAIN`. But when using `GRND_NONBLOCK`, it may still return less bytes than requested without failing, in that case we better act like if it failed with `EAGAIN`, considering the `EAGAIN` error to always be considered when `getrandom()` returns less than requested.

slipher: OK. But I would rather see a unique message at each location so that if we have a log file we can tell which operation actually failed. So for this one you could put for example `"getrandom returned insufficient data"`. And instead of `"Failed getrandom syscall: %s", strerror(ENOSYS)`, `"getrandom syscall not supported by kernel"`.

illwieckz: Done.

slipher:
LGTM

PR #1283, opened 2024-09-08
Fix lightMapping_fp compile error

VReaperV:
Use the correct cvars and initializer that is supported on earlier GLSL versions.


illwieckz:
Fixup for #1234:

- https://github.com/DaemonEngine/Daemon/pull/1234

I had to fix some of those after the rebase because of the upstream having diverged from the original branch my PR was based on, but I missed this.

illwieckz (approved):
LGTM.

illwieckz:
@VReaperV No need to wait one day for merging this one, let's consider it's part of #1234. If I haven't missed it, I would have done it yesterday before merging. 🙂️

VReaperV:
Sure.

PR #1284, opened 2024-09-08
Optimize material compute

VReaperV:
Since we always set `instanceCount = 1` and `baseVertex = 0`, we don't need to put those into the `surfaceCommands` buffer.
Use a macro to get the correct uvec2 part of the uvec4 array to avoid adding padding.

This slightly improves performace since there's less memory to fetch/write.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -1868,13 +1868,11 @@ void MaterialSystem::AddDrawCommand( const uint32_t materialID, const uint32_t m
 	}
 
 	cmd.cmd.count = count;
-	cmd.cmd.instanceCount = 1;
 	cmd.cmd.firstIndex = firstIndex;
-	cmd.cmd.baseVertex = 0;
 	cmd.cmd.baseInstance = materialsSSBOOffset;
 	cmd.materialsSSBOOffset = materialsSSBOOffset;
 
-	materialPacks[materialPackID].materials[materialID].drawCommands.emplace_back(cmd);
+	materialPacks[materialPackID].materials[materialID].drawCommands.emplace_back( cmd );
 	lastCommandID = materialPacks[materialPackID].materials[materialID].drawCommands.size() - 1;
 	cmd.textureCount = 0;
 }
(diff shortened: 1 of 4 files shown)


src/engine/renderer/Material.cpp
```diff
@@ -1870,13 +1870,11 @@ void MaterialSystem::AddDrawCommand( const uint32_t materialID, const uint32_t m
 	}
 
 	cmd.cmd.count = count;
-	cmd.cmd.instanceCount = 1;
```
slipher, line 1873: Why are those two fields needed at all? If I grep for `baseVertex` and `instanceCount` I don't see anything that reads them

src/engine/renderer/glsl_source/processSurfaces_cp.glsl
```diff
@@ -108,7 +112,7 @@ void main() {
 	                              + gl_GlobalInvocationID.x
 	                              + 1; // Add 1 because the first surface command is always reserved as a fake command
 	// Each surfaceBatch encompasses 64 surfaceCommands with the same material, padded to 64 as necessary
-	const uvec2 materialID = surfaceBatches[globalGroupID / 2].materialIDs[globalGroupID % 2];
+	const uvec2 materialID = UVEC2_ID_TO_UVEC4( surfaceBatches, globalGroupID );
```
slipher, line 115: The name of this macro doesn't make sense to me. Maybe `UVEC2_FROM_UVEC4_ARRAY`

src/engine/renderer/Material.cpp
VReaperV: It's used by `glMultiDrawElementsIndirectCountARB()` eventually.

src/engine/renderer/glsl_source/processSurfaces_cp.glsl
VReaperV: Yeah, I was looking for a better name for it.

VReaperV: Done.

slipher:
LGTM

PR #1287, opened 2024-09-08
Some more material cleanup

VReaperV:
Builds on #1285.

Add `common_cp` shader. This shader is for use with #insert in other compute shaders, and contains a few #define's and structs that are shared between shaders, as well as #define's for workgroup and invocation ID's.

Don't actually add `surfaceCommand` 0, just do `-1` when adding the command. Simplify some code a bit, remove `culledCommandsOffset`.

Replace `MAX_FRAMES * MAX_VIEWS` with `MAX_VIEWFRAMES`.

### DIFF
--- a/src/engine/renderer/glsl_source/common_cp.glsl
+++ b/src/engine/renderer/glsl_source/common_cp.glsl
@@ -1,0 +1,94 @@
+/*
+===========================================================================
+
+Daemon BSD Source Code
+Copyright (c) 2024 Daemon Developers
+All rights reserved.
+
+This file is part of the Daemon BSD Source Code (Daemon Source Code).
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+	* Redistributions of source code must retain the above copyright
+	  notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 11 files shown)


src/engine/renderer/glsl_source/common_cp.glsl
```diff
@@ -0,0 +1,94 @@
...
+
+/* Common defines */
+
+/* Allows accessing each element of a uvec4 array with a singular ID
+Useful to avoid wasting memory due to alignment requirements
+array must be in the form of uvec4 array[] */
+
+#define UINT_FROM_UVEC4_ARRAY( array, id ) array[id / 4][id % 4]
+#define UVEC2_FROM_UVEC4_ARRAY( array, id ) id % 2 == 0 ? array[id / 2].xy : array[id / 2].zw;
+
```
slipher, line 45: Add parentheses around contents of macro definitions

VReaperV: Fixed.

slipher:
LGTM to the new commits here. So just waiting on #1285

VReaperV:
Rebased, now only the new commits here.

slipher:
LGTM (in case there is any doubt)

PR #1288, opened 2024-09-08
Fix compile error

VReaperV:


illwieckz:
Fixup for #1270:

- https://github.com/DaemonEngine/Daemon/pull/1270

illwieckz:
This is a fix for a merge conflict coming from #1269:

- https://github.com/DaemonEngine/Daemon/pull/1269

I remember having written the exact same patch that I wanted to be part of #1270, so it looks like I simply forgot to push the fixed branch before merging.

This is what I wanted to merge when merging #1270 so there is no need to wait more. I'll just wait for the AppVeyor CI to finish and merge this.

VReaperV:
Merging given the above.

PR #1290, opened 2024-09-09
Fix skybox debug draw

VReaperV:
Skybox is rendered with a mesh that has only position, so it can't be used with generic shader. Fixes the log spam and correctly shows where the skybox tris are.

Also remove GLDeformStage from skybox.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -2501,8 +2501,7 @@ GLShader_skybox::GLShader_skybox( GLShaderManager *manager ) :
 	u_AlphaThreshold( this ),
 	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
-	u_InverseLightFactor( this ),
-	GLDeformStage( this )
+	u_InverseLightFactor( this )
 {
 }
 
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -114,7 +114,12 @@ jobs:
     - bash: |
         set -e
         sudo apt-get update
-        sudo apt-get -y -q --no-install-recommends install zlib1g-dev libncursesw5-dev libgeoip-dev nettle-dev libgmp-dev libcurl4-gnutls-dev libsdl2-dev libogg-dev libvorbis-dev libopusfile-dev libwebp-dev libjpeg8-dev libpng-dev libfreetype6-dev libglew-dev libopenal-dev liblua5.2-dev ninja-build $(EXTRA_PACKAGES)
+        sudo apt-get -q install -y gdebi-core
+        wget http://mirrors.kernel.org/ubuntu/pool/universe/g/glew/libglew2.2_2.2.0-4_amd64.deb
+        wget http://mirrors.kernel.org/ubuntu/pool/universe/g/glew/libglew-dev_2.2.0-4_amd64.deb
+        sudo gdebi -q -n libglew2.2_2.2.0-4_amd64.deb
+        sudo gdebi -q -n libglew-dev_2.2.0-4_amd64.deb
```
DolceTriade, line 121: Leave comments about why we need this and when we can remove it.

slipher:
Please remove commits from other PRs

azure-pipelines.yml
VReaperV: Done in another pr, this one's gonna be rebased on master instead.

VReaperV:
Right, fixed now.

slipher:
That's a lot of code to draw 12 tris! Maybe it should be in tr_sky.cpp next to the code it is based on?

I tested it and noticed the color changes randomly. We should always set all of the uniforms to avoid random results.

VReaperV:
Actually, I just found a much better solution that simply modifies `Tess_StageIteratorSky()` a bit. It's in this pr now.

illwieckz (approved):
LGTM.

PR #1291, opened 2024-09-09
Renderer cleanup: tessellation and portals

slipher:
Mostly cleanup that's needed for #1281. There's one bug fix for scenes with lots of portals on the material system.


illwieckz (approved):
LGTM.

I still see some wrongness with portals with material system, but that should be a good step forward.

PR #1292, opened 2024-09-10
tr_shade: factorise GLSL header generation more

illwieckz:
In fact after factorizing the code more I brought back the vectors.

---

Original message:

@slipher recently taught me we can use shiny C++ for-each feature with classic C arrays, so better use C arrays instead of vectors where the data is known in advance, also make it static.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -390,6 +390,55 @@ static std::string BuildDeformSteps( deformStage_t *deforms, int numDeforms )
 	return steps;
 }
 
+struct addedExtension_t {
+	bool &available;
+	int minGlslVersion;
+	std::string name;
+};
+
+// Fragment and vertex version declaration.
+static const std::vector<addedExtension_t> fragmentVertexAddedExtensions = {
+	{ glConfig2.gpuShader4Available, 130, "EXT_gpu_shader4" },
+	{ glConfig2.gpuShader5Available, 400, "ARB_gpu_shader5" },
+	{ glConfig2.textureGatherAvailable, 400, "ARB_texture_gather" },
+	{ glConfig2.textureIntegerAvailable, 0, "EXT_texture_integer" },
+	{ glConfig2.textureRGAvailable, 0, "ARB_texture_rg" },
+	{ glConfig2.uniformBufferObjectAvailable, 140, "ARB_uniform_buffer_object" },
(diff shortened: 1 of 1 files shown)


slipher:
I guess there is really no need for an array at all here. It can just be a series of `addExtension` calls.

illwieckz:
Yes but such array is easier to edit than copy-pasting function calls. 😜️

That code is not critical for performance so I don't see the need to unroll such loop when we can structure the data nicely instead.

illwieckz:
In fact we can remove much more boiler plate, so I did this. The new implementation doesn't seem to be compatible with C arrays as iterators seems to not be available with C arrays once passed as argument (even with `static const` C arrays), so I brought back vectors.

src/engine/renderer/gl_shader.cpp
```diff
@@ -417,84 +466,26 @@ static void AddConst( std::string& str, const std::string& name, float v1, float
 	str += Str::Format("const vec2 %s = vec2(%.8e, %.8e);\n", name, v1, v2);
 }
 
-static std::string GenVersionDeclaration() {
-	// Basic version declaration
+inline std::string GenVersionDeclaration( const std::vector<addedExtension_t> &addedExtensions ) {
```
slipher, line 469: It never makes sense to use `inline` in a .cpp file. Use `static`

src/engine/renderer/gl_shader.cpp
```diff
@@ -390,6 +390,55 @@ static std::string BuildDeformSteps( deformStage_t *deforms, int numDeforms )
 	return steps;
 }
 
+struct addedExtension_t {
+	bool &available;
+	int minGlslVersion;
+	std::string name;
+};
+
+// Basic version declaration.
```
slipher, line 399: "Basic" is not totally self-explanatory so it would be good to say that this for vertex + fragment shaders.

src/engine/renderer/gl_shader.cpp
illwieckz: Fixed. This was a leftover from when I tried various things to see if the compiler would accept to use iterators on C arrays passed as argument (spoiler: it didn't work 🥲️).

src/engine/renderer/gl_shader.cpp
illwieckz: I used `FragmentVertex` instead.

slipher:
LGTM

PR #1293, opened 2024-09-10
renderer: cancel overbright on opaque decal

illwieckz:
In #1162 I made sure decals are not cancelling overbright, this was to fix impact marks being too faint:

- https://github.com/DaemonEngine/Daemon/pull/1162

In fact only blended decals should not cancel overbright, opaque decals are like any fullbright surface and should then cancel overbright.

The commit also rewrites a bit the code in a better way.

In the future I may rewrite that with a more theoretical operation (based on which blending is an addition, which one is a multiplication, etc. as it's basically a commutativity/associativity problem), but for now listing the various known use cases makes easier to know what is expected. This is good enough for now.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5218,14 +5218,28 @@ static void FinishStages()
 			stage->deformIndex = gl_shaderManager.getDeformShaderIndex( shader.deforms, shader.numDeforms );
 		}
 
-		// We should cancel overBrightBits if there is no light stage.
-		stage->shaderHasNoLight = shaderHasNoLight;
+		// We should cancel overbright if there is no light stage.
+		stage->cancelOverBright = shaderHasNoLight;
 
-		// We should not cancel overbright on decals.
-		if ( shader.sort == Util::ordinal(shaderSort_t::SS_DECAL) )
+		if ( shaderHasNoLight )
 		{
-			// HACK: Reuse that boolean.
-			stage->shaderHasNoLight = false;
+			// We should not cancel overbright if there is no light and it's not using blendFunc dst_color.
+			bool blendFunc_dstColor = ( stage->stateBits & GLS_SRCBLEND_BITS ) == GLS_SRCBLEND_DST_COLOR;
+
+			if ( blendFunc_dstColor )
+			{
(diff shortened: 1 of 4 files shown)


illwieckz:
Before:

unvanquished_2024-09-10_152949_000

After:

unvanquished_2024-09-10_182223_000

Impact marks are still OK:

unvanquished_2024-09-10_182233_000

src/engine/renderer/tr_shader.cpp
```diff
@@ -5218,14 +5218,29 @@ static void FinishStages()
...
+
+			if ( blendFunc_dstColor )
+			{
+				stage->cancelOverBright = false;
+			}
+
+			// We should not cancel overbright if that's a non-opaque decal.
+			bool isDecal = shader.sort == Util::ordinal(shaderSort_t::SS_DECAL);
+			// SRC1 and DST0 are reset to zero in ParseStage (no blending).
+			bool isOpaque = !( ( stage->stateBits & GLS_SRCBLEND_BITS )
```
VReaperV, line 5237: This can be simplified to `!( stage->stateBits & GLS_SRCBLEND_BITS | GLS_DSTBLEND_BITS )`.

illwieckz: Fixed.

VReaperV (approved):
LGTM.

slipher:
Why do decals need special treatment anyway?

illwieckz:
> Why do decals need special treatment anyway?

Decals do not necessarily need special treatment:

> In the future I may rewrite that with a more theoretical operation (based on which blending is an addition, which one is a multiplication, etc. as it's basically a commutativity/associativity problem), but for now listing the various known use cases makes easier to know what is expected.

Though, decals are meant to blend with other surfaces (when non-opaque), so they are a bit like an extra material stage, but with their own surface. This _may_ mean they may require special treatment. Or not.

For example a translucent decal will be blended over a texture with a lightmap, meaning the multiplied light is the one of the texture the decal is blended over.

We can probably implement some generic solution but for now I prefer to address issues I find one by one so I can actually know what use case needs what, so I can later refactorize that. Let's first list the known uses cases with their proper solutions and later it would help us to figure out the factorized code that can satisfy all of them.

PR #1295, opened 2024-09-12
Use blackCube image instead of nothing for missing skybox image

slipher:
This can be seen on the Pulse map.


slipher:
Turns out that setting the alpha to 255 (making the image be uploaded as RGB intead of transparent RGBA) was not really needed for this to work, since the sky shader is presumably always opaque. But it seems more correct anyway.

illwieckz:
I guess you meant blackCube image?

slipher:
Yeah it's a black image, not `blackImage` :stuck_out_tongue:

illwieckz:
The PR title was confusing. 😄️

illwieckz (approved):
LGTM.

PR #1296, opened 2024-09-13
Make possible to build a native engine and dll/exe game on FreeBSD (nexe can run in Linuxulator)

illwieckz:
I needed to take my mind off things so I toyed with this.

It's possible to build a native engine on FreeBSD and native exe/dll game.

But better than that, FreeBSD provides a “Linuxulator” service that provides a compatibility layer for Linux binaries, once this service is enabled, the nacl loader runs and it's possible to a native FreeBSD engine with the nexe running in the Linuxulated nacl loader.

The trick is a bit similar to what we do on Linux arm64 where the native 64-bit engine runs a 32-bit nacl loader with an 32-bit arm nexe: because the engine and the nacl loader are two different binaries exchanging data through a pipe and shared memory, it's possible to run builds of them for different platforms or architectures at the same time if the operating system knows how to run the different binaries.

Unlike the arm64/armhf trick that is totally transparent to the user (nothing to install), the FreeBSD user has to enable the linuxulator by themselve (the Linuxulator enablement commands require root permissions):

```sh
sysrc linux_enable="YES"
service linux start
pkg install linux_base-c7
```


illwieckz:
This doesn't make the nexe buildable on FreeBSD, to make the pexe buildable on FreeBSD (running PNaCl in Linuxulator), one can patch `daemon/external_deps/linux-amd64-default_10/pnacl/bin/pydir/driver_tools.py` this way:

```diff
--- a/external_deps/linux-amd64-default_10/pnacl/bin/pydir/driver_tools.py	2023-12-07 21:58:28.000000000 +0100
+++ b/external_deps/linux-amd64-default_10/pnacl/bin/pydir/driver_tools.py	2024-09-13 02:17:37.311989783 +0200
@@ -156,6 +156,8 @@
   name = platform.system().lower()
   if name.startswith('cygwin_nt') or 'windows' in name:
     name = 'windows'
+  if name == 'freebsd':
+    name = 'linux'
   if name not in ('linux', 'nacl', 'darwin', 'windows'):
     Log.Fatal("Unsupported platform '%s'", name)
   return name
```

This is enough to make the pexe buildable, but then it fails at translating the nexe:

```
[ 99%] Linking CXX executable sgame.pexe
[ 99%] Built target sgame-nacl
[ 99%] Translating sgame (amd64)
external_deps/linux-amd64-default_10/pnacl/bin/le32-nacl-ld.gold: fatal error: build/nacl-vms/../sgame-amd64.nexe.tmp: Invalid argument
gmake[6]: *** [CMakeFiles/sgame-amd64.dir/build.make:77: build/sgame-amd64.nexe] Error 1
```

More details with `make VERBOSE=1`:

```
env PNACLPYTHON=/usr/local/bin/python3.11\  build/pnacl/bin/pnacl-translate --allow-llvm-bitcode-input --pnacl-allow-exceptions    -O2  -arch x86-64 build/nacl-vms/../sgame.pexe -o build/nacl-vms/../sgame-amd64.nexe
external_deps/linux-amd64-default_10/pnacl/bin/le32-nacl-ld.gold: fatal error: build/nacl-vms/../sgame-amd64.nexe.tmp: Invalid argument
```

It's not a problem if nexe are not buildables on FreeBSD. As long as the players can run the engine to join servers with nexe on Linuxulator, or can even build and run an exe server without Linuxulator, this is fine.

illwieckz:
I was able to host a FreeBSD server I could join with a Linux client, and I was able to run a FreeBSD client and join Sweet's server, with the engine downloading custom nexe game and other dpk, executing the custom nexe and letting me play a complete online game for 20 minutes.

srclibs.cmake
```diff
@@ -192,7 +192,7 @@ if (WIN32)
         ${LIB_DIR}/breakpad/src/client/windows/handler/exception_handler.cc
         ${LIB_DIR}/breakpad/src/client/windows/handler/exception_handler.h
     )
-elseif (LINUX)
+elseif (LINUX OR FREEBSD)
```
slipher, line 195: Does Breakpad really work? I'd be shocked.

CMakeLists.txt
```diff
@@ -572,7 +572,7 @@ endif()
 
 # Look for OpenGL here before we potentially switch to looking for static libs.
 if (BUILD_CLIENT)
-    if (LINUX)
+    if (LINUX OR FREEBSD)
```
slipher, line 575: Does it really have this Legacy vs. GLVND thing?

src/common/FileSystem.cpp
```diff
@@ -192,7 +192,7 @@ inline int my_open(Str::StringRef path, openMode_t mode)
 	int fd = _open_osfhandle(reinterpret_cast<intptr_t>(h), modes[mode_] | O_BINARY | O_NOINHERIT);
 	if (fd == -1)
 		CloseHandle(h);
-#elif defined(__APPLE__)
+#elif defined(__FreeBSD__) || defined(__APPLE__)
 	// O_CLOEXEC is supported from 10.7 onwards
```
slipher, line 196: Comment should be clarified as to which OS, or deleted

CMakeLists.txt
illwieckz: I expect FreeBSD to be exactly like Linux on that part: they both use Mesa as the de facto standard implementation, not only for drivers, but for the library itself.

The only other non-Mesa OpenGL implementation available on FreeBSD is likely the proprietary Nvidia one for their own proprietary GL driver, and GLVND was actually initiated by Nvidia (because having a vendor-neutral dispatch library only makes sense if there exist more than one vendor library available, and on Linux and FreeBSD, the only two ones are likely Mesa… and Nvidia).

Also FreeBSD provides both legacy `/usr/local/lib/libGL.so` and GLVND `/usr/local/lib/libOpenGL.so`.

illwieckz: I confirm both link and run.

srclibs.cmake
illwieckz: Right, it does not, removed.

src/common/FileSystem.cpp
illwieckz: Comment clarified.

slipher (commented):
LGTM modulo comment

CMakeLists.txt
```diff
@@ -777,7 +777,7 @@ if (USE_BREAKPAD)
         set(LIBS_ENGINE_BASE ${LIBS_ENGINE_BASE} srclibs-breakpad-exception_handler
             srclibs-breakpad-crash_generation_client srclibs-breakpad-common)
 
-    elseif (LINUX)
+    elseif (LINUX OR FREEBSD)
```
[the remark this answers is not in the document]

PR #1301, opened 2024-09-14
q_shared: inline some vector functions

illwieckz:
This helps to get more performance in some place of the code.

One good example that benefits from it is the `R_TBNtoQtangents()` function. This function is used for rendering models when the models is too large for some GPU limit. It's possible to force the engine to use this code path for all models with `r_vboVertexSkinning 0`. A good example of scene that puts the engine on knees is the `plat23` map with `setviewpos 1920 1920 0 0 0`.

This simple change increases performance by `15%` on this heavy scene, with the `ondemand` CPU profile, it makes the game run at `77fps` instead of  `67fps` on my machine.

Before:

unvanquished_2024-09-14_035249_000

After:

unvanquished_2024-09-14_035455_000

When I set the CPU profile to `performance` and do some tricks like suspending other apps and setting `-20` priority to the renderer thread, that patch makes the game run at `83fps` instead of `72fps` on my machine.

Before:

unvanquished_2024-09-14_035903_000

After:

unvanquished_2024-09-14_040318_000


illwieckz:
I can reproduce the framerate difference very consistently, not only the framerate is stable and lasts for minutes or hours during tests, but I get consistently the same result if when re-run the test multiple times in a row, and when I re-run the test hours later.

slipher:
LGTM

illwieckz:
@VReaperV is the CI build error in Material uncovered by this change?

- https://dev.azure.com/UnvanquishedDevelopment/Daemon/_build/results?buildId=9738&view=logs&jobId=e4954a55-bc0b-5fbe-d8f6-05775b1a9559&j=e4954a55-bc0b-5fbe-d8f6-05775b1a9559&t=6fedcdd4-7b5b-58c7-109c-77b7f76a6cd5

VReaperV:
> @VReaperV is the CI build error in Material uncovered by this change?

Yep, needs a `vec3_t` somewhere in `MaterialSystem` for PVS lock.

illwieckz:
For now I `VectorClear()` it, it seems to make the CI happy.

VReaperV:
Alright, I'll pr a proper fix a bit later.

src/engine/renderer/Material.cpp
```diff
@@ -1652,6 +1652,7 @@ void MaterialSystem::CullSurfaces() {
 		}
 		matrix_t viewMatrix;
 		if ( PVSLocked ) {
+			VectorClear( origin );
```
VReaperV, line 1655: This can now be removed, the fix is on `master`.

illwieckz: Now removed. Let's see if the CI is now happy with it and I'll merge this.

PR #1303, opened 2024-09-15
Fix material system PVS lock

VReaperV:
Fixes the issue uncovered in #1301, as well as an issue that resulted in every view using the main view's origin.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -1646,11 +1646,12 @@ void MaterialSystem::CullSurfaces() {
 
 		vec4_t frustumPlanes[6];
 		for ( int i = 0; i < 6; i++ ) {
-			VectorCopy( PVSLocked ? lockedFrustum[i].normal : frustum[0][i].normal, frustumPlanes[i] );
-			frustumPlanes[i][3] = PVSLocked ? lockedFrustum[i].dist : frustum[0][i].dist;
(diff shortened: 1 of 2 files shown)


src/engine/renderer/Material.cpp
```diff
@@ -1647,11 +1647,12 @@ void MaterialSystem::CullSurfaces() {
...
 		vec4_t frustumPlanes[6];
 		for ( int i = 0; i < 6; i++ ) {
-			VectorCopy( PVSLocked ? lockedFrustum[i].normal : frustum[0][i].normal, frustumPlanes[i] );
-			frustumPlanes[i][3] = PVSLocked ? lockedFrustum[i].dist : frustum[0][i].dist;
+			VectorCopy( PVSLocked ? lockedFrustums[view][i].normal : frustum[0][i].normal, frustumPlanes[i] );
+			frustumPlanes[i][3] = PVSLocked ? lockedFrustums[view][i].dist : frustum[0][i].dist;
 		}
 		matrix_t viewMatrix;
 		if ( PVSLocked ) {
+			VectorCopy( lockedOrigins[view], origin );
```
slipher, line 1655: Does locking PVS actually hold the list of views fixed? Otherwise this doesn't make sense.

VReaperV: You're right, I added that now.

slipher:
LGTM

PR #1304, opened 2024-09-15
Minor build system cleanup

slipher:


cmake/DaemonFlags.cmake
```diff
@@ -305,7 +305,7 @@ else()
 
 	# The -pthread flag sets some preprocessor defines,
 	# it is also used to link with libpthread on Linux.
-	if (NOT APPLE)
+	if (LINUX OR FREEBSD)
 		try_c_cxx_flag(PTHREAD "-pthread")
```
illwieckz, line 309: See #1305:

- https://github.com/DaemonEngine/Daemon/pull/1305

cmake/DaemonFlags.cmake
```diff
@@ -315,7 +315,7 @@ else()
 	endif()
 
 	# Hardening.
-	if (USE_HARDENING OR NOT MINGW)
+	if (USE_HARDENING OR (NOT MINGW AND NOT NACL))
```
illwieckz, line 318: If that should not be used with NaCl, the `USE_HARDENING` name can be removed from `set(DEFAULT_NACL_VM_INHERITED_OPTIONS …)` in `CMakeLists.txt` (better do both).

PR #1308, opened 2024-09-16
tr_init: set default r_reliefDepthScale to 0.02 like DarkPlaces

illwieckz:
Most of our heightmaps come from Xonotic which are supposed to use DarkPlaces 0.02 default relief depth scale. It's easier for use to align with that default than to convert any height map with import from Xonotic. Our heightmaps are actually wrong without this.

I haven't seen any relief mapping code after a quick search in ioq3, so we're not breaking any usage but we may actually be defining the standard instead. If we are in position to set the standard for idtech3 games, better align with friendly projects like Xonotic/DarkPlaces.


slipher:
LGTM

PR #1309, opened 2024-09-17
Material debug improvements

VReaperV:
Add an `r_showGlobalMaterials` cvar that allows visualising material system materials by color-coding surfaces, in 3 different modes: depth, opaque, opaque + transparent.

Add `r_showTris` to `gl_genericShaderMaterial`.

Add `r_materialDebug` cvar to allow to quickly add debug output to shaders. This is intended for use with compute shaders and Nsight Graphics (due to formatting interface).

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -2136,9 +2136,14 @@ void MaterialSystem::RenderMaterial( Material& material, const uint32_t viewID )
 
 	if ( r_showTris->integer
 		&& ( material.stateBits & GLS_DEPTHMASK_TRUE ) == 0
-		&& material.shaderBinder == &BindShaderLightMapping )
+		&& ( material.shaderBinder == &BindShaderLightMapping || material.shaderBinder == &BindShaderGeneric3D ) )
 	{
-		gl_lightMappingShaderMaterial->SetUniform_ShowTris( 1 );
+		if ( material.shaderBinder == &BindShaderLightMapping ) {
+			gl_lightMappingShaderMaterial->SetUniform_ShowTris( 1 );
+		} else if ( material.shaderBinder == &BindShaderGeneric3D ) {
+			gl_genericShaderMaterial->SetUniform_ShowTris( 1 );
+		}
+
 		GL_State( GLS_DEPTHTEST_DISABLE );
 		glMultiDrawElementsIndirectCountARB( GL_LINES, GL_UNSIGNED_INT,
 			BUFFER_OFFSET( material.surfaceCommandBatchOffset * SURFACE_COMMANDS_PER_BATCH * sizeof( GLIndirectBuffer::GLIndirectCommand )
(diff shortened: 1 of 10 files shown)


slipher:
r_showGlobalMaterials causes a GLSL compiler error if I turn it on.

What's an example of how r_materialDebug would be used?

VReaperV:
Fixed the compile error.

VReaperV:
> What's an example of how r_materialDebug would be used?

E. g. with this surface blinking between tranparent and black: temp0
I can tell with the debug output that it's not being culled and the 2 draw commands that it uses:
image

I've used a buffer like this to fix a number of issues with material system, like culling etc.

src/engine/renderer/Material.cpp
```diff
@@ -2051,6 +2052,80 @@ void MaterialSystem::RenderMaterial( Material& material, const uint32_t viewID )
...
+				// Even though this is for depth materials, we don't actually draw anything on depth pass
+				if ( material.sort != 1 ) {
+					return;
+				}
+
+				const float id = ( float ) material.id / ( materialPacks[0].materials.size() + 2 ) + 1;
+
+				color[0] = fmod( id, 1 / 3.0 ) * colors[int( material.id * 6.0
+					/ materialPacks[0].materials.size() )][0];
+				color[1] = fmod( std::max( 0.0, id - 1 / 3.0 ), 1 / 3.0 ) * colors[int( material.id * 6.0
```
slipher, line 2098: What are these color calculations trying to do? This part does (id - 1/3) modulo 1/3 which doesn't make sense as that's just id modulo 1/3.

VReaperV: The idea was to assign a different rgb colour to each material. The calculation is definitely wrong though looking at it again, gonna fix it later.

VReaperV: Fixed.

illwieckz (approved):
LGTM.

illwieckz:
I had to test with `llvmpipe` because of remaining bugs in `radeonsi` with material system, but that's not this PR's fault.

src/engine/renderer/Material.cpp
```diff
@@ -2051,6 +2052,80 @@ void MaterialSystem::RenderMaterial( Material& material, const uint32_t viewID )
...
+			case Util::ordinal( MaterialDebugMode::DEPTH ):
+			{
+				// Even though this is for depth materials, we don't actually draw anything on depth pass
+				if ( material.sort != 1 ) {
+					return;
+				}
+
+				const float id = ( float ) material.id / ( materialPacks[0].materials.size() + 2 ) + 1;
+
+				color[0] = std::min( id, 1 / 3.0f ) * 3.0 * colors[int( material.id * 6.0
```
slipher, line 2096: Use `Math::Clamp` for the min + max

slipher:
LGTM

PR #1314, opened 2024-09-19
computeLight_fp: fix compilation when r_highPrecisionRendering is disabled

illwieckz:
Fix compilation when `r_highPrecisionRendering` is disabled.

I don't know if the fix is right, but it fixes building the `lightMapping` GLSL and I see no visual bugs.

What makes me doubt is that I replace a `sampler3D` with an `usampler3D` but I don't know the difference.

This fixes a bug introduced in f960177aba3c68e17e8ebfc865e8e8682268667a from #1105:

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -167,22 +167,24 @@ void computeLight(in vec3 lightColor, vec4 diffuseColor, inout vec4 color) {
 #if defined(r_realtimeLighting) && r_realtimeLightingRenderer == 1
 #if defined(HAVE_EXT_texture_integer) && defined(r_highPrecisionRendering)
 const int lightsPerLayer = 16;
-uniform usampler3D u_LightTilesInt;
+#define lightTilesSampler_t usampler3D
+#define lightTilesUniform u_LightTilesInt
 #define idxs_t uvec4
-idxs_t fetchIdxs( in vec3 coords, in usampler3D u_LightTilesInt ) {
-  return texture3D( u_LightTilesInt, coords );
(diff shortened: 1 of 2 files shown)


src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -177,20 +177,20 @@ int nextIdx( inout idxs_t idxs ) {
   idxs = idxs >> 2;
   return int( tmp.x + tmp.y + tmp.z + tmp.w );
 }
-#else // !HAVE_EXT_texture_integer
+#else // !HAVE_EXT_texture_integer || !r_highPrecisionRendering
 const int lightsPerLayer = 4;
-uniform sampler3D u_LightTiles;
+uniform usampler3D u_LightTilesInt;
```
VReaperV, line 182: This sampler should stay as `sampler3D u_LightTiles`. The reason for the distinction is so that `GLShaderManager::ShaderPostProcess()` can work correctly without having to process macros in shaders, which is what we used previously to select the correct sampler type.

illwieckz: Why do `u_LightTilesInt` is passed as argument in the other use case? It looks to never be modified.

Can it be used as a global variable instead? This would prevent us to duplicate some lines with the variants (up to 4 variants with `computeDynamicLights`…

VReaperV: Both use cases should pass it as argument, I forgot to add it to the non-integer one in #1105.

It can't be global with bindless textures whose handles are sourced from a buffer. We can just ditch the version that uses non-integer textures, since integer textures are supported pretty much everywhere.

illwieckz: OK, the current code should now be correct.

VReaperV (approved):
LGTM.

PR #1315, opened 2024-09-19
Make VFS file operations case-insensitive

### LINKED ISSUE #1302: Should the VFS be case-insensitive?
I saw a map where an image failed to load because the file's extension was written in upper case, and was surprised to learn that VFS paths are case-sensitive. I found this odd because everything else (like commands or any sort of config file) seems to be case-insensitive. So making VFS file opening case-insensitive seems more consistent, and might improve asset compatibility. @sweet235 Do you often see case-sensitivity problems when porting Tremulous maps?

Note: this should be fine for dpkdirs on case-sensitive filesystems because we always open VFS files by looking them up in the loaded file list first, rather than blindly passing the path to the OS's open function. We'd just need a 1-line change to PakPath::ReadFile to use the name from the file list rather than the requested name.

slipher:
Make single-file VFS operations (open, read, timestamp, copy to homepath) case-insensitive with regard to the path name.

The ListFiles stuff appears to be case-insensitive already.

Fixes #1302.


illwieckz:
Excellent !

Before:

unvanquished_2024-09-20_030938_000

After:

unvanquished_2024-09-20_031058_000

illwieckz (approved):
LGTM.

PR #1317, opened 2024-09-20
Make some uniforms global

VReaperV:
Don't put data that doesn't change into the material buffer.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -260,8 +260,6 @@ void UpdateSurfaceDataGeneric3D( uint32_t* materials, Material& material, drawSu
 		gl_genericShaderMaterial->SetUniform_DepthScale( pStage->depthFadeValue );
 	}
 
-	gl_genericShaderMaterial->SetUniform_VertexInterpolation( false );
-
 	gl_genericShaderMaterial->WriteUniformsToBuffer( materials );
 }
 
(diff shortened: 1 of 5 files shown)


src/engine/renderer/Material.cpp
```diff
@@ -1104,6 +1105,107 @@ void BindShaderLightMapping( Material* material ) {
 	}
 	// FIXME: else
 
+	// bind u_LightGrid1
```
slipher, line 1108: Why is there a lot of new code here?

src/engine/renderer/gl_shader.h
```diff
@@ -3496,7 +3496,7 @@ class u_VertexInterpolation :
 {
 public:
 	u_VertexInterpolation( GLShader *shader ) :
-		GLUniform1f( shader, "u_VertexInterpolation" )
+		GLUniform1f( shader, "u_VertexInterpolation", true )
```
slipher, line 3499: This seems wrong. This one is set for surfaces that are part of an MD3 model

src/engine/renderer/gl_shader.h
```diff
@@ -2497,7 +2497,7 @@ class u_EnvironmentMap0 :
 	GLUniformSamplerCube {
 	public:
 	u_EnvironmentMap0( GLShader* shader ) :
-		GLUniformSamplerCube( shader, "u_EnvironmentMap0" ) {
+		GLUniformSamplerCube( shader, "u_EnvironmentMap0", true ) {
```
slipher, line 2500: I imagine this should be dependent on the surface's position? Currently it is determined from the viewing position but I assume that's just because the old reflective specular code was half unfinished and written while drunk.

src/engine/renderer/gl_shader.h
VReaperV: I suppose that uniform can just be removed from the shader since it will likely use other means to do vertex skinning/animation.

src/engine/renderer/Material.cpp
VReaperV: Hmm, that can be removed since light/deluxe grid textures don't change and they're set always set elsewhere.

src/engine/renderer/gl_shader.h
VReaperV: It should be, yeah. I was just going off of whatever code we already had.

VReaperV: To be more specific, binding cubemaps based on where the surface is will also give bad results, especially if the surface is large. Ideally we'd further divide the frustum (further from the current tiled implementation) into clusters and assign a cubemap to each cluster.

VReaperV: Actually, this would pretty much require changing from tiled renderer to clustered, because I plan to merge stages with the same data in the buffer, which means that they will point to the wrong cubemaps for most surfaces (right now having separate data for each stage in the material buffer makes it take up about ~10x the space on average and decreases performance).

src/engine/renderer/gl_shader.h
VReaperV: I've removed this uniform entirely from the material shaders for now.

src/engine/renderer/Material.cpp
VReaperV: I removed the same code from `UpdateSurfaceDataLightMapping()` because other shaders also set global uniforms in `BindShader*()`.

VReaperV:
Rebased/ready for review again.

src/engine/renderer/gl_shader.h
slipher: This change can be removed then since it's in principle a per-surface value.

slipher (commented):
LGTM modulo comment

src/engine/renderer/gl_shader.h
VReaperV: Ah, yea, removed.

PR #1320, opened 2024-09-25
cmake: add USE_EXTERNAL_DEPS option to prevent the download of external_deps packages at build time

illwieckz:
Add `USE_EXTERNAL_DEPS` option to prevent the download of `external_deps` packages at build time.

It's up to the user to know what he is doing and to provide everything he needs when he disables this option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -244,7 +244,9 @@ endif()
 # Dependencies version, this must match the number in external_deps/build.sh
 set(DEPS_VERSION 10)
 
-if (NOT NACL)
+option(USE_EXTERNAL_DEPS "Download or reuse dependencies from EXTERNAL_DEPS_DIR (mandatory for building and running NaCl .nexe binaries)." ON)
+
+if (USE_EXTERNAL_DEPS AND NOT NACL)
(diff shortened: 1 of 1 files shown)


illwieckz:
It is assumed the rest of the code already skips tasks using `DEPS_DIR` when `DEPS_DIR` is not set, thanks to:

- https://github.com/DaemonEngine/Daemon/pull/805

CMakeLists.txt
```diff
@@ -244,7 +244,9 @@ endif()
 # Dependencies version, this must match the number in external_deps/build.sh
 set(DEPS_VERSION 10)
 
-if (NOT NACL)
+option(USE_EXTERNAL_DEPS "Download build dependencies, required for building an running NaCl nexe binaries." ON)
```
slipher, line 247: I find the description not totally accurate as it not only controls downloading, but also whether dependencies that are already there (which may even have been built by the user). How about "Download or reuse dependencies from EXTERNAL_DEPS_DIR (mandatory for building and running NaCl .nexe binaries)"

slipher:
LGTM

Issue #1321, opened 2024-09-25
Entity baselines are of questionable utility


slipher:
When the server transmits an entity state to a client, it is always delta encoded against some base entity state. This base entity state is one of the following, in order of priority:
1. The entity's state in the client's most recently acknowledged snapshot
2. The entity's state in the "entity baseline"
3. An all-zeroes entity state

I think (1) should be used the majority of the time under normal conditions. But when an entity first enters the client's PVS, (1) won't be available. Assuming entities stay in roughly the same position from frame to frame, and the network is not too flaky, (1) will start being available after a delay roughly equal to the client's ping. I assume (2) is intended to make things more efficient during that window.

This issue is about (2). So what is the entity baseline? I always imagined that entity baselines and delta compression work like keyframes in a video: periodically you get a new baseline/keyframe, and then for the next few seconds each frame is encoded via the diff from that base. (1) is kind of like that, but (2) is stupider: a baseline is taken only once, when the map is first loaded. It's never updated, not even on a `map_restart` (the cause of the inconsistent behavior in https://github.com/Unvanquished/Unvanquished/issues/2124). It's like if the entire video were encoded based on diffs from the first frame. This was probably good for Quake, since most entities that existed at the start could never be destroyed. But we have buildables that can be destroyed and moved, and beacons for those buildables. Entity numbers are recycled, so it's likely that the baseline entity with a given number isn't even the same type as the currently existing entity. It's likely that using the entity baseline results in more bandwidth consumption in such cases.

What should we do about this? The easiest thing would be to NUKE or disable baselines. Maybe they can be considered an unnecessary micro-optimization from today's standpoint. If not, we could try some of the following ideas:
- Remove an entity from the baseline once it dies. If we do this (and never add any new entities), we could keep the advantages of the baseline just for permanent stuff like doors.
- When a new entity spawns, update its baseline entity.
- Transmit baselines lazily instead of all at once.
- Remember states of entities received by the client even after they leave the PVS. This would be more an extension of (1) that would replace the baseline concept.

slipher:
@DolceTriade and I tested this on a server, having around 200 ping. On a torture test layout with around 600 buildables in a single room, baselines do have a significant advantage. With baselines turned off, the lag spike during the interval between the buildables entering the PVS and the client acknowledging the first snapshot is very noticeable. A bad thing is that the server has to keep sending the entities' full state every frame for several frames during this period, since there is not yet an acknowledged snapshot to delta against.

[...]

PR #1322, opened 2024-09-25
Allow disabling entity baselines (can load big layouts)

slipher:
Entity baselines are supposed to make netcode more efficient, but it's dubious (see #1321). Let server owners disable entity baselines with `set sv_useBaseline 0` as a workaround to https://github.com/Unvanquished/Unvanquished/issues/2124. This is compatible with vanilla clients. Probably no one will notice a difference wrt bandwith efficiency.

### DIFF
--- a/src/engine/server/sv_init.cpp
+++ b/src/engine/server/sv_init.cpp
@@ -263,12 +263,9 @@ baseline will be transmitted
 */
 void SV_CreateBaseline()
 {
-	sharedEntity_t *svent;
-	int            entnum;
-
-	for ( entnum = 1; entnum < sv.num_entities; entnum++ )
+	for ( int entnum = MAX_CLIENTS; entnum < sv.num_entities; entnum++ )
 	{
-		svent = SV_GentityNum( entnum );
+		sharedEntity_t *svent = SV_GentityNum( entnum );
 
 		if ( !svent->r.linked )
 		{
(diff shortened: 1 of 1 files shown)


DolceTriade (commented):
Seems like this would be more visible with high ping clients then?

slipher:
Yes, high ping would make there be more of a difference. I imagine that it's even more important to have a room with a lot of entities in it, so that there are enough packets to congest the network. (And the entities should have existed since the server started, so that the baseline is helpful).

slipher:
Server owners may use this if they want to load layouts with more than 600 or so buildables. But they should make sure the buildables are spread out somewhat. If you have hundreds of them in one room, clients will get lag spikes when the buildables enter PVS. The baseline does provide some advantage in the case that you have hundreds of entities in one place and those entities were present when the map loaded.

slipher:
Added latch flag so that it is more obvious that you need to change map for a change to take effect.

DolceTriade:
I feel like this is too big a change to add at the last minute. We can merge this after. Server hosts can cherry-pick this change in themselves if they need it since it's not compat breaking.

slipher:
In what sense is it a big change? The diff is +11/-4 and it doesn't change anything unless you set the cvar.

DolceTriade:
I mean it has the possibility to be an impactful change if there ends up being a regression. This is a nice to have thing with potential drawbacks. I don't think we need to force it in at the last minute.

slipher:
I'm confused. Are you talking about the possibility of a regression if the user flips the cvar? Or the possibility of a regression if it is left unchanged?

DolceTriade:
A regression if unchanged. As I said, if we wanted this merged, we should have done it sooner. This is not a required change and we've already shipped RC2. Let's ship 0.55 and we can merge this. If people want this, they can build it into their own servers.

slipher:
Is this OK to merge now?

illwieckz (approved):
LGTM.

PR #1325, opened 2024-09-29
Add missing `break;` to reliefMaping_fp

VReaperV:
This can slightly increase performance if the shader compiler isn't optimising out the rest of the loop already.

### DIFF
--- a/src/engine/renderer/glsl_source/reliefMapping_fp.glsl
+++ b/src/engine/renderer/glsl_source/reliefMapping_fp.glsl
@@ -171,6 +171,7 @@ vec2 ReliefTexOffset(vec2 rayStartTexCoords, vec3 viewDir, mat3 tangentToWorldMa
 			if(currentDepth >= heightMapDepth)
 			{
 				bestDepth = currentDepth;
+				break;
 			}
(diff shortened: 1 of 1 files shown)


illwieckz:
Impressive! On an AMD Radeon Pro W7600 (RDNA3/navi33) with Mesa radeonsi 24.0.9, the framerate goes from `290fps` to `425fps` on a 4K screen with `ultra` preset!

Before:

unvanquished_2024-09-29_115452_000

After:

unvanquished_2024-09-29_115947_000

This is to compare to the `610fps` I get when I disable relief mapping:

unvanquished_2024-09-29_120530_000

This means that maybe a Radeon RX 7900 XTX can hit the 999fps cap on `ultra` preset @ 4K? 🤪️

VReaperV:
Nice, that's quite a big improvement.

illwieckz:
Indeed! And with a Mesa build from today, I reach `446fps` with that single-slot average Radeon PRO W7600 card:

unvanquished_2024-09-29_122048_000

illwieckz (approved):
LGTM.

PR #1329, opened 2024-10-01
sdl_glimp: add r_arb_framebuffer_object cvar

illwieckz:
Add `r_arb_framebuffer_object` cvar, I noticed it was missing when debugging something. Having it makes easy to test the workaround.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -72,6 +72,8 @@ static Cvar::Cvar<bool> r_arb_compute_shader( "r_arb_compute_shader",
 	"Use GL_ARB_compute_shader if available", Cvar::NONE, true );
 static Cvar::Cvar<bool> r_arb_depth_clamp( "r_arb_depth_clamp",
 	"Use GL_ARB_depth_clamp if available", Cvar::NONE, true );
+static Cvar::Cvar<bool> r_arb_framebuffer_object( "r_arb_framebuffer_object",
+	"Use GL_ARB_framebuffer_object if available", Cvar::NONE, true );
 static Cvar::Cvar<bool> r_arb_explicit_uniform_location( "r_arb_explicit_uniform_location",
 	"Use GL_ARB_explicit_uniform_location if available", Cvar::NONE, true );
 static Cvar::Cvar<bool> r_arb_gpu_shader5( "r_arb_gpu_shader5",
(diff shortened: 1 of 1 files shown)


slipher:
Shouldn't there just be one GL extension cvar for the ARB prefix version and one GL extension cvar for the EXT prefixed version? No one is violating the specification by not providing an extension, right? So it's not a workaround.

illwieckz:
> Also there is this in LoadExt:
> 
> ```
> // Required extension can't be made optional
> ASSERT( !( flags & ExtFlag_REQUIRED ) );
> ```
> 
> So I would suggest to instead structure it like
> 
> ```
> if (LOAD_EXTENSION_WITH_TEST(ARB_blabla) {
> enable arb functions..
> } else if (LOAD_EXTENSION_WITH_TEST(EXT_blabla) {
> enable ext functions...
> } else {
> Sys:Error("missing xxx extension");
> }
> ```

I don't see which needs that fulfills. Also ARB and EXT are not the only existing prefixes. For example OpenGL 3.1 Core made `GL_NV_primitive_restart` standard and required for 3.1.

Also, `EXT_framebuffer_object` is not an alternative to `ARB_framebuffer_object`, it only happens that the sum of `EXT_framebuffer_object`  plus `EXT_framebuffer_blit` is a featureful alternative to `ARB_framebuffer_object` for our need. We can't make macros find the multiple alternative extensions needed to replace a single ARB extension. The extension equivalence is at best a surjective function. So if we did:

```
load(framebuffer_object);

load(blabla)
{
  if !loadext(ARB_${blabla});
    loadext(EXT_${blabla});
}
```

When `ARB_framebuffer_object` would be missing we would successfully load `EXT_framebuffer_object` but we would fail to load `EXT_framebuffer_blit` required to complement `EXT_framebuffer_object` to reach a similar level of feature of `ARB_framebuffer_object` alone.

illwieckz:
Hum, now that I re-read your comment, maybe you didn't suggested to add the `if ARB else if EXT` snippet in `LoadExt`, but to modify the existing `if ARB else if EXT` snippet because of `LoadExt`. 🤭️

illwieckz:
I rewrote the code to avoid making required something with a test (and then optional).

Testing this make me uncover a bug in Mesa: using `MESA_EXTENSION_OVERRIDE='-ARB_framebuffer_object'` produces a broken `EXT_framebuffer_object` that leads to a crash, while on a device missing `ARB_framebuffer_object` for real, there is no crash. So our own `r_arb_framebuffer_object` cvar is more reliable: it makes us able to test `EXT_framebuffer_object` functions even if `ARB_framebuffer_object` is available.

slipher:
LGTM

PR #1330, opened 2024-10-01
tr_shader: transmit `RSF_FITSCREEN` and `RSF_NOMIP` flag from gamecode to shaders

illwieckz:
When the game code requests the engine to load a texture and sets the `RSF_FITSCREEN` or the `RSF_NOMIP` flag, it was only used on images without shaders.


slipher:
LGTM

PR #1331, opened 2024-10-01
More renderer debug logs

illwieckz:
Some renderer debug logs I needed to debug things:

- sdl_glimp: tell if used extension is required or optional

When it is suspected an extension may be buggy, it is useful to know which ones are optional so one can disable them to verify the assumption.

- renderer: add more debug log about image creation

When the OpenGL driver reports errors when creating or uploading an image, it's useful to have such debug logs to know exactly which image failed, just by reading the immediate previous line. It is useful to identify which image format may be broken in an OpenGL driver.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1897,7 +1897,14 @@ static bool LoadExt( int flags, bool hasExt, const char* name, bool test = true 
 	{
 		if ( test )
 		{
-			logger.WithoutSuppression().Notice( "...using GL_%s", name );
+			if ( flags & ExtFlag_REQUIRED )
+			{
+				logger.WithoutSuppression().Notice( "...using required extension GL_%s.", name );
+			}
+			else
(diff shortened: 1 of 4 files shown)


VReaperV (approved):
LGTM.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1897,7 +1897,14 @@ static bool LoadExt( int flags, bool hasExt, const char* name, bool test = true
 	{
 		if ( test )
 		{
-			logger.WithoutSuppression().Notice( "...using GL_%s", name );
+			if ( flags & ExtFlag_REQUIRED )
+			{
+				logger.WithoutSuppression().Notice( "...using required core extension GL_%s.", name );
```
slipher, line 1902: The message is wrong, taking this branch doesn't mean it must be core

illwieckz: Right, fixed.

PR #1333, opened 2024-10-01
Tiled renderer improvements

VReaperV:
- Removed the non-UBO and non-integer texture code paths. The non-UBO and non-integer texture code was supposed to be used when those extensions are not available, however:
1. These extensions are available on nearly all hardware nowadays. These extensions became core in 3.1 as well.
2. Hardware that doesn't support those extensions probably won't be using `r_realtimeighting on` anyway due to performance reasons.
3. These old code paths were broken.
- Made the way light IDs are stored in the lighttile texture sane: this was a mess of bitwise operations and strange ways of storing the ID by putting different bits into different parts of a uvec4. Now this will just store each 8-bit ID consecutively.
- Cleaned up the code in `lighttile_fp` and `computeLight_fp`.

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -38,7 +38,6 @@ uniform float u_EnvironmentInterpolation;
 #endif // USE_REFLECTIVE_SPECULAR
 
 #if defined(r_realtimeLighting) && r_realtimeLightingRenderer == 1
-#if defined(HAVE_ARB_uniform_buffer_object)
 struct light {
   vec4  center_radius;
   vec4  color_type;
@@ -49,17 +48,6 @@ layout(std140) uniform u_Lights {
   light lights[ MAX_REF_LIGHTS ];
 };
 #define GetLight(idx, component) lights[idx].component
-#else // !HAVE_ARB_uniform_buffer_object
-uniform sampler2D u_LightsTexture;
-#define idxToTC( idx, w, h ) vec2( floor( ( idx * ( 1.0 / w ) ) + 0.5 ) * ( 1.0 / h ), \
-				   fract( ( idx + 0.5 ) * (1.0 / w ) ) )
-const struct GetLightOffsets {
-  int center_radius;
-  int color_type;
-  int direction_angle;
-} getLightOffsets = GetLightOffsets(0, 1, 2);
-#define GetLight(idx, component) texture2D( u_LightsTexture, idxToTC(3 * idx + getLightOffsets.component, 64.0, float( 3 * MAX_REF_LIGHTS / 64 ) ) )
-#endif // HAVE_ARB_uniform_buffer_object
 
 uniform int u_numLights;
 #endif // defined(r_realtimeLighting) && r_realtimeLightingRenderer == 1
@@ -165,34 +153,21 @@ void computeLight(in vec3 lightColor, vec4 diffuseColor, inout vec4 color) {
 #endif // !defined(USE_DELUXE_MAPPING) && !defined(USE_GRID_DELUXE_MAPPING)
 
 #if defined(r_realtimeLighting) && r_realtimeLightingRenderer == 1
-#if defined(HAVE_EXT_texture_integer) && defined(r_highPrecisionRendering)
 const int lightsPerLayer = 16;
+
 #define lightTilesSampler_t usampler3D
 #define lightTilesUniform u_LightTilesInt
 #define idxs_t uvec4
+
 idxs_t fetchIdxs( in vec3 coords, in lightTilesSampler_t lightTilesUniform ) {
   return texture3D( lightTilesUniform, coords );
 }
+
 int nextIdx( inout idxs_t idxs ) {
   uvec4 tmp = ( idxs & uvec4( 3 ) ) * uvec4( 0x40, 0x10, 0x04, 0x01 );
   idxs = idxs >> 2;
   return int( tmp.x + tmp.y + tmp.z + tmp.w );
 }
-#else // !HAVE_EXT_texture_integer || !r_highPrecisionRendering
-const int lightsPerLayer = 4;
-#define lightTilesSampler_t sampler3D
-#define lightTilesUniform u_LightTiles
-#define idxs_t vec4
-idxs_t fetchIdxs( in vec3 coords, in lightTilesSampler_t lightTilesUniform ) {
-  return texture3D( lightTilesUniform, coords ) * 255.0;
-}
-int nextIdx( inout idxs_t idxs ) {
-  vec4 tmp = idxs;
-  idxs = floor(idxs * 0.25);
-  tmp -= 4.0 * idxs;
-  return int( dot( tmp, vec4( 64.0, 16.0, 4.0, 1.0 ) ) );
-}
-#endif // !HAVE_EXT_texture_integer || !r_highPrecisionRendering
 
 uniform lightTilesSampler_t lightTilesUniform;
 
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -159,10 +159,10 @@ layout(std140) uniform u_Lights {
 uniform int u_numLights;
 
 #if defined(USE_REFLECTIVE_SPECULAR)
-void computeDynamicLight( int idx, vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse,
+void computeDynamicLight( uint idx, vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse,
 	vec4 material, inout vec4 color, in samplerCube u_EnvironmentMap0, in samplerCube u_EnvironmentMap1 )
 #else // !USE_REFLECTIVE_SPECULAR
-void computeDynamicLight( int idx, vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse,
+void computeDynamicLight( uint idx, vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse,
 	vec4 material, inout vec4 color )
 #endif // !USE_REFLECTIVE_SPECULAR
 {
@@ -211,20 +211,19 @@ const int lightsPerLayer = 16;
 
 #define idxs_t uvec4
 
+uniform usampler3D u_LightTilesInt;
+
+const uint numLayers = MAX_REF_LIGHTS / 256;
+const vec3 tileScale = vec3( r_tileStep, 1.0 / numLayers );
+
 idxs_t fetchIdxs( in vec3 coords, in usampler3D u_LightTilesInt ) {
 	return texture3D( u_LightTilesInt, coords );
 }
 
-int nextIdx( inout idxs_t idxs ) {
-	uvec4 tmp = ( idxs & uvec4( 3 ) ) * uvec4( 0x40, 0x10, 0x04, 0x01 );
-	idxs >>= 2;
-	return int( tmp.x + tmp.y + tmp.z + tmp.w );
+// 8 bits per light ID
+uint nextIdx( in uint count, in idxs_t idxs ) {
+	return ( idxs[count / 4] >> ( 8 * ( count % 4 ) ) ) & 0xFF;
 }
-
-uniform usampler3D u_LightTilesInt;
-
-const int numLayers = MAX_REF_LIGHTS / 256;
-const vec3 tileScale = vec3( r_tileStep, 1.0 / numLayers );
 
 #if defined(USE_REFLECTIVE_SPECULAR)
 void computeDynamicLights( vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse, vec4 material,
@@ -236,19 +235,26 @@ void computeDynamicLights( vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse, vec4
 #endif // !USE_REFLECTIVE_SPECULAR
 {
 	vec2 tile = floor( gl_FragCoord.xy * ( 1.0 / float( TILE_SIZE ) ) ) + 0.5;
+	
+	// NOT the amount of lights that can actually be put into each layer of the lighttile texture
+	uint globalLightsPerLayer = ( u_numLights + numLayers - 1 ) / numLayers;
 
-	#if defined(r_showLightTiles)
-		float numLights = 0.0;
-	#endif
-
-	for( int layer = 0; layer < numLayers; layer++ ) {
+	uint lightCount = 0;
+	for( uint layer = 0; layer < numLayers; layer++ ) {
 		idxs_t idxs = fetchIdxs( tileScale * vec3( tile, float( layer ) + 0.5 ), u_LightTilesInt );
-		for( int i = 0; i < lightsPerLayer; i++ ) {
-			int idx = numLayers * nextIdx( idxs ) + layer;
 
-			if( idx >= u_numLights ) {
+		uint lightOffset = layer * globalLightsPerLayer;
+		uint layerLightCount = lightOffset < u_numLights ? min( min( u_numLights - lightOffset, globalLightsPerLayer ), lightsPerLayer ) : 0;
+		for( uint i = 0; i < layerLightCount; i++ ) {
+			uint idx = nextIdx( lightCount, idxs );
+
+			if( idx == 0 ) {
 				break;
 			}
+
+			/* Light IDs are stored relative to the layer
+			Subtract 1 because 0 means there's no light */
+			idx += ( layer * numLayers ) + layer - 1;
 	  
 			#if defined(USE_REFLECTIVE_SPECULAR)
 				computeDynamicLight( idx, P, normal, viewDir, diffuse, material, color, u_EnvironmentMap0, u_EnvironmentMap1 );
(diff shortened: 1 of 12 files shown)


illwieckz:
I confirm that the alternative code for missing `EXT_texture_integer` is working, while the alternative code for missing `ARB_uniform_buffer_object` doesn't.

In my database of collected `glxinfo` dumps, I see ATI RV500 and RV400 cards that are capable of compiling the tiled renderer, but don't support either `EXT_texture_integer` and `ARB_uniform_buffer_object`. Olders ATI cards like RV300 are not able to compile the tiled renderer shaders anyway.  I also see Nvidia NV40 cards that miss both. So actually having an alternate code working for one extension is useless if the alternate code for the other extension is not working.

The glxinfo dumps for those cards are from 2020 so I may re-run them with latest mesa.

Intel gma 3 misses them but is not able to compile the tiled renderer shaders to begin with. The gma 4 can probably compile them but I have no dump and it would probably be super slow (I'll have to restart the related computer to check).

VReaperV:
Yeah, both extensions seem to really be supported by the same set of hardware.

src/engine/renderer/tr_shade.cpp
```diff
@@ -48,6 +48,16 @@ static void EnableAvailableFeatures()
 				glConfig2.realtimeLighting = false;
 			}
 
+			if ( !glConfig2.uniformBufferObjectAvailable ) {
```
slipher, line 45: These could be else-if's

slipher:
In the commit `Cleanup lighttile_fp`, there's a suspicious change `for( int i = u_lightLayer; i < u_numLights; i += numLayers ) {` -> `for( int i = u_lightLayer; i < u_numLights; i++ ) {` (counting up by 1 instead of 4). Is that what you mean by lights being added multiple times? This would make "most" lights be processed 4 times.

VReaperV:
> In the commit `Cleanup lighttile_fp`, there's a suspicious change `for( int i = u_lightLayer; i < u_numLights; i += numLayers ) {` -> `for( int i = u_lightLayer; i < u_numLights; i++ ) {` (counting up by 1 instead of 4). Is that what you mean by lights being added multiple times? This would make "most" lights be processed 4 times.

That one was probably an erroneus change. Lights being added multiple times is something I noticed with frame debugging (might've got it wrong with all the weird scattering of bits of the same ID across 4 components of a uvec4... Either way, now it's easy to confirm that it is working properly).

VReaperV:
Also, with the old code the same lights were being added to multiple layers.

src/engine/renderer/tr_shade.cpp
VReaperV: Perhaps, I just did the same thing as the original code.

VReaperV:
> In the commit `Cleanup lighttile_fp`, there's a suspicious change `for( int i = u_lightLayer; i < u_numLights; i += numLayers ) {` -> `for( int i = u_lightLayer; i < u_numLights; i++ ) {` (counting up by 1 instead of 4). Is that what you mean by lights being added multiple times? This would make "most" lights be processed 4 times.

I've now removed this from the commit.

illwieckz:
I also wondered if such alternative code would help if one day we optionally target OpenGL ES as a fallback, it looks like none of those extensions exist in OpenGL ES, so for such alternative code to work, we would need to fix the alternative code for `ARB_uniform_buffer_object` anyway.

illwieckz:
One interest in the alternate code to `EXT_texture_integer` is that it allows to use `RGBA8` textures instead of `RGBA32UI` ones and then, use less VRAM. Here with a 4K screen to make it obvious:

```
]/r_ext_texture_integer on
]/vid_restart
]/listImages _lighttileRender
------------------------------------------------------------------
num   width height layers mm  type format   wrap.t   wrap.s   name
------------------------------------------------------------------
51      240    135      4 no  3D   RGBA32UI t.clmp   s.clmp   _lighttileRender
------------------------------------------------------------------
129600 total texels (not including mipmaps)
1.97 MB total image memory (estimated)
817 total images
------------------------------------------------------------------
```

```
]/r_ext_texture_integer off
]/vid_restart
]/listImages _lighttileRender
------------------------------------------------------------------
num   width height layers mm  type format   wrap.t   wrap.s   name
------------------------------------------------------------------
51      240    135      4 no  3D   RGBA8    t.clmp   s.clmp   _lighttileRender
------------------------------------------------------------------
129600 total texels (not including mipmaps)
0.49 MB total image memory (estimated)
817 total images
------------------------------------------------------------------
```

illwieckz:
More detailed example:

```
]/r_highPrecisionRendering on
]/vid_restart
]/listImages _*tile*Render
------------------------------------------------------------------
num   width height layers mm  type format   wrap.t   wrap.s   name
------------------------------------------------------------------
49      960    540      1 no  2D   RGBA32F  t.1clmp  s.1clmp  _depthtile1Render
50      240    135      1 no  2D   RGBA32F  t.clmp   s.clmp   _depthtile2Render
51      240    135      4 no  3D   RGBA32UI t.clmp   s.clmp   _lighttileRender
------------------------------------------------------------------
680400 total texels (not including mipmaps)
10.38 MB total image memory (estimated)
817 total images
------------------------------------------------------------------
```

```
]/r_highPrecisionRendering off
]/vid_restart
]/listImages _*tile*Render   
------------------------------------------------------------------
num   width height layers mm  type format   wrap.t   wrap.s   name
------------------------------------------------------------------
49      960    540      1 no  2D   RGBA16F  t.1clmp  s.1clmp  _depthtile1Render
50      240    135      1 no  2D   RGBA16F  t.clmp   s.clmp   _depthtile2Render
51      240    135      4 no  3D   RGBA8    t.clmp   s.clmp   _lighttileRender
------------------------------------------------------------------
680400 total texels (not including mipmaps)
4.69 MB total image memory (estimated)
817 total images
------------------------------------------------------------------
```

VReaperV:
> I also wondered if such alternative code would help if one day we optionally target OpenGL ES as a fallback, it looks like none of those extensions exist in OpenGL ES, so for such alternative code to work, we would need to fix the alternative code for `ARB_uniform_buffer_object` anyway.

Wait, which extensions? UBOs and integer textures are supported in OpenGL ES, though as a core feature rather than extensions.

VReaperV:
> One interest in the alternate code to EXT_texture_integer is that it allows to use RGBA8 textures instead of RGBA32UI ones and then, use less VRAM. Here with a 4K screen to make it obvious:

I plan on making an alternative implementation/path that would use a UBO instead of a texture to avoid wasting so much memory. Right now we have a bunch of problems because of using a texture:
- a tile cannot hold more than 16 lights, any extra ones will not affect it
- we can't add more info to tiles (e. g. which cubeprobe to use; I plan to replace one light with a probe index) without either reducing the amount of lights per tile or using up a lot more memory
- most of time the time most of this data is unused

If we want to reduce the memory usage from this texture, we should just add a cvar that would set the amount of layers. The effect would be practically the same (RGBA32UI texture with 1 layer can hold the same amount of lights as an RGBA8 texture with 4 layers), but without the extra complexity and code paths.

VReaperV:
> More detailed example:

I don't think `r_highPrecisionRendering` should affect the lighttile texture anyway. After all, it's not about representing more or less colours/depth, but just about how many lights each tile can hold.

illwieckz:
> > I also wondered if such alternative code would help if one day we optionally target OpenGL ES as a fallback, it looks like none of those extensions exist in OpenGL ES, so for such alternative code to work, we would need to fix the alternative code for `ARB_uniform_buffer_object` anyway.
> 
> Wait, which extensions? UBOs and integer textures are supported in OpenGL ES, though as a core feature rather than extensions.

Ah, good news! I confess I went the lazy way and grepped https://mesamatrix.net/ but maybe it's not listed because of being a basic feature. 😄️

illwieckz:
> I don't think `r_highPrecisionRendering` should affect the lighttile texture anyway. After all, it's not about representing more or less colours/depth, but just about how many lights each tile can hold.

Oh I see! So that means we can NUKE more code.

If you go on NUKING the `r_highPrecisionRendering` test on light tile, better add a comment saying that “it's not about representing more or less colours/depth, but just about how many lights each tile can hold” to avoid future misunderstanding to occur again.

> I plan on making an alternative implementation/path that would use a UBO instead of a texture to avoid wasting so much memory.

OK so we would not avoid UBO anyway.

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -24,273 +24,253 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #define COMPUTELIGHT_GLSL
 
 #if !defined(USE_BSP_SURFACE)
-	#define USE_MODEL_SURFACE
+    #define USE_MODEL_SURFACE
```
illwieckz, line 27: It looks like you're actually rewriting the whole file so I prefer if you use this opportunity to switch to tabs.

src/engine/renderer/glsl_source/lighttile_fp.glsl
```diff
@@ -37,140 +37,121 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 
-struct light {
-  vec3  center;
-  float radius;
-  vec3  color;
-  float type;
-  vec3  direction;
-  float angle;
+struct Light {
+    vec3 center;
```
illwieckz, line 41: Same concern about tabs.

src/engine/renderer/glsl_source/computeLight_fp.glsl
VReaperV: Hmm, I thought I did replace it with tabs. Maybe I changed some setting in my IDE that converted it into spaces, I'll recheck this later.

VReaperV:
> > > I also wondered if such alternative code would help if one day we optionally target OpenGL ES as a fallback, it looks like none of those extensions exist in OpenGL ES, so for such alternative code to work, we would need to fix the alternative code for `ARB_uniform_buffer_object` anyway.
> > 
> > 
> > Wait, which extensions? UBOs and integer textures are supported in OpenGL ES, though as a core feature rather than extensions.
> 
> Ah, good news! I confess I went the lazy way and grepped https://mesamatrix.net/ but maybe it's not listed because of being a basic feature. 😄️

Yeah, it's part of the core spec. We can easily target OpenGL ES 3.2 since it's very widely supported, and lower versions might not have all the features we use.

VReaperV:
> If you go on NUKING the r_highPrecisionRendering test on light tile, better add a comment saying that “it's not about representing more or less colours/depth, but just about how many lights each tile can hold” to avoid future misunderstanding to occur again.

Yeah, can do. 

> OK so we would not avoid UBO anyway.

Yeah... I mean, is there any reason *to* avoid it? I would make it the only code path as well, but I'd need to either move the lighttile code to CPU side or bind it as an SSBO for writing in a shader. The latter would make it require more extensions, which is why that would be an alternative code path unless moved to CPU.

src/engine/renderer/glsl_source/computeLight_fp.glsl
VReaperV: Fixed.

src/engine/renderer/glsl_source/lighttile_fp.glsl
VReaperV: Fixed.

src/engine/renderer/tr_shade.cpp
illwieckz: I prefer to have twice the warnings.

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -235,36 +234,41 @@ void computeDynamicLights( vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse, vec4
...
-			#endif
+			lightCount++;
 		}
 	}
 
 	#if defined(r_showLightTiles)
-		if ( numLights > 0.0 ) {
-			color = vec4( numLights / ( lightsPerLayer * numLayers ), numLights / ( lightsPerLayer * numLayers ),
-				numLights / ( lightsPerLayer * numLayers ), 1.0 );
+		if ( lightCount > 0.0 ) {
```
illwieckz, line 269: No need to compare with a float if it's a uint I guess.

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -32,196 +32,199 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+
+	#if defined(USE_REFLECTIVE_SPECULAR)
+		// not implemented for PBR yet
+		vec4 envColor0 = textureCube(u_EnvironmentMap0, reflect( -viewDir, normal ) );
+		vec4 envColor1 = textureCube(u_EnvironmentMap1, reflect( -viewDir, normal ) );
+
+		materialColor.rgb *= mix( envColor0, envColor1, u_EnvironmentInterpolation ).rgb;
+	#endif // USE_REFLECTIVE_SPECULAR
+
+	color.rgb += lightColor.rgb * NdotL * diffuseColor.rgb;
```
illwieckz, line 125: This block should be moved one tab right.

src/engine/renderer/glsl_source/lighttile_fp.glsl
```diff
@@ -9,14 +9,14 @@ This file is part of the Daemon BSD Source Code (Daemon Source Code).
...
-      names of its contributors may be used to endorse or promote products
-      derived from this software without specific prior written permission.
+	* Redistributions of source code must retain the above copyright
+	  notice, this list of conditions and the following disclaimer.
+	* Redistributions in binary form must reproduce the above copyright
+	  notice, this list of conditions and the following disclaimer in the
+	  documentation and/or other materials provided with the distribution.
+	* Neither the name of the Daemon developers nor the
+	  names of its contributors may be used to endorse or promote products
+	  derived from this software without specific prior written permission.
```
illwieckz, line 19: The header uses spaces.

illwieckz (approved):
LGTM % the nitpicking.

src/engine/renderer/glsl_source/lighttile_fp.glsl
```diff
@@ -37,140 +37,121 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+	vec4 plane5 = vec4( 0.0, 0.0,  1.0,  minmax.y );
+	vec4 plane6 = vec4( 0.0, 0.0, -1.0, -minmax.x );
+
+	idxs_t idxs = uvec4( 0, 0, 0, 0 );
+
+	// NOT the amount of lights that can actually be put into each layer of the lighttile texture
+	uint globalLightsPerLayer = min( ( u_numLights + numLayers - 1 ) / numLayers, lightsPerLayer );
+	uint lightOffset = u_lightLayer * globalLightsPerLayer;
+	uint layerLightCount = lightOffset < u_numLights ? min( u_numLights - lightOffset, globalLightsPerLayer ) : 0;
```
slipher, line 123: This looks wrong: it makes it so that no more than 16 lights are ever considered. There are up to 256 input lights and they should all be considered as long as less than 16 lights have been output after culling.

slipher (commented):
What's the bug in the old code being referred to with `Process each realtime light only once per fragment`?

VReaperV:
> What's the bug in the old code being referred to with `Process each realtime light only once per fragment`?

Old code processed all of the values in lighttile image unless the light ID exceeded `u_numLights`, including ones that weren't set to any light by `lighttile_fp`.

slipher:
> Old code processed all of the values in lighttile image unless the light ID exceeded `u_numLights`, including ones that weren't set to any light by `lighttile_fp`.

There is supposed to be a 255 at the end of the list to trigger the id > numlights condition (the `#define idx_initializer uvec4(3)` bit). Although it wouldn't work correctly if there are more than 1020 (MAX_REF_LIGHTS - 4) lights.

VReaperV:
> > Old code processed all of the values in lighttile image unless the light ID exceeded `u_numLights`, including ones that weren't set to any light by `lighttile_fp`.
> 
> There is supposed to be a 255 at the end of the list to trigger the id > numlights condition (the `#define idx_initializer uvec4(3)` bit). Although it wouldn't work correctly if there are more than 1020 (MAX_REF_LIGHTS - 4) lights.

You're right, I checked and tested it again and that seems to be the case. I'll keep the changes though, since they should be a bit faster and would allow to easily change the maximum amount of lights per tile.

src/engine/renderer/glsl_source/lighttile_fp.glsl
VReaperV: Fixed.

src/engine/renderer/glsl_source/lighttile_fp.glsl
VReaperV: Fixed.

src/engine/renderer/glsl_source/computeLight_fp.glsl
VReaperV: Done.

src/engine/renderer/glsl_source/computeLight_fp.glsl
VReaperV: Right, done.

illwieckz:
Just for the knowledge, I just tested an R500 card (an ATI Radeon X1950 PRO, RV570), and this high-end-but-old R500 card features no `GL_EXT_texture_integer` neither `GL_ARB_uniform_buffer_object`. It's a pretty decent card that can achieve some 60 fps with a low preset plus deluxe/specular mapping on 720p screen:

unvanquished_2024-10-12_011252_000

_Note_: This framerate is expected to be way below what the card can do, this was tested with the R500 being a secondary card and not driving the display directly and requiring to copy the renderer to the primary card through PCIe, which is known to be slower on older cards. Four years ago I got 70fps with it on 1080p screen, and since that time both engine and drivers improved.

The high-end cards R400/R500 families are probably the only cards for which an alternative implementation for the tiled renderer makes sense. Other GL 2.1 cards are either too weak for the tiled renderer, or are powerful enough for it and have the features.

PR #1334, opened 2024-10-01
Replace new with R_GetRenderCommand()

VReaperV:
Same as every other usage of it.


slipher:
It's different for a reason. https://github.com/DaemonEngine/Daemon/blob/83244ed8442530e2776d4297b66dbb7a1ceacda6/src/engine/renderer/tr_cmds.cpp#L232

VReaperV:
That makes sense. Closing this now.

PR #1335, opened 2024-10-02
sdl_glimp: rewrite a bit some GL extension and workaround detection

illwieckz:
Rewrite a bit some GL extension and workaround detection. This basically unify the wording of both logs and the way it is implemented.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -125,6 +125,10 @@ static Cvar::Cvar<bool> r_khr_debug( "r_khr_debug",
 static Cvar::Cvar<bool> r_khr_shader_subgroup( "r_khr_shader_subgroup",
 	"Use GL_KHR_shader_subgroup if available", Cvar::NONE, true );
 
+static Cvar::Cvar<bool> workaround_glDriver_amd_adrenalin_disableBindlessTexture(
+	"workaround.glDriver.amd.adrenalin.disableBindlessTexture",
+	"Disable ARB_bindless_texture on AMD Adrenalin driver",
+	Cvar::NONE, true );
 static Cvar::Cvar<bool> workaround_glDriver_amd_oglp_disableBindlessTexture(
 	"workaround.glDriver.amd.oglp.disableBindlessTexture",
 	"Disable ARB_bindless_texture on AMD OGLP driver",
@@ -1480,9 +1484,9 @@ static void GLimp_CheckGLEW( const glConfiguration &requestedConfiguration )
 static bool IsSdlVideoRestartNeeded()
 {
 	/* We call RV600 the first generation of R600 cards, to make a difference
-	with RV700 and RV800 cards that are also supported by the Mesa R600 driver.
+	with RV700 and RV800 cards that are also supported by the Mesa r600 driver.
 
-	The Mesa R600 driver has broken Hyper-Z wth RV600, not RV700 nor RV800. */
+	The Mesa r600 driver has broken Hyper-Z wth RV600, not RV700 nor RV800. */
 	if ( workaround_glDriver_mesa_ati_rv600_disableHyperZ.Get() )
 	{
 		if ( getenv( "R600_HYPERZ" ) )
@@ -1490,13 +1494,14 @@ static bool IsSdlVideoRestartNeeded()
 			return false;
 		}
 
-		if ( !Q_stricmp( glConfig.vendor_string, "Mesa" ) || !Q_stricmp( glConfig.vendor_string, "X.Org" ) )
+		if ( glConfig2.driverVendor == glDriverVendor_t::MESA
+			&& glConfig2.hardwareVendor == glHardwareVendor_t::ATI )
 		{
 			bool foundRv600 = false;
 
 			std::string cardName = "";
 
-			static const std::vector<std::string> codenames = {
+			static const std::string codenames[] = {
 				// Radeon HD 2000 Series
 				"R600", "RV610", "RV630",
 				// Radeon HD 3000 Series
@@ -1516,7 +1521,7 @@ static bool IsSdlVideoRestartNeeded()
 
 			if ( foundRv600 )
 			{
-				logger.Warn( "...found buggy Mesa driver with %s card, Hyper-Z disabled", cardName );
+				logger.Warn( "Found buggy Mesa driver with %s card, disabling Hyper-Z.", cardName );
 
 				Sys::SetEnv( "R600_HYPERZ", "false" );
 
@@ -2029,7 +2034,7 @@ static void GLimp_InitExtensions()
 	}
 	glConfig2.shadingLanguageVersion = majorVersion * 100 + minorVersion;
 
-	logger.Notice("...found shading language version %i", glConfig2.shadingLanguageVersion );
+	logger.Notice("...using shading language version %i", glConfig2.shadingLanguageVersion );
 
 	// Texture formats and compression
 	glGetIntegerv( GL_MAX_CUBE_MAP_TEXTURE_SIZE_ARB, &glConfig2.maxCubeMapTextureSize );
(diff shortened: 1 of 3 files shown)


src/engine/sys/sdl_glimp.cpp
```diff
@@ -2285,77 +2286,84 @@ static void GLimp_InitExtensions()
...
-					if ( Sys::isRunningOnWine() )
-					{
-						// Detect AMD ATI driver on macOS not being used.
-						if ( !Q_stristr( glConfig.version_string, "ATI-" ) )
+				bool foundOglp = false;
+
+				if ( glConfig2.hardwareVendor == glHardwareVendor_t::ATI
+					&& glConfig2.driverVendor != glDriverVendor_t::MESA )
+				{
+					#if defined(_WIN32)
```
VReaperV, line 2341: I've received a report that it's also not working on the windows AMD proprietary driver (which is the Adrenalin one I guess).

src/engine/sys/sdl_glimp.cpp
```diff
@@ -2285,77 +2286,84 @@ static void GLimp_InitExtensions()
...
 						}
-					}
-				#elif defined(__APPLE__)
-					// AMD ATI driver for macOS is believed to not implement bindless texture.
-				#else
-					foundOglp = true;
-				#endif
-			}
+					#elif defined(__APPLE__)
+						// AMD ATI driver for macOS is believed to not implement bindless texture.
```
VReaperV, line 2352: AFAIK Apple is stuck on 4.1 completely due to their petty bullshit of trying to push Metal.

cu-kai: Yes bindless texture does not work on macOS and probably never will.

illwieckz: Maybe with Mesa zink over MoltenVK over Metal, but then it's Mesa, and then not the Apple AMD/ATI driver. 😄️

VReaperV: > Yes

Thanks for the confirmation.

This can then set `foundOglp` to `true`.

VReaperV: From dev channel:
> in fact, iirc, there is even a warning printed in the console about missing bindless texture extension when you launch unv on a mac

So looks like we don't even need to special-case Mesa on Apple here.

src/engine/sys/sdl_glimp.cpp
VReaperV: Apparently not Adrenalin, but I don't think that changes anything here, it should still be disabled on non-Mesa ATI, regardless of if running in Wine.

illwieckz: > Apparently not Adrenalin

What does that mean? Does that mean the AMD proprietary driver is not Adrenalin? or that Adrenalin is not buggy?

Is there a difference between Windows provided drivers and separate download from AMD?

VReaperV: No idea, I was told that Adrenain had to be uninstalled because it was apparently hard-locking in unv after some Adrenalin update, and that they're using the generic AMD driver. I ask to clarify later.

illwieckz: I assume Windows stock driver is just older Adrenalin, there is no reason to have two drivers there.

src/engine/sys/sdl_glimp.cpp
illwieckz: What I did is that I now detect what is the host system running wine. I don't test for anything being Apple anymore.

src/engine/sys/sdl_glimp.cpp
illwieckz: I split the workaround cvar between OGLP and Adrenalin because there is strong reasons to believe the current OGLP driver is the last one and will never be updated. While AMD has no Adrenalin alternative on Windows and then is very likely to update it.

But I make no difference between AMD stock driver on Windows and self-installed Adrenalin driver (I name them both Adrenalin).

illwieckz:
Do someone knows how to silence this warning? The code works but the warning breaks the MinGW CI.

```
src/common/System.cpp: In function ‘const char* Sys::getWineHostSystem()’:
src/common/System.cpp:93:49: warning: cast between incompatible function types from ‘FARPROC’
 {aka ‘long long int (*)()’} to ‘wghv_t’ {aka ‘long long int (*)(const char**, const char**)’} [-Wcast-function-type]
   93 |                 wghv_t pwine_get_host_version = (wghv_t) GetProcAddress( hNTDLL, "wine_get_host_version" );
```

src/engine/sys/sdl_glimp.cpp
VReaperV: Yeah, probably. To me the description they have on their website reads like Adrenalin is just a companion app like Geforce Experience, that makes getting a driver update faster, changing some setting etc.

Anyway, this should work fine.

illwieckz:
It looks like I found a trick for the function pointer casting.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -2285,77 +2290,98 @@ static void GLimp_InitExtensions()
...
+					{ "1.0", { "1.0-devel", "1.0-rc1", "1.0-rc2", "1.0-rc3", "1.0-rc4", "1.0", } },
+					{ "1.",  { "1.1", "1.2", "1.3", "1.4", } },
+					{ "2.0", { "2.0-devel", "2.0-rc1", } },
+				};
 
-			for ( auto& vp : versions ) {
-				mesaVersion = Str::Format( "Mesa 24.%s", vp.first );
+				for ( auto& vp : versions )
+				{
+					mesaVersion = Str::Format( "Mesa 24.%s", vp.first );
```
VReaperV, line 2314: Is there any reason not to just put all the more complete version numbers (`1.0-devel`, `1.0-rc1` etc.) into one array and check against them? The nested loop here seems redundant.

illwieckz: We can iterate all the numbers indeed, but I thought it would be faster to not test all strings so I went the divide & conquer way. I can make it simpler if considered better.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -2142,25 +2149,23 @@ static void GLimp_InitExtensions()
...
-			bool foundNvidia340 = ( glConfig2.driverVendor == glDriverVendor_t::NVIDIA
-			&& Q_stristr( glConfig.version_string, "NVIDIA 340." ) );
-
-			if ( foundNvidia340 )
+			if ( glConfig2.driverVendor == glDriverVendor_t::NVIDIA
+				&& Q_stristr( glConfig.version_string, " NVIDIA 340." ) )
 			{
 				// No need for WithoutSuppression for something which can only be printed once per renderer restart.
-				logger.Notice("...found buggy Nvidia 340 driver");
+				logger.Warn("Found buggy Nvidia 340 driver, disabling ARB_texture_gather.");
```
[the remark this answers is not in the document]
illwieckz: Fixed.

src/engine/sys/sdl_glimp.cpp
illwieckz: I rewrote the code to walk the string and not retest what is already tested. I kept the dichotomous search, and I updated the known version strings.

illwieckz:
For knowledge, this is what I do to know if a Mesa tree has the fixes for the bindless textures (starting with Mesa-24.2.0):

```
$ git log | grep -E 'glsl: make use of new tex src deref intrinsic|nir: add nir_tex_src_\{sampler,texture\}_deref_intrinsic|nir: create validate_tex_src_texture_deref\(\) helper|nir/glsl: set deref cast mode during function inlining'
    nir/glsl: set deref cast mode during function inlining
    glsl: make use of new tex src deref intrinsic
    nir: add nir_tex_src_{sampler,texture}_deref_intrinsic
    nir: create validate_tex_src_texture_deref() helper
```

The four lines should be printed. Because of backports in some branches I grep for commit names and not for commit references.

I updated the list of Mesa versions affected by the bindless texture bug. I actually tested a bunch of them to verify it's correct and all the tests gave me the expected resultes.

src/engine/sys/sdl_glimp.cpp
illwieckz: So for matching `24.2.0` it matches `Mesa 24.`, then `2.` then `0`.

VReaperV (approved):
LGTM.

PR #1338, opened 2024-10-03
GLSL cleanup: make var_FadeDepth a scalar

slipher:
Divide out the w coordinate in the vertex shader instead of the fragment shader.

Also add comment about hacky depth fraction calc.

### DIFF
--- a/src/engine/renderer/glsl_source/generic_fp.glsl
+++ b/src/engine/renderer/glsl_source/generic_fp.glsl
@@ -36,7 +36,7 @@ IN(smooth) vec2		var_TexCoords;
 IN(smooth) vec4		var_Color;
 
 #if defined(USE_DEPTH_FADE)
-IN(smooth) vec2         var_FadeDepth;
+IN(smooth) float        var_FadeDepth;
 uniform sampler2D       u_DepthMap;
 #endif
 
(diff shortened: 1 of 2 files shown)


src/engine/renderer/glsl_source/generic_fp.glsl
```diff
@@ -55,7 +55,14 @@ void	main()
 
 #if defined(USE_DEPTH_FADE) || defined(USE_VERTEX_SPRITE)
 	float depth = texture2D(u_DepthMap, gl_FragCoord.xy / r_FBufSize).x;
-	float fadeDepth = 0.5 * var_FadeDepth.x / var_FadeDepth.y + 0.5;
+
+	// convert z from normalized device coordinates [-1, 1]
+	// to window coordinates [0, 1]
+	float fadeDepth = 0.5 * var_FadeDepth + 0.5;
+
+	// HACK: the (distance from triangle to object behind it) / (shader's depthFade distance) ratio
```
VReaperV, line 63: Why not linearize it then? I assume the different fading based on near/far plane etc. is not intended.

slipher: Seemed like more effort than it's worth since I would have to plumb through the near/far values (which can apparently be different for every view) with a uniform. There's an `r_zNear` GLSL define, but it seems to have a tenuous (at best) connection with the value which is actually used.

VReaperV: Yeah, zFar is supposed to cover the whole map in whatever direction, while zNear can change for portals. `depthtile1_vp` does this, albeit with `r_zNear`, but if it's discreet enough then it's fine as is I guess.

slipher:
Retargeted this to 0.55 to avoid future merge conflicts.

PR #1340, opened 2024-10-03
Fix GL_ARB_bindless_texture not being loaded on non-Mesa drivers

VReaperV:


illwieckz:
That would have conflicted with https://github.com/DaemonEngine/Daemon/pull/1335

But since that was not the only mistake, I also implemented the bindless enablement fix in #1335 and fixed the other similar mistake.

VReaperV:
So this is to be considered obsolete then?

illwieckz:
yes, it is now meant to be obsoleted by #1335

VReaperV:
Okay, closing this then.

PR #1341, opened 2024-10-05
Lighting-related cleanup; make r_dynamicLight non-latch

slipher:
- Remove unused and unneeded lighting code
- Gate some tiled renderer-specific code behind the tiled cvar
- Make r_dynamicLight non-latch, to quickly toggle dynamic lights


VReaperV (commented):
If you're gonna nuke occlusion queries, then nuke the following as well, and all the related functionality:
https://github.com/DaemonEngine/Daemon/blob/14e7fb04fa9459e5248116212e201d2cf5137d36/src/engine/renderer/tr_backend.cpp#L2696
https://github.com/DaemonEngine/Daemon/blob/14e7fb04fa9459e5248116212e201d2cf5137d36/src/engine/renderer/tr_local.h#L2520
https://github.com/DaemonEngine/Daemon/blob/14e7fb04fa9459e5248116212e201d2cf5137d36/src/engine/renderer/tr_local.h#L179
https://github.com/DaemonEngine/Daemon/blob/14e7fb04fa9459e5248116212e201d2cf5137d36/src/engine/renderer/tr_local.h#L309 (this one is unused, but the comment at https://github.com/DaemonEngine/Daemon/blob/14e7fb04fa9459e5248116212e201d2cf5137d36/src/engine/renderer/tr_main.cpp#L2787 and the nature of this algorithm suggests it's part of occlusion queries too).
You might also want to get rid of the infinite zFar projection at https://github.com/DaemonEngine/Daemon/blob/14e7fb04fa9459e5248116212e201d2cf5137d36/src/engine/renderer/tr_main.cpp#L2788 but I haven't looked much into that part of the code.

In commit "Remove redundant realtime light enabled tests" I assume that "If a type of realtime light is **enabled**, just don't add it to the scene." is a typo.

VReaperV:
(moved from review since this would require targeting for-0.55) This: https://github.com/DaemonEngine/Daemon/blob/14e7fb04fa9459e5248116212e201d2cf5137d36/src/engine/renderer/tr_backend.cpp#L2696 apparently uses the same functionality and can be nuked as well later. It's only used for lightflares and seems to be pretty broken, so I assume it was just some proof of concept. Having to do extra IPC calls makes it even worse.

slipher:
I'm not planning to nuke vis tests. What's being deleted here is
1. An extension cvar r_ext_occlusion_query which has no effect. Although we do use GL_ARB_occlusion query functionality, this usage is not actually gated by the cvar. I updated the commit message to clarify that it's the cvar that is unused, not the extension.
2. The vis bounds expansion code which was apparently needed for a different occlusion test code path that was removed in 2014. The vis tests that we still have are used for a similar purpose (seeing if a "light" would shine on certain areas), the vis bounds change code wouldn't do anything useful for it because light flares are just textures, not dynamic lights.

VReaperV:
I see, thanks for the explanation.

slipher:
Regrettably, I have added more stuff to the PR:
- NUKE static lights
- Rename `r_dynamicLight` to `r_drawDynamicLights` and remove latch flag

src/engine/renderer/tr_light.cpp
```diff
@@ -1622,34 +1568,6 @@ void R_ComputeFinalAttenuation( shaderStage_t *pStage, trRefLight_t *light )
 =================
 R_CullLightTriangle
```
VReaperV, line 1569: Leftover comment.

slipher: It wasn't me. The banner comment for R_CullLightWorldBounds already says "R_CullLightTriangle"

slipher:
The default diff algorithm generates garbage for the static light nuking commit in tr_bsp.cpp. Reviewers may wish to view commits locally with a better algorithm, for example:
```
git fetch origin pull/1341/head
git -c diff.algorithm=patience show -9 FETCH_HEAD
```

src/engine/renderer/tr_cmds.cpp
```diff
@@ -252,6 +250,12 @@ R_AddSetupLightsCmd
 */
 void R_AddSetupLightsCmd()
 {
+	if ( !glConfig2.realtimeLighting ||
+	     r_realtimeLightingRenderer.Get() != Util::ordinal( realtimeLightingRenderer_t::TILED ) )
+	{
+		return;
+	}
```
illwieckz, line 257: Totally nitpicking but I prefer that syntax:

```c++
	if ( !glConfig2.realtimeLighting )
	{
		return;
	}
	
	if ( r_realtimeLightingRenderer.Get() != Util::ordinal( realtimeLightingRenderer_t::TILED ) )
	{
		return;
	}
```

This is the pattern we usually implement (and I usually write new code or modify code this way), for example:

https://github.com/DaemonEngine/Daemon/blob/926d8f6680af598efe49eea791a27567da3ad7f9/src/engine/renderer/tr_scene.cpp#L288-L301

illwieckz (approved):
LGTM.

src/engine/renderer/tr_cmds.cpp
slipher: I see it as one condition. "Is the tiled renderer active"

PR #1342, opened 2024-10-07
Fix heat haze shaders not rendered due to bad sorting

slipher:
All of the heat haze shaders in our weapon assets suffer from having a wrong sort order value (SS_OPAQUE). Only some more or less worked, due to the luck of being drawn later than most other opaque surfaces. It would seem the first properly working heat haze shader in our project was https://github.com/UnvanquishedAssets/map-plat23_src.dpkdir/pull/1: this shader received a SS_BLEND0 sort due to other stages.

Targeting 0.55 since some suddenly-visible assets probably need to be fixed. The Lucifer Cannon secondary fire impact looks stupid: the impacted surface suddenly assumes a distorted aspect, then 3 seconds later snaps back to normal.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5755,6 +5755,90 @@ static void ValidateStage( shaderStage_t *pStage )
 	}
 }
 
+// Note: this code was written to be exactly equivalent to previous sort determination code, which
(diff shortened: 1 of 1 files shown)


VReaperV (approved):
LGTM, assets can be adjusted later.

PR #1344, opened 2024-10-09
Runtime VBO layout: make half float usage optional

slipher:
Choose vertex attribute layouts at runtime. For attributes designated as wanting half float, use half float if GL_ARB_half_float_vertex is available or full-sized float otherwise,  #1179 explains the situation in detail.

TODO: grab the commit adding a *workaround* for the cards that support the half float vertex extension, but with poor quality of implementation.

The first 4 commits are from #1341.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -4572,8 +4572,8 @@ void DebugDrawVertex(const vec3_t pos, unsigned int color, const vec2_t uv) {
 	tess.verts[ tess.numVertexes ].xyz[ 2 ] = pos[ 2 ];
 	tess.verts[ tess.numVertexes ].color = colors;
 	if( uv ) {
-		tess.verts[ tess.numVertexes ].texCoords[ 0 ] = floatToHalf( uv[ 0 ] );
-		tess.verts[ tess.numVertexes ].texCoords[ 1 ] = floatToHalf( uv[ 1 ] );
+		tess.verts[ tess.numVertexes ].texCoords[ 0 ] = uv[ 0 ];
+		tess.verts[ tess.numVertexes ].texCoords[ 1 ] = uv[ 1 ];
 	}
 	tess.indexes[ tess.numIndexes ] = tess.numVertexes;
 	tess.numVertexes++;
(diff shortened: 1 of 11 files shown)


src/engine/renderer/tr_backend.cpp
```diff
@@ -4572,8 +4572,8 @@ void DebugDrawVertex(const vec3_t pos, unsigned int color, const vec2_t uv) {
 	tess.verts[ tess.numVertexes ].xyz[ 2 ] = pos[ 2 ];
 	tess.verts[ tess.numVertexes ].color = colors;
 	if( uv ) {
-		tess.verts[ tess.numVertexes ].texCoords[ 0 ] = floatToHalf( uv[ 0 ] );
-		tess.verts[ tess.numVertexes ].texCoords[ 1 ] = floatToHalf( uv[ 1 ] );
+		tess.verts[ tess.numVertexes ].texCoords[ 0 ] = uv[ 0 ];
```
illwieckz, line 4575: This can use Vector2Copy.

src/engine/renderer/tr_surface.cpp
```diff
@@ -1056,8 +1056,8 @@ static void Tess_SurfaceMD5( md5Surface_t *srf )
 			VectorCopy( position, tessVertex->xyz );
 
 			// FIXME: don't store data as half float?
-			tessVertex->texCoords[ 0 ] = halfToFloat( surfaceVertex->texCoords[ 0 ] );
-			tessVertex->texCoords[ 1 ] = halfToFloat( surfaceVertex->texCoords[ 1 ] );
+			tessVertex->texCoords[ 0 ] = surfaceVertex->texCoords[ 0 ];
```
illwieckz, line 1059: Is the comment above still up-to-date?

src/engine/renderer/tr_backend.cpp
illwieckz: An all the other occurrences too.

illwieckz:
OK I've finished reading everything. I'll do some tests on r300 hardware to see the performances when I disable the extension.

src/engine/renderer/tr_backend.cpp
slipher: Done.

illwieckz:
For some unnknown reasons I don't reproduce the 55fps I had on default Vega scene on one of the slowest r300 ever [some months ago](https://github.com/DaemonEngine/Daemon/issues/1172#issuecomment-2146467761), I only get 37 fps today.

But the good news is that I also get 37 fps with my own branch #1179, so that's not your branch's fault. Either we merged something bad in between, either I got a regression in the system (I updated from Ubuntu 23.10 to Ubuntu 24.04 since the last time and then I got updated systems libs including Mesa).

I confirm this unlocks the fps on the hardware and does the job, as with half float vertexes I get 5 fps on the same vega scene.

I can't test right now on Intel GMA3 as I'm not next to it and probably not before some months, but I assume it's fine as well.

So, LGTM.

illwieckz:
The 37fps was likely due to some bad options, I reapplied the `lowest` preset and I now get 75fps:

unvanquished_2024-10-10_061046_000

illwieckz:
In fact the 75fps was caped by the monitor refresh rate. For some reasons with that driver the vsync is always forced, I had to write some Xorg config to unlock it.

I get 90fps:

unvanquished_2024-10-10_132342_000

I even get 31fps with a 720p resolution:

unvanquished_2024-10-10_133359_000

This is the lowest-end range of hardware among the slowest family of ATI cards that are supported, and it's possible to play at 90fps.

For the curious people the Xorg trick to unlock the framerate is this file `/etc/X11/xorg.conf.d/00-disable-vsync.conf`:

```
Section "Device"
	Identifier  "Card0"
	Driver "All"
	Option "SwapbuffersWait" "false"
EndSection
```

slipher:
So now we're just waiting on #1341.

PR #1345, opened 2024-10-09
A branch that enables static lights with every light modes

illwieckz:
Includes:

- https://github.com/DaemonEngine/Daemon/pull/1343

Some other commits can be moved to other PRs.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -701,9 +701,6 @@ static std::string GenEngineConstants() {
 		AddDefine( str, "r_realtimeLightingRenderer", r_realtimeLightingRenderer.Get() );
 	}
 
-	if ( r_precomputedLighting->integer )
-		AddDefine( str, "r_precomputedLighting", 1 );
-
 	if ( r_showNormalMaps->integer )
 	{
 		AddDefine( str, "r_showNormalMaps", 1 );
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -75,8 +75,20 @@ void ReadLightGrid(in vec4 texel, out vec3 ambientColor, out vec3 lightColor) {
 }
 #endif
 
-#if defined(USE_DELUXE_MAPPING) || defined(USE_GRID_DELUXE_MAPPING) || (defined(r_realtimeLighting) && r_realtimeLightingRenderer == 1)
+#if defined(USE_DELUXE_MAPPING) || defined(USE_GRID_DELUXE_MAPPING) || defined(r_realtimeLighting)
+#if !defined(USE_PHYSICAL_MAPPING)
+#if defined(r_specularMapping)
 uniform vec2 u_SpecularExponent;
+
+vec3 computeSpecularity(vec3 lightColor, vec4 materialColor, float NdotH)
+{
+	return lightColor * materialColor.rgb * pow(NdotH, u_SpecularExponent.x * materialColor.a + u_SpecularExponent.y) * r_SpecularScale;
+}
+#endif
+#endif
+#endif
+
+#if defined(USE_DELUXE_MAPPING) || defined(USE_GRID_DELUXE_MAPPING) || (defined(r_realtimeLighting) && r_realtimeLightingRenderer == 1)
 
 #if defined(USE_REFLECTIVE_SPECULAR)
 void computeDeluxeLight( vec3 lightDir, vec3 normal, vec3 viewDir, vec3 lightColor,
@@ -152,7 +164,7 @@ void computeDeluxeLight( vec3 lightDir, vec3 normal, vec3 viewDir, vec3 lightCol
   color.rgb += lightColor.rgb * NdotL * diffuseColor.rgb;
 #if defined(r_specularMapping)
   // The minimal specular exponent should preferably be nonzero to avoid the undefined pow(0, 0)
-  color.rgb += lightColor.rgb * materialColor.rgb * pow( NdotH, u_SpecularExponent.x * materialColor.a + u_SpecularExponent.y) * r_SpecularScale;
+  color.rgb += computeSpecularity(lightColor.rgb, materialColor, NdotH);
 #endif // r_specularMapping
 #endif // !USE_PHYSICAL_MAPPING
 }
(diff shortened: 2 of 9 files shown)


illwieckz:
For unknown reasons that doesn't work with the legacy forward lighting renderer, the renderer for which we have a lot of dedicated static light code, it only works with the tiled renderer.

illwieckz:
Hu no, last commit breaks it.

illwieckz:
Ah no, it probably works, it's likely that stahl had no `noRadiosity` lights but some bug processed all lights.

illwieckz:
If I hack the engine to consider all lights to be processed over a black lightmap, and I build chasm with `-keeplights`, I get this:

unvanquished_2024-10-09_164425_000

Those may be the only light entities in chasm and sun/light emitting shaders is not considered:

20241009-164258-000 netradiant-chasm-lights

slipher:
It's titled "A branch that enables static lights with every light modes", but AFAICT it doesn't enable static lights in any case they weren't enabled before.

Do you have a test asset? There are no known maps with `noRadiosity` lights.

We have two possible use cases in mind, (A) approximating q3map2 lighting with the engine without lightmaps, and (B) adding a realtime light that is not considered by q3map2. I believe the current implementation does not have the parameters calibrated to the same strength as q3map2 like it would need to work for (A). So there's not really any value there. We could potentially use it for (B) as-is, but then that would mess up case (A) later: if we adjusted the parameters. So even if we are interested in (A), (B), or both, I still think we should NUKE the current implementation.

If you want to have a map realtime light entity that we can start using right now, I would suggest some other entity name instead of `light`. Then the two cases would not step on each other. Also it would make sense that compiling without `-keeplights` strips q3map2-calculated lights, but not the purely realtime lights.

illwieckz:
The static lights now work with the legacy forward renderer:

unvanquished_2024-10-09_165522_000

illwieckz:
@slipher you can use that chasm map compiled with `-keeplights`:

- https://dl.illwieckz.net/b/unvanquished/wip/static-lights/map-chasm_1.2.3-20230502-160927-5bead47-keeplights.dpk

I pushed to the branch a HACK that makes the engine process all static lights even if they don't have the `noRadiosity` property, and renders all lights over a dark lightmap.

illwieckz:
An example of use case for a `noRadiosity` light can be what we see in `test-dlight` map (there should be a build of it [there](https://github.com/UnvanquishedAssets/UnvanquishedTestAssets/releases/tag/v2.1).

The `test-dlight` map features a train map, the player plays inside the train, and the tunnel wall texture is moving to give the illusion the train is moving. Some dynamic lights on the tunnel walls move at the same speed of the tunnel wall textures, adding more to the illusion. Currently the light is done with a particle, but it could be done with a static bsp light instead.

I guess we may move the light entity, and even maybe disable it (or at least moving it outside the playable area), making possible for map lights that can be moved or disabled/enabled in game without relying on particle systems.

I agree that to make them useful, those lights intensities should somewhat match what q3map2 produces.

illwieckz:
Those lights are not really useful as an alternative to q3map2 as they traverse the geometry.

The entry corridor on the left is lit from the outside top, the red light on the right is from within the pipes:

unvanquished_2024-10-09_171655_000

illwieckz:
> You couldn't do that with these renderer `light` entities. They can't move. If you want an entity that can move, an sgame-level entity would be a better idea.

Oh, so the whole if `staticLight else if dynamicLight` is about optimizing BSP lights that can move? If yes we better consider everything a dynamic light and allow us to move BSP lights!

Maybe the ability to enable/disable them with non-movable BSP lights can be done though?

illwieckz:
I meant “optimizing BSP lights that can't move” (now corrected).

slipher:
> Those lights are not really useful as an alternative to q3map2 as they traverse the geometry.
> 
> The entry corridor on the left is lit from the outside top, the red light on the right is from within the vents:

Lol I guess use case (A) has to wait for ray tracing.

> Oh, so the whole `if staticLight else if dynamicLight` is about optimizing BSP lights that can't move?

Yes, 90% of static light code is about optimizing lights that can't move in the deprecated forward renderer.

It seems that most cases where you'd want a dynamic light as part of the map involve the light changing in some way, for example moving or being toggled on and off. Light entities parsed by the renderer cannot do this as they can't interact with game state. It seems that we would be better served by a light entity managed by the sgame (as any normal entity), rather than the "static light" code.

illwieckz:
OK, so reusing the curent dynamic light code for every light including possible BSP ones would be a better option. Maybe we don't need to “optimize” something for which the only thing doable is to do less good than lightmap.

illwieckz:
unvanquished_2024-10-09_180844_000

illwieckz:
> > Those lights are not really useful as an alternative to q3map2 as they traverse the geometry.
> > The entry corridor on the left is lit from the outside top, the red light on the right is from within the vents:
> 
> Lol I guess use case (A) has to wait for ray tracing.

The light bleed may be cancelled by shadows. 🤪️

illwieckz:
> > > Those lights are not really useful as an alternative to q3map2 as they traverse the geometry.
> > > The entry corridor on the left is lit from the outside top, the red light on the right is from within the vents:
> > 
> > 
> > Lol I guess use case (A) has to wait for ray tracing.
> 
> The light bleed may be cancelled by shadows. 🤪️

Actually doing that would require to uncaulk all invisible surfaces that should obstruct light…

unvanquished_2024-10-09_181614_000

unvanquished_2024-10-09_181622_000

illwieckz:
The fact only world shadows work may indicate that maybe only `R_AddPrecachedWorldInteractions` works…

VReaperV:
> I meant “optimizing BSP lights that can't move” (now corrected).

I don't really see what you could optimise there, other than just not re-uploading the data if it hasn't changed.

slipher:
> > I meant “optimizing BSP lights that can't move” (now corrected).
> 
> I don't really see what you could optimise there, other than just not re-uploading the data if it hasn't changed.

That's part of it (creating VBOs). Also the forward renderer makes a list of surfaces in range of the light, so it saves this list.

VReaperV:
I don't really see what lights would use VBOs for. A list of affected surfaces would also be useless for tiled lights, although it can be of use for shadows.

illwieckz:
This branch was basically a demo to investigate what this code was. This code is now NUKED.

To fetch this branch (even after deletion), one can do:

```sh
git fetch https://github.com/DaemonEngine/Daemon.git pull/1345/head
```

PR #1346, opened 2024-10-10
glsl: fix specular mapping with forward lighting

illwieckz:
Better merge this before so I can rebase over it:

- https://github.com/DaemonEngine/Daemon/pull/1333

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -48,9 +48,20 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	}
 #endif
 
-#if defined(USE_DELUXE_MAPPING) || defined(USE_GRID_DELUXE_MAPPING) || (defined(r_realtimeLighting) && r_realtimeLightingRenderer == 1)
+#if defined(USE_DELUXE_MAPPING) || defined(USE_GRID_DELUXE_MAPPING) || defined(r_realtimeLighting)
+#if !defined(USE_PHYSICAL_MAPPING)
+#if defined(r_specularMapping)
 	uniform vec2 u_SpecularExponent;
 
+vec3 computeSpecularity(vec3 lightColor, vec4 materialColor, float NdotH)
+{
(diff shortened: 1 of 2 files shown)


slipher:
What's the difference? Formula looks the same.

illwieckz:
The difference is that `uniform vec2 u_SpecularExponent;` is now done when `if /* … */ || defined(r_realtimeLighting)` is true (when realtime lighting is enabled, whatever the light renderer), previously it was only done if realtime lighting was enabled and light renderer was the tiled one, so compilation failed. Then I factorized the code using the said `u_SpecularExponent`.

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -75,9 +75,21 @@ void ReadLightGrid(in vec4 texel, out vec3 ambientColor, out vec3 lightColor) {
 }
 #endif
 
-#if defined(USE_DELUXE_MAPPING) || defined(USE_GRID_DELUXE_MAPPING) || (defined(r_realtimeLighting) && r_realtimeLightingRenderer == 1)
+#if defined(USE_DELUXE_MAPPING) || defined(USE_GRID_DELUXE_MAPPING) || defined(r_realtimeLighting)
+#if !defined(USE_PHYSICAL_MAPPING)
+#if defined(r_specularMapping)
 uniform vec2 u_SpecularExponent;
```
illwieckz, line 81: The fix is that this `uniform vec2 u_SpecularExponent;` line is now used for both the forward and the tiled light renderer.

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -75,9 +75,21 @@ void ReadLightGrid(in vec4 texel, out vec3 ambientColor, out vec3 lightColor) {
...
 
+vec3 computeSpecularity(vec3 lightColor, vec4 materialColor, float NdotH)
+{
+	return lightColor * materialColor.rgb * pow(NdotH, u_SpecularExponent.x * materialColor.a + u_SpecularExponent.y) * r_SpecularScale;
```
illwieckz, line 85: This refactor makes sure the code using `u_SpecularExponent` is right next to the declaration of it, within the same `ifdef`, so it reduces the risk to reintroduce the bug in the future.

illwieckz:
Now rebased.

slipher:
LGTM

PR #1347, opened 2024-10-11
Detect RV300 and VC4 hardware and disable slow half float vertex on them

illwieckz:
Import from:

- https://github.com/DaemonEngine/Daemon/pull/1179

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -137,9 +137,17 @@ static Cvar::Cvar<bool> workaround_glDriver_mesa_ati_rv300_disableRgba16Blend(
 	"workaround.glDriver.mesa.ati.rv300.disableRgba16Blend",
 	"Disable misdetected RGBA16 on Mesa driver on RV300 hardware",
 	Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_glDriver_mesa_ati_rv300_useFloatVertex(
+	"workaround.glDriver.mesa.ati.rv300.useFloatVertex",
+	"Use float vertex instead of supported-but-slower half-float vertex on Mesa driver on ATI RV300 hardware",
+	Cvar::NONE, true );
 static Cvar::Cvar<bool> workaround_glDriver_mesa_ati_rv600_disableHyperZ(
 	"workaround.glDriver.mesa.ati.rv600.disableHyperZ",
 	"Disable Hyper-Z on Mesa driver on RV600 hardware",
+	Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_glDriver_mesa_broadcom_vc4_useFloatVertex(
+	"workaround.glDriver.mesa.broadcom.vc4.useFloatVertex",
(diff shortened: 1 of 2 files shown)


illwieckz:
I also added VC4 detection, it's twice faster with float vertex on it.

Before (25fps):

unvanquished_2024-10-11_183704_000

After (55fps):

unvanquished_2024-10-11_184004_000

src/engine/sys/sdl_glimp.cpp
```diff
@@ -2210,10 +2214,65 @@ static void GLimp_InitExtensions()
...
+	instead. */
+	{
+		bool halfFloatVertexEnabled = r_arb_half_float_vertex.Get();
+
+		if ( glConfig2.driverVendor == glDriverVendor_t::MESA
+			&& glConfig2.hardwareVendor == glHardwareVendor_t::ATI )
+		{
+			bool foundRv300 = false;
+
+			std::string cardName = "unknown ATI RV3xx";
```
VReaperV, line 2231: The initial value is unused.

illwieckz: It is used if I forgot to list a card name and the `glConfig.hardwareType == glHardwareType_t::GLHW_R300 && glConfig2.maxAluInstructions == 64` test matches.

slipher: It isn't. It's assigned the first time the for loop is entered.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -2210,10 +2214,65 @@ static void GLimp_InitExtensions()
...
+			}
+		}
+
+		// VAO and VBO
+		// made required in OpenGL 3.0
+		glConfig2.halfFloatVertexAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_CORE, ARB_half_float_vertex, halfFloatVertexEnabled );
+
+		if ( !halfFloatVertexEnabled )
+		{
+			logger.Warn( "Missing half-float vertex, using float vertex instead." );
```
slipher, line 2273: How about Notice for this one. We shouldn't use Warn so much unless there is a serious problem.

illwieckz: I like to have a bolder log for such things, especially when it is reporting about making decisions, it makes it easier to catch them in logs. Though I agree Warn is maybe too much.

src/engine/sys/sdl_glimp.cpp
illwieckz: Oh I see. Right.

illwieckz:
I addressed the various comments.

slipher:
> I also added VC4 detection, it's twice faster with float vertex on it.

Forgot to push something? I don't see any new stuff.

illwieckz:
> > I also added VC4 detection, it's twice faster with float vertex on it.
> 
> Forgot to push something? I don't see any new stuff.

Indeed, now fixed.

slipher:
LGTM

PR #1348, opened 2024-10-11
Report more texture size and handle case when there is no 3D texture support

illwieckz:
3D texture support is not optional in OpenGL _but_ some low-end OpenGL ES devices may lack hardware support for it and Linux systems may still implement OpenGL for them.

Mesa is likely to provide such incomplete OpenGL drivers in a “better than nothing” way, this to unlock out-of-the-box OpenGL compatibility with mainstream OpenGL software that don't use any 3D textures. This includes window managers, media players, visualisation software and some production software not using 3D textures, etc. For example I guess NetRadiant would work out of the box on such incomplete OpenGL driver. 

This can happen with some Arm boards, especially since some Arm platforms may have targeted first Android phones before being integrated in computer boards. A good example of device with no hardware support for 3D texture but Mesa providing an as-much-complete-as-possible OpenGL driver for it is the Broadcom VC4 used in some Raspberry Pi.

We have 4 usages of 3D textures:

1. optional tiled lighting renderer (not needed to render the game)
2. optional colorgrading (not needed to render the game)
3. optional grid-based deluxe mapping for models (not needed to render the game)
4. light grid

Light grid is supposedly the only way to lit models with light that matches the position of the model in the map. If 3D textures are not supported, the models will be fullbright. This can be considered a cheat if intentional (models will not be black in dark areas), so I don't plan to give the user a way to toggle this, and this is not something we should enable in a preset. This is a fallback compatibility feature, and I consider it's fine to have it because if you actually play on such low-end hardware, you have stronger handicaps to deal with (tiny screen resolution, low framerate…). Using such hardware is hard to consider an unfair advantage.

I may add a cheat cvar to test the fallback though.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -3969,6 +3969,20 @@ R_LoadLightGrid
 */
 void R_LoadLightGrid( lump_t *l )
 {
+	if ( glConfig2.max3DTextureSize == 0 )
+	{
+		Log::Warn( "Grid lighting disabled because of missing 3D texture support." );
+
+		if ( glConfig2.deluxeMapping )
+		{
+			Log::Warn( "Grid deluxe mapping disabled because of missing 3D texture support." );
+		}
+
+		tr.lightGrid1Image = nullptr;
+		tr.lightGrid2Image = nullptr;
+		return;
+	}
+
 	int            i, j, k;
 	vec3_t         maxs;
 	world_t        *w;
@@ -5036,15 +5050,13 @@ void RE_LoadWorldMap( const char *name )
 	//R_BuildCubeMaps();
 
 	tr.worldLight = tr.lightMode;
+	tr.modelLight = lightMode_t::FULLBRIGHT;
 	tr.modelDeluxe = deluxeMode_t::NONE;
 	tr.mapLightFactor = 1.0f;
 	tr.mapInverseLightFactor = 1.0f;
 
-	if ( tr.worldLight == lightMode_t::FULLBRIGHT )
-	{
-		tr.modelLight = lightMode_t::FULLBRIGHT;
-	}
-	else
+	// Use fullbright lighting for everything if the world is fullbright.
+	if ( tr.worldLight != lightMode_t::FULLBRIGHT )
 	{
 		if ( tr.worldLight == lightMode_t::MAP )
 		{
@@ -5058,6 +5070,14 @@ void RE_LoadWorldMap( const char *name )
 				to match the color of the nearby lightmaps. We better not want to use
 				the grid light as a fallback as it would be close but not close enough. */
 
+				tr.worldLight = lightMode_t::VERTEX;
+			}
+		}
+		else if ( tr.worldLight == lightMode_t::GRID )
+		{
+			if ( !tr.lightGrid1Image )
+			{
+				// Use vertex light on world surface if light color grid is missing.
 				tr.worldLight = lightMode_t::VERTEX;
 			}
 		}
@@ -5086,19 +5106,10 @@ void RE_LoadWorldMap( const char *name )
 		algorithm for emulating the deluxe map from light direction grid.
 		See https://github.com/DaemonEngine/Daemon/issues/32 */
 
-		// Game model surfaces use grid lighting, they don't have vertex light colors.
-		tr.modelLight = lightMode_t::GRID;
-
-		if ( !tr.lightGrid1Image )
+		if ( tr.lightGrid1Image )
 		{
-			// Use fullbright light on game models if light color grid is missing.
-			tr.modelLight = lightMode_t::FULLBRIGHT;
-
-			if ( tr.worldLight == lightMode_t::GRID )
-			{
-				// Use vertex light on world surface if light color grid is missing.
-				tr.worldLight = lightMode_t::VERTEX;
-			}
+			// Game model surfaces use grid lighting, they don't have vertex light colors.
+			tr.modelLight = lightMode_t::GRID;
 		}
 
 		if ( glConfig2.deluxeMapping )
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -2047,8 +2047,46 @@ static void GLimp_InitExtensions()
 
 	logger.Notice("...using shading language version %i", glConfig2.shadingLanguageVersion );
 
-	// Texture formats and compression
+
+	// OpenGL driver constants.
+
+	glGetIntegerv( GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS, &glConfig2.maxTextureUnits );
+	glGetIntegerv( GL_MAX_TEXTURE_SIZE, &glConfig.maxTextureSize );
+	glGetIntegerv( GL_MAX_3D_TEXTURE_SIZE, &glConfig2.max3DTextureSize );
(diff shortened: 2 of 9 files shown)


illwieckz:
Here on a VC4 (Raspberry Pi 3B+):

unvanquished_2024-10-11_184444_000

src/engine/sys/sdl_glimp.cpp
```diff
@@ -2033,9 +2033,27 @@ static void GLimp_InitExtensions()
...
+
+	// OpenGL driver constants.
+
+	glGetIntegerv( GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS, &glConfig2.maxTextureUnits );
+	glGetIntegerv( GL_MAX_TEXTURE_SIZE, &glConfig.maxTextureSize );
+	glGetIntegerv( GL_MAX_3D_TEXTURE_SIZE, &glConfig2.max3DTextureSize );
 	glGetIntegerv( GL_MAX_CUBE_MAP_TEXTURE_SIZE_ARB, &glConfig2.maxCubeMapTextureSize );
 
+	// Stubbed or broken drivers may have reported garbage.
+	glConfig.maxTextureSize = std::max( 0, glConfig.maxTextureSize );
```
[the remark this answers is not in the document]
illwieckz: I have no idea, but that can't break anything I guess.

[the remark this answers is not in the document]
slipher: If it's actually true that drivers report garbage, then we shouldn't disable features on the basis of said garbage.

Is it possible to check whether the 3D texture creation succeeded by any chance? If not maybe we'd be better off with detecting hardware/drivers.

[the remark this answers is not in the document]
illwieckz: I removed that part. That very likely was a workaround for broken OpenGL 1 drivers in the nineties, drivers that we will never support for hardware that we will never support.

This workaround with this comment was already there in commit
https://github.com/id-Software/Quake-III-Arena/commit/dbe4ddb10315479fc00086f08e25d968b4b43c49 “_The Quake III Arena sources as originally released under the GPL license on August 20, 2005._”

If a driver in 2024 mistakenly reports 0 there (or a negative number we would clamp to 0), we would have a deepest problems than non-working lightgrid: no one texture would be renderable and I even don't know if the engine would survive this.

> Is it possible to check whether the 3D texture creation succeeded by any chance?

Yes that would be the same `GL_PROXY` mechanism that I use in this WIP PR (this one is for downscaling images if size is not accepted):

-  https://github.com/DaemonEngine/Daemon/pull/1131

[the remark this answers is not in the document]
slipher: > > Is it possible to check whether the 3D texture creation succeeded by any chance?
> 
> Yes that would be the same `GL_PROXY` mechanism that I use in this WIP PR (this one is for downscaling images if size is not accepted):

I meant when you create it for real, not with the proxy mechanism.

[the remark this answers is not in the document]
illwieckz: Hmm, it looks like I never submitted the comment I wrote. I was basically saying that I would prefer to not purposely use some feature that doesn't work and have the driver cry to the user about it.

For example Mesa currently prints things like that if we try to do 3D textures on VC4:

```
Mesa: User error: GL_INVALID_VALUE in glTexImage3D(level=0)
Mesa 22.3.6 implementation error: Bad t->Target in _mesa_test_texobj_completeness
Please report at https://gitlab.freedesktop.org/mesa/mesa/-/issues
```

Actually that third line telling the user to report a bug is wrong (the feature is not there, it will never be fixed, the bug is in the code attempting to use an unsupported feature), but the first line is totally legit, the user should not do something the driver/card doesn't support.

Also I experienced some crash with some driver versions before I've modified the engine to not attempt to do those things the hardware can't do. Of course it would be better if the driver would be more robust, but I cannot blame it for not behaving correctly if the mistake is purposely made by us.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -4007,6 +4007,20 @@ static void R_LoadFogs( lump_t *l, lump_t *brushesLump, lump_t *sidesLump )
 
 static void R_SetDefaultLightGrid()
 {
+	if ( glConfig2.max3DTextureSize == 0 )
```
slipher, line 4010: SetDefaultLightGrid is only called by LoadLightGrid, so this code is unreachable.

src/engine/renderer/ShadeCommon.h
```diff
@@ -156,6 +156,16 @@ template<typename Obj> void SetLightDeluxeMode( Obj* obj,
 		lightMode = tr.modelLight;
 		deluxeMode = tr.modelDeluxe;
 	}
+
+	if ( lightMode == lightMode_t::GRID && !tr.lightGrid1Image )
```
slipher, line 160: Why don't you set `tr.modelLight` instead of checking it every time?

src/engine/sys/sdl_glimp.cpp
```diff
@@ -2033,9 +2033,21 @@ static void GLimp_InitExtensions()
...
+	// OpenGL driver constants.
+
+	glGetIntegerv( GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS, &glConfig2.maxTextureUnits );
+	glGetIntegerv( GL_MAX_TEXTURE_SIZE, &glConfig.maxTextureSize );
+	glGetIntegerv( GL_MAX_3D_TEXTURE_SIZE, &glConfig2.max3DTextureSize );
 	glGetIntegerv( GL_MAX_CUBE_MAP_TEXTURE_SIZE_ARB, &glConfig2.maxCubeMapTextureSize );
 
+	logger.Notice( "...using up to %d texture size.", glConfig.maxTextureSize );
+	logger.Notice( "...using up to %d 3D texture size.", glConfig2.max3DTextureSize );
+	logger.Notice( "...using up to %d CubeMap texture size.", glConfig2.maxCubeMapTextureSize );
```
slipher, line 2046: Weird capitalization here. "Cubemap" is usually treated as 1 word

illwieckz: Well, we have both:

```
$ rg '[Cc]ubeMap' | wc -l
87
$ rg '[Cc]ubemap' | wc -l
110
```

We can unify that later.

slipher: I mean the capitalization should never be like that if you are writing text, not a function name or whatever. If you want it to be two words then "cube map".

illwieckz:
It looks like we will need some `std::max`…

```
GL_MAX_3D_TEXTURE_SIZE: -2147483648
```

slipher:
So drivers do return garbage after all?

In that case I think we should not merge the 3d texture thing for 0.55. I would hate to see lighting being disabled based on some faulty capability query thing.

illwieckz:
This driver returning garbage is that because this driver doesn't implement 3D textures at all. I'll report the bug on their side so they can at least return `0` here. You really should not be afraid of full-featured drivers returning garbage.

Our code already downscales 2D and Cube textures down to those limits. If drivers were reporting negative values, we would already know it!

Those checks are basically for less tested or incomplete drivers, it means they are used when we actually need them. We should not expect major drivers to return garbage, and if they do, not only we would already know about it, but we would have other problems than lights being fullbright.

src/engine/sys/sdl_glimp.cpp
illwieckz: I fixed the capitalization.

PR #1349, opened 2024-10-11
Revert computeLight_fp refactoring (worsens light layer distribution)

slipher:
#1333 changed how the dynamic lights are distributed into layers: before, the layer was set by the light's index modulo 4. After, the first quarter of lights are in layer 0, the 2nd quarter in layer 1, etc. The distribution is (hopefully) 'random' with the previous algorithm. With the new algorithm they are grouped by when they are added. So the new algorithm is more likely to fail to render some lights because all the lights in an area ended up in one layer instead of being well distributed. 

Also the changes there made the GLSL fail to compile on my machine. I had to add a float cast to the following line:
```
const vec3 tileScale = vec3( r_tileStep, 1.0 / numLayers );
```

I demonstrated the issue with a trail system:
```
fartfire
{
	beam
	{
		shader gfx/weapons/prifle/trail
		segments 300
		width 4 4
		alpha 1 0
		segmentTime 30
		fadeOutTime 2.5
		textureType stretch 1 .1

		dynamicLight 80 { .5 .5 .5 }

		jitter 1 10
	}
}
```

With the old algorithm it looks like this:
OLDunvanquished_2024-10-11_165403_000

With the new algorithm it looks like this:
NEWunvanquished_2024-10-11_165132_000

To reproduce,
1. Save the trail system definition above somewhere in the VFS as `scripts/something.trail`.
2. Execute these commands
```
/cg_thirdperson 1
/cg_thirdpersonangle 180
/testts fartfire
```
3. Run forward a while.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -307,7 +307,7 @@ void UpdateSurfaceDataLightMapping( uint32_t* materials, Material& material, dra
 
 		// bind u_LightTiles
 		if ( r_realtimeLightingRenderer.Get() == Util::ordinal( realtimeLightingRenderer_t::TILED ) ) {
-			gl_lightMappingShaderMaterial->SetUniform_LightTilesBindless(
+			gl_lightMappingShaderMaterial->SetUniform_LightTilesIntBindless(
 				GL_BindToTMU( BIND_LIGHTTILES, tr.lighttileRenderImage )
 			);
 		}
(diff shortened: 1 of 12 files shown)


VReaperV:
> After, the first quarter of lights are in layer 0, the 2nd quarter in layer 1, etc.

That's not true, light ID retrieval was just using an incorrect formula.

> Also the changes there made the GLSL fail to compile on my machine. I had to add a float cast to the following line:

That's weird, the only thing that changed there is `numLayers` int -> uint. Anyway, both issues are fixed by #1351.

slipher:
> > After, the first quarter of lights are in layer 0, the 2nd quarter in layer 1, etc.
> 
> That's not true, light ID retrieval was just using an incorrect formula.

I must have been supposing that based on `uint lightOffset = layer * globalLightsPerLayer;`

Issue #1289, opened 2024-09-09
Proposal for overbright cvars


slipher:
I suggest (1) to change the overbright cvar semantics to facilitate testing, and (2) to retain overbright clamping behavior on by default. I don't expect (1) to be controversial, but we may endlessly argue (2) as it is a matter of taste.

## 1. Change cvars so that all overbright settings can easily be tested.

The current overbright-related cvars have a wonky asymmetric design that can make it impossible to test a setting without editing the map. We should fix that. Something like `r_overbrightDefaultExponent` (0-3) and `r_overbrightDefaultClamp` (on/off) to control the default settings when the map doesn't say anything, and one more cvar `r_overbrightIgnoreMapSettings` to make those cvar values override the worldspawn values even if present. In the worldspawn entity `mapOverBrightBits` can stay as is, but `forceLegacyOverBrightClamping` should drop the "force" (since it is no longer a one-way switch) and rename to just `overbrightClamping` or something.

## 2. Set overbright clamping on by default.
If nothing else we should be true to the intentions of the authors of our own official maps, who were definitely targeting a renderer with clamping. For older maps, the `.ent` file may be used in restoration efforts if non-default settings look better. See the discussion below for more arguments on this topic.

## IRC discussion

```
<perturbed>	yeah well atcshd will be the only map which looks good with the new overbright capability :D
<Sweet>	it is unclear if the shall make the new overbright settings the default
<Sweet>	i argued against that
<Sweet>	reason: new maps can easily enable the new settings
<Sweet>	old maps are about 250, and cannot
<Sweet>	if * we* ahall
<Sweet>	jfc
<Sweet>	if *we* shall
<perturbed>	currently it's on by default
<Sweet>	that is not a final thing
<perturbed>	i think maybe 60-70% of maps look preferable with the overbright on
<Sweet>	focus on the open source maps
<perturbed>	i guess illwieckz thinks it is 90% lol
<Sweet>	he does
<Sweet>	but people are tweaking the lighting all the time
<Sweet>	even right now, probably
<Sweet>	we should be conservative with big lighting changes
<Sweet>	at least in the default settings
<Sweet>	for the hundreds of old maps without license and source: who has seen them all
<Sweet>	i certainly have not
<perturbed>	right
<perturbed>	good news is the .ent file can be used to tweak it, regardless of which way becomes the default
<Sweet>	the overbright thing is actually 6 different settings, afaik
<Sweet>	you can switch it on/off
[...]

PR #1350, opened 2024-10-12
Disable overbright by default + make fully configurable

slipher:
1. Change cvars so that all overbright settings can easily be tested. The current overbright-related cvars have a wonky asymmetric design that can make it impossible to test a setting without editing the map. Fix that. With this commit we can use r_overbrightDefaultExponent (0-3) and r_overbrightDefaultClamp (on/off) to control the default settings when the map doesn't say anything, and one more cvar r_overbrightIgnoreMapSettings to make those cvar values override the worldspawn values even if present. In the worldspawn entity mapOverBrightBits can stay as is, but forceLegacyOverBrightClamping drops the "force" (since it is no longer a one-way switch) and renames to just overbrightClamping.

2. Set overbright clamping on by default. If nothing else we should be true to the intentions of the authors of our own official maps, who were definitely targeting a renderer with clamping. For older maps, the .ent file may be used in restoration efforts if non-default settings look better.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -667,7 +667,7 @@ static void R_LoadLightmaps( lump_t *l, const char *bspName )
 					lightMapBuffer[( index * 4 ) + 2 ] = buf_p[( ( x + ( y * internalLightMapSize ) ) * 3 ) + 2 ];
 					lightMapBuffer[( index * 4 ) + 3 ] = 255;
 
-					if ( tr.forceLegacyOverBrightClamping )
+					if ( tr.legacyOverBrightClamping )
 					{
 						R_ColorShiftLightingBytes( &lightMapBuffer[( index * 4 ) + 0 ] );
 					}
@@ -1034,7 +1034,7 @@ static void ParseFace( dsurface_t *ds, drawVert_t *verts, bspSurface_t *surf, in
 		cv->verts[ i ].lightColor = Color::Adapt( verts[ i ].color );
 
 
-		if ( tr.forceLegacyOverBrightClamping )
+		if ( tr.legacyOverBrightClamping )
 		{
 			R_ColorShiftLightingBytes( cv->verts[ i ].lightColor.ToArray() );
 		}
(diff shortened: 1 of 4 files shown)


illwieckz:
There are physical evidences that not-clamped light was the true intention of the authors, this is true for both game engine authors and map authors.

## 1. Game engine authors

This quote from the official Quake3 manual:

> 6.3.1 RgbGen identityLighting
Colors will be (1.0,1.0,1.0) if running without overbright bits (NT, linux, windowed modes), or (0.5, 0.5, 0.5) if running with overbright. Overbright allows a greater color range at the expense of a loss of precision. Additive and blended stages will get this by default.
> -- https://icculus.org/gtkradiant/documentation/Q3AShader_Manual/ch05/pg5_1.htm

This quote is about `rgbGen identityLighting` which only has different values than `rgbGen identity` if and only if there is a need to adjust for _not_ using a brightening alternative technique to (clamped) overbright bits that was only implemented on Windows 9x.

Actually our GLSL overbright that is done in postprocess is similar to what an hardware overbright would do, and we also have to adjust the lighting on some surfaces to compensate.

## 2. Map authors

Those levelshots for A.T.C.S. Zone Alpha,

- [`map-atcszalpha.pk3`](https://users.unvanquished.net/~illwieckz/dump/betaserv.tk/map-atcszalpha.pk3):`levelshots/atcszalpha.jpg`, file date `2006-09-02 14:09:30`:

atcszalpha

- [`map-atcszalpha-b1.pk3`](https://users.unvanquished.net/~illwieckz/dump/betaserv.tk/map-atcszalpha.pk3):`levelshots/atcszalpha.jpg`, file date `2006-09-08 19:11:56`:

atcszalpha

Here is how looks the map in Dæmon engine with clamped overbright:

unvanquished_2024-10-12_033608_000

Here is how looks the map in Dæmon engine without clamped overbright:

unvanquished_2024-10-12_033519_000

I haven't found the exact `viewpos` but `-440 -365 532 -45 40` is close enough, and resolution is `720×570`.

We have minor difference in blending compared to what the mapper experienced, but the lighting is what the map author expects the map to look like in 2006. Tremulous was released in 2006 as well, so this is what Tremulous players playing fullscreen on Windows 9x experienced at release time.

slipher:
That's a cool demonstration that at least one person enjoyed a genuine overbright implementation. (And even had r_gamma 1, what a good boy!) But even in 2006, surely Windows 9x users were greatly outnumbered by users of other systems? If I'm not mistaken, Tremulous always had a substantial Linux playerbase. I suppose most mappers/players would have been working with the crippled version of overbright.

Forgot to link #1289 in the OP which contains a discussion of various people on whether we should make this the default.

PR #1351, opened 2024-10-12
Fix light ID computation

VReaperV:
Fix a bug introduced in #1333.

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -239,8 +239,8 @@ void computeDynamicLights( vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse, vec4
 	// NOT the amount of lights that can actually be put into each layer of the lighttile texture
 	uint globalLightsPerLayer = ( u_numLights + numLayers - 1 ) / numLayers;
 
-	uint lightCount = 0;
 	for( uint layer = 0; layer < numLayers; layer++ ) {
+		uint lightCount = 0;
 		idxs_t idxs = fetchIdxs( tileScale * vec3( tile, float( layer ) + 0.5 ), u_LightTiles );
 
 		uint lightOffset = layer * globalLightsPerLayer;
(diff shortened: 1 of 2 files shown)


src/engine/renderer/glsl_source/lighttile_fp.glsl
```diff
@@ -123,8 +123,8 @@ void main() {
 	only process 1 / 4 of different lights for each layer, extra lights going into the last layer. This can fail to add some lights
 	if 1 / 4 of all lights is more than the amount of lights that each layer can hold (16). To fix this, we'd need to either do this on CPU
 	or use compute shaders with atomics so we can have a variable amount of lights for each tile. */
-	for( uint i = u_lightLayer; i < u_numLights; i += 4 ) {
```
slipher, line 126: The prior version seems more readable to me.

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -239,8 +239,8 @@ void computeDynamicLights( vec3 P, vec3 normal, vec3 viewDir, vec4 diffuse, vec4
 	// NOT the amount of lights that can actually be put into each layer of the lighttile texture
 	uint globalLightsPerLayer = ( u_numLights + numLayers - 1 ) / numLayers;
 
-	uint lightCount = 0;
 	for( uint layer = 0; layer < numLayers; layer++ ) {
+		uint lightCount = 0;
 		idxs_t idxs = fetchIdxs( tileScale * vec3( tile, float( layer ) + 0.5 ), u_LightTiles );
 
 		uint lightOffset = layer * globalLightsPerLayer;
```
slipher, line 246: Why don't we make the loop condition below just `i < 16`? It will break on idx being 0 anyway.

VReaperV: I suppose that's better, yeah.

VReaperV: Done now.

src/engine/renderer/glsl_source/lighttile_fp.glsl
slipher: Also the proposed version is going to have an off by 1 bug if the number of lights is not divisible by 4.

VReaperV: Hmm, true.

VReaperV: Fixed now.

slipher:
LGTM

PR #1355, opened 2024-10-14
Implement cubemap probe grid, cubemap caching, and fix a few bugs

VReaperV:
Replace the hashtable with a new Grid<T> struct which acts as a wrapper around a region of memory accessable with 3 coordinates.

This fixes several issues:
- Cubemap probes are selected regardless of how far away the closest one is
- Both the nearest and second nearest cubemaps are always selected
- Performance is much better now while selecting the cubemap probes at all times, access time is low and pretty much constant, with a few extra operations for interpolation. The code supports sampling up to 8 grid points, but I didn't wanna bother with adding a lot of extra GLSL code to support that, so it still uses 2 cubemaps.

Implemented cubemap caching on disks. This saves the cubemaps themselves along with a file containing the grid, cubemap probe list and version + parts of BSP header + cubemap size and spacing in `homePath/reflectionCubemaps/mapName/`, and loads them from disk if available and the settings match.

Added an `r_autoBuildCubeMaps` cvar for it:
0: off
1: cache on disk
2: always rebuild on map load

Also fixed static reflections with material system and a few minor bugs.

Additionally fixed the issue that was resulting in the engine looking for images in the format of `name.ext.ext`, which was obviously failing.

`r_showCubeProbes` is now a 0-2 range cvar:
0: off
1: show cubemap grid
2: show only unique cubemaps


slipher:
I just did a bit of testing. My impressions:
- It could use more interpolation. The discontinuous cubemap changes are pretty noticeable.
- plat23 has way too much specularity on its textures. When you walk around in the outside area, the impression at some places is that the whole map is just randomly blinking. (I guess it's still choosing only one cubemap for all surfaces in the view.)
- The r_gamma setting gets baked into the cubemaps. I suppose they ought to be rendered without any screen-space effects.
- It's annoying that `r_autoBuildCubeMaps` builds them even if reflection mapping is disabled via cvars. I was confused for a while that nothing was happening even though cubemaps were built; turns out `r_specularMapping` was disabled.

This Parpax tank looks cool although it shows the discontinuousness problem.

[parpax-reflection.webm](https://github.com/user-attachments/assets/e74e963a-54bb-4e72-94dc-785118ce6be0)

VReaperV:
> It could use more interpolation. The discontinuous cubemap changes are pretty noticeable.

I agree, I've tried adjusting the formula for interpolation and the results seemed to be smoother, so I'll take another look at it. 

> plat23 has way too much specularity on its textures. When you walk around in the outside area, the impression at some places is that the whole map is just randomly blinking.

That'd be an asset issue then I suppose, though the blinking may be improved with better interpolation etc. 

> (I guess it's still choosing only one cubemap for all surfaces in the view.)

It is, yeah (well, 2 cubemaps to be specific). The problem with assigning a cubemap to each surface is that: 
- It would look wrong on large surfaces that have multiple cubemaps near them
- It would break batching
As I've said elsewhere, I believe that selecting one or more cubemaps per tile with the tiled renderer would fix this issue. After the rewrite in #1333 it would be easy to reserve some memory in each tile for a cubemap/cubemaps.

This would probably require using an array texture as having hundreds of sampler uniforms is not gonna work, and storing the cubemap ID in the tile, which would also be its layer in the texture. I have some code that adds support for array textures (only for 2D ones there, but it can be extended to cubemaps) in https://github.com/VReaperV/Daemon/tree/material-clusters, but it has some bugs still and would need to be ported over the newer commits, so that will have to wait until after 0.55 is released. The cubemap grid would then be uploaded as a 3D texture and accessed in `lighttile_fp`. The code there already knows the (minimum?) depth at each fragment, so getting the grid position from that would be trivial.

Then cubemap selection would be based on where a fragment is spatially and would work properly in most cases (other than perhaps where there's a large depth discontinuity within one tile).

This would, of course, not work without the tiled renderer, but even then I believe it can be brought to an acceptable quality.

> The r_gamma setting gets baked into the cubemaps. I suppose they ought to be rendered without any screen-space effects.

I think so too. The problem I ran into is that overbright settings are "baked" into the shaders (the shader struct that is) at map loading time, so it would probably require doing a `vid_restart` twice (some extra shader code under a define can be done easily though). `r_gamma` can be changed to 1.0 for the duration of cubemap rendering easily though.

> It's annoying that r_autoBuildCubeMaps builds them even if reflection mapping is disabled via cvars. I was confused for a while that nothing was happening even though cubemaps were built; turns out r_specularMapping was disabled.

I'll recheck the cvar combinations to make sure it's not built when it's unusable. I don't like how reflections require a combination of cvars to be enabled (and one disabled) to work, but I don't know a solution for that right now. I can add a warn or something in console if `r_reflectionMapping` is enabled but the other required cvars are not, but it would be easy to miss. I did add it in the description of the setting in https://github.com/Unvanquished/Unvanquished/pull/3110 at least (this should be done for some other graphics settings too...). Maybe adding a popup after the map is loaded if such a combination of cvars is detected, with options to auto-adjust settings, close, or never show again, would work.

VReaperV:
> * The r_gamma setting gets baked into the cubemaps. I suppose they ought to be rendered without any screen-space effects.

I made it set `r_gamma` to 1.0 when building cubemaps now. `cameraEffects` shader still has to be used for overbright.

VReaperV:
> I can add a warn or something in console if `r_reflectionMapping` is enabled but the other required cvars are not, but it would be easy to miss.

I've done this now. It will also not build cubemaps anymore if some cvar is missing. I don't like the fact the warn is quite high up in the console, but that can be improved in the future. There are other cvar combinations that require this, so some generalised solution would be better, but it's out of scope here.

VReaperV:
Also fixed a shader macro conflict with material system.

VReaperV:
> * It could use more interpolation. The discontinuous cubemap changes are pretty noticeable.

So this is apparently not an issue with interpolation, but an issue of cubemap placement. Right now they just placed in empty BSP leafs with a distance check for nearby cubemaps, but that doesn't mean that they properly encompass an area.

I've changed the default `r_cubeProbeSpacing` to 256 now, which helps somewhat, but a proper fix would require a bunch more work and wouldn't fit into this pr.

VReaperV:
Notably, with the change to `r_cubeProbeSpacing` there's no longer any issue with the tank on parpax.

illwieckz (approved):
```
LGTM LGTM LGTM
LGTM LGTM LGTM
LGTM LGTM LGTM
LGTM LGTM LGTM
```

illwieckz:
> Additionally fixed the issue that was resulting in the engine looking for images in the format of name.ext.ext, which was obviously failing.

@VReaperV This is wanted. A q3 shader can contain image path without ext. So for example `image.blend.jpg` can be written as `image.blend`, meaning the ext split will provide `image` and `blend` but we need to look for `image.blend.ext` so `image.ext1.ext2`, see: 15815d42ec5c006b7bb254f12fc827fb66fafec1.

VReaperV:
> > Additionally fixed the issue that was resulting in the engine looking for images in the format of name.ext.ext, which was obviously failing.
> 
> @VReaperV This is wanted. A q3 shader can contain image path without ext. So for example `image.blend.jpg` can be written as `image.blend`, meaning the ext split will provide `image` as basename and `blend` as ext but we need to look for `image.blend.ext` so `image.ext1.ext2`, see: [15815d4](https://github.com/DaemonEngine/Daemon/commit/15815d42ec5c006b7bb254f12fc827fb66fafec1).

I don't think it's the same issue. In this case engine was looking for `image.ktx.ktx` and failing because of that, that makes 0 sense.

VReaperV:
> > Additionally fixed the issue that was resulting in the engine looking for images in the format of name.ext.ext, which was obviously failing.
> 
> @VReaperV This is wanted. A q3 shader can contain image path without ext. So for example `image.blend.jpg` can be written as `image.blend`, meaning the ext split will provide `image` as basename and `blend` as ext but we need to look for `image.blend.ext` so `image.ext1.ext2`, see: [15815d4](https://github.com/DaemonEngine/Daemon/commit/15815d42ec5c006b7bb254f12fc827fb66fafec1).

In fact, I don't even think I changed the code you linked, since I changed `R_FindCubeImage()` rather than `R_LoadImage()`.

PR #1356, opened 2024-10-14
Tweak material system cvars

slipher:
- Disable GPU occlusion culling by default. @VReaperV and I agreed in an IRC chat a while back that it is not ready yet.
- Make it so that material system can be toggled with a single cvar, and updates on map change like all other latched renderer cvars. Right now it requires setting 2 cvars and only updates on a full renderer reload with `vid_restart`.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -29,6 +29,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 /*
 =================================================================================
 THIS ENTIRE FILE IS BACK END!
+^ Not true, EnableAvailableFeatures() right below this isn't.
 
 This file deals with applying shaders to surface data in the tess struct.
 =================================================================================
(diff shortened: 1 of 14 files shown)


src/engine/renderer/tr_shade.cpp
```diff
@@ -143,6 +143,7 @@ static void EnableAvailableFeatures()
 	}
 
 	glConfig2.usingMaterialSystem = r_materialSystem.Get() && glConfig2.materialSystemAvailable;
+	glConfig2.usingBindlessTextures = glConfig2.usingMaterialSystem || r_preferBindlessTextures.Get();
```
VReaperV, line 146: This seems wrong, if `r_preferBindlessTextures` is `true` and the extension is unavailable or disabled, `glConfig2.usingBindlessTextures` would still be true and would lead to errors/crashes.

slipher: Good catch. Fixed

VReaperV (approved):
LGTM

PR #1357, opened 2024-10-15
Don't do dynamic light computations if there are no dynamic lights

VReaperV:
Avoid the overhead of fetching all or most of the lighttile texture and running shaders that do nothing when there are no dynamic lights.


illwieckz (approved):
LGTM.

PR #1359, opened 2024-10-15
Don't process binds if KEYCATCH_UI was just disabled

slipher:
Decide whether to process binds based on the key catcher state before the key down event(s) are processed, not after. This is important to avoid double-processing a key if the UI turns off its key catcher in response to a key press.

Fixes https://github.com/Unvanquished/Unvanquished/issues/3113.


DolceTriade (approved):
wfm

PR #1361, opened 2024-10-16
Fix R_CreateImage() crash with r_logFile

### LINKED ISSUE #1020: r_logFile is broken
Doing /r_logFile 1, then /devmap plat23 produces a black screen upon map load for a few seconds, then goes back to the main menu with the following error:
image

VReaperV:
Fixes a crash resulting from `va()` overflow. Fixes #1020.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -3101,8 +3101,9 @@ qhandle_t RE_GenerateTexture( const byte *pic, int width, int height )
 		return 0;
 	}
 
-	const char *name = va( "rocket%d", numTextures++ );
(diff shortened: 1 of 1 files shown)


slipher:
Where are the contents of the "log file" printed? I tried that once and couldn't figure it out.

VReaperV:
They're added into a GL log apparently, I can see that in a frame debugger:
image

Supposedly it should print to console with r_glDebugProfile, but that seems not to work.

VReaperV:
> Can you actually see the message there?
> 
> LGTM

Yeah, the part after "buf" (--- R_BindVBO...) is the message.

PR #1362, opened 2024-10-16
Fix R_ListImages_f() image size

VReaperV:
Account for mipmaps to report more accurate memory data.


illwieckz:
Yay!

illwieckz (approved):
If it works for you LGTM.

VReaperV:
It does, when comparing to the image sizes reported by a frame debugger, this branch is (more) correct.

PR #1363, opened 2024-10-16
renderer: reverse r_fastSky as r_drawSky

illwieckz:
We usually do end-users options the positive way, and this would be more consistent in option window where people ticks the boxes for enabling things.


DolceTriade:
Is this really required for 0.55.0?

slipher:
LGTM

illwieckz:
> Is this really required for 0.55.0?

This makes the graphics option more polished:

https://github.com/Unvanquished/Unvanquished/pull/3117

The presets were already modifying that option so not having a toggle in UI is weird, and this makes the UI better.

PR #1364, opened 2024-10-16
Make overbright settings fully configurable (default no clamping version)

slipher:
Same as #1350 but overbright clamping is NOT disabled by default. So this version would go ahead and change how all Unvanquished and most Tremulous maps are displayed


illwieckz (approved):
I notice that `r_mapOverBrightBits` was renamed and that name dated back to quake3, but we also renamed `r_fastsky` and `r_dynamiclight` so I don't see any reason to care about that if we have not cared about the others (and I'm the culprit for the others)!

LGTM.

PR #1366, opened 2024-10-17
common: do not cap timedemo at 999fps

illwieckz:
This is something I totally forgot to submit, it was one of the purpose of me tracking many divisions by zero and other bugs related so small `minMsec` 20 months ago:

- https://github.com/DaemonEngine/Daemon/pull/812

The purpose is to not cap `timedemo` benchmarks at ~999fps (in fact more around 970fps because of time precision).

It goes against the idea of providing a benchmark feature if all hardware and driver would only report ~970fps if they can reach at least reach that.

### DIFF
--- a/src/engine/qcommon/common.cpp
+++ b/src/engine/qcommon/common.cpp
@@ -870,11 +870,10 @@ void Com_Frame()
 	}
 	else
 	{
-		// Bad things happen if minMsec is 0.
 		// It looks like demo played with cvar_demo_timedemo enabled
 		// are not affected by the rendering bugs related to having
 		// minMsec being lower than 3.
-		minMsec = 1;
+		minMsec = 0;
 	}
 
 	Com_EventLoop();


illwieckz:
This was tested with multiple demos, including portals, going out of bounds, doing SMP or not, using ultra and lowest presets, actually doing more than 1000fps, and running a build using `feenableexcept(FE_DIVBYZERO);` and NOT catching any division by zero.

DolceTriade:
Again, this is not a regression for 0.55. This is an engine change that *does not* need to go to users. It can be postponed to a later release.

illwieckz:
This is a bug. It's like having implemented normal mapping 20 months ago, having announced the release will feature normal mapping, and discovering while testing the RC that the `r_normalMapping` cvar only enables half of the normal mapping code and only does it on models and not on world.

The ability to benchmark the game above 999fps has been implemented 20 months ago, this is also because of this effort done 20 months ago to improve the benchmark ability I have said “_prepare your benchmark_” in the blog post. I also already told Michael Larrabel in Phoronix forums that this release will be interesting to benchmark. I plan to write him an email with details about benchmarking this release, this is why I did a couple of benchmarking tests over the RC2:

- to check that a demo recorded with GL3+ can run on a GL2 card,
- to check that it is true we are not benchmarking `Sys::Milliseconds()` above 999fps anymore.

First test was valid, second test was invalid.

This is a bugfix that should have been merged no later than December of 2023 when we were releasing 0.54.1.

What happens is that the `demo.timedemo` cvar only does half the job: it starts and stops the timer but I forgot to make it disable the sleep function too.

It was forgotten because at the time the commits written to fix bugs that were previously worked around by the sleep function were split in various PR, and it was expected to merge this last bit once all of them were merged. I simply forgot about it once all the child PRs were merged.

This change should have been merged 20 months ago and the rest of the code merged 20 months ago was tested with this code. This bugfix already missed a release.

PR #1367, opened 2024-10-17
Fix latch cvars not being archived

slipher:
Fixes https://github.com/Unvanquished/Unvanquished/issues/3120.

I suspect this is only now showing up due to removing useless cvar sets on cgame load with
https://github.com/Unvanquished/Unvanquished/pull/3099.


VReaperV (approved):
LGTM

PR #1370, opened 2024-10-18
Set default of r_autobuildcubemaps to always build

slipher:
The default of never build makes it unusable without setting 2 cvars. Always seems like a good conservative default so that it won't consume a bunch of disk space.


VReaperV (approved):
LGTM

PR #1372, opened 2024-10-19
Work around crash on Mac with null tess.verts

slipher:


slipher:
Hmm this really kills the perf. 45 fps in plat23 alien base, compared to 300 for 0.54.1.

PR #1373, opened 2024-10-19
Fix crash with glMapBufferRange-based vertex buffer

### LINKED ISSUE #1371: Mapping the dynamic VBO fails on Mac
On Mac, glMapBufferRange can return null. Will add more info later.

slipher:
Fixes #1371.


illwieckz:
It works. Thank you very much!

illwieckz:
Wait, why did it fail to detect I merged it from command line?

PR #1376, opened 2024-10-20
Implement GPU lane usage profiling

VReaperV:
Add `r_profilerRenderSubGroups` and `r_profilerRenderSubGroupsMode`.

If `r_profilerRenderSubGroups` is enabled and the required subgroup extensions are suported, lightMapping and generic shaders will change their output colour to `dim red->bright red->dim green->bright green` based on how many active lanes in either the VS or FS there are.

Modes:
0: VS opaque
1: VS transparent
2: VS all
3: FS opaque
4: FS transparent
5: FS all

Also improved shader post-processing to leave global uniforms in-place to make this work, and made it consider comments on the same line.

More red = ~more scary~ worse (less active lanes), more green = better (more active lanes)
This is an example of using mode 0 on plat23:
unvanquished_2024-10-20_225919_000
Mode 2:
unvanquished_2024-10-20_225939_000
The surfaces resulting from rendering commands with few triangles are the more red ones.
Mode 3:
unvanquished_2024-10-20_225947_000
Mode 5:
unvanquished_2024-10-20_225956_000
The triangle edges, for example, have more inactive lanes.

Dense vegetation on map `moonbase`, VS lanes are mostly good:
unvanquished_2024-10-20_230210_000
But there's a lot of wasted FS lanes:
unvanquished_2024-10-20_230220_000

I've originally thought of doing this because of something that caught my attention in one of the error logs that @illwieckz has sent me for some shader change or another. Lanes can become inactive for many reasons: not enough work for the SIMD unit to do, triangle edges, too much register pressure etc... The log illwieckz sent me had a line that said something along the lines of only 1 invocation being active at maximum, so it'd be interesting to see the results of using this on Mesa.

Requires subgroup extensions.

### DIFF
--- a/src/engine/renderer/glsl_source/shaderProfiler_vp.glsl
+++ b/src/engine/renderer/glsl_source/shaderProfiler_vp.glsl
@@ -1,0 +1,48 @@
+/*
+===========================================================================
+
+Daemon BSD Source Code
+Copyright (c) 2024 Daemon Developers
+All rights reserved.
+
+This file is part of the Daemon BSD Source Code (Daemon Source Code).
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+	* Redistributions of source code must retain the above copyright
+	  notice, this list of conditions and the following disclaimer.
+	* Redistributions in binary form must reproduce the above copyright
+	  notice, this list of conditions and the following disclaimer in the
+	  documentation and/or other materials provided with the distribution.
+	* Neither the name of the Daemon developers nor the
+	  names of its contributors may be used to endorse or promote products
+	  derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL DAEMON DEVELOPERS BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+===========================================================================
+*/
+
+/* shaderProfiler_vp.glsl */
+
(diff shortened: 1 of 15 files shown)


VReaperV:
> What happens in a multi-stage shader? Would you only be able to see the last stage?

Yeah, only the last stage. I can add a cvar to make it skip all stages except one though (or a specific material ID, since one material can be e. g. first stage on one surface, and second stage on another surface).

src/engine/renderer/glsl_source/shaderProfiler_vp.glsl
```diff
@@ -0,0 +1,49 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+===========================================================================
+*/
+
+/* shaderProfiler_vp.glsl */
+
+#if defined(r_profilerRenderSubGroups) && !defined(GENERIC_2D) && defined(HAVE_KHR_shader_subgroup_basic) && defined(HAVE_KHR_shader_subgroup_arithmetic)
+	#define SHADER_PROFILER_OUT \
```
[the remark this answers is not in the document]
VReaperV: True, I did it for `shaderProfiler_fp` earlier, but forgot to make that change to the `_vp` counterpart. Fixed now.

VReaperV:
> What happens in a multi-stage shader? Would you only be able to see the last stage?

I've now added `r_profilerRenderSubGroupsStage`. When set to 0 or more, it will only render the specific stage/material, otherwise if set to -1 it will render all stages on top of each other (like it did before I added this cvar).

src/engine/renderer/tr_init.cpp
```diff
@@ -256,12 +256,15 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
 
-	Cvar::Cvar<bool> r_profilerRenderSubGroups( "r_profilerRenderSubGroups", "Enable subgroup profiling in rendering shaders", Cvar::NONE, false );
+	Cvar::Cvar<bool> r_profilerRenderSubGroups( "r_profilerRenderSubGroups", "Enable subgroup profiling in rendering shaders", Cvar::CHEAT, false );
 	Cvar::Range<Cvar::Cvar<int>> r_profilerRenderSubGroupsMode( "r_profilerRenderSubGroupsMode", "Red: more wasted lanes, green: less wasted lanes; "
 		"0: VS opaque, 1: VS transparent, 2: VS all, 3: FS opaque, 4: FS transparent, 5: FS all", Cvar::NONE,
 		Util::ordinal( shaderProfilerRenderSubGroupsMode::VS_OPAQUE ),
 		Util::ordinal( shaderProfilerRenderSubGroupsMode::VS_OPAQUE ),
 		Util::ordinal( shaderProfilerRenderSubGroupsMode::FS_ALL ) );
+	Cvar::Cvar<int> r_profilerRenderSubGroupsStage( "r_profilerRenderSubGroupsStage", "Select a specific"
+		"stage/material ID (if material system is enabled) for subgroup profiling (-1 to always profile the last stage/material); "
```
slipher, line 266: Should this say 'all materials' instead of 'last material'?

VReaperV: I've changed this to say "-1 to profile all stages/materials, rendered in theuir usual order" now.

slipher: That doesn't match the non-material behavior now which only does the last stage with -1.

VReaperV: Right, I changed stage behaviour to match now.

VReaperV:
Added a missing check for depth pre-pass surfaces in tr_shade.cpp.

src/engine/renderer/tr_shade.cpp
```diff
@@ -2743,14 +2767,21 @@ void Tess_StageIteratorColor()
 	}
 
 	// call shader function
-	uint stage = 0;
+	int stage = 0;
```
slipher, line 2770: it's always 0

VReaperV: Fixed.

slipher:
LGTM

Issue #850, opened 2023-05-01
Investigating the non-working liquid shader


illwieckz:
Do, in `tr_shade.cpp` the code is doing that in `Render_liquid()`:

```c++
GL_BindToTMU( 1, tr.portalRenderImage );
```

I happens that `portalRenderImage` is actually not initialized while it should.

This image is expected to be set in `tr_backend.cpp` in `RB_RenderView()`, but it is not set because it happens that `backEnd.viewParms.portalLevel` is `0`.

It is expected that `backEnd.viewParms.portalLevel` is set in `R_MirrorViewBySurface()` in `tr_main.cpp`, but `R_MirrorViewBySurface` seems to not be called.

illwieckz:
On thing is that `R_MirrorViewBySurface` is called only if `shader.sort` is `SS_PORTAL`, but the water material both set `mirror` then `sort postProcess`, meaning `shader.sort` is first set to `SS_PORTAL` then overwritten to `SS_POST_PROCESS`. If I remove `postProcess` for now `R_MirrorViewBySurface` is called .

But then `R_MirrorViewBySurface` returns `false`.

illwieckz:
OK, the map needed a `misc_portal_surface` entity to be close to the water surface, now I get the portal texture be properly initialized, but there are many other bugs?

illwieckz:
So, by looking at the way it is rendered in game, it _looks like_ the surface is rendered as a mirror, not as a liquid shader…

illwieckz:
I actually got “good enough” proof that the mirror is rendered above the surface, instead of feeding the liquid shader.

VReaperV:
Is it expected to remder as mirror with some sort of distortions by water?

VReaperV:
For portal image we can just generate a portal entity near liquid surfaces. Not sure what stage it should be though.

Also, related issue: #645.

illwieckz:
> Is it expected to render as a mirror with some sort of distortions by water?

I believe the rendered mirror is reused as a texture in the GLSL shader itself to apply some transformation.

VReaperV:
It might be better to just render liquids in 2 stages: a portal stage, and a liquid stage blended over it with additive blending (portal image would then be removed from the shader).

Liquid shader for some reason also uses portal colour as material colour in `computeDeluxeLight()`, which seems completely wrong. Other than that, we're already essentially doing additive blending, just in the shader.

[...]

PR #1377, opened 2024-10-20
Fix liquid shader compile error

VReaperV:
Reported by someone in-game.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -1191,6 +1191,8 @@ void BindShaderLiquid( Material* material ) {
 	// Select shader permutation.
 	gl_liquidShaderMaterial->SetHeightMapInNormalMap( material->hasHeightMapInNormalMap );
 	gl_liquidShaderMaterial->SetReliefMapping( material->enableReliefMapping );
+	gl_liquidShaderMaterial->SetGridDeluxeMapping( material->enableGridDeluxeMapping );
+	gl_liquidShaderMaterial->SetGridLighting( material->enableGridLighting );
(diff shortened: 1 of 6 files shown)


slipher:
Why do we even have this file? It doesn't work at all so shouldn't we just nuke it?

VReaperV:
There's an issue for it: #850. It is used however, e. g. on map https://users.unvanquished.net/~sweet/pkg/map-citadel_0%2B0.dpk.

VReaperV:
> It doesn't work at all so shouldn't we just nuke it?

It does sort of work, but there are some issues with it that need to be fixed.

slipher:
Why add a GLSL compile macro that's always enabled? The point of those is to build multiple variants of the shader.

Also it's possible that grid lighting or deluxe mapping are globally disabled.

VReaperV:
> Why add a GLSL compile macro that's always enabled? The point of those is to build multiple variants of the shader.
Also it's possible that grid lighting or deluxe mapping are globally disabled.

Well, it's required for `ReadLightGrid()` and `ComputeDeluxeLight()` to be available. Perhaps liquid shader should just be disabled if those aren't available.

slipher:
> Well, it's required for `ReadLightGrid()` and `ComputeDeluxeLight()` to be available.

Could just write `#define USE_GRID_LIGHTING` or whatever in the liquid shader.

VReaperV:
I've changed this to actually check if lightgrid/deluxegrid are available. Otherwise it would probably break if the map lacks one.

slipher:
LGTM

PR #1381, opened 2024-10-21
Small fixes

necessarily-equal:
This PR backports small fixes from me and freem. The "fix missing headers" commit is especially useful as I needed it to be able to compile Daemon.

### DIFF
--- a/src/common/cm/cm_load.cpp
+++ b/src/common/cm/cm_load.cpp
@@ -771,7 +771,7 @@ void CM_LoadMap(Str::StringRef name)
 	std::error_code err;
 	std::string mapData = FS::PakPath::ReadFile(mapFile, err);
 	if (err) {
-		Sys::Drop("Could not load %s", mapFile.c_str());
+		Sys::Drop("Could not load %s: %s (code: %d)", mapFile.c_str(), err.message(), err.value() );
 	}
 
 	std::string externalEntities = FS::PakPath::ReadFile( "maps/" + name + ".ent", err );
--- a/src/common/cm/cm_load.cpp
+++ b/src/common/cm/cm_load.cpp
@@ -50,9 +50,9 @@ clipMap_t cm;
 int       c_pointcontents;
 int       c_traces, c_brush_traces, c_patch_traces, c_trisoup_traces;
 
-cmodel_t  box_model;
-cplane_t  *box_planes;
-cbrush_t  *box_brush;
+static cmodel_t  box_model;
+static cplane_t  *box_planes;
+static cbrush_t  *box_brush;
 
 void      CM_InitBoxHull();
 void      CM_FloodAreaConnections();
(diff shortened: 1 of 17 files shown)


src/engine/sys/con_tty.cpp
```diff
@@ -85,9 +85,9 @@ void WriteToStdout(const char* text) {
 
 	if (ret < 0) {
 		// Do not log to stdout
-		Log::Dispatch( Str::Format("Error writing to the terminal: %d", errno),
+		Log::Dispatch( Str::Format("Error writing to the terminal: %s", strerror(errno)),
```
DolceTriade, line 88: let's print the code and the string?

DolceTriade (commented):
otherwise lgtm

src/common/cm/cm_load.cpp
```diff
@@ -771,7 +771,7 @@ void CM_LoadMap(Str::StringRef name)
 	std::error_code err;
 	std::string mapData = FS::PakPath::ReadFile(mapFile, err);
 	if (err) {
-		Sys::Drop("Could not load %s", mapFile.c_str());
+		Sys::Drop("Could not load %s: %s (code:%02x)", mapFile.c_str(), err.message(), err.value() );
```
slipher, line 774: The use of an unprefixed hexadecimal number is misleading. If it had only digits 0-9, you couldn't tell it's supposed to be hexadecimal.

src/common/Assert.h
```diff
@@ -31,6 +31,12 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef COMMON_ASSERT_H_
 #define COMMON_ASSERT_H_
 
+#include <utility>
```
slipher, line 34: Our guidance on the common headers for now is they are not actually expected to be self-contained. What you are supposed to to is always include `Common.h` first in every .cpp file.

src/engine/sys/con_tty.cpp
necessarily-equal: The standard in Daemon code seems to be to print only `strerror(errno)`, and not the number. I guess even worst case if the message is translated it's often clear which one it is. If that's a problem though, then we should rework more than just this line

src/common/cm/cm_load.cpp
necessarily-equal: Good point, fixed

src/common/Assert.h
necessarily-equal: I've tried to reproduce the bug to try to find an alternative solution. But somehow, the bug doesn't trigger anymore. Do you have suggestions on how to reproduce again?

In the meantime, I removed the commit so be able to merge this faster. Will merge tomorrow most likely

slipher:
LGTM

PR #1389, opened 2024-10-25
sv_main: let LAN server created with client respond to queries

illwieckz:
By default, no one can list a LAN server that someone else created with a client, change that.


DolceTriade (approved):

slipher:
My proposal https://github.com/DaemonEngine/Daemon/pull/1395  keeps the restriction on non-LAN addresses and gets rid of the useless option number 3 for `server.private` (which is split into two cvars).

illwieckz:
Obsoleted by https://github.com/DaemonEngine/Daemon/pull/1395

PR #1394, opened 2024-10-26
tr_image: do not use R_FindImageLoader with cube maps, test multifile cubemaps first (or preview may be loaded as cubemap instead)

### LINKED ISSUE #1393: Broken wasteland skybox
It worked before™ but it doesn't anymore:

```
Warn: cube map face 'env/wasteland/wasteland_rt' has DXTn compression, cube map unusable 
Warn: could not find cubemap 'env/wasteland/wasteland' for outer skybox in shader 'textures/perseus_env/sky' 
```

I first reproduced it on atcshd that ships wasteland, but one can reproduce it with stock assets as perseus ships a copy of wasteland.

illwieckz:
Deep rewrite of image loading code.

- do not walk .shader files in tr_image.cpp
- do not use R_FindImageLoader with cube maps
- only use home read for KTX with IF_HOMEPATH
- load home KTX with IF_HOMEPATH before doing any lookup
- do not do any other home loading
- test multifile cubemaps before singlefile ones (or preview may be loaded as cubemap instead)
- fix a crash when singlefile cubemap is freed because it was loaded but not created
- look for cubemap file with explicit extension
- add many debug logs and factorize them

Fixes #1393:

- https://github.com/DaemonEngine/Daemon/issues/1393

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -1616,27 +1616,29 @@ image_t *R_Create3DImage( const char *name, const byte *pic, int width, int heig
 	return image;
 }
 
-struct imageExtToLoaderMap_t
+using imageLoader_t = void( * )( const char *, byte **, int *, int *, int *, int *, int *, byte );
+
+struct imageExtLoader_t
 {
 	const char *ext;
-	void ( *ImageLoader )( const char *, unsigned char **, int *, int *, int *, int *, int *, byte );
+	const char *name;
+	imageLoader_t imageLoader;
+	bool cubemap;
 };
 
-// Note that the ordering indicates the order of preference used
-// when there are multiple images of different formats available
-static const imageExtToLoaderMap_t imageLoaders[] =
+/* The ordering indicates the order of preference used when
+there are multiple images of different formats available. */
+static const imageExtLoader_t imageLoaders[] =
 {
-	{ "webp", LoadWEBP },
-	{ "png",  LoadPNG  },
-	{ "tga",  LoadTGA  },
-	{ "jpg",  LoadJPG  },
-	{ "jpeg", LoadJPG  },
-	{ "dds",  LoadDDS  },
-	{ "crn",  LoadCRN  },
-	{ "ktx",  LoadKTX  },
+	{ "webp", "WebP", LoadWEBP, false },
+	{ "png",  "PNG",  LoadPNG,  false },
+	{ "tga",  "TGA",  LoadTGA,  false },
+	{ "jpg",  "JPEG", LoadJPG,  false },
+	{ "jpeg", "JPEG", LoadJPG,  false },
+	{ "dds",  "DDS",  LoadDDS,  false },
+	{ "crn",  "CRN",  LoadCRN,  true  },
+	{ "ktx",  "KTX",  LoadKTX,  true  },
 };
-
-static int                   numImageLoaders = ARRAY_LEN( imageLoaders );
 
 /*
 =================
@@ -1646,44 +1648,40 @@ Finds and returns an image loader for a given basename,
 tells the extra prefix that may be required to load the image.
 =================
 */
-static int R_FindImageLoader( const char *baseName, const char **prefix ) {
+static const imageExtLoader_t* R_FindImageLoader( const char *baseName, const char **prefix ) {
 	const FS::PakInfo* bestPak = nullptr;
-	int i;
+	const imageExtLoader_t* bestLoader = nullptr;
 
-	int bestLoader = -1;
 	*prefix = "";
+
 	// try and find a suitable match using all the image formats supported
 	// prioritize with the pak priority
-	for ( i = 0; i < numImageLoaders; i++ )
+	for ( const imageExtLoader_t &loader : imageLoaders )
 	{
-		std::string altName = Str::Format( "%s.%s", baseName, imageLoaders[i].ext );
+		std::string altName = Str::Format( "%s.%s", baseName, loader.ext );
 		const FS::PakInfo* pak = FS::PakPath::LocateFile( altName );
 
-		// We found a file and its pak is better than the best pak we have
-		// this relies on how the filesystem works internally and should be moved
-		// to a more explicit interface once there is one. (FIXME)
+		/* We found a file and its pak is better than the best pak we have
+		this relies on how the filesystem works internally.
+		FIXME: Move to a more explicit interface once there is one. */
 		if ( pak != nullptr && ( bestPak == nullptr || pak < bestPak ) )
 		{
+			/* We found a file and its pak is better than the best pak we have
+			this relies on how the filesystem works internally. */
 			bestPak = pak;
-			bestLoader = i;
-		}
-
-		if( pak == nullptr ) {
-			if ( FS::HomePath::FileExists( altName ) ) {
-				bestLoader = i;
-			}
+			bestLoader = &loader;
 		}
 
 		// DarkPlaces or Doom3 packages can ship alternative texture path in the form of
 		//   dds/<path without ext>.dds
-		if ( bestPak == nullptr && !Q_stricmp( "dds", imageLoaders[i].ext ) )
+		if ( bestPak == nullptr && !Q_stricmp( "dds", loader.ext ) )
 		{
 			std::string prefixedName = Str::Format( "dds/%s.dds", baseName );
 			bestPak = FS::PakPath::LocateFile( prefixedName );
 			if ( bestPak != nullptr ) {
 				*prefix = "dds/";
 				bestPak = pak;
-				bestLoader = i;
+				bestLoader = &loader;
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


src/engine/renderer/tr_image.cpp
```diff
@@ -2038,6 +2038,8 @@ struct face_t
 	int height;
 };
 
+
+
```
VReaperV, line 2042: Extraneous newlines.

src/engine/renderer/tr_image.cpp
```diff
@@ -2195,6 +2176,47 @@ image_t *R_FindCubeImage( const char *imageName, imageParams_t &imageParams )
...
+		// We found a file and its pak is better than the best pak we have
+		// this relies on how the filesystem works internally and should be moved
+		// to a more explicit interface once there is one. (FIXME)
+		if ( pak != nullptr && ( bestPak == nullptr || pak < bestPak ) )
+		{
+			bestPak = pak;
+			foundLoader = true;
+		}
+
+		if ( pak == nullptr ) {
```
slipher, line 2198: We don't want this homepath thing (yes it should be deleted from the other one too)

src/engine/renderer/tr_image.cpp
```diff
@@ -2195,6 +2176,47 @@ image_t *R_FindCubeImage( const char *imageName, imageParams_t &imageParams )
 		R_FreeCubePics( pic, i );
 	}
 
+	char baseName[ MAX_QPATH ];
+	COM_StripExtension3( buffer, baseName, sizeof( baseName ) );
+
+	const FS::PakInfo* bestPak = nullptr;
```
slipher, line 2182: Could use at least a  TODO comment about respecting explicit extensions.

illwieckz:
I finally reworded the thing in a better way, including the non-cubemap code.

src/engine/renderer/tr_image.cpp
illwieckz: Done.

src/engine/renderer/tr_image.cpp
illwieckz: @VReaperV Is it used to load cached reflection cubemaps?

src/engine/renderer/tr_image.cpp
illwieckz: Done.

src/engine/renderer/tr_image.cpp
VReaperV: The one in `R_FindImageLoader()` is.

illwieckz: Well, something is wrong, `R_FindImageLoader()` is not expected to be used to load a cached cubemap.

For now I'll keep the home loader in both, then later we can decide what is to be kept when things are not weird.

illwieckz: > 🔹️ xreaperx: Well, I thought you were asking about the homepath usage in the loader?
> 🔹️ xreaperx: I added it so it would look for .ktx under homepath
> 🔹️ illwieckz: yes
> 🔹️ illwieckz: but you said only R_FindImageLoader requires to read home path
> 🔹️ xreaperx: Well yeah otherwise it doesn't find the files
> 🔹️ illwieckz: but this is a cubemap
> 🔹️ illwieckz: R_FindImageLoader is for non-cubemap files
> 🔹️ xreaperx: Yeah there's some weirdness with that

illwieckz: > 🔹️ xreaperx: @illwieckz `R_FindCubeImage()` use `R_FindImageLoader()`, not `R_FindCubeImage()`
> 🔹️ xreaperx: So if that's not what it should do, then the error is in that

I say: we merge this that fixes the skybox, then later we fix the reflection cubemap code.

VReaperV: Sounds reasonable to me. I don't know if there's supposed to be some other way to load images from homepath, but other functionality that I looked at (log, glsl cache) used `FS::HomePath`.

slipher: This is not the code that's used for reflection maps. The reflection map search sets an `IF_HOMEPATH` flag instructing to search the homepath instead of the VFS. It's bad to search the homepath for every random image.

VReaperV: Yes and no. `loader.ImageLoader()` is the only call that uses `IF_HOMEPATH`, but `R_FindImageLoader()`, which preceeds it, doesn't use that, and is used to know if the file exists in the first place. I suppose a proper solution might be to pass `imageParams.bits` to it and then use `IF_HOMEPATH`.

illwieckz: The purpose of this PR is to fix skybox loading, other things like “attempting to remove some code and see if that breaks reflections“ later.

illwieckz:
I added some `TODO` comments saying the removal of homepath reading have to be investigated later.

src/engine/renderer/tr_image.cpp
```diff
@@ -1772,21 +1774,21 @@ static void R_LoadImage( const char **buffer, byte **pic, int *width, int *heigh
 	// and it may be possible the file has a dot in its name that was mistakenly
 	// taken as an extension
 	const char *prefix;
-	int bestLoader = R_FindImageLoader( filename, &prefix );
+	const imageExtLoader_t* loader = R_FindImageLoader( filename, &prefix );
 
-	if ( *ext && bestLoader == -1 )
+	if ( *ext && loader )
```
VReaperV, line 1779: This should be `!loader`.

illwieckz:
I noticed some functions were copying with `Q_strncpyz` strings received as argument before using them. I guess it's some security check in case some user-provided data is intentionally wrong. I don't know if such copy is needed but I renamed the symbols to make them more obvious.

src/engine/renderer/tr_image.cpp
```diff
@@ -2075,112 +2046,162 @@ struct face_t
...
+				{
+					/* Ignore silently, skip this format. This may silence incomplete multifile cubemap
+					but this is an hardly decidable problem since some formats can share some suffixes */
+					break;
+				}
+
+				if ( IsImageCompressed( imageParams.bits ) )
+				{
+					Log::Warn("Cube map face '%s' has DXTn compression, cube map unusable", fileName );
+					R_FreeCubePics( pic, i );
```
slipher, line 2174: Should be `i + 1`

src/engine/renderer/tr_image.cpp
```diff
@@ -2075,112 +2046,162 @@ struct face_t
...
-				break;
-			}
+				Log::Debug( "Looking for %s cube map face '%s'", format.name, fileName );
 
-			if ( numLayers > 0 )
-			{
-				Log::Warn("cubemap face '%s' is a multilayer image with %d layer(s), cube map unusable", filename, numLayers);
-				break;
-			}
+				R_LoadImage( &fileName[ 0 ], &pic[ i ], &width, &height, &numLayers, &numMips, &imageParams.bits );
```
slipher, line 2162: Just `fileName`

src/engine/renderer/tr_image.cpp
```diff
@@ -2075,112 +2046,162 @@ struct face_t
...
 
-			if ( width > greatestEdge )
-			{
-				greatestEdge = width;
+				if ( !pic[ i ] )
+				{
+					/* Ignore silently, skip this format. This may silence incomplete multifile cubemap
+					but this is an hardly decidable problem since some formats can share some suffixes */
+					break;
```
slipher, line 2168: Also free here?

src/engine/renderer/tr_image.cpp
```diff
@@ -2223,13 +2244,73 @@ image_t *R_FindCubeImage( const char *imageName, imageParams_t &imageParams )
...
+						}
+					}
+
+					// We have to break because the expected loader was found.
+					break;
+				}
+			}
+		}
+
+		/* If the image path is “some.thing” for a file named “some.thing.crn”,
```
slipher, line 2295: These comments are written in a confusing way. You could say `If we are give the name "some.thing", first we look for some.thing.webp, some.thing.tga etc... Then, some.webp, some.tga, etc.`

src/engine/renderer/tr_image.cpp
```diff
@@ -2223,13 +2244,73 @@ image_t *R_FindCubeImage( const char *imageName, imageParams_t &imageParams )
...
+		first try loading with “some.thing” as a basename. */
+		const imageExtLoader_t *loader = R_FindCubeLoader( fileName );
+
+		/* Then if the image path is “something.crn” for a file named “something.crn”,
+		try loading with “something” as a basename. */
+		if ( *ext && !loader )
+		{
+			COM_StripExtension3( token, fileName, sizeof(fileName) );
```
slipher, line 2303: The `token` temporary pointer shouldn't be used after calling a bunch of other functions.

src/engine/renderer/tr_image.cpp
illwieckz: Done.

slipher:
LGTM

PR #1398, opened 2024-10-27
Fix vertex-lit surfaces with material system

VReaperV:
Don't overwrite `tess.svars.color` by `Tess_ComputeColor()` after setting it with `SetVertexLightingSettings()`. Fixes some surfaces being whitened out, like crates, bricks and boxes on `terminus`.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -294,6 +294,8 @@ void UpdateSurfaceDataLightMapping( uint32_t* materials, Material& material, dra
 	colorGen_t rgbGen = SetRgbGen( pStage );
 	alphaGen_t alphaGen = SetAlphaGen( pStage );
 
+	Tess_ComputeColor( pStage );
+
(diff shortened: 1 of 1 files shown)


slipher:
LGTM

illwieckz:
Does it also fix `r_lightMode 1` with material system?

VReaperV:
...was it broken? I would need to check because I wasn't aware of that.

VReaperV:
> Does it also fix `r_lightMode 1` with material system?

Just tried and it does, indeed, fix that.

illwieckz:
Yes, I noticed it was broken, but I didn't find time reporting it before with the release rush, and since `r_lightMode 1` is meant to be used in presets for hardware not supporting the material system, and because material system is experimental and not enabled by default, it was not blocking the release. Thanks for fixing that! 😀️

VReaperV:
Rebasing first because CI failed to start for some reason.

PR #1399, opened 2024-10-27
Add setWindowOrigin command

slipher:
Could be useful in scripting automatic testing and stuff. I can put it at, say, (-2000 -2000) to make the window not visible. (Tested on Windows; not sure if that works on all systems.)

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -2992,3 +2992,23 @@ void GLimp_LogComment( const char *comment )
 					 strlen( buf ), buf );
 	}
 }
+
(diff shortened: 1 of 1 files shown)


DolceTriade:
We should probably test this elsewhere

slipher:
Now tested. On Linux (KDE, Wayland), it works. You can place the window completely off-screen in any direction.
On Mac it also works. The window can't be moved past the top of the screen (*y* is clamped to 0). It can be placed partially off the 3 other sides but is clamped at the point where just a little bit of the window remains on-screen.

On all systems, screenshots work even if the window is offscreen. (This was a concern for me even since screenshots while minimized often don't work.)

PR #1402, opened 2024-10-29
readme: use same color as Unvanquished readme for irc link, do not advertise #daemon-engine anymore

illwieckz:
readme: use same color as Unvanquished readme for irc link, do not advertise #daemon-engine anymore


illwieckz:
The `#daemon-engine` was used when we were doing things with Xonotic guys, but no active developer use it anymore, we better redirect anyone to `#unvanquished-dev` that is enough for the job.

DolceTriade (approved):

PR #1404, opened 2024-10-30
Fix overbright bugs

illwieckz:
- `renderer: cancel overbright on added lightmap`

It fixes the perseus glass_02 door.

The code was slightly rewritten too.

- `renderer: disable GLSL overbright on fullbright lightmode globally instead of doing it per-texture`

It fixes rendering fullbright lightmap stages when light mode is not fullbright,
like the starchart512 hologram in procyon map.

Fix Unvanquished/Unvanquished#3161 and #1405:

- https://github.com/Unvanquished/Unvanquished/issues/3161
- https://github.com/DaemonEngine/Daemon/issues/1405

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5239,33 +5239,55 @@ static void FinishStages()
 		? gl_shaderManager.getDeformShaderIndex( shader.deforms, shader.numDeforms )
 		: 0;
 
+	bool isOpaqueShader = false;
+
 	for ( size_t s = 0; s < numStages; s++ )
 	{
 		shaderStage_t *stage = &stages[ s ];
 
 		stage->deformIndex = deformIndex;
 
-		// We should cancel overbright if there is no light stage.
-		stage->cancelOverBright = shaderHasNoLight;
+		// SRC1 and DST0 are reset to zero in ParseStage (no blending).
+		bool isOpaque = !( stage->stateBits & ( GLS_SRCBLEND_BITS | GLS_DSTBLEND_BITS ) );
+
+		// A shader is not opaque if all stages are not opaques.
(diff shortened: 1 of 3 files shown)


illwieckz:
Perseus:

unvanquished_2024-10-30_054726_000

Procyon:

unvanquished_2024-10-30_075957_000

src/engine/renderer/tr_shader.cpp
```diff
@@ -5239,35 +5239,57 @@ static void FinishStages()
...
 
 		stage->deformIndex = deformIndex;
 
-		// We should cancel overbright if there is no light stage.
-		stage->cancelOverBright = shaderHasNoLight;
+		// SRC1 and DST0 are reset to zero in ParseStage (no blending).
+		bool isOpaque = !( stage->stateBits & ( GLS_SRCBLEND_BITS | GLS_DSTBLEND_BITS ) );
+
+		// A shader is not opaque if all stages are not opaques.
+		isOpaqueShader |= isOpaque;
```
VReaperV, line 5254: To clarify: is the intention to make the shader opaque when at least 1 stage is opaque, or if all stages are opaque?

illwieckz: It is opaque if at least one stage is opaque.

illwieckz: In fact the information for the whole shader is not used, but when processing a stage, we must know if there is nothing opaque between the background the shader will be painted over, and the current stage of the shader being processed.

VReaperV (approved):
LGTM.

slipher:
Thoughts on #1406?

illwieckz:
> Thoughts on #1406?

To be investigated after 0.55.1.

PR #1406, opened 2024-10-30
Simplify overbright implementation

slipher:
AIUI overbright simply means that the lightmaps and light grid should be scaled from 0 to 4, instead of 0 to 1, so we can represent a larger range of brightness. But image colors are only allowed to go from 0 to 1. This is a big problem... if we are using GL 1 (hence, perhaps, the mysterious "hardware implementations"). But we have shaders, so all we have to do is multiply by 4 whenever we read one of those special light textures. Problem solved?

I tried this approach and it looks the same to me, consistent with the complex cancellng system that we had before. Also fixes the Perseus glass and Procyon star chart issues addressed in #1404.

An explanation for the canceling approach is given in #1050:
> We cannot overBright separate stage as stage are clamped within [0.0, 1.0] when blending multiple stages together, meaning the result of the multiplication of a separate light stage with a light factor will be clamped before blending it with the color map.

But this is not an accurate representation of how shaders work. Each stage acts independently and directly on the color and depth buffers. There is no scratch area where shader stages are blended before going into the color buffer.

I would expect this to ameliorate the color banding problems with low-precision color buffers since now we don't give up 2 bits of precision to make room for canceling.

I haven't tried vertex lighting yet. Not sure how it's supposed to work with overbright.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -222,15 +222,12 @@ void UpdateSurfaceDataGeneric3D( uint32_t* materials, Material& material, drawSu
 	// u_AlphaThreshold
 	gl_genericShaderMaterial->SetUniform_AlphaTest( pStage->stateBits );
 
-	// u_InverseLightFactor
-	float inverseLightFactor = pStage->cancelOverBright ? tr.mapInverseLightFactor : 1.0f;
-	gl_genericShaderMaterial->SetUniform_InverseLightFactor( inverseLightFactor );
-
 	// u_ColorModulate
 	colorGen_t rgbGen = SetRgbGen( pStage );
 	alphaGen_t alphaGen = SetAlphaGen( pStage );
 
-	gl_genericShaderMaterial->SetUniform_ColorModulate( rgbGen, alphaGen );
+	bool mayUseVertexOverbright = pStage->type == stageType_t::ST_COLORMAP && drawSurf->bspSurface;
+	gl_genericShaderMaterial->SetUniform_ColorModulate( rgbGen, alphaGen, mayUseVertexOverbright );
 
 	Tess_ComputeColor( pStage );
 	gl_genericShaderMaterial->SetUniform_Color( tess.svars.color );
@@ -291,12 +288,9 @@ void UpdateSurfaceDataLightMapping( uint32_t* materials, Material& material, dra
 	bool enableGridLighting = ( lightMode == lightMode_t::GRID );
 	bool enableGridDeluxeMapping = ( deluxeMode == deluxeMode_t::GRID );
 
-	// u_InverseLightFactor
-	/* HACK: use sign to know if there is a light or not, and
-	then if it will receive overbright multiplication or not. */
-	bool cancelOverBright = pStage->cancelOverBright || lightMode == lightMode_t::FULLBRIGHT;
-	float inverseLightFactor = cancelOverBright ? tr.mapInverseLightFactor : -tr.mapInverseLightFactor;
-	gl_lightMappingShaderMaterial->SetUniform_InverseLightFactor( inverseLightFactor );
+	// u_LightFactor
+	gl_lightMappingShaderMaterial->SetUniform_LightFactor(
+		lightMode == lightMode_t::FULLBRIGHT ? 1.0f : tr.mapLightFactor );
 
 	// u_ColorModulate
 	gl_lightMappingShaderMaterial->SetUniform_ColorModulate( rgbGen, alphaGen );
@@ -470,9 +464,6 @@ void UpdateSurfaceDataSkybox( uint32_t* materials, Material& material, drawSurf_
 	// u_AlphaThreshold
 	gl_skyboxShaderMaterial->SetUniform_AlphaTest( GLS_ATEST_NONE );
 
-	// u_InverseLightFactor
-	gl_skyboxShaderMaterial->SetUniform_InverseLightFactor( tr.mapInverseLightFactor );
-
 	gl_skyboxShaderMaterial->WriteUniformsToBuffer( materials );
 }
 
@@ -619,9 +610,6 @@ void UpdateSurfaceDataFog( uint32_t* materials, Material& material, drawSurf_t* 
 	drawSurf->initialized[stage] = true;
 
 	const fog_t* fog = material.fog;
-
-	// u_InverseLightFactor
-	gl_fogQuake3ShaderMaterial->SetUniform_InverseLightFactor( tr.mapInverseLightFactor );
 
 	// u_Color
 	gl_fogQuake3ShaderMaterial->SetUniform_Color( fog->color );
--- a/src/engine/renderer/gl_shader.h
+++ b/src/engine/renderer/gl_shader.h
@@ -2150,21 +2150,6 @@ public:
 	}
 };
 
-class u_InverseLightFactor :
-	GLUniform1f
-{
-public:
-	u_InverseLightFactor( GLShader *shader ) :
-		GLUniform1f( shader, "u_InverseLightFactor" )
-	{
-	}
-
-	void SetUniform_InverseLightFactor( const float inverseLightFactor )
-	{
-		this->SetValue( inverseLightFactor );
-	}
-};
-
 class u_ColorMap :
 	GLUniformSampler2D {
 	public:
@@ -3617,7 +3602,7 @@ public:
 	{
 		this->SetValue( v );
 	}
-	void SetUniform_ColorModulate( colorGen_t colorGen, alphaGen_t alphaGen )
+	void SetUniform_ColorModulate( colorGen_t colorGen, alphaGen_t alphaGen, bool vertexOverbright = false )
 	{
 		vec4_t v;
 		bool needAttrib = false;
@@ -3631,7 +3616,16 @@ public:
 		{
 			case colorGen_t::CGEN_VERTEX:
 				needAttrib = true;
-				VectorSet( v, 1, 1, 1 );
+				if ( vertexOverbright )
+				{
+					// vertexOverbright is only needed for non-lightmapped cases. When there is a
+					// lightmap, this is done by multiplying with the overbright-scaled white image
+					VectorSet( v, tr.mapLightFactor, tr.mapLightFactor, tr.mapLightFactor );
+				}
+				else
+				{
+					VectorSet( v, 1, 1, 1 );
+				}
 				break;
 
 			case colorGen_t::CGEN_ONE_MINUS_VERTEX:
@@ -3946,7 +3940,6 @@ class GLShader_generic :
 	public u_AlphaThreshold,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
-	public u_InverseLightFactor,
 	public u_ColorModulate,
 	public u_Color,
 	public u_Bones,
@@ -3976,7 +3969,6 @@ class GLShader_genericMaterial :
 	public u_AlphaThreshold,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
-	public u_InverseLightFactor,
 	public u_ColorModulate,
 	public u_Color,
 	public u_DepthScale,
@@ -4015,7 +4007,7 @@ class GLShader_lightMapping :
 	public u_ViewOrigin,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
-	public u_InverseLightFactor,
+	public u_LightFactor,
 	public u_Bones,
 	public u_VertexInterpolation,
 	public u_ReliefDepthScale,
@@ -4067,7 +4059,7 @@ class GLShader_lightMappingMaterial :
 	public u_ViewOrigin,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
-	public u_InverseLightFactor,
+	public u_LightFactor,
 	public u_ReliefDepthScale,
 	public u_ReliefOffsetBias,
 	public u_NormalScale,
@@ -4113,7 +4105,6 @@ class GLShader_forwardLighting_omniXYZ :
 	public u_ViewOrigin,
 	public u_LightOrigin,
 	public u_LightColor,
-	public u_InverseLightFactor,
 	public u_LightRadius,
 	public u_LightScale,
 	public u_LightAttenuationMatrix,
@@ -4157,7 +4148,6 @@ class GLShader_forwardLighting_projXYZ :
 	public u_ViewOrigin,
 	public u_LightOrigin,
 	public u_LightColor,
-	public u_InverseLightFactor,
 	public u_LightRadius,
 	public u_LightScale,
 	public u_LightAttenuationMatrix,
@@ -4208,7 +4198,6 @@ class GLShader_forwardLighting_directionalSun :
 	public u_ViewOrigin,
 	public u_LightDir,
 	public u_LightColor,
-	public u_InverseLightFactor,
 	public u_LightRadius,
 	public u_LightScale,
 	public u_LightAttenuationMatrix,
@@ -4275,7 +4264,6 @@ class GLShader_reflection :
 	public u_NormalScale,
 	public u_VertexInterpolation,
 	public u_CameraPosition,
-	public u_InverseLightFactor,
 	public GLDeformStage,
 	public GLCompileMacro_USE_VERTEX_SKINNING,
 	public GLCompileMacro_USE_VERTEX_ANIMATION,
@@ -4300,7 +4288,6 @@ class GLShader_reflectionMaterial :
 	public u_ReliefOffsetBias,
 	public u_NormalScale,
 	public u_CameraPosition,
-	public u_InverseLightFactor,
 	public GLDeformStage,
 	public GLCompileMacro_USE_HEIGHTMAP_IN_NORMALMAP,
 	public GLCompileMacro_USE_RELIEF_MAPPING {
@@ -4319,8 +4306,7 @@ class GLShader_skybox :
 	public u_UseCloudMap,
 	public u_AlphaThreshold,
 	public u_ModelMatrix,
-	public u_ModelViewProjectionMatrix,
-	public u_InverseLightFactor
+	public u_ModelViewProjectionMatrix
 {
 public:
 	GLShader_skybox( GLShaderManager *manager );
@@ -4337,8 +4323,7 @@ class GLShader_skyboxMaterial :
 	public u_UseCloudMap,
 	public u_AlphaThreshold,
 	public u_ModelMatrix,
-	public u_ModelViewProjectionMatrix,
-	public u_InverseLightFactor {
+	public u_ModelViewProjectionMatrix {
 	public:
 	GLShader_skyboxMaterial( GLShaderManager* manager );
 	void SetShaderProgramUniforms( shaderProgram_t* shaderProgram ) override;
@@ -4349,7 +4334,6 @@ class GLShader_fogQuake3 :
 	public u_FogMap,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
-	public u_InverseLightFactor,
 	public u_Color,
 	public u_Bones,
 	public u_VertexInterpolation,
@@ -4370,7 +4354,6 @@ class GLShader_fogQuake3Material :
 	public u_FogMap,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
-	public u_InverseLightFactor,
 	public u_Color,
 	public u_FogDistanceVector,
 	public u_FogDepthVector,
@@ -4389,7 +4372,6 @@ class GLShader_fogGlobal :
 	public u_ViewMatrix,
 	public u_ModelViewProjectionMatrix,
 	public u_UnprojectMatrix,
-	public u_InverseLightFactor,
 	public u_Color,
 	public u_FogDistanceVector,
 	public u_FogDepthVector
@@ -4484,8 +4466,7 @@ public:
 class GLShader_contrast :
 	public GLShader,
 	public u_ColorMap,
-	public u_ModelViewProjectionMatrix,
-	public u_InverseLightFactor
+	public u_ModelViewProjectionMatrix
 {
 public:
 	GLShader_contrast( GLShaderManager *manager );
@@ -4499,7 +4480,6 @@ class GLShader_cameraEffects :
 	public u_ColorModulate,
 	public u_TextureMatrix,
 	public u_ModelViewProjectionMatrix,
-	public u_LightFactor,
 	public u_DeformMagnitude,
 	public u_InverseGamma
 {
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -23,6 +23,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 #define COMPUTELIGHT_GLSL
 
+uniform float u_LightFactor;
+
 #if !defined(USE_BSP_SURFACE)
 	#define USE_MODEL_SURFACE
 #endif
@@ -55,6 +57,8 @@ vec4 EnvironmentalSpecularFactor( vec3 viewDir, vec3 normal )
 		float directedScale = 2.0 - ambientScale;
 		ambientColor = ambientScale * texel.rgb;
 		lightColor = directedScale * texel.rgb;
+		ambientColor *= u_LightFactor;
+		lightColor *= u_LightFactor;
 	}
 #endif
 
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -2210,7 +2210,6 @@ GLShader_generic::GLShader_generic( GLShaderManager *manager ) :
 	u_AlphaThreshold( this ),
 	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
-	u_InverseLightFactor( this ),
 	u_ColorModulate( this ),
 	u_Color( this ),
 	u_Bones( this ),
@@ -2243,7 +2242,6 @@ GLShader_genericMaterial::GLShader_genericMaterial( GLShaderManager* manager ) :
 	u_AlphaThreshold( this ),
 	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
-	u_InverseLightFactor( this ),
 	u_ColorModulate( this ),
 	u_Color( this ),
 	u_DepthScale( this ),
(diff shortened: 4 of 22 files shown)


DolceTriade:
Screenies?

VReaperV:
> I haven't tried vertex lighting yet. Not sure how it's supposed to work with overbright.

Vertex lighting seems to use inverse light factor the same way.

slipher:
From a recent master, default settings:
unvanquished-atcshd-outside
unvanquished-nano-overmind

With this PR:
unvanquished-atcshd-outside
unvanquished-nano-overmind

illwieckz:
I'm all in favor of a simpler implementation.

In the past I first tried what you did there, but I got bugs. Maybe they were mistakes I did?

My main concern was about surfaces where the lightmap is applied more than once (and then multiplied more than once), basically terrain blended surfaces. Fortunately with your branch those terrain blended surfaces look fine (thunder outdoor and station15 garden):

unvanquished_2024-10-31_035141_000

unvanquished_2024-10-31_034606_000

On the station15 garden, plant leaves are wrongly lit, maybe they are vertex lit.

So, here is the light sources that should be multiplied:

- lightmap light
- vertex light
- grid light
- white fullbright light WHEN the lightmode is not fullbright

illwieckz:
```
> - white fullbright light WHEN the lightmode is not fullbright
```

Doing this properly is one of the fix from #1404 (the starchart issue) and it should be done in all implementations.

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -45,6 +47,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 		float directedScale = 2.0 - ambientScale;
 		ambientColor = ambientScale * texel.rgb;
 		lightColor = directedScale * texel.rgb;
+		ambientColor /= u_InverseLightFactor;
+		lightColor /= u_InverseLightFactor;
```
illwieckz, line 51: Better do some `something *= u_LightFactor` instead.

illwieckz:
On my end when I first tried your approach, I was having such places reproducing the same clamping as we get whih the legacy clamped overbright:

unvanquished_2024-10-31_041056_000

But as you see in this screenshot, your branch doesn't reproduce the clamping. That clamping in GLSL was also what prompted me to follow a more complex route to avoid that clamping.

I don't know why I got any clamping in the past even when doing overbright in GLSL that way. The GLSL code received heavy rewrites, maybe we fixed some bugs in the meantime?

slipher:
> So, here are the light sources that should be multiplied:
> 
>     * lightmap light
> 
>     * vertex light
> 
>     * grid light
> 
>     * white fullbright light WHEN the lightmode is not fullbright

Lightmap and grid light are what is implemented in the initial draft of this PR. I don't think the last case exists for the simplified approach; it sounds like a problem of canceling.

Vertex light is a tiny bit more complex than the other cases since everything has the vertex color attributes, not only things with precomputed lighting. Maybe I can check `tess.bspSurface` and scale u_ColorModulate by 4 if is true.

illwieckz:
Here is an example of GLSL overbrigth clamping when I first applied the overbright on light per stage instead of screen:

unvanquished_2024-02-06_175946_000

I don't get why I got it and you're not getting it, but if we're not getting it with the per-light approach, that's great!

illwieckz:
> I don't think the last case exists for the simplified approach; it sounds like a problem of canceling.

Probably not, the fullbright light mode is to just apply a white lightmap, so it's just using the lightmap code. Maybe there is nothing to do.

Though, we must make sure the (multiplied) vertex light code doesn't multiple the (multiplied) white lightmap fallback.

illwieckz:
I reproduce the clamped overbright with that GLSL implementation, now I remember when it's happening.

The clamped overbright happens with such GLSL implementation when the lightmap is blended as a separate stage (i.e. no collapsing is doable), when the engine processes a lightmap stage over a colormap stage (or the colormap stage over a lightmap stage), instead of a singe diffusemap stage:

This:

```
stage colormap
stage lightmap
```

Or this:

```
stage lightmap
stage colormap
```

Instead of:

```
stage diffusemap
```

Here is an example with a build of the pulse map attached (the one hosted on sweet folder has broken lightstyles so it doesn't reproduce the light clamping bug):

The `master` branch:

unvanquished_2024-10-31_043029_000

This branch:

unvanquished_2024-10-31_044425_000

[map-pulse_0-20230207-114641-e9a0a5e.dpk.zip](https://github.com/user-attachments/files/17581693/map-pulse_0-20230207-114641-e9a0a5e.dpk.zip)

I guess I don't reproduce the bug with plat23 anymore because I improved the collapse code and we now collapse more materials, getting them processed as a single diffusemap stage instead of separate colormap + lightmap stages.

illwieckz:
To make it more obvious, the problem is where the red arrows are pointing at:

unvanquished_2024-10-31_044425_001

illwieckz:
So I actually reproduce the bug with a released Tremulous map that is the original map build by the original author (only repackaged, not rebuilt):

Map: https://users.unvanquished.net/~sweet/pkg/map-metro-b1-2_0%2B0.dpk

viewpos: `2123 -231 -166 -89 26`

With `master`:

unvanquished_2024-10-31_051304_000

With this branch:

unvanquished_2024-10-31_051423_000

illwieckz:
Interestingly, your branch fixes the (what is assumed to be a) floating light shader that should not be visible:

Map: https://users.unvanquished.net/~sweet/pkg/map-metro-b1-2_0%2B0.dpk

viewpos; `2267 -429 -223 -51 -19`

branch `master` or `illwieckz/obb`:

unvanquished_2024-10-31_052550_000

branch `slipher/overbright-simple`:

unvanquished_2024-10-31_052655_000

illwieckz:
So basically, I'm all in favor of investing this more and if we can make it handle light styles properly, let's do it.

But for the 0.55.1 release we better merge my fix, we'll have all the time to investigate a different implementation after that.

slipher:
I looked at the Metro light style shaders. So apparently it builds a composite lightmap in the color buffer from 4 additive stages, then does a standard multiplicative blend with the diffusemap in the final stage.

I guess we could keep the cancellation approach for light styles only. There is already detection and special handling for light styles anyway.

illwieckz:
> I guess we could keep the cancellation approach for light styles only. There is already detection and special handling for light styles anyway.

If that's doable that would be awesome! I dislike the cancelling-everywhere approach a lot (it's very fragile because of the very large amounts of uses case…).

VReaperV:
> > So, here are the light sources that should be multiplied:
> > 
> > * lightmap light
> > * vertex light
> > * grid light
> > * white fullbright light WHEN the lightmode is not fullbright
> 
> Lightmap and grid light are what is implemented in the initial draft of this PR. I don't think the last case exists for the simplified approach; it sounds like a problem of canceling.
> 
> Vertex light is a tiny bit more complex than the other cases since everything has the vertex color attributes, not only things with precomputed lighting. Maybe I can check `tess.bspSurface` and scale u_ColorModulate by 4 if is true.

Vertex lighting is multiplied with the diffuse colours, so multiplying it by 4 again would be wrong. 

Although on master vertex lighting is also somewhat broken with overbright: e. g. on map `railgun` from Wolf:ET the icicles are almost completely black without overbright clamping, owing to the fact that they're entirely vertex-lit. Perhaps there needs to be a check for surfaces that use only vertex lighting, which would then be multiplied or whatever. That shouldn't be done in tess however, just multiply the vertex colours themselves instead.

illwieckz:
> Although on master vertex lighting is also somewhat broken with overbright: e. g. on map railgun from Wolf:ET the icicles are almost completely black without overbright clamping

Probably the same bug as the procyon hologram?

VReaperV:
> > Although on master vertex lighting is also somewhat broken with overbright: e. g. on map railgun from Wolf:ET the icicles are almost completely black without overbright clamping
> 
> Probably the same bug as the procyon hologram?

I doubt it, the procyon one has a lightmap.

slipher:
Besides canceling, another possible approach to light styles would be using multiple collapse lightmap stages with additive blending, and disable dynamic lights in all but one.

illwieckz:
> disable dynamic lights in all but one.

I guess it is already the case, at least with the tiled renderer: light styles light map are expected to be rendered with `Renger_generic3D` and not with `Render_lightMapping` as far as I know.

illwieckz:
> multiple collapse lightmap

Actually I don't see how that would works? to collapse lightmaps with what?

illwieckz:
This may be theoretical, but we may not exclude that some non-lightstyle lightmaps rendered with the lightMapping stage get blended like the light style ones and reproduce the bugs that happens with light styles.

It's just that the collapsing has been well improved so standalone non-light-style lightmap stages are more rare nowadays, but it is not guaranteed we collapse all of them or that we even can collapse all of them.

illwieckz:
@slipher do you have any understanding about how the light style lightmap blending is working and then how we may workaround the related bugs?

Actually we may workaround some wrongness by normalizing to `1` at the end of the GLSL shader, to avoid bad colors popping up there and there. But that will still clamp them (it would still be better than the legacy lightmap shift that is not actually clamping but turns 75%s of the data into garbage).

illwieckz:
I added a commit to multiply the light factor instead of dividing the inverse light factor, that also cleaned-up code, so it helps to have a better look at things.

src/engine/renderer/glsl_source/computeLight_fp.glsl
illwieckz: I added some commits to do that.

VReaperV:
> Interestingly, your branch fixes the (what is assumed to be a) floating light shader that should not be visible:
> 
> Map: https://users.unvanquished.net/~sweet/pkg/map-metro-b1-2_0%2B0.dpk
> 
> viewpos; `2267 -429 -223 -51 -19`
> 
> branch `master` or `illwieckz/obb`:
> 
> unvanquished_2024-10-31_052550_000
> 
> branch `slipher/overbright-simple`:
> 
> unvanquished_2024-10-31_052655_000
> 
> The metro map is the only map known to reproduce this floating transparent surface bug, so it's low priority, but anything helping to pinpoint what is happening is welcome.

I'm guessing it's the weird square on the wall that is the issue, right?

illwieckz:
> I'm guessing it's the weird square on the wall that is the issue, right?

For the “floating light shader”, yes.

illwieckz:
I rebased on current master, fixed some merge conflict, and cleaned-up more cancelling leftovers.

@VReaperV I would love if you could look at that broken light style at said position in map with your magical tools and if you can actually describe what's going on and what could explain the light being that bad.

Maybe light styles are the only things that require a workaround when we multiply the lightmap directly in color/lightmapping shader.

slipher:
> I rebased on current master, fixed some merge conflict, and cleaned-up more cancelling leftovers.

@illwieckz I would suggest getting rid of your commit `fixup: only multiply generic3D with light factor if it's a light style lightmap`. This means treating the affected light map like as if overbright bits were configured to 0 (instead of 2).

VReaperV:
> I rebased on current master, fixed some merge conflict, and cleaned-up more cancelling leftovers.
> 
> @VReaperV I would love if you could look at that broken light style at said position in map with your magical tools and if you can actually describe what's going on and what could explain the light being that bad.
> 
> Maybe light styles are the only things that require a workaround when we multiply the lightmap directly in color/lightmapping shader.

Apparently it's some surface that has white image as both diffuse and lightmap, and is rendered using the `lightMapping` shader. It's also vertex-lit. On master at least, that is.

VReaperV:
> I rebased on current master, fixed some merge conflict, and cleaned-up more cancelling leftovers.
> 
> @VReaperV I would love if you could look at that broken light style at said position in map with your magical tools and if you can actually describe what's going on and what could explain the light being that bad.
> 
> Maybe light styles are the only things that require a workaround when we multiply the lightmap directly in color/lightmapping shader.

So the reason it doesn't show up on this branch is because the surface is rendered with blendfunc GL_DST_COLOR , GL_ZERO and GL_FUNC_ADD, and this branch produces a completely white fragment, while master doesn't.

The "before" value is essentially multiplied with the fragment value:
On master:
image
On this branch:
image

While on master `u_InverseLightFactor` is set to -0.25 for this surface, which means it's not used and the fragment essentially gets the vertex colours, this branch sets `u_LightFactor` to 4.0 for this surface, and then the lightmap colour is multiplied by it, resulting in light colour of {4.0, 4.0, 4.0, 1.0}, which is then multiplied with the vertex colours, which in turn is enough to create a {1.0, 1.0, 1.0, 1.0} fragment.

illwieckz:
> So the reason it doesn't show up on this branch is because the surface is rendered with blendfunc GL_DST_COLOR , GL_ZERO and GL_FUNC_ADD, and this branch produces a completely white fragment, while master doesn't.

Hmm, is this an explanation for why the square light emitting shader is displayed with `master` branch?

unvanquished_2024-10-31_052550_001

I was looking for an explanation for why the light style is broken with this branch, so that weirdly colored halo on the wall:

unvanquished_2024-10-31_052655_001

I suspect some kind of clamping.

VReaperV:
> Hmm, is this an explanation for why the square light emitting shader is displayed with `master` branch?

Yep.

> I was looking for an explanation for why the light style is broken with this branch, so that weirdly colored halo on the wall:

...oh. I didn't realise that was the issue lol.

illwieckz:
That thread was a big confusing I agree. 😄️

Before comment https://github.com/DaemonEngine/Daemon/pull/1406#issuecomment-2454011193 I was talking about the light emitting shader, but starting with that one I was talking about light styles. 🙂️

illwieckz:
unvanquished_2024-11-05_115510_000

🤔️

VReaperV:
That looks more like overdark to me.

slipher:
Dark map on material system fixed.

Trying a new approach where we configure the FBO to allow colors greater than 1. This also has side effects though...

unvanquished_2024-11-05_051138_000

VReaperV:
Given that `u_LightFactor` is only used for lightmap and lightgrid, can't we just multiply those and avoid the shader hack altogether?

illwieckz:
> Given that `u_LightFactor` is only used for lightmap and lightgrid, can't we just multiply those and avoid the shader hack altogether?

This is the original implementation making it clamped…

VReaperV:
> > Given that `u_LightFactor` is only used for lightmap and lightgrid, can't we just multiply those and avoid the shader hack altogether?
> 
> This is the original implementation making it clamped…
> 
> Edit: lightmap, lightgrid, and vertex color.

We could try using a different image format there.

illwieckz:
I added two changes:

1. Also multiply the lightmaps used in light style.
2. Do not multiply fullbright light, I first believed this should also be multiplied, but it was not the case before.

The second change fixes the procyon hologram:

unvanquished_2024-11-06_182740_000

illwieckz:
The metro visible “visible light emitting shader” is a tricky one.

As a remind, the lantern, the candle, etc. come with some shaders having a single stage like this:

```
	{
		map $lightmap
		blendfunc filter
	}
```

So we get floating lightmaps multiplied over whatever is behind.

I even don't know why it worked on Tremulous…

Maybe we can just strip them out…

illwieckz:
I added a commit to fix the lantern/candle. I assume it's a shader bug, it doesn't make sense. That patch has nothing to do with overbright and could be done in another PR if we wanted to.

illwieckz:
I don't see any bugs anymore, in all maps I've tested (metro, procyon, pulse…)

illwieckz:
So, I noticed there was something still wrong with leaves in station15 garden.

Here is 0.54.1, the leaves on the floor are correct:

unvanquished_2024-11-07_135251_000

Here is this branch before we only multiplied light style/light map/diffuse map stages, the leaves on the floor are too dark:

unvanquished_2024-11-07_135339_000

HEre is this branch when I multiplied all colormaps (which is wrong), just to verify something, the leaves on the floor are correct, but the ones in tree are too bright:

unvanquished_2024-11-07_135416_000

Here is once I added the patch multiplying color map with lightFactor when rgbGen is `vertex`:

unvanquished_2024-11-07_140242_000

So here is an example of such leaf-on-the-ground shader:

```
textures/station15_custom/leave02d
{
	qer_editorImage textures/station15_custom_src/leave02d_d
	qer_alphaFunc greater .5

	q3map_cloneShader textures/station15_custom/leave02d_back
	q3map_vertexScale 1.5

	surfaceparm alphashadow
	surfaceparm trans
	surfaceparm nonsolid
	surfaceparm nomarks

	{
		map textures/station15_custom_src/leave02d_d
		rgbGen vertex
		depthWrite
		alphaFunc GE128
	}
}
```

Such shader doesn't have a lightmap, so it's processed as a generic shader.

But the colormap is blended with `rgbGen vertex` meaning it's multiplied with the vertex light.

So this is an old trick to do single-stage vertex-based lighting.

And indeed the old overbright code was multiplying vertex colors at map load.

So we need to multiply the vertex color like if it is a light.

I don't know yet why the foam is too bright, though.

illwieckz:
I guess we should do something when using `rgbgen oneMinusVertex` because the vertex light was multiplied too before.

illwieckz:
Here is the foam shader:

```
gfx/station15/haze
{
	entityMergable
	cull none
	{
		map gfx/station15/water_haze
		blendFunc blend
		rgbGen vertex
		alphaGen vertex
	}
}
```

I don't know what to do to fix it.

illwieckz:
> I guess we should do something when using `rgbGen oneMinusVertex` because the vertex light was multiplied too before.
> 
> But I don't know what to do in such case.

I haven't found a single occurrence of `rgbGen oneMinusVertex` in the whole betaserv dump.

There are some `alphGen oneMinusVertex` in some maps like transit, but I can't find usage of `oneMinusVertex` with `rgbGen`.

illwieckz:
Maybe the haze always have been too bright but clamping hidden the bug.

illwieckz:
If I remove `rgbgen vertex` from the haze shader, it looks correct:

unvanquished_2024-11-07_150056_000

So I guess the previous clamped overbright just set the vertex color to `{1, 1, 1}`, making the texture fullbright, but that was wrong.

illwieckz:
It is possible to precompute the per-stage light factor at map loading time, but that would require to rewrite some of the BSP loading code (most shaders are loaded before the global light factor is known, and there are some tricky chicken-and-egg problem to solve).

illwieckz:
The haze effect is not a map surface, it's a particle emitted surface, so I guess it has no vertex color, and we may work this around.

illwieckz:
> The haze effect is not a map surface, it's a particle emitted surface, so I guess it has no vertex color, and we may work this around.

Now done:

unvanquished_2024-11-07_162100_000

In fact such surface should better be collapsed as a diffuse map with vertex lighting, so they can also receive dynamic lights.

They are meant to receive light, it's just a trick to use vertex light explicitely instead of a lightmap, and the light map code already calls functions like `isExplicitelyVertexLitSurface()`.

Such collapse is to be postponed for the future.

illwieckz:
I added a commit to add `gl_generic3DShader`, this way every call of `gl_genericShader` don't need to receive any `lightFactor` (it no need to spam all the debug code to set it to 1). I have not stripped `gl_genericShader` from all the macros that are now useless, but we can do it.

illwieckz:
Once material system is fixed, this will be ready to merge.

VReaperV:
> @VReaperV I reported the changes to the material system, but I probably did a stupid mistake somewhere as most textures are black, can you look at it?

Yeah, will do in a bit.

src/engine/renderer/Material.cpp
```diff
@@ -214,45 +214,45 @@ void UpdateSurfaceDataGeneric3D( uint32_t* materials, Material& material, drawSu
...
 
 	// u_AlphaThreshold
-	gl_genericShaderMaterial->SetUniform_AlphaTest( pStage->stateBits );
+	gl_generic3DShaderMaterial->SetUniform_AlphaTest( pStage->stateBits );
 
-	// u_InverseLightFactor
-	float inverseLightFactor = pStage->cancelOverBright ? tr.mapInverseLightFactor : 1.0f;
-	gl_genericShaderMaterial->SetUniform_InverseLightFactor( inverseLightFactor );
+	// u_LightFactor
+	float lightFactor = ( pStage->type == stageType_t::ST_STYLELIGHTMAP || pStage->type == stageType_t::ST_STYLECOLORMAP || ( tess.bspSurface && pStage->rgbGen == colorGen_t::CGEN_VERTEX ) ) ? tr.mapLightFactor : 1.0f;
```
VReaperV, line 226: Redundant parentheses.

VReaperV, line 226: Use `drawSurf->bspSurface` here, `tess.bspSurface` won't be valid in this functions.

src/engine/renderer/gl_shader.h
```diff
@@ -2140,7 +2140,7 @@ class u_LightFactor :
 {
 public:
 	u_LightFactor( GLShader *shader ) :
-		GLUniform1f( shader, "u_LightFactor" )
+		GLUniform1f( shader, "u_LightFactor", true )
```
VReaperV, line 2143: If `u_LightFactor` can change between surfaces, then this should not be global.

VReaperV (commented):
Why the change to have 3 generic shaders instead of 2?

VReaperV:
I'll check now if fixing the above makes it work with material system.

VReaperV:
@illwieckz I'm not getting any black textures:
image
Can you confirm if the issue is still there on your end?

illwieckz:
> Why the change to have 3 generic shaders instead of 2?

To avoid the need of calling `SetUniform_InverseLightFactor( 1.0 )` on every function using the generic shader (like debug tools), and the risk of forgetting about it.

illwieckz:
> Can you confirm if the issue is still there on your end?

I'll test this after sleeping. 🙂️

slipher:
I'd like to avoid adding u_LightFactor to the generic shader at all. I think we can do it via u_Color instead. We can add a flag to make Tess_ComputeColor scale by the light factor. I'll try this later.

illwieckz:
I still have one bug with the material system, off:

unvanquished_2024-11-09_134108_000

on:

unvanquished_2024-11-09_134213_000

You'll see the plants are too bright.

(Don't mind if the floor and the trees look a bit different than stock map, I'm also experimenting other things with the map, but those plants are not modified).

VReaperV:
It's because the vertex colour attribute is disabled. I'll look at why that happens.

VReaperV:
> I still have one bug with the material system, off:
> 
> unvanquished_2024-11-09_134108_000
> 
> on:
> 
> unvanquished_2024-11-09_134213_000
> 
> You'll see the plants are too bright.
> 
> (Don't mind if the floor and the trees look a bit different than stock map, I'm also experimenting other things with the map, but those plants are not modified).

Fixed now.

VReaperV (approved):
LGTM

VReaperV:
Since I only did a small change here and the pool of people who can review this isn't much larger than this pr's contributors, I took the liberty of reviewing it, especially as I reviewed most of it earlier anyway.

illwieckz:
I squashed the “fixup” commits.

This looks ready to me.

There are some improvements that may be doable:

- setting per-stage lightFactor at shader load time (this requires to rewrite some BSP loading code),
- pre-multiplying lightFactor with u_Color to not have to pass it to GLSL (this may benefit from the former),
- turn colormap stage with vertex color blending to a diffusemap one when the shader has no “nolightmap” keyword (so it can also receive dynamic light).

But that can be done in the future in subsequent PRs.

illwieckz:
I moved the fix for the metro lanterns here:

- https://github.com/DaemonEngine/Daemon/pull/1426

illwieckz:
I also renamed the PR to fit what it does, as it is now in good shape for merging and we will going to do things that way.

slipher:
Please remember that according to our PR etiquette, no one should merge another person's PR :)

slipher:
I've now reimplemented the three special cases treated by illwieckz in way that avoids adding `u_LightFactor` to the generic shader. Those fixes are separated into a temporary fixup commit for convenient reading.

Let us call BASE the source tree obtained from merging #1426, #1431 and #1434 into master. Let us call SIMPLE this PR's tree, the `slipher/overbright/simple` branch (which incorporates the three aforementioned PRs). To verify the changes I have, using the scenes listed [here](https://github.com/slipher/imgdiff/blob/master/imgdiff.cfg), done a four-way comparison of BASE with all default settings, BASE with material system, SIMPLE with all default settings, and SIMPLE with material system. (Please suggest more scenes!)

## Some bugs fixed by these overbright changes

### Non-lightmapped vertex-lit surface not overbright scaled (station15 `462 -1428 -114 178 -8`)
BASE

SIMPLE

### Improperly scaled 'glow map' (zittrig-arena `1201 -301 -338 87 0`)
BASE ( + material system)
unvanquished_2024-11-12_010216_000

SIMPLE ( + material system)
unvanquished_2024-11-12_010340_000

### Procyon star chart darkened with material system
BASE + material system

unvanquished-procyon-starchart

SIMPLE + material system
unvanquished-procyon-starchart

## Consequences for framebuffers with only 1 byte of color

In some cases things are worse because of clamping in the color buffer that wasn't there before. In some cases things are a bit better because of the effective increased precision. It may be preferable to enable overbright clamping on low-end systems that don't have 16-bit float framebuffers.

### Color clamping regression - Pulse `4822 -1312 -1314 98 13`
BASE + `r_highPrecisionRendering 0`
unvanquished-pulse-lowprecision-clamping

SIMPLE + `r_highPrecisionRendering 0`
unvanquished-pulse-lowprecision-clamping

### Decreased noise in dark area - Procyon `-955 3443 276 94 7`
BASE + `r_highPrecisionRendering 0`
unvanquished-procyon-lowprecision-noise

SIMPLE + `r_highPrecisionRendering 0`
unvanquished-procyon-lowprecision-noise

src/engine/renderer/Material.cpp
```diff
@@ -222,15 +222,12 @@ void UpdateSurfaceDataGeneric3D( uint32_t* materials, Material& material, drawSu
...
-	float inverseLightFactor = pStage->cancelOverBright ? tr.mapInverseLightFactor : 1.0f;
-	gl_genericShaderMaterial->SetUniform_InverseLightFactor( inverseLightFactor );
-
 	// u_ColorModulate
 	colorGen_t rgbGen = SetRgbGen( pStage );
 	alphaGen_t alphaGen = SetAlphaGen( pStage );
 
-	gl_genericShaderMaterial->SetUniform_ColorModulate( rgbGen, alphaGen );
+	bool mayUseVertexOverbright = pStage->type == stageType_t::ST_COLORMAP && drawSurf->bspSurface;
+	gl_genericShaderMaterial->SetUniform_ColorModulate( rgbGen, alphaGen, mayUseVertexOverbright );
```
illwieckz, line 230: What is preventing to multiply `tess.svars.color` in `Tess_ComputeColor`?

Why is the material variant calling ``Tess_ComputeColor` but not the non-material one? Is that a dormant bug we missed?

VReaperV: Non-material code does `Tess_ComputeColor()` in `Tess_StageIteratorColor()`. Material system does not have that.

illwieckz: Ah so in non-material code, `Tess_ComputeColor()` is done before calling `Render_generic3D()`. Is it too early to multiply the light factor? Because both `tess` and `pStage` are known as well.

illwieckz (approved):
LGTM.

src/engine/renderer/Material.cpp
VReaperV: > Ah so in non-material code, `Tess_ComputeColor()` is done before calling `Render_generic3D()`. Is it too early to multiply the light factor? Because both `tess` and `pStage` are known as well.

I suppose that was just to avoid having to add it to all relevant stage renderers.

PR #1408, opened 2024-10-31
Make fog into a stage, fix some issues with it

### LINKED ISSUE #537: SSAO displayed above fog from behind wall
On this screenshot (hangar28 map), there is a fog between the player's eyes and the wall, though, the SSAO dark corner is visible.

[ssao above fog from behind wall](https://dl.illwieckz.net/b/daemon/bugs/ssao-above-fog-from-behind-wall/unvanquished_2021-10-13_020303_000.jpg)

I noticed SSAO is not done on skybox textures, so maybe we can exclude such surface as well.

For a start, we must check if SSAO is drawn on surfaces with `nolightmap`, I guess it must not.

VReaperV:
Changes fog from being done based on an if statement in `Tess_StageIteratorColor()` to being added as a surface with one of two new defaut fog shaders - similar to how we add depth pre-pass surfaces. The 2 new default shaders are for fogs with FP_EQUAL and FP_LE fog passes.

- Adds fog to material system. Fog surfaces are added as stages to the main surface, like it's already done with depth pre-pass surfaces.
- Slightly reworked how depth and fog surfaces are linked to the main surface, now `R_AddDrawSurf()` returns the index of the surface, to get the correct one.
- Moved some early-return fog checks to shader parsing/`R_AddDrawSurf()`.
- Added `u_FogMap` since all fogs use the same image, so it can be made global, and cleaned up `Render_fog()` and fog processing a bit.
- Fixes #537 by rendering all fog surfaces with `SS_FOG`, which is done after SSAO.
- Fixed incorrect max stage count for material system.

### DIFF
--- a/src/engine/renderer/glsl_source/fogQuake3_fp.glsl
+++ b/src/engine/renderer/glsl_source/fogQuake3_fp.glsl
@@ -24,7 +24,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 #define FOGQUAKE3_GLSL
 
-uniform sampler2D	u_ColorMap;
+uniform sampler2D u_FogMap;
 
 uniform float u_InverseLightFactor;
 IN(smooth) vec3		var_Position;
@@ -37,7 +37,7 @@ void	main()
 {
 	#insert material_fp
 
-	vec4 color = texture2D(u_ColorMap, var_TexCoords);
+	vec4 color = texture2D(u_FogMap, var_TexCoords);
 
 	color *= var_Color;
 
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -3520,7 +3520,7 @@ inline bool checkGLErrors()
 	void Render_portal( shaderStage_t *pStage );
 	void Render_heatHaze( shaderStage_t *pStage );
 	void Render_liquid( shaderStage_t *pStage );
-	void Render_fog( shaderStage_t* /* pStage */ );
+	void Render_fog( shaderStage_t* pStage );
 
 	/*
 	============================================================
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1034,6 +1034,7 @@ enum class shaderProfilerRenderSubGroupsMode {
 	  ST_PORTALMAP,
 	  ST_HEATHAZEMAP, // heatHaze post process effect
 	  ST_LIQUIDMAP,
+	  ST_FOGMAP,
 	  ST_LIGHTMAP,
 	  ST_STYLELIGHTMAP,
 	  ST_STYLECOLORMAP,
@@ -1288,6 +1289,7 @@ enum class shaderProfilerRenderSubGroupsMode {
 		} altShader[ MAX_ALTSHADERS ]; // state-based remapping; note that index 0 is unused
 
 		struct shader_t *depthShader;
+		struct shader_t *fogShader;
 		struct shader_t *next;
 	};
 
@@ -1609,6 +1611,7 @@ enum class shaderProfilerRenderSubGroupsMode {
 		shader_t      *shader;
 		uint64_t      sort;
 		bool          bspSurface;
+		int fog;
 
 		uint materialsSSBOOffset[ MAX_SHADER_STAGES ];
 		bool initialized[ MAX_SHADER_STAGES ];
@@ -1618,6 +1621,7 @@ enum class shaderProfilerRenderSubGroupsMode {
 		uint drawCommandIDs[ MAX_SHADER_STAGES ];
 
 		drawSurf_t* depthSurface;
+		drawSurf_t* fogSurface;
 		bool materialSystemSkip = false;
 
 		inline int index() const {
@@ -2737,6 +2741,8 @@ enum class shaderProfilerRenderSubGroupsMode {
 
 		// internal shaders
 		shader_t *defaultShader;
+		shader_t *fogEqualShader;
+		shader_t *fogLEShader;
 		shader_t *defaultPointLightShader;
 		shader_t *defaultProjectedLightShader;
 		shader_t *defaultDynamicLightShader;
@@ -3093,7 +3099,7 @@ inline bool checkGLErrors()
 
 	void           R_AddPolygonSurfaces();
 
-	void           R_AddDrawSurf( surfaceType_t *surface, shader_t *shader, int lightmapNum, int fogNum, bool bspSurface = false );
+	int R_AddDrawSurf( surfaceType_t *surface, shader_t *shader, int lightmapNum, int fogNum, bool bspSurface = false );
 
 	void           R_LocalNormalToWorld( const vec3_t local, vec3_t world );
 	void           R_LocalPointToWorld( const vec3_t local, vec3_t world );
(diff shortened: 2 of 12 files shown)


slipher:
What are some maps with fog to try it out? The only one I know off the top of my head is atcshd.

VReaperV:
The other 2 I know of are https://users.unvanquished.net/~sweet/pkg/map-assault_0%2Bv7.1-beta4.dpk and https://users.unvanquished.net/~sweet/pkg/map-hangar28_1.dpk.

src/engine/renderer/tr_local.h
```diff
@@ -3093,7 +3099,7 @@ inline bool checkGLErrors()
 
 	void           R_AddPolygonSurfaces();
 
-	void           R_AddDrawSurf( surfaceType_t *surface, shader_t *shader, int lightmapNum, int fogNum, bool bspSurface = false );
+	int           R_AddDrawSurf( surfaceType_t *surface, shader_t *shader, int lightmapNum, int fogNum, bool bspSurface = false );
```
illwieckz, line 3102: Same (and in fact alignment is already lost, so better strip that space).

illwieckz (approved):
LGTM! And it works well on my end!

src/engine/renderer/tr_shader.cpp
```diff
@@ -7155,11 +7177,41 @@ static void CreateInternalShaders()
...
 	stages[ 0 ].bundle[ 0 ].image[ 0 ] = tr.defaultImage;
 	stages[ 0 ].active = true;
 	stages[ 0 ].stateBits = GLS_DEFAULT;
 	tr.defaultShader = FinishShader();
+
+	Q_strncpyz( shader.name, "<fogEqual>", sizeof( shader.name ) );
+
+	shader.type = shaderType_t::SHADER_3D_DYNAMIC;
+	shader.sort = Util::ordinal( shaderSort_t::SS_FOG );
+	shader.noFog = true;
```
slipher, line 7192: I don't see why we should specially disable fog for the not-found shader.

src/engine/renderer/tr_shade.cpp
```diff
@@ -2325,52 +2325,43 @@ void Render_liquid( shaderStage_t *pStage )
...
-	{
+	if ( materialSystem.generatingWorldCommandBuffer ) {
+		Tess_DrawElements();
 		return;
 	}
 
-	// ydnar: no world, no fogging
-	if ( backEnd.refdef.rdflags & RDF_NOWORLDMODEL )
+	// no fog pass in snooper
+	// WTF is a snooper?
```
slipher, line 2336: This I guess: https://wolfenstein.fandom.com/wiki/Snooper_rifle

src/engine/renderer/tr_shader.cpp
```diff
@@ -5971,6 +5963,14 @@ static shader_t *FinishShader()
 		}
 	}
 
+	if ( shader.sort <= Util::ordinal( shaderSort_t::SS_OPAQUE ) ) {
+		shader.fogPass = fogPass_t::FP_EQUAL;
+	} else if ( shader.contentFlags & CONTENTS_FOG ) {
+		shader.fogPass = fogPass_t::FP_LE;
+	}
+
+	shader.noFog = shader.noFog || shader.fogPass == fogPass_t::FP_NONE || ( backEnd.refdef.rdflags & RDF_NOWORLDMODEL );
```
slipher, line 5980: rdflags can change each frame so you can't use it here.

src/engine/renderer/tr_shade.cpp
```diff
@@ -2352,19 +2347,20 @@ static void Render_fog()
...
 	fogDistanceVector[ 1 ] = -backEnd.orientation.modelViewMatrix[ 6 ];
 	fogDistanceVector[ 2 ] = -backEnd.orientation.modelViewMatrix[ 10 ];
 	fogDistanceVector[ 3 ] = DotProduct( local, backEnd.viewParms.orientation.axis[ 0 ] );
 
 	// scale the fog vectors based on the fog's thickness
-	fogDistanceVector[ 0 ] *= fog->tcScale;
-	fogDistanceVector[ 1 ] *= fog->tcScale;
-	fogDistanceVector[ 2 ] *= fog->tcScale;
-	fogDistanceVector[ 3 ] *= fog->tcScale;
+	VectorScale( fogDistanceVector, fog->tcScale, fogDistanceVector );
```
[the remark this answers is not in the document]
slipher: Apparently it was fixed in a subsequent commit

src/engine/renderer/glsl_source/fogQuake3_fp.glsl
```diff
@@ -24,7 +24,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 #define FOGQUAKE3_GLSL
 
-uniform sampler2D	u_ColorMap;
+uniform sampler2D	u_FogMap;
```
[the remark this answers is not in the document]
slipher: What do you mean?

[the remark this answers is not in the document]
illwieckz: The `uniform sampler2D u_FogMap;`, actually that's a tab and not many spaces, but fixing the ugliness of the code when modifying is the best way to prevent noisy restyling commits. Nothing is blocking there, just noticing the code can be beautified at the same time and that would save us a noisy restyle commit.

src/engine/renderer/tr_shader.cpp
VReaperV: Hmm, I thought it's only used in menus and such, but apparently it's only used for hyperspace.

src/engine/renderer/tr_shade.cpp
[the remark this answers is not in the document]
VReaperV: Yeah, I didn't realize I put the initial changed into the previous commit at the time.

src/engine/renderer/tr_shader.cpp
VReaperV: In fact, given that it's not used for anything else and hyperspace already has its own flag, I think `RDF_NOWORLDMODEL` can just be NUKED. Out of scope for this pr though.

src/engine/renderer/tr_shade.cpp
VReaperV: Perhaps. Apparently Wolf:ET had a check for `tr.refdef.rdflags & RDF_SNOOPERVIEW` at https://github.com/id-Software/Enemy-Territory/blob/40342a9e3690cb5b627a433d4d5cbf30e3c57698/src/renderer/tr_shade.c#L803.

src/engine/renderer/tr_shader.cpp
VReaperV: Anyway, fixed the initial issue.

src/engine/renderer/tr_shade.cpp
VReaperV: I've removed that comment now because it doesn't apply to daemon.

src/engine/renderer/glsl_source/fogQuake3_fp.glsl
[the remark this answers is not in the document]
VReaperV: Alright, done.

src/engine/renderer/tr_local.h
VReaperV: Also done now.

src/engine/renderer/tr_shader.cpp
VReaperV: I suppose that line can be dropped.

VReaperV: I tried it, it seems to work fine, so I added that change.

PR #1409, opened 2024-11-01
sort the output of command listBinds

sweet235:
I have no idea how else people are supposed to find anything in that list.

### DIFF
--- a/src/engine/client/key_binding.cpp
+++ b/src/engine/client/key_binding.cpp
@@ -465,6 +465,7 @@ public:
 
 	void Run(const Cmd::Args&) const override
 	{
+		std::vector<std::string> lines;
 		for (auto& kv: keys)
 		{
 			bool teamSpecific = false;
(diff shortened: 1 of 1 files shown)


src/engine/client/key_binding.cpp
```diff
@@ -491,11 +493,18 @@ class ListBindsCmd: public Cmd::StaticCmd
...
-						Print( "%s [%s] → %s", ExtraInfoKeyString( kv.first ), teamName[ team ], kv.second.binding[ team ].value() );
+						std::string line = ExtraInfoKeyString( kv.first ) + " [" + teamName[ team ] + "] → " + kv.second.binding[ team ].value();
+						lines.push_back( line );
 					}
 				}
 			}
 		}
+
+		std::sort( lines.begin(), lines.end() );
+		for ( std::string line : lines )
```
illwieckz, line 504: Does `for ( std::string& line : lines )` works? It may avoid a copy.

slipher:
It would be preferable if sorting only changed the order of keys, rather than the order of lines within a key. So that the default team bind stayed at the beginning instead of coming after aliens.

> I have no idea how else people are supposed to find anything in that list.

If you want to see what a particular key is bound to, use `/bind <key>`.

slipher:
LGTM

PR #1411, opened 2024-11-02
Fix compile error

VReaperV:


DolceTriade:
What's the compile error

VReaperV:
Some function calls that were removed in #1317 and accidentally re-introduced in #1408.

slipher:
LGTM

Issue #119, opened 2018-08-28
partial void effect on station15 tree model


illwieckz:
See that weird void effect at the tip of the conifer branch:

[![incorrect](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/unvanquished_2018-08-28_230842_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/unvanquished_2018-08-28_230842_000.jpg)

It must look like this:

[![correct](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/unvanquished_2018-08-28_233039_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/unvanquished_2018-08-28_233039_000.jpg)

Note that the bug seems to occur on the collision with the plant leaf model. For information, there is also an alphagen brush in source map at this place:

[![netradiant](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/20180828-232944-000.netradiant-station15-tree-plant-alphagen-brush.png)](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/20180828-232944-000.netradiant-station15-tree-plant-alphagen-brush.png)

This bug looks somewhat related to that bug noticed in wolf:et map (at least for the tree):
https://github.com/Unvanquished/Unvanquished/issues/799#issuecomment-253367127

illwieckz:
@gimhael the somwhat fix for #118 somewhat fix this (the texture is right but lighting is wrong):

[![incorrect](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/unvanquished_2018-08-28_230842_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/unvanquished_2018-08-28_230842_000.jpg)

[![somewhat correct](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/unvanquished_2018-08-28_231610_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/nolightmap-bug/unvanquished_2018-08-28_231610_000.jpg)

It also somewhat fix Unvanquished/Unvanquished#799 bug at least for the tree part:

[...]

PR #1412, opened 2024-11-04
Allow disabling depth prepass with r_depthShaders

slipher:
This cvar can be used to demonstrate that #119 is caused by a faulty depth shader. Or to quantify the performance benefits of depth shaders.

AFAIK disabling them shouldn't affect the final result of rendering, so it does not need to be a cheat var. But I could be wrong. So in case the performance benefits turn out to be negative on some system, depth shaders could be turned off.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -81,6 +81,12 @@ Cvar::Cvar<bool> r_dpBlend("r_dpBlend", "Enable DarkPlaces blend compatibility, 
 static Cvar::Cvar<float> r_portalDefaultRange(
 	"r_portalDefaultRange", "Default portal range", Cvar::NONE, 1024);
 
+// This can be turned off to debug problems with depth shaders.
+// Almost everything can be rendered correctly without them, except real-time lights
+// (since light tiles are populated using the depth buffer).
+static Cvar::Cvar<bool> r_depthShaders(
+	"r_depthShaders", "use depth pre-pass shaders", Cvar::CHEAT, true);
+
 /*
 ================
 return a hash value for the filename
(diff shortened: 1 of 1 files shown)


slipher:
Another bug suspected to be caused by a faulty depth shader is #334. This could be tested by reverting the workaround 3c7061af2e2da7b219380d726f062f7eb9466e95. I don't know which map has this though.

VReaperV:
> But I could be wrong.

The benefit is performance, yes. Also, I think you can make it non-latch and instead do the check in `R_AddDrawSurface()`.

VReaperV:
> AFAIK disabling them shouldn't affect the final result of rendering, so it does not need to be a cheat var. But I could be wrong.

Actually, there is. We have some shaders that use the current depth image: generic, global fog, motion blur and SSAO. While global fog and SSAO would be rendered after all depth is written anyway, the other 2 won't.

VReaperV:
Also, it would probably break portals.

slipher:
> Also, I think you can make it non-latch and instead do the check in `R_AddDrawSurface()`.

Just tried this but it wasn't totally convenient since the depth mask bits of the main shader are modified depending on whether there is a depth shader. I'd have to make a backup copy of the depth bits.



> We have some shaders that use the current depth image: generic, global fog, motion blur, SSAO and dynamic light shaders (depthtile1/2 and lighttile). While global fog and SSAO would be rendered after all depth is written anyway, the other 5 won't.

Hmm yeah the light tile is a deal breaker. I don't think any of the others are problems but lighting definitely is. I'll make this a cheat cvar for debugging then.

VReaperV (approved):
LGTM

slipher:
I have realized that even dynamic lights can be made to work, if we disable the last two planes in the lighttile frustum bounds check. But I will do this later.

PR #1413, opened 2024-11-04
system: load `<libpath>/../pkg` if `<libpath>/pkg` doesn't exist but `<libpath>/../pkg` does

illwieckz:
Load `<libpath>/../pkg` if `<libpath>/pkg` doesn't exist but `<libpath>/../pkg` does.

This to workaround the fact most file browsers will create a subfolder when extracting an archive that doesn't contain one.

This makes `Unvanquished_0.56.0/linux-amd64/daemon` load `Unvanquished_0.56.0/pkg` the same way `Unvanquished_0.56.0/daemon` loads `Unvanquished_0.56.0/pkg`.


illwieckz:
Using `FS::RawPath::ListFiles()` is overkill but we don't have yet a dedicated function to test that a directory exists.

slipher:
I'm against it. Going into `..` to look for resources seems like a fishy and insecure design.

Curiously, this could cause a build in `Unvanquished/build` to automatically find the repository's `pkg/`. This could seem "convenient" but I think it is a bad idea. I don't want stuff to unpredictably change behavior based on how you choose the build dir. And testing with only released assets is an often desired use case.

illwieckz:
If we had a builtin mini-GUI in the engine to run when there is no dpk, we could display a minimal file browser where people could just walk directories for a pakpath (basically going to parent dir and selecting pkg/). I believe Darkplaces does something a bit like that.

I'm also unhappy with the idea of the engine looking at the parent directory. But I'm also unhappy abot the zip thing almost never working as expected. We may also make that mechanism stronger by only looking at the parent dir if the basename of the current dir is something like `linux-amd64` (and variants).

slipher:
How about adding some `extract.sh` to the zip file?

Or the README instructions could be changed to recommend moving `pkg` into the new folder which might be easier.

illwieckz:
> How about adding some `extract.sh` to the zip file?

It would not solving the problem of starting a binary properly if the solution requires to start a binary properly. Also Linux systems doesn't run .sh file this easily in file browser, making it more complicated than a simple binary.

> Or the README instructions could be changed to recommend moving `pkg` into the new folder which might be easier.

That still needs a README to give instruction, so it doesn't change much. Actually such README instruction to move pkg in may help on macOS.

PR #1416, opened 2024-11-04
Some material system fixes and cleanup

VReaperV:
Fixes a material system crash where the first dynamic surface shader has a non-0 padding.
Fixes incorrect drawSurf index being returned by `R_AddDrawSurf()`.

Made `u_DepthMap` and `u_PortalMap` into global uniforms, and cleaned up some of the material system code.


src/engine/renderer/Material.cpp
```diff
@@ -705,6 +705,13 @@ void MaterialSystem::GenerateWorldMaterialsBuffer() {
 			const uint32_t padding = ( offset % paddedSize == 0 ) ? 0 : paddedSize - ( offset % paddedSize );
 
 			offset += padding;
+			
+			// Make sure padding is taken into account for dynamicDrawSurfsOffset
+			if ( !dynamicDrawSurfOffsetSet ) {
+				dynamicDrawSurfsOffset = offset;
```
slipher, line 700: So this variable needs to point to the end of the first material? That seems weird.

VReaperV: To the start of the first dynamic material, rather. Since the buffer is structured like this:
```
staticMaterial_0
...
staticMaterial_last
padding
dynamicMaterial_0
...
```
Padding after material `n` depends on the material `n + 1`, and is added at the `offset += padding;` line (last dynamic material never has any padding after it). `dynamicDrawSurfsOffset` is then used as the start of the range when mapping the buffer to update it, so it has to point to the index directly after the last static material padding.

slipher:
LGTM

PR #1419, opened 2024-11-05
Material system: u_LightFactor must be global

slipher:
This makes the Procyon star chart render, albeit incorrectly.

But that should be fixed soon by #1406.


slipher:
Never mind, mixed up stuff between branches.

PR #1421, opened 2024-11-06
Clean up snapshot mess with #including gamelogic

slipher:
A shared header defined snapshot_t which was used in both the client and cgame, but with different definitions of the player state struct. For the cgame definition of playerState_t it included the file shared/bg_public.h from the gamelogic.
Instead of this define an ipcSnapshot_t type which has a consistent definition and is used for IPC messages. Now snapshot_t will be defined and used only in the cgame.

Companion: https://github.com/Unvanquished/Unvanquished/pull/3182

### DIFF
--- a/src/engine/client/cg_api.h
+++ b/src/engine/client/cg_api.h
@@ -25,10 +25,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define CGAPI_H
 
 #include "engine/qcommon/q_shared.h"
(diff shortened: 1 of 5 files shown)


VReaperV:
This can be rebased on `for-0.56.0/sync` now.

slipher:
@illwieckz Does the LGTM on #1430 apply to this also?

slipher:
Bump

VReaperV (approved):
LGTM

PR #1422, opened 2024-11-07
Remove random aspect-based screen clearing

slipher:
The logic was this: if the user is not in-game and the window's aspect ratio is greater than 4:3, clear the screen. Otherwise do nothing.


VReaperV (approved):
LGTM

PR #1423, opened 2024-11-07
Don't use reflective specular for dynamic lights

### LINKED ISSUE #1358: Reflections are added for each dynamic light
`computeDynamicLight()` calls `computeDeluxeLight()`, which then adds the reflection colour if reflections are available. A simple fix would be to have an extra input or something to distinguish between the main light (lightmap) and the dynamic lights (or having 2 functions). However, if screen-space reflections are also added we'd need to put the reflection data into some sort of g-buffer (although doing SSR in the lightMapping shader might work too...), so this would no longer be an issue if both are enabled. Either way, this can be fixed as this issue isn't noticeable, at least visually.

slipher:
It's wrong because in the reflection cubemaps,
the world geometry is already pre-multiplied with the world lighting. For example if the location were completely dark but you added a static light, the specular reflection of the light should appear (but wouldn't previously).

Fixes #1358.

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -32,9 +32,19 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #endif
 
 #if defined(USE_REFLECTIVE_SPECULAR)
-	uniform samplerCube u_EnvironmentMap0;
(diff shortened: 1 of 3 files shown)


slipher:
I was motivated to do this as a start to untangling the Gordian knot of `computeDeluxeLight` (which combines the loosely related functions of deluxe mapping, half-Lambert lighting, physical mapping, and specular mapping). I need to do that to fix #1424.

SomaZ (approved):
Looks good. Makes more sense now.

PR #1425, opened 2024-11-09
Fix dynamic lights intensity, cleanup

### LINKED ISSUE #61: Very bright dynamic lights on some maps
From https://github.com/Unvanquished/Unvanquished/issues/1010:
> Screenshots:
> 
> unvanquished_2017-06-01_193056_000
> unvanquished_2017-06-01_193102_000
> 
> Here's map where it happens (I needed to put pk3 in another zip to stop github from complaining):
> 
> [map-tremtest_1.0.pk3.zip](https://github.com/Unvanquished/Unvanquished/files/1046026/map-tremtest_1.0.pk3.zip)
> 
> For me this bug happens on 0.50. @illwieckz says that he has this bug on master.
> 
Crossposting it here since it looks like a postprocessing issue to me.

VReaperV:
Cgame-side pr: https://github.com/Unvanquished/Unvanquished/pull/3184
Dynamic light config update for weapon flash colours: https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/32

- NUKED the weird 4x dynamic light scaling
- NUKED the weird `r_lightScale` cvar, which is used for dynamic lights, but it's a cheat cvar and always set to 2.0, and essentially only used for dynamic lights spawned from movers
- NUKED `RE_AddDynamicLightToSceneQ3A()` and renamed `RE_AddDynamicLightToSceneET()` to `RE_AddDynamicLightToScene()` because they're essentially the same
- NUKED now unused `trap_R_AddAdditiveLightToScene()`
- NUKED light intensity/scale, which was only used to scale the light's colour, just multiply the colour earlier on instead
- Fixed some incorrect memory read before bounds check

Fixes #61:
`hq-beta28`:
Before:
unvanquished_2024-11-09_184524_000
After:
unvanquished_2024-11-09_184725_000

`tremtest`:
Before:
unvanquished_2024-11-09_184614_000
After:
unvanquished_2024-11-09_184754_000

### DIFF
--- a/src/engine/renderer/tr_scene.cpp
+++ b/src/engine/renderer/tr_scene.cpp
@@ -285,10 +285,8 @@ RE_AddDynamicLightToScene
 ydnar: modified dlight system to support separate radius and intensity
 =====================
 */
-void RE_AddDynamicLightToScene( const vec3_t org, float radius, float intensity, float r, float g, float b, int flags )
+void RE_AddDynamicLightToScene( const vec3_t org, float radius, float r, float g, float b, int flags )
 {
-	trRefLight_t *light;
-
 	if ( !glConfig2.realtimeLighting || !r_drawDynamicLights.Get() )
 	{
 		return;
@@ -304,11 +302,12 @@ void RE_AddDynamicLightToScene( const vec3_t org, float radius, float intensity,
 		return;
 	}
 
-	if ( intensity <= 0 || radius <= 0 )
+	if ( radius <= 0 )
 	{
 		return;
 	}
 
+	trRefLight_t* light;
 	// set last lights restrictInteractionEnd if needed
 	if ( r_numLights > r_firstSceneLight ) {
 		light = &backEndData[ tr.smpFrame ]->lights[ r_numLights - 1 ];
@@ -347,10 +346,9 @@ void RE_AddDynamicLightToScene( const vec3_t org, float radius, float intensity,
 
 	light->additive = true;
 
-	if( light->l.inverseShadows )
-		light->l.scale = -intensity;
-	else
-		light->l.scale = intensity;
+	if ( light->l.inverseShadows ) {
+		VectorNegate( light->l.color, light->l.color );
+	}
 }
 
 static void RE_RenderCubeProbeFace( const refdef_t* originalRefdef ) {
--- a/src/engine/renderer/tr_scene.cpp
+++ b/src/engine/renderer/tr_scene.cpp
@@ -299,14 +299,6 @@ void RE_AddDynamicLightToScene( const vec3_t org, float radius, float intensity,
 		return;
 	}
 
-	// set last lights restrictInteractionEnd if needed
-	if ( r_numLights > r_firstSceneLight ) {
-		light = &backEndData[ tr.smpFrame ]->lights[ r_numLights - 1 ];
-		if( light->restrictInteractionFirst >= 0 ) {
-			light->restrictInteractionLast = r_numEntities - r_firstSceneEntity - 1;
-		}
-	}
-
 	if ( r_numLights >= MAX_REF_LIGHTS )
 	{
 		return;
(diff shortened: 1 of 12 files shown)


slipher:
Why don't we nuke the `intensity` argument? It seems to be redundant with just changing the r/g/b values.

VReaperV:
> Why don't we nuke the `intensity` argument? It seems to be redundant with just changing the r/g/b values.

Right, done now.

src/engine/renderer/tr_scene.cpp
```diff
@@ -354,10 +353,9 @@ void RE_AddDynamicLightToScene( const vec3_t org, float radius, float intensity,
 
 	light->additive = true;
 
-	if( light->l.inverseShadows )
-		light->l.scale = -intensity;
-	else
-		light->l.scale = intensity;
+	if ( light->l.inverseShadows ) {
```
slipher, line 356: The forward lighting shader checks the sign of u_LightScale to check for inverse lights. These are said to be already broken, but it would be nice to set the scale to 1 or -1 to at least gesture at how it was supposed to work

VReaperV: `scale` is already removed there, so there's nothing to set.

VReaperV:
> I tested this branch vs. master and found the dynamic lights are not as bright. I guess if you deleted a factor of 8 of random scaling, you need to add that back in somewhere to get the same result?
> 
> Master: unvanquished-plat23-dlight-psaw
> 
> This branch: unvanquished-plat23-dlight-psaw

I have created a PR in res-weapons repo for this: https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/32.

DolceTriade (approved):
tested ingame and feels good

slipher:
There are some maps with particle systems with dynamic lights, so not everything can be made the same by only adjusting the game’s own assets. 

Also there are some places where dynamic light intensities from config files are clamped to [0, 1] (yeah that’s stupid and shouldn’t be done), so that’s something we would need to watch out for if adjusting configs.

VReaperV:
> There are some maps with particle systems with dynamic lights, so not everything can be made the same by only adjusting the game’s own assets.

Which maps?

> Also there are some places where dynamic light intensities from config files are clamped to [0, 1] (yeah that’s stupid and shouldn’t be done), so that’s something we would need to watch out for if adjusting configs.

Yes, that is particles and beams/trails. Those look fine though.

slipher:
> > There are some maps with particle systems with dynamic lights, so not everything can be made the same by only adjusting the game’s own assets.
> 
> Which maps?

For example usstremor at viewpos 112 -306 86 6 2. Here is an [index](https://users.unvanquished.net/~slipher/particle-keyword-directory.txt) where all map particle systems with one can be found under `dynamiclight`. (Map trail systems could use them too, but none do.)


> 
> > Also there are some places where dynamic light intensities from config files are clamped to [0, 1] (yeah that’s stupid and shouldn’t be done), so that’s something we would need to watch out for if adjusting configs.
> 
> Yes, that is particles and beams/trails. Those look fine though.

We have particle and trail systems using dynamic lights in `res-weapons`, but I don't see any updates for them in https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/32.

Why don't we just add back the scaling factors to anywhere the lights are used on the gamelogic side? This is an easy way to avoid unwanted changes to assets. I doubt designers are coming in with a specific expectation like 'the illumination should be equivalent to fullbright at distance 200 with intensity 17.4', so the arbitrary factor doesn't hurt anything.

VReaperV:
> For example usstremor at viewpos 112 -306 86 6 2. Here is an [index](https://users.unvanquished.net/~slipher/particle-keyword-directory.txt) where all map particle systems with one can be found under `dynamiclight`. (Map trail systems could use them too, but none do.)

Thanks, I'll test those once I update this/Unvanquished-side branch.

> We have particle and trail systems using dynamic lights in `res-weapons`, but I don't see any updates for them in [UnvanquishedAssets/res-weapons_src.dpkdir#32](https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/32).

I tried with those initially, but those ended up looking wrong because of the 8-bit precision limit.

> Why don't we just add back the scaling factors to anywhere the lights are used on the gamelogic side? This is an easy way to avoid unwanted changes to assets. I doubt designers are coming in with a specific expectation like 'the illumination should be equivalent to fullbright at distance 200 with intensity 17.4', so the arbitrary factor doesn't hurt anything.

I agree, I've been thinking about doing just that.

VReaperV:
> We have particle and trail systems using dynamic lights in `res-weapons`, but I don't see any updates for them in [UnvanquishedAssets/res-weapons_src.dpkdir#32](https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/32).
> 
> Why don't we just add back the scaling factors to anywhere the lights are used on the gamelogic side? This is an easy way to avoid unwanted changes to assets. I doubt designers are coming in with a specific expectation like 'the illumination should be equivalent to fullbright at distance 200 with intensity 17.4', so the arbitrary factor doesn't hurt anything.

I've added a fix for that to the Unvanquished-side pr. Both the dynamic lights on maps and those produced by weapons look fine to me with that change.

PR #1427, opened 2024-11-09
Fix entity baseline overflow

VReaperV:
Split net messages if there are too many baselines. I don't like the solution too much because it only works for this exact issue, but I guess it'll do.

The layouts in https://users.unvanquished.net/~reaper/maps/layouts/spacetracks/ can be used to test this with map `spacetracks`.

Fixes https://github.com/Unvanquished/Unvanquished/issues/2124.

### DIFF
--- a/src/engine/client/cl_parse.cpp
+++ b/src/engine/client/cl_parse.cpp
@@ -403,11 +403,13 @@ void CL_ParseGamestate( msg_t *msg )
 
 	clc.connectPacketCount = 0;
 
-	// wipe local client state
-	CL_ClearState();
+	if ( !cl.reading ) {
+		// wipe local client state
+		CL_ClearState();
 
-	// a gamestate always marks a server command sequence
-	clc.serverCommandSequence = MSG_ReadLong( msg );
+		// a gamestate always marks a server command sequence
+		clc.serverCommandSequence = MSG_ReadLong( msg );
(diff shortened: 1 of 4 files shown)


DolceTriade:
Note that currently, servers can already work around the issue using https://github.com/DaemonEngine/Daemon/pull/1322 which will allow these to load at the host of high ping players getting a little bit of lag.

Also see https://github.com/DaemonEngine/Daemon/issues/1321 for some prior art on this topic.

Indeed this might improve the situation, but is it worth it when we have a workaround?

VReaperV:
CI now fails on an unrelated file (#1645 should fix that).

VReaperV:
I never said it would, it's targeting the for-0.56.0/sync branch rather than master.

slipher:
What's the motivation for rushing to merge it without review then?

VReaperV:
I wouldn't call it rushing given that the pr has existed in its current state for half a year. The only objection to it is similar to the one for map scripting: theoretical bullshit being used as an argument over a small change that works in practice and fixes crashes and situations where a server can be left largely inaccessible. So I'm doing what @illwieckz has said in regards to being bold with merges and not going into the mindset of having to wait for approval for clear improvements (and the theoretical argument presented earlier has so far generated 0 practical value).

PR #1428, opened 2024-11-09
NUKE unused IPC calls

VReaperV:


DolceTriade (commented):
If they are unused, do you think this change is safe in master (with a corresponding Unvanquished change)? I worry that deleting them in 0.56.0 will allow this to regress.

slipher:
LGTM. Doing it in the 0.56 branch is the right way. Having syscalls that work in some engine builds of the same ABI version but not others wouldn't be good.

DolceTriade:
Fair enough. Probably won't gain any uses of the syscalls and we'll jsut discover it when we release 0.56

PR #1430, opened 2024-11-10
Add a trivial cgame module which can view a map

slipher:
Stacked on #1421.

Add a cgame module that simply loads and displays a map, with no entities. It can be useful to quickly test how a map renders or to make sure a simple gamelogic builds correctly in the CI. `loadmap` is the command to enter a map. The only way to change position for now is `setviewpos`.

Future directions:
- Add controls for moving
- Draw some help text when not in a map
- Move commands like `testmodel` to Daemon and build them in
- Load and draw stationary entities

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,6 +147,10 @@ option(USE_ADDRESS_SANITIZER "Try to use the address sanitizer" OFF)
 option(BE_VERBOSE "Tell the compiler to report all warnings" OFF)
 option(USE_STATIC_LIBS "Tries to use static libs where possible. Only works for Linux" OFF)
 
+if (NOT DAEMON_PARENT_SCOPE_DIR)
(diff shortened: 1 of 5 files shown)


illwieckz:
is there some plan to make it loaded by default if there is no game?

slipher:
> is there some plan to make it loaded by default if there is no game?

No. The uses I have in mind are renderer testing, and making sure there are no build errors in the shared gamelogic. The way I deploy it for renderer testing is building it in a separate directory and using `-libpath` to grab its gamelogic.

slipher:
Fixed color grading in a less hacky way and changed a shader to `white` instead of `gfx/colors/white` so it will work with the daemon base pak. Now ready, though #1421 is still waiting.

illwieckz:
Is there something preventing to name this command `devmap`?

slipher:
There is an engine command named `devmap`. I don't use that since I want to do it without an sgame.

illwieckz (approved):
LGTM.

PR #1431, opened 2024-11-10
Set vertex colour attr correctly with material system

slipher:
Cherry-picked from #1406.

Before:
unvanquished-station15-vertexlight-tree

After:
unvanquished-station15-vertexlight-tree

I think this is still not complete actually, as there are also implicitly vertex lit surfaces that don't have `CGEN_VERTEX` set. Vertex lighting can be automatically selected when there is no lightmap. The computer console in Vega is an example. It uses an image loaded as shader.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -1315,11 +1315,13 @@ void ProcessMaterialLightMapping( Material* material, shaderStage_t* pStage, dra
 
 	DAEMON_ASSERT( !( enableDeluxeMapping && enableGridDeluxeMapping ) );
 
-	colorGen_t rgbGen = SetRgbGen( pStage );
+	// useAttrColor has no effect since the lightMapping shader has ATTR_COLOR forced to be always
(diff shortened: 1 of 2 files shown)


slipher:
The previous screenshots were with full-range overbright. Although this commit makes the material system match the standard renderer, the result is wrong (too dark).

With clamping enabled, the 'before' picture remains the same (too bright). But clamping + this commit looks just right:
unvanquished-station15-vertexlight-tree

illwieckz (approved):
LGTM.

The reason why it's too dark is because the overbright code is buggy, something that will be fixed with #1406. So actually getting it too dark is what is expected.

illwieckz:
What is “implicit vertex lighting”?

slipher:
I mean when the shader does not have `rgbgen vertex`, but vertex lighting is selected because it's a lighting stage of a BSP surface that has no lightmap.

illwieckz (approved):
LGTM.

PR #1434, opened 2024-11-12
Fix instant quad 'stageIteratorFunc == NULL' error

slipher:
This happens when I load plat23 with the dummy gamelogic. The surface rendering function calls Tess_CheckVBOAndIBO, which calls Tess_EndBegin, which is unhappy that there is no stage iterator set.


VReaperV (approved):
LGTM

Issue #1435, opened 2024-11-12
If two shaders / permutations produce the same code, do we compile them twice?


illwieckz:
If two shaders / permutations produce the same code, do we compile them twice?

illwieckz:
If not, we may do an index of cached checksum, not just an index of shader+permutation id so we can deduplicate both compile time and storage.

VReaperV:
The glsl caching code is a mess. It should be changed to be a straightforward "do all of the shader code processing, including adding macros, *then* check if we can load it based on that".

VReaperV:
> If two shaders / permutations produce the same code, do we compile them twice?

Is there an example of such permutations or shaders? If two permutations exist for the same shader that are created by setting different macros, yet produce the same results, then some macros should just be nuked.

illwieckz:
I wonder if that happens with generic2D.

VReaperV:
> I wonder if that happens with generic2D.

You mean permutations of it?

illwieckz:
Yes.

illwieckz:
I mean I wonder if some permutations of generic2D and some permutations of generic may do the same.

VReaperV:
> Yes.

There's only one: deph fade.

VReaperV:
> I mean I wonder if some permutations of generic2D and some permutations of generic may do the same.

Well, technically generic2D shader always has:
```
#if defined(GENERIC_2D)
	gl_FragDepth = 0;
#endif
```

VReaperV:
(the other one doesn't)

VReaperV:
The above is no longer true, now that the generic2D shader no longer uses that and is basically the same as generic one now.

VReaperV:
> If not, we may do an index of cached shader text checksum, not just an index of shader+permutation id so we can automatically deduplicate both compile time and storage.

This is probably the best option, especially considering some of the post-processing our shader sources go through, which is not cached properly I think.

VReaperV:
Note that `generic2D` shader and the relevant macro no longer exist.

VReaperV:
For only building the required shaders, see: https://github.com/DaemonEngine/Daemon/issues/1294#issuecomment-2378910816.

slipher:
> If two shaders / permutations produce the same code, do we compile them twice?

Yes. Our shader cache files things under the shader name and permutation number, so it would never notice if a checksum is equal to a different shader's checksum.

[...]

Issue #384, opened 2020-10-12
Display a Dæmon logo at startup, before everything else can be displayed and before the game can do anything


illwieckz:
When the engine starts, there is a certain amount of time between the moment the window is created and the game starts to display something. Basically, the game has to load the VFS and run the UI code to display something, but before that the displayed window is empty.

To avoid garbage being displayed at this early stage (see #373), PR #374 just fill the window with a neutral-like dark grey color as early as possible.

The next step to make things look good and more welcoming would be to display a simple logo.

The logo would have to be displayed as early as possible, so before VFS is loaded. Because of that the logo would be unique whatever the game, and would have to be embedded in engine for easy use.

To make it neutral regarding games running on Dæmon, we may imagine a Dæmon logo, about the engine itself. Also that would take part in building reputation about the engine itself, something like _“Powered by Dæmon”_.

I made a simple mockup, reusing the “Unvanquished Development” icon and a random font:

[![Powered by Dæmon](https://dl.illwieckz.net/b/daemon/features/splash/20201004-055144-000.unvanquished-empty-window-engine-logo-mockup.png)](https://dl.illwieckz.net/b/daemon/features/splash/20201004-055144-000.unvanquished-empty-window-engine-logo-mockup.png)

[Edit: as showcased by this mockup, a monochrome logo would be perfect]

illwieckz:
One idea that live rent-free in my mind for a long time now, would be for the logo to be a monochrome silhouette of a rocket engine, the word “dæmon” would be written on it as the engine name or the engine brand. The same way there is a “raptor engine” in real life, that would be the “dæmon engine” in some lore (yeah, the engine can have its own lore 🤣️.

Some inspirations:

1. Porco Rosso and the “Ghibli engine” for the idea of an engine with a name on it (here a plane engine):

![porco-rosso-ghibli-engine-0004](https://github.com/user-attachments/assets/f544257a-cafd-481c-988d-f04a7b2f11f4)

The id Tech 5 logo (for the kind of monochrome outline logo):

2. [![Id_Tech_5_Logo svg](https://github.com/user-attachments/assets/272876b5-0387-415f-b1ba-5a269d9b24dc)](https://fr.wikipedia.org/wiki/Fichier:Id_Tech_5_Logo.svg)

So, instead of a plane engine and instead of a light bulb, it would be some cool rocket engine. This can even be some avant-garde rocket engine (aerospike or something else) or sci-fi design, as long as it is recognized as a rocket engine.

[...]

PR #1436, opened 2024-11-12
renderer: split generic and generic3D

illwieckz:
Split `gl_genericShader` and `gl_generic3DShader`.

- Dedicate a simple shader without extra uniforms and no permutations for simple things like debug tools.
- Make sure to not have to call this boiler plate everywhere:
```c
gl_genericShader->SetVertexSkinning( false );
gl_genericShader->SetVertexAnimation( false );
gl_genericShader->SetTCGenEnvironment( false );
gl_genericShader->SetTCGenLightmap( false );
gl_genericShader->SetDepthFade( false );
// should be done there: gl_genericShader->BindProgram( 0 );
gl_genericShader->SetUniform_AlphaTest( GLS_ATEST_NONE );
```
- Leverage more of the compiler type checking at build time.

Here are the current usages:

- `gl_genericShader`: Mostly used by debug tools and things only requiring a very simple shader.
- `gl_generic2DShader`: Used for 2D UI elements, it has special code for RmlUi transforms.
- `gl_generic3DShader`: Full-featured shader for stages with a single texture.
- `gl_lightMapping`: Full-featured shader for multi-textured stages.

The `generic2D` and `generic3D` names are temporary names, we can rename them later to something more meaningful, as 2D and 3D can be confusing with 2D and 3D textures but they both receive 2D textures, we may rename them like `genericUI` and `generic<SOMETHING>`. The `Render_generic2D`, `Render_generic3D` and the `Render_generic` wrapper (that only calls code using `gl_generic2DShader` and  `gl_generic3DShader`) may be renamed too in the future.

If it happens that by doing so when compiling `gl_genericShader` we recompile a shader that already exists as a permutation of `gl_generic3DShader`, we may improve caching to deduplicate them (this may even deduplicate some `gl_generic2DShader` too):

- https://github.com/DaemonEngine/Daemon/issues/1435

This split has nothing to do with light overbright, I first implemented it as part of my overbright efforts because such overbright effort was one of the use case where such changes having already be made before would have make things easier. The overbright effort just reminded me I wanted to do that for a long time, so I decided it was the right time to finally do it.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -43,9 +43,10 @@ ShaderKind shaderKind = ShaderKind::Unknown;
 
 // *INDENT-OFF*
 
-GLShader_generic2D                       *gl_generic2DShader = nullptr;
 GLShader_generic                         *gl_genericShader = nullptr;
-GLShader_genericMaterial                 *gl_genericShaderMaterial = nullptr;
+GLShader_generic2D                       *gl_generic2DShader = nullptr;
+GLShader_generic3D                       *gl_generic3DShader = nullptr;
+GLShader_generic3DMaterial               *gl_generic3DShaderMaterial = nullptr;
 GLShader_cull                            *gl_cullShader = nullptr;
 GLShader_depthReduction                  *gl_depthReductionShader = nullptr;
 GLShader_clearSurfaces                   *gl_clearSurfacesShader = nullptr;
@@ -2173,6 +2174,25 @@ void GLShader::WriteUniformsToBuffer( uint32_t* buffer ) {
 	}
 }
 
+GLShader_generic::GLShader_generic( GLShaderManager *manager ) :
+	GLShader( "generic", ATTR_POSITION | ATTR_TEXCOORD | ATTR_QTANGENT, manager ),
+	u_ColorMap( this ),
+	u_DepthMap( this ),
+	u_TextureMatrix( this ),
+	u_ModelMatrix( this ),
+	u_ModelViewProjectionMatrix( this ),
+	u_ColorModulate( this ),
+	u_Color( this ),
+	GLDeformStage( this )
+{
+}
+
+void GLShader_generic::SetShaderProgramUniforms( shaderProgram_t *shaderProgram )
+{
+	glUniform1i( glGetUniformLocation( shaderProgram->program, "u_ColorMap" ), 0 );
+	glUniform1i( glGetUniformLocation( shaderProgram->program, "u_DepthMap" ), 1 );
+}
+
 GLShader_generic2D::GLShader_generic2D( GLShaderManager *manager ) :
 	GLShader( "generic2D", "generic", ATTR_POSITION | ATTR_TEXCOORD | ATTR_QTANGENT, manager ),
 	u_ColorMap( this ),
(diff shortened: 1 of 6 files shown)


illwieckz:
Also, there may be no need for any `gl_genericMaterialShader` variant.

slipher:
> gl_generic2DMaterialShader 

Should never need to exist.

We could get most of the benefit of boilerplate reduction by switching off GLSL macros en masse rather than one at a time, replacing
```
gl_genericShader->SetVertexSkinning( false );
gl_genericShader->SetVertexAnimation( false );
gl_genericShader->SetTCGenEnvironment( false );
gl_genericShader->SetTCGenLightmap( false );
gl_genericShader->SetDepthFade( false );
```
with
```
gl_genericShader->DelMacroBit(-1);
```

illwieckz:
Yes but there is still a one-line boilerplate to be copy-pasted every time `gl_genericShader` is used in those functions (currently 12 times), plus the uniform boilerplate to do 12 times too.

And a type can do it better.

VReaperV:
> * `gl_generic3DShader`: Full-featured shader for stages with a single texture.
> * `gl_lightMapping`: Full-featured shader for multi-textured stages.

I lighting here is the more important distinction (height maps, normal maps etc. also have to with lighting).

VReaperV:
> gl_generic2DMaterialShader

Yeah, I don't see the point in that either, UI wouldn't benefit from it, and generating materials for UI wuld require rewriting a lot of code.

src/engine/renderer/gl_shader.cpp
```diff
@@ -2173,6 +2174,27 @@ void GLShader::WriteUniformsToBuffer( uint32_t* buffer ) {
...
+	GLShader( "generic", ATTR_POSITION | ATTR_TEXCOORD | ATTR_QTANGENT, manager ),
+	u_ColorMap( this ),
+	u_DepthMap( this ),
+	u_TextureMatrix( this ),
+	u_ModelMatrix( this ),
+	u_ModelViewProjectionMatrix( this ),
+	u_ColorModulate( this ),
+	u_Color( this ),
+	u_ProfilerZero( this ),
+	u_ProfilerRenderSubGroups( this ),
```
VReaperV, line 2187: You don't really need these 2 uniforms for the non-3D shader. I guess you could technically use it with UI, but I don't see much benefit to it.

illwieckz:
> > gl_generic2DMaterialShader
> 
> Yeah, I don't see the point in that either, UI wouldn't benefit from it, and generating materials for UI wuld require rewriting a lot of code.

That part was just my memory being wrong. That doesn't change the fact there may be no need for a material variant of such stripped down generic shader, unlike the generic3D one, which is what I wanted to say.

src/engine/renderer/gl_shader.cpp
illwieckz: Done.

illwieckz:
I wonder if such simple shader would be reusable for displaying a startup logo as soon as possible:

- https://github.com/DaemonEngine/Daemon/issues/384

It would then be the very first shader being compiled, maybe not having permutations and being simple may help for having it built as soon as possible?

VReaperV:
> gl_generic2DShader: Used for 2D UI elements, it has special code for RmlUi transforms.

This will no longer be the case after #1439 is merged, so it can then be merged with the 3D shader. The `GENERIC_2D` will then only be used by the shader profiler.

slipher:
Since it turned out that 2D drawing doesn't actually need any special GLSL, I might make a very basic shader to be shared by 2D and debug drawing. But as a different GLSL source file please, not more macros on `generic`!

illwieckz:
The new `insert` mechanism may be usable to share possible common code if needed, the macro creep was basically a workaround for no `insert`.

slipher:
This shader would have less than 10 lines in its `main` functions so no need to share anything.

illwieckz:
Does 2D transformation require some vertex complex code?

VReaperV:
Nope, it's just `vertexSimple`.

illwieckz:
Nice!

VReaperV:
So once #1483 is merged, `gl_generic2DShader` can be removed from here (the 2D macro can be moved to `gl_genericShader`, not using it at all for `gl_generic3DShader`.

illwieckz:
Yes.

PR #1439, opened 2024-11-18
Remove useless depth-related stuff from 2D rendering

slipher:
Disable depth fade, depth testing, depth writing for 2D rendering. Don't create a depth or stencil channel in the default framebuffer.

Instead of using the "depth clamp" GL extension (which disables clipping based on the z-coordinate), set the projection matrix to make z always 0. This is relevant if we use 3d CSS transformations like [these](https://polypane.app/css-3d-transform-examples/).


VReaperV (approved):
LGTM.

PR #1441, opened 2024-11-18
Bloom fixes

### LINKED ISSUE #1438: Bloom has no effect
`r_bloom` no longer does anything. It worked in 0.54.1 and is broken in 0.55.0.

VReaperV:
- Removed mip-levels from contrast and bloom FBOs, other than the base one since that's the only one that we use.
- NUKED one of the blur shaders and moved everything into just one of them, since the difference was only the components of a vec2 being switched, which is better done with a uniform.
- Fixes #1438.


slipher:
It doesn't seem to do the same thing as in 0.54.0. Here's map "zandronum" `-175 -2529 -82 -97 3`. I tested this scene since bloom has a very large effect on it. For testing the PR I set `r_overbrightDefaultClamp 1` for fair comparison.

0.54.0:
unvanquished-zandronum-bloom-cage

This PR:
unvanquished-zandronum-bloom-cage


There is a lot of added brightness around the top left corner of the photograph that shouldn't be there.

While reviewing the first commit, I noticed that in the old code a "contrast" shader is the first one executed, while the commit changes it so that a "blur" shader is first.

VReaperV:
It's still executed first, at lines #3075-3090. Both use the blur shader(s) after the contrast one.

> There is a lot of added brightness around the top left corner of the photograph that shouldn't be there.

Could it be due to overbright and/or framebuffer format changes?

slipher:
> It's still executed first, at lines #3075-3090. Both use the blur shader(s) after the contrast one.

I see.


> Could it be due to framebuffer format changes?

I tried reverting `483d83d` and setting `r_highPrecisionRendering 0` and there wasn't any difference.

VReaperV:
Hmm, that's weird. I'll look at what the issue is later. On my end on the same map + viewpos the effect is not as strong as on the 2nd screenshot, but still brighter than the 1st one.

VReaperV:
> It doesn't seem to do the same thing as in 0.54.0. Here's map "zandronum" `-175 -2529 -82 -97 3`. I tested this scene since bloom has a very large effect on it. For testing the PR I set `r_overbrightDefaultClamp 1` for fair comparison.
> 
> 0.54.0: unvanquished-zandronum-bloom-cage
> 
> This PR: unvanquished-zandronum-bloom-cage
> 
> There is a lot of added brightness around the top left corner of the photograph that shouldn't be there.
> 
> While reviewing the first commit, I noticed that in the old code a "contrast" shader is the first one executed, while the commit changes it so that a "blur" shader is first.

Fixed, both look the same now:
0.54.1:
unvanquished_2024-11-24_131644_000
This pr:
unvanquished_2024-11-24_131543_000

slipher:
LGTM

PR #1442, opened 2024-11-18
Fix stage dynamics computation with material system

### LINKED ISSUE #1418: Broken pulsating shader on map station12
I remember it working on 0.54, the walls here should be pulsating with different colours:
unvanquished_2024-11-01_015947_000

Apparently the same should be true for some wall or something outside the alien base.

VReaperV:
Fixes #1418.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -54,6 +54,7 @@ MaterialSystem materialSystem;
 
 static void ComputeDynamics( shaderStage_t* pStage ) {
 	// TODO: Move color and texMatrices stuff to a compute shader
+	pStage->colorDynamic = false;
 	switch ( pStage->rgbGen ) {
 		case colorGen_t::CGEN_IDENTITY:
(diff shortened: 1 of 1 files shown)


slipher:
In my renderer test suite this fixed light styles on Metro, but caused a regression in the outdoor area on Pulse. Everything became more orange.

unvanquished-pulse-windowglass


unvanquished-pulse-windowglass

VReaperV:
> In my renderer test suite this fixed light styles on Metro, but caused a regression in the outdoor area on Pulse. Everything became more orange.
> 
> unvanquished-pulse-windowglass
> 
> unvanquished-pulse-windowglass

It seems to look the same as the second screenshot with `r_materialSystem off`.

slipher:
My bad, I accidentally made the screenshots with 2 different versions of the map. I reran the tests and it eliminates some diffs (vs. non-material renderer) on Metro and Pulse, without causing any new ones.

LGTM

PR #1443, opened 2024-11-18
Fix specular with material system

### LINKED ISSUE #1440: Material system: non-reflective specular mapping has no effect
If you take Daemon default settings + `r_materialSystem 1` + `r_specularMapping 0`, you get the same results as with `r_materialSystem 1` + `r_specularMapping 1`.

Reflective specular does do something.

VReaperV:
Fixes #1440.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -1031,7 +1031,7 @@ void BindShaderLightMapping( Material* material ) {
 	gl_lightMappingShaderMaterial->SetReliefMapping( material->enableReliefMapping );
 	/* Reflective specular setting is different here than in ProcessMaterialLightMapping(),
 	because we don't have cubemaps built yet at this point, but for the purposes of the material ordering there's no difference */
-	gl_lightMappingShaderMaterial->SetReflectiveSpecular( material->enableSpecularMapping && !( tr.refdef.rdflags & RDF_NOCUBEMAP ) );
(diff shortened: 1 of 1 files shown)


slipher:
LGTM

src/engine/renderer/Material.cpp
```diff
@@ -1031,7 +1031,7 @@ void BindShaderLightMapping( Material* material ) {
 	gl_lightMappingShaderMaterial->SetReliefMapping( material->enableReliefMapping );
 	/* Reflective specular setting is different here than in ProcessMaterialLightMapping(),
 	because we don't have cubemaps built yet at this point, but for the purposes of the material ordering there's no difference */
-	gl_lightMappingShaderMaterial->SetReflectiveSpecular( material->enableSpecularMapping && !( tr.refdef.rdflags & RDF_NOCUBEMAP ) );
+	gl_lightMappingShaderMaterial->SetReflectiveSpecular( glConfig2.reflectionMapping && material->enableSpecularMapping && !( tr.refdef.rdflags & RDF_NOCUBEMAP ) );
```
illwieckz, line 1034: Maybe there is a way to test for `glConfig2.reflectionMapping` when setting `material->enableSpecularMapping` earlier? Anyway, this is not blocking.

slipher: You can use specular mapping while reflection mapping is disabled. That's the broken case.

illwieckz: Ah I see, for some reasons my mind misread `glConfig2.specularMapping` instead. 😅️

Issue #1199, opened 2024-06-21
Reading from tess.verts


slipher:
Some renderer functions attempt to read back a surface's polygons from `tess.verts`. There are 3 possibilities, none of them good:
- The surface uses a static VBO with multidraw. `tess.numVerts` will be 0: there is nothing for it to read.
- The surface uses a static VBO with a single data range. `tess.numVerts` will be nonzero so the function will think there is data there. But really it is random data left over from whichever surfaces used the buffer last.
- The surface does not use a static VBO; it actually writes data to `tess.verts`. The function may at least work correctly in this case. But reading data from this GPU-mapped memory may cause poor performance as in #849.

I haven't attempted to search for all functions like this, but happened to notice the following ones:
- <s>SurfIsOffscreen</s> PortalOffScreenOrOutOfRange
- AutospriteDeform
- Autosprite2Deform

Note that some "autosprite" code is already known to be broken (#730). We might detect more suspicious functions by mapping the buffer as write-only, or setting it to null when a static VBO is active.

slipher:
Another culprit: `R_GetPortalOrientations`. That one also has a more serious problem: #1216.

slipher:
For historical interest, some previous examples were fixed in f870d9f392f512ad908a71156f9708944ceabaff :)

slipher:
#1444 only fixes the portal case, not the autosprite case.

VReaperV:
> #1444 only fixes the portal case, not the autosprite case.

Doesn't autosprite only write to it rather than read from it?

slipher:
Autosprite is read-modify-write.

VReaperV:
I thought it only reads from tess.vertsBuffer, which is supposedly CPU-side?

slipher:
Oh right, I changed it to use the CPU buffer so that performance problem should no longer happen. It's kinda hacky but good enough to close this I guess.

PR #1444, opened 2024-11-18
Fix portals with r_smp 1

### LINKED ISSUE #1216: r_smp is broken with portals
R_GetPortalOrientations is a render frontend function, but it writes into the tess.verts buffer which is supposed to be used by the renderer backend. Sure enough, if I launch daemon with args `-set r_smp 1 -set common.framerate.max -1 +devmap test_portals`, it immediately crashes when the map loads.

Note: it has always been like this; the problem is unrelated to recent portal changes.

Assuming there is no quick fix, I suggest making the USE_SMP build option disabled by default.

VReaperV:
Save portal AABBs when loading the map, then use them to cull portals instead of using the vertex buffer. I've considered using a rect instead since portals only really work on flat surfaces, but it wasn't important enough for me to bother given that the surface can be rotated. It might also be worth just using `R_CullBox()` instead of the projection into window coordinates and clipping.

Also fixes the issue mentioned in a comment that said that portal distance should be between the portal surface and view origin, instead of between the closest vertex and view origin.

This also improves performance since the vertex buffer is no longer read from on the CPU.

Fixes #1216 and #1199.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -2929,6 +2929,7 @@ static void R_CreateWorldVBO()
 	numVerts = 0;
 	numTriangles = 0;
 	numSurfaces = 0;
+	int numPortals = 0;
 
 	for ( k = 0; k < s_worldData.numSurfaces; k++ )
 	{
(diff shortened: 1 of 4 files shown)


illwieckz (approved):
LGTM.

PR #1446, opened 2024-11-19
FreeType submodule

illwieckz:
It is written against the Saigo branch:

- https://github.com/DaemonEngine/Daemon/pull/981

There is no submodule yet, one has to clone:

- `https://gitlab.freedesktop.org/freetype/freetype.git` as `libs/freetype`

When `USE_SYSTEM_LIBS` is set to `ON`, system libraries are used instead of submodules.

See also:

- https://github.com/Unvanquished/Unvanquished/pull/3195

See this issue for the motivation behind it:

- https://github.com/Unvanquished/Unvanquished/issues/3189


illwieckz:
If there is no comment about mistakes to fix, I will merge it in two days.

See this comment  for the reasons why I want and need FreeType to be a submodule:

- https://github.com/Unvanquished/Unvanquished/issues/3189#issuecomment-2550068802

slipher:
I see that for now, Freetype in the engine would still use the external_deps or system package by default. But what is the future of Freetype in external_deps? Is the plan to remove that and make everyone build it from scratch later?

illwieckz:
> I see that for now, Freetype in the engine would still use the external_deps or system package by default. But what is the future of Freetype in external_deps? Is the plan to remove that and make everyone build it from scratch later?

The plan is to never drop the ability to build against system library. Being able to not rebuild Freetype when the system provides it a wanted feature. This is what classic Linux distributions want. If Debian does an `unvanquished.deb`, they want it to rely on `freetype.deb`.

CMakeLists.txt
```diff
@@ -791,6 +791,8 @@ if (USE_BREAKPAD)
     endif()
 endif()
 
+option(USE_SYSTEM_LIBS "Tries to use system libs where possible." ON)
```
slipher, line 787: This is not a good name since it really means to use any kind of precompiled one. How about inverting the meaning and calling it USE_SOURCE_LIBS? Meaning prefer to build stuff from source.

.gitmodules
```diff
@@ -4,6 +4,9 @@
 [submodule "libs/crunch"]
 	path = libs/crunch
 	url = https://github.com/DaemonEngine/crunch.git
+[submodule "libs/freetype"]
+	path = libs/freetype
```
slipher, line 8: Thoughts on committing the source code directly instead of a submodule?

We could have done that with Googletest too...

freetype.cmake
```diff
@@ -0,0 +1,18 @@
+set(FT_DISABLE_BROTLI ON CACHE STRING "Disable Brotli" FORCE)
```
slipher, line 1: If you are adding a new cmake file it should go in `cmake/`. This is small enough that personally I would just put it in the main file.

cmake/DaemonFlags.cmake
```diff
@@ -208,9 +208,11 @@ else()
 	if (USE_RECOMMENDED_C_STANDARD)
 		# GNU89 or later standard is required when building gzip or the compiler
 		# will complain about implicitly defined lseek, read, write and close.
-		try_c_flag(GNU89 "-std=gnu89")
+		# GNU99 or later standard is required when building lua or lua will
+		# complain that the compiler doesn't support 'long long'.
+		try_c_flag(GNU89 "-std=gnu99")
 		if (NOT FLAG_GNU89)
-			message(FATAL_ERROR "GNU89 or C99 not supported by compiler")
+			message(FATAL_ERROR "GNU99 is not supported by the compiler")
```
slipher, line 215: Test variable's name is now mismatched with the flag

illwieckz: Fixed.

freetype.cmake
illwieckz: Why not. I did it that way because I mimicked how `rmlui.cmake` is stored in Unvanquished repository, but the Unvanquished repository has no `cmake/` folder (maybe it's time to do one there too).

CMakeLists.txt
illwieckz: Reverting it may not be a good idea if I change this to some `PREFER_SYSTEM_LIBS` with a fallback mechanism.

We can name it `USE_EXTERNAL_LIBS`, that name will fit both the system libraries and the ones from the `external_deps` folder.

illwieckz: I implemented `PREFER_EXTERNAL_LIBS` instead.

illwieckz:
I implemented `PREFER_EXTERNAL_LIBS`.

.gitmodules
illwieckz: I prefer submodules, committing may be good for things like some dead single-file include or so, and even in such situation it depends… Submodules make it easier to check for updates, to sync with upstream, and to commit that change.

freetype.cmake
illwieckz: I will keep it that way for now, to be consistent with all other cmake files in both Unvanquished and Daemon repositories for building sources, like `src.cmake`, `srclibs.cmake` (and the `lua.cmake` and `rmlui.cmake` ones in Unvanquished).

We can define a new convention later and migrate those to a better name later, but for now this is consistent.

freetype.cmake
```diff
@@ -0,0 +1,18 @@
...
+set(FT_DISABLE_BZIP2 ON CACHE STRING "Disable bzip2" FORCE)
+set(FT_DISABLE_HARFBUZZ ON CACHE STRING "Disable HarfBuzz" FORCE)
```
slipher, line 3: The subdirectory stuff showing up in the option list is always annoying. Let's put mark_as_advanced for all those

CMakeLists.txt
```diff
@@ -791,6 +791,22 @@ if (USE_BREAKPAD)
...
 endif()
 
+option(PREFER_EXTERNAL_LIBS "Tries to use system libs where possible." ON)
+
+macro(prefer_package LIB_NAME LIB_CMAKE)
+    if (PREFER_EXTERNAL_LIBS AND NOT NACL)
+        find_package(${LIB_NAME})
+
+        if (NOT ${LIB_NAME}_FOUND)
+            message(WARNING "PREFER_EXTERNAL_LIBS is enabled but external ${LIB_NAME} is not found, falling back to vendored FreeType.")
```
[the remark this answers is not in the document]

freetype.cmake
illwieckz: Done. I have also marked as advanced another one we don't modify.

While I was at it, I modified those lines as `option()`, and we don't force them anymore, we just set the default value we want, and this takes precedence over FreeTypes's own defaults, but still allow us to tweak them if we want.

The only forced FreeType option is `FT_DISABLE_ZLIB` (option to use builtin FreeType zlib) because it is always ON with NaCl and we already rely on `PREFER_EXTERNAL_LIBS` instead when building the engine.

slipher:
LGTM. Yay!

PR #1449, opened 2024-11-23
VirtualMachine: make receive timeout configurable

illwieckz:
Make receive timeout configurable.

I used this to make sure the Saigo-compiled code was stuck in an infinite loop and not just slower.

It may also help when loading the game from a very slow file system (like network file system over Internet), because the time waiting for the first receive is waiting for the nexe to be actually loaded (downloaded if a network file system).

### DIFF
--- a/src/engine/framework/VirtualMachine.cpp
+++ b/src/engine/framework/VirtualMachine.cpp
@@ -83,6 +83,11 @@ static Cvar::Cvar<bool> vm_nacl_bootstrap(
 	"Use NaCl bootstrap helper",
 	Cvar::INIT, true);
 
+static Cvar::Cvar<int> vm_timeout(
+	"vm.timeout",
+	"Receive timeout in seconds",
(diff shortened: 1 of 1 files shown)


src/engine/framework/VirtualMachine.cpp
```diff
@@ -83,6 +83,11 @@ static Cvar::Cvar<bool> vm_nacl_bootstrap(
 	"Use NaCl bootstrap helper",
 	Cvar::INIT, true);
 
+static Cvar::Cvar<int> vm_timeout(
+	"vm.timeout",
+	"Receive timeout in seconds",
+	Cvar::INIT, 2);
```
slipher, line 89: Why INIT?

illwieckz: Lazy copy-paste. 🤪️

Fixed.

slipher:
LGTM

PR #1450, opened 2024-11-26
Add trap_GetPings, so that engine does not mutate playerstate

slipher:
Have sgame retrieve pings by using a syscall, instead of the engine writing the ping into shared memory.

This is an attempt to simplify the interface between the engine and the VM, which could benefit future sandboxing technology migrations. The player state portion of the shared memory will now be readonly for the engine, rather than a two-way communication channel. Hopefully I can do the same for the entity state next, so that all of the shared memory is read-only for the engine.

Companion: https://github.com/Unvanquished/Unvanquished/pull/3207

### DIFF
--- a/src/engine/qcommon/msg.cpp
+++ b/src/engine/qcommon/msg.cpp
@@ -1336,10 +1336,10 @@ MSG_WriteDeltaPlayerstate
 
 =============
 */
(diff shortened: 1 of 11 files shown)


illwieckz (approved):
LGTM.

Issue #1312, opened 2024-09-17
Model bug with latest and Clang 19


illwieckz:
I remember @DolceTriade said one day he seen bugs while compiling with GCC 14.

I confirm I see bugs affecting models in a weird way when compiling with GCC 14 (no issue with GCC 13):

![unvanquished_2024-09-13_181819_000](https://github.com/user-attachments/assets/045c0ea7-99dc-4763-8224-0f59efb50cb1)

![unvanquished_2024-09-13_181834_000](https://github.com/user-attachments/assets/c7a5e571-4981-4796-bc26-5360549cca69)

First stable version of Clang 19 was released today and then out of curiosity I test it, and I got another bug but affecting the models too, but in a different way (no issue with Clang 18):

![unvanquished_2024-09-17_200312_000](https://github.com/user-attachments/assets/f8fc4c2a-a913-401c-abd3-fcf8e0abfbee)

![unvanquished_2024-09-17_200930_000](https://github.com/user-attachments/assets/c24be8ca-2251-434e-a580-1b9566db9c0e)

Maybe we actually do something wrong if two compilers fail on it.

illwieckz:
An easy way to test both GCC 14 and Clang 19 is to use Ubuntu Noble (it ships GCC 14) and to install Clang 19 from the LLVM upstream repository for Ubuntu:  https://apt.llvm.org

slipher:
As usual this doesn't happen with GCC 14.2 on Windows.

Does it matter what `vm.cgame.type` you use? Does it matter if you disable `r_vboModels`?

illwieckz:
I get the same results with both DLL and EXE cgame.

I get the same results with `r_vboVertexSkinning 0` and `r_vboModel 0` (actually `r_vboVertexSkinning 0` has to be preferred over `r_vboModel 0`).

DolceTriade:
Removing `-ffast-math` fixes issue on GCC14 for me.

illwieckz:
With current Saigo (LLVM 19):

![unvanquished_2024-11-19_234514_000](https://github.com/user-attachments/assets/fa26522e-34d0-421a-bcce-361b0a9074de)

![unvanquished_2024-11-19_234532_000](https://github.com/user-attachments/assets/ea6bb39f-5ce9-42dd-887a-b961ef167c76)

So that looks to be a game code bug.

illwieckz:
[...]

PR #1452, opened 2024-12-02
Fix Clang model bug, improve SSE intrinsics code and add tests for q_math transform functions

slipher:


illwieckz:
I confirm it fixes the bugs with Clang 19 from:

But it doesn't fix the bugs with GCC 14.

illwieckz:
I confirm that I get no bug when combining:

- an engine built with either Clang 19.1.5 or GCC 14.2.0,
- a nexe game built with either PNaCl 3.6.0 or Saigo 19.0.0.

DolceTriade (approved):
if it fixes bug, then lgtm

Issue #1454, opened 2024-12-04
Using PDCursesMod on all systems?


illwieckz:
We have a `pdcursesmod` submodule for Windows, so this code is always there.

Should we be use it instead of ncurses on Linux and macOS?

Not only we have pdcurses always available in our sourcetree, but building or linking ncurses isn't that straightforward.

For example in the flatpak we disable ncurses because of various errors, and I myself faced many errors in various situations dealing with ncurses for building the engine.

PDCursesMod provides VT support:

- https://github.com/Bill-Gray/PDCursesMod/blob/master/vt/README.md

So I guess everything is there for Linux and macOS.

slipher:
That's OK with me. We should check with Linux users that the quality of implementation lives up to what they get from ncurses. The Windows integration leaves some things to be desired: there is no way to resize the window, scrolling with mouse doesn't work (only page up/down), and the number of colors is small.

illwieckz:
> The Windows integration leaves some things to be desired: there is no way to resize the window, scrolling with mouse doesn't work (only page up/down), and the number of colors is small.

Ah, that is annoying!

If ncurses is much better, maybe we can look at an option to force the uses on pdcurses on Linux system failing to build with ncurses, but use ncurses otherwise.

Distro packages like Arch ones would use system's ncurses dynamic library, while some other packages may use pdcurses.

illwieckz:
Quote from IRC:

```
<illwieckz> perturbed`, I see USE_CURSES is always ON on Windows, what happens if it is forced to OFF?
<perturbed`> probably it doesn't compile
<perturbed`> the alternative code is based on non-curses unix terminal functions
```

I also noticed that `USE_CURSES` was always disabled on macOS with this comment:

> Not supported on mac because the included version is too old

So using PDCusrsesMod on macOS would make possible to uses `-curses` on macOS.

illwieckz:
PDCursesMod for VT has a `resize_term()` function that seem to work like ncurses `resizeterm()` one.

I can resize and scroll, but the colors are less good.

So, using PDCursesMod looks to be a good fallback.

illwieckz:
@slipher I noticed the call to `resize_term` is commented out on Windows in our code, what if we enable it?

The function is named `resize_term()` on PDCursesMod while it is named `resizeterm()` on ncurses.

illwieckz:
Hmm, maybe what we miss on Windows is a code to query the new size.

illwieckz:
[...]

PR #1455, opened 2024-12-04
Make possible to use PDCursesMod on every system

illwieckz:
- PDCursesMod is forced on Windows as `USE_CURSES` is always enabled since there is no other console code.
- PDCursesMod is used on macOS if `USE_CURSES` is enabled, it couldn't be enabled before as system ncurses isn't supported.
- PDCursesMod is used on Linux if `USE_CURSES` is enabled, but `USE_CURSES_NCURSES` is disabled.
- ncurses is used on Linux if both `USE_CURSES` and `USE_CURSES_NCURSES` are enabled.

See:

- https://github.com/DaemonEngine/Daemon/issues/1454


slipher:
I tested on Mac and the cursor was broken. But it was easy to guess the culprit: some `#ifdef _WIN32`s should be *if using pdcurses*. I pushed a fixup commit for that.

LGTM with the fixup

illwieckz:
I confirm the PDCursesMod builds properly in the dockerized build script (and runs properly on macOS).

PR #1456, opened 2024-12-06
cmake: only use _FORTIFY_SOURCE with hardening

illwieckz:
That test was almost always true (well, except with non-release MinGW builds no one run).

The previous code means that `_FORTIFY_SOURCE`  was used on MinGW on recent releases because `USE_HARDENING` is used.

I don't really know what means that test anyway.

I noticed this because doing ASAN build on macOS spammed lots of complain that `_FORTIFY_SOURCE`. It looks like an ASAN build on macOS forces a different `_FORTIFY_SOURCE` value so we need to be able to disable it, something we can do by disabling hardening.


slipher:
This makes most builds less secure just to fix warnings on one obscure use case.

PR #1460, opened 2024-12-12
Unit test for Q_rsqrt

slipher:
Add a unit test so we can verify that magic square root algorithms meet the claimed tolerances. If #1458 adds a second version of the function, we can copy and paste the test and just change the `relativeTolerance`.


illwieckz (approved):
LGTM.

illwieckz:
Or the magic one, but not the SSE one, I guess the relative error bound may differ.

PR #1461, opened 2024-12-13
Don't set -pie flag with AppleClang

slipher:
Apparently it doesn't do anything, and triggers a warning about an unused command line argument. When USE_HARDENING is enabled, this warning breaks a bunch of other try_compile's, causing flags to be rejected that should have been accepted.


slipher:
Added another commit removing another flag which causes unused argument warning spam (`-stdlib=c++`).

illwieckz:
It works.

PR #1462, opened 2024-12-14
Networking bug fixes and cleanup

slipher:
Note: nothing release-critical here

### DIFF
--- a/src/engine/qcommon/net_ip.cpp
+++ b/src/engine/qcommon/net_ip.cpp
@@ -1630,7 +1630,7 @@ static void NET_OpenIP()
 
 	if ( net_enabled->integer & NET_ENABLEV4 )
 	{
-		for ( i = ( port6 == PORT_ANY ? 1 : MAX_TRY_PORTS ); i; i-- )
+		for ( i = ( port == PORT_ANY ? 1 : MAX_TRY_PORTS ); i; i-- )
 		{
 			ip_socket = NET_IPSocket( net_ip->string, port, &boundto4, &err );
 
--- a/src/engine/qcommon/net_ip.cpp
+++ b/src/engine/qcommon/net_ip.cpp
@@ -935,7 +935,13 @@ SOCKET NET_IPSocket( const char *net_interface, int port, struct sockaddr_in *bi
 
 	Log::Notice( "Opening IP socket: %s:%s", net_interface ? net_interface : "0.0.0.0", port ? va( "%i", port ) : "*" );
 
-	if ( ( newsocket = socket( PF_INET, SOCK_DGRAM, IPPROTO_UDP ) ) == INVALID_SOCKET )
+#ifdef _WIN32
+	newsocket = WSASocketW( PF_INET, SOCK_DGRAM, IPPROTO_UDP, nullptr, 0, WSA_FLAG_NO_HANDLE_INHERIT );
+#else
+	newsocket = socket( PF_INET, SOCK_DGRAM, IPPROTO_UDP );
+#endif
+
+	if ( newsocket == INVALID_SOCKET )
 	{
 		*err = socketError;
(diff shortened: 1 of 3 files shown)


slipher:
I will merge this in 3 days if there are no objections.

src/engine/qcommon/net_ip.cpp
```diff
@@ -1020,7 +1022,13 @@ SOCKET NET_IP6Socket( const char *net_interface, int port, struct sockaddr_in6 *
 	// Print the name in brackets if there is a colon:
 	Log::Notice( "Opening IP6 socket: %s%s%s:%s", brackets ? "[" : "", net_interface ? net_interface : "[::]", brackets ? "]" : "", port ? va( "%i", port ) : "*" );
 
-	if ( ( newsocket = socket( PF_INET6, SOCK_DGRAM, IPPROTO_UDP ) ) == INVALID_SOCKET )
+#ifdef _WIN32
+	newsocket = WSASocketW( PF_INET6, SOCK_DGRAM, IPPROTO_UDP, nullptr, 0, WSA_FLAG_NO_HANDLE_INHERIT );
```
DolceTriade, line 1026: Why is this better than socket?

src/engine/qcommon/net_ip.cpp
```diff
@@ -935,7 +935,13 @@ SOCKET NET_IPSocket( const char *net_interface, int port, struct sockaddr_in *bi
 
 	Log::Notice( "Opening IP socket: %s:%s", net_interface ? net_interface : "0.0.0.0", port ? va( "%i", port ) : "*" );
 
-	if ( ( newsocket = socket( PF_INET, SOCK_DGRAM, IPPROTO_UDP ) ) == INVALID_SOCKET )
+#ifdef _WIN32
+	newsocket = WSASocketW( PF_INET, SOCK_DGRAM, IPPROTO_UDP, nullptr, 0, WSA_FLAG_NO_HANDLE_INHERIT );
```
DolceTriade, line 935: Should we use SO_REUSEPORT/SO_REUSEADDR instead? That should be consistent across all OSs (well SO_REUSEPORT on Linux and SO_REUSEADDR on others)...

src/engine/qcommon/net_ip.cpp
```diff
@@ -1948,8 +1877,12 @@ NET_Restart_f
 */
 void NET_Restart_f()
 {
-	NET_Config( false );
+	NET_DisableNetworking();
 	SV_NET_Config();
 	Net::ShutDownDNS();
-	NET_Config( true );
```
DolceTriade, line 1936: So this was a bug?

src/engine/server/sv_init.cpp
```diff
@@ -756,7 +756,7 @@ void SV_Shutdown( const char *finalmsg )
 
 	Cvar_Set( "sv_running", "0" );
 #ifndef BUILD_SERVER
-	NET_Config( true );
+	NET_EnableNetworking( false );
```
DolceTriade, line 759: This probably deserves a comment? Why was this changed?

slipher: It's not really doing something different. The diff may be confusing because the boolean argument on the left side is `networkingEnabled` and on the right side it is `serverMode` (the responsibility of deciding serverMode was put on the caller).

The `ifndef` stuff is indeed quite confusing, so I added a comment. Thanks for the suggestion.

src/engine/qcommon/net_ip.cpp
slipher: It's not really doing something different. The diff may be confusing because the boolean argument on the left side is `networkingEnabled` and on the right side it is `serverMode` (the responsibility of deciding serverMode was put on the caller).

DolceTriade (approved):

PR #1464, opened 2024-12-15
Make NACL_VMS_PROJECTS accessible to outer scope

slipher:
Also remove duplicative nacl-vms-pexe target.

### DIFF
--- a/cmake/DaemonGame.cmake
+++ b/cmake/DaemonGame.cmake
@@ -77,9 +77,8 @@ function(GAMEMODULE)
                     "-D${inherited_option}=${${inherited_option}}")
             endforeach(inherited_option)
 
-            add_custom_target(nacl-vms ALL)
-
             if (USE_NACL_SAIGO)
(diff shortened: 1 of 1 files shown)


illwieckz:
With that patch re-adding the separate generic `nacl-vms` target and toolchain-specific `nacl-vms-pexe` target, LGTM.

illwieckz:
And actually I like the fact the `nacl-vms` target just exists to depend on the toolchain-specific targets anyway.

slipher:
> Not removing the duplicative nacl-vms-pexe target fixes [Unvanquished/Unvanquished#3245 (comment)](https://github.com/Unvanquished/Unvanquished/pull/3245#issuecomment-2546433433)

Fixed. The Saigo-specific line creating a dummy target was mistakenly not in its `if` block.

illwieckz:
It works, you can merge it, but I will add a generic `vms` target that is independent to the underlying format and toolchain anyway.

PR #1465, opened 2024-12-16
Add the `contact` cvar and server info string to contain contact information to reach out the server admin when needed

illwieckz:
- sv_init: add the `contact` cvar and server info string to contain contact information to reach out the server admin when needed
- common: disable logging suppression in `Info_Print()` (used by commands like `/systeminfo` and `/serverinfo`)

### DIFF
--- a/src/engine/server/sv_init.cpp
+++ b/src/engine/server/sv_init.cpp
@@ -52,6 +52,9 @@ static Cvar::Cvar<std::string> cvar_pakname(
 	"pakname", "pak containing current map", Cvar::SERVERINFO | Cvar::ROM, "");
 static Cvar::Cvar<std::string> sv_paks(
 	"sv_paks", "currently loaded paks", Cvar::SYSTEMINFO | Cvar::ROM, "");
+static Cvar::Cvar<std::string> cvar_contact(
+	"contact", "contact information to reach out to the server admin when needed: forum or chat nickname, mail address…",
+	Cvar::SERVERINFO, "" );
 static Cvar::Cvar<bool> sv_useBaseline(
 	"sv_useBaseline", "send entity baseline for non-snapshot delta compression", Cvar::NONE, true);
 
@@ -437,6 +440,14 @@ void SV_SpawnServer(std::string pakname, std::string mapname)
 	SV_ShutdownGameProgs();
 
 	PrintBanner( "Server Initialization" )
+
+	if ( !SV_Private(ServerPrivate::NoAdvertise)
+		&& sv_networkScope.Get() >= 2
+		&& cvar_contact.Get().empty() )
+	{
+		Log::Warn( "The contact information isn't set, it is requested for public servers." );
+	}
+
 	Log::Notice( "Map: %s", mapname );
 
 	// if not running a dedicated server CL_MapLoading will connect the client to the server
--- a/src/engine/qcommon/common.cpp
+++ b/src/engine/qcommon/common.cpp
@@ -124,11 +124,11 @@ void Info_Print( const char *s )
 			*o = 0;
 		}
 
-		Log::Notice( key );
+		Log::defaultLogger.WithoutSuppression().Notice( key );
 
 		if ( !*s )
 		{
-			Log::Notice( "MISSING VALUE" );
+			Log::defaultLogger.WithoutSuppression().Notice( "MISSING VALUE" );
 			return;
 		}
 
@@ -147,7 +147,7 @@ void Info_Print( const char *s )
 			s++;
 		}
 
-		Log::Notice( "%s", value );
+		Log::defaultLogger.WithoutSuppression().Notice( value );
 	}
 }
 


illwieckz:
The `abuse` system info string can be read (if set) inside the game with the `/serverinfo` command, and outside of the game with this command:

```
qstat -R -unvanquisheds <server>
```

The idea is that we will encourage public server owners to give some contact information, be it a chat or forum nick, or email address, or whatever we can use to reach them.

This could be usable to reach out server admins outside of the game for processing abusive behavior from players. It is why it is named `abuse`, like the `abuse` mail address convention.

This could also be usable to reach out server admin to tell them to update their servers.

Actually writing there the forum nick may allow us to integrate the web server list with the forum private message feature. At some point we may recommend this.

illwieckz:
At some point we may enforce this string to be set for starting public servers. People will still be free to put garbage there so giving information will still not be mandatory, but it would be mandatory for the public server owner to decide to not give contact information on purpose.

So people wanting to remain fully anonymous (even not disclosing their forum/chat nick name) but actually doing things right would not be bothered.

But if we need to contact a server owner because of abusive behavior not being processed _and_ we discover the server owner intentionally doesn't want to be reached out, we may decide to blacklist the server. So basically it would help to make the difference between neglect and intentional ill-will.

That's why I believe we may require that value to be set in the future for starting public servers, as making it a requirement:

- doesn't require good people to leave anonymity if the admin job is done right,
- would help making the difference between good will and bad will if the admin job isn't done right,
- it would encourage nice people to give contact information so it help us coordinating things in a better way.

What we can do is to have it optional (with the warning) for the next version and the next-next version would make it required.

src/engine/server/sv_init.cpp
```diff
@@ -437,6 +440,12 @@ void SV_SpawnServer(std::string pakname, std::string mapname)
 	SV_ShutdownGameProgs();
 
 	PrintBanner( "Server Initialization" )
+
+	if ( SV_Private( ServerPrivate::Public ) && cvar_abuse.Get().empty() )
```
slipher, line 444: The condition is inverted here. Also I suggest including sv_networkScope so that it matches the condition for advertising to the master server. So `!SV_Private(ServerPrivate::NoAdvertise) && sv_networkScope >= 2`

src/engine/server/sv_init.cpp
```diff
@@ -52,6 +52,9 @@ static Cvar::Cvar<std::string> cvar_pakname(
 	"pakname", "pak containing current map", Cvar::SERVERINFO | Cvar::ROM, "");
 static Cvar::Cvar<std::string> sv_paks(
 	"sv_paks", "currently loaded paks", Cvar::SYSTEMINFO | Cvar::ROM, "");
+static Cvar::Cvar<std::string> cvar_abuse(
```
slipher, line 55: Why call it "abuse" if that's not the main motivation for providing it? How about "contact" or "admin"

src/engine/server/sv_init.cpp
```diff
@@ -52,6 +52,9 @@ static Cvar::Cvar<std::string> cvar_pakname(
...
+	"abuse", "contact information to reach out the server admin when needed (forum or chat nickname, mail address…)",
```
VReaperV, line 56: `out the server admin` -> `out t the server admin`

src/engine/server/sv_init.cpp
illwieckz: This can be named `contact`, I first named it `admin` but then it sounded weird because there are other `admin*` commands and this would be confusing. Then my next example in mind was the `abuse@` mail address convention. But I'm fully open to other names, like `contact`.

src/engine/server/sv_init.cpp
illwieckz: Thanks, if some have better wording for the whole sentence, it's welcome too.

src/engine/server/sv_init.cpp
illwieckz: I renamed to `contact`.

src/engine/server/sv_init.cpp
illwieckz: I fixed the string.

src/engine/server/sv_init.cpp
slipher: This is still not fixed

illwieckz: Done.

slipher:
A commented-out example in `dist/configs/config/server.cfg` would be good.

LGTM

sweet235:
I do not like the idea of partitioning people into good and bad ones.

Gireen:
But its unavoidable. What matters is where the line is drawn and to preserve as much freedom as possible. 
When there is a DDOS attack it should be easy to judge, but "admin job isn't done right" sounds way to squishy.

sweet235:
> "admin job isn't done right" sounds way to squishy

It does.

> But if we need to contact a server owner because of abusive behavior not being processed and we discover the server owner intentionally doesn't want to be reached out, we may decide to blacklist the server.

We cannot simply blacklist random servers just because someone did not leave their email address and is on vacation.

PR #1466, opened 2024-12-19
Bug fix + cvar migration for r_textureMode

slipher:


src/engine/renderer/tr_init.cpp
```diff
@@ -830,8 +831,8 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 
 		tr.currenttextures.resize( glConfig2.maxTextureUnits );
 
-		GL_TextureMode( r_textureMode->string );
-		r_textureMode->modified = false;
+		GL_TextureMode( r_textureMode.Get().c_str() );
```
VReaperV, line 834: This looks like it can be just `GL_TextureMode( r_textureMode.GetModifiedValue().c_str() );` since it should always be `modified` at this point?

slipher: It will not always be modified. I believe this code runs on every map change.

src/common/CvarTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+    cv.GetModifiedValue();
+
+    // Make sure modified flag is set upon unlatching
+    Latch(cv);
+    Util::optional<float> modified = cv.GetModifiedValue();
+    ASSERT_TRUE(modified);
+    ASSERT_EQ(modified.value(), -5.0f);
+}
+
+} // namespace Math
```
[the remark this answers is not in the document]
slipher: Done.

src/engine/renderer/tr_init.cpp
VReaperV: I see.

VReaperV (approved):
LGTM

PR #1468, opened 2024-12-19
cmake: remove a warning about scope

illwieckz:
Remove this warning:

```
CMake Warning (dev) at CMakeLists.txt:136 (set):
  Cannot set "NACL_TARGETS": current scope has no parent.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at /usr/share/cmake-3.28/Modules/CMakeDependentOption.cmake:89 (message):
  Policy CMP0127 is not set: cmake_dependent_option() supports full Condition
  Syntax.  Run "cmake --help-policy CMP0127" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.
  CMakeLists.txt:167 (cmake_dependent_option)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

That lines look like a mistake, everything is either in the same scope, either passed through command line to the child cmake, there is nothing to bring up.

### DIFF
--- a/cmake/DaemonGame.cmake
+++ b/cmake/DaemonGame.cmake
@@ -32,11 +32,61 @@ include(DaemonFlags)
 # Function to setup all the Sgame/Cgame libraries
 include(CMakeParseArguments)
 
+## About the different ways to host/play games:
+## Native DLL: no sandboxing, no cleaning up but debugger support. Use for dev.
+## NaCl exe: sandboxing, no leaks, slightly slower, hard to debug. Use for regular players.
+## Native exe: no sandboxing, no leaks, hard to debug. Might be used by server owners for perf.
+## See VirtualMachine.h for code.
+
+# can be loaded by daemon with vm.[sc]game.type 3
+option(BUILD_GAME_NATIVE_DLL "Build the shared library files, mostly useful for debugging changes locally." ON)
+
+# can be loaded by daemon with vm.[sc]game.type 2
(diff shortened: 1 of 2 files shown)


slipher:
NACL_TARGETS is set in Daemon's CMakeLists, but consumed by DaemonGame.cmake which is executed in the context of Unvanquished's CMakeLists. So I expect this to be needed for the variable to show up (including daemon with `add_subdirectory` is what makes it have a separate scope). Otherwise we would have the same problem as with NACL_VMS_PROJECTS being empty. 

Probably the right thing to move the code seting NACL_TARGETS inside DaemonGame.cmake so we don't run this code for building gamelogic when only the engine is built.

illwieckz:
Ah, that may explain why I added that line back in the days.

Is the new patch better?

slipher:
I mean move the entire block of code that builds the target list in there.

illwieckz:
OK, now it may be correct.

cmake/DaemonGame.cmake
```diff
@@ -200,4 +250,6 @@ function(GAMEMODULE)
             endforeach()
         endif()
     endif()
+
+    set(NACL_TARGETS "${NACL_TARGETS}" PARENT_SCOPE)
```
slipher, line 254: Now NACL_TARGETS is only used within this function, right? So it is not needed in another scope.

illwieckz: Now removed.

DolceTriade (approved):

Issue #1448, opened 2024-11-22
Better material data packing for material system


VReaperV:
(I wrote this for #1414, but it's too lengthy and out of scope for that issue) 

Right now, the highest total amount of shader stages I've seen on a map is 1551, which I believe can be decreased further since some of them will only differ in the textures that they use. 

Currently, these are the non-global uniforms used by the material shaders: 
```
genericMaterial:

// 32 components (128 bytes)
struct Material {
  mat4 u_TextureMatrix;
  vec4 u_ColorModulate;
  vec4 u_Color;
  uvec2 u_ColorMap;
  uvec2 u_DepthMap
  float u_AlphaThreshold;
  float u_DepthScale;
  int material_padding0;
  int material_padding1;
};

lightMappingMaterial:
// 48 components (192 bytes)
struct Material {
  vec4 u_ColorModulate;
  vec4 u_Color;
  mat4 u_TextureMatrix;
  vec3 u_NormalScale;
  int u_NormalScale_padding;
  uvec2 u_DiffuseMap;
  uvec2 u_NormalMap;
  uvec2 u_HeightMap;
  uvec2 u_MaterialMap;
  uvec2 u_LightMap;
  uvec2 u_DeluxeMap;
  uvec2 u_GlowMap;
  vec2 u_SpecularExponent;
  float u_AlphaThreshold;
  float u_LightFactor;
  float u_ReliefDepthScale;
  float u_ReliefOffsetBias;
};

skyboxMaterial:
// 24 components (96 bytes)
struct Material {
  mat4 u_TextureMatrix;
  uvec2 u_ColorMapCube;
  uvec2 u_CloudMap;
  float u_CloudHeight;
  float u_AlphaThreshold;
  int material_padding0;
  int material_padding1;


};

fogQuake3Material:
// 8 components (32 bytes)
struct Material {
  vec4 u_Color;
  uvec2 u_FogMap;
  int material_padding0;
  int material_padding1;
};

heatHazeMaterial:
// 36 components (144 bytes)
struct Material {
  vec4 u_ColorModulate;
  vec4 u_Color;
  mat4 u_TextureMatrix;
  vec3 u_NormalScale;
  int u_NormalScale_padding;
  uvec2 u_CurrentMap;
  uvec2 u_NormalMap;
  uvec2 u_HeightMap;
  float u_DeformMagnitude;
  int material_padding0;

};

liquidMaterial:
// 38 components (152 bytes)
struct Material {
[...]

PR #1473, opened 2024-12-25
Improve material system data packing

VReaperV:
Partially implements #1448.

Stage data is allocated for stages rather than `surface * stage`, which greatly reduces the amount of required data. Texture handles and texture matrix are moved into a different buffer, since stage data ends up being often under 1kb this way. Lightmap/deluxemap handles are moved into a different UBO, because they depend on the `drawSurf` rather than on shader stage. All 3 are encoded into `baseInstance` (12 bits for shader stage, 12 bits for textures, and 8 more for lightmaps).

Optimised `u_Color` by changing it from a `vec4` to a `uint` (the colors we use only have 8-bit precision in engine).
Also added `u_ColorModulateColorGen` for use with shaders other than `cameraEffects`: these only use 12 distinct states, encoded as 5 bits, which also allows using a `uint` here instead of a `vec4`. Added `unpackUnorm4x8()` define for devices that don't support `ARB_gpu_shader5`.

Changed `u_TextureMatrix` from a `mat4` to `mat3x2`, reducing its size from 16 to just 6 components since only these 6 components have any actual effect, which allows making the `TexData` struct be exactly 64 bytes.

All of the above combined reduces the memory usage of material system by ~50-100x, and improves its performance and load times.

Also added `r_materialSystemSkip` cvar, which allows instantaneously switching between core and material renderer (as long as the latter is enabled in the first place), for much faster debugging and performance comparison.


slipher:
Is there any way to split this up into multiple PRs?

VReaperV:
I believe so, yes.

VReaperV:
> Is there any way to split this up into multiple PRs?

Parts of it are now in: #1475, #1476, #1477, and #1478. I think that's as far as I can split it (`u_TextureMatrix` change would end up with a bunch of merge conflicts here).

slipher:
Light styles are broken, e.g. map `metro-b1-2` at `setviewpos -1172 627 107`.

unvanquished_2025-01-10_192237_000

VReaperV:
Thought I fixed those, gonna need to recheck it then.

VReaperV:
> Light styles are broken, e.g. map `metro-b1-2` at `setviewpos -1172 627 107`.
> 
> unvanquished_2025-01-10_192237_000

The black spot is a shader caching issue, deleting the `glsl` folder should fix it.

VReaperV:
Nevermind, that does not fix it. It's an issue with dynamic lights though.

VReaperV:
Both issues fixed now.

src/engine/renderer/gl_shader.h
```diff
@@ -2703,17 +2745,26 @@ class u_ShadowClipMap4 :
...
+		GLUniformMatrix32f( shader, "u_TextureMatrix", true )
 	{
 	}
 
 	void SetUniform_TextureMatrix( const matrix_t m )
 	{
-		this->SetValue( GL_FALSE, m );
+		/* We only actually need these 6 components to get the correct texture transformation,
+		the other ones are unused */
+		static vec_t m2[6];
```
slipher, line 2760: Why static?

slipher (commented):
Tests look good. Fixes #1499 and doesn't alter anything else.

Currently I have reviewed all commits except the first one.

VReaperV:
That is very good, thanks.

src/engine/renderer/gl_shader.h
VReaperV: Oh, that might be a leftover from some code I added and removed earlier.

VReaperV: Removed it now.

src/engine/renderer/Material.h
```diff
@@ -143,6 +135,69 @@ struct Material {
...
+			}
+		}
+
+		return true;
+	}
+
+	TextureData() {
+	}
+
+	TextureData( const TextureData& other ) {
```
slipher, line 193: Could use the auto-generated copy constructor here (either by putting `= default` or removing it entirely)

src/engine/renderer/Material.cpp
```diff
@@ -642,125 +459,151 @@ void UpdateSurfaceDataFog( uint32_t* materials, Material& material, drawSurf_t*
...
 }
 
-// This generates the buffer GLIndirect commands
+void MaterialSystem::GenerateMaterialsBuffer( std::vector<shaderStage_t*>& stages, const uint32_t size, uint32_t* materialsData ) {
+	// Shader uniforms are set to 0 if they're not specified, so make sure we do that here too
+	memset( materialsData, 0, size * sizeof( uint32_t ) );
+	for ( shaderStage_t* pStage : stages ) {
+		Material& material = materialPacks[pStage->materialPackID].materials[pStage->materialID];
+
+		/* Stage variants are essentially copies of the same stage with slightly different values that
```
slipher, line 487: So what's the ultimate purpose of variants? If I trace e.g. `mayUseVertexOverbright`/`VERTEX_OVERBRIGHT`, I can't find any decisions ever being made based on the variant, other than segregating it from other variants.

src/engine/renderer/Material.cpp
```diff
@@ -1392,8 +1305,131 @@ void ProcessMaterialFog( Material* material, shaderStage_t* pStage, drawSurf_t*
...
+		dynamicStages.emplace_back( pStage );
+	}
+}
+
 void MaterialSystem::ProcessStage( drawSurf_t* drawSurf, shaderStage_t* pStage, shader_t* shader, uint32_t* packIDs, uint32_t& stage,
 	uint32_t& previousMaterialID ) {
+	lightMode_t lightMode;
+	deluxeMode_t deluxeMode;
+	SetLightDeluxeMode( drawSurf, pStage->type, lightMode, deluxeMode );
+	pStage->mayUseVertexOverbright = pStage->type == stageType_t::ST_COLORMAP && drawSurf->bspSurface && pStage->shaderBinder == BindShaderGeneric3D;
```
slipher, line 1356: This data should be stored somewhere else. It's hacky to store drawsurf-specific data on the shader.

src/engine/renderer/Material.cpp
VReaperV: Once some of the `pStage` members are set in `MaterialSystem::GenerateMaterialsBuffer()`, they are then used in `pStage->surfaceDataUpdater()` to set the correct surface data. Essentially, for each unique surface data all of the active variants are stored consecutively, with the differences dictated by variant.

src/engine/renderer/Material.cpp
VReaperV: Perhaps, though I don't really see where else it could be stored. It could be something like a wrapper around `shaderStage_t` passed to `GenerateMaterialsBuffer` instead, and it could store all the variants, I suppose.

src/engine/renderer/Material.cpp
slipher: Seems wrong to store them on `pStage` since if there is more than one draw surf using the same shader (which is the whole point of having variant bit masks, right?) the different values would clobber each other.

VReaperV: The offset stored in `drawSurf_t` includes the offset to the specific variant used for it (https://github.com/DaemonEngine/Daemon/blob/685684fed01feccf884bdb1a5f75aa4ecaf2b1d6/src/engine/renderer/Material.cpp#L452), computed and set in `ProcessStage()` and `AddStage()`.

src/engine/renderer/Material.cpp
VReaperV: Well, it's not truly stored in the shader, the `shaderStage_t` is just used as a proxy, so that `surfaceDataUpdater()` no longer needs a `drawSurf_t` as input. When the materials are initially generated, these use `drawSurf` as input to avoid using up memory and doing processing for stage variants that are never used.

src/engine/renderer/Material.cpp
VReaperV: The mapping between drawSurfs, shader stages, and material buffer can be thought of like this: 
```
shaderStage_t* pStage = drawSurf->shader->stages[stage];

uint32_t baseSurfaceDataIndex = pStage->bufferOffset;
uint32_t variant = drawSurf->shaderVariant[stage]; // Always in the range of [0; active stage variants]
uint32_t variantOffset = pStage->variantOffsets[variant];
void* material = materials + baseSurfaceDataIndex + variantOffset;
uint32_t materialOffset = pStage->materialOffset // Different from bufferOffset for dynamic stages
    + variantOffset; // materialOffset is what's then used in the GLSL shaders to index the material array 
```

src/engine/renderer/Material.cpp
slipher: OK I see that they are just temporary; used during one iteration of GenerateMaterialsBuffer.

If it's too burdensome to add an argument to those "updater" functions, it would still be better to store them in a global variable, instead of misleadingly them with shader stages.

src/engine/renderer/Material.cpp
```diff
@@ -38,6 +38,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "ShadeCommon.h"
 
 GLSSBO materialsSSBO( "materials", 0, GL_MAP_WRITE_BIT, GL_MAP_INVALIDATE_RANGE_BIT );
+GLSSBO texDataSSBO( "texData", 6, GL_MAP_WRITE_BIT, GL_MAP_FLUSH_EXPLICIT_BIT );
+GLUBO lightmapDataUBO( "texData", 7, GL_MAP_WRITE_BIT, GL_MAP_FLUSH_EXPLICIT_BIT );
```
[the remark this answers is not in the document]
VReaperV: Oh, right, fixed.

src/engine/renderer/Material.h
VReaperV: Done.

src/engine/renderer/Material.cpp
VReaperV: Made it work with propagating it through `UpdateSurfaceData*()`.

src/engine/renderer/tr_local.h
```diff
@@ -1064,10 +1102,17 @@ enum class shaderProfilerRenderSubGroupsMode {
 	struct drawSurf_t;
 
 	using stageRenderer_t = void(*)(shaderStage_t *);
-	using stageSurfaceDataUpdater_t = void(*)(uint32_t*, Material&, drawSurf_t*, const uint32_t);
+	using surfaceDataUpdater_t = void(*)(uint32_t*, Material&, shaderStage_t*);
 	using stageShaderBinder_t = void(*)(Material*);
 	using stageMaterialProcessor_t = void(*)(Material*, shaderStage_t*, drawSurf_t*);
 
+	enum class ShaderStageVariant {
```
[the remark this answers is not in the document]
VReaperV: Hm, true, done now.

slipher:
LGTM

PR #1475, opened 2024-12-26
Change u_Color to uint

VReaperV:
Picked from #1473.

`u_Color` was a vec4 that was being set from a `Color` object, which only has 8-bit precision per colour, so using up 4 times more memory than that to transfer it was quite atrocious. Instead, pack it into a uint32_t, then use `unpackUnorm4x8()` to unpack in the shader.

Also adds a macro for unpacking for hardware that doesn't support `GL_ARB_gpu_shader5`.


slipher:
I'm getting some findings with light styles on my diff testing system. I wasn't sure if the difference is real at first, but then I noticed that the correlation between the brightness of the light fixture and the brightness of the rest of the room is broken. Observe how in this screenshot, the closest fixture is bright but the walls are dark. On master they change together.

unvanquished-metro124

Thanks for splitting up the PR, now there will be a lot less stuff to narrow down.



> `u_Color` was a vec4 that was being set from a `Color` object, which only has 8-bit precision per colour

This is not true btw. `tess.svars.color` is floating-point. But I added a log to check for ones outside [0, 1] and didn't find any, so that's apparently not the issue.

VReaperV:
> `tess.svars.color` is floating-point.

Apaprently it is indeed, but all the values that are ever assigned to it are either all 0s, all 1s, an 8-bit precision colour (like `pStage->constantColor` and `backEnd.currentEntity->e.shaderRGBA`), or are clamped to [0; 1] range.

VReaperV:
> I'm getting some findings with light styles on my diff testing system. I wasn't sure if the difference is real at first, but then I noticed that the correlation between the brightness of the light fixture and the brightness of the rest of the room is broken. Observe how in this screenshot, the closest fixture is bright but the walls are dark. On master they change together.

Interestingly enough, I don't get this issue with a build of #1473 on my end, so I might have messed up the cherry-pick somehow (or fixed this issue later without knowing).

VReaperV:
> I'm getting some findings with light styles on my diff testing system. I wasn't sure if the difference is real at first, but then I noticed that the correlation between the brightness of the light fixture and the brightness of the rest of the room is broken. Observe how in this screenshot, the closest fixture is bright but the walls are dark. On master they change together.

I don't think I'm getting this issue here:
image

VReaperV:
I've added an `r_forceRendererTime` cvar to make it easier to test time-based effects.

slipher:
The time for my screenshot is 2375.

slipher:
The bright light with dark walls can also happen on master with some combination of cvars. Need to track down which ones.

VReaperV:
> The time for my screenshot is 2375.

I'm getting this at 2375:
image

Could be some combination of graphics settings...

VReaperV:
> I'm getting some findings with light styles on my diff testing system. I wasn't sure if the difference is real at first, but then I noticed that the correlation between the brightness of the light fixture and the brightness of the rest of the room is broken. Observe how in this screenshot, the closest fixture is bright but the walls are dark. On master they change together.

What settings are you using there? I've tried both high and low, the result is the same as on my screenshot.

slipher:
> The bright light with dark walls can also happen on master with some combination of cvars. Need to track down which ones.

That was `r_overbrightDefaultClamp 1`. So we should make sure clamping is disabled to clearly see the differences. Turning down your r_gamma would also be a good idea.

In the testing script all renderer settings are default, except some resolution/fullscreen stuff.

Try with `r_forceRendererTime 2425`, that gives a bigger difference than 2375 and is probably closer to my screenshot. I may have over-corrected for the screenshot lag time. Here's before/after shots with 2425.


unvanquished-metro101

unvanquished-metro101

VReaperV:
Apparently this is because `ST_STYLELIGHTMAP` and `ST_STYLECOLORMAP` stages are multiplied by the map light factor.

VReaperV:
I've added a fix for this, although I plan to move the value into `u_ColorModulateColorGen` in #1476 since most bits are unused there, and I wanna keep these 2 PR's separate.

slipher:
Merging everything together in #1476 seems better to me, rather than spending any more time reviewing or testing some "temp" stuff.

VReaperV:
Sure, I'll do that.

VReaperV:
Updated version now in #1476.

PR #1476, opened 2024-12-26
Add a uint version of u_ColorModulate

VReaperV:
Picked from #1473.

`u_Color` was a vec4 that was being set from a `Color` object, which only has 8-bit precision per colour, so using up 4 times more memory than that to transfer it was quite atrocious. Instead, pack it into a uint32_t, then use `unpackUnorm4x8()` to unpack in the shader.

Also adds a macro for unpacking for hardware that doesn't support `GL_ARB_gpu_shader5`.

Add `u_ColorModulateColorGen` for use with shaders other than `cameraEffects`, since the latter can get an arbitrary vec4 for it. All other shaders only use 12 distinct states of `u_ColorModulateColorGen`, encoded as 5 bits. The map lightFactor is put into 8 bits of `u_ColorModulateColorGen`.

Also added an `r_forceRendererTime` cvar to debug time-based renderer effects, and since it's a very simple change that I used to test the `u_Color` change, it might as well be here.

Fixes an issue with material shader post-processing that was removing global uniforms that had similar names to a non-global one.


slipher:
This seems to make the heat haze effect on the Plat23 forcefields lose its effect. It renders the same as if `r_heatHaze 0` is set. The preexisting bug #1499 may be relevant. Weapon heat haze effects still appear fine.

VReaperV:
> This seems to make the heat haze effect on the Plat23 forcefields lose its effect. It renders the same as if `r_heatHaze 0` is set. The preexisting bug #1499 may be relevant. Weapon heat haze effects still appear fine.

It is the same issue, yeah. It just so happens that for plat23 on master the size of the struct generated for the heatHaze shader at least somewhat lines up to prdouce a (non-moving) heatHaze effect, but here changing the size of `u_Color` and `u_ColorModulate` changes the size of the struct, and the buffer offsets for dynamic surfaces no longer line up at all. In any case, #1473 fixes that issue entirely, and makes the code that computes the offsets simpler.

src/engine/renderer/tr_init.cpp
```diff
@@ -271,6 +271,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 		"(-1 to profile all stages/materials, rendered in their usual order); "
 		"for materials, material IDs start from opaque materials, depth pre-pass materials are ignored", Cvar::NONE, -1 );
 
+	Cvar::Cvar<int> r_forceRendererTime( "r_forceRendererTime", "Set a specific time (in ms, since the start of the map) for time-based shader effects; -1 to disable", Cvar::NONE, -1 );
```
slipher, line 274: Probably should be Cvar::CHEAT

src/engine/renderer/shaders.cpp
```diff
@@ -65,69 +66,70 @@
...
 
-std::unordered_map<std::string, std::string> shadermap({
-	{ "blur_fp.glsl", std::string(reinterpret_cast<const char*>(blur_fp_glsl), sizeof(blur_fp_glsl)) },
-	{ "blur_vp.glsl", std::string(reinterpret_cast<const char*>(blur_vp_glsl), sizeof(blur_vp_glsl)) },
-	{ "cameraEffects_fp.glsl", std::string(reinterpret_cast<const char*>(cameraEffects_fp_glsl), sizeof(cameraEffects_fp_glsl)) },
-	{ "cameraEffects_vp.glsl", std::string(reinterpret_cast<const char*>(cameraEffects_vp_glsl), sizeof(cameraEffects_vp_glsl)) },
-	{ "computeLight_fp.glsl", std::string(reinterpret_cast<const char*>(computeLight_fp_glsl), sizeof(computeLight_fp_glsl)) },
-	{ "contrast_fp.glsl", std::string(reinterpret_cast<const char*>(contrast_fp_glsl), sizeof(contrast_fp_glsl)) },
-	{ "contrast_vp.glsl", std::string(reinterpret_cast<const char*>(contrast_vp_glsl), sizeof(contrast_vp_glsl)) },
+std::unordered_map<std::string, std::string> shadermap( {
```
slipher, line 69: Please remove random whitespace changes

src/engine/renderer/glsl_source/common.glsl
```diff
@@ -0,0 +1,64 @@
...
+
+/* common.glsl */
+
+/* Common defines */
+
+/* Allows accessing each element of a uvec4 array with a singular ID
+Useful to avoid wasting memory due to alignment requirements
+array must be in the form of uvec4 array[] */
+
+#define UINT_FROM_UVEC4_ARRAY( array, id ) ( array[id / 4][id % 4] )
```
slipher, line 43: These could use more parentheses for safety. `(array)[(id)/4]` etc.

src/engine/renderer/gl_shader.h
```diff
@@ -2151,6 +2151,19 @@ class u_LightFactor :
...
+// HACK: Light factor is set as a global uniform here so that genericMaterial struct can fit into 8 bytes
+class u_ColorModulateLightFactor :
+	GLUniform1f {
+	public:
+	u_ColorModulateLightFactor( GLShader* shader ) :
+		GLUniform1f( shader, "u_ColorModulateLightFactor", true ) {
+	}
+
+	void SetUniform_ColorModulateLightFactor( const float lightFactor ) {
+		this->SetValue( lightFactor / 4.0f ); // Multiplied by 4 in the shader to save on instructions
```
slipher: The constant 4 looks suspicious... I find it produces the wrong result for some vertex lighting if `r_overbrightDefaultExponent` is set to 1. Station15 `47 -1507 27 75 55`. (I don't see any multiplication by 4 in the shader.)

src/engine/renderer/gl_shader.h
```diff
@@ -3603,68 +3616,76 @@ class u_ColorModulate :
...
 			case alphaGen_t::AGEN_VERTEX:
 				needAttrib = true;
-				v[ 3 ] = 1.0f;
+				colorModulate |= Util::ordinal( ColorModulate::ALPHA_ADD );
 				break;
 
 			case alphaGen_t::AGEN_ONE_MINUS_VERTEX:
 				needAttrib = true;
-				v[ 3 ] = -1.0f;
+				colorModulate |= Util::ordinal( ColorModulate::ALPHA_ADD ) | Util::ordinal( ColorModulate::ALPHA_NEGATE );
```
slipher, line 3657: This looks wrong as we would access a length 3 array with index 3.

src/engine/renderer/glsl_source/common.glsl
```diff
@@ -62,3 +63,7 @@ vec4 ColorModulateToColor( const in uint colorMod, const in float lightFactor )
 	colorModulate.a = ( colorModArray[( colorMod & 24 ) >> 3] );
 	return colorModulate;
 }
+
+float ColorModulateToLightFactor( const in uint colorMod ) {
+	return ( colorMod >> 5 ) & 0xF;
```
slipher, line 68: There's some discrepancy here on how many bits it is allowed to be. This takes 4 bits but the comment above colorModArray describes it as 8 bits. Meanwhile in the engine, the worldspawn (`mapOverBrightBits`) limits it to 3 bits and the cvar (`r_overbrightDefaultExponent`) has no limit.

I think 3 bits like the worldspawn should be enough.

slipher (commented):
This fixes the red and black dretch banners in `metro-b1-2`. Previously they were not displayed with the material system.

unvanquished-metro-autosprite2-flame-chains

src/engine/renderer/shaders.cpp
VReaperV: Apparently this got automatically done by my IDE at some point, fixed now.

src/engine/renderer/glsl_source/common.glsl
VReaperV: Done.

src/engine/renderer/gl_shader.h
VReaperV: Fixed, this was a leftover of a different implementation that did not use a look-up array.

src/engine/renderer/glsl_source/common.glsl
VReaperV: Hmm, true, didn't notice that about `mapOverBrightBits`. Changed it to just be 3 bits now.

src/engine/renderer/gl_shader.h
VReaperV: The shader was doing `( colorMod & 4 ) * lightFactor`, and the first part would always produce 4 if it needed to produce `lightFactor`, but this doesn't work for `lightFactor` < 4. So I've changed this now to do a bit-shift in the shader before the multiplication. This fixes the issue on `station15`.

src/engine/renderer/tr_init.cpp
VReaperV: True, added that.

VReaperV:
> This fixes the red and black dretch banners in `metro-b1-2`. Previously they were not displayed with the material system.
> 
> unvanquished-metro-autosprite2-flame-chains

Oh, nice, I saw a few places that work correctly now with this pr, but hadn't seen this one.

slipher:
Light styles are off again. It seems that the light factor is not being applied to light style stages: if I set `r_overbrightDefaultBits 0`, the diff goes away.

VReaperV:
> Light styles are off again. It seems that the light factor is not being applied to light style stages: if I set `r_overbrightDefaultBits 0`, the diff goes away.

This is now fixed.

src/engine/renderer/gl_shader.cpp
```diff
@@ -1434,9 +1438,12 @@ std::string GLShaderManager::ShaderPostProcess( GLShader *shader, const std::str
 		bool skip = false;
 		if ( line.find( "uniform" ) < line.find( "//" ) && line.find( ";" ) != std::string::npos ) {
 			for ( GLUniform* uniform : shader->_uniforms ) {
-				if ( !uniform->IsGlobal() && ( line.find( uniform->GetName() ) != std::string::npos ) ) {
-					skip = true;
-					break;
+				if ( !uniform->IsGlobal() ) {
+					const size_t pos = line.find( uniform->GetName() );
+					if ( pos != std::string::npos && !std::isalpha( line[pos + strlen( uniform->GetName() ) ]) ) {
```
slipher, line 1443: Use `Str::cisalpha`

slipher (commented):
LGTM (with one last comment)

src/engine/renderer/gl_shader.cpp
VReaperV: Fixed.

Issue #645, opened 2022-07-18
water shader is broken


illwieckz:
This is known since a long time, but we now have a proof it worked before, [at least in 2012](https://unvanquished.net/the-human-model-and-new-map-features/):

[![water shader](https://unvanquished.net/wp-content/uploads/2012/09/2012.human-model-and-map-features.Cdc91.jpg)](https://unvanquished.net/wp-content/uploads/2012/09/2012.human-model-and-map-features.Cdc91.jpg)

Here the depicted map is Citadel, but it likely required a modified `.shader` in a modified `.pk3`.

We have a map to investigate this:

- https://github.com/UnvanquishedAssets/UnvanquishedTestAssets/tree/master/src/map-test-water_src.dpkdir

Here is a related talk to improve the material syntax, but improving it doesn't make sense if the shader doesn't work to begin with:

- https://github.com/DaemonEngine/Daemon/issues/261

PR #1477, opened 2024-12-26
Don't set global uniforms in UpdateSurfaceData*

VReaperV:
Picked from #1473.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -215,9 +215,6 @@ void UpdateSurfaceDataGeneric3D( uint32_t* materials, Material& material, drawSu
 
 	gl_genericShaderMaterial->BindProgram( material.deformIndex );
 
-	gl_genericShaderMaterial->SetUniform_ModelMatrix( backEnd.orientation.transformMatrix );
-	gl_genericShaderMaterial->SetUniform_ModelViewProjectionMatrix( glState.modelViewProjectionMatrix[glState.stackIndex] );
-
 	// u_AlphaThreshold
 	gl_genericShaderMaterial->SetUniform_AlphaTest( pStage->stateBits );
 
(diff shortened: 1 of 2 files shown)


slipher:
I don't see the global flag in the constructor on u_ModelViewMatrix or u_UnprojectMatrix.

slipher:
I don't see the global flag in the constructor on u_UnprojectMatrix. (Maybe we should delete the liquid shader so we don't have to discuss that...)

VReaperV:
> I don't see the global flag in the constructor on u_UnprojectMatrix.
Indeed. I have made it global now.

> (Maybe we should delete the liquid shader so we don't have to discuss that...)
Then how would we render water?

illwieckz:
We want water to be fixed, not NUKED:

- https://github.com/DaemonEngine/Daemon/issues/645

slipher:
LGTM

> > (Maybe we should delete the liquid shader so we don't have to discuss that...)
> 
> Then how would we render water?

What do you mean? We don't have any assets using that shader so everything would continue working as before.







> We want water to be fixed, not NUKED:
> 
>     * [water shader is broken #645](https://github.com/DaemonEngine/Daemon/issues/645)

Then the issue can have a pointer to the old version of the code, if indeed it would be useful. Meanwhile we can delete it so it is not a drag on development, what with having to migrate APIs, fix GLSL compile errors, etc. in the dead code.

illwieckz:
Water is basically a variant of both the heatHaze code and the mirror code, so 99% of the code is already here. And the missing 1% is things that had been broken when refactoring other things because we had no asset to test it.

We have a test map: https://github.com/UnvanquishedAssets/UnvanquishedTestAssets/tree/master/pkg/map-test-water_src.dpkdir

And the code is there, probably complete or near to be. The water code was never removed.

VReaperV:
Some tremulous maps have water. https://users.unvanquished.net/~sweet/pkg/map-citadel_0%2B0.dpk IIRC uses the water shader.

VReaperV:
Anyway, if someone has more comments on liquid shaders, please make those in the relevant issues.

PR #1478, opened 2024-12-26
Add r_materialSystemSkip

VReaperV:
Picked from #1473.

This makes debugging issues that are present on material system but not core renderer (or vice versa), as well as comparing performance, much faster, since it can be changed without any sort of restart.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -2061,6 +2061,10 @@ void MaterialSystem::RenderMaterials( const shaderSort_t fromSort, const shaderS
 		return;
 	}
 
+	if ( r_materialSystemSkip.Get() ) {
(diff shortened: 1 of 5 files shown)


slipher:
LGTM, successfully tested.

illwieckz:
Nitpicking, but maybe `r_skipMaterialSystem` is a better name?

VReaperV:
I suppose so. My thinking originally was to group material system cvars under `r_material*`, but then GPU culling has different naming anyway... One other `r_skip*` cvar also exists, so that makes sense.

VReaperV:
But then there's also `r_materialDebug`... I'll keep at as is for now. If needed, it can be renamed later, especially as this is cvar is intended for debugging/profiling.

PR #1480, opened 2024-12-27
Implement MSAA

VReaperV:
Implements #532.

Add `r_msaa`. When set to > 0, an MSAA FBO will be created with that many samples. This FBO will be used for rendering, other than when it requires sampling from current render/depth image. When such rendering is required the MSAA FBO will be blit into the main FBO and vice versa, to resolve the MSAA texture.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -208,6 +208,26 @@ GLuint64 GL_BindToTMU( int unit, image_t *image )
 	return 0;
 }
 
+void GL_BlitFBOToMSAA( FBO_t* fbo ) {
+	R_BindFBO( GL_READ_FRAMEBUFFER, fbo );
+	R_BindFBO( GL_DRAW_FRAMEBUFFER, tr.msaaFBO );
+	glBlitFramebuffer( 0, 0, fbo->width, fbo->height, 0, 0, tr.msaaFBO->width, tr.msaaFBO->height,
+		GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT, GL_NEAREST );
+
+	R_BindFBO( GL_DRAW_FRAMEBUFFER, fbo );
+	glState.currentFBO = nullptr;
+}
+
+void GL_BlitMSAAToFBO( FBO_t* fbo ) {
+	R_BindFBO( GL_READ_FRAMEBUFFER, tr.msaaFBO );
+	R_BindFBO( GL_DRAW_FRAMEBUFFER, fbo );
+	glBlitFramebuffer( 0, 0, tr.msaaFBO->width, tr.msaaFBO->height, 0, 0, fbo->width, fbo->height,
+		GL_COLOR_BUFFER_BIT /* | GL_DEPTH_BUFFER_BIT */, GL_NEAREST );
+
+	R_BindFBO( GL_READ_FRAMEBUFFER, fbo );
+	glState.currentFBO = nullptr;
+}
+
 void GL_BlendFunc( GLenum sfactor, GLenum dfactor )
 {
 	if ( glState.blendSrc != ( signed ) sfactor || glState.blendDst != ( signed ) dfactor )
@@ -3082,6 +3102,9 @@ void RB_RenderBloom()
 
 		GL_PopMatrix(); // special 1/4th of the screen contrastRenderFBO ortho
 
+		if ( r_msaa.Get() ) {
+			GL_BlitMSAAToFBO( tr.mainFBO[backEnd.currentMainFBO] );
+		}
 		R_BindFBO( tr.contrastRenderFBO );
 		GL_ClearColor( 0.0f, 0.0f, 0.0f, 1.0f );
 		glClear( GL_COLOR_BUFFER_BIT );
@@ -3148,6 +3171,11 @@ void RB_RenderBloom()
 		                  backEnd.viewParms.viewportWidth, backEnd.viewParms.viewportHeight );
 	}
 
+	if ( r_msaa.Get() ) {
+		GL_BlitFBOToMSAA( tr.mainFBO[backEnd.currentMainFBO] );
+		R_BindFBO( tr.msaaFBO );
+	}
+
 	// go back to 3D
 	GL_PopMatrix();
 
@@ -3169,6 +3197,10 @@ void RB_RenderMotionBlur()
 	GL_Cull( cullType_t::CT_TWO_SIDED );
 
 	gl_motionblurShader->BindProgram( 0 );
+
+	if ( r_msaa.Get() ) {
+		GL_BlitMSAAToFBO( tr.mainFBO[backEnd.currentMainFBO] );
+	}
 
 	// Swap main FBOs
 	gl_motionblurShader->SetUniform_ColorMapBindless(
@@ -3195,6 +3227,11 @@ void RB_RenderMotionBlur()
 	Tess_InstantQuad( *gl_motionblurShader,
 	                   backEnd.viewParms.viewportX, backEnd.viewParms.viewportY,
 	                   backEnd.viewParms.viewportWidth, backEnd.viewParms.viewportHeight );
+
+	if ( r_msaa.Get() ) {
+		GL_BlitFBOToMSAA( tr.mainFBO[backEnd.currentMainFBO] );
+		R_BindFBO( tr.msaaFBO );
+	}
 
 	GL_PopMatrix();
 
@@ -4628,7 +4665,11 @@ static void RB_RenderView( bool depthPass )
 	backEnd.pc.c_surfaces += backEnd.viewParms.numDrawSurfs;
 
 	// disable offscreen rendering
-	R_BindFBO( tr.mainFBO[ backEnd.currentMainFBO ] );
+	if ( r_msaa.Get() ) {
+		R_BindFBO( tr.msaaFBO );
+	} else {
+		R_BindFBO( tr.mainFBO[backEnd.currentMainFBO] );
+	}
 
 	// we will need to change the projection matrix before drawing
 	// 2D images again
--- a/src/engine/renderer/tr_fbo.cpp
+++ b/src/engine/renderer/tr_fbo.cpp
@@ -303,7 +303,8 @@ R_AttachFBOTexture2D
 */
 void R_AttachFBOTexture2D( int target, int texId, int index )
 {
-	if ( target != GL_TEXTURE_2D && ( target < GL_TEXTURE_CUBE_MAP_POSITIVE_X || target > GL_TEXTURE_CUBE_MAP_NEGATIVE_Z ) )
+	if ( target != GL_TEXTURE_2D  && target != GL_TEXTURE_2D_MULTISAMPLE
+		&& ( target < GL_TEXTURE_CUBE_MAP_POSITIVE_X || target > GL_TEXTURE_CUBE_MAP_NEGATIVE_Z ) )
 	{
 		Log::Warn("R_AttachFBOTexture2D: invalid target %i", target );
 		return;
@@ -355,6 +356,11 @@ void R_AttachFBOTexturePackedDepthStencil( int texId )
 	GL_fboShim.glFramebufferTexture2D( GL_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D, texId, 0 );
 }
 
+void R_AttachFBOTexturePackedDepthStencilMSAA( int texId ) {
+	GL_fboShim.glFramebufferTexture2D( GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D_MULTISAMPLE, texId, 0 );
+	GL_fboShim.glFramebufferTexture2D( GL_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D_MULTISAMPLE, texId, 0 );
+}
+
 /*
 ============
 R_BindFBO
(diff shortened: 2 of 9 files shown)


illwieckz:
Nice! Finally some antialiasing!

It looks like when MSAA is enabled the heathaze from back surface is applied on front surfaces too:
https://imgsli.com/MzMyODgx

illwieckz:
My previous comment was about screenshots from the original post.

When I try the branch and enable MSAA I get a black screen (AMD Radeon with Mesa radeonsi driver).

VReaperV:
> When I try the branch and enable MSAA I get a black screen (AMD Radeon with Mesa radeonsi driver).

That's interesting. Can you try with effects that use the main render target disabled, like heatHaze, bloom etc (or just use the lowest profile to make sure)? Also, does it work in the main menu?

slipher:
I get a completely black screen (other than 2D drawing) with `r_msaa` 4 or 16 on Spacetracks. Plat23, on the other hand, renders. I tested with Nvidia on Windows; default cvars except with reliefmapping enabled.

slipher:
Actually it's just black from whenever the 2nd map is loaded. Must be some cleanup bug

illwieckz:
Can this be rebased on current `master`? 🙂️

VReaperV:
> Can this be rebased on current `master`? 🙂️

Done... but I haven't tested it after rebase, so I don't know if it still works lol.

illwieckz:
It looks like the rebase went wrong! 😅️

```
/home/vsts/work/1/s/src/engine/renderer/tr_backend.cpp:3105:1: error: expected expression
<<<<<<< HEAD
^
```

slipher:
Works great on my favorite jaggies test case from Spacetracks!

Default:
unvanquished-spacetracks-aliasing-jaggies

With `r_msaa 1` - somehow a lot better already. Is that even supposed to be different from having it off?
unvanquished-spacetracks-aliasing-jaggies

`r_msaa 64`
unvanquished-spacetracks-aliasing-jaggies

illwieckz:
On my end it now works. I still get a black screen on AMD and Mesa when I use values higher than 8, but maybe we can cap that.

VReaperV:
> On my end it now works. I still get a black screen on AMD and Mesa when I use values higher than 8, but maybe we can cap that.

Does it say in the log that the samples requested were too high and it's falling back on some other value? According to https://opengl.gpuinfo.org/displaycapability.php?name=GL_MAX_COLOR_TEXTURE_SAMPLES, 8 is indeed the max for AMD, but the renderer should be automatically using that value if you specify a higher one.

illwieckz:
We probably need something similar to this:

```
	// HACK: We want to set the current value, not the latched value
	Cvar::ClearFlags("r_customwidth", CVAR_LATCH);
	Cvar::ClearFlags("r_customheight", CVAR_LATCH);
	Cvar_Set( "r_customwidth", va("%d", windowConfig.vidWidth ) );
	Cvar_Set( "r_customheight", va("%d", windowConfig.vidHeight ) );
	Cvar::AddFlags("r_customwidth", CVAR_LATCH);
	Cvar::AddFlags("r_customheight", CVAR_LATCH);
```

VReaperV:
Ah, yes. I'll add that a bit later.

slipher:
> We probably need something similar to this, but for `r_MSAA`:

Well that's only needed if you really want to use the cvar as a status display for the currently used value. Normally we try to avoid modifying cvars, and would store the current value in `glConfig.msaa` or something like that.

VReaperV:
That's an option too.

src/engine/renderer/tr_image.cpp
```diff
@@ -1217,149 +1235,147 @@ void R_UploadImage( const char *name, const byte **dataArray, int numLayers, int
 	GL_CheckErrors();
 
 	// set filter type
-	switch ( image->filterType )
-	{
-		case filterType_t::FT_DEFAULT:
+	if ( !samples ) {
```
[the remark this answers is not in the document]
VReaperV: I added a commit that does that.

src/engine/renderer/tr_fbo.cpp
```diff
@@ -356,24 +357,31 @@ void R_AttachFBOTexturePackedDepthStencil( int texId )
...
 		R_BindNullFBO();
 		return;
 	}
 
 	GLIMP_LOGCOMMENT( "--- R_BindFBO( %s ) ---", fbo->name );
 
-	if ( glState.currentFBO != fbo )
-	{
-		GL_fboShim.glBindFramebuffer( GL_FRAMEBUFFER, fbo->frameBuffer );
+	if ( glState.currentFBO != fbo ) {
```
slipher, line 383: How about don't read or update `currentFBO` is the `target` is not the default one? Then the `glState.currentFBO = nullptr` stuff is not needed

src/engine/renderer/tr_fbo.cpp
```diff
@@ -430,6 +438,17 @@ void R_InitFBOs()
...
 	R_CheckFBO( tr.mainFBO[1] );
 
+	if( glConfig.MSAA ) {
+		tr.msaaFBO = R_CreateFBO( "msaa", width, height );
+		R_BindFBO( tr.msaaFBO );
+		GL_CheckErrors();
+		R_AttachFBOTexture2D( GL_TEXTURE_2D_MULTISAMPLE, tr.currentRenderImageMSAA->texnum, 0 );
+		GL_CheckErrors();
+		R_AttachFBOTexturePackedDepthStencilMSAA( tr.currentDepthImageMSAA->texnum );
+		GL_CheckErrors();
```
slipher, line 448: Do we want to keep CheckErrors after every line?

slipher:
Also depth fade isn't working. It looks the same as if depth fade were disabled.

unvanquished-vega-depthfade-glitch-decal

slipher:
Also I confirm what illwieckz observed long ago about heat haze penetrating opaque surfaces. Try the following commands and you will see heat haze on the other side of a door being rendered.

```
devmap chasm
setviewpos -634 67 79 -178 4
testshader gfx/weapons/flamer/haze
```

unvanquished-chasm-heathaze-behind-door

src/engine/renderer/tr_fbo.cpp
VReaperV: Done.

VReaperV:
> Also depth fade isn't working. It looks the same as if depth fade were disabled.
> 
> unvanquished-vega-depthfade-glitch-decal

Fixed.



> Also I confirm what illwieckz observed long ago about heat haze penetrating opaque surfaces. Try the following commands and you will see heat haze on the other side of a door being rendered.
> 
> ```
> devmap chasm
> setviewpos -634 67 79 -178 4
> testshader gfx/weapons/flamer/haze
> ```
> 
> unvanquished-chasm-heathaze-behind-door

Also fixed now.

VReaperV:
> There's a bug - dynamic lights are not rendered when MSAA is enabled.

Fixed this too.

VReaperV:
> In the latest version, some transparent surfaces are no longer rendered. For example in the Vega scene below, the cone under the globe and the alpha-tested grates are missing. Also fogs are gone.
> 
> unvanquished_2025-09-05_130336_000

Both fixed now.

slipher:
Turning on `r_msaa` causes tone mapping to be disabled in my tests. This is the same issue as I saw in https://github.com/DaemonEngine/Daemon/pull/1684#issuecomment-3624398313 (but this time I took screenshots with `r_toneMapping 0` to confirm that this is what is happening). I can't reproduce it manually, only in scripted tests, so I need to do more investigation on how to reproduce it.

VReaperV:
By numerical imprecision I mean using the same vertices for the two draws producing slightly different results due to the rasterisation specifics.

VReaperV:
> * SSAO does not work.

Fixed.

VReaperV:
> Turning on `r_msaa` causes tone mapping to be disabled in my tests. This is the same issue as I saw in [#1684 (comment)](https://github.com/DaemonEngine/Daemon/pull/1684#issuecomment-3624398313) (but this time I took screenshots with `r_toneMapping 0` to confirm that this is what is happening). I can't reproduce it manually, only in scripted tests, so I need to do more investigation on how to reproduce it.
> Also color grading is missing in my tests so it's like the whole cameraEffects shader is skipped?

I'm unable to reproduce this on NV and on Intel Arc (both with proprietary drivers).

slipher:
Figured out what the tonemapping issue is. Nothing to do with my scripts; the problem is that screenshots are not the same as what you seen is on-screen. It seems to show the 3D drawing framebuffer before any color effects or 2D drawing.

slipher:
Yeah looks like RB_ReadPixels doesn't bother to select which framebuffer to read from. So it depends on the correct one being active by accident and this could depend on the sequence of `RenderCommand`s.

src/engine/renderer/tr_backend.cpp
```diff
@@ -199,6 +199,47 @@ GLuint64 GL_BindToTMU( int unit, image_t *image )
 	return 0;
 }
 
+static void BlitFBOToMSAA( FBO_t* fbo, const GLbitfield mask ) {
+	R_BindFBO( GL_READ_FRAMEBUFFER, fbo );
+	R_BindFBO( GL_DRAW_FRAMEBUFFER, tr.msaaFBO );
+	glBlitFramebuffer( 0, 0, fbo->width, fbo->height, 0, 0, tr.msaaFBO->width, tr.msaaFBO->height,
+		mask, GL_NEAREST );
+
+	R_BindFBO( GL_DRAW_FRAMEBUFFER, fbo );
```
slipher: This `R_BindFBO` call seems pointless since the calling function immediately does its own `R_BindFBO`.

src/engine/renderer/tr_shade.cpp
```diff
@@ -880,11 +887,21 @@ void Render_generic3D( shaderStage_t *pStage )
 	bool hasDepthFade = pStage->hasDepthFade;
 	bool needDepthMap = pStage->hasDepthFade;
 
-	if ( needDepthMap )
+	const bool needMSAATransion = needDepthMap && backEnd.dirtyDepthBuffer;
+
+	if ( needDepthMap && backEnd.dirtyDepthBuffer && glConfig.textureBarrierAvailable )
```
slipher: I suggest writing the code like this:
```
if ( needDepthMap )
{
    RB_PrepareForSamplingDepthMap();
    
    if ( glConfig.MSAA && backEnd.dirtyDepthBuffer )
    {
        TransitionMSAAToMain( GL_DEPTH_BUFFER_BIT );
        R_BindFBO( tr.msaaFBO );
    }
}
``` 

Since RB_PrepareForSamplingDepthMap already checks the texture barrier available and dirty depth buffer conditions. And the following TransitionMSAAToMain and R_BindFBO calls always occur together, so the intent is clearer if they are in the same if block.

src/engine/renderer/tr_image.cpp
```diff
@@ -1150,12 +1159,21 @@ void R_UploadImage( const char *name, const byte **dataArray, int numLayers, int
...
 				else
 				{
-					GL_TexImage2D( target, 0, internalFormat, scaledWidth, scaledHeight, 0, format, GL_UNSIGNED_BYTE, scaledBuffer, isSRGB );
+					if ( samples ) {
+						GL_TexImage2DMultisample( target, samples, internalFormat, scaledWidth, scaledHeight, fixedSampleLocations, isSRGB );
+					} else {
+						GL_TexImage2D( target, 0, internalFormat, scaledWidth, scaledHeight, 0, format, GL_UNSIGNED_BYTE, scaledBuffer, isSRGB );
+					}
 				}
+				GL_CheckErrors();
```
slipher: GL_CheckErrors not needed here since it is done after the switch statement.

VReaperV: Where after the switch statement?

src/engine/renderer/tr_shade.cpp
VReaperV: Oh, yea, `RB_PrepareForSamplingDepthMap()` must've been changed since this was written. Fixed now.

src/engine/renderer/tr_backend.cpp
```diff
@@ -199,6 +199,47 @@ GLuint64 GL_BindToTMU( int unit, image_t *image )
...
+}
+
+static void BlitMSAAToFBO( FBO_t* fbo, const GLbitfield mask ) {
+	R_BindFBO( GL_READ_FRAMEBUFFER, tr.msaaFBO );
+	R_BindFBO( GL_DRAW_FRAMEBUFFER, fbo );
+	glBlitFramebuffer( 0, 0, tr.msaaFBO->width, tr.msaaFBO->height, 0, 0, fbo->width, fbo->height,
+		mask, GL_NEAREST );
+
+	R_BindFBO( GL_READ_FRAMEBUFFER, fbo );
+	glState.currentFBO = fbo;
```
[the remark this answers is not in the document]
VReaperV: Changed to just bind `GL_READ_FRAMEBUFFER`.

src/engine/renderer/tr_backend.cpp
VReaperV: Changed to just bind `GL_READ_FRAMEBUFFER`.

src/engine/renderer/tr_image.cpp
slipher: Maybe I should say after the outermost if/else. Right before the new location of `switch ( image->internalFormat )`

VReaperV:
> Unfortunately I found another bug when trying a different graphics preset... when MSAA is on and SSAO is off, depth fade doesn't work (no alpha modification occurs). I reproduced it with Nvidia so you should hopefully be able to see it.

Fixed I think. Also added the missing MSAA resolve to global fog.

PR #1482, opened 2024-12-30
Add `bounds_t` with pre-aligned `mins` and `maxs`

illwieckz:
This is not urgent at all. This is complementary to:

- https://github.com/DaemonEngine/Daemon/pull/1142

The `BoxOnPlaneSide()` function is known to be a hot spot, being called by multiple recursive functions. Right now we spend a lot of time in `_mm_loadu_ps`, we have to call `sseLoadVec3Unsafe()` explicitly because we can't guess if the data is aligned or not. This comes with multiple downside:

- `_mm_loadu_ps` is said to be slower than `_mm_load_ps`, and that fits my experience¹.
- The compiler doesn't optimize `_mm_loadu_ps` and will always call it if we ask for it explicitely, even if the data is already aligned, and by experience, even if no copy is needed.

So the idea is to introduce some nice `bounds_t` struct that uses aligned `mins` and `maxs`, same for `cplane_t` with an aligned `normal`. When doing that, we can write an explicit `_mm_load_ps` that is faster than `_mm_loadu_ps`, and most of the cases, because the compiler knows the data is `const` and already aligned, the compiler just removes the `_mm_load_ps` and process the data without any copy.

___

¹ Some times ago I tried to write optimized SSE code for some other functions, but the code was slower because of the explicit `_mm_loadu_ps` call. I even noticed copying an unaligned vec3 to an aligned vec3 before calling  `_mm_load_ps` could make the code faster when the compiler notices the data is already aligned and skips the copy and calls `_mm_load_ps` (or even doesn't need to call it at all).


illwieckz:
For now there is a bug somewhere I have not found yet. The bug can be seen by loading the `plat23` map and looking to the left (some surfaces will disappear).

In my first iteration of the branch I made a stupid mistakes by passing the `bounds_t` by value instead of by reference, meaning functions modifying it would not modify it. I assume this is now fixed but I guess I left somewhere another mistake as stupid as that.

illwieckz:
Note: the percentage of time spent is not reliable in those screenshots because I use a `viewpos` on some acid tubes spamming acid gas and the amount of particles is very variable.

Before (C), a lot of time is spent in `sseLoadVec3Unsafe()`:

20241230-201855-000 orbit-unvanquished-aligned-sse-load

After (C), the first loads are just completely skipped:

20241230-202108-000 orbit-unvanquished-aligned-sse-load

Before (Asm), we see three `movups` instructions:

20241230-201857-000 orbit-unvanquished-aligned-sse-load

After (Asm), se see only two `movaps` instructions:

20241230-202121-000 orbit-unvanquished-aligned-sse-load

illwieckz:
So, as I said, this is not urgent. But I would appreciate people re-reading what I did in hope someone finds the stupid mistake I may have done. The code change should be straightforward because it's basically rewriting with a different data layout, there is no algorithm change to be expected. Unfortunately this data struct is used in many places so the patch is massive.

There are many functions in `CM` not using the new structure yet that can be migrated, but that can be done later as the patch is already heavy as it is. Migrating other functions would be mostly code clean-up and only about code purity, it is not required to make it work. Though maybe it can also help the compiler to optimize in other places. For the same reasons, I only modified the game to be compatible with the new structs and shared functions, using some wrapping around the new struct, to keep the patch light on game side (porting the game to the new struct for the sake of purity can be done later as well).

illwieckz:
I also added some constness to some function input, which may help the compiler to optimize a bit more some other places of the code. I doubt this extra-constness is the cause of the bug I'm facing because the compiler doesn't complain I'm writing to some `const` structs, so hopefully I only added some `const` keywords to structs that are only read and not written.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -3612,11 +3612,11 @@ static void R_LoadNodesAndLeafs( lump_t *nodeLump, lump_t *leafLump )
...
-		for ( j = 0; j < 3; j++ )
-		{
-			out->mins[ j ] = LittleLong( in->mins[ j ] );
-			out->maxs[ j ] = LittleLong( in->maxs[ j ] );
-		}
+		out->bounds.mins[ 0 ] = LittleLong( in->mins[ 0 ] );
+		out->bounds.mins[ 1 ] = LittleLong( in->mins[ 1 ] );
+		out->bounds.mins[ 2 ] = LittleLong( in->mins[ 2 ] );
+		out->bounds.maxs[ 1 ] = LittleLong( in->maxs[ 1 ] );
+		out->bounds.maxs[ 2 ] = LittleLong( in->maxs[ 2 ] );
```
illwieckz, line 3619: Here I forgot:

```c++
out->bounds.maxs[ 0 ] = LittleLong( in->maxs[ 0 ] );
```

illwieckz: That was the bug I was looking for.

illwieckz:
So I reread _everything_ and found the bug I was looking for. I missed the conversion of one line (I just did not re-added the new one, that's why there was no compile type error).

It works.

illwieckz:
I found another bug affecting BSP movers like doors. One can test it in the Vega map. The doors disappear according to the viewing angle / point of view.

DolceTriade:
Any fps difference?

slipher:
I think you can make a drop-in aligned replacement for vec3_t like this:
```
struct alignas(16) alignedVec3_t : public std::array<float, 3> {};
```

This can be used in arrays so then you don't need `bounds_t` with the ugly `at` function.

src/engine/renderer/tr_model_iqm.cpp
```diff
@@ -880,31 +878,26 @@ R_CullIQM
...
+	bounds_t localBounds;
+	BoundsCopy( model->bounds, localBounds );
 
-	if ( anim && ( bounds = anim->bounds ) ) {
+	if ( anim && anim->bounds ) {
 		// merge bounding box provided by the animation
-		BoundsAdd( localBounds[ 0 ], localBounds[ 1 ],
-			   bounds, bounds + 3 );
+		bounds_t bounds;
+		BoundsSet( bounds, anim->bounds, anim->bounds + 3 );
```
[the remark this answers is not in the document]
VReaperV: Ah, wait, it's still just float[6] here, nevermind.

src/engine/qcommon/q_math.cpp
```diff
@@ -815,142 +814,153 @@ int BoxOnPlaneSide( const vec3_t emins, const vec3_t emaxs, const cplane_t *p )
...
-	maxs[ 0 ] = maxs[ 1 ] = maxs[ 2 ] = 0;
+	b = {};
 }
 
-void ClearBounds( vec3_t mins, vec3_t maxs )
+void ClearBounds( bounds_t &b )
 {
-	mins[ 0 ] = mins[ 1 ] = mins[ 2 ] = 99999;
-	maxs[ 0 ] = maxs[ 1 ] = maxs[ 2 ] = -99999;
+	bounds_t cleared = {
```
VReaperV, line 838: `static const`?

src/engine/qcommon/q_shared.h
```diff
@@ -1718,7 +1720,7 @@ void MatrixTransformBounds( const matrix_t m, const bounds_t &b, bounds_t &o );
 		vec_t dist;
 		byte   type; // for fast side tests: 0,1,2 = axial, 3 = nonaxial
 		byte   signbits; // signx + (signy<<1) + (signz<<2), used as lookup during collision
-		byte   pad[ 2 ];
+		byte pad[ 12 ];
```
VReaperV, line 1723: This looks wrong. The previous variables in the struct add up to 6 bytes.

src/shared/client/cg_api.cpp
```diff
@@ -532,14 +531,22 @@ int trap_R_BlendSkeleton( refSkeleton_t *skel, const refSkeleton_t *blend, float
...
+    bounds_t bounds;
+
     for ( i = 0; i < 3; i++ )
     {
-        bounds[ 0 ][ i ] = skel->bounds[ 0 ][ i ] < blend->bounds[ 0 ][ i ] ? skel->bounds[ 0 ][ i ] : blend->bounds[ 0 ][ i ];
-        bounds[ 1 ][ i ] = skel->bounds[ 1 ][ i ] > blend->bounds[ 1 ][ i ] ? skel->bounds[ 1 ][ i ] : blend->bounds[ 1 ][ i ];
+        bounds.maxs[ i ] = skel->bounds.maxs[ i ]
+            < blend->bounds.maxs[ i ]
+            ? skel->bounds.maxs[ i ]
+            : blend->bounds.maxs[ i ];
```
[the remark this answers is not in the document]
illwieckz: Good catch! Fixed.

src/engine/renderer/tr_world.cpp
```diff
@@ -304,15 +304,11 @@ void R_AddBSPModelSurfaces( trRefEntity_t *ent )
 	bspModel = pModel->bsp;
 
 	// copy local bounds
-	for ( i = 0; i < 3; i++ )
-	{
-		ent->localBounds[ 0 ][ i ] = bspModel->bounds[ 0 ][ i ];
-		ent->localBounds[ 1 ][ i ] = bspModel->bounds[ 1 ][ i ];
-	}
+	BoundsCopy( ent->localBounds, bspModel->bounds );
```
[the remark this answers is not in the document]
illwieckz: Good catch! Fixed.

src/engine/qcommon/q_math.cpp
illwieckz: Done.

src/engine/qcommon/q_shared.h
illwieckz: That was because of:

- https://github.com/Unvanquished/Unvanquished/pull/3265#issuecomment-2566006183

But maybe I'm wrong with the values.

illwieckz:
> I think you can make a drop-in aligned replacement for vec3_t like this:
> 
> ```
> struct alignas(16) alignedVec3_t : public std::array<float, 3> {};
> ```
> 
> This can be used in arrays so then you don't need `bounds_t` with the ugly `at` function.

Interesting, how such type can be used and where?

slipher:
As a direct replacement for vec3_t. `alignedVec3_t   localBounds[ 2 ];` This would automatically work with indexing operations and functions like `VectorCopy`. But you would need a `.data()` or whatever to pass it to functions taking a `float*` (this includes functions with `vec3_t` parameter)

src/engine/qcommon/q_shared.h
VReaperV: Sure, but should be `byte pad[ 10 ]`, rather than `byte pad[ 12 ]`.

VReaperV: (small correction to my previous comment: they add up to 18 bytes, not 6)

illwieckz:
I like the idea of having have an `alignedVec3_t` but I don't see the benefit of making `bounds_t` an array of it. There is only a few usage of `at()`, while using `.data()` everywhere would be much verbose. Also I like the idea of explicit `mins` and `maxs` naming, I guess we may be able to do some `mins()` and `maxs()` and I tried to implement that with an `alignedVec3_t` but that because much more complex.

I just rewrote `at()` in a way it is now guaranteed to be branchless.

slipher:
> I like the idea of having have an `alignedVec3_t` but I don't see the benefit of making `bounds_t` an array of it.

I'm suggesting that no `bounds_t` is needed. As I wrote in the previous comment, it could be used like `alignedVec3_t   localBounds[ 2 ]`;

illwieckz:
But I like having a explicit `bounds_t`! It being implemented the way I did it or doing it like `using bounds_t = alignedVec3_t[2]`, I want to do it…

cmake/DaemonFlags.cmake
```diff
@@ -68,10 +68,19 @@ macro(set_c_cxx_flag FLAG)
     set_c_flag(${FLAG} ${ARGN})
     set_cxx_flag(${FLAG} ${ARGN})
 endmacro()
+
+macro(set_kind_linker_flag KIND FLAG)
```
slipher, line 72: There are random cmake changes in here by mistake

src/engine/qcommon/q_shared.h
```diff
@@ -296,6 +296,19 @@ void  Com_Free_Aligned( void *ptr );
...
+struct bounds_t
+{
+	alignas(16) vec3_t mins;
+	vec_t pad_mins;
+	alignas(16) vec3_t maxs;
+	vec_t pad_maxs;
+
+	vec3_t& at( bool index )
+	{
+		return ( &mins + ( index * 16 ) )[ 0 ];
```
slipher, line 308: That's undefined behavior, like the `(&v.x)[i]` thing with GLM.

illwieckz: But how can this be undefined if the struct layout is guaranteed? And if it isn't guaranteed, how can we use structs for defining file storage format or network packet formats if the layout of the data isn't guaranteed?

Is there a type in C/C++ that guarantees that `alignas(16) vec4_t, alignas(16) vec4_t` has an offset of `sizeof(vec4_t)` or `16`?

illwieckz: Can this be solved with static assertion (if the compiler doesn't do as expected, don't let the code be compiled?).

slipher: It doesn't matter that the layout is guaranteed. It's illegal to use an index that goes outside the bounds of an array. (A non-array object is treated like an array of length 1.)

illwieckz: But this is not an index, this is a pointer. Pointer arithmetic isn't undefined.

slipher: I just said "index" to explain it in a simple way. But in fact the pointer arithmetic IS undefined on its own, even if you don't dereference anything. If you have `int a[3]`, the only pointers it is legal to form are `a + 0`, `a + 1`, `a + 2`, and `a + 3`. If you do `int *b = a - 1`, that's UB. If you do `int *c = a + 4` that's UB.

illwieckz: Why would this be undefined?

```c++
	vec_t* dereference( vec3_t array )
	{
		return &array[ 0 ];
	}

	vec_t* offset( vec_t *pointer, bool index )
	{
		return pointer + ( index * 16 );
	}

	vec_t* at( bool index )
	{
		vec_t *pointer = dereference( mins );
		pointer = offset( pointer, index );
		return &pointer[ 0 ];
	}
```

illwieckz: What if we just do that instead?

```c++
	vec_t* at( bool index )
	{
		return &mins[ 0 ] + ( index * 16 );
	}
```

slipher: > Why would this be undefined?

Because the C++ standard says so.

> What if we just do that instead?

That's the same as before. Pointer arithmetic that goes outside the bounds of the object (`mins`).

illwieckz: …but it doesn't go outside of the struct this object is! There is no way to find something else than `maxs` there!

illwieckz: Or one would say that if you have `float array[3]`, doing `&array[0] + 2` isn't defined because that goes outside of the object (`array[0]`).

slipher: OK I didn't explain it precisely. Let's cite people who can. https://en.cppreference.com/w/cpp/language/operator_arithmetic#Pointer_arithmetic

> Or one would say that if you have float array[3], doing &array[0] + 2 isn't defined because that goes outside of the object (array[0]).

This is the case `Otherwise, if P points to the ith element of an array object x with n elements, given the value of J as j, P is added or subtracted as follows: `. J can be from 0 to 3.

If you do `&array` instead of `&a[0]`, it's the case `Otherwise, if P points to a complete object, a base class subobject or a member subobject y, given the value of J as j, P is added or subtracted as follows: `. J can be 0 or 1. Note that if it is 1, the resulting pointer is a "pointer past the end" and is illegal to dereference, even if it would happen to have the same numerical value as a valid object.

illwieckz: So, is this defined?

```c++
	vec_t* at( bool index )
	{
		return (vec_t*) this + ( index * 16 );
	}
```

The pointer doesn't go past the end of the object `this`.

illwieckz: Well, that doesn't work, actually only the `if (index) return maxs; return mins` work:

https://godbolt.org/z/d85zMKh7P

What's curious is that the compilers (GCC, Clang) seem to produce the same code for all three functions, but produce garbage on the functions doing pointer arithmetic.

VReaperV: > _Edit:_ In fact the computed offsets are wrong.

The wrong offsets are the ones in your `at()` functions, i. e. it should be `index * 4`, rather than `index * 16`.

VReaperV: I've seen claims that, essentially, doing this would not be UB:
`return (vec_t*) ( (char*) &mins[0] + ( index ? offsetof( boundsA_t, maxs ) : offsetof( boundsA_t, mins ) ) );`
However, don't quote me on that, & it produces the same assembly anyway.

PR #1483, opened 2024-12-31
Merge generic2D shader into generic

VReaperV:
Based on #1479.

The only difference is in a compile macro that's used to skip UI for subgroup profiling, which doesn't require a whole new shader.


illwieckz (approved):
If it works, LGTM.

VReaperV:
Alright, I just need to rebase without #1479's commits... I thought I did some changes to the generic shader there, but apparently not.

PR #1486, opened 2025-01-03
Revert "Merge generic2D shader into generic"

slipher:
This reverts #1483.

This is a loading time regression that causes several useless shader permutations to be built.
With "Merge generic2D shader into generic":
	built 241 glsl shaders in 19483 msec
Without it:
	built 224 glsl shaders in 18827 msec

The GENERIC_2D macro would need to be removed to merge these without introducing unnecessary permutations.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -43,6 +43,7 @@ ShaderKind shaderKind = ShaderKind::Unknown;
 
 // *INDENT-OFF*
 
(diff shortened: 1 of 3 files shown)


illwieckz:
Oopsie.

VReaperV:
Can probably be done with `IsUnusedPermutation()`.

VReaperV:
Well, that or just temporarily skipping subgroup profiler.

slipher:
Superseded by #1487.

PR #1487, opened 2025-01-03
FIx unnecessary generic shader permutations

VReaperV:
NUKE the GENERIC_2D macro and add `trGlobals_t::skipSubgroupProfiler` to skip the subgroup profiler instead.

Alternative to #1486.


illwieckz:
- with https://github.com/DaemonEngine/Daemon/pull/1483
  `built 341 glsl shaders in 3488 msec`
- with https://github.com/DaemonEngine/Daemon/pull/1486
  `built 324 glsl shaders in 3129 msec`
- with this:
  `built 323 glsl shaders in 3200 msec`

illwieckz (approved):
LGTM.

PR #1488, opened 2025-01-05
Add -noforward/-forward-only command line args

### LINKED ISSUE #942: feature request: add -forward-only option to not start engine if there is no engine already running
Right now we can do things like `daemonded +devmap <parpax>` to forward the command to an existing engine already running. But if the engine is not already running, the engine will start.

This is sometime wanted, for example `daemon +connect <servername>` will either connect to the server, or start the game and connect to the server, which is the mechanism used by the `unv://` protocol, which is very cool.

This is sometime unwanted, like if you're a server owner and all you want to do is to forward a command to an existing server, and you only want to forward commands to existing servers. For example, let's say you have a systemd service managing the `daemonded` server, you may want to do `sudo -u unvanquished daemonded +say hello` and be sure the engine doesn't start an instance in your current tty if for some reason the engine is not running yet.

So I suggest to implement an option like `-forward-only` that would not start the engine if there is no existing instance to forward the command to.

slipher:
By default Daemon forwards commands passed on the command line if there is one, or otherwise starts a new instance. The new command-line arguments can be used to only allow one or the other of the behaviors.

Fixes #942.

### DIFF
--- a/src/engine/framework/System.cpp
+++ b/src/engine/framework/System.cpp
@@ -491,7 +491,8 @@ struct cmdlineArgs_t {
 	cmdlineArgs_t()
 		: homePath(Application::GetTraits().defaultHomepath), libPath(FS::DefaultLibPath()),
 		  reset_config(false), use_crash_handlers(true),
-		  use_curses(Application::GetTraits().useCurses) {}
+		  use_curses(Application::GetTraits().useCurses),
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

illwieckz (approved):
Thank you very much, that's greatly appreciated! LGTM.

PR #1490, opened 2025-01-08
fixup: cmake: set variables before using them

illwieckz:
Fixup for:

- https://github.com/DaemonEngine/Daemon/pull/1468

I missed this mistake.

`DaemonPlatform` should always be included for now because of some other chicken-and-egg problem (not in the scope of that PR).


illwieckz:
This is a fixup for a mistake currently in `master` and is obvious, we better not wait to merge that.

PR #1492, opened 2025-01-08
Fix out of order shader log messages

VReaperV:
This can sometimes happen, leading to missing log entries when printing the log. Fixes that by sorting them first.


slipher:
LGTM

PR #1494, opened 2025-01-10
Migrate render commands to C++ interface

slipher:
Migrate commands to the new interface. That has the benefit of show descriptions for the commands, and using `Print` so that the output is not affected by log suppression.

Most of the diffs are indentation changes, so you might like to review it as follows:
```
git fetch origin pull/1494/head
git show -4 FETCH_HEAD --ignore-space-change
```

### DIFF
--- a/src/engine/client/cl_main.cpp
+++ b/src/engine/client/cl_main.cpp
@@ -2474,18 +2474,15 @@ void CL_Shutdown()
 
 	Cmd_RemoveCommand( "cmd" );
 	Cmd_RemoveCommand( "configstrings" );
-	Cmd_RemoveCommand( "userinfo" );
 	Cmd_RemoveCommand( "snd_restart" );
 	Cmd_RemoveCommand( "vid_restart" );
 	Cmd_RemoveCommand( "disconnect" );
 	Cmd_RemoveCommand( "connect" );
 	Cmd_RemoveCommand( "localservers" );
 	Cmd_RemoveCommand( "globalservers" );
-	Cmd_RemoveCommand( "rcon" );
 	Cmd_RemoveCommand( "ping" );
 	Cmd_RemoveCommand( "serverstatus" );
 	Cmd_RemoveCommand( "showip" );
-	Cmd_RemoveCommand( "model" );
 
 	CL_ClearKeyBinding();
 	CL_ClearInput();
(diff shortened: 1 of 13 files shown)


VReaperV:
The descriptions for e. g. `shaderexp` and `buildcubemaps` show up as `--`, rather than `evaluate a q3shader expression (RB_EvalExpression)` and `generate cube probes for reflection mapping`, is this supposed to happen?

slipher:
> The descriptions for e. g. `shaderexp` and `buildcubemaps` show up as `--` when using `/listCmds`, rather than `evaluate a q3shader expression (RB_EvalExpression)` and `generate cube probes for reflection mapping`, is this supposed to happen?

Can't reproduce. Are you sure you're building and running the right code?

VReaperV:
I'm certain of that.

Apparently, it works properly with auto-completion, but not otherwise:
Regular usage:
> ]/listcmds gfx
> 0 commands
> ]/listcmds shader
> 0 commands
> ]/listcmds build
>   build an old style vm command
> 1 commands
> ]/listcmds gl
>   globalservers --
> 1 commands

Auto-completion:
> ^3-> ^*/shader
>    shaderexp  evaluate a q3shader expression (RB_EvalExpression)
>    shaderpath cvar - "^*" - text - path to load GLSL source files at runtime
> ^3-> ^*/build
>    build         an old style vm command
>    buildcubemaps generate cube probes for reflection mapping
> ^3-> ^*/gl
>    globalservers --
>    glsl_restart  recompile GLSL shaders (useful when shaderpath is set)
> ^3-> ^*/listcmds g
>    g_                  
>    gfxinfo             dump graphics driver and configuration info

slipher:
OK so there's a bug with `listCmds` that it won't show commands without a flag.
```
static ListCmdsCmd listCmdsRegistration("listCmds", BASE, "lists all the commands", ~(CVAR|ALIAS));
```

But I don't see `--` anywhere with the migrated commands.

VReaperV:
> But I don't see `--` anywhere with the migrated commands.

This works fine on my end after a full rebuild.

VReaperV:
> OK so there's a bug with `listCmds` that it won't show commands without a flag.

Yeah, found the same issue.

VReaperV (approved):
LGTM

PR #1496, opened 2025-01-12
NUKE leftover CPU lightGrid code, minor fixes and cleanup

VReaperV:
- NUKE some code that was used before lightGrid was moved to the GPU, and which was now just waiting resources.
- Made `r_forceAmbient` (hopefully) work by actually using it with the GLSL lightGrid.
- `r_forceAmbient` and `r_ambientScale` changed to new-style cvars, +some cleanup.


src/engine/renderer/tr_local.h
```diff
@@ -2890,9 +2889,9 @@ enum class shaderProfilerRenderSubGroupsMode {
 
 	extern cvar_t *r_wolfFog;
 	extern cvar_t *r_noFog;
-
-	extern cvar_t *r_forceAmbient;
-	extern cvar_t *r_ambientScale;
+
```
illwieckz, line 2892: nitpick: there is a useless tab there.

illwieckz (approved):
If it WFY, LGTM.

illwieckz:
Speaking of alternate lightgrid code, @VReaperV do you think it's technically doable to have a “lowest” lightgrid implementation (only meaningful for models), that would just sample the lightgrid color at the origin of the model, and uniformly lit the whole model with that color (probably using the rgbgen fields)?

VReaperV:
> Speaking of alternate lightgrid code, @VReaperV do you think it's technically doable to have a “lowest” lightgrid implementation (only meaningful for models), that would just sample the lightgrid color at the origin of the model (CPU side), and uniformly lit the whole model with that color (probably using the rgbgen fields), skipping entirely the lightgrid GLSL, relying on the fullbright GLSL code?

Yeah, that would be easily doable. In fact, particles and trails already do pretty much that (but they interpolate between a few points) - which is kinda stupid, because it actually involves an IPC round-trip for no reason...
Why do the uniform lighting for models though?

src/engine/renderer/tr_bsp.cpp
```diff
@@ -4338,13 +4352,12 @@ void R_LoadEntities( lump_t *l, std::string &externalEntities )
 		// check for ambient color
 		else if ( !Q_stricmp( keyname, "_color" ) || !Q_stricmp( keyname, "ambientColor" ) )
 		{
-			if ( r_forceAmbient->value <= 0 )
-			{
-				sscanf( value, "%f %f %f", &tr.worldEntity.ambientLight[ 0 ], &tr.worldEntity.ambientLight[ 1 ],
-				        &tr.worldEntity.ambientLight[ 2 ] );
+			if ( r_forceAmbient.Get() == 0 ) {
```
slipher, line 4355: If `r_forceAmbient` is overriding some value specified by the BSP, shouldn't the cvar be off by default?

VReaperV: Perhaps. I just used the value that was already there for it.

VReaperV:
(rebased)

src/engine/renderer/tr_bsp.cpp
VReaperV: There's also the issue that there could be maps relying on it.

VReaperV:
I'm merging it as is, if there are further concerns for the default value of `r_forceAmbient`, they can be fixed in a different pr, since it's just a single number change. This pr does not modify the default value, nor its range from what we already had (with reasoning outlined in the review).

VReaperV:
> > Why do the uniform lighting for models though?
> 
> Just as an ugly and cheap alternative for the “lowest” preset that would not be a cheat, because the model would still be dark in a dark room and lit in a lit room. It would save on GLSL processing and on GPU memory, and not involve 3D textures in GPU memory. Just a cheap alternative that would not be a cheat.

You might be able to just use https://github.com/DaemonEngine/Daemon/blob/5a071e2cfe78b815f17fc75b2df5e192b8ca236a/src/engine/renderer/tr_light.cpp#L188
Interpolating probably shouldn't be too much of an issue? If it is, just simply getting the closest point can be done easily enough.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -4116,6 +4124,12 @@ void R_LoadLightGrid( lump_t *l )
 		tmpDirected[ 2 ] = in->directed[ 2 ];
 		tmpDirected[ 3 ] = 255;
 
+		if ( tmpAmbient[0] < r_forceAmbient.Get() &&
```
slipher, line 4127: This doesn't actually do anything because it's compariing a 0-255 value with a 0-1 value.

VReaperV: https://github.com/DaemonEngine/Daemon/pull/1510

PR #1500, opened 2025-01-15
cmake: do not use space in compiler string definition, it is not compatible with the standalone MinGW for Windows

illwieckz:
Do not use space in compiler string definition, it is not compatible with the standalone MinGW for Windows.

Fix #1497:

- https://github.com/DaemonEngine/Daemon/pull/1497


illwieckz:
I also added  patch to fix PCH build:

- https://github.com/DaemonEngine/Daemon/issues/1508

And some other touch-ups.

illwieckz:
Delaying the merge of this just makes every commit merged between the original merge and those fix-ups harder to bisect, so I just merge this.

PR #1509, opened 2025-01-18
Set engine command flags

slipher:
As @VReaperV discovered while reviewing #1494, `listCmds` can't list a command if it has no flags defined. Fix this by defining flags for all the engine commands and disabling the constructor that doesn't define flags.

### DIFF
--- a/src/common/Command.cpp
+++ b/src/common/Command.cpp
@@ -493,11 +493,13 @@ namespace Cmd {
         return *Cmd::GetEnv();
     }
 
+#ifndef BUILD_ENGINE
     StaticCmd::StaticCmd(std::string name, std::string description)
     :CmdBase(0){
         //Register this command statically
         AddCommand(std::move(name), *this, std::move(description));
     }
+#endif
 
     StaticCmd::StaticCmd(std::string name, const int flags, std::string description)
     :CmdBase(flags){
@@ -509,9 +511,11 @@ namespace Cmd {
         return {};
     }
 
+#ifndef BUILD_ENGINE
     LambdaCmd::LambdaCmd(std::string name, std::string description, RunFn run, CompleteFn complete)
     :StaticCmd(std::move(name), std::move(description)), run(run), complete(complete) {
     }
+#endif
     LambdaCmd::LambdaCmd(std::string name, int flags, std::string description, RunFn run, CompleteFn complete)
     :StaticCmd(std::move(name), flags, std::move(description)), run(run), complete(complete) {
     }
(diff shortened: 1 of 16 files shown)


src/common/Command.cpp
```diff
@@ -493,11 +493,13 @@ namespace Cmd {
         return *Cmd::GetEnv();
     }
 
+#ifndef BUILD_ENGINE
```
DolceTriade, line 496: Please leave a comment for this ifdef? Are we basically only allowing this in the dummy app?

src/common/Command.h
```diff
@@ -41,15 +41,18 @@ namespace Cmd {
...
-        AUDIO            = BIT(5),
-        SGAME_VM         = BIT(6),
-        CGAME_VM         = BIT(7),
-        KEY_BINDING      = BIT(8),
+        BASE             = BIT(0), // anything in the dummy app?
+        SERVER           = BIT(1),
+        CLIENT           = BIT(2),
+        RENDERER         = BIT(3) | CLIENT,
+        AUDIO            = BIT(4) | CLIENT,
+        KEY_BINDING      = BIT(5) | CLIENT,
```
VReaperV, line 49: This results in some of the commands polluting `/listClientCmds`. Just use the `BIT()` part and add `CLIENT` to `cmdFlags` for list cmds, rather than `~CLIENT` for `showCmdFlags`?

src/common/Command.cpp
slipher: There is a comment explaining this where the constructor is declared in the headers. It is allowed only for gamelogic.

src/common/Command.h
slipher: > This results in some of the commands polluting `/listClientCmds`.

That is intentional - renderer/audio/keyboard are components of the client. I could make a command to list only the exclusively-CLIENT ones if I had a succinct way to describe it (basically, commands that exist in the tty client).



> Just use the `BIT()` part and add `CLIENT` to `cmdFlags` for list cmds, rather than `~CLIENT` for `showCmdFlags`?

That's another equally valid way to get to the same result. I can implement it that way if you think it is better.

VReaperV: > That is intentional - renderer/audio/keyboard are components of the client. I could make a command to list only the exclusively-CLIENT ones if I had a succinct way to describe it (basically, commands that exist in the tty client).

Oh, I see.

> That's another equally valid way to get to the same result. I can implement it that way if you think it is better.

Yes, I think it's a more straightforward way of doing it.

slipher: Done.

DolceTriade (approved):

PR #1510, opened 2025-01-21
Fix r_forceAmbient comparison

VReaperV:


DolceTriade:
for the peanut gallery, why was the before comparison wrong?

VReaperV:
`r_forceAmbient` ranges in [0.0; 0.3], while `tmpAmbient` is in range of [0; 255].

DolceTriade:
I meant in the commit message or PR description.

DolceTriade (approved):

VReaperV:
Oh, sure, I added it to the commit description.

slipher:
Setting it to 0 ought to mean no modification to the light grid. But if I set `r_forceAmbient` to 0 it makes stuff way brighter anyway.

Before r_forceAmbient changes (with any cvar value as it makes no difference):


unvanquished-sokolov-alienbase

With this PR and `r_forceAmbient 0`:

unvanquished-sokolov-alienbase


Also I think the light factor handling should be different. IMO it should target the final light value, not the pre-lightfactor-scaling value. With the max value of .3 something even seems to overflow somehow and it is less bright than with 0.

VReaperV:
> Setting it to 0 ought to mean no modification to the light grid. But if I set `r_forceAmbient` to 0 it makes stuff way brighter anyway.
> 
> Before r_forceAmbient changes (with any cvar value as it makes no difference):

The original code was supposed to set it to the value provided by the map in this case, it was just broken.

On `hangar28` it works as you'd expect when the map doesn't set any value there:
0:
unvanquished_2025-01-23_005938_000
0.3:
unvanquished_2025-01-23_005958_000

Perhaps it should be changed such that this behaviour is used if the cvar is set to -1.0 instead?

> Also I think the light factor handling should be different. IMO it should target the final light value, not the pre-lightfactor-scaling value. With the max value of .3 something even seems to overflow somehow and it is less bright than with 0.

That I agree with.

VReaperV:
I've added both of these changes here.

slipher:
IIUC the worldspawn `_color` or `forceAmbient` keyword is not supposed to do anything unless the map light grid is broken or missing. Is that correct?

VReaperV:
Not entirely.

This seems to be an XreaL feature, and from what I can tell:
If there's no lightGrid, either the `worldspawn` or `r_forceAmbient` value is used:
https://github.com/QuakeEngines/ET-XreaL/blob/7e5bcc9082d85ecf9cc3b29ccbddba21e80d7c06/src/engine/rendererGL3/tr_light.c#L375-L386
If the interpolated grid light average of the three colour channels is < `r_forceAmbient` (for ambient light), then `r_forceAmbient` value is used instead:
https://github.com/QuakeEngines/ET-XreaL/blob/7e5bcc9082d85ecf9cc3b29ccbddba21e80d7c06/src/engine/rendererGL3/tr_light.c#L266-L271

There might be a slight difference in doing this check on lightGrid points rather than the interpolated value, but I don't think it would make much of a difference.

slipher:
> If the interpolated grid light average of the three colour channels is < `r_forceAmbient` (for ambient light), then `r_forceAmbient` value is used instead:
> https://github.com/QuakeEngines/ET-XreaL/blob/7e5bcc9082d85ecf9cc3b29ccbddba21e80d7c06/src/engine/rendererGL3/tr_light.c#L266-L271

Oh right, `R_LightForPoint` is also affected (not that this makes any sense, if you're not going to modify the light grid samples anywhere else). But for the vast majority of maps, which have a valid light grid and no `realLight` particles, nothing should be affected.

VReaperV:
> Oh right, R_LightForPoint is also affected (not that this makes any sense, if you're not going to modify the light grid samples anywhere else).

I think that's because it was originally used for entities as well, but the `r_forceAmbient` use was lost with the GLSL transition.

VReaperV:
Anyway, if there are no further concerns with these changes, I'll merge it a bit later.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -4124,18 +4124,19 @@ void R_LoadLightGrid( lump_t *l )
...
-			VectorSet( tmpAmbient, r_forceAmbient.Get(), r_forceAmbient.Get(), r_forceAmbient.Get() );
-		}
-
 		if ( tr.legacyOverBrightClamping )
 		{
 			R_ColorShiftLightingBytes( tmpAmbient );
 			R_ColorShiftLightingBytes( tmpDirected );
 		}
 
+		const byte forceAmbientNormalised = floatToUnorm8( r_forceAmbient.Get() );
```
slipher, line 4133: The overbright is still not taken into account in non-clamped mode.

src/engine/renderer/tr_init.cpp
```diff
@@ -216,8 +216,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_wolfFog;
 	cvar_t      *r_noFog;
 
-	Cvar::Range<Cvar::Cvar<float>> r_forceAmbient( "r_forceAmbient", "Minimal light amount in lightGrid", Cvar::NONE,
-		0.125f, 0.0f, 0.3f );
+	Cvar::Range<Cvar::Cvar<float>> r_forceAmbient( "r_forceAmbient", "Minimal light amount in lightGrid; -1 to use map value", Cvar::NONE,
+		0.125f, -1.0f, 0.3f );
```
slipher, line 220: The `_color` keyword is used in a lot of extant maps, sometimes with pretty large values. I imagine it must have some meaning for q3map2 which is distinct from this XReal renderer's meaning. Maybe we should remove the `_color` worldspawn key (leaving just `ambientColor`) so that it doesn't mess up those maps.

Should probably be cheat, that's a big cheat to make models always displayed brightly. Maybe a bigger cheat even than r_gamma since you could make only the targets be bright, rather than everything

slipher (commented):
On some maps, with all default settings, this somehow makes the light grid *darker* rather than brighter. Before/after on nano
unvanquished-nano-overmind


unvanquished-nano-overmind

VReaperV:
> On some maps, with all default settings, this somehow makes the light grid _darker_ rather than brighter. Before/after on nano unvanquished-nano-overmind
> 
> unvanquished-nano-overmind

Fixed.

src/engine/renderer/tr_bsp.cpp
VReaperV: Isn't that what the multiplication by `lightFactor` does in the shader?

slipher: If we request minimum light of 0.4 then that should be applied as x = max(0.1, x) in the light grid since it will be multiplied later.

src/engine/renderer/tr_init.cpp
VReaperV: There's this in q3map2:
https://github.com/id-Software/Quake-III-Arena/blob/dbe4ddb10315479fc00086f08e25d968b4b43c49/q3map/lightv.c#L4941
But I can't find any notion of that being used in the game's code.

slipher: Yeah that's what I'm saying. `_color` is intended for the map compiler and interpreting it like that in the renderer seems bad and wrong.

src/engine/renderer/tr_bsp.cpp
VReaperV: 0.1 being `0.4 / tr.mapLightFactor`? Or where does it come from?

slipher: Yes that's right

VReaperV: Well, that gives wrong results.
This pr without the scaling:
unvanquished_2025-01-27_023602_000
Dividing by `tr.mapLightFactor`:
unvanquished_2025-01-27_023954_000
Both are with default overbright values.

VReaperV: If the scaling is done after comparison, it's also wrong:
unvanquished_2025-01-27_023523_000

VReaperV: But beside that, I don't really see where the lightGrid values would be multiplied by anything extra in case of `r_forceAmbient` being < the value in the lightGrid. It's just substituting the values after they're normalised to [0; 1] range.

src/engine/renderer/tr_init.cpp
VReaperV: Removed the check for `_color` key.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -4142,6 +4136,18 @@ void R_LoadLightGrid( lump_t *l )
...
 		}
 
+		const float forceAmbient = r_forceAmbient.Get();
+		if ( ambientColor[0] < forceAmbient &&
+			ambientColor[1] < forceAmbient &&
+			ambientColor[2] < forceAmbient &&
+			/* Make sure we don't change the (0, 0, 0) points because those are points in walls,
+			which we'll fill up by interpolating nearby points later */
+			ambientColor[0] != 0 &&
+			ambientColor[1] != 0 &&
```
slipher, line 4146: Shouldn't these be `||` for the zero check?

src/engine/renderer/tr_init.cpp
slipher: What do you think about the cheat comment?

slipher (commented):
Still getting altered grid lighting for all the maps I test (nano, plat23 and sokolov). Sokolov has the biggest difference (much brighter). I'm still confused why this is happening... reading the code it looks like there should be no difference as long as `ambientColor` and `r_forceAmbient` are not set.

src/engine/renderer/tr_bsp.cpp
VReaperV: Oh, right, fixed now.

VReaperV:
> Still getting altered grid lighting for all the maps I test (nano, plat23 and sokolov). Sokolov has the biggest difference (much brighter). I'm still confused why this is happening... reading the code it looks like there should be no difference as long as `ambientColor` and `r_forceAmbient` are not set.

Any specific viewpos? I checked on Sokolov at spectator spawn after rebasing and fixing `&&` to `||`, looked the same to me compared to 0.55.2.

slipher:
> Any specific viewpos? I checked on Sokolov at spectator spawn after rebasing and fixing `&&` to `||`, looked the same to me compared to 0.55.2.

How are you going to check at the spectator spawn? There need to be players or buildables to see the light grid changes. I'm seeing the differences on buildables as in the screenshots posted upthread.

VReaperV:
Well, I just built the same buildings in that are on this branch and on 0.55.2.

VReaperV:
(it also affects the first-person weapon)

slipher:
If I check out this PR and go to the alien base on Sokolov, it is still a lot brighter.  Closer to the second screenshot than the first in https://github.com/DaemonEngine/Daemon/pull/1510#issuecomment-2606929677

VReaperV:
I get this in alien base on Sokolov:
unvanquished_2025-01-30_024722_000
Do note though that `r_forceAmbient` should be set to -1 to disable it.

slipher:
I see. I assumed -1 is supposed to be the default value. So is that an intentional change, to light everything up so it is easier to see for gameplay purposes or something?

VReaperV:
Not really, I just didn't touch the default value.

slipher:
If that is not intentional then let's set it to -1

VReaperV:
Yeah, that makes sense, added now.

src/engine/renderer/tr_init.cpp
slipher: Interesting idea, maybe this could be an alternative to `r_gamma` that degrades the map's aesthetics less.

VReaperV: Yeah, I was thinking that too.

VReaperV:
I suppose lighting up entities in dark areas with this isn't as useful, now that tonemapping has been merged.

slipher:
> I suppose lighting up entities in dark areas with this isn't as useful, now that tonemapping has been merged.

Just make it CHEAT then? Or NUKE the whole thing...

VReaperV:
Set to cvar cheat now.

slipher:
LGTM

PR #1511, opened 2025-01-22
Clean-up GLBuffer interface, use DSA

VReaperV:
The `GLBuffer` has started to become unwieldy when setting up buffers, so I cleaned up its interface and made it use Direct State Access versions of the buffer commands where possible instead. These commands allow modyfing buffers without actually binding them, which makes it much easier. I've also removed a lot of duplicate code from `GLBuffer` itself, as well as the derivative buffers.

`glBufferData` has been moved to `GLBuffer`, in line with the other buffer functions.

Cleaned up buffer binding points as well, now they use an enum, and the shaders use defines for it.

Moved buffer init/free to a more appropriate location as well.

Also NUKED the unused `GLIndirectBuffer` and moved `GLIndirectCommand` to `MaterialSystem.h`.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -37,18 +37,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "Material.h"
 #include "ShadeCommon.h"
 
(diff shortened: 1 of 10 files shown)


VReaperV:
Rebased, doesn't require #1507 anymore.

slipher:
LGTM, no diffs in my tests

PR #1513, opened 2025-01-22
Fix backEnd.currentVAO

VReaperV:
This was previously getting reset in `R_Init()`, so it was always 0 after a map was loaded. This also meant that the call to `glDeleteVertexArrays()`  in `RE_Shutdown()` was silently erroring out if a map was loaded (or `vid_restart` used).


slipher:
Is it even used for anything?

VReaperV:
> Is it even used for anything?

Yeah, a VAO is required for any vertex specification commands (and by extension the typical rendering commands).

slipher:
LGTM

PR #1514, opened 2025-01-22
Use colorModulateColorGen instead of disabling vertex color

VReaperV:
Originally, vertex color array was disabled if colorGen CGEN_VERTEX or CGEN_ONE_MINUS_VERTEX or alphaGen CGEN_VERTEX or CGEN_ONE_MINUS_VERTEX were used, resulting in the shader receiving the default OpenGL values for the disabled arrays (0.0, 0.0, 0.0, 1.0).

Now it will instead be set via setting a bit in `u_ColorModulateColorGen`, which allows skipping the vertex format change. It will also be necessary for the geometry cache.


VReaperV:
I've added a fix for #1522 here, since this pr changes the relevant code anyway.

src/engine/renderer/gl_shader.h
```diff
@@ -3731,10 +3732,12 @@ class u_ColorModulateColorGen :
...
 
-		if ( needAttrib ) {
-			_shader->AddVertexAttribBit( ATTR_COLOR );
-		} else {
-			_shader->DelVertexAttribBit( ATTR_COLOR );
+		if ( !needAttrib ) {
+			/* Originally, this controlled whether the vertex color array was used,
+			now it does the equivalent by setting the color in the shader in such way as if it was using
+			the default OpenGL values for the disabled arrays (0.0, 0.0, 0.0, 1.0)
+			This allows to skip the vertex format change */
```
slipher, line 3739: Is that related to "Vertex shader in program 11 is being recompiled based on GL state." warnings?

slipher:
LGTM

PR #1515, opened 2025-01-22
Implement geometry cache

VReaperV:
Requires #1511, #1513, and #1514.

The geometry cache represents 3 buffers: a VBO, IBO, and inputVBO. These buffers will replace the old separate input VBOs and IBOs when enabled. This allows binding only a VAO before rendering, rather than setting up vertex pointers with the correct offset and formats every time. Additionally, it means that there's no need to switch between different VBOs/IBOs. The `inputVBO` is currently unused, but it will be needed to make models work with the geometry cache (model animations and world transform will be processed in a compute shader).

This further improves performance and removes the VBO/IBO check for materials.

Currently this is only used by the material system, but it should be possible to make the core renderer use it as well.

This will also be required to support entities with material system.

### DIFF
--- a/src/engine/renderer/GeometryCache.cpp
+++ b/src/engine/renderer/GeometryCache.cpp
@@ -33,6 +33,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 // GeometryCache.cpp
 
+#include "common/Common.h"
+
 #include "GeometryCache.h"
 
 #include "tr_local.h"
@@ -57,9 +59,6 @@ void GeometryCache::FreeGLBuffers() {
 	IBO.DelBuffer();
 
 	VAO.DelVAO();
-}
-
-void GeometryCache::Free() {
 }
 
 void GeometryCache::AllocBuffers() {
@@ -89,7 +88,7 @@ void GeometryCache::AddMapGeometry( const uint32_t verticesNumber, const uint32_
 	for ( const vertexAttributeSpec_t* spec = attrBegin; spec < attrEnd; spec++ ) {
 		vboAttributeLayout_t& attr = VAO.attrs[spec->attrIndex];
 
-		CopyVertexAttribute( attr, *spec, mapVerticesNumber, ( byte* ) VBOVerts );
+		R_CopyVertexAttribute( attr, *spec, mapVerticesNumber, ( byte* ) VBOVerts );
 	}
 	VBO.UnmapBuffer();
 
--- a/src/engine/renderer/GeometryCache.cpp
+++ b/src/engine/renderer/GeometryCache.cpp
@@ -1,0 +1,103 @@
+﻿/*
+===========================================================================
+
+Daemon BSD Source Code
+Copyright (c) 2025 Daemon Developers
+All rights reserved.
+
+This file is part of the Daemon BSD Source Code (Daemon Source Code).
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+	* Redistributions of source code must retain the above copyright
+	  notice, this list of conditions and the following disclaimer.
+	* Redistributions in binary form must reproduce the above copyright
+	  notice, this list of conditions and the following disclaimer in the
+	  documentation and/or other materials provided with the distribution.
+	* Neither the name of the Daemon developers nor the
+	  names of its contributors may be used to endorse or promote products
+	  derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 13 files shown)


src/engine/renderer/GeometryCache.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+void GeometryCache::FreeGLBuffers() {
+	inputVBO.DelBuffer();
+	VBO.DelBuffer();
+	IBO.DelBuffer();
+
+	VAO.DelVAO();
+}
+
+void GeometryCache::Free() {
```
slipher, line 62: Why do we need an empty function?

src/engine/renderer/tr_local.h
```diff
@@ -3715,6 +3716,10 @@ inline bool checkGLErrors()
 
 	============================================================
 	*/
+	uint32_t ComponentSize( GLenum type );
```
slipher, line 3719: Add an `R_` namespace now that they are used beyond one file

src/engine/renderer/gl_shader.h
```diff
@@ -1530,6 +1532,78 @@ class GLAtomicCounterBuffer : public GLBuffer {
...
+	void Bind() {
+		glBindVertexArray( id );
+	}
+
+	void SetAttrs( const vertexAttributeSpec_t* attrBegin, const vertexAttributeSpec_t* attrEnd ) {
+		uint32_t ofs = 0;
+		for ( const vertexAttributeSpec_t* spec = attrBegin; spec < attrEnd; spec++ ) {
+			vboAttributeLayout_t& attr = attrs[spec->attrIndex];
+			ASSERT_NQ( spec->numComponents, 0U );
+			// vbo->attribBits |= 1 << spec->attrIndex;
```
slipher, line 1555: Remove commented code

src/engine/renderer/GeometryCache.cpp
```diff
@@ -0,0 +1,103 @@
...
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+===========================================================================
+*/
+// GeometryCache.cpp
+
+#include "GeometryCache.h"
```
slipher, line 36: Please include the precompiled header `common/Common.h` first in each .cpp file

src/engine/renderer/GeometryCache.cpp
VReaperV: Good question, removed it now.

src/engine/renderer/tr_local.h
VReaperV: Done.

src/engine/renderer/gl_shader.h
VReaperV: Done.

src/engine/renderer/GeometryCache.cpp
VReaperV: Added it.

src/engine/renderer/tr_init.cpp
```diff
@@ -102,6 +103,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	Cvar::Cvar<bool> r_gpuFrustumCulling( "r_gpuFrustumCulling", "Use frustum culling on the GPU for the Material System", Cvar::NONE, true );
 	Cvar::Cvar<bool> r_gpuOcclusionCulling( "r_gpuOcclusionCulling", "Use occlusion culling on the GPU for the Material System", Cvar::NONE, false );
 	Cvar::Cvar<bool> r_materialSystemSkip( "r_materialSystemSkip", "Temporarily skip Material System rendering, using only core renderer instead", Cvar::NONE, false );
+	Cvar::Cvar<bool> r_geometryCache( "r_geometryCache", "Use Geometry Cache", Cvar::NONE, true );
```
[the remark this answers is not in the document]
VReaperV: I've removed the cvar for now, I'll look into making it work with core renderer in some other pr perhaps.

src/engine/renderer/Material.h
```diff
@@ -288,7 +285,10 @@ enum class BufferBind {
 	COMMAND_COUNTERS_ATOMIC = 0,
 	COMMAND_COUNTERS_STORAGE = 4, // Avoid needlessly rebinding buffers
 	PORTAL_SURFACES = 5,
-	DEBUG = 10
+	GEOMETRY_CACHE_INPUT_VBO = 6,
```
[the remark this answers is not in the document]
VReaperV: Right now it will work fine since buffers are still always being bound before the draws/dispatches, and later I'll probably add something to avoid rebinding both old and new-style buffers.

slipher:
LGTM

PR #1516, opened 2025-01-23
cmake: build Saigo arm nexe with -O0 to avoid a bug

illwieckz:
Fixes Unvanquished/Unvanquished#3297:

- https://github.com/Unvanquished/Unvanquished/issues/3297

In the future we may either enable some optimizations from `-O2` or enable `-O2` and disable some specific optimizations to avoid doing a _not-optimized-at-all_ build, but this is enough to make the arm nexe run and then unblock the migration to Saigo.

### DIFF
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -231,6 +231,17 @@ else()
 		endif()
 	endif()
 
+	if (NACL AND USE_NACL_SAIGO AND SAIGO_ARCH STREQUAL "arm")
+		# This should be set for every build type because build type flags
+		# are set after the other custom flags and then have the last word.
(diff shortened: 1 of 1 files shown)


cmake/DaemonFlags.cmake
```diff
@@ -231,6 +231,16 @@ else()
 		endif()
 	endif()
 
+	if (NACL AND USE_NACL_SAIGO AND SAIGO_ARCH STREQUAL "arm")
+		# This should be set for every build type because build type flags
```
slipher, line 235: Please add a link here to the relevant issue

illwieckz: Done.

slipher:
LGTM

PR #1519, opened 2025-01-25
Cvar migration: handle remaining cvars that use `modified` flag

slipher:
Also improved the behavior of `sv_maxclients`.


src/engine/renderer/tr_init.cpp
```diff
@@ -55,8 +55,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_showSmp;
 	cvar_t      *r_skipBackEnd;
 
-	cvar_t      *r_measureOverdraw;
```
VReaperV, line 58: Remove the corresponding field from `backEndCounters_t` too?

slipher: Done.

VReaperV (approved):
LGTM

PR #1520, opened 2025-01-25
Fix bug with 2d drawing GL state

slipher:
The drawing state can get messed up if you call CG_FillRect right before RmlUi draws anything, and almost everything looks wrong with wrong colors and text being drawn as solid boxes.

Regression in 7b210353ce0b73117baeec32521746c0990bd68c.

unvanquished_2025-01-25_094335_000

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -830,9 +830,6 @@ static void RB_SetGL2D()
 	GL_LoadProjectionMatrix( proj );
 	GL_LoadModelViewMatrix( matrixIdentity );
 
-	// TODO: remove this, state is set wherever drawing is done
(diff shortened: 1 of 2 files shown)


VReaperV:
Why does the bitmask contain `~GLS_DEPTHTEST_DISABLE`? Depth write is supposed to be disabled for 2D drawing.

slipher:
> Why does the bitmask contain `~GLS_DEPTHTEST_DISABLE`?

Which bitmask?

> Depth write is supposed to be disabled for 2D drawing.

Yes, I set the GL_DEPTHMASK_TRUE bit to 0 and then prevent it from being changed.

VReaperV:
> Which bitmask?

`glState.glStateBitsMask`

> Yes, I set the GL_DEPTHMASK_TRUE bit to 0 and then prevent it from being changed.

I was incorrect there, so nevermind that.

VReaperV (approved):
This works on my end, LGTM on the code too.

PR #1521, opened 2025-01-25
Fix crash in GenerateWorldMaterials()

slipher:
If the cgame sends some 2D rendering commands in the first frame before calling RenderScene as described in
https://github.com/Unvanquished/Unvanquished/pull/3305, it could cause a crash. Fix this by not setting generatingWorldCommandBuffer prematurely while those commands are executing.


VReaperV (approved):
LGTM

PR #1523, opened 2025-01-26
Add RSF_FORCE_LIGHTMAP

VReaperV:
Cgame-side pr: https://github.com/Unvanquished/Unvanquished/pull/3306

Replaces RSF_NOLIGHTSCALE.

If this flag is used when registering a shader, all `ST_COLORMAP` stages will be changed to `ST_DIFFUSEMAP`. This is intended for use with particles and trails, to make them use the lightGrid in the GLSL code.

Improves performance with particle systems and trails, e. g. with firebombs, and a massive performance improvement on map https://users.unvanquished.net/~sweet/pkg/map-jota_3.0.dpk in the outside area.

There's for the most part no difference in how the particles/trails look, the only differences are due to the GLSL lightGrid being more precise.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5869,6 +5869,16 @@ static shader_t *FinishShader()
 	numStages = MAX_SHADER_STAGES;
 	GroupActiveStages();
 
+	if ( shader.forceLightMap ) {
+		for( size_t stage = 0; stage < numStages; stage++ ) {
+			shaderStage_t* pStage = &stages[numStages];
+
+			if ( pStage->type == stageType_t::ST_COLORMAP ) {
+				pStage->type = stageType_t::ST_DIFFUSEMAP;
+			}
+		}
+	}
+
 	// set appropriate stage information
 	for ( size_t stage = 0; stage < numStages; stage++ )
 	{
@@ -6231,6 +6241,11 @@ shader_t       *R_FindShader( const char *name, shaderType_t type, int flags )
 	if ( flags & RSF_SPRITE )
 	{
 		shader.entitySpriteFaceViewDirection = true;
+	}
+
+	if ( flags & RSF_FORCE_LIGHTMAP )
+	{
+		shader.forceLightMap = true;
 	}
 
 	// attempt to define shader from an explicit parameter file
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -5870,12 +5870,10 @@ static shader_t *FinishShader()
 	GroupActiveStages();
 
 	if ( shader.forceLightMap && numStages == 1 ) {
-		for( size_t stage = 0; stage < numStages; stage++ ) {
-			shaderStage_t* pStage = &stages[stage];
+		shaderStage_t* pStage = &stages[0];
 
-			if ( pStage->type == stageType_t::ST_COLORMAP ) {
-				pStage->type = stageType_t::ST_DIFFUSEMAP;
-			}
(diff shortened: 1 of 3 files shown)


slipher:
FWIW the `realLight` on Jota doesn't even do anything currently. The shader must have an rgbgen `vertex` or `entity` for the entity color to have any effect. There are only 4 maps where `realLight` is actually used with a compatible shader:
- gantry (blood drops from fan)
- sectorb17 (smokestack)
- nova (purports to be smoke; haven't tested)
- `map-edge1_v3_ye_m.pk3` (purports to be smoke; haven't tested)


Haven't read the code yet but the approach sounds fine.

src/engine/renderer/tr_shader.cpp
```diff
@@ -5888,6 +5888,17 @@ static shader_t *FinishShader()
 	numStages = MAX_SHADER_STAGES;
 	GroupActiveStages();
 
+	if ( shader.forceLightMap ) {
+		for( size_t stage = 0; stage < numStages; stage++ ) {
+			shaderStage_t* pStage = &stages[numStages];
+
+			// TODO: Add lightmap as the first stage if there's a heatHaze stage here?
```
slipher, line 5895: The TODO seems unnecessary as this is only needed for compatibility with a few old assets. Hopefully no one should be making new particle systems with `realLight`. (Maybe there should be a deprecation warning.)

src/engine/renderer/tr_shader.cpp
```diff
@@ -5888,6 +5888,17 @@ static shader_t *FinishShader()
...
+			if ( pStage->type == stageType_t::ST_COLORMAP ) {
+				pStage->type = stageType_t::ST_DIFFUSEMAP;
```
slipher, line 5897: I recall a shader in StarATCS that used `realLight` (uselessly, without rgbgen vertex/entity) and also had a lightmap stage. A shader like that would end up with two lightmaps. To avoid this we could restrict the `forceLightMap` effect to shaders with only a single stage, or perhaps do it after `CollapseStages` (so that it wouldn't have the ST_COLORMAP type anymore).

src/engine/renderer/tr_shader.cpp
VReaperV: Yeah, I was thinking of the `heatHaze` particles on `usstremor`, but apparently they don't have `realLight`. Removed the comment.

src/engine/renderer/tr_shader.cpp
VReaperV: That shader ends up with 2 stages anyway.

slipher: True. So what about limiting the effect to shaders with a single stage?

VReaperV: That seems reasonable. I checked all the shaders for particles with `realLight` in the particle directory, and only the `staratcs` one has 2 stages. Without the single stage check it ends up with a lightgrid and a regular lightmap stage, with the check it ends up as generic + lightgrid.

VReaperV: Added the check for the amount of stages.

src/engine/renderer/tr_shader.cpp
```diff
@@ -5869,6 +5869,16 @@ static shader_t *FinishShader()
 	numStages = MAX_SHADER_STAGES;
 	GroupActiveStages();
 
+	if ( shader.forceLightMap && numStages == 1 ) {
+		for( size_t stage = 0; stage < numStages; stage++ ) {
```
slipher, line 5873: The loop is not needed now.

slipher (commented):
LGTM modulo comment

PR #1525, opened 2025-01-26
Make Square() a template function instead of macro

slipher:
Missed that one when converting the vector macros, as noticed by freem


illwieckz (approved):
LGTM.

PR #1526, opened 2025-01-27
tr_bsp: unroll a loop by hand to avoid a GCC compiler bug

### LINKED ISSUE #1518: Building with GCC 13+ and `-msse4 -O3 -ffast-math` breaks the light grid
When I build with GCC 13 and `-march=native` on Zen 2 it breaks the light grid.

I get that when I load the plat23 map:

```
Warn: light grid mismatch, default light grid used 
```

This may be a GCC bug, or maybe we have a shady code relying on some undefined behavior and using other instructions triggers an hidden bug.

System tested:

- System: Ubuntu 24.04.1
- Compiler: GCC 13.3.0

Hardware tested:

- CPU: AMD Ryzen Threadripper PRO 3955WX (Castle Peak, Zen 2)
- CPU: AMD Custom Sony PlayStation 5 CPU, ASRock BC-250 variant (Renoir, Zen2)

I haven't tested other compilers or other systems yet.

illwieckz:
Fixes #1518:

- https://github.com/DaemonEngine/Daemon/issues/1518

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -4026,7 +4026,6 @@ void R_LoadLightGrid( lump_t *l )
 	}
 
 	int            i, j, k;
-	vec3_t         maxs;
 	world_t        *w;
 	float          *wMins, *wMaxs;
 	dgridPoint_t   *in;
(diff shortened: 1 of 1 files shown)


illwieckz:
Don't ask me why it workarounds the bug… 🤷‍♀️️

illwieckz:
Obsoleted by:

- https://github.com/DaemonEngine/Daemon/pull/1529

PR #1528, opened 2025-01-27
Some dynamic buffer-related clean-up and fixes

VReaperV:
- NUKED useless GL_ARB_buffer_storage and GL_ARB_sync macro checks.
- Fix bogus checks for SHADER_MAX_VERTEXES and SHADER_MAX_INDEXES for cases where the dynamic buffers are not being used by adding a check for `r_vboModels`. Also changed it to a new-style cvar.
- Some small clean-up of Ringbuffer functions and more stuff in logs.

### DIFF
--- a/src/engine/renderer/tr_vbo.cpp
+++ b/src/engine/renderer/tr_vbo.cpp
@@ -93,7 +93,6 @@ static uint32_t ComponentSize( GLenum type )
 	Sys::Error( "VBO ComponentSize: unknown type %d", type );
 }
 
-#if defined( GL_ARB_buffer_storage ) && defined( GL_ARB_sync )
 /*
 ============
 R_InitRingbuffer
@@ -102,7 +101,6 @@ R_InitRingbuffer
 static void R_InitRingbuffer( GLenum target, GLsizei elementSize,
 			      GLsizei segmentElements, glRingbuffer_t *rb ) {
 	GLsizei totalSize = elementSize * segmentElements * DYN_BUFFER_SEGMENTS;
-	int i;
 
 	glBufferStorage( target, totalSize, nullptr,
 			 GL_MAP_WRITE_BIT | GL_MAP_PERSISTENT_BIT );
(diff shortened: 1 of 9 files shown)


src/engine/renderer/tr_vbo.cpp
```diff
@@ -125,13 +123,15 @@ static GLsizei R_RotateRingbuffer( glRingbuffer_t *rb ) {
...
+	if ( rb->activeSegment >= DYN_BUFFER_SEGMENTS ) {
 		rb->activeSegment = 0;
+	}
 
-	// wait until next segment is ready in 1 sec intervals
-	while( glClientWaitSync( rb->syncs[ rb->activeSegment ], GL_SYNC_FLUSH_COMMANDS_BIT,
-				 10000000 ) == GL_TIMEOUT_EXPIRED ) {
-		Log::Warn("long wait for GL buffer" );
+	// wait until next segment is ready in 10ms intervals
+	static const GLuint64 TIMEOUT = 10000000;
```
slipher, line 131: No static, just const or constexpr

src/engine/renderer/tr_vbo.cpp
```diff
@@ -125,13 +123,15 @@ static GLsizei R_RotateRingbuffer( glRingbuffer_t *rb ) {
...
-	if( rb->activeSegment >= DYN_BUFFER_SEGMENTS )
+	if ( rb->activeSegment >= DYN_BUFFER_SEGMENTS ) {
 		rb->activeSegment = 0;
+	}
 
-	// wait until next segment is ready in 1 sec intervals
-	while( glClientWaitSync( rb->syncs[ rb->activeSegment ], GL_SYNC_FLUSH_COMMANDS_BIT,
-				 10000000 ) == GL_TIMEOUT_EXPIRED ) {
-		Log::Warn("long wait for GL buffer" );
+	// wait until next segment is ready in 10ms intervals
```
slipher, line 130: Heh, I've seen that warning occasionally. Maybe the timeout should be longer. (non-blocking comment)

slipher (commented):
LGTM modulo comment

src/engine/renderer/tr_vbo.cpp
VReaperV: Hmm, I don't think I've ever actually seen that one. I've got some further in-progress changes to these dynamic VBOs/IBOs to fix `navedit` performance, it might fix the warn too.

src/engine/renderer/tr_vbo.cpp
VReaperV: Fixed.

PR #1529, opened 2025-01-27
Fix moronic light grid array size calculation

### LINKED ISSUE #1518: Building with GCC 13+ and `-msse4 -O3 -ffast-math` breaks the light grid

slipher:
First it calculated the number of points along the negative and positive axes of a dimension of the light grid. Then, instead of just adding those two together to determine the total number of points on the axis, it calculated the coordinates of the minimum and maximum points, subtracted them, and divided by the cell size to get the number of points on the axis. This division was floored, not rounded, so if the difference in coordinates had an error of -0.000001 it would get the wrong answer.

Surprising that this didn't break earlier.
Fixes #1518.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -4026,7 +4026,6 @@ void R_LoadLightGrid( lump_t *l )
 	}
 
 	int            i, j, k;
-	vec3_t         maxs;
 	world_t        *w;
 	float          *wMins, *wMaxs;
 	dgridPoint_t   *in;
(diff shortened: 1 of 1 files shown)


illwieckz (approved):
I confirm it fixes #1518:

- https://github.com/DaemonEngine/Daemon/issues/1518

LGTM.

PR #1530, opened 2025-01-27
Change most of the remaining renderer cvars to new-style

VReaperV:
Based on #1515 because I don't eant to bother with merge conflicts.

Also NUKED unused r_noLightVisCull.


slipher:
Needs rebase

PR #1531, opened 2025-01-28
Use dGPU if both dGPU and iGPU are available on Windows

VReaperV:
Normally, to use the dGPU on Windows 10/11, you need to manually set it for the executable in Windows settings. This will set some registry values to tell the NVidia/AMD drivers to use the dGPU here.

I've confirmed that this works on Win10 with an NVidia dGPU, now just waiting for someone to test it with AMD.

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -26,6 +26,16 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "DetectGLVendors.h"
 #include "Material.h"
 
+#ifdef _WIN32
+	extern "C" {
+		// Use dGPU if both dGPU and iGPU are available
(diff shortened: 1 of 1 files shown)


slipher:
We have a thing for that in the updater: https://github.com/Unvanquished/updater/blob/bd169b5f8539c5ef61cd8af87a1484ff8ebc383d/win.cpp#L322. But this way seems easier; maybe we can NUKE the updater bit.

VReaperV:
> We have a thing for that in the updater: https://github.com/Unvanquished/updater/blob/bd169b5f8539c5ef61cd8af87a1484ff8ebc383d/win.cpp#L322. But this way seems easier; maybe we can NUKE the updater bit.

Hmm, the thing in the updater never worked for me I suppose, since I still had to set it manually.

VReaperV:
I've got confirmation that this works on AMD too.

illwieckz (approved):
LGTM.

PR #1532, opened 2025-01-28
q_math: do tracing in id Software/J.M.P. van Waveren's QuatFromMatrix() implementation like Martin John Baker's one did, fix model distortion with GCC 14

### LINKED ISSUE #1527: Model distortion bug with GCC 14
Splitting #1312 so #1312 only focuses on the Clang bug and can be closed as fixed, and using this thread for the GCC variant.

Quote:

> I remember @DolceTriade said one day he seen bugs while compiling with GCC 14.
>
> I confirm I see bugs affecting models in a weird way when compiling with GCC 14 (no issue with GCC 13):
>
> unvanquished_2024-09-13_181819_000
>
> unvanquished_2024-09-13_181834_000

illwieckz:
Do tracing in [id Software/J.M.P. van Waveren's QuatFromMatrix() implementation](https://web.archive.org/web/20100818052330/http://cache-www.intel.com/cd/00/00/29/37/293748_293748.pdf) like the one we have by Martin John Baker's did, fix model distortion with GCC 14.

Fix #1527:

- https://github.com/DaemonEngine/Daemon/issues/1527

It adds `+1` to the first trace

Fix Unvanquished/Unvanquished#3140:

- https://github.com/Unvanquished/Unvanquished/issues/3140

What's curious is that the Martin John Baker's page says they also removed the `+1`:

> // I removed + 1.0f; [see discussion with Ethan](http://www.euclideanspace.com/maths/geometry/rotations/conversions/matrixToQuaternion/ethan.htm)

But it also says (but we didn't use any epsilon):

> // I changed M_EPSILON to 0

The said “_discussion with Ethan Tira-Thompson_” says:

> Hi Martin, 
>
> I was looking at the matrixToQuaternion page, and I think the "Better code" and "C++ code" sections may have a problem with numerical stability as Tr + 1 may approach 0 in some cases. The test for greater-than-epsilon helps, but a better solution may be found in the implementation from the Wild Magic package by Dave Eberly:
> http://www.geometrictools.com/LibFoundation/Mathematics/Wm4Quaternion.inl
> (See "FromRotationMatrix" function)
>
> There, they compare Tr > 0 (instead of Tr+1 > 0) for the initial conditional, thus avoiding getting close to 0 in the 0.5 / sqrt(Tr + 1) which follows it.
>
> Do you have any thoughts, does this make sense to you? The implications of falling into the 'else' clause are the main concern here, but that seems safe, and appears to work in practice.
>
> Thanks!

Then follows a lengthy discussion about “danger of division by zero or square root of a negative number” and “danger of division by zero or square root of a negative number”, maybe it was safe to rule out those concerns with no fast math but we need to take care of this with fast math?

Actually using that `+1` means one compute is twice the same, so it may even be faster.

On a side note, the document by id Software/J.M.P. van Waveren mentions this:

> The above routine also uses a fast reciprocal square root approximation, referencing works like the Chris Lomont one I quoted there when improving `Q_rsqrt()`:

- https://github.com/DaemonEngine/Daemon/pull/1458

So I made the code also use `Q_rsqrt()`.

### DIFF
--- a/src/engine/qcommon/q_math.cpp
+++ b/src/engine/qcommon/q_math.cpp
@@ -2870,11 +2870,11 @@ void QuatFromMatrix( quat_t q, const matrix_t m )
 	/* For the +1 that deviates from the original implementation,
 	see https://github.com/DaemonEngine/Daemon/issues/1527 */
 
-	float t, s;
+	float t = m[ 0 ] + m[ 5 ] + m[ 10 ] + 1.0f;
+	float s;
 
-	if ( m[ 0 ] + m[ 5 ] + m[ 10 ] + 1.0f > 0.0f )
+	if ( t > 0.0f )
 	{
-		t = m[ 0 ] + m[ 5 ] + m[ 10 ] + 1.0f;
 		s = ( 1.0f / sqrtf( t ) ) * 0.5f;
 
 		q[ 3 ] = s * t;
--- a/src/engine/qcommon/q_math.cpp
+++ b/src/engine/qcommon/q_math.cpp
@@ -2866,9 +2866,13 @@ void QuatFromMatrix( quat_t q, const matrix_t m )
 	 *
 	 *	   http://www.intel.com/cd/ids/developer/asmo-na/eng/293748.htm
 	 */
+
+	/* For the +1 that deviates from the original implementation,
+	see https://github.com/DaemonEngine/Daemon/issues/1527 */
+
 	float t, s;
 
-	if ( m[ 0 ] + m[ 5 ] + m[ 10 ] > 0.0f )
+	if ( m[ 0 ] + m[ 5 ] + m[ 10 ] + 1.0f > 0.0f )
 	{
 		t = m[ 0 ] + m[ 5 ] + m[ 10 ] + 1.0f;
 		s = ( 1.0f / sqrtf( t ) ) * 0.5f;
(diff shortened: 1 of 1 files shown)


illwieckz:
It's also worth noting that the [id Software document](https://web.archive.org/web/20100818052330/http://cache-www.intel.com/cd/00/00/29/37/293748_293748.pdf) also provides some SSE implementations, it may be interesting to look at them.

illwieckz:
So it looks like the same bug was affecting different binaries:

- native cgame (model bug)
- native engine (portal bug)

illwieckz:
So, with `grangermaze` map on `-176 204 203 -44 38` viewpos and a native GCC 14 engine and game build, I can reproduce both bugs:

- https://github.com/DaemonEngine/Daemon/issues/1527
- https://github.com/Unvanquished/Unvanquished/issues/3140

Before:

unvanquished_2025-01-28_115655_000

After:

unvanquished_2025-01-28_115739_000

slipher:
LGTM

I won't dive into the math here since there is really no need to be using the QuatFromMatrix function in either case... For the cgame model rotation we should just use only matrices (without the pointless round trips matrix -> quat -> matrix). The other caller, QuatFromAngles, ought to be constructing the quat directly (there is a commented-out implementation for that) instead of constructing a matrix and converting from a matrix.

PR #1534, opened 2025-01-29
Fix bad return type for Square() template

### LINKED ISSUE #1533: Square function change severely breaks the game
With MSVC on release build with dll cgame/sgame, after pressing forward once:
Image

I've isolated this issue to 96bd0213514e4a49f557a624807ed8bd9825b248.

slipher:
It didn't take into account promotion to integer for types smaller than int.

Fixes #1533.

### DIFF
--- a/src/engine/qcommon/q_shared.h
+++ b/src/engine/qcommon/q_shared.h
@@ -459,7 +459,7 @@ inline void CrossProduct( const vec3_t v1, const vec3_t v2, vec3_t cross )
 }
 
 template<typename A>
-A Square( A a )
+decltype(std::declval<A>() * std::declval<A>()) Square( const A &a )
 {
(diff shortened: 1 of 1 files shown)


VReaperV (approved):
This fixes the issue on myend/haven't found any other regressions. LGTM.

illwieckz:
I'm in favor of merging this kind of fixup as soon as possible, to prevent future bisect runs to be polluted by this bug. We better keep such fixup as close as possible to the commit introducing the bug. The alternative would be to delay other merges to prevent them to slip in between (but we don't do that, and we better not do that).

So @slipher you can merge this as soon as possible, no need to wait for one day.

illwieckz (approved):
LGTM, fixes it for me.

PR #1539, opened 2025-01-31
Engine support for server engine version in client

VReaperV:
Allows showing the engine version the server was built with in server browser etc.


VReaperV:
Closing because #1547 implemented the required functionality.

PR #1541, opened 2025-02-03
Send entire server info string to cgame instead of filtering to certain fields

slipher:
This will let us start using any data we want from the server info without having to make an engine change.

Companion: https://github.com/Unvanquished/Unvanquished/pull/3311


VReaperV (approved):
LGTM

PR #1543, opened 2025-02-03
WIP: implement colorspace support (sRGB, etc.)

illwieckz:
The purpose of this change is to achieve linear blending, this is a continuation of:

Some context:

- Everything should be displayed in sRGB space.
- Everything should be computed in linear space.

Textures are in sRGB space, lightmaps are usually in linear space. So the computation for applying a lightmap on a texture and rendering it is:

- `convertToSRGB( convertFromSRGB( texture ) * light )`

Q3map2 has a trick to also store lightmaps in sRGB space, this makes possible to bias the storage and allocate more precision to some wanted values (makes black less banded if I'm right). When this Q3map2 option is used, the computation for applying a lightmap on a texture and rendering it is:

- `convertToSRGB( convertFromSRGB( texture ) * convertFromSRGB( light ) )`

OpenGL provides features to store images while saying if they are in linear or sRGB space, when using those features, OpenGL would automatically convert the colors to linear space when sampling them, and convert them back to sRGB when displaying.

I still want to start with a fully explicit implementation where we do the conversions ourselves, because it makes easier to control what is happening at every step with our current engine design. It's very unlikely that our engine works with graphics card that don't support sRGB image OpenGL formats (even my 23 years old Radeon 2700 Pro supports it if I'm right). Though, our engine design may not makes it easy to use an implicit OpenGL implementation because of its current design:

- When we load a light style light map, we may not know yet it's a lightmap so we may not know yet in which format it is when we need to know it (when uploading it to GPU).
- When we load an image we may not know yet what kind of image it is, for example a normal map is always in linear space, but some legacy shader keywords make possible to load and upload first the image to GPU memory, then tell if it is a normal map or not only after the OpenGL format is already selected.

There can be other shortcomings like that, but also, there can be things that require some in-engine conversions. For example the rgbgen colors are assumed to be in sRGB space, because if people copy the RGB values from a color picker in an image editor like GIMP or Photoshop, those values are in sRGB. So we need to convert those colors back to linear in the C++ engine code.

Migrating to OpenGL sRGB features is something we can postpone for the future, it will reduce the amount of GLSL code and may limit precision loss (so, more performance and less color imprecision), but we don't require that to get a viable product. And we better want to get proper colorspace management as soon as possible.

My effort to achieve proper colorspace management is now 6 years old.

@sweet235 wrote today in chat:

> \<Sweet\> since almost a year, i cannot decide on what specularity to use for the atcshd texture package
> \<Sweet\> because there is always a change coming up

So we better merge this as soon as possible.

Once this is merged, there would not be _strong_ changes to expect. For example if one day we implement HDR rendering, this will not change the colorspace neither the blend computations, it will just add more precision.

This effort is also required (but not enough) to get true PBR.

It is assumed that once this is merged, the light computation formulæ will be stable (except for PBR which is still work-in-progress).

### DIFF
--- a/src/engine/renderer/gl_shader.h
+++ b/src/engine/renderer/gl_shader.h
@@ -2879,6 +2879,36 @@ public:
 	}
 };
 
+class u_LinearizeTexture :
+	GLUniform1i
+{
+public:
+	u_LinearizeTexture( GLShader *shader ) :
+		GLUniform1i( shader, "u_LinearizeTexture" )
+	{
+	}
+
+	void SetUniform_LinearizeTexture( const int value )
+	{
+		this->SetValue( value );
+	}
+};
+
+class u_DelinearizeScreen :
+	GLUniform1i
+{
+public:
+	u_DelinearizeScreen( GLShader *shader ) :
+		GLUniform1i( shader, "u_DelinearizeScreen" )
+	{
+	}
+
+	void SetUniform_DelinearizeScreen( const int value )
+	{
+		this->SetValue( value );
+	}
+};
+
 class u_ShadowBlur :
 	GLUniform1f
 {
@@ -3907,6 +3937,7 @@ class GLShader_generic :
 	public u_AlphaThreshold,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
+	public u_LinearizeTexture,
 	public u_ColorModulateColorGen,
 	public u_Color,
 	public u_Bones,
@@ -3936,6 +3967,7 @@ class GLShader_genericMaterial :
 	public u_AlphaThreshold,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
+	public u_LinearizeTexture,
 	public u_ColorModulateColorGen,
 	public u_Color,
 	public u_DepthScale,
@@ -3974,6 +4006,7 @@ class GLShader_lightMapping :
 	public u_ViewOrigin,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
+	public u_LinearizeTexture,
 	public u_Bones,
 	public u_VertexInterpolation,
 	public u_ReliefDepthScale,
@@ -4025,6 +4058,7 @@ class GLShader_lightMappingMaterial :
 	public u_ViewOrigin,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
+	public u_LinearizeTexture,
 	public u_ReliefDepthScale,
 	public u_ReliefOffsetBias,
 	public u_NormalScale,
@@ -4268,6 +4302,7 @@ class GLShader_skybox :
 	public u_CloudHeight,
 	public u_UseCloudMap,
 	public u_AlphaThreshold,
+	public u_LinearizeTexture,
 	public u_ModelViewProjectionMatrix
 {
 public:
@@ -4283,6 +4318,7 @@ class GLShader_skyboxMaterial :
 	public u_CloudHeight,
 	public u_UseCloudMap,
 	public u_AlphaThreshold,
+	public u_LinearizeTexture,
 	public u_ModelViewProjectionMatrix {
 	public:
 	GLShader_skyboxMaterial( GLShaderManager* manager );
@@ -4294,6 +4330,7 @@ class GLShader_fogQuake3 :
 	public u_FogMap,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
+	public u_LinearizeTexture,
 	public u_ColorGlobal,
 	public u_Bones,
 	public u_VertexInterpolation,
@@ -4314,6 +4351,7 @@ class GLShader_fogQuake3Material :
 	public u_FogMap,
 	public u_ModelMatrix,
 	public u_ModelViewProjectionMatrix,
+	public u_LinearizeTexture,
 	public u_ColorGlobal,
 	public u_FogDistanceVector,
 	public u_FogDepthVector,
@@ -4330,6 +4368,7 @@ class GLShader_fogGlobal :
 	public u_DepthMap,
 	public u_ModelViewProjectionMatrix,
 	public u_UnprojectMatrix,
+	public u_LinearizeTexture,
 	public u_Color,
 	public u_FogDistanceVector
 {
@@ -4427,7 +4466,8 @@ class GLShader_cameraEffects :
 	public u_ColorModulate,
 	public u_TextureMatrix,
 	public u_ModelViewProjectionMatrix,
-	public u_InverseGamma
+	public u_InverseGamma,
+	public u_DelinearizeScreen
 {
 public:
 	GLShader_cameraEffects( GLShaderManager *manager );
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -49,12 +49,32 @@ vec4 EnvironmentalSpecularFactor( vec3 viewDir, vec3 normal )
 
 // lighting helper functions
 
+#if defined(USE_LIGHT_MAPPING)
+	void TransformLightMap( inout vec3 lightColor, in float lightFactor, bool linearizeLightMap )
+	{
+		convertFromSRGB(lightColor, linearizeLightMap);
+
+		// When doing vertex lighting with full-range overbright, this reads out
+		// 1<<overbrightBits and serves for the overbright shift for vertex colors.
+		lightColor *= lightFactor;
+	}
+#endif
+
 #if defined(USE_GRID_LIGHTING) || defined(USE_GRID_DELUXE_MAPPING)
-	void ReadLightGrid( in vec4 texel, in float lightFactor, out vec3 ambientColor, out vec3 lightColor ) {
+	void ReadLightGrid( in vec4 texel, in float lightFactor, in bool linearizeLightMap, out vec3 ambientColor, out vec3 lightColor) {
 		float ambientScale = 2.0 * texel.a;
 		float directedScale = 2.0 - ambientScale;
 		ambientColor = ambientScale * texel.rgb;
 		lightColor = directedScale * texel.rgb;
+
+		#if defined(r_cheapSRGB)
+			/* The light grid conversion from sRGB is done in C++ code
+			when loading it, meaning it's done before interpolation. */
+		#else
+			convertFromSRGB(lightColor, linearizeLightMap);
+			convertFromSRGB(ambientColor, linearizeLightMap);
+		#endif
+
 		ambientColor *= lightFactor;
 		lightColor *= lightFactor;
 	}
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -67,13 +67,8 @@ vec4 EnvironmentalSpecularFactor( vec3 viewDir, vec3 normal )
 		ambientColor = ambientScale * texel.rgb;
 		lightColor = directedScale * texel.rgb;
 
-		#if defined(r_cheapSRGB)
-			/* The light grid conversion from sRGB is done in C++ code
-			when loading it, meaning it's done before interpolation. */
-		#else
-			convertFromSRGB(lightColor, linearizeLightMap);
-			convertFromSRGB(ambientColor, linearizeLightMap);
-		#endif
+		/* The light grid conversion from sRGB is done in C++ code
+		when loading it, so it's done before interpolation. */
 
 		ambientColor *= lightFactor;
 		lightColor *= lightFactor;
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -202,6 +202,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_rimExponent;
 
 	Cvar::Cvar<bool> r_highPrecisionRendering("r_highPrecisionRendering", "use high precision frame buffers for rendering and blending", Cvar::NONE, true);
+	Cvar::Cvar<bool> r_cheapSRGB("r_cheapSRGB", "use cheap sRGB computation when converting images", Cvar::NONE, false);
 
 	cvar_t      *r_gamma;
 	cvar_t      *r_lockpvs;
@@ -1288,6 +1289,7 @@ ScreenshotCmd screenshotPNGRegistration("screenshotPNG", ssFormat_t::SSF_PNG, "p
 		AssertCvarRange( r_rimExponent, 0.5, 8.0, false );
 
 		Cvar::Latch( r_highPrecisionRendering );
+		Cvar::Latch( r_cheapSRGB );
 
 		r_drawBuffer = Cvar_Get( "r_drawBuffer", "GL_BACK", CVAR_CHEAT );
 		r_lockpvs = Cvar_Get( "r_lockpvs", "0", CVAR_CHEAT );
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -4757,6 +4757,15 @@ static bool ParseShader( const char *_text )
 	return true;
 }
 
+static int packLinearizeTexture( bool linearizeColorMap, bool linearizeMaterialMap, bool linearizeLightMap )
+{
+	/* HACK: emulate three-bits bitfield
+	even: no color map linearization (first bit)
+	less than 2: no light map linearization (second bit)
+	positive: no material map linearization (extra bit) */
+	return ( int(linearizeColorMap) + ( 2 * int(linearizeLightMap) ) ) * ( linearizeMaterialMap ? -1 : 1 );
+}
+
 /*
 ========================================================================================
 
@@ -5298,6 +5307,31 @@ static void FinishStages()
 		// see https://github.com/DaemonEngine/Daemon/issues/376
 		stage->hasHeightMapInNormalMap = stage->hasHeightMapInNormalMap
 			&& ( stage->enableNormalMapping || stage->enableReliefMapping );
+
+		switch ( stage->type )
+		{
+			case stageType_t::ST_COLORMAP:
+			case stageType_t::ST_COLLAPSE_COLORMAP:
+			case stageType_t::ST_SKYBOXMAP:
+				stage->linearizeTexture = tr.worldLinearizeTexture;
+				break;
+			case stageType_t::ST_STYLELIGHTMAP:
+			case stageType_t::ST_STYLECOLORMAP:
+				stage->linearizeTexture = tr.worldLinearizeLightMap;
+				break;
+			case stageType_t::ST_LIGHTMAP:
+				stage->linearizeTexture = packLinearizeTexture( false, false, tr.worldLinearizeLightMap );
+				break;
+			case stageType_t::ST_DIFFUSEMAP:
+			case stageType_t::ST_COLLAPSE_DIFFUSEMAP:
+				stage->linearizeTexture = packLinearizeTexture(
+					tr.worldLinearizeTexture,
+					hasMaterialMap && stage->collapseType == collapseType_t::COLLAPSE_PHONG,
+					tr.worldLinearizeLightMap );
+				break;
+			default:
+				break;
+		}
 
 		// Bind fallback textures if required.
 		if ( !stage->enableNormalMapping && !( stage->enableReliefMapping && stage->hasHeightMapInNormalMap) )
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -4767,7 +4767,9 @@ static int packLinearizeTexture( bool linearizeColorMap, bool linearizeMaterialM
 	even: no color map linearization (first bit)
 	less than 2: no light map linearization (second bit)
 	positive: no material map linearization (extra bit) */
-	return ( int(linearizeColorMap) + ( 2 * int(linearizeLightMap) ) ) * ( linearizeMaterialMap ? -1 : 1 );
+	return linearizeColorMap << 0
+	     | linearizeLightMap << 1
+	     | linearizeMaterialMap << 2;
 }
 
 /*
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -3233,6 +3233,10 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
 		{
 			ParseExpression( text, &stage->deformMagnitudeExp );
 		}
+		else if ( !Q_stricmp( token, "specularSRGB" ) )
+		{
+			stage->specularSRGB = true;
+		}
 		else
 		{
 			Log::Warn("unknown shader stage parameter '%s' in shader '%s'", token, shader.name );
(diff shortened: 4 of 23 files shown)


VReaperV:
> Here is an arachnid2 build, to be used with that branch:
> 
> * https://dl.illwieckz.net/b/unvanquished/pkg/preview/map-arachnid2_1.2.2%2Bsrgb.dpk
> 
> I need help to understand why the creep is so bright:
> 
> unvanquished_2025-02-03_192748_000
> 
> Here is the creep material:
> 
> ```
> gfx/buildables/creep/creep
> {
> 	polygonoffset
> 	twoSided
> 	imageMinDimension 16
> 	{
> 		clamp
> 		diffuseMap gfx/buildables/creep/creep_d
> 		specularMap gfx/buildables/creep/creep_s
> 		normalMap gfx/buildables/creep/creep_n
> 		blendfunc blend
> 		rgbGen identity
> 		alphaGen vertex
> 		alphaFunc GE128
> 	}
> }
> ```
> 
> I guess it should be blended with the lightgrid. Models lit with the lightgrid don't look that wrong.

What happens if you use overbright clamping? That looks like a light factor multiplication bug.

VReaperV:
I also don't really get why you want to store everything in sRGB and do extra conversions (which include `pow`) back and forth instead of just using more precision for textures, which should be enough.

slipher:
> Q3map2 has a trick to also store lightmaps in sRGB space, this makes possible to bias the storage and allocate more precision to some wanted values (makes black less banded if I'm right). When this Q3map2 option is used, the computation for applying a lightmap on a texture and rendering it is:
> 
>     * `convertToSRGB( convertFromSRGB( texture ) * convertFromSRGB( light ) )`

How does that interact with overbright? Does it define some extended sRGB curve that can operate on the range [0, 2^overbrightBits]? Or does it divide by 2^n first and then use standard sRGB?

> * some legacy shader keywords make possible to load and upload first the image to GPU memory, then tell if it is a normal map or not only after the OpenGL format is already selected.

Which ones are those?


> For example the rgbgen colors are assumed to be in sRGB space, because if people copy the RGB values from an color picker in an image editor like GIMP or Photoshop, those values are in sRGB.

On the other hand, it would be kinda weird if stuff like `rgbgen wave` is not in linear space. But we could do it differently depending on which function is used. And even have an `rgbgenColorspace` override if needed.

illwieckz:
Answers to @VReaperV:

> What happens if you use overbright clamping? That looks like a light factor multiplication bug.

The same happen if I disable overbright to begin with. Overbright clamping is not a fix, neither a feature, and precomputed overbright (and then clamping) cannot be used here anyway so it is forcibly disabled when such map is loaded.

> I also don't really get why you want to store everything in sRGB and do extra conversions (which include `pow`) back and forth instead of just using more precision for textures, which should be enough.

Because this is the industry standard. The industry standard here is DarkPlaces/Xonotic, it was DarkPlaces/Xonotic guys who implemented the sRGB stuff in q3map2 (I had nothing to implement in q3map2), and then, the de-facto test maps for the feature are Xonotic maps. Xonotic went in the way to also convert the lightmap to sRGB as a storage precision trick, likely because it makes possible to trick the storage precision while retaining compatibility with in-BSP lightmaps (8-bit per channel array with fixed-size dimension), and then, without breaking the BSP format.

We Unvanquished discourage to use in-BSP lightmaps, but thanks to that trick DarkPlaces/Xonotic people made possible for mappers to just add the single `-sRGB` q3map2 command line option to already used compile commands (using internal lightmaps or not) to both get proper colorspace conversion when raytracing the lightmap, bias the lightmap for precision in some wanted areas, and keep the Q3 BSP format compatibility, all at once.

I would like to see HDR lightmap being implemented, there exist multiple implementations of it in some q3map2 forks, most of them being dead. But this is for the future.

First step is to achieve what DarkPlaces/Xonotic already does, which is the easiest way to achieve because that's their standard pipeline and we have a standard test bed for that in the form of their own assets.

This branch already has some code to decide to convert textures and lightmaps separately, on purpose for a later HDR work that would store lightmaps as linear HDR while having q3map2 still does the conversions when raytracing (q3map2 already have separate options too). But right now, the best we can do is to do the same as DarkPlaces/Xonotic: it is the least effort.

So I dont want to “store everything in sRGB”, I want to do the least effort, have a test bed and another software as reference, so I do the way DarkPlaces/Xonotic does.

illwieckz:
Replying to @slipher:

> > Q3map2 has a trick to also store lightmaps in sRGB space, this makes possible to bias the storage and allocate more precision to some wanted values (makes black less banded if I'm right). When this Q3map2 option is used, the computation for applying a lightmap on a texture and rendering it is:
> > ```
> > * `convertToSRGB( convertFromSRGB( texture ) * convertFromSRGB( light ) )`
> > ```
> 
> How does that interact with overbright? Does it define some extended sRGB curve that can operate on the range [0, 2^overbrightBits]? Or does it divide by 2^n first and then use standard sRGB?

Actually, thank you for aking, because I forgot that patch was initially targeting my overbright implementation (that multiplied at the end of the render pipeline after the light is blended) and then I had to port it to the new overbright implementation (that multiplies before blending the light).

Here is the pseudo-code with overbright:

* `convertToSRGB( convertFromSRGB( texture ) * overbright( convertFromSRGB( light ) ) )`

This would also work, because of multiplication commutativity:

* `convertToSRGB( overbright( convertFromSRGB( texture ) * convertFromSRGB( light ) ) )`

The first one is what is implemented (simpler).

The very important thing to remember is that ALL computations should be done in linear space.

> > * some legacy shader keywords make possible to load and upload first the image to GPU memory, then tell if it is a normal map or not only after the OpenGL format is already selected.
> 
> Which ones are those?

This legacy Doom 3 syntax:

```
textures/castle/brick
{
	{
		map textures/castle/brick_d
		blend diffuseMap
	}
	{
		map textures/castle/brick_n
		blend normalMap
	}
}
```

Or this legacy XreaL syntax:

```
textures/castle/brick
{
	{
		map textures/castle/brick_d
		stage diffuseMap
	}
	{
		map textures/castle/brick_n
		stage normalMap
	}
}
```

But actually, by re-reading `tr_shader.cpp`, it looks like the parsing of `map textures/castle/brick_n` only calls `ParseMap()` and then the call to `LoadMap()` may be delayed to after having parsed `stage normalMap`…


> 
> > For example the rgbgen colors are assumed to be in sRGB space, because if people copy the RGB values from an color picker in an image editor like GIMP or Photoshop, those values are in sRGB.
> 
> On the other hand, it would be kinda weird if stuff like `rgbgen wave` is not in linear space. But we could do it differently depending on which function is used. And even have an `rgbgenColorspace` override if needed.

Yes, that's a problem. We may just assume that they are in linear space. Old maps are supposed to mix different color spaces, but new maps would use linear space. One thing to know is that the q3map2 `-sRGB` option also implies `-sRGBcolor` which means material lines like `q3map_lightRGB` are converted from sRGB, so it would be weird if in the same material the `rgbGen` lines were considered linear…

illwieckz:
Here is an example of such material:

```
textures/shared_vega/squarelight01_blue_1500
{
	qer_editorImage textures/shared_vega_src/squarelight01_blue_p

	// blue #73C0D7
	q3map_lightRGB .6078 .7529 .8431
	q3map_surfacelight 1500

	{
		diffuseMap textures/shared_vega_src/squarelight01_d
		normalMap textures/shared_vega_src/squarelight01_n
		normalFormat -X -Y Z
		specularMap textures/shared_vega_src/squarelight01_s
	}
	{
		map textures/shared_vega_src/squarelight01_a
		blend add
		red .6078
		green .7529
		blue .8431
	}
}
```

We better assume `q3map_lightRGB` and `red`, `green`, `blue` are in the same colorspace.

When `-sRGB` or `-sRGBcolor` is used as q3map2 option, `q3map_lightRGB` converted from sRGB.

illwieckz:
Also, something very bold currently done with my implementation is that specular map is assumed to be in sRGB, of course it's wrong. Specular maps should be in linear space.

But people are doing specular maps from far before people cared about computing lights in linear spaces. Many specular maps may have just been loosely sampled from textures, and then done in sRGB space. And anyway, they were tested with engines doing wrong computations.

Here is the same branch with specular maps assumed to be in sRGB space:

unvanquished_2025-02-04_135240_000

Now the same branch with specular maps assumed to be in linear space:

unvanquished_2025-02-04_135406_000

See how on the first image, the egg looks correct and the overmind not that bad.

See also how on the second image, the egg looks wrong and the overmind looks better.

So it looks like wrongly assuming a specular map is in sRGB space is less problematic than wrongly assuming a specular map is in linear space, that's why as a proof of concept I started with that intentional wrong assumption.

But I guess I will implement some material keyword to tell the format (like we already do with normal maps), so we can assume specular maps are in linear space as they should be, and configure the wrong ones like the egg one to be in linear space.

Actually this kind of material configuration keyword would need a redesign if we want to use OpenGL sRGB formats instead of explicit GLSL code as the `specularMap path/to/image` keyword loads the image before finishing parshig the stage, if I'm right.

This wrong assumption has nothing to do with the creep being too bright anyway.

illwieckz:
So, if I implement a special material keyword to tell the specular map is in sRGB colorspace and use it for the egg, and use the linear space for specular maps by default, here is how that looks:

unvanquished_2025-02-04_162805_000

illwieckz:
But then some specular maps are too strong. I guess some of our texture packs have specular maps in sRGB.

unvanquished_2025-02-04_163040_000

unvanquished_2025-02-04_163054_000

So we will have to add such keyword to whole texture packs.

illwieckz:
The creep is too bright because it it is not linearized (so it gets delinearized at the end, it's like doing gamma 2.2 on it). It is not linearized because the texture is loaded before the map is loaded.

illwieckz:
In the cgame, `trap_R_LoadWorldMap()` is called _before_ `trap_R_RegisterShader("gfx/buildables/creep/creep", RSF_DEFAULT );`, but maybe `trap_R_LoadWorldMap()` is asynchronous…

VReaperV:
> The creep is too bright because it it is not linearized (so it gets delinearized at the end, it's like doing gamma 2.2 on it). It is not linearized because the texture is loaded before the map is loaded, so it is not known yet that it should be linearized.
> 
> ```
> Debug: Uploading image gfx/buildables/creep/creep_n (2048×2048, 1 layers, 0xde1 type, 0x8dbd format) 
> …
> Debug: ----- RE_LoadWorldMap( maps/plat23.bsp ) ----- 
> ```

This doesn't require changing the image data itself, right? Just some `shaderStage_t` data?

illwieckz:
> This doesn't require changing the image data itself, right? Just some `shaderStage_t` data?

Yes.

VReaperV:
> > This doesn't require changing the image data itself, right? Just some `shaderStage_t` data?
> 
> Yes.

What if you just go through all non-UI shaders after loading a map and change those values then?

illwieckz:
I did that instead:

- https://github.com/Unvanquished/Unvanquished/pull/3312

illwieckz:
The `r_rimLighting` feature produces wrong results in linear space. In some way I'm not surprised as it as been likely created for engines not using a linear space.

On the other hand, `r_bloom` works out of the box in linear space, maybe the values were meant for an engine in linear space, but then was broken in our engine, and such brokenness was cancelled by the clamping bug, so maybe it was just not modified to fit the non-linear space.

illwieckz:
I added code to disable bloom when not using a map in linear space (so bloom will be disabled with all maps produced until today).

I also modified rim-lighting to multiply with linearized factors when doing a compute in linear space, I don't know if it's right, but this looks OK.

src/engine/renderer/tr_init.cpp
```diff
@@ -203,6 +203,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	cvar_t      *r_rimExponent;
 
 	Cvar::Cvar<bool> r_highPrecisionRendering("r_highPrecisionRendering", "use high precision frame buffers for rendering and blending", Cvar::NONE, true);
+	Cvar::Cvar<bool> r_cheapSRGB("r_cheapSRGB", "use cheap sRGB computation when converting images", Cvar::NONE, false);
```
slipher, line 206: "cheap" in what sense?

src/engine/renderer/tr_shader.cpp
```diff
@@ -4757,6 +4761,15 @@ static bool ParseShader( const char *_text )
 	return true;
 }
 
+static int packLinearizeTexture( bool linearizeColorMap, bool linearizeMaterialMap, bool linearizeLightMap )
+{
+	/* HACK: emulate three-bits bitfield
```
slipher, line 4766: What's with the bit tests written in a way that's very hard to read instead of just `& 0x2` or whatever? `u_ColorModulate` stuff uses normal bitwise operations.

illwieckz:
> So this doesn't do anything different unless you have a map compiled with sRGB?

It is thought to render legacy maps like they were rendered before.

> I'm afraid it would be a nightmare to design any non-map graphical assets (player and buildable models, weapon effects, etc.) if they have to work with 2 very different blending modes.

Why? When a map is using the linear blending mode, any texture is linearized and delinearized, so for example with a fullbright light it should produce the same. Player, weapon and buildable models use the map light (light grid), in either case. It just happens that with the old way, the light attenuation and things like that are not correct. Like, you can get a very dark shadow in a very lit room in a non-physically correct way. Basically with the old way some contrast is too strong to be real.

> Is there any way we could get linear blending to work with existing maps?

The whole data used for linear blending is done in q3map2 when raytracing the lightmaps/lightgrid before releasing the assets. So no. Here the engine just blends the data computed in q3map2. Existing maps have broken pre-computed data we can't fix afterward.

What is done in engine with that branch and compatible maps is just to make sure the data is in the right colorspace when blending texture and precomputed light, the precomputed light is already generated with either incorrect or correct equations by q3map2. The engine just makes sure the data is correctly processed when q3map2 was using the correct equations. When q3map2 wasn't using the correct equations, the engine just does “as before” for backward compatibility purpose.

The only thing we really have to care about is about real time light, because here we do the job of q3map2 but in real time, hence this thread:

- https://github.com/DaemonEngine/Daemon/issues/1544

src/engine/renderer/tr_shader.cpp
illwieckz: As far as I know bitwise operators are only supported in GLSL starting with GLSL 1.30, so OpenGL 3.0.

VReaperV: I don't know how far back you want to go, but GLSL 1.20 definitely supports bit-wise operators.

illwieckz: OK thanks, GLSL 1.20 is fine (OpenGL 2.1), so let's use them then.

src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -49,12 +49,27 @@ vec4 EnvironmentalSpecularFactor( vec3 viewDir, vec3 normal )
...
+
+		// When doing vertex lighting with full-range overbright, this reads out
+		// 1<<overbrightBits and serves for the overbright shift for vertex colors.
+		lightColor *= lightFactor;
+	}
+#endif
+
 #if defined(USE_GRID_LIGHTING) || defined(USE_GRID_DELUXE_MAPPING)
-	void ReadLightGrid( in vec4 texel, in float lightFactor, out vec3 ambientColor, out vec3 lightColor ) {
+	void ReadLightGrid( in vec4 texel, in float lightFactor, in bool linearizeLightMap, out vec3 ambientColor, out vec3 lightColor) {
```
VReaperV, line 64: `linearizeLightMap` is unused.

src/engine/renderer/gl_shader.h
```diff
@@ -2806,6 +2806,36 @@ class u_ShadowTexelSize :
 	}
 };
 
+class u_LinearizeTexture :
```
VReaperV, line 2809: As I said in #1034, this should go into `u_ColorModulateColorGen`, instead of creating multiple bit-fields.

illwieckz: Yep, I can investigate that later, I'm currently focusing on getting everything rendering correctly first.

slipher:
> > I'm afraid it would be a nightmare to design any non-map graphical assets (player and buildable models, weapon effects, etc.) if they have to work with 2 very different blending modes.
> 
> Why? When a map is using the linear blending mode, any texture is linearized and delinearized, so for example with a fullbright light it should produce the same. Player, weapon and buildable models use the map light (light grid), in either case.

I don't think you can count on anything working both ways except single-stage opaque shaders. Anything with translucency (there is a lot of that with flames, explosions, projectiles etc.) will be affected, and any multi-stage shader will be affected by blending changes.

slipher:
What about `r_halfLambertLighting`? That may be another part of the universe of hacks to work around sRGB-unaware rendering pipelines.

slipher:
As an example of potential problems with non-map assets in a new color space, the rifle impact dust puff is too bright and looks bad. The shader `gfx/weapons/rifle/puff` has 3 stages so it may be more affected by blending differences.


unvanquished_2025-02-07_223950_000

illwieckz:
> What about `r_halfLambertLighting`? That may be another part of the universe of hacks to work around sRGB-unaware rendering pipelines.

Yes, very likely. In fact even `q3map2` has the `-lightanglehl 1` to compile the map with it. 😅️ Though we only use [half lambert lighting](https://developer.valvesoftware.com/wiki/Half_Lambert) in Dæmon engine for models. We may test without. 🙂️

slipher:
I pushed a couple of commits to the branch [slipher/srgb-hax](https://github.com/DaemonEngine/Daemon/tree/slipher/srgb-hax) for making it easier to test:
- `srgb: don't depend on loading map before or after textures`
- `srgb u_LinearizeTexture - use normal bitwise ops`

illwieckz:
Just to explain why our current maps are bogus. For simplicity I will only consider light emitting shaders.

The color of light are sampling the light emitting shader textures, and the color of bounced lights are also sampling the color of the texture the light bounces on.

So to simplify a lot, the lightmap is computed by doing `light texture * intensity`. This is done in `q3map2`.

Then we apply the computed light on diffuse textures, we do that by doing `diffuse_texture * lightmap`.
This is like doing `diffuse_texture * (light_texture * intensity)`.

A correct computation and display should be:

```c
// q3map2
linear_lightmap = linear_light_texture * linear_light_intensity;
// dæmon
linear_screen = linear_diffuse_texture * linear_lightmap;
srgb_screen = convertToSRGB(linear_screen);
display(srgb_screen);
```

But we do instead:

```c
// q3map2
bogus_lightmap = srgb_light_texture * linear_light_intensity;
// dæmon
bogus_screen = srgb_diffuse_texture * bogus_lightmap);
display(bogus_screen);
```

So it's like doing:

```c
// gimp
srgb_diffuse_texture = convertToSRGB(linear_diffuse_texture);
srgb_light_texture = convertToSRGB(linear_light_texture);
// q3map2
bogus_lightmap = srgb_light_texture * linear_light_intensity;
// dæmon
bogus_screen = srgb_diffuse_texture * bogus_lightmap;
display(bogus_screen);
```

People using with `r_gamma 2.2` are somewhat doing:

```c
// gimp
srgb_light_texture = convertToSRGB(linear_light_texture);
srgb_diffuse_texture = convertToSRGB(linear_diffuse_texture);
// q3map2
bogus_lightmap = srgb_light_texture * linear_light_intensity;
// dæmon
bogus_screen = srgb_diffuse_texture * bogus_lightmap;
bogus_screen = cheapConvertFromSRGB(bogus_screen);
display(bogus_screen);
```

So right now, any post processing done (bloom, tonemapping…) is done this way:

```c
// gimp
srgb_light_texture = convertToSRGB(linear_light_texture);
srgb_diffuse_texture = convertToSRGB(linear_diffuse_texture);
// q3map2
bogus_lightmap = srgb_light_texture * linear_light_intensity;
// dæmon
bogus_screen = srgb_diffuse_texture * bogus_lightmap;
bogus_screen = postprocess(bogus_screen);
display(bogus_screen);
```

So here are two different things we can do, that are both correct:

```c
// gimp
srgb_light_texture = convertToSRGB(linear_light_texture);
srgb_diffuse_texture = convertToSRGB(linear_diffuse_texture);
// q3map2
linear_light_texture = convertFromSRGB(srgb_light_texture);
linear_lightmap = linear_light_texture * linear_light_intensity;
// daemon
linear_diffuse_texture = convertFromSRGB(srgb_diffuse_texture);
linear_screen = linear_diffuse_texture * linear_lightmap;
srgb_screen = convertToSRGB(linear_screen);
display(srgb_screen);
```

```c
// gimp
srgb_light_texture = convertToSRGB(linear_light_texture);
srgb_diffuse_texture = convertToSRGB(linear_diffuse_texture);
// q3map2
linear_light_texture = convertFromSRGB(srgb_light_texture);
linear_lightmap = linear_light_texture * linear_light_intensity;
srgb_lightmap = convertToSRGB(linear_lightmap);
// daemon
linear_lightmap = convertFromSRGB(srgb_lightmap);
linear_diffuse_texture = convertFromSRGB(srgb_diffuse_texture);
linear_screen = linear_diffuse_texture * linear_lightmap;
srgb_screen = convertToSRGB(linear_screen);
display(srgb_screen);
```

The second one is what Xonotic does, this branch supports both.

To get the first computation, one can compile a map with `-sRGBcolor -sRGBtex`.
To get the second computation, one can compile a map with `-sRGB` or `-sRGBcolor -sRGBtex -sRGBlight`.
They both produce the same visual results in game, but the second one biases the precision for wanted values while retaining the 8-bit storage.

Most postprocessing operations are probably expecting `linear_screen` as input, but some may have been written while expecting `srgb_screen` as input, or more precisely, `bogus_screen`.

For example right now the colorgrading feature is expecting `bogus_screen` as input and assumes that bogus data is in srgb space, and it produces something the engine displays assuming it is already in srgb space.

illwieckz:
Here are some comparison with `edge-one`. Bloom without linear light computation is just shown for knowledge, it's expected to be broken.

||bloom off|bloom on
|-|-|-
|legacy, obb2|unvanquished_2025-02-12_170722_000|unvanquished_2025-02-12_170756_000
|colorspace, obb2|unvanquished_2025-02-12_211204_000|unvanquished_2025-02-12_211229_000
|colorspace, obb3|unvanquished_2025-02-12_211725_000|unvanquished_2025-02-12_211753_000

“legacy” is our current engine, like Quake3 always did, but with complete overbright.
“colorspace” is that branch. I like the look of a map that is very sunny so I also done overbrightbits 3 just for the look of it.

As we see bloom only works with this branch.

VReaperV:
So on those screenshots the only thing missing is interpretation of textures/colours in q3map2 as sRGB, right?

illwieckz:
> So on those screenshots the only thing missing is interpretation of textures/colours in q3map2 as sRGB, right?

No, q3map2 option `-sRGB` was used to rebuild the map so textures and colours were linearized before computing the lightmap. Q3map2 was aware that textures and colours were in sRGB space and did the conversions.

illwieckz:
For the `legacy`, I used the usual build we currently play and ran Unvanquished 0.55.2.

For `colorspace`, I rebuilt the map with `q3map2 -sRGB` and I ran this branch.

VReaperV:
Oh, I see. Is this fixed version of the map uploaded anywhere?

illwieckz:
> Oh, I see. Is this fixed version of the map uploaded anywhere?

Here it is: https://dl.illwieckz.net/b/unvanquished/pkg/preview/map-edge-one_1.0+srgb.dpk

Note that the lightmap is in sRGB space too (like Xonotic does), but this branch does the conversion.

PR #1546, opened 2025-02-08
Fix shader parse error with `\r` new-lines

VReaperV:
Fixes a shader stage parsing error on https://users.unvanquished.net/~sweet/pkg/map-monk_0%2B0.dpk.


slipher:
`gantry` is another map exhibiting this issue.

Counting both `\r` and `\n` as newlines will mess up the line counter for other parsers for files using CRLF. How about replacing `\r` with `\n` in COM_Compress instead? That function is used only for shaders, and the shader parser doesn't use line numbers.

VReaperV:
That makes sense, I changed this to modify `COM_Compress()` now instead.

slipher:
LGTM

PR #1548, opened 2025-02-10
Fix alpha = 1 colorMod

VReaperV:


VReaperV:
Any that don't use vertex-based `rgbGen`/`alphaGen`, since it's set to 32 here:
https://github.com/DaemonEngine/Daemon/blob/55bf48108af139f7c5368258e392c0ad2e53e49c/src/engine/renderer/gl_shader.h#L3643

slipher:
LGTM

Issue #1551, opened 2025-02-12
Weird dæmon compiler string transformation when compiler has `linux` in its name


illwieckz:
The compilation command line does `-DDAEMON_CXX_COMPILER_STRING=GCC_12.2.0:arm-linux-gnueabihf-g++` but the engine is printing this:

```
Unvanquished 0.55.2 Linux armhf (GCC_12.2.0:arm-1-gnueabihf-g++) Feb 12 2025
```

See how `arm-linux-gnueabihf-g++` is transformed into `arm-1-gnueabihf-g++`.

I don't know what's happening there.

illwieckz:
```
$ echo '' | /usr/bin/arm-linux-gnueabihf-g++ -dM -E - | grep linux
#define __linux 1
#define __gnu_linux__ 1
#define linux 1
#define __linux__ 1
```

So, the compile sets some `linux` definition to `1`, but then, how can a definition be expanded into another definition?

illwieckz:
Well, this is likely affecting any compiler having `linux` in its name:

```
Unvanquished 0.55.2 Linux arm64 (GCC_12.2.0:aarch64-1-gnu-g++) Feb 12 2025
```

slipher:
As seen in #1497 it is risky to pass aribitrary data through command-line defines.  Should generate a file like the GLSL embedded data

illwieckz:
Fixed by:

- https://github.com/DaemonEngine/Daemon/pull/1553

PR #1553, opened 2025-02-12
generate a DaemonBuildInfo.cpp file instead of using compiler definitions

illwieckz:
Generate a builtin.cpp file instead of using compiler definitions.

It produces those kind of files:

- `builtin/builtin.cpp`

```c++
// Automatically generated, do not modify!
const char* DAEMON_ARCH_STRING="amd64";
const char* DAEMON_NACL_ARCH_STRING="amd64";
const char* DAEMON_C_COMPILER_STRING="GCC_13.3.0:x86_64-linux-gnu-gcc";
const char* DAEMON_CXX_COMPILER_STRING="GCC_13.3.0:x86_64-linux-gnu-g++";
const char* DAEMON_OPENGL_ABI_STRING="LEGACY";
```

- `builtin/builtin.h`

```c++
// Automatically generated, do not modify!
extern const char* DAEMON_ARCH_STRING;
extern const char* DAEMON_NACL_ARCH_STRING;
extern const char* DAEMON_C_COMPILER_STRING;
extern const char* DAEMON_CXX_COMPILER_STRING;
extern const char* DAEMON_OPENGL_ABI_STRING;
```

Needed by:

- https://github.com/Unvanquished/Unvanquished/pull/3317

Fixes:

- https://github.com/DaemonEngine/Daemon/issues/1551

Would have fixed (if a workaround wasn't already implemented):

- https://github.com/DaemonEngine/Daemon/pull/1497
- https://github.com/DaemonEngine/Daemon/pull/1500

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,6 @@ if (NOT (BUILD_CLIENT OR BUILD_TTY_CLIENT OR BUILD_SERVER OR BUILD_DUMMY_APP))
     message(NOTICE "You can safely ignore the following reported architecture, it is not used.")
     message(NOTICE "You can safely ignore the following reported compilers, they are unused.")
 endif()
-include(DaemonPlatform)
 
 if (Daemon_OUT)
     set(CMAKE_CURRENT_BINARY_DIR ${Daemon_OUT})
@@ -85,6 +84,9 @@ if (Daemon_OUT)
     set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${Daemon_OUT})
     set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${Daemon_OUT})
 endif()
+
+include(DaemonBuildInfo)
+include(DaemonPlatform)
 
 ################################################################################
 # Configuration options
@@ -605,8 +607,9 @@ if (BUILD_CLIENT)
         message(STATUS "Found OpenGL ABI: ${OpenGL_selected_ABI}")
         message(STATUS "Found OpenGL library: ${OpenGL_selected_LIBRARY}")
 
-        add_definitions("-DDAEMON_OpenGL_ABI=${OpenGL_selected_ABI}")
-        add_definitions("-DDAEMON_OpenGL_ABI_${OpenGL_selected_ABI}")
+        daemon_add_buildinfo("char*" "DAEMON_OPENGL_ABI_STRING" "\"${OpenGL_selected_ABI}\"")
+        add_definitions("-DDAEMON_OPENGL_ABI_${OpenGL_selected_ABI}")
+        add_definitions("-DDAEMON_OPENGL_ABI")
     endif()
 endif()
 
@@ -655,7 +658,8 @@ endif()
 
 if (BUILD_CLIENT OR BUILD_TTY_CLIENT OR BUILD_SERVER OR BUILD_DUMMY_APP)
     if (NACL_RUNTIME_PATH)
-        add_definitions("-DNACL_RUNTIME_PATH=${NACL_RUNTIME_PATH}")
+        daemon_add_buildinfo("char*" "DAEMON_NACL_RUNTIME_PATH_STRING" "\"${NACL_RUNTIME_PATH}\"")
+        add_definitions("-DDAEMON_NACL_RUNTIME_PATH")
     endif()
 
     # FindLocale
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -605,8 +607,9 @@ if (BUILD_CLIENT)
         message(STATUS "Found OpenGL ABI: ${OpenGL_selected_ABI}")
         message(STATUS "Found OpenGL library: ${OpenGL_selected_LIBRARY}")
 
-        add_definitions("-DDAEMON_OpenGL_ABI=${OpenGL_selected_ABI}")
+        daemon_add_builtin("char*" "DAEMON_OpenGL_ABI_STRING" "\"${OpenGL_selected_ABI}\"")
```
slipher, line 610: Use all caps for consistency?

src/engine/framework/VirtualMachine.cpp
```diff
@@ -235,7 +235,7 @@ static std::pair<Sys::OSHandle, IPC::Socket> CreateNaClVM(std::pair<IPC::Socket,
 	CheckMinAddressSysctlTooLarge();
 	const std::string& libPath = FS::GetLibPath();
 #ifdef NACL_RUNTIME_PATH
```
slipher, line 237: That's not right since the new definition does not use a macro.

cmake/DaemonBuiltin.cmake
```diff
@@ -0,0 +1,29 @@
+set(DAEMON_BUILTIN_HEADER "// Automatically generated, do not modify!\n")
+set(DAEMON_BUILTIN_C "${DAEMON_BUILTIN_HEADER}")
+set(DAEMON_BUILTIN_H "${DAEMON_BUILTIN_HEADER}")
+
+macro(daemon_add_builtin TYPE NAME VALUE)
+	set(DAEMON_BUILTIN_C "${DAEMON_BUILTIN_C}const ${TYPE} ${NAME}=${VALUE};\n")
+	set(DAEMON_BUILTIN_H "${DAEMON_BUILTIN_H}extern const ${TYPE} ${NAME};\n")
+endmacro()
+
+set(DAEMON_BUILTIN_DIR "${CMAKE_CURRENT_BINARY_DIR}/builtin")
```
slipher, line 10: It might be nice if `builtin.cpp` etc. were not a singleton... so we could have a Daemon one and an Unvanquished one with other information e.g. an Unvanquished git revision. Non-blocking comment.

slipher (commented):
The name "builtin" is not very meaningful. How about BuildInfo?

If I'm not mistaken, this will cause all files to be rebuilt every time cmake runs because when cmake runs it will re-create the generated header, which is included by all files. We should avoid this, for example by writing the header by hand instead of generating it, or only including the header where it is used.

CMakeLists.txt

src/engine/framework/VirtualMachine.cpp
illwieckz: Fixed.

cmake/DaemonBuiltin.cmake
illwieckz: Yes but that would make it more complicated. One good news is that VM buildt as subproject already use their own `builtin.cpp` file. We can split it later if we face the need for it.

slipher:
> The name "builtin" is not very meaningful. How about BuildInfo?

Looks like this is the only thing left to address.

illwieckz:
> > The name "builtin" is not very meaningful. How about BuildInfo?
> 
> Looks like this is the only thing left to address.

I renamed to `DaemonBuildInfo`, I also made sure that all those static variables are prefixed with `DAEMON_` for consistency (the same way they are now all uppercase).

cmake/DaemonBuildInfo.cmake
```diff
@@ -0,0 +1,29 @@
...
+endmacro()
+
+set(DAEMON_BUILDINFO_DIR "${CMAKE_CURRENT_BINARY_DIR}/DaemonBuildInfo")
+set(DAEMON_BUILDINFO_CPP_FILE "${DAEMON_BUILDINFO_DIR}/DaemonBuildInfo.cpp")
+set(DAEMON_BUILDINFO_H_FILE "${DAEMON_BUILDINFO_DIR}/DaemonBuildInfo.h")
+
+file(MAKE_DIRECTORY "${DAEMON_BUILDINFO_DIR}")
+include_directories("${DAEMON_BUILDINFO_DIR}")
+
+set(BUILTINLIST "${DAEMON_BUILDINFO_CPP_FILE}" "${DAEMON_BUILDINFO_H_FILE}")
```
slipher, line 17: This one remains to be renamed

slipher (commented):
LGTM modulo comment

PR #1555, opened 2025-02-14
cmake: refactor game build configuration to reuse the same code in parent build or subproject build and mutualize code between vm type build

illwieckz:
Refactor game build configuration to reuse the same code in parent build or subproject build and mutualize code between vm type build.

Extracted from:

- https://github.com/DaemonEngine/Daemon/pull/1484

This one brings nothing new, it just refactors the code in a better way so it makes future changes easier.

Amont various things,

- it makes building vm subproject independent from NaCl
- it deduplicates and mutualises the code to avoid maintaining multiple copy-paste

### DIFF
--- a/cmake/DaemonGame.cmake
+++ b/cmake/DaemonGame.cmake
@@ -106,173 +106,170 @@ endif()
 
 daemon_write_buildinfo("Game")
 
+function(buildGameModule module_slug)
+	set(module_target "${GAMEMODULE_NAME}-${module_slug}")
+
+	set(module_target_args "${module_target}" ${PCH_FILE} ${GAMEMODULE_FILES} ${SHAREDLIST_${GAMEMODULE_NAME}} ${SHAREDLIST} ${BUILDINFOLIST} ${COMMONLIST})
+
+	if (module_slug STREQUAL "native-dll")
+		add_library(${module_target_args})
+		set_target_properties(${module_target} PROPERTIES
+			PREFIX ""
+			COMPILE_DEFINITIONS "BUILD_VM_IN_PROCESS")
+	else()
+		add_executable(${module_target_args})
+	endif()
+
+	set_target_properties(${module_target} PROPERTIES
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -164,15 +164,16 @@ set(DEFAULT_NACL_VM_INHERITED_OPTIONS
     USE_COMPILER_INTRINSICS
     USE_DEBUG_OPTIMIZE
     USE_HARDENING
+    USE_LTO
```
slipher, line 167: Not needed now

CMakeLists.txt
```diff
@@ -153,7 +153,7 @@ endif()
 
 # Game VM modules are built with a recursive invocation of CMake, by which all the configuration
 # options are lost, except ones we explicitly choose to pass.
-set(DEFAULT_NACL_VM_INHERITED_OPTIONS
+set(DEFAULT_VMS_INHERITED_OPTIONS
```
slipher, line 156: Can we drop the variable renaming changes (`VM_INHERITED_OPTIONS`, `VMS_PROJECTS`) that force Unvanquished CMakeLists.txt changes? Given than NaCl remains the only relevant type of VM for the near future, the current names seem clearer as to the variables' purpose. And the annoying mystery errors due to un-updated submodules are not really worth it for a minor refactoring change like this.

illwieckz: I have dropped the renaming, this can be postponed for later.

illwieckz: It means only this PR is now needed (not the game one).

slipher: The variable NACL_VMS_PROJECTS which is used by the other CMakeLists also needs to preserve its original name

CMakeLists.txt
illwieckz: Removed.

CMakeLists.txt
illwieckz: Renamed back.

illwieckz:
There is no need for game-side patch anymore (which is now closed):

- https://github.com/Unvanquished/Unvanquished/pull/3318

This PR is standalone.

illwieckz:
Now rebased.

slipher:
LGTM

PR #1556, opened 2025-02-14
cmake: use -fPIC when compiling all object files, but only -pie when linking executables

illwieckz:
Use `-fPIC` when compiling all object files, but only `-pie` when linking executables.

### DIFF
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -68,6 +68,11 @@ macro(set_c_cxx_flag FLAG)
     set_c_flag(${FLAG} ${ARGN})
     set_cxx_flag(${FLAG} ${ARGN})
 endmacro()
+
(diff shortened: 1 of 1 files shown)


illwieckz:
@slipher I removed the `if (NOT APPLE)` for the `-pie` flag. Setting the `-pie` flag in all cases should not have worked to begin with, so I assume macOS was just more picky about that obvious error. Can you test it on macOS?

slipher:
I tested the Mac build with USE_HARDENING on. It builds but I still get the warning when linking an executable:
```
[100%] Linking CXX executable ../daemonded
clang: warning: argument unused during compilation: '-pie' [-Wunused-command-line-argument]
```

illwieckz:
Oh I see… I re-added the `if (NOT APPLE)` test then.

slipher:
LGTM

Issue #1535, opened 2025-01-29
Some information for bindless texture workaround on AMD drivers


VReaperV:
I'm just putting this here so it's not forgotten, as I haven't been able to fully make it work.

Render targets (i. e. the textures bound to FBOs) - leave as non-bindless. When doing that on Adrenaline driver - a few surfaces are visible, but most are still black. Main menu works fine, including on release, since it does not use FBOs other than the default one, and doesn't sample from render targets.

For material system, `processSurfaces_cp` (irrelevant text omitted):
```
 156: #if defined(HAVE_KHR_shader_subgroup_basic) && defined(HAVE_KHR_shader_subgroup_arithmetic)\
 157: 	&& defined(HAVE_KHR_shader_subgroup_ballot) && defined(HAVE_ARB_shader_atomic_counter_ops)
 158: 	#define HAVE_processSurfaces_subgroup
 159: #endif
 160: 
 161: void AddDrawCommand( in uint commandID, in uvec2 materialID ) {
 162: 	SurfaceCommand command = surfaceCommands[commandID + u_SurfaceCommandsOffset];
 163: 
 164: 	#if defined(HAVE_processSurfaces_subgroup)
 165: 		const uint count = subgroupBallotBitCount( subgroupBallot( command.enabled ) );
 166: 		// Exclusive scan so we can determine the offset for each lane without any synchronization
 167: 		const uint subgroupOffset = subgroupExclusiveAdd( command.enabled ? 1u : 0u );
 168: 		
 169: 		uint atomicCmdID = 0u;
 170: 		// Once per subgroup
 171: 		if( subgroupElect() ) {
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR: 0:171: 'assign' :  cannot convert from ' const float' to ' temp uint'

 172: 			atomicCmdID = atomicCounterAddARB( atomicCommandCounters[materialID.x
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR: 0:172: 'atomicCounterAddARB' : no matching overloaded function found 

^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR: 0:172: '' : missing #endif 

[...]

PR #1558, opened 2025-02-21
Workaround for atomic ops on AMD

VReaperV:
Fixes one of the issues in #1535.

Adrenaline/OGLP drivers erroneously disregard the ARB extension functionality (which has different naming than the core one) with `#version 460 core`, which is the one we end up using on such drivers, because it's the context version that the driver provides.


slipher:
LGTM, not tested

Issue #29, opened 2017-09-09
HDR lighting


illwieckz:
I just noticed yesterday that XreaL had HDR lighting, and gh's wip tremulous has it too. That's something I would like to see in Unvanquished since ages but I didn't know the code was already there somewhere. What is preventing us to get that effect in Dæmon?

DolceTriade:
We had it too, but it was broken. We nuked it as a result. GH WIP one is prolly either ported from TremFusion (most likely. I think gimhael was the one who implemented it there...) or XreaL

PR #1559, opened 2025-02-21
Implement adaptive exposure

VReaperV:
Requires #1550 and #1558
Implements #29

This builds up on #1550 by adding automatic exposure based on scene brightness.

When `r_tonemapAdaptiveExposure` is enabled, a compute shader gets the geometric mean of luminance before tonemapping. Then in the tonemapper this value is mapped to a curve with coefficients chosen to try to avoid sudden changes and too high/too low exposure values, and multiplied with the HDR colour.

An example of this in practice: https://users.unvanquished.net/~reaper/adaptiveLighting/adaptiveLightingGlobal.mp4

Also moved the material stuff from `common_cp` to `material_cp`. This wasn't an issue prior because only the material system used compute shaders, but now this would result in shaders failing to compile if `r_materialSystem` is disabled (and using macros would result in unnecessary shader re-compilations).


VReaperV:
I'll see if I can add a non-compute path later. The compute one can also easily give histogram information for https://github.com/Unvanquished/Unvanquished/issues/221.

VReaperV:
I've put some of these changes into #1563 to make this pr lighter.

VReaperV:
Hmm, I might've broken something with the more recent push, the changes were pretty mild when I was testing it.

VReaperV:
Rebased + fixed some issues. That should hopefully fix the drastic exposure changes as seen on the above screenshot too.

sweet235:
LGTM

slipher:
I found another buggy place with huge exposure changes between spots that look about the same in mxl-school.

unvanquished_2025-03-15_013605_000
unvanquished_2025-03-15_013607_000

slipher (commented):
With my laptop's Nvidia card, taking a screenshot goes from no noticeable delay to a 2-second delay when adaptive exposure is enabled.

Any plans to smooth out the exposure changes over time, like how the eye adapts to light maybe? For now the exposure variation seems too large and rapid to be really usable; all world surfaces are flickering when you move.

Doors feel kind of buggy in general. There is often a bright flash when the door is opening. But maybe that's just the same thing as the preceding paragraph.

VReaperV:
> I found another buggy place with huge exposure changes between spots that look about the same in mxl-school.

I miscalculated the max value for luminosity sum, so it does overflow sometimes. Will add a fix later.

slipher:
Also the instant exposure adjustments mess up other effects that are supposed to cause rapid changes in brightness. The flickering of the firelight in `metro-b1-2` alien base is almost not visible at all, being canceled out by exposure changes. When a human weapon is fired, the whole scene perversely darkens due to the luminance of the muzzle flash.

VReaperV:
> With my laptop's Nvidia card, taking a screenshot goes from no noticeable delay to a 2-second delay when adaptive exposure is enabled.

Hmm, it might be best to improve the screenshotting in some way. I don't think there's much that can be done with this from the adaptive exposure side.

> Any plans to smooth out the exposure changes over time, like how the eye adapts to light maybe? For now the exposure variation seems too large and rapid to be really usable; all world surfaces are flickering when you move.

Yeah, I plan to add some exponential smoothing there.

> Doors feel kind of buggy in general. There is often a bright flash when the door is opening. But maybe that's just the same thing as the preceding paragraph.

It might be due to the overflow too.

> Also the instant exposure adjustments mess up other effects that are supposed to cause rapid changes in brightness. The flickering of the firelight in metro-b1-2 alien base is almost not visible at all, being canceled out by exposure changes. When a human weapon is fired, the whole scene perversely darkens due to the luminance of the muzzle flash.

Yeah, smoothing it over time should make it better, maybe some value tweaking too.

VReaperV:
> I found another buggy place with huge exposure changes between spots that look about the same in mxl-school.
> 
> unvanquished_2025-03-15_013605_000 unvanquished_2025-03-15_013607_000

This should be fixed now.

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
```diff
@@ -53,16 +63,28 @@ vec3 TonemapLottes( vec3 color ) {
...
 	vec4 color = texture2D(u_CurrentMap, st);
 	color *= u_GlobalLightFactor;
 
 #if defined(r_highPrecisionRendering) && defined(HAVE_ARB_texture_float)
-	color.rgb = TonemapLottes( color.rgb * u_TonemapExposure );
+	if( u_Tonemap ) {
+		#if defined(HAVE_ARB_explicit_uniform_location) && defined(HAVE_ARB_shader_atomic_counters)
+			if( u_TonemapAdaptiveExposure ) {
+					const float l = GetAverageLuminance( luminanceU ) - 8;
+					color.rgb *= clamp( 0.18f / exp2( l * 0.8f + 0.1f ), 0.0f, 2.0f );
```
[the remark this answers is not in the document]
VReaperV: It won't, the lowest it can go is ~0.0016.

src/engine/renderer/gl_shader.h
```diff
@@ -3906,6 +3918,18 @@ class u_TonemapParms :
...
 };
 
+class u_TonemapParms2 :
+	GLUniform4f {
+	public:
+	u_TonemapParms2( GLShader* shader ) :
+		GLUniform4f( shader, "u_TonemapParms2", true ) {
+	}
+
+	void SetUniform_TonemapParms2( vec4_t tonemapParms2 ) {
```
slipher, line 3928: This could use a more readable interface. Arguments with named float(s) instead of vec4

src/engine/renderer/tr_backend.cpp
```diff
@@ -3211,6 +3211,37 @@ void RB_FXAA()
...
+
+	uint32_t globalWorkgroupX = ( width + 7 ) / 8;
+	uint32_t globalWorkgroupY = ( height + 7 ) / 8;
+
+	GL_Bind( tr.currentRenderImage[backEnd.currentMainFBO] );
+
+	gl_luminanceReductionShader->SetUniform_ViewWidth( width );
+	gl_luminanceReductionShader->SetUniform_ViewHeight( height );
+	vec4_t parms { log2f( r_toneMappingHDRMax.Get() ) };
+	parms[1] = UINT32_MAX / ( width * height * ( uint32_t( parms[0] ) + 8 ) );
```
slipher, line 3235: Why is UINT32_MAX used?

src/engine/renderer/glsl_source/luminanceReduction_cp.glsl
```diff
@@ -0,0 +1,86 @@
...
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+===========================================================================
+*/
+
+/* adaptiveLightingReduction_cp.glsl */
```
slipher, line 35: doesn't match

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
```diff
@@ -61,6 +75,13 @@ void main()
 
 #if defined(r_highPrecisionRendering) && defined(HAVE_ARB_texture_float)
 	if( u_Tonemap ) {
+		#if defined(HAVE_ARB_explicit_uniform_location) && defined(HAVE_ARB_shader_atomic_counters)
```
slipher, line 78: It seems better for this ifdef to depend directly on `adaptiveExposureAvailable` rather than on some partial list of required extensions.

src/engine/renderer/tr_backend.cpp
VReaperV: That is the maximum value that can be held in an `atomic_uint`, so it ensures that there will be no overflow.

src/engine/renderer/glsl_source/luminanceReduction_cp.glsl
VReaperV: Fixed.

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
VReaperV: Done.

src/engine/renderer/gl_shader.h
VReaperV: Perhaps so. I'd just made it as an "extension" to `u_TonemapParms` that was already there.

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
```diff
@@ -25,34 +25,48 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
          / ( pow( color, vec3( u_TonemapParms[0] * u_TonemapParms[1] ) ) * u_TonemapParms[2] + u_TonemapParms[3] );
 }
+
+#if defined(ADAPTIVE_EXPOSURE_AVAILABLE)
+	layout(std140, binding = BIND_LUMINANCE) uniform ub_LuminanceUBO {
+		uint luminanceU;
+	};
+#endif
+
+float GetAverageLuminance( const in uint luminance ) {
```
slipher, line 62: needs to be ifdef'd

src/engine/renderer/glsl_source/common_cp.glsl
```diff
@@ -56,6 +56,11 @@ array must be in the form of uvec4 array[] */
 
 /* Macro combinations for subgroup ops */
 
+#if defined(HAVE_KHR_shader_subgroup_basic) && defined(HAVE_KHR_shader_subgroup_arithmetic)\
+	&& defined(HAVE_ARB_shader_atomic_counter_ops)
+	#define SUBGROUP_ATOMIC
```
slipher, line 61: unused?

VReaperV: Oh, I had the wrong ifdef in the `luminanceReduction` shader. Fixed it to use `SUBGROUP_ATOMIC`.

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
VReaperV: Fixed.

illwieckz:
@VReaperV you'll notice I kept the release published this way. So I'm not taking my own opinion as something absolute.

PR #1561, opened 2025-02-23
NUKE some unused ET stuff

VReaperV:


slipher:
`RenderFinishCommand` could be deleted too

VReaperV:
> `RenderFinishCommand` could be deleted too

Good point, done now.

VReaperV:
I've added another commit, which removes the `tr_flare.cpp` stuff. This was a feature in ET, but most for the code was removed in Daemon so it wasn't actually working. And the way it worked originally is not sound either (doing a `glReadPixels()` to compute visibility or something...).

slipher:
LGTM

PR #1562, opened 2025-02-23
Fix a crash with r_noportals

VReaperV:
`r_noportals` is non-latch, so loading a map with `r_noportals 1` that uses alphaGen portals, then setting `r_noportals 0` will result in a crash. This can be observed on e. g. map `pulse` with the camera screens.

Fix this by always loading the portal shader.

### DIFF
--- a/src/engine/renderer/tr_shade.cpp
+++ b/src/engine/renderer/tr_shade.cpp
@@ -332,11 +332,9 @@ static void GLSL_InitGPUShadersOrError()
 		gl_shaderManager.load( gl_contrastShader );
 	}
 
-	if ( !r_noportals->integer || r_liquidMapping->integer )
-	{
-		// portal process effect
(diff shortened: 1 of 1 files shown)


src/engine/renderer/tr_shade.cpp
```diff
@@ -332,7 +332,7 @@ static void GLSL_InitGPUShadersOrError()
 		gl_shaderManager.load( gl_contrastShader );
 	}
 
-	if ( !r_noportals->integer || r_liquidMapping->integer )
+	if ( r_liquidMapping->integer )
```
slipher, line 335: This does not always load the portal shader.

VReaperV: Fixed.

slipher:
LGTM

PR #1563, opened 2025-02-24
Some more material system clean-up

VReaperV:
These are mostly changes that I picked from #1559.

Previously only the material system used compute shaders at all, but now that other systems will be using them, all this stuff needs to be moved to another shader. Otherwise the other shaders would fail to compile, or all of them would require a bunch of macros to check for it.

Also moved the material system per-frame compute dispatches to `StartFrame()` since they work there now.

### DIFF
--- a/src.cmake
+++ b/src.cmake
@@ -147,6 +147,7 @@ set(GLSLSOURCELIST
     ${ENGINE_DIR}/renderer/glsl_source/cull_cp.glsl
     ${ENGINE_DIR}/renderer/glsl_source/depthReduction_cp.glsl
     ${ENGINE_DIR}/renderer/glsl_source/processSurfaces_cp.glsl
+    ${ENGINE_DIR}/renderer/glsl_source/material_cp.glsl
     ${ENGINE_DIR}/renderer/glsl_source/material_vp.glsl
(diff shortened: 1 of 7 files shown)


DolceTriade:
Tested and didn't notice any regressions, though I did spot a material issue: https://github.com/DaemonEngine/Daemon/issues/1572

But this happens with or without this PR (and on vanilla 0.55.2)

slipher:
The StartFrame change seems mess up time-dependent stuff a bit (`tcMod rotate`, `tcMod scroll`, light styles). Like the time is off by 1 frame. On master if I take a screenshot at a given game time, I get the same result with or without material system. With this branch it is a bit off.

VReaperV:
> The StartFrame change seems to mess up time-dependent stuff a bit (`tcMod rotate`, `tcMod scroll`, light styles). Like the time is off by 1 frame. On master if I take a screenshot at a given game time, I get the same result with or without material system. With this branch it is a bit off.

Oh, that's because it's using the incorrect `refdef` there. I've removed that change for now.

VReaperV:
Also added a fix for a crash on maps where `MST_FLARE` surfaces are present.

slipher:
LGTM

PR #1564, opened 2025-02-24
NUKE unused refractionMap/dispersionMap

VReaperV:
These were in XreaL, but never actually used on any map. We also don't even have them implemented properly (whatever that should even mean in this context), and neither did XreaL (there's some refractionMap code there that is #ifdef'd out, and dispersionMap is not used at all).


slipher:
LGTM

PR #1568, opened 2025-02-26
Turn off fast-math mode by default

slipher:
Fixes https://github.com/Unvanquished/Unvanquished/issues/2311.


DolceTriade (approved):
RIP bugs

slipher:
Had to add another commit to stop warning spam that makes the CI fail.

PR #1570, opened 2025-03-01
renderer: add non-bitwise alternative for u_Color and u_ColorModulateColorGen

illwieckz:
Add non-bitwise alternative for `u_Color` and `u_ColorModulateColorGen`, it is when GLSL 1.20 isn't supported.

The material system assumes GLSL 1.20 is supported.

Written on top of:

- https://github.com/DaemonEngine/Daemon/pull/1576
- https://github.com/DaemonEngine/Daemon/pull/1577
- https://github.com/DaemonEngine/Daemon/pull/1578


illwieckz:
Context: I noticed that since 0.55.2 we broke GL 2.1 (GLSL 1.20) compatibility.

What broke compatibility was the usage of `uint` and bitwise operations that starts to be supported only with GL 3. (GLSL 1.30).

I don't mind if we use bitwise operations and uint in any feature that is not part of `low` and `lowest` presets.

So I ported the `u_Color` and `u_ColorModulate` code to emulate both `uint` and bitwise operations on GLSL 1.20.

I have noticed that writing a `GLUniform1ui` when `uint` isn't supported leads to a segfault, so I use `GLUniform1i` to store the bitfield to keep the code unique on C++ size. Both `int` and int32` having the same size, it works as the data is properly carried anyway.

I don't plan to do this for any other code that is not required by `low` and `lowest` presets. If it happens a not-`low*` preset uses uint or bitwise operations we can disable the feature on GLSL 1.20.

That also means that any feature from some advanced rendering code path (like the bindless one) can rely on `uint` and bitwise operations as much as we want.

illwieckz:
It's possible to test the code on non-GL 2.1 hardware when running Mesa and setting those environment variables:

```sh
export MESA_GL_VERSION_OVERRIDE='2.1'
export MESA_GLSL_VERSION_OVERRIDE='120'
```

slipher:
Speaking of those bitwise operations, on master I've been getting compile errors again with the Apple Silicon GPU driver because it doesn't like if you `&` signed with unsigned operands or something. I wonder if this can work around it :)

slipher:
Is https://github.com/DaemonEngine/Daemon/pull/1543#discussion_r1946532883 saying 1.20 can use bitwise operators not true?

VReaperV:
> Is [#1543 (comment)](https://github.com/DaemonEngine/Daemon/pull/1543#discussion_r1946532883) saying 1.20 can use bitwise operators not true?

The spec for version 120 has bitwise operators, though it does not support `uint` as a type.

src/engine/renderer/glsl_source/common.glsl
```diff
@@ -43,33 +43,97 @@ array must be in the form of uvec4 array[] */
 #define UINT_FROM_UVEC4_ARRAY( array, id ) ( ( array )[( id ) / 4][( id ) % 4] )
 #define UVEC2_FROM_UVEC4_ARRAY( array, id ) ( ( id ) % 2 == 0 ? ( array )[( id ) / 2].xy : ( array )[( id ) / 2].zw )
 
+// Common functions
+
+vec4 UnpackColor( const in int color )
+{
+#if __VERSION__ > 120
+	uint uColor = color;
+	return unpackUnorm4x8( uColor );
```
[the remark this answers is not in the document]
VReaperV: Actually, is this even correct given that you're directly copying uint bytes into an int on the cpp side, but doing a cast here instead?

VReaperV:
> Speaking of those bitwise operations, on master I've been getting compile errors again with the Apple Silicon GPU driver because it doesn't like if you `&` signed with unsigned operands or something. I wonder if this can work around it :)

Probably so. Some drivers fail to do the implicit conversions.

src/engine/renderer/Material.cpp
```diff
@@ -1059,7 +1059,7 @@ void BindShaderFog( Material* material ) {
 	gl_fogQuake3ShaderMaterial->SetUniform_FogDepthVector( fogDepthVector );
 	gl_fogQuake3ShaderMaterial->SetUniform_FogEyeT( eyeT );
 
-	gl_fogQuake3ShaderMaterial->SetUniform_ColorGlobal( fog->color );
+	gl_fogQuake3ShaderMaterial->SetUniform_Color( fog->color );
```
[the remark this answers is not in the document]
illwieckz: Reverted.

src/engine/renderer/glsl_source/common.glsl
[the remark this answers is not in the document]
illwieckz: The bits should be the same regardless of the type, right? This is just a bit field…

src/engine/renderer/glsl_source/common.glsl
```diff
@@ -43,33 +43,97 @@ array must be in the form of uvec4 array[] */
...
+#endif
+}
+
 /* Bit 0: color * 1
 Bit 1: color * ( -1 )
 Bit 2: color += lightFactor
 Bit 3: alpha * 1
 Bit 4: alpha * ( -1 )
 Bit 5: alpha = 1
+// There should be no bit above.
```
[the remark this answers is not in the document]
illwieckz: My code assumes that every bits higher than the lightFactor are zero. Anyway I moved the lightFactor to the higher bits so we are safe for the future.

src/engine/renderer/glsl_source/common.glsl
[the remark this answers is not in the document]
VReaperV: `int` in GLSL is stored as two's complement, so presumably the conversion can yield different results.

illwieckz:
> > Is [#1543 (comment)](https://github.com/DaemonEngine/Daemon/pull/1543#discussion_r1946532883) saying 1.20 can use bitwise operators not true?
> 
> The spec for version 120 has bitwise operators, though it does not support `uint` as a type.

What? I just tested… and it works, with the original code, just by using `int` instead of `uint`…

But that document says:

- https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.1.20.pdf

> The OpenGL Shading Language has the following operators.
> Those marked reserved are illegal.
>
> bit-wise shift (reserved) << >>

Actually it's funny because they say `%` is reserved and then illegal too. Though for this one we can just use `mod()` to be safe.

illwieckz:
The GLSL 1.30 version doesn't say bitwise operators are reserved:

- https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.1.30.pdf

VReaperV:
> > > Is [#1543 (comment)](https://github.com/DaemonEngine/Daemon/pull/1543#discussion_r1946532883) saying 1.20 can use bitwise operators not true?
> > 
> > 
> > The spec for version 120 has bitwise operators, though it does not support `uint` as a type.
> 
> What? I just tested… and it works, with the original code, just by using `int` instead of `uint`…
> 
> But that document says:
> 
> * https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.1.20.pdf
> 
> > The OpenGL Shading Language has the following operators.
> > Those marked reserved are illegal.
> > bit-wise shift (reserved) << >>
> 
> Actually it's funny because they say `%` is reserved and then illegal too. Though for this one we can just use `mod()` to be safe.

Oh, I was looking at the pre-processor one. But if it works, then whatever I guess.

slipher:
So bitwise operations are not part of GLSL 1.20 but the Mesa driver allows it? Assuming Mesa is usually used for the ancient systems I guess we can just use bitwise operators.

illwieckz:
Something to keep in mind is that I'm currently on the move and testing GL 2.1 using the Mesa environment variables to downgrade the version, but the hardware itself supports 4.6 and all the extensions are listed as available. This can lead to some bugs as seen in one commit where I explicitly disable a GL 4.6 extension when running GL 2.1. So maybe it just works because the driver and the hardware has implemented it and Mesa isn't good enough at disabling what should not be there. Il will have physical access back to Intel, Nvidia and ATI GL 2.1 hardware starting with tomorrow.

illwieckz:
I remember we had to do a non-bitwise fix in the past:

- https://github.com/DaemonEngine/Daemon/commit/df6b609d4c38b55f178cf24d250f99ec5fe602f6
- https://github.com/DaemonEngine/Daemon/issues/490

It looks like the extension we look for is `EXT_gpu_shader4`:

- https://registry.khronos.org/OpenGL/extensions/EXT/EXT_gpu_shader4.txt

> - Integer bitwise operators are now enabled.

We currently load `EXT_gpu_shader4` when available, so me currently testing GL 2.1 on a 4.6 hardware explains why bitwise operations are working.

> Though it should also allow `uint` but I don't get it:

> * Full signed integer and unsigned integer support in the OpenGL Shading Language:

illwieckz:
So yes with `MESA_EXTENSION_OVERRIDE='-GL_EXT_gpu_shader4'` I get:

```
error: operator '%' is reserved in GLSL 1.20 (GLSL 1.30 or GLSL ES 3.00 required)
```

So replaced `x % y` with `mod(x, y)`, but it looks like that in GLSL those two operations are not that equivalent, because my UI is currently pitch black, but the whole map look properly rendered. The translucency of 2D works, but colors are all pitch black.

At this point I can't exclude it may uncovers a dormant bug in the 2D code.

unvanquished_2025-03-02_053018_000

slipher:
Then the question would be whether Mesa implements the extension on that hardware.

illwieckz:
> Then the question would be whether Mesa implements the extension on that hardware.

Mesa doesn't, I've checked my previous logs, this extension isn't provided.

illwieckz:
So, the function that turns the UI black when I enable the alternative code is `vec4 ColorModulateToColor( const in int colorMod, const in float lightFactor )`, which has no reason to be used for UI.

illwieckz:
I foud the cause of the black UI. The GLSL 1.20 specification says:

> Because of their intended (limited) purpose, integers are limited to 16 bits of precision, plus a sign representation in both the vertex and fragment languages

I had moved the bits for `lightFactor` at the end at the 32-bit integer to guarantee there is no higher bits used…

If I move the bits for `lightFactor` at the end of the 16-bit integer, it works again.

It is said the sign representation is not part of those 16-bits, so it means the code can be purged from my sign hack, and it actually bumps the amount of bits to kinda 17-bit with a bit recoverable with an explicit sign test that would not pollute any division like a stray upper bit can do.

It also means we should encode colorgen on 16-bits instead of 32-bits on GLSL 1.20, which is fine for a `low` preset. The good news is that I first implemented 16-bit packing after doing 32-bit-with-sign-hack. So I just have to recover that code from an old branch.

Being limited to 16-bit plus sign means we still have 6 bits available, plus the sign, (so 7 kinda-bits) to encode future information.

slipher:
> It also means we should encode colorgen on 16-bits instead of 32-bit on GLSL 1.20, which is fine for a `low` preset (we can keep 32-bit on higher GLSL, it works). The good news is that I first implemented 16-bit packing before doing 32-bit-with-sign-hack. So I just have to recover that code from an old branch.

Maybe it would be easier to just change `u_Color` back to a float for that case? The uniform golfing thing with integer-packed color data was supposed to benefit the material system somehow. But for the non-material system it's not good; we might as well send u_Color as a float uniform whenever material system is off.

illwieckz:
That is `lightFactor` that is encoded in the `u_ColorModulateColorGen` bitfield, while `u_Color` encodes 4 numbers. If convenient, we can go back to float to some of them. I also want to see the performance of emulated bitwise operations on old hardware (on my laptop when downgrading the GLSL version, I see no difference). Of course the material system can go full `uint` and bitwise operations.

slipher:
Yes I'm suggesting to move `u_Color` back to vec4 for non-material rendering. I suspect that in terms of minimizing the amount of code, it will be a win to avoid the uniform golfing for `u_Color` but keep it for `u_ColorModulate`.

illwieckz:
> Yes I'm suggesting to move `u_Color` back to vec4 for non-material rendering. I suspect that in terms of minimizing the amount of code, it will be a win to avoid the uniform golfing for `u_Color` but keep it for `u_ColorModulate`.

I tried this but this is not easy. For example fogGlobal has no material variant.

The current branch implements packed data for both `u_Color` and `u_ColodModulateColorGen` whenever bitwise operations are supported.

Because `int` is used as a packed format and GLSL 1.20 doesn't support 32-bit integers, when bitwise operations are not supported `u_Color` is packed in 16-bit. This is fine for a `low` or `lowest` preset. In year 2006 my latptop could not do 32-bit OpenGL in full resolution so I played Tremulous in 16-bit and it was fine. Here only `u_Color` is 16-bit while everything else remains 32-bit. So this feels really fine to me.

I haven't checked the performance on real GL 2.1 hardware yet.

illwieckz:
I'm not sure we can do 32-bit bitwise operations on 32-bit `int`, because GLSL 1.30 spec says:

> The shift operators (<<) and (>>). For both operators, the operands must be signed or unsigned
> integers or integer vectors. One operand can be signed while the other is unsigned. In all cases, the
> resulting type will be the same type as the left operand. If the first operand is a scalar, the second
> operand has to be a scalar as well. If the first operand is a vector, the second operand must be a scalar
> or a vector, and the result is computed component-wise. The result is undefined if the right operand is
> negative, or greater than or equal to the number of bits in the left expression's base type. The value of
> E1 << E2 is E1 (interpreted as a bit pattern) left-shifted by E2 bits. The value of E1 >> E2 is E1 right-
> shifted by E2 bit positions. If E1 is a signed integer, the right-shift will extend the sign bit. If E1 is an
> unsigned integer, the right-shift will zero-extend.
> -- https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.1.30.pdf

So to be safe I reconstruct the `uint` in GLSL by copying the 31 bits of the `int` then writing the 32th bit using the sign.

illwieckz:
All this bitwise emulation has huge impact on performance indeed.

I tested on one of the oldest and lowest-end card (ATI RV370):

||plat23 spectator|plat23 `1920 1920 0 0 0`|0.55.2 benchmark demo
|-|-|-|-
|0.55.2|116 fps|46 fps|6204 frames, 126.3s: 49.1 fps
|this branch|71 fps|22 fps|6204 frames, 195.5s: 31.7 fps

Though this is an intentional choice of very slow hardware, there exists GL 2.1 cards that are known to be playable in 1080p, so this would probably regress them to 720p or so. I have some of them but I'm too lazy to test right now.

I believe this is ready for now. I may investigate doing `u_Color` as float on them later to recover some performance.

illwieckz:
So I managed to implement a float variant of `u_Color` to be used outside of the material system, I kept the 16-bit packed int `u_Color` because of `fogGlobal` that has no material variant (I can also make it float in both material and non-material instead).

It increases a bit the performance, but not that much. So I guess what really destroy performance is the bitwise emulation on `u_ColorModulateColorGen`…

||plat23 spectator|plat23 `1920 1920 0 0 0`|0.55.2 benchmark demo
|-|-|-|-
|0.55.2|116 fps|46 fps|6204 frames, 126.3s: 49.1 fps
|bitfield color, bitfield modulate|71 fps|22 fps|6204 frames, 195.5s: 31.7 fps
|float color, bitfield modulate|75 fps|25 fps|6204 frames, 170.1s: 36.5 fps

illwieckz:
We may have some performance regression elsewhere in the code, when I reimplement some float-based u_ColorModulateColorGen, I still don't fully recover the performance we had in 0.55.2:

||plat23 spectator|plat23 `1920 1920 0 0 0`|0.55.2 benchmark demo
|-|-|-|-
|0.55.2|116 fps|46 fps|6204 frames, 126.3s: 49.1 fps
|bitfield color, bitfield modulate|71 fps|22 fps|6204 frames, 195.5s: 31.7 fps
|float color, bitfield modulate|75 fps|25 fps|6204 frames, 170.1s: 36.5 fps
|float color, float modulate|82 fps|39 fps|6204 frames, 127.9s: 48.5 fps

illwieckz:
So, the remaining performance eater is the tonemapper, it looks like it is not fully disabled when disabled.

VReaperV:
> So, the remaining performance eater is the tonemapper, it looks like it is not fully disabled when disabled.

It's a 21-year old GPU. It wouldn't be surprising if it has bad performance with branches.

illwieckz:
||plat23 spectator|plat23 `1920 1920 0 0 0`|0.55.2 benchmark demo
|-|-|-|-
|0.55.2|116 fps|46 fps|6204 frames, 126.3s: 49.1 fps
|float color, float modulate|115 fps|45 fps|6204 frames, 106.5s: 58.2 fps

illwieckz:
So with the tonemapper being disabled with some `ifdef`, I get better numbers:

||plat23 spectator|plat23 `1920 1920 0 0 0`|0.55.2 benchmark demo
|-|-|-|-
|0.55.2|116 fps|46 fps|6204 frames, 126.3s: 49.1 fps
|bitfield color, bitfield modulate|94 fps|23 fps|6204 frames, 174.3s: 35.6 fps
|float color, bitfield modulate|101 fps|28 fps|6204 frames, 150.3s: 41.3 fps
|float color, float modulate|116 fps|45 fps|6204 frames, 106.5s: 58.2 fps

VReaperV:
That loses the ability to switch tonemapping on/off at run-time and increases the amount of shader permutations. It's better to `#ifdef` the GLSL version instead.

If the float thing works, it should be used instead of the much more complex bit-wise emulation (and the restriction to 16 bits) on ancient hw (the emulation is then not needed at all).

illwieckz:
> That loses the ability to switch tonemapping on/off at run-time

Yes, I don't see that as an issue.

> and increases the amount of shader permutations.

Why? The shader is either built with or rebuilt without, there is only one shader.

> It's better to #ifdef the GLSL version instead.

You mean to force disable the tone mapper when GLSL is old?

> If the float thing works, it should be used instead of the much more complex bit-wise emulation (and the restriction to 16 bits) on ancient hw (the emulation is then not needed at all).

Yes, if that works, the emulation is not needed at all. Though it doesn't fully work yet, there have been some deep rewrite of the lightFactor code so right now the vertex lighting mode doesn't get any light factor.

Also I don't know what to do with this code in `u_ColorModulateColorGenFloat`, I suspect this is legacy (the non-float code removed it):

```c++
	if ( needAttrib )
	{
		_shader->AddVertexAttribBit( ATTR_COLOR );
	}
	else
	{
		_shader->DelVertexAttribBit( ATTR_COLOR );
	}
```

VReaperV:
> I guess you meant “That loses the ability to switch tonemapping on/off without `vid_restart`”.
> Yes, I don't see that as an issue.

Yeah, that is what I meant. Switching it on/off is useful to see the original (clamped) input.

> Why? The shader is either built with or rebuilt without, there is only one shader.

If you want to turn it off, you have to re-build the shader.

> You mean to force disable the tone mapper when GLSL is old?

Yes.

> Also I don't know what to do with this code in u_ColorModulateColorGenFloat(), I suspect this is legacy (the non-float code removed it):
> ```
> if ( needAttrib )
> {
> 	_shader->AddVertexAttribBit( ATTR_COLOR );
> }
> else
> {
> 	_shader->DelVertexAttribBit( ATTR_COLOR );
> }
> ```

Yes, it was removed to avoid vertex format changes. You can use an extra uniform for it for the float approach.

illwieckz:
So, the float alternative is fully implemented. The bitwise emulation is deleted. Performance is as expected.

||plat23 spectator|plat23 `1920 1920 0 0 0`|0.55.2 benchmark demo
|-|-|-|-
|0.55.2|116 fps|46 fps|6204 frames, 126.3s: 49.1 fps
|this branch|116 fps|44 fps|6204 frames, 112.1s: 55.4 fps

The float packing is different than the one that existed before the uint packing:

Before:

```
colorMod[ 0 ]: red * f
colorMod[ 1 ]: green * f
colorMod[ 2 ]: blue * f
colorMod[ 4 ]: alpha * f
```

```
colorMod[ 0 ]: color * f
colorMod[ 1 ] absolute value: lightFactor
colorMod[ 1 ] minus sign: color += lightFactor
colorMod[ 2 ]: alpha = 1
colorMod[ 3 ]: alpha * f
```

The fact we store the rgb color in a single float means we can store much more data in a single vec4, like the lightFactor, and other things. We can actually store more information with sign and magic values in `colorMod[ 2 ]`. Or storing an index in `colorMod[ 0 ]` and `colorMod[ 3 ]` (like the `uint` packing does) will also make the signs available for storage of more data.

illwieckz:
Thansk to me finding a way to better use the 4 floats to store data, the `u_ColorModulateColorGen_Float` and `u_ColorModulateColorGen_Uint` code is exactly the same except the packing format. The exact same data is packed, and it is processed the same way after unpacking. Just the packing format differs. So this prevents the trouble of having an implementation lagging behind another one. When adding more data to pack in, only the packing/unpacking code has to be modified. Everyting else would be left untouched. I'm very happy with this, that's robust and future-proof.

So Material system uses uint packing, and non-Material system uses float packing.

Actually with the current data we are packing, we can store the colorGen into an `ivec3` instead of a `vec4`, but doing such thing is not within scope of this PR.

illwieckz:
@VReaperV I forgot to answer to this question:

> > Disable ARB_shader_draw_parameters on GLSL 1.20, it's not working.
> Is there any specific system where that happens? If a higher GLSL version is not supported, that extension shouldn't even be reported as enabled.

Yes, Mesa does that. For example if you're running a GL 4.6 hardware and GL 4.6 is implemented, if you force the GL 2.1 version, Mesa will still list all the available extensions.

illwieckz:
So this looks ready to me.

When reviewing, you can ignore the `truly-disable tonemapper` commit. This commit belongs to a future PR and is probably not ready, but I need it to test this branch when comparing performance with 0.55.2. This commit will be removed before merging (and submitted to another PR).

VReaperV:
> @VReaperV I forgot to answer to this question:
> 
> > > Disable ARB_shader_draw_parameters on GLSL 1.20, it's not working.
> > 
> > 
> > Is there any specific system where that happens? If a higher GLSL version is not supported, that extension shouldn't even be reported as enabled.
> 
> Yes, Mesa does that. For example if you're running a GL 4.6 hardware and GL 4.6 is implemented, if you force the GL 2.1 version, Mesa will still list all the available extensions.

But that's not a realistic use-case.

illwieckz:
> But that's not a realistic use-case.

Yep but I need my test environment to work. 😃️

illwieckz:
I created a `ColorModulateColorGen()` function that is used by both `u_ColorModulateColorGen_Float::SetUniform_ColorModulateColorGen()` and `u_ColorModulateColorGen_Uint::SetUniform_ColorModulateColorGen()`.

Each `SetUniform_ColorModulateColorGen()` function is now just doing their respective packing.

It ensures the logic is kept exactly the same between both format, and that we only have to care about the packing when adding something new to pack there.

src/engine/renderer/glsl_source/common.glsl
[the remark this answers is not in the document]
illwieckz: This is now using `uint`.

illwieckz:
I rewrote a bit the GLSL code to deduplicate it, now only differing in the packing/unpacking code, and sharing the processing code whatever the packing format.

src/engine/renderer/glsl_source/common.glsl
```diff
@@ -43,33 +43,129 @@ array must be in the form of uvec4 array[] */
...
+colorMod[ 1 ] absolute value: lightFactor
+colorMod[ 1 ] minus sign: color += lightFactor
+colorMod[ 2 ]: alpha = 1
+colorMod[ 3 ]: alpha * f */
+
+vec4 ColorModulateToColor( const in colorModulatePack colorMod )
+{
+#if defined(USE_MATERIAL_SYSTEM)
+	vec3 colorModArray = vec3( 0.0f, 1.0f, -1.0f );
+	uint rgbIndex = colorMod & uint( 3 );
```
VReaperV, line 88: `uint( 3 )`->`3u` (same for other places like this)

src/engine/renderer/glsl_source/common.glsl
```diff
@@ -43,33 +43,129 @@ array must be in the form of uvec4 array[] */
...
+	float lightFactor = ColorModulateToLightFactor( colorMod );
+
+	// This is used to skip vertex colours if the colorMod doesn't need them.
+	color.a = modBits.alphaAddOne ? 1.0 : color.a;
+
+	colorModulation.rgb += vec3( modBits.isLightStyle ? lightFactor : 0 );
+
+	vec4 unpackedColor = UnpackColor( packedColor );
+
+	vec4 multipliedColor = unpackedColor;
```
VReaperV, line 167: `vec4 multipliedColor = UnpackColor( packedColor ) * lightFactor;`?

illwieckz: Only rgb is multiplied, not the alpha channel.

VReaperV: Oh, right. The second line is still extraneous though.

src/engine/renderer/glsl_source/common.glsl
```diff
@@ -43,33 +43,139 @@ array must be in the form of uvec4 array[] */
...
+
+/* colorMod uint format:
+
+colorMod << 0: color * 1
+colorMod << 1: color * ( -1 )
+colorMod << 2: alpha * 1
+colorMod << 3: alpha * ( -1 )
+colorMod << 4: color += lightFactor
+colorMod << 5: alpha = 1
+colorMod << 6-27: available for future usage
```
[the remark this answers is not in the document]
illwieckz: I fixed the comment.

src/engine/renderer/glsl_source/common.glsl
illwieckz: Done.

src/engine/renderer/glsl_source/common.glsl
illwieckz: Well, that's more about style I guess.

Would you prefer something like that?

```
vec4 multipliedColor = unpackedColor * vec4( vec3( lightFactor ), 1 ) );
```

illwieckz: I removed the temporary `multipliedColor` and reuse `unpackedColor`.

illwieckz:
So now the PR just contains what is related to GLSL compatibility.

slipher:
Fog doesn't appear with material system enabled.

unvanquished-atcshd-fog

illwieckz:
OK. I have a fix but I'm not sure to understand why.

@VReaperV is the fact `fogQuake3`'s `u_ColorGlobal` is “global” prevents it to use a different type between material and non-material system?

VReaperV:
> OK. I have a fix but I'm not sure to understand why.
> 
> @VReaperV is the fact `fogQuake3`'s `u_ColorGlobal` is “global” prevents it to use a different packing format between material and non-material system?

Global means that it will be set as a uniform in `BindShader*()`, while non-global ones are put into a uniform buffer.

illwieckz:
I guess that means yes? 😅️

Also @VReaperV I have another question, is doing that will pass two uniforms to the GLSL shader (one being unset), or only one?

```c++
class GLShader_fogQuake3Material :
	public GLShader,
	public u_FogMap,
	public u_ModelMatrix,
	public u_ModelViewProjectionMatrix,
	public u_ColorGlobal_Float,
	public u_ColorGlobal_Uint,
	public u_FogDistanceVector,
	public u_FogDepthVector,
	public u_FogEyeT,
	public GLDeformStage {
	public:
	GLShader_fogQuake3Material( GLShaderManager* manager );
	void SetShaderProgramUniforms( shaderProgram_t* shaderProgram ) override;
};

template<typename Shader> void SetUniform_ColorGlobal( Shader* shader, const Color::Color& color )
{
	if( glConfig2.shadingLanguageVersion > 120 )
	{
		shader.SetUniform_ColorGlobal_Uint( color );
	}
	else
	{
		shader.SetUniform_ColorGlobal_Float( color );
	}
}

class u_ColorGlobal_Float :
	GLUniform4f
{
public:
	u_ColorGlobal_Float( GLShader *shader ) :
		GLUniform4f( shader, "u_ColorGlobal", true )
	{
	}

	void SetUniform_ColorGlobal_Float( const Color::Color& color )
	{
		this->SetValue( color.ToArray() );
	}
};

class u_ColorGlobal_Uint :
	GLUniform1ui {
	public:
	u_ColorGlobal_Uint( GLShader* shader ) :
		GLUniform1ui( shader, "u_ColorGlobal", true ) {
	}

	void SetUniform_ColorGlobal_Uint( const Color::Color& color ) {
		this->SetValue( packUnorm4x8( color.ToArray() ) );
	}
};
```

```c++
SetUniform_ColorGlobal( gl_fogQuake3Shader, fog->color );
```

VReaperV:
> I guess that means yes? 😅️

Well, depends on what you mean by packing format.

> Also @VReaperV I have another question, is doing something like that will pass two uniforms to the GLSL shader (one being unset), or only one?

That will result in just one uniform being set.

illwieckz:
> Well, depends on what you mean by packing format.

Well, let's rephrase it:

is the fact fogQuake3's `u_ColorGlobal` is “global” prevents to either call `SetUniform_ColorGlobal_Float()` or `SetUniform_ColorGlobal_Uint` between material and non-material system and only one of them will be called anyway?

illwieckz:
So I used some templated dispatch making it possible to always use `uint` packing when `EXT_gpu_shader4` is available, whatever the renderer, whatever the GLSL shader.

VReaperV:
I suppose that answers the previous questions then?

illwieckz:
Just to make it clear: I now consider this ready. There may be ideas of further improvements to be done over this but that would then belongs to future PRs.

slipher:
This PR fixes the GLSL compile errors with the Apple Silicon driver. Yay! Note that this driver does not support extensions needed for material system. For the record, the errors on master look like this:

```Warn: Compile log:
ERROR: 0:10056: '&' does not operate on 'unsigned int' and 'int'
ERROR: 0:10057: Use of undeclared identifier 'colorModulate'
ERROR: 0:10057: '&' does not operate on 'unsigned int' and 'int'
ERROR: 0:10058: Use of undeclared identifier 'colorModulate'
ERROR: 0:10062: '&' does not operate on 'unsigned int' and 'int'
ERROR: 0:10062: '&' does not operate on 'unsigned int' and 'int'
ERROR: 0:10063: Use of undeclared identifier 'colorModulate'
ERROR: 0:10063: '&' does not operate on 'unsigned int' and 'int'
ERROR: 0:10064: Use of undeclared identifier 'colorModulate'
ERROR: 0:10068: '&' does not operate on 'unsigned int' and 'int'
ERROR: 0:10073: '&' does not operate on 'unsigned int' and 'int'
 
Warn: Unhandled exception (15ShaderException): Couldn't compile vertex shader: generic 
```

VReaperV:
> > That will result in just one uniform being set.
> 
> Great, that means we can use the GLSL version to select the packing format, instead of material/non-material.
> 
> Do you know if there is a way to create a wrapper that is part of the shader object? So I can do `shader->function( input )` that either calls `shader->function_Uint( input )` or `shader->function_Float( input )` without doing such `function( shader, input )` template as I gave as example?

Well, you could have something like a `u_ColorGlobal` inherited from both `u_ColorGlobal_Uint` and `u_ColorGlobal_Float`, and call whichever base method based on the extension availability... That doesn't really sound better to me though, since you'd have some nastier inheritance in shaders.

illwieckz:
> This PR fixes the GLSL compile errors with the Apple Silicon driver. Yay! Note that this driver does not support extensions needed for material system. For the record, the errors on master look like this:
> 
> ```
> ERROR: 0:10056: '&' does not operate on 'unsigned int' and 'int'
> ```

Yes, this is fixed by writing `a & 1` as `a & 1u`.

I also got those errors on Mesa at some point. I wasn't getting them by default, but by toying with multiple configurations I got them. I'm not surprised macOS is less permissive, as it is more complete (one funny thing is that macOS was even still giving us a GL 2.1 context until 2021).

illwieckz:
I squashed and rebased.

illwieckz:
I added a commit to disable `EXT_gpu_shader4` (bitwise operations and `uint` type) on Intel GMA because that's buggy:

- https://gitlab.freedesktop.org/mesa/mesa/-/issues/12803

illwieckz:
> I added a commit to disable `EXT_gpu_shader4` (bitwise operations and `uint` type) on Intel GMA because that's buggy:
> 
>     * https://gitlab.freedesktop.org/mesa/mesa/-/issues/12803

I removed it because there is no need to delay that PR more. It's ready. I just need an approval.

VReaperV:
I suppose the `renderer: also require EXT_gpu_shader4 for the material system` commit isn't needed now, since it's disabled in other ways when not supported?

illwieckz:
> I suppose the `renderer: also require EXT_gpu_shader4 for the material system` commit isn't needed now, since it's disabled in other ways when not supported?

In that branch the material system assumes the `EXT_gpu_shader4` feature is there, it directly calls the Uint functions without any wrapper, so we better check it is there.

src/engine/renderer/Material.cpp
```diff
@@ -148,10 +148,10 @@ void UpdateSurfaceDataGeneric3D( uint32_t* materials, shaderStage_t* pStage, boo
 	alphaGen_t alphaGen = SetAlphaGen( pStage );
 
 	const bool styleLightMap = pStage->type == stageType_t::ST_STYLELIGHTMAP || pStage->type == stageType_t::ST_STYLECOLORMAP;
-	gl_genericShaderMaterial->SetUniform_ColorModulateColorGen( rgbGen, alphaGen, mayUseVertexOverbright, styleLightMap );
+	gl_genericShaderMaterial->SetUniform_ColorModulateColorGen_Uint( rgbGen, alphaGen, mayUseVertexOverbright, styleLightMap );
```
illwieckz, line 151: @VReaperV here the `Uint` function variant is called without any wrapper.
We don't have to take care of old cards in the material system.

illwieckz:
This was tested successfully on Nvidia, Intel, ATI and Broadcom GL 2.1 hardware.

VReaperV:
> > I suppose the `renderer: also require EXT_gpu_shader4 for the material system` commit isn't needed now, since it's disabled in other ways when not supported?
> 
> In that branch the material system assumes the `EXT_gpu_shader4` feature is there, it directly calls the Uint functions without any wrapper, so we better check it is there.

I see. I didn't bother adding some of the extension checks because it's pretty much impossible for the other extensions to be present but not the older ones. Certainly not a mistake to add them though, so it's fine to keep the change.

src/engine/renderer/Material.cpp
VReaperV: Yeah, that should work fine.

src/engine/renderer/glsl_source/common.glsl
```diff
@@ -43,33 +43,129 @@ array must be in the form of uvec4 array[] */
...
+	float rgb = colorMod.r;
+	float alpha = colorMod.a;
+#endif
+
+	return vec4( rgb, rgb, rgb, alpha );
 }
 
-float ColorModulateToLightFactor( const in uint colorMod ) {
-	return ( colorMod >> 6 ) & 0xF;
+struct modBits_t {
```
VReaperV, line 102: It'd be preferrable if this followed CamelCase, and for `{` to have the same indents as here in the rest of the file (or the other way around), although non-blocking.

VReaperV:
`r_tonemapping` has no effect on this branch.

VReaperV:
Apparently, that's a regression from #1577.

src/engine/renderer/glsl_source/common.glsl
illwieckz: Done.

src/engine/renderer/gl_shader.h
```diff
@@ -3689,89 +3743,187 @@ class u_ColorModulate :
...
+	const alphaGen_t alphaGen,
+	const bool vertexOverbright,
+	const bool useMapLightFactor )
+{
+	colorModulation_t colorModulation = {};
+
+	switch ( colorGen )
+	{
+		case colorGen_t::CGEN_VERTEX:
+			colorModulation.alphaAddOne = false;;
```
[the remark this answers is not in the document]
illwieckz: Fixed.

src/engine/renderer/gl_shader.h
```diff
@@ -3689,89 +3743,187 @@ class u_ColorModulate :
...
+
+			if ( vertexOverbright )
+			{
+				// vertexOverbright is only needed for non-lightmapped cases. When there is a
+				// lightmap, this is done by multiplying with the overbright-scaled white image
+				colorModulation.isLightStyle = true;
```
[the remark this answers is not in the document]
illwieckz: Renamed to `useVertexLightFactor`.

VReaperV (approved):
LGTM

PR #1571, opened 2025-03-02
cmake: use -flto=auto compiler flag when supported

illwieckz:
Use -flto=auto compiler flag when supported, this silence this GCC warning:

> lto-wrapper: warning: using serial compilation of # LTRANS jobs

This also greatly speeds-up the linkage time as it enables LTO multithreading in GCC (either by using Make jobserver if detected, either by detecting CPU cores).

### DIFF
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -420,9 +420,13 @@ else()
 	# Saigo NaCl compiler doesn't support LTO, the flag is accepted but linking fails
 	# with “unable to pass LLVM bit-code files to linker” error.
 	if (USE_LTO AND NOT NACL)
-		try_c_cxx_flag(LTO "-flto")
+		try_c_cxx_flag(LTO_AUTO "-flto=auto")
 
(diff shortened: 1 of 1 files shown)


slipher:
LGTM. I tried on MSYS2 and it didn't break anything, although the parallelism didn't really appear to work. Running under Ninja, there were 6 processes but only one of them ever used the CPU at a time.

DolceTriade:
:shrug: If we want faster linking, we should just use mold or one of the new fancy linkers, but no harm so LGTM

illwieckz:
I tested with:

```
clang-10
clang-11
clang-12
clang-13
clang-14
clang-15
clang-16
clang-17
clang-18
clang-19
clang-20
gcc-10
gcc-11
gcc-12
gcc-13
gcc-14
mingw
icx
aocc
icc
zig
apple-clang
freebsd-clang
```

Either the flag is used and there is no error, either the flag is not used because detected as unsupported. I haven't checked one by one how it ended, but in all cases this never failed neither printed obvious warnings.

PR #1574, opened 2025-03-03
Fix mistake when setting linker flags

### LINKED ISSUE #1573: Still getting `argument unused during compilation: '-pie'` with dll
Here on Linux with whatever clang version I use:

```
[100%] Linking CXX shared module sgame-native-dll.so
clang: warning: argument unused during compilation: '-pie' [-Wunused-command-line-argument]
```

This flag is not meant to be used with dll, and our CMake code is supposed to only set this flag with exe…

illwieckz:
There was a stupid mistake inverting between executable-linker-flags and everythng-linker-flags.

Fixes #1573:

- https://github.com/DaemonEngine/Daemon/issues/1573

I verified that even after fixing this maCOS still complains about `-pie` not being used on executables so I kept the macOS test to disable `-pie`:

```
[100%] Linking CXX executable daemonded
clang++: warning: argument unused during compilation: '-pie' [-Wunused-command-line-argument]
```

### DIFF
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -147,7 +147,7 @@ macro(try_linker_flag PROP FLAG)
     check_C_compiler_flag(${FLAG} FLAG_${PROP})
     set(CMAKE_REQUIRED_FLAGS "")
     if (FLAG_${PROP})
-        set_exe_linker_flag(${FLAG} ${ARGN})
+        set_linker_flag(${FLAG} ${ARGN})
     endif()
 endmacro()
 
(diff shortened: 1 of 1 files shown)


illwieckz:
I also noticed that `-pie` reruires `-fPIC`, so since `-fPIC` is optional, this now tests for this flag being supported before testing for `-pie`. This is probably useless and this may skip a flag test if -fPIC isn't supported, this should be very rare but this is more consistent when written this way.

```
$ clang -pie a.cpp && echo done
ld: error: relocation R_X86_64_64 cannot be used against local symbol; recompile with -fPIC
>>> defined in /tmp/a-b05268.o
>>> referenced by a.cpp
>>>               /tmp/a-b05268.o:(main)
clang: error: linker command failed with exit code 1 (use -v to see invocation)

$ clang -fPIC -pie a.cpp && echo done
done
```

DolceTriade (approved):

illwieckz:
Thanks for the review!

PR #1576, opened 2025-03-04
renderer: disable ARB_shader_draw_parameters on GLSL 1.20

### LINKED ISSUE #1575: Some material shader code is built even when `glConfig2.usingMaterialSystem` is `false`
The material system shaders are built even when `glConfig2.usingMaterialSystem` is `false`.

This is why I get errors about features from `ARB_shader_draw_parameters` not working on old GLSL, the related GLSL code should not be compiled to begin with.

illwieckz:
Disable ARB_shader_draw_parameters on GLSL 1.20.

Extracted from:

- https://github.com/DaemonEngine/Daemon/pull/1570

Fixes #1575:

- https://github.com/DaemonEngine/Daemon/issues/1575

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -175,6 +175,10 @@ static Cvar::Cvar<bool> workaround_glExtension_missingArbFbo_useExtFbo(
 	"workaround.glExtension.missingArbFbo.useExtFbo",
 	"Use EXT_framebuffer_object and EXT_framebuffer_blit when ARB_framebuffer_object is not available",
 	Cvar::NONE, true );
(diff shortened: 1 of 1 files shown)


illwieckz:
> How about just disabling the extension on version 120? I think that would be a cleaner approach.

That was my first implementation but then I thought this would be better instead. 🤭️

I reverted to disabling the extension on version 120.

VReaperV (approved):
Non-blocking, might want to name all of these as `*ShaderDrawParameter`**s** for consistency.

LGTM.

illwieckz:
> Non-blocking, might want to name all of these as `*ShaderDrawParameter`**s** for consistency (and with camelCase for variables).

Good catch, fixed.

PR #1577, opened 2025-03-04
renderer: rework tone mapper and SSAO enablement

illwieckz:
Rework tone mapper and ssao enablement.

Extracted from:

- https://github.com/DaemonEngine/Daemon/pull/1570

### DIFF
--- a/src/engine/renderer/tr_init.cpp
+++ b/src/engine/renderer/tr_init.cpp
@@ -209,20 +209,24 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 	cvar_t      *r_gamma;
 
-	Cvar::Cvar<bool> r_tonemap( "r_tonemap", "Use  HDR->LDR tonemapping", Cvar::NONE, true );
-	Cvar::Cvar<float> r_tonemapExposure( "r_tonemapExposure", "Tonemap exposure", Cvar::NONE, 1.0f );
-	Cvar::Range<Cvar::Cvar<float>> r_tonemapContrast( "r_tonemapContrast", "Makes dark areas light up faster",
+	Cvar::Cvar<bool> r_toneMapping(
+		"r_toneMapping", "Use  HDR->LDR tonemapping", Cvar::NONE, true );
+	Cvar::Cvar<float> r_toneMappingExposure(
+		"r_toneMappingExposure", "Tonemap exposure", Cvar::NONE, 1.0f );
+	Cvar::Range<Cvar::Cvar<float>> r_toneMappingContrast(
+		"r_toneMappingContrast", "Makes dark areas light up faster",
 		Cvar::NONE, 1.6f, 1.0f, 10.0f );
-	Cvar::Range<Cvar::Cvar<float>> r_tonemapHighlightsCompressionSpeed( "r_tonemapHighlightsCompressionSpeed",
-		"Highlights saturation",
+	Cvar::Range<Cvar::Cvar<float>> r_toneMappingHighlightsCompressionSpeed(
+		"r_toneMappingHighlightsCompressionSpeed", "Highlights saturation",
 		Cvar::NONE, 0.977f, 0.0f, 10.0f );
-	Cvar::Range<Cvar::Cvar<float>> r_tonemapHDRMax( "r_tonemapHDRMax", "HDR white point",
+	Cvar::Range<Cvar::Cvar<float>> r_toneMappingHDRMax(
+		"r_toneMappingHDRMax", "HDR white point",
 		Cvar::NONE, 8.0f, 1.0f, 128.0f );
-	Cvar::Range<Cvar::Cvar<float>> r_tonemapDarkAreaPointHDR( "r_tonemapDarkAreaPointHDR",
-		"Cut-off for dark area light-up",
+	Cvar::Range<Cvar::Cvar<float>> r_toneMappingDarkAreaPointHDR(
+		"r_toneMappingDarkAreaPointHDR", "Cut-off for dark area light-up",
 		Cvar::NONE, 0.18f, 0.0f, 1.0f );
-	Cvar::Range<Cvar::Cvar<float>> r_tonemapDarkAreaPointLDR( "r_tonemapDarkAreaPointLDR",
-		"Convert to this brightness at dark area cut-off",
+	Cvar::Range<Cvar::Cvar<float>> r_toneMappingDarkAreaPointLDR(
+		"r_toneMappingDarkAreaPointLDR", "Convert to this brightness at dark area cut-off",
 		Cvar::NONE, 0.268f, 0.0f, 1.0f );
 
 	cvar_t      *r_lockpvs;
@@ -316,7 +320,13 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	Cvar::Cvar<float> r_bloomBlur( "r_bloomBlur", "Bloom strength", Cvar::NONE, 1.0 );
 	Cvar::Cvar<int> r_bloomPasses( "r_bloomPasses", "Amount of bloom passes in each direction", Cvar::NONE, 2 );
 	cvar_t      *r_FXAA;
-	cvar_t      *r_ssao;
+	Cvar::Range<Cvar::Cvar<int>> r_ssao( "r_ssao",
+		"Screen space ambient occlusion: "
+		"-1: show, 0: disabled, 1: enabled",
+		Cvar::NONE,
+		Util::ordinal( ssaoMode::DISABLED ),
+		Util::ordinal( ssaoMode::SHOW ),
+		Util::ordinal( ssaoMode::ENABLED ) );
 
 	cvar_t      *r_evsmPostProcess;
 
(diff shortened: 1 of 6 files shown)


illwieckz:
Note that it renames `r_tonemap` to `r_toneMapping` to be consistent with other cvars, fell free to comment about that.

Other related `r_tonemap*` cvars were not renamed, but maybe that's fine as they may be seen as options for the tonemap.

src/engine/renderer/tr_init.cpp
```diff
@@ -312,7 +312,13 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 	Cvar::Cvar<float> r_bloomBlur( "r_bloomBlur", "Bloom strength", Cvar::NONE, 1.0 );
 	Cvar::Cvar<int> r_bloomPasses( "r_bloomPasses", "Amount of bloom passes in each direction", Cvar::NONE, 2 );
 	cvar_t      *r_FXAA;
-	cvar_t      *r_ssao;
+	Cvar::Range<Cvar::Cvar<int>> r_ssao( "r_ssao",
+		"Screen space ambient occlusion: "
+		"-1: show, 0: disabled, 1: enabled",
+		Cvar::NONE,
```
VReaperV, line 318: This should be `Cvar::ARCHIVE`.

VReaperV (commented):
This still has the issue of not being able to instantly turn tonemapping on/off.

VReaperV:
> Other related r_tonemap* cvars were not renamed, but maybe that's fine as they may be seen as options for the tonemap.

The naming should be consistent since that works better with console auto-complete.

src/engine/renderer/tr_init.cpp
illwieckz: Do we need to put everything as archive? We have `/seta` for that.

slipher: ARCHIVE shouldn't be used any more except a few cvars that have a special UI to set them which is not controlled by the cgame. For example r_fullscreen with Alt+Enter.

illwieckz: Thanks, I don't remembered what ARCHIVE was still good for, but in my memory we could drop it on most stuff.

illwieckz:
> The naming should be consistent since that works better with console auto-complete.

Now unified.

illwieckz:
> This still has the issue of not being able to instantly turn tonemapping on/off.

I'm not considering this an issue. I can see the annoyance when testing stuff, but normal usage never randomly switches such option.

For the annoyance of having to do a `vid_restart`, at some point we may implement a mechanism that automatically recompiles GLSL shaders when said shaders don't require any file to be loaded, this would benefit most screen shaders.

slipher:
> > This still has the issue of not being able to instantly turn tonemapping on/off.
> 
> I'm not considering this an issue. I can see the annoyance when testing stuff, but normal usage never randomly switches such option.

I don't agree with the assumption "normal users should never want to toggle settings". How are you supposed to decide whether you want to use a setting without turning it off and on to compare the two configurations?

illwieckz:
Using an uniform to disable the feature means the code is actually executed on hardware not supporting modern branching and then the renders still have to wait for the compute to be done and discarded. Here it was eating 30% of the performance on the testbed I was using when noticing it

I'm not saying it would be cool to be able to toggle that without any `vid_restart`, I'm saying using an uniform for that is not a good solution because it fails to actually disable the code on some hardware. We better implement something like lazy permutations, allowing such permutations to be lazily built without having to do a `vid_restart`.

src/engine/renderer/tr_init.cpp
VReaperV: Oh, there are some places where it should be removed then.

VReaperV:
> For the annoyance of having to do a `vid_restart`, at some point we may implement a mechanism that automatically recompiles GLSL shaders when said shaders don't require any file to be loaded, this would benefit most screen shaders.

The problem here is that headers need to be regenerated. Shaders sources are all in a hashmap for built-in shaders.

illwieckz:
That's part of implementation details to be discussed there:

- https://github.com/DaemonEngine/Daemon/issues/1580

VReaperV:
> Using an uniform to disable the feature means the code is actually executed on hardware not supporting modern branching and then the renders still have to wait for the compute to be done and discarded. Here the disabled feature was still eating 30% of the performance on the testbed I was using when noticing it

That is why I suggested just disabling based on GLSL version. Clearly such hardware is incapable of running it at a playable framerate when it's enabled.

illwieckz:
> That is why I suggested just disabling based it on the GLSL version. Clearly such hardware is incapable of running it at a playable framerate when it's enabled.

One thing is that I don't know what kind of hardware can truly branch. For example AMD TeraScale (the previous-to-current architecture) is GL 4.5 but has some VLIW architecture and I'm not sure they are good at branching because of that. I can test on them, but in general I prefer to skip building the disabled code.

VReaperV:
> One thing is that I don't know what kind of hardware can truly branch.

Proper branching at ~GTX 600 series, so probably around the same for AMD.

illwieckz:
In fact, tone-mapping is cheap on high-end GLSL 120 cards, so, GLSL version doesn't matter. What matters is: is the hardware low-end or not?

Here is an ATI RV570, it only supports OpenGL 2.1 with GLSL 120, but it has no problem at running the tonemapping code:

resolution|preset|r_toneMapping off|r_toneMapping on r_highPrecisionRendering on
-|-|-|-
640×480|lowest|310 fps|282 fps
640×480|low|295 fps|270 fps
1024×768|lowest|153 fps|138 fps
1024×768|ultra|65 fps|64 fps
1280×720|lowest|132 fps|116 fps
1280×720|low|125 fps|113 fps
1280×720|medium|84 fps|81 fps
1280×720|high (image 256)|83 fps|80 fps
1280×720|ultra (image 256)|60 fps|58 fps
1920×1080|low|60 fps|60 fps
1920×1080|medium|40 fps|38 fps

This runs at 80 fps on “HDReady” 1280×720p on medium preset, or 60 fps on “FullHD” 1920×1080 on `low` preset. on `medium` preset it only eats between 1 and 3 fps. When testing presets above `medium` I still set the same image size as `low` or some images may be broken and performance may be unreliable. But with fairly acceptable image resolution `256×256`, this card can run maps on ultra preset (with deluxe mapping, relief mapping, etc.) and tonemapping at 1280×720. Of course a real game with plenty of models and particles will differ a bit, but the GLSL version is definitely not a good determinant.

It should be noted that this card is currently plugged in as a secondary card and is rendering through `DRI_PRIME` (so basically transferring the render through PCIe in order to be displayed by another card), which is expected to reduce performance.

This is actually a good example of GL 2.1 card that is brings a very nice playing experience. The 0.55.2 benchmark gets 50 fps on 1024×768 with ultra presets and tone mapping, and 67 fps on 1024×768 with ultra presets and tone mapping, if I reduce images to `256×256` of course.

Actually it looks like this card even does branching properly, as the non-latched implementation also shows the framerate difference on disablement.

So we can't use the GL or GLSL version to detect the card would suffer from tonemapping, we would have to resort on card family detection if we want to disable the feature (that would be really annoying).

But the good thing is that by doing this testing I noticed that framerate was the same with or without tone mapping on `low` and `lowest` presets, and that the render looked the same. So I just got a revelation.

Tone mapping requires `r_highPrecisionRendering` to be enable, and this cvar is latched and will always be (changing it requires to recreate the framebuffer), so we can use that cvar to prevent the compilation of the tonemapping code on `low` and `lowest` presets which are expected to be used low-end cards not supporting tonemapping.

Actually I would prefer to disable the build of the tone mapper when the tone mapper is actually disabled, this is much much cleaner.

VReaperV:
Something like an `#ifdef` for `r_highPrecisionRendering` would be fine?

illwieckz:
> Something like an `#ifdef` for `r_highPrecisionRendering` would be fine?

That's what I just did. This may be an acceptable compromise for now, but this is ugly. 🫣

The day we can do some lazy shader permutation rebuild I will likely revert that.

VReaperV:
> The day we can do some lazy shader permutation rebuild I will likely revert that.

Revert to what?

src/engine/renderer/tr_shade.cpp
```diff
@@ -136,14 +136,29 @@ static void EnableAvailableFeatures()
...
+		{ &glConfig2.textureGatherAvailable, "ARB_texture_gather" },
+		{ &glConfig2.gpuShader4Available, "EXT_gpu_shader4" },
+	};
+
+	for ( auto& e: ssaoRequiredExtensions )
+	{
+		if ( !*e.first )
+		{
+			Log::Warn( "SSAO disabled because %s is not available.", e.second );
+			glConfig2.ssao = false;
```
VReaperV, line 158: Non-blocking, but you can do `break;` here.

illwieckz:
> > The day we can do some lazy shader permutation rebuild I will likely revert that.
> 
> Revert to what?

Revert to my first implementation of the commit:

```glsl
#if defined(r_toneMapping)
…
#endif
```

src/engine/renderer/tr_shade.cpp
illwieckz: This is actually intentional: I prefer the log to be exhaustive for all causes of disablement.

VReaperV: I see.

VReaperV:
> So we can't use the GL or GLSL version to detect the card would suffer from tone mapping, we would have to resort on card family detection if we want to disable the feature (that would be really annoying).

Another option is to run this shader a few times while measuring performance to see if it has an effect, then cache the result in some cvar (and only check again if the detected hardware changes).

PR #1582, opened 2025-03-07
Mac: don't vid_restart when toggling full-screen

slipher:
The keyboard shortcut (Command+F) for toggling fullscreen mode unnecessarily issued vid_restart. It works fine without that as can be seen by just setting the cvar r_fullscreen.

### DIFF
--- a/src/engine/client/cl_keys.cpp
+++ b/src/engine/client/cl_keys.cpp
@@ -511,7 +511,7 @@ static bool DetectBuiltInShortcut( Keyboard::Key key )
 		if ( key == Key::FromCharacter('f') )
 		{
 			Key_ClearStates();
-			Cmd::BufferCommandText("toggle r_fullscreen; vid_restart");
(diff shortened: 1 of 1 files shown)


slipher:
There's an ominous comment in SDL documentation: "Note that for some renderers, this function may trigger an [SDL_RENDER_TARGETS_RESET](https://wiki.libsdl.org/SDL2/SDL_RENDER_TARGETS_RESET) event. Your application should be prepared to handle this event by reuploading textures!" But people say this doesn't happen for OpenGL contexts. Anyway, if there were a possible bug, you could already trigger it by setting the cvar directly.

VReaperV (approved):
Not tested, but code LGTM.

illwieckz:
It was likely a relic of some very old code. For example Tremulous GPP was issuing a `vid_restart` on `Alt+Enter` (likely because of having to switch the overbright mode). This relic still being there is likely related to the fact macOS is less used and it was overlooked.

PR #1584, opened 2025-03-08
ci: use submodules for freetype

illwieckz:
Our submodules may provide symbols not provided by system libraries.


illwieckz:
Bad branch name, see #1585 instead:

- https://github.com/DaemonEngine/Daemon/pull/1585

PR #1433, opened 2024-11-11
WIP: external_deps, use CMake when packages do, update versions

illwieckz:
This is WIP, for now I verified the external_deps built for Linux amd64 and MinGW amd64, I have not tested if the produced files are usable for a working engine build.

- Use CMake when packages provide a `CMakeLists.txt` file
- Factorize much code
- Make a nice list of package base url, so it's easy to click all of them and check for new versions
- Update versions

PR #1585, opened 2025-03-08
ci: use submodules for freetype

illwieckz:
Our submodules may provide symbols not provided by system libraries.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -73,14 +73,14 @@ jobs:
       run: |
         set -x
         sudo apt-get update
-        sudo apt-get -y -q --no-install-recommends install zlib1g-dev libncursesw5-dev libgeoip-dev nettle-dev libgmp-dev libcurl4-gnutls-dev libsdl2-dev libogg-dev libvorbis-dev libopusfile-dev libwebp-dev libjpeg8-dev libpng-dev libfreetype6-dev libglew-dev libopenal-dev ninja-build
+        sudo apt-get -y -q --no-install-recommends install zlib1g-dev libncursesw5-dev libgeoip-dev nettle-dev libgmp-dev libcurl4-gnutls-dev libsdl2-dev libogg-dev libvorbis-dev libopusfile-dev libwebp-dev libjpeg8-dev libpng-dev libglew-dev libopenal-dev ninja-build
         git submodule update --init --recursive
 
         curl -sS https://gitlab.com/illwieckz/git-checkout-modules/raw/master/git-checkout-modules -o ~/git-checkout-modules
         bash ~/git-checkout-modules --update --sub-ref="${GITHUB_HEAD_REF}:has=/sync$" --print
 
         export CMAKE_BUILD_PARALLEL_LEVEL="$(nproc)"
-        cmake -S. -Bbuild -GNinja -DBUILD_CLIENT=ON -DBUILD_TTY_CLIENT=ON -DBUILD_SERVER=ON -DBUILD_DUMMY_APP=ON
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -765,7 +765,7 @@ if (USE_BREAKPAD)
     endif()
 endif()
 
-option(PREFER_EXTERNAL_LIBS "Tries to use system libs where possible." ON)
+option(PREFER_EXTERNAL_LIBS "Tries to use system libs where possible." OFF)
```
slipher, line 768: This shouldn't be changed because Unvanquished doesn't even build in the default configuration that way (it defines the Freetype lib twice).

illwieckz: The submodule Freetype build bug is now fixed.

When building both the engine and a native game, the native game now reuses the Freetype target already sets by the engine.

CMakeLists.txt
```diff
@@ -769,17 +769,21 @@ endif()
 option(PREFER_EXTERNAL_LIBS "Tries to use system libs where possible." ON)
 
 macro(prefer_package LIB_NAME LIB_CMAKE)
-    if (PREFER_EXTERNAL_LIBS AND NOT NACL)
-        find_package(${LIB_NAME})
+	if (NOT ${LIB_NAME}_FOUND)
+		if (PREFER_EXTERNAL_LIBS AND NOT NACL)
```
slipher, line 773: The rest of this file is using spaces, not tabs

CMakeLists.txt
slipher: > The Freetype submodule double build bug is now fixed.

Great!

What's the motivation for the change though? It makes builds longer and people building Daemon themselves can no longer benefit from distro security updates.

illwieckz:
> What's the motivation for the change though? It makes builds longer

The motivation is that we better better strictly control what we build against when doing our own official release builds. We already rebuild every deps using the `external_deps` script in the docker release build. So this changes nothing for the release build, either rebuilding the `external_deps` or rebuilding the submodule is the same (though submodule is less cumbersome).

And when a library is used in both game and engine, like freetype, we better use the exact same freetype library in engine so we don't deal with different bugs and have consistent version to track down.

Then we better build stuff in our CI the closest way we build the release, building the submodule in both cases is the easiest way to achieve that.

Note that for non-Linux systems where we usually ship many prebuilt libraries in external deps archives, we may modify the external_deps script to build from submodule so we ensure prebuilt libraries are exactly the same as submodule ones. This can be implemented above #1433:

>  and people building Daemon themselves can no longer benefit from distro security updates.

This is not true at all. The whole reason why `PREFER_EXTERNAL_LIBS` exists and it is actually implemented at first and all the extra code was written by myself is to keep the possibility for distro packages to build engines fully relying on system libraries, and to allow people to link against their system libraries for whatever reason (binary size, updates…).

If this prevent to benefit from distro security updates or just simply rely on system libraries, the whole `PREFER_EXTERNAL_LIBS`  mechanism could be entirely NUKEd and would have not been implemented to begin with.

slipher:
SGTM to use the submodule in the release script and to make the CI match what is done with the release. I'm just disputing what should be the default value. To benefit players who build their own engine from source, the default should be to use the system libs. Release and CI can customize everything so the default doesn't matter for them.

illwieckz:
I extracted the non “default switching” changes there and there:

- https://github.com/DaemonEngine/Daemon/pull/1617
- https://github.com/Unvanquished/Unvanquished/pull/3353

We better merge them in all case. I'm myself not entirely convinced about what's the best to do for the default, but we can already fix the existing bugs without deciding on the default.

CMakeLists.txt

PR #1587, opened 2025-03-10
Implement shader pipelines

VReaperV:
Requires #1586

This further improves the shader building by using GL_ARB_separate_shader_objects when it is available. This extension allows linking different shader stages separately, thereby decreasing the total amount of linking required.

As an example, on map `station15` with all the shader stuff except liquid mapping and static reflections enabled, it went from 433 permutations/~866 compiled shaders/433 linked programs to 433 permutations/287 compiled shaders/286 linked programs, and cut the total amount of time it took to build shaders almost in half.

I'd had to add some ugly workarounds to make it work with both shader pipelines and regular shader programs due to how our GL uniforms work. Because of that this is currently draft, as I'm thinking of moving all these uniforms to a UBO when shader pipelines are available.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -1070,7 +1070,8 @@ void GLShaderManager::BuildShaderProgram( ShaderProgramDescriptor* descriptor ) 
 	Log::Debug( "Program creation + linking: %i", time );
 }
 
-ShaderProgramDescriptor* GLShaderManager::FindShaderProgram( std::vector<ShaderEntry>& shaders, GLShader* mainShader ) {
+ShaderProgramDescriptor* GLShaderManager::FindShaderProgram( std::vector<ShaderEntry>& shaders, GLShader* mainShader,
+	GLShader::UniformData* uniformData ) {
 	std::vector<ShaderProgramDescriptor>::iterator it = std::find_if( shaderProgramDescriptors.begin(), shaderProgramDescriptors.end(),
 		[&]( const ShaderProgramDescriptor& program ) {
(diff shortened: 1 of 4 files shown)


VReaperV:
> I'd had to add some ugly workarounds to make it work with both shader pipelines and regular shader programs due to how our GL uniforms work. Because of that this is currently draft, as I'm thinking of moving all these uniforms to a UBO when shader pipelines are available.

This would be much easier if we drop support for hardware straight out of Pleistocene.

illwieckz:
It looks like Mesa supports `GL_ARB_separate_shader_objects` on every hardware. I have verified Intel, Nvidia, ATI, Mali, and Videocore. I even seen this extension listed with something as crazily old as an Nvidia RIVA TNT2 (nv05), something we cannot support as we dropped the GL1 renderer 10 years ago anyway. This extension is probably just a matter of API, and is very likely totally unrelated to the hardware.

VReaperV:
> It looks like Mesa supports `GL_ARB_separate_shader_objects` on every hardware. I have verified Intel, Nvidia, ATI/AMD, Arm (Mali) and Broadcom (VideoCore), this included some of the oldest of each brand I own. I even seen this extension listed with something as crazily old as an Nvidia RIVA TNT2 (nv05), something we cannot support as we dropped the GL1 renderer 10 years ago anyway, and Mesa already supported it on it 5 years ago.. This extension is probably just a matter of API implementation, and is very likely totally unrelated to the hardware. It looks like OpenGL made it mandatory only with GL 4.1, but it looks like that doesn't prevented Mesa to implement it on older hardware, even hardware from Pliocene 😅️.

It doesn't make sense for GL1 cards, since they don't support the programmable pipeline anyway.

PR #1588, opened 2025-03-10
Fix UI being affected by subgroup profiler

### LINKED ISSUE #1583: Shader subgroup profiler affecting 2D rendering
Something broke recently, wherein using `r_profilerRenderSubGroups on` results in UI elements using the subgroup profiler as well.

VReaperV:
Fixes #1583.


illwieckz (approved):
LGTM.

PR #1592, opened 2025-03-11
Fix cursor not rendering right in main menu

### LINKED ISSUE #1590: The cursor is black on master
Image

slipher:
Some 2D shaders are automatically assigned CGEN_IDENTITY_LIGHTING. This causes a bug of the cursor being black in the main menu (where tr.identityLighting is 0) and the wrong color if r_overbrightQ3 is used while in a map. Fix by making it behave as CGEN_IDENTITY during 2D rendering.

Fixes #1590.


VReaperV (approved):
That fixes the issue on my end. LGTM.

slipher:
Added a 2nd commit that cleans up another outdated comment about overbright.

PR #1594, opened 2025-03-11
Fix --help output not appearing through a pipe

### LINKED ISSUE #631: `daemon --help` only prints help in interactive terminals
`daemon --help` only prints help in interactive terminals:

```
$ ./daemon --help
Unvanquished 0.52.1 Linux x86_64 May 28 2022
cmdline: --help
Usage: daemon [-OPTION]... [-connect <uri> | +COMMAND...]

Possible options are:
  -h, -help                print this help and exit
  -v, -version             print version and exit
  -homepath <path>         set the path used for user-specific configuration files and downloaded dpk files
  -libpath <path>          set the path containing additional executables and libraries
  -pakpath <path>          add another path from which dpk files are loaded
  -resetconfig             reset all cvars and keybindings to their default value
  -curses                  activate the curses interface
  -set <variable> <value>  set the value of a cvar
  -connect unv://<address>[:<port>]>
                           connect to server at startup
  +<command> <args>        execute an ingame command after startup

Order is important, -options must be set before +commands.
Nothing is read and executed after -connect option and the following URI.
If another instance is already running, commands will be forwarded to it.
```

```
$ ./daemon --help | cat
Unvanquished 0.52.1 Linux x86_64 May 28 2022
cmdline: --help
```

```
$ echo "$(./daemon --help)"
Unvanquished 0.52.1 Linux x86_64 May 28 2022
cmdline: --help
```

```
$ unbuffer ./daemon --help | cat
Unvanquished 0.52.1 Linux x86_64 May 28 2022
cmdline: --help
Usage: ./daemon [-OPTION]... [-connect <uri> | +COMMAND...]

Possible options are:
  -h, -help                print this help and exit
  -v, -version             print version and exit
  -homepath <path>         set the path used for user-specific configuration files and downloaded dpk files
  -libpath <path>          set the path containing additional executables and libraries
  -pakpath <path>          add another path from which dpk files are loaded
  -resetconfig             reset all cvars and keybindings to their default value
  -curses                  activate the curses interface
  -set <variable> <value>  set the value of a cvar
  -connect unv://<address>[:<port>]>
                           connect to server at startup
  +<command> <args>        execute an ingame command after startup

Order is important, -options must be set before +commands.
Nothing is read and executed after -connect option and the following URI.
If another instance is already running, commands will be forwarded to it.
```

slipher:
On *nix using a pipe makes the output fully buffered, as opposed to a TTY where the default is line buffering. We did a fast shutdown without flushing so the output could be lost.

Fixes #631.

### DIFF
--- a/src/engine/framework/System.cpp
+++ b/src/engine/framework/System.cpp
@@ -739,9 +739,11 @@ static void ParseCmdline(int argc, char** argv, cmdlineArgs_t& cmdlineArgs)
 				"Nothing is read and executed after -connect option and the following URI.\n"
 				"If another instance is already running, commands will be forwarded to it.\n"
 			);
+			FS::FlushAll();
 			OSExit(0);
 		} else if (!strcmp(argv[i], "--version") || !strcmp(argv[i], "-version") || !strcmp(argv[i], "-v")) {
 			printf(PRODUCT_NAME " " PRODUCT_VERSION "\n");
+			FS::FlushAll();
 			OSExit(0);
 		} else if (!strcmp(argv[i], "-set")) {
 			if (i >= argc - 2) {


illwieckz (approved):
Thank you very much. It works.

LGTM.

PR #1595, opened 2025-03-12
renderer: disable EXT_gpu_shader4 on GLSL 1.20

illwieckz:
Disable EXT_gpu_shader4 on GLSL 1.2.

`EXT_gpu_shader4` behaves slightly differently on GLSL 1.20.

See: https://gitlab.freedesktop.org/mesa/mesa/-/issues/12803#note_2819461

The patch is similar to: 

- https://github.com/DaemonEngine/Daemon/pull/1576

Also does a bit of rewording of #1576.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1970,6 +1970,8 @@ static bool LoadExt( int flags, bool hasExt, const char* name, bool test = true 
 	return false;
 }
 
+#define SILENTLY_CHECK_EXTENSION( ext ) ( GLEW_##ext )
+
 #define LOAD_EXTENSION(flags, ext) LoadExt(flags, GLEW_##ext, #ext)
 
 #define LOAD_EXTENSION_WITH_TEST(flags, ext, test) LoadExt(flags, GLEW_##ext, #ext, test)
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -2600,7 +2600,7 @@ static void GLimp_InitExtensions()
 		&& glConfig2.geometryCacheAvailable;
 
 	// This requires GLEW 2.2+, so skip if it's a lower version
-#ifdef GL_KHR_shader_subgroup
+#if defined(GLEW_KHR_shader_subgroup)
 	// not required by any OpenGL version
 	glConfig2.shaderSubgroupAvailable = LOAD_EXTENSION_WITH_TEST( ExtFlag_NONE, KHR_shader_subgroup, r_khr_shader_subgroup.Get() );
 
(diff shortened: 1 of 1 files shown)


src/engine/sys/sdl_glimp.cpp
```diff
@@ -2171,8 +2175,21 @@ static void GLimp_InitExtensions()
 		}
 	}
 
+	bool gpuShader4Enabled = r_arb_shader_draw_parameters.Get();
+
+	if ( gpuShader4Enabled
+		&& GL_EXT_gpu_shader4
```
VReaperV, line 2181: Isn't this just the GLEW define that we don't need to check for?

illwieckz: I used `GLEW_` prefix for consistency with `LOAD_EXTENSION()` macros.

illwieckz: In fact I added a `SILENTLY_CHECK_EXTENSION()` macro that makes it obvious.

illwieckz: From `GL/glew.h`:

```c++
#define GLEW_EXT_gpu_shader4 GLEW_GET_VAR(__GLEW_EXT_gpu_shader4)
GLEW_VAR_EXPORT GLboolean __GLEW_EXT_gpu_shader4;
```

The `GLEW_EXT_gpu_shader4` symbol is a definition that can be used at build time to know if GLEW implements the querying of that extension.

When it is defined, this definition expands to a boolean which contains the runtime status of the extension being implemented or not by the driver.

illwieckz: Actually you were right about the `GL_` prefixed symbol: it expands to 1 when the feature is implemented in GLEW, but says nothing about the runtime existence of it. The `GLEW_` prefixed symbol does both.

illwieckz: We can use either `GL_*` or `GL_*` in `if defined(x)`, but only `GLEW_*` is usable at run time.

I added a commit renaming the ifdef for `GL_KHR_shader_subgroup` to `GLEW_KHR_shader_subgroup` for consistency because I guess it works as well as long as it is an `ifdef`.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -2173,8 +2177,21 @@ static void GLimp_InitExtensions()
 		}
 	}
 
+	bool gpuShader4Enabled = r_arb_shader_draw_parameters.Get();
```
VReaperV, line 2180: Wrong cvar used?

illwieckz: Oupsie, fixed.

VReaperV (approved):
LGTM

PR #1597, opened 2025-03-13
DaemonArchitecture: NUKE x86 revision detection

### LINKED ISSUE #1596: Architecture detection fails with recent 32-bit MinGW
With MSYS GCC 14.2 targeting (32-bit) x86, the architecture detection reports an "unknown" architecture (and consequently doesn't fetch external_deps). It seems recent MinGW GCC versions by default advertise the arch as i386 while previous ones claimed i686. DaemonArchitecture.cpp is compiled without any of the configured compiler flags, so the default is what we get. Building it without the flags doesn't work right since they can change the architecture version.

slipher:
Always assume i686, instead of trying to use the numeric value of Q_PROCESSOR_X86, which isn't accurate in our build.

For architecture detection the compiler is invoked without any of the flags that we usually pass. So the -march=XXX, etc. flags are not there. This means the architecture revision that the compiler advertises during architecture detection may not be the same as in the real build. You may react, "just do the architecture detection with all flags on", but this would probably lead to chicken-and-egg problems as architecture is used to choose flags. In principle it could be done with two passes of architecture detection but implementing that would obviously be a waste of time.

Fixes #1596.

### DIFF
--- a/cmake/DaemonArchitecture/DaemonArchitecture.cpp
+++ b/cmake/DaemonArchitecture/DaemonArchitecture.cpp
@@ -80,7 +80,10 @@ platform variants we cannot support anyway. */
 #elif defined(Q_PROCESSOR_X86_64)
 	#pragma message("DAEMON_ARCH_amd64")
 
-#elif defined(Q_PROCESSOR_X86_32) && Q_PROCESSOR_X86 >= 6
+#elif defined(Q_PROCESSOR_X86_32)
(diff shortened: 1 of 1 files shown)


illwieckz:
Right, we only need to know it's 32-bit, it is us who chose the revision.

illwieckz (approved):
LGTM.

PR #1602, opened 2025-03-14
Don't print some network information on startup

slipher:
Don't print the hostname and network addresses that may make people worry about posting logs with confidential information. Instead the `showip` command can be used. Make the command available for all engines.

Fixes https://github.com/Unvanquished/Unvanquished/issues/1573.


illwieckz (approved):
LGTM.

PR #1605, opened 2025-03-17
client: make the error popup disableable

illwieckz:
Make the error popup disableable.

The error popup is very useful for end users, to let them know what happened when the engine has to stop in an unexpected way.

But such error popup turns testing into torture, requiring a manual action after every test run hits a fatal error, and requiring a direct input access to the test bed.

This addes a cvar names `client.errorPopup` which is enabled by default. When disabled, the error popup never is displayed, letting the engine stop without manual intervention on the computer running the engine.


src/engine/client/ClientApplication.cpp
```diff
@@ -126,7 +128,7 @@ class ClientApplication : public Application {
 
         void Shutdown(bool error, Str::StringRef message) override {
             #if defined(_WIN32) || defined(BUILD_GRAPHICAL_CLIENT)
-                if (error) {
+                if (error && client_errorPopup.Get()) {
```
DolceTriade, line 131: should we print to stdout instead?

illwieckz: The code calling for `Shutdown()` always prints the message as a warning as far as I know, before actually calling `Shutdown()`.

illwieckz: https://github.com/DaemonEngine/Daemon/blob/7d15e9236bb8fd40f027d1d8248fcb42f65f5e55/src/engine/framework/System.cpp#L551-L562

DolceTriade (approved):

src/engine/client/ClientApplication.cpp
```diff
@@ -42,6 +42,8 @@ extern SDL_Window *window;
 #endif
 #endif
 
+static Cvar::Cvar<bool> client_errorPopup("client.errorPopup", "Enable the error popup window", Cvar::TEMPORARY, true);
```
slipher, line 45: Remove TEMPORARY? TEMPORARY, like ARCHIVE, has little reason to be used now that cvar commands don't automatically archive things.

illwieckz: Removed.

PR #1607, opened 2025-03-17
cmake: make sure the dll game is built as module, not as static library

illwieckz:
Make sure the dll game is built as module, not as static library.

Fixup for:

This wasn't caught by the CI, but a bug in #1555 made the dll game being compiled to `cgame-native-dll.a` instead of `cgame-native-dll.so`. Also, this hid another bug where CMake would overwrite `COMPILE_DEFINITIONS` and then lose the `BUILD_VM_IN_PROCESS` definition.


illwieckz:
I merge this quickly because it fixes a regression breaking the build, and we better not merge anything between the breakage and the fix, to limit the impact on future bisects.

PR #1608, opened 2025-03-17
renderer: rewrite `GLimp_LogComment()` calls

illwieckz:
Rewrite `GLimp_LogComment()` calls.

Also don't require an ending `\n` on every call.

Also only call `GLimp_LogComment()` when the function doesn't return early.

No more noisy:

```c++
if ( r_logFile->integer )
{
	GLimp_LogComment( va( "bla: %d\n", bla ) );
}

if ( condition )
{
	return;
}
```

Just this:

```c++
if ( condition )
{
	return;
}

GLIMP_LOGCOMMENT( "bla: %d", bla );
```

_Edit:_ `GLIMP_LOGCOMMENT()` is a macro wrapping both the `if ( r_logFile->integer )` test and a `Str::Format()` call.

### DIFF
--- a/src/engine/renderer/gl_shader.h
+++ b/src/engine/renderer/gl_shader.h
@@ -685,11 +685,8 @@ protected:
 		}
 
 #if defined( LOG_GLSL_UNIFORMS )
-		if ( r_logFile->integer )
-		{
-			GLimp_LogComment( va( "GLSL_SetUniform1i( %s, shader: %s, value: %d ) ---\n",
-				this->GetName(), _shader->GetName().c_str(), value ) );
-		}
+		GLIMP_LOGCOMMENT( "GLSL_SetUniform1i( %s, shader: %s, value: %d )",
+			this->GetName(), _shader->GetName().c_str(), value );
 #endif
 
 		if ( _shader->UseMaterialSystem() && !_global ) {
@@ -738,10 +735,8 @@ class GLUniform1ui : protected GLUniform {
 		}
 
 #if defined( LOG_GLSL_UNIFORMS )
-		if ( r_logFile->integer ) {
-			GLimp_LogComment( va( "GLSL_SetUniform1i( %s, shader: %s, value: %d ) ---\n",
-				this->GetName(), _shader->GetName().c_str(), value ) );
-		}
+		GLIMP_LOGCOMMENT( "GLSL_SetUniform1i( %s, shader: %s, value: %d )",
+			this->GetName(), _shader->GetName().c_str(), value );
 #endif
 
 		if ( _shader->UseMaterialSystem() && !_global ) {
@@ -789,10 +784,8 @@ class GLUniform1Bool : protected GLUniform {
 		}
 
 #if defined( LOG_GLSL_UNIFORMS )
-		if ( r_logFile->integer ) {
-			GLimp_LogComment( va( "GLSL_SetUniform1i( %s, shader: %s, value: %d ) ---\n",
-				this->GetName(), _shader->GetName().c_str(), value ) );
-		}
+		GLIMP_LOGCOMMENT( "GLSL_SetUniform1i( %s, shader: %s, value: %d )",
+			this->GetName(), _shader->GetName().c_str(), value );
 #endif
 
 		if ( _shader->UseMaterialSystem() && !_global ) {
@@ -843,11 +836,8 @@ protected:
 		}
 
 #if defined( LOG_GLSL_UNIFORMS )
-		if ( r_logFile->integer )
-		{
-			GLimp_LogComment( va( "GLSL_SetUniform1f( %s, shader: %s, value: %f ) ---\n",
-				this->GetName(), _shader->GetName().c_str(), value ) );
-		}
+		GLIMP_LOGCOMMENT( "GLSL_SetUniform1f( %s, shader: %s, value: %f )",
+			this->GetName(), _shader->GetName().c_str(), value );
 #endif
 
 		if ( _shader->UseMaterialSystem() && !_global ) {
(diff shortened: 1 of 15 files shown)


src/engine/renderer/tr_backend.cpp
```diff
@@ -2966,10 +2880,8 @@ void RB_RenderGlobalFog()
 	{
 		fog_t* fog = &tr.world->fogs[ tr.world->globalFog ];
 
-		if ( r_logFile->integer )
-		{
-			GLimp_LogComment( va( "--- RB_RenderGlobalFog( fogNum = %i, originalBrushNumber = %i ) ---\n", tr.world->globalFog, fog->originalBrushNumber ) );
-		}
+		GLimp_LogComment( "--- RB_RenderGlobalFog( fogNum = %i, originalBrushNumber = %i ) ---",
```
VReaperV, line 2883: `originalBrushNumber` can be removed here, as it's always -1.

src/engine/renderer/tr_backend.cpp
```diff
@@ -5640,7 +5544,7 @@ const RenderCommand *SwapBuffersCommand::ExecuteSelf( ) const
 		RB_ShowImages();
 	}
 
-	GLimp_LogComment( "***************** RB_SwapBuffers *****************\n\n\n" );
+	GLimp_LogComment( "***************** RB_SwapBuffers *****************" );
```
VReaperV, line 5547: This should also be removed.

src/engine/renderer/tr_init.cpp
```diff
@@ -123,7 +123,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 		Cvar::NONE, 1, 0, 2);
 
 	cvar_t      *r_checkGLErrors;
-	cvar_t      *r_logFile;
+	Cvar::Cvar<bool> r_logFile( "r_logFile", "Emit GL logs", Cvar::CHEAT, false );
```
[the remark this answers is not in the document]

src/engine/renderer/tr_shade.cpp
```diff
@@ -2321,7 +2314,7 @@ void Tess_ComputeColor( shaderStage_t *pStage )
 	float blue;
 	float alpha;
 
-	GLimp_LogComment( "--- Tess_ComputeColor ---\n" );
+	GLimp_LogComment( "--- Tess_ComputeColor ---" );
```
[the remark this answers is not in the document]
illwieckz: Done (both).

src/engine/renderer/tr_backend.cpp
illwieckz: Done.

src/engine/renderer/tr_backend.cpp
illwieckz: Done

src/engine/renderer/gl_shader.h
```diff
@@ -685,11 +685,8 @@ class GLUniform1i : protected GLUniform
 		}
 
 #if defined( LOG_GLSL_UNIFORMS )
-		if ( r_logFile->integer )
-		{
-			GLimp_LogComment( va( "GLSL_SetUniform1i( %s, shader: %s, value: %d ) ---\n",
-				this->GetName(), _shader->GetName().c_str(), value ) );
-		}
+		GLimp_LogComment( "GLSL_SetUniform1i( %s, shader: %s, value: %d )",
```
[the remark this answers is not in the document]
illwieckz: Removed too.

DolceTriade:
Hm, maybe reaper and slipher can disagree, but I'm not sure I agree with the premise of the PR.

Keep in mind that your change isn't 1:1 with the old behavior.

```
GLimp_LogComment("%d", SomeVeryExpensiveCall())
```

`SomeVeryExpensiveCall` will not be run unconditionally. We do have `DoDebugCode` for this reason, but in that case, I'd like to see all calls to GLimp_LogComment wrapped in that function (which from a noisy-ness perspective, does not really get us much).

illwieckz:
@DolceTriade this is the part you look for:

```c++
inline bool GLimp_isLogging() {
	return r_logFile.Get() && GLEW_ARB_debug_output;
}

#define GLimp_LogComment( ... ) { \
	if ( GLimp_isLogging() ) \
	{ \
		GLimp_LogComment_( Str::Format( __VA_ARGS__ ) ); \
	} \
}
```

src/engine/renderer/tr_local.h
```diff
@@ -3383,7 +3382,19 @@ inline bool checkGLErrors()
...
+	inline bool GLimp_isLogging() {
+		return r_logFile.Get() && GLEW_ARB_debug_output;
+	}
+
+	#define GLimp_LogComment( ... ) { \
+		if ( GLimp_isLogging() ) \
+		{ \
+			GLimp_LogComment_( Str::Format( __VA_ARGS__ ) ); \
+		} \
+	}
```
illwieckz, line 3394: @DolceTriade here

illwieckz:
Well, I can rename `GLimp_LogComment` as upper case `GLIMP_LOGCOMMENT` to hint this is now a macro.

illwieckz:
> Well, I can rename `GLimp_LogComment` as upper case `GLIMP_LOGCOMMENT` to hint this is now a macro.

Now done.

illwieckz:
I also added a commit asserting that the `GLimp_LogComment()` function is never called directly and not wrapped by `GLIMP_LOGCOMMENT()`.

There is actually one place where we can call it directly, but the extra calls of `GLimp_isLogging()` will be optimized out by the compiler anyway.

So we better make sure `GLIMP_LOGCOMMENT()` is always used instead.

src/engine/renderer/tr_local.h
DolceTriade: I see

illwieckz:
> I see. I missed that it was converted into a macro

My fault, I forgot to mention it in the first post. 😊️

src/engine/renderer/tr_local.h
```diff
@@ -3383,7 +3382,19 @@ inline bool checkGLErrors()
 	void     GLimp_SyncRenderThread();
 	void     GLimp_WakeRenderer( void *data );
 
-	void     GLimp_LogComment( const char *comment );
+inline bool GLimp_isLogging() {
+	return r_logFile.Get() && GLEW_ARB_debug_output;
+}
+
+// Never use GLimp_LogComment() directly, use the GLIMP_LOGCOMMENT() wrapper instead.
+void GLimp_LogComment( std::string comment, bool wrapped = false );
```
slipher, line 3390: `bool wrapped` seems like overkill. Just give the function a name hinting it is not intending to be used normally, like with the trailing underscore on `GL_CheckErrors_`

illwieckz: Done.

VReaperV (approved):
LGTM

PR #1609, opened 2025-03-18
framework: sort the output of the `listCvars` command

illwieckz:
Sort the output of the `listCvars` command.

Inspired from a similar code from the `listCmds` command.

The `matchesValues` vector was unused so it is dropped and not ported to the new code.


slipher:
LGTM

PR #1610, opened 2025-03-18
NUKE surface sorting by fog

VReaperV:
#1408 had removed the hacky way of drawing fog surfaces and made them into actual stages. This change removes the sorting by fog index since it now has no positive effect (and potentially has a negative one, preventing surfaces from being merged). The fognum bits are redistributed towards drawsurf indexes and entity count.

This will also allow increasing `MAX_REF_ENTITIES`.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1602,8 +1602,7 @@ enum class ssaoMode {
 	// 1. shaderNum
 	// 2. lightmapNum
 	// 3. entityNum
-	// 4. fogNum
-	// 5. index
+	// 4. index
 
 	static const uint64_t SORT_INDEX_BITS = 20;
 	static const uint64_t SORT_ENTITYNUM_BITS = 13;


slipher:
> This change removes the sorting by fog index since it now has no positive effect (and potentially has a negative one, preventing surfaces from being merged).

Can you really merge surfaces of different fogs though? It seems like they have some parameters which could be different such as `fog->color` and `fog->tcScale`.

VReaperV:
> Can you really merge surfaces of different fogs though? It seems like they have some parameters which could be different such as `fog->color` and `fog->tcScale`.

You can't merge the actual fog surfaces, but you can merge the "main" surfaces that emitted them. Although I'll need to make the main surfaces not use the actual fognum for that to work, otherwise the backend still won't merge them.

slipher:
Right, now I get it. You want to sort by fog num for fog surfaces, but not surfaces in the fog. LGTM

To get fog surfaces sorted/merged properly, maybe it would be easiest to just generate a `shader_t` for each fog so that we could take advantage of existing machinery for sorting.

VReaperV:
That is correct, yeah.

I've thought of generating the `shader_t`s too, that should work.

VReaperV:
I've added two more commits that set fogNum on main surfaces to 0 so they can be merged, and fix a comment on drawSurf sort.

Issue #1612, opened 2025-03-18
Weird `Unhandled exception (St12length_error): basic_string::_M_replace_aux ` error


illwieckz:
While testing a branch attempting to fix #1611, I get this error:

```
Warn: Unhandled exception (St12length_error): basic_string::_M_replace_aux 
```

My branch only modifies GLSL so this my change doesn't cause this bug, this bug is hiding in our current code base.

Unfortunately this exception raises an annoying error popup instead of returning to debugger, so I don't know how to backtrace it.

slipher:
> Unfortunately this exception raises an annoying error popup instead of returning to debugger, so I don't know how to backtrace it.

Configure the debugger to break when `length_error` is thrown.

illwieckz:
I commented out the catcher, and got this:

```
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_M_replace_aux

Thread 1 "daemon" received signal SIGABRT, Aborted.

Thread 1 (Thread 0x7c6a465e5b00 (LWP 1863202) "daemon"):
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007c6a5264527e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007c6a526288ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007c6a52aa5ff5 in __gnu_cxx::__verbose_terminate_handler () at ../../../../src/libstdc++-v3/libsupc++/vterminate.cc:95
#6  0x00007c6a52abb0da in __cxxabiv1::__terminate (handler=<optimized out>) at ../../../../src/libstdc++-v3/libsupc++/eh_terminate.cc:48
#7  0x00007c6a52aa5a55 in std::terminate () at ../../../../src/libstdc++-v3/libsupc++/eh_terminate.cc:58
[...]

PR #1614, opened 2025-03-19
renderer: catch shader source processing exceptions and print raw source

illwieckz:
Catch shader source processing exceptions and print raw source if something bad happened while processing the source.

See #1612:

- https://github.com/DaemonEngine/Daemon/issues/1612

This doesn't fix the underlying cause of the processing error, but at least the engine doesn't crash, the source get printed, the error message get printed, and it is possible to debug the GLSL shader code in some kind of degraded mode.


slipher:
I think we should just fix the text processing code. That code is a good candidate for unit testing.

PR #1617, opened 2025-03-24
cmake: make possible to build both game and engine against the engine Freetype submodule

illwieckz:
Make possible to build both game and engine against the engine Freetype submodule.

Extracted from #1585:

- https://github.com/DaemonEngine/Daemon/pull/1585

Without the controversial change about default building.

See also:

- https://github.com/Unvanquished/Unvanquished/pull/3353

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -778,16 +778,20 @@ endif()
 option(PREFER_EXTERNAL_LIBS "Tries to use system libs where possible." ON)
 
 macro(prefer_package LIB_NAME LIB_CMAKE)
-    if (PREFER_EXTERNAL_LIBS AND NOT NACL)
-        find_package(${LIB_NAME})
+    if (NOT ${LIB_NAME}_FOUND)
+        if (PREFER_EXTERNAL_LIBS AND NOT NACL)
+            find_package(${LIB_NAME})
+
+            if (NOT ${LIB_NAME}_FOUND)
+                message(WARNING "PREFER_EXTERNAL_LIBS is enabled but external ${LIB_NAME} is not found, falling back to vendored ${LIB_NAME}.")
+            endif()
+        endif()
 
         if (NOT ${LIB_NAME}_FOUND)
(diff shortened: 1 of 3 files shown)


slipher:
LGTM

slipher:
Wait, what is FREETYPE_INTERNAL_ZLIB supposed to mean? We have the submodule configured so that there is no zlib dependency at all, right? The zlib dependency only comes in now for our poorly built Windows or Mac external_deps Freetype.

illwieckz:
Freetype ships with an internal zlib, this is good enough for the nacl game because this is the only usage of zlib in the nacl game, so we use `FREETYPE_INTERNAL_ZLIB`, it prevents us to require a zlib submodule just to build the nacl game.

But then the engine may share zlib with freetype and libpng, so we don't use the internal freetype zlib to avoid having two zlib in the engine binary (they don't conflict as far as I know but it's just stupid to have zlib twice).

illwieckz:
Hmm, rereading the code, we may use `FREETYPE_INTERNAL_ZLIB` only with NACL then.

slipher:
We disable Brotli, bzip2, Harfbuzz, and PNG which means that we won't have a zlib dependency according to https://stackoverflow.com/a/29456684.

illwieckz:
The `configure`'s `--with-zlib=no` option also doesn't allow to entirely disable zlib:

```
$ cd libs/freetype
$ ./autogen.sh
$ ./configure -h
…
 --with-zlib=[yes|no|auto]
                          use system zlib instead of internal library
                          [default=auto]
…
```

slipher:
I read the Freetype CMakeLists.txt and it doesn't have internal zlib. I think it only implements a subset of the options of the configure script. When using CMake, if you disable zlib, there is really no zlib at all. `make VERBOSE=1 freetype` seems to confirm this.

slipher:
I was wrong, it does implement internal zlib. I couldn't see it being built because it works by #including many `.c` source files. No, they couldn't just build it normally, gotta make hundreds of lines of patches to zlib so we can #include it!

There *is* an `FT_CONFIG_OPTION_USE_ZLIB` that we could theoretically use to turn off zlib. I guess we'd have to make our own customized `ftoptions.h`.

slipher:
LGTM

PR #1621, opened 2025-03-25
cmake/DaemonGame: do not reset BUILD_GAME_NACL in cmake subproject

illwieckz:
Mysteriously, a commit discussed in #1555 and implemented before merging was missing.

Another bug slipped in resetting `BUILD_GAME_NACL`.

Fixup for:

Fix Unvanquished/Unvanquished#3355:

- https://github.com/Unvanquished/Unvanquished/issues/3355


illwieckz:
Unfortunately the CI could not catch that, because everything to build was built successfully, but some targets were not added, so not built at all.

PR #1624, opened 2025-03-26
NUKE CMake dependent options

slipher:
cmake_dependent_option, as implemented in the GUI/TUI, is a pointless misfeature. It could be useful if it, say, showed the option as grayed out and then enabled itself as soon as the dependee changed. But what it actually does is hide the dependent option completely, and not show it until you reconfigure. So it's just a nuisance when you want to use the option: you have to configure multiple times. The goal of hiding options that are not relevant is hopeless anyway since there are always zillions of options that don't affect the current build.

We had just 2 cmake_depend_option's, USE_SMP and USE_SLIM_LTO.

Fixes https://github.com/Unvanquished/Unvanquished/issues/1921.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,14 +117,16 @@ if(NOT DAEMON_EXTERNAL_APP)
         option(USE_MUMBLE "Build Daemon with mumblelink sumpport" ON)
     endif()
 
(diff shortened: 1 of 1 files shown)


illwieckz:
I also seen CMake warnings about them, like they are not done properly or so.

I agree with the principle of removing the option dependency. We don't need that.

illwieckz (approved):
LGTM.

PR #1625, opened 2025-03-27
Fix NaCl build breakage

slipher:
FORK=1 is needed to know whether the nacl-vms target has been created already.

### DIFF
--- a/cmake/DaemonGame.cmake
+++ b/cmake/DaemonGame.cmake
@@ -38,6 +38,7 @@ option(BUILD_GAME_NATIVE_DLL "Build the shared library files, mostly useful for 
 # can be loaded by daemon with vm.[sc]game.type 2
 option(BUILD_GAME_NATIVE_EXE "Build native executable, which might be used for better performances by server owners" OFF)
 
+include(ExternalProject)
 include(DaemonBuildInfo)
 include(DaemonPlatform)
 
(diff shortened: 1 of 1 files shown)


slipher:
Also this fixes another regression that you can't build the sgame and cgame at the same time...

cmake/DaemonGame.cmake
```diff
@@ -171,28 +172,23 @@ function(GAMEMODULE)
...
 
-		if (BUILD_GAME_NATIVE_EXE)
-			buildGameModule("native-exe")
-		endif()
+	if (BUILD_GAME_NATIVE_EXE)
+		buildGameModule("native-exe")
 	endif()
 
-	if (FORK EQUAL 1)
+	if (NOT FORK) # create the nacl-vms target only the first time that GAMEMODULE is called
```
illwieckz, line 183: You don't need to set the PARENT_SCOPE if you do:

```cmake
if (FORK EQUAL 1 AND NOT TARGET nacl-vms)
```

slipher:
Well I tried to preserve your notion of making it extensible to multiple gamelogic toolchains with multiple sub-invocations. But if you don't care about that now we could check just `nacl-vms` instead

illwieckz:
Instead of `nacl-vms` we could just name that `sub-vms`, the various toolchains (wasm one, etc.) could just be integrated in the exact same places the NaCl ones are.

cmake/DaemonGame.cmake
illwieckz: In fact that would activate the following `else()`.

slipher:
Bored of bikeshedding now

illwieckz:
It was not a suggestion to do this in this PR anyway (it would break the docker release script if I'm right), just me sharing some thoughts on what we may do in the future.

PR #1629, opened 2025-03-28
Merging part 1 of external_deps work from #1433

slipher:
For my convenience as reviewer, I would like to break off digestible pieces of #1433. I will take on the responsibility of rebasing to solve merge conflicts and ensuring that the final state after all patches is the same as the original. Everything in this PR was written by illwieckz and now has my LGTM.

Tested using `--download-only` that all 20 packages can be downloaded (for at least one target). Also tested the new `--require-theirs`.

I have taken the liberty of removing a trailing slash from NASM_BASEURL.


illwieckz:
LGTM I guess.

PR #1630, opened 2025-03-28
Print error when curl command not found

slipher:
Merges without conflict with #1433.


slipher:
Added another commit to fix the case of a log message messed up by the commit fixing building on Mac from #1642.

VReaperV (approved):
LGTM

PR #1631, opened 2025-03-30
Move entityType_t in gamelogic

VReaperV:
Gamelogic-side pr: https://github.com/Unvanquished/Unvanquished/pull/3359

Patch shared by freem, this moves `entityType_t` to gamelogic, since the engine doesn't need to know about it.

This should be fine to merge into master, since the ABI is still the same and merge conflicts from this change are unlikely.

### DIFF
--- a/src/engine/qcommon/q_shared.cpp
+++ b/src/engine/qcommon/q_shared.cpp
@@ -288,45 +288,6 @@ int Com_HashKey( char *string, int maxlen )
 /*
 ============================================================================
 
-q_shared.h-enum to name conversion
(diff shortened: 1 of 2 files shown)


slipher:
If you push the branches to the origin repos with a name ending in `/sync` (same in both repos), we can see whether it compiles. (may have to re-create the PRs)

illwieckz (approved):

illwieckz:
CI was run in https://github.com/Unvanquished/Unvanquished/pull/3362

PR #1635, opened 2025-04-01
Get rid of a bunch of duplicate code in map loading and srf*_t

VReaperV:
I've run into a bunch of duplicate code while making some improvements to map processing, any improvements to which required to add more such duplication, while the code underneath is the exact same.

This pr NUKES `srfTriangles_t` and `srfSurface_t`, and moves all the common information between them, `srfGridMesh_t`, and `srfVBOMesh_t `, into `srfGeneric_t`.

The common parts of `ParseFace()` and `ParseTriSurf()` are moved into a new `ParseTriangleSurface()`, which is all of it except the plane calculation. The comment about `.lwo` models in `ParseTriSurf()` is irrelevant, because we don't support those.

`ParseTriangleSurface()` and `ParseMesh()` now use `SphereFromBounds()` instead of `FinishGenericSurface()`, because the only difference between the 2 was just duplicated code.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -2605,7 +2605,7 @@ static void R_CreateWorldVBO()
 		if ( *surface->data == surfaceType_t::SF_FACE || *surface->data == surfaceType_t::SF_GRID
 			|| *surface->data == surfaceType_t::SF_TRIANGLES )
 		{
-			srfSurfaceFace_t* srf = ( srfSurfaceFace_t * ) surface->data;
+			srfGeneric_t* srf = ( srfGeneric_t* ) surface->data;
 
 			numVerts += srf->numVerts;
 			numTriangles += srf->numTriangles;
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -746,62 +746,6 @@ static void R_LoadVisibility( lump_t *l )
 
 //===============================================================================
 
-/*
-===============
-ShaderForShaderNum
-===============
-*/
-static shader_t *ShaderForShaderNum( int shaderNum )
-{
-	shader_t  *shader;
-	dshader_t *dsh;
-
-	shaderNum = LittleLong( shaderNum ) + 0;  // silence the warning
-
-	if ( shaderNum < 0 || shaderNum >= s_worldData.numShaders )
-	{
-		Sys::Drop( "ShaderForShaderNum: bad num %i", shaderNum );
-	}
-
-	dsh = &s_worldData.shaders[ shaderNum ];
-
-	shader = R_FindShader( dsh->shader, shaderType_t::SHADER_3D_STATIC, RSF_DEFAULT );
-
-	// if the shader had errors, just use default shader
-	if ( shader->defaultShader )
-	{
-		return tr.defaultShader;
-	}
-
-	return shader;
-}
-
-/*
-SphereFromBounds() - ydnar
-creates a bounding sphere from a bounding box
-*/
-
-static void SphereFromBounds( vec3_t mins, vec3_t maxs, vec3_t origin, float *radius )
-{
-	vec3_t temp;
-
-	VectorAdd( mins, maxs, origin );
-	VectorScale( origin, 0.5, origin );
-	VectorSubtract( maxs, origin, temp );
-	*radius = VectorLength( temp );
-}
-
-/*
-FinishGenericSurface() - ydnar
-handles final surface classification
-*/
-
-static void FinishGenericSurface( dsurface_t *ds, srfGeneric_t *gen, vec3_t pt )
-{
-	// set bounding sphere
-	SphereFromBounds( gen->bounds[ 0 ], gen->bounds[ 1 ], gen->origin, &gen->radius );
-}
-
 // Generate the skybox mesh and add it to world
 static void FinishSkybox() {
 	// Min and max coordinates of the skybox cube corners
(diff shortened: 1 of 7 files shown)


VReaperV:
CI fail is unrelated, and is due to Ubuntu 20.04 being phased out on azure pipelines.

src/engine/renderer/tr_local.h
```diff
@@ -1706,17 +1706,15 @@ enum class ssaoMode {
...
 // as soon as it is called
-	struct srfPoly_t
-	{
+	struct srfPoly_t {
 		surfaceType_t surfaceType;
-		qhandle_t     hShader;
-		int16_t       numVerts;
-		int16_t       fogIndex;
-		polyVert_t    *verts;
+		qhandle_t hShader;
```
slipher, line 1711: Please remove random whitespace changes from `Move some common stuff from srf* to srfGeneric_t `

slipher (commented):
There are some tiny model lighting changes from `Move some common stuff from srf* to srfGeneric_t` and `Move duplicate code from ParseFace() and ParseTriSurf() into ParseTriangleSurface()`. I will try to check later if those diffs are anything.

VReaperV:
> There are some tiny model lighting changes from `Move some common stuff from srf* to srfGeneric_t` and `Move duplicate code from ParseFace() and ParseTriSurf() into ParseTriangleSurface()`. I will try to check later if those diffs are anything.

For BSP models? I don't think other models can even be affected by this.

slipher:
No, it was a buildable model.

VReaperV:
That's weird... Buildings should only ever use `srfVBOMD5Mesh_t` or `srfVBOMDVMesh_t`.

src/engine/renderer/tr_local.h
VReaperV: Done.

slipher:
I can reproduce a similar small diff by adding unused fields to `srfGeneric_t`, e.g.

```diff
diff --git a/src/engine/renderer/tr_local.h b/src/engine/renderer/tr_local.h
index fe658c0b2..c3c4a9b3d 100644
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1738,9 +1738,15 @@ enum class ssaoMode {
        {
                surfaceType_t surfaceType;
 
+               int argle;
+
                // culling information
                vec3_t   bounds[ 2 ];
+
+               int bargle;
+
                vec3_t   origin;
+
                float    radius;
        };
```

Maybe the layout is assumed somewhere else.

slipher:
Without/with the above patch, the lower part of the Overmind looks a bit more brightly lit on the before side.

unvanquished-nano-overmind

unvanquished-nano-overmind

slipher:
The lighting diffs reproduce with different OS, compiler, and GPU.

I tried getting rid of `srfGeneric_t` and commenting out or rewriting all places that use it. After that I still get the same diff by adding extra fields in `srfVBOMesh_t` only.

The diffs go away if I disable `r_deluxeMapping`.

slipher:
I found the cause of that stuff in https://github.com/DaemonEngine/Daemon/pull/1639.

VReaperV:
Thanks, that is very helpful!

src/engine/renderer/tr_local.h
```diff
@@ -1736,97 +1734,53 @@ enum class ssaoMode {
...
-		float  lodRadius;
-		int    lodFixed;
-		int    lodStitched;
+		cplane_t plane;
 
-		// triangle definitions
-		int           width, height;
-		float         *widthLodError;
-		float         *heightLodError;
+		int numVerts;
```
slipher, line 1747: Please add comments to indicate which surface types each field is defined in. It's really annoying to have fields that are only valid in a random subset of the objects and not know which ones.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -887,152 +887,127 @@ static void FinishSkybox() {
...
-		for( i = 0, tri = cv->triangles; i < numTriangles; i++, tri++ ) {
-			int minVertex = std::min( std::min( components[ tri->indexes[ 0 ] ].minVertex,
-						  components[ tri->indexes[ 1 ] ].minVertex ),
-					     components[ tri->indexes[ 2 ] ].minVertex );
-			for( j = 0; j < 3; j++ ) {
-				int vertex = tri->indexes[ j ];
-				if( components[ vertex ].minVertex != minVertex ) {
+		for ( int i = 0; i < cv->numTriangles; i++ ) {
+			srfTriangle_t* tri = &cv->triangles[i];
+			int minVertex = std::min( std::min( components[tri->indexes[0]].minVertex,
```
slipher, line 948: There are a ton of random whitespace changes in this commit (` Move duplicate code from ParseFace() and ParseTriSurf() into ParseTriangleSurface()`). Can you configure your formatter to add the spaces inside square brackets? That seems to be the cause of most of the diffs.

src/engine/renderer/tr_main.cpp
```diff
@@ -1218,19 +1217,9 @@ void R_PlaneForSurface( surfaceType_t *surfType, cplane_t *plane ) {
...
 		case surfaceType_t::SF_TRIANGLES:
 		{
-			srfGeneric_t* tri = ( srfGeneric_t* ) surfType;
-			v1 = tri->verts + tri->triangles[0].indexes[0];
-			v2 = tri->verts + tri->triangles[0].indexes[1];
-			v3 = tri->verts + tri->triangles[0].indexes[2];
-			PlaneFromPoints( plane4, v1->xyz, v2->xyz, v3->xyz );
-			VectorCopy( plane4.normal, plane->normal );
-			plane->dist = plane4.dist;
+			*plane = ( ( srfSurfaceFace_t* ) surfType )->plane;
```
slipher: Use srfGeneric_t cast

slipher (commented):
I think the last commit, `NUKE srfSurfaceFace_t`, should be dropped. It will just cause confusion for one of the surface types to not have its own struct. But a couple changes from that commit of changing some casts to `srfGeneric_t *` should be kept (these fix comments I made on earlier commits complaining of casting to an inappropriate type).

VReaperV:
> I think the last commit, `NUKE srfSurfaceFace_t`, should be dropped. It will just cause confusion for one of the surface types to not have its own struct.

But this is also the case with `srfTriangles_t`.

VReaperV:
I guess I can revert those, though I don't see much point in keeping around structs that have nothing in them other than stuff they inherit.

src/engine/renderer/tr_bsp.cpp
VReaperV: Done.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -3174,21 +3142,12 @@ static void R_CreateWorldVBO()
 
 			oldViewCount = surf1->viewCount;
 
-			if ( *surf1->data == surfaceType_t::SF_FACE )
+			if ( *surf1->data == surfaceType_t::SF_FACE || *surf1->data == surfaceType_t::SF_TRIANGLES
+				|| *surf1->data == surfaceType_t::SF_GRID )
 			{
-				srfSurfaceFace_t *face = ( srfSurfaceFace_t * ) surf1->data;
+				srfGeneric_t* face = ( srfGeneric_t* ) surf1->data;
```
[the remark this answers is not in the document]
VReaperV: Fixed.

src/engine/renderer/tr_local.h
```diff
@@ -1736,97 +1734,57 @@ enum class ssaoMode {
...
-	};
-
-	struct srfSurfaceFace_t : srfGeneric_t
-	{
-		cplane_t     plane;
-
-		// triangle definitions
-		int           numTriangles;
-		srfTriangle_t *triangles;
+		cplane_t plane; // Valid in: SF_FACE, SF_TRIANGLES
```
slipher, line 1754: Does SF_TRIANGLES really have a plane? It doesn't make much sense and the old `srfTriangles_t` doesn't have it.

Also I don't think it's necessary to have comments on ones that are always valid; you can remove those if you want.

VReaperV: I thought that too, but `R_PlaneForSurface()` was already using the first triangle's plane as surface plane for `SF_TRIANGLES`.

slipher: I see. It seems R_PlaneForSurface is only used for portals.

slipher:
LGTM

Issue #1504, opened 2025-01-17
NaCl amd64 crash dumps no longer work on Linux


slipher:
With 0.55.2, the command `daemonded +map chasm +delay 50f sgame.injectFault intdiv` fails to produce a dump in `crashdump/` in the homepath.

It still works on Windows.

illwieckz:
Here is the results of my tests on Linux:

||amd64|i686|armhf
|-|-|-|-
|pnacl|❌️|✅️|✅️
|saigo|❌️|✅️|✅️

So, only linux-amd64 is affected.

slipher:
It appears that the NaCl exception handling method is broken in general. I tried the following patch:

```diff
diff --git a/src/common/System.cpp b/src/common/System.cpp
index 6e598f2ff..843d2681c 100644
--- a/src/common/System.cpp
+++ b/src/common/System.cpp
@@ -335,12 +335,15 @@ static void CrashHandler(const void* data, size_t n)
     Sys::Error("Crashed with NaCl exception");
 }

+void Crash(struct NaClExceptionContext *)
+{
+       Log::Warn("nacl crash!");
+       _exit(99);
+}
+
 void SetupCrashHandler()
 {
-#if !defined(__saigo__)
-    nacl_minidump_register_crash_handler();
-    nacl_minidump_set_callback(CrashHandler);
-#endif
+        nacl_exception_set_handler(Crash);
 }
 #else
 NORETURN static void CrashHandler(int sig)
```

Testing like `./daemonded -set vm.sgame.type 1 +map vega +delay 50f sgame.injectFault intdiv`, the `nacl crash!` message appears on Windows, but not on Linux. All testing is with PNaCl.

P.S. for testing crashes on older Unvanquished versions with no `injectFault`, these [behavior tree snippets](https://github.com/Unvanquished/Unvanquished/pull/3232) are nice.

slipher:
I wondered whether it ever worked at all and had to dig deep for the proof. Finally, I found 3 NaCl crash dumps from when I was developing the original Breakpad PR on an Ubuntu VM snapshot from 2015. And voila, Unvanquished 0.47.0 (the first release to include crash dumps) successfully produces a dump.

Also I tried old Debian in Docker which resulted in old glibc with new kernel. This didn't work so maybe it's the kernel that broke compatibility with something.

- Ubuntu Wily: works. glibc 2.21, kernel 4.2.0, Unvanquished 0.47.0
[...]

PR #1637, opened 2025-04-02
Improve VM crash dumping

slipher:
- Produce NaCl crash dumps when there is a fatal C++ exception. Previously it was printing the exception but failing to trigger a crash dump.
- Fix MSVC native exe terminate handler which could never be called.
- Add a cvar `sgame.nativeExeCrashHandler` (likewise for cgame) which can be used to disable the crash handler in order to produce traditional core dumps for native exe gamelogic.

Note: to test NaCl crash handling on amd64 Linux, you need the fixed `nacl_loader` from https://github.com/DaemonEngine/Daemon/issues/1504#issuecomment-2669666958.


slipher:
I guess I will merge this in honor of merging without review day.

illwieckz:
Yes, we better merge that.

PR #1639, opened 2025-04-05
Fix some bugs from address sanitizer 

slipher:
The out-of-bounds light grid read one led to visibly different lighting on Nano. This map tends to read on the edges of the grid due to its very large cell size: the light grid is only 3x4x2 for the whole map.

### DIFF
--- a/src/engine/server/sv_init.cpp
+++ b/src/engine/server/sv_init.cpp
@@ -108,7 +108,7 @@ static void SendConfigStringToClient( int cs, client_t *cl )
 
 		*out++ = c;
 
-		if ( !*in )
+		if ( !c )
 		{
 			break;
 		}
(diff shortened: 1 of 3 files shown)


src/engine/server/sv_init.cpp
```diff
@@ -108,7 +108,7 @@ static void SendConfigStringToClient( int cs, client_t *cl )
 
 		*out++ = c;
 
-		if ( !*in )
+		if ( !c )
```
VReaperV, line 111: Irrelevant change.

VReaperV (approved):
LGTM

src/engine/server/sv_init.cpp
slipher: `in` was incremented after the last character was read

VReaperV: Ah, I see.

PR #1640, opened 2025-04-06
Add GeometryOptimiser, move some stuff there from tr_bsp

VReaperV:
Requires #1635

This moves some surface merging stuff into GeometryOptimiser.cpp, and adds a function that removes duplicate vertices there. Also removes some more duplicate code, and puts all renderer surfaces into temp memory at the start of `R_CreateWorldVBO()`, to avoid going over more of them (i. e. it excludes portals, autosprite surfaces, `SF_SKIP` etc. at the start, rather than every time when iterating through all of the surfaces.

Adds some functionality that will be needed for further improvements to geometry processing with material system (I've done some tests around that, which already show good results, but I need to make it use surfaces of up to a specific limit of triangles to make it work well in all cases).

### DIFF
--- a/src/engine/renderer/GeometryOptimiser.cpp
+++ b/src/engine/renderer/GeometryOptimiser.cpp
@@ -1,0 +1,282 @@
+﻿/*
+===========================================================================
+Copyright (C) 1999-2005 Id Software, Inc.
+Copyright (C) 2006-2011 Robert Beckebans <trebor_7@users.sourceforge.net>
+Copyright (C) 2009 Peter McNeill <n27@bigpond.net.au>
+
+This file is part of Daemon source code.
+
+Daemon source code is free software; you can redistribute it
+and/or modify it under the terms of the GNU General Public License as
+published by the Free Software Foundation; either version 2 of the License,
+or (at your option) any later version.
+
+Daemon source code is distributed in the hope that it will be
+useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with Daemon source code; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+===========================================================================
+*/
+// GeometryOptimiser.cpp
+
+#include "common/Common.h"
+
+#include "GeometryOptimiser.h"
+
+static int LeafSurfaceCompare( const void* a, const void* b ) {
+	bspSurface_t* aa, * bb;
+
+	aa = *( bspSurface_t** ) a;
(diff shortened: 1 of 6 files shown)


src/engine/renderer/GeometryOptimiser.cpp
```diff
@@ -0,0 +1,259 @@
+﻿/*
+===========================================================================
+
+Daemon BSD Source Code
```
slipher, line 4: This file seems to have old code copy and pasted so it should have the GPL license.

src/engine/renderer/tr_bsp.cpp
```diff
@@ -2755,322 +2519,86 @@ static void R_CreateClusters()
...
+		srfVert_t* materialVerts = ( srfVert_t* ) ri.Hunk_AllocateTempMemory( numVerts * sizeof( srfVert_t ) );
+		glIndex_t* materialIdxs = ( glIndex_t* ) ri.Hunk_AllocateTempMemory( numTriangles * 3 * sizeof( glIndex_t ) );
 
-		for ( ; surfTriangle < surfTriangleEnd; surfTriangle++ )
-		{
-			vboIdxs[ vboNumIndexes++ ] = vboNumVerts + surfTriangle->indexes[ 0 ];
-			vboIdxs[ vboNumIndexes++ ] = vboNumVerts + surfTriangle->indexes[ 1 ];
-			vboIdxs[ vboNumIndexes++ ] = vboNumVerts + surfTriangle->indexes[ 2 ];
-		}
+		// OptimiseMapGeometryMaterial( &s_worldData, numSurfaces, materialVerts, numVerts, materialIdxs, numTriangles * 3 );
```
slipher, line 2591: What's up with that?

illwieckz:
Seeing work for more geometry optimization is great!

Just a thought brought by reading the PR:

For multi-licenses issues, when we have to put new functions into a file with a more permissive license, I think we may add specific comments to split the file without actually doing two files.

I'm thinking about doing those for huge code blocks I entirely rewritten from `tr_shader.cpp` for example, where some of the code is like, entirely brand new for hundreds of lines.

It's harder to do when the new/old code is intricately mixed (not just whole functions or large code blocks living side to side).

The easiest way is to use the old license when mixing new code with the old license to avoid marking each block, but we may start marking blocks at some point.

slipher:
The task of rewriting the entire engine with BSD-licensed code seems a bit hopeless, so personally I just write new code in old files if it would make my life any more difficult to do otherwise. Dealing with multiple licenses in one file sounds painful.

VReaperV:
I'll deal with that when #1635 is merged, I'll need to rebase on it anyway.

VReaperV:
Mac ci is tripping as the variable is incremented in the loop (possibly optmised out, but the next pr ensures it won't be, so, whatever...).

slipher:
Why is this being merged without LGTM? Doesn't seem very urgent.

VReaperV:
I'm simply following what the project head has said.

Issue #1599, opened 2025-03-14
Generated build info missing if BUILD_DUMMY_GAMELOGIC is changed from 0 to 1


slipher:
If in an already existing Daemon build directory I configure `cmake . -DBUILD_DUMMY_GAMELOGIC=1`, the build breaks with this message:

```
CMake Error at cmake/DaemonGame.cmake:117 (add_library):
  Cannot find source file:

    C:/lang/msys64/home/meh/daemonb32/GeneratedSource/DaemonBuildInfo/Game.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .ixx .cppm
  .ccm .cxxm .c++m .h .hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90
  .f95 .f03 .hip .ispc
Call Stack (most recent call first):
  src/dummygame/CMakeLists.txt:3 (GAMEMODULE)
  CMakeLists.txt:1073 (include)
```

It doesn't happen if `-DBUILD_DUMMY_GAMELOGIC=1` is passed the first time CMake is invoked.

I wish generating gamelogic build info would be optional (configured by gamelogic's build) because dummy gamelogic doesn't use that.

slipher:
This was fixed by 3c32c24fcf3730572437412b5bd6c89f5f6a0e08. I still want to move the `<whatever>_generate_buildinfo` call out of DaemonGame.cmake to CMakeLists.txt though.

PR #1641, opened 2025-04-07
cmake: make DaemonPlatform, Architecture, Compiler, etc. autonomous and reusable as Yokai framework

illwieckz:
Some revamp of the `DaemonPlatform`/`DaemonArchitecture`/`DaemonCompiler`/`DaemonBuildInfo` helper code for CMake.

I want this framework to be reusable for other projects, the first user outside of the engine may be the NaCl loader itself, and I may use it for NetRadiant as well at some point. So I made the code fully autonomous and contained in a single folder (so it can be copied and kept it sync easily).

It happens that the endian stuff we did in the `DaemonCompiler` code using `Endian.h` was useless (it was not doing better than what the `qprocessordetection.h` code does anyway), so I removed it.

I also renamed `GAME_PIE` to `NACL_PIE` because it's an NaCl option, it was named `GAME_PIE` because we only build games as NaCl, but for a self-contained framework we better name the NaCl option `NACL`. For now this code sets `GAME_PIE` to the `NACL_PIE` value but once the game is modified to use `NACL_PIE` the code setting the `GAME_PIE` alias will be deleted.

---

_Edit:_ That framework now has its own name: _**Yokai**_.

Yokai is the name of that reusable generic CMake framework. CMake variables are prefixed with `YOKAI_`, C variables and definitions are prefixed with `YOKAI_` as well, and the folder for generated files is prefixed with `Yokai`.

The name is inspired from that [third race mod](https://unvanquished.net/surviving-the-third-race/) that was named Unvanquished, from which we already inherited the Dæmon name (Dæmon was the third team name along Humans and Aliens, the Yokai model was one of those Dæmons).


slipher:
Why would we need a variable called NACL_PIE? There is no PIE for NaCl and never will be.

cmake/DaemonPlatform/Platform.cmake
```diff
@@ -0,0 +1,40 @@
...
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+################################################################################
+# Collection of reusable CMake helpers written for the Dæmon engine and related
+# projects.
+################################################################################
+
```
slipher: Why do we need this file? Are some of the files dependent on each other and so have to be included in a specific order?

illwieckz: You'll end up loading them anyway.

The only thing that seems to be interesting to do is to provide an option to not generate the source files with the various discovered things, for projects not caring about reporting the build environment to the user. For example we don't need such reporting in NaCl.

illwieckz: With the latest version, if you only include `Detection.cmake` without having included `SourceGenerator.cmake` first (or the `Platform.cmake` entry point which is meant to include everything supported), then no source is generated when detecting the build environment.

illwieckz:
There is now a unique `DaemonPlatform/Detection/detect` script that can be used to test the compilation-based detection outside of CMake:

```
$ ./detect clang
DAEMON_COMPILER_GCC_COMPATIBILITY=ON
DAEMON_COMPILER_GCC_VERSION=4.2.1
DAEMON_COMPILER_Clang_COMPATIBILITY=ON
DAEMON_COMPILER_Clang_VERSION=18.1.3
DAEMON_COMPILER_Clang_VERSION_STRING=18.1.3 (1ubuntu1)
DAEMON_COMPILER_generic_VERSION_STRING=Ubuntu Clang 18.1.3 (1ubuntu1)
DAEMON_COMPILER_NAME=Clang
DAEMON_SYSTEM_Linux_COMPATIBILITY=ON
DAEMON_SYSTEM_Unix_COMPATIBILITY=ON
DAEMON_SYSTEM_NAME=Linux
DAEMON_ARCH_NAME=amd64
```

```
$ ./detect i686-w64-mingw32-gcc
DAEMON_COMPILER_GCC_COMPATIBILITY=ON
DAEMON_COMPILER_GCC_VERSION=13.0.0
DAEMON_COMPILER_generic_VERSION_STRING=13-posix
DAEMON_COMPILER_NAME=MinGW
DAEMON_SYSTEM_NAME=Windows
DAEMON_ARCH_NAME=i686
```

```
$ ./detect arm-nacl-clang
DAEMON_COMPILER_GCC_COMPATIBILITY=ON
DAEMON_COMPILER_GCC_VERSION=4.2.1
DAEMON_COMPILER_Clang_COMPATIBILITY=ON
DAEMON_COMPILER_Clang_VERSION=21.0.0
DAEMON_COMPILER_Clang_VERSION_STRING=21.0.0git (https://chromium.googlesource.com/native_client/nacl-llvm-project-v10.git 9c7f0369cfdd591e580c5ccfc1f00fedee58029f)
DAEMON_COMPILER_generic_VERSION_STRING=Clang 21.0.0git (https://chromium.googlesource.com/native_client/nacl-llvm-project-v10.git 9c7f0369cfdd591e580c5ccfc1f00fedee58029f)
DAEMON_COMPILER_NAME=Saigo
DAEMON_SYSTEM_Unix_COMPATIBILITY=ON
DAEMON_SYSTEM_NAME=NaCl
DAEMON_ARCH_NAME=nacl
```

illwieckz:
As a side quest, I'm questioning the prefix. I wanted the variables produced by this framework to have a single prefix to 1. make it obvious it's generated by the framework, 2. have them share a namespace, 3. have them explicitly different with stock `CMAKE_` prefixed variables. Now, the `DAEMON_` prefix may not be the best, as we have genuine need for `DAEMON_` prefixed variables that don't need to be generated by this framework.

The thing is that I want this framework reusable. I plan to use it for the NaCl loader, I want to use it in NetRadiant, I may use it in XQF if the need arise. In those context, another name would be fine, though `DAEMON_` isn't bad because anyway it's the framework from Dæmon, but then it may look better if that's more obvious it's the platform than the engine itself. For example `DPLATFORM_` would also work.

illwieckz:
@slipher I would like to merge this so I can move on. Also I need to rebase the NaCl loader cmake code above it.

In the future I may find a more generic prefix than `DAEMON_` but for now it's good enough. If you think `DPLATFORM_` is fine or you have a better suggestion, I may do it. I still think it would be good if things from that platform would have a prefix different to the engine-specific stuff prefixed with `DAEMON_`. Actually we may go the bold way with a specific name (like Sloth, Urcheon…) for example to reuse [some name historically associated with the Unvanquished project](https://unvanquished.net/surviving-the-third-race/): `YOKAI_`.

illwieckz:
Actually I just did that. This CMake framework was written for the Dæmon engine first but isn't specific to the engine and is meant to be reusable for any CMake-based project, like the NaCl loader, NetRadiant, etc.

So I gave it a name: _**Yokai**_. Yokai is the name of that reusable generic CMake framework. CMake variables are prefixed with `YOKAI_`, C definitions are prefixed with `YOKAI_` as well, and folders for generated files are prefixed with `Yokai`.

The name is inspired from that third race mod that was named Unvanquished, from which we already inherited the Dæmon name (Dæmon was the third team name along Humans and Aliens, the Yokai model was one of those Dæmons).

illwieckz:
One commit renames the framework to Yokai and but keeps a variable substitution mechanism to not break the existing `DAEMON_` prefixed things in C/C++ code and CMake code outside of Yokai, then the next commit removes the substitutions and fully embraces the Yokai namespace in the C/C++ and CMake code all the way down.

slipher:
> A good example of usage of them is to avoid the ambiguity of `WIN32`.
> 
> Instead of `WIN32` to detect the target we can use `DAEMON_TARGET_SYSTEM_Windows` and instead of `MSVC` and `NOT MSVC AND WIN32` to detect the compiler we can use `DAEMON_CXX_COMPILER_MSVC` and `DAEMON_CXX_COMPILER_MinGW`.

Please no. The built-in `WIN32` works fine. It is obnoxious to replace a standard language feature with a custom implementation. Anyone who was worked with CMake in other projects will not be able to use their knowledge. Anyone working on the project, familiar with CMake or not, will be inconvenienced by the use of a homemade version which does not appear in search engines, forcing them to read a bunch of crap to figure out what it is.

cmake/Yokai/SourceGenerator.cmake
```diff
@@ -1,5 +1,5 @@
-# Daemon BSD Source Code
-# Copyright (c) 2025, Daemon Developers
+# YOKAI BSD Source Code
+# Copyright (c) 2025, YOKAI Developers
```
slipher: It's annoying to proliferate different author names because then we would have to update the author lists whenever moving things between files.

srclibs.cmake
```diff
@@ -179,7 +179,7 @@ if (WIN32)
         ${LIB_DIR}/breakpad/src/client/windows/handler/exception_handler.cc
         ${LIB_DIR}/breakpad/src/client/windows/handler/exception_handler.h
     )
-elseif (LINUX)
+elseif (YOKAI_SYSTEM_Linux_COMPATIBILITY)
```
slipher: Does Breakpad Linux code seriously work on some non-Linux OS? Would be surprising since debugger APIs are very OS-specific. But if you want to try it I have a testing script: https://github.com/DaemonEngine/Daemon/pull/1785

CMakeLists.txt
```diff
@@ -534,7 +533,7 @@ else()
 
     # Look for OpenGL here before we potentially switch to looking for static libs.
     if (BUILD_CLIENT)
-        if (LINUX OR FREEBSD)
+        if (YOKAI_SYSTEM_Unix_COMPATIBILITY)
```
slipher: I assume "Unix compatibility" would include Mac? Then this is wrong because this GLVND stuff should not be set on Mac.

cmake/DaemonFlags.cmake
```diff
@@ -604,15 +604,15 @@ endif()
 option(USE_CPU_RECOMMENDED_FEATURES "Use some common hardware features like SSE2, NEON, VFP, MCX16, etc." ON)
 
 # Target options.
-if (MSVC)
-    if (ARCH STREQUAL "i686")
+if (YOKAI_CXX_COMPILER_MSVC)
```
slipher: Replacing MSVC conditionals with this breaks the `clang-cl` compiler which is both Clang and MSVC. It is correctly detected by CMake:
```
-- The C compiler identification is Clang 19.1.5 with MSVC-like command-line
-- The CXX compiler identification is Clang 19.1.5 with MSVC-like command-line
```

clang-cl works on master but this PR breaks it by feeding it GCC-specific options.

Was there anything wrong with CMake's compiler detection that you saw?

slipher:
Besides clang-cl being broken, regular old MSVC also doesn't work. 

``` NaCl VM is not supported on this platform. Set vm.cgame.type and vm.sgame.type to 3 (native DLL) and use devmap instead of map.```

File: cmake/Yokai/SourceGenerator.cmake (hunk 2 above)
Comment (illwieckz) [reply cmake/Yokai/SourceGenerator.cmake]: Oh lol yes this one was just my `sed` command being not specific enough, I'll revert that. 😆️

File: CMakeLists.txt (hunk 4 above)
Comment (illwieckz) [reply CMakeLists.txt]: Right, it doesn't intend to include macOS. I'm looking for a name for Linux-like systems, the ones with XDG desktops, Mesa drivers…

File: cmake/Yokai/SourceGenerator.cmake (hunk 2 above)
Comment (illwieckz) [reply cmake/Yokai/SourceGenerator.cmake]: Fixed.

File: CMakeLists.txt (hunk 4 above)
Comment (slipher) [reply CMakeLists.txt]: What did you mean by having two different variables called Linux_COMPATIBILITY and Unix_COMPATIBILITY then?

Anyway my reaction is that the whole thing is a bad idea. Replacing `WIN32`, `LINUX`, `MSVC` with hand-rolled alternatives just makes the project hostile to contributors.

Comment (illwieckz) [reply CMakeLists.txt]: Maybe I can just name it Linux compatibility, for things that are exactly the same as Linux, even if not Linux.

File: cmake/DaemonFlags.cmake (hunk 5 above)
Comment (illwieckz) [reply cmake/DaemonFlags.cmake]: This is a perfect example of why I wrote this code, to avoid such CMake ambiguity.

When rewriting that, I made the wrong assumption that “MSVC” meant “MSVC compiler” and not “MSVC compiler compatibility”. To avoid this mistake of mine I would have to read CMake docs. But to feel the need to read CMake docs, it was required from to to have in mind the knowledge that there exists non-MSVC compilers compatible with MSVC, information I totally forgot when I rewrote this.

With this framework which is meant to remove ambiguity from CMake, someone reading `COMPILER_MSVC_COMPATIBILITY` doesn't need to read any doc to know what's happening: it's for any compiler that behaves like MSVC.

So the same way I already implemented `COMPILER_GCC_COMPATIBILITY` and `COMPILER_Clang_COMPATIBILITY`, I now implemented `COMPILER_MSVC_COMPATIBILITY` and I used it where ambiguous and misleading `MSVC` was used.

And `COMPILER_MSVC` can be used to target non-Clang MSVC if needed.

Comment (slipher) [reply cmake/DaemonFlags.cmake]: Because no docs at all is so much better than reading CMake docs!

**Comment by illwieckz** [comment]
> Besides clang-cl being broken, regular old MSVC also doesn't work.
> 
> ` NaCl VM is not supported on this platform. Set vm.cgame.type and vm.sgame.type to 3 (native DLL) and use devmap instead of map.`

This wasn't MSVC-specific, it was a mistake that happened when rebasing over the box64 code.
By merging this sooner than later, we would avoid such kind of rebase mistakes to pile up.

File: CMakeLists.txt (hunk 4 above)
Comment (illwieckz) [reply CMakeLists.txt]: Now named `SYSTEM_Linux_COMPATIBILITY`.

**Comment by illwieckz** [comment]
> > A good example of usage of them is to avoid the ambiguity of `WIN32`.
> > Instead of `WIN32` to detect the target we can use `DAEMON_TARGET_SYSTEM_Windows` and instead of `MSVC` and `NOT MSVC AND WIN32` to detect the compiler we can use `DAEMON_CXX_COMPILER_MSVC` and `DAEMON_CXX_COMPILER_MinGW`.
> 
> Please no. The built-in `WIN32` works fine. It is obnoxious to replace a standard language feature with a custom implementation. Anyone who was worked with CMake in other projects will not be able to use their knowledge. Anyone working on the project, familiar with CMake or not, will be inconvenienced by the use of a homemade version which does not appear in search engines, forcing them to read a bunch of crap to figure out what it is.

As said [above](https://github.com/DaemonEngine/Daemon/pull/1641#discussion_r3362946597), CMake is ambiguous and not only requires to read docs, but is ambiguous in a way it even doesn't give the incentive to read docs.

Things like `COMPILER_MSVC_Compatibility` or `SYSTEM_Windows` are very much less ambiguous and don't need to read any doc to be understood. That's the whole purpose.

If we want to remove even more ambiguity we may rename to `TARGET_SYSTEM_Windows`. We currently have `HOST_SYSTEM_Windows` and `SYSTEM_Windows`, but the  `SYSTEM_Windows` keyword being for the target isn't context-free for the reader.

**Comment by illwieckz** [comment]
>  Anyone who was worked with CMake in other projects will not be able to use their knowledge.

Someone writing new code using `WIN32` or `MSVC` will still get his CMake code working, he will be able to use his knowledge, but he will just introduce more error-prone code like we did before.

We would be able to encourage such contributor to use our unambiguous symbols when reviewing, or just rewrite his code later to make it more robust.

Writers don't lose their knowledge, and readers would even require _less_ knowledge.

File: cmake/Yokai/Architecture.cmake [hunk 6]
```diff
@@ -0,0 +1,128 @@
...
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of the <organization> nor the
+#    names of its contributors may be used to endorse or promote products
+#    derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
```
slipher: more license junk

cmake/Yokai/Architecture.cmake
```diff
@@ -0,0 +1,128 @@
...
+	add_definitions(-DYOKAI_ARCH_${arch_name})
+
+	set(nacl_arch "${arch_name}")
+
+	if (YOKAI_SYSTEM_Linux_COMPATIBILITY)
+		set(armhf_usage "arm64;armel")
+		if ("${arch_name}" IN_LIST armhf_usage)
+		# Load 32-bit armhf nexe on 64-bit arm64 engine on Linux with multiarch.
+		# The nexe is system agnostic so there should be no difference with armel.
+		set(nacl_arch "armhf")
```
slipher: I imagine this NaCl and box64 stuff shouldn't be in here as it's Daemon-specific.

src/common/Common.h
```diff
@@ -31,6 +31,14 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef COMMON_COMMON_H_
 #define COMMON_COMMON_H_
 
+// TODO: Remove when the game is ported to Yokai.
+#if !defined(DAEMON_ARCH_STRING)
+#define DAEMON_ARCH_STRING YOKAI_ARCH_STRING
+#endif
+#if !defined(YOKAI_CXX_COMPILER_STRING)
```
slipher: this doesn't make sense

cmake/Yokai/Architecture.cmake
illwieckz: Is the junk being the `<COPYRIGHT HOLDER>` unexpanded template?

This template is used unexpanded in:

- CMakeLists.txt
- cmake/DaemonBuildTypeGeneratorExpression.cmake
- cmake/FindWasmtime.cmake
- cmake/DaemonNacl.cmake
- cmake/toolchain-pnacl.cmake
- cmake/DaemonGame.cmake
- cmake/DaemonFlags.cmake
- cmake/DaemonCBSE.cmake

I can fix it in Yokai's files, but that is something generic to our CMake code.

cmake/DaemonFlags.cmake
slipher: > When rewriting that, I made the wrong assumption that “MSVC” meant “MSVC compiler” and not “MSVC compiler compatibility”. To avoid this mistake of mine I would have to read CMake docs. But to feel the need to read CMake docs, it was required from to to have in mind the knowledge that there exists non-MSVC compilers compatible with MSVC, information I totally forgot when I rewrote this.
> 
> With this framework which is meant to remove ambiguity from CMake, someone reading `COMPILER_MSVC_COMPATIBILITY` doesn't need to read any doc to know what's happening: it's for any compiler that behaves like MSVC.

Seems backward to me. 99% of the time what you want is the MSVC-compatible options/ABI thing. So the person who is blissfully unaware of `clang-cl` will write `if (MSVC)` and get it right, even if by accident. While in the proposed framework they will have multiple options like YOKA_MSVC_COMPATIBILITY or YOKAI_COMPILE_MSVC and they can just as easily get it wrong.

srclibs.cmake
illwieckz: I totally forgot if I tested that on FreeBSD or not. Extending support isn't required for the framework to be merged so I returned to `SYSTEM_Linux`.

slipher:
> If we want to remove even more ambiguity we may rename to `TARGET_SYSTEM_Windows`. We currently have `HOST_SYSTEM_Windows` and `SYSTEM_Windows`, but the `SYSTEM_Windows` keyword being for the target isn't context-free for the reader.

At least that would *actually* be less ambiguous than what we have. The `SYSTEM_` word by itself does not provide any information.

cmake/Yokai/Architecture.cmake
slipher: > Is the junk being the `<COPYRIGHT HOLDER>` unexpanded template?

right

CMakeLists.txt
```diff
@@ -463,23 +462,23 @@ endif()
 
 # Native client
 include(DaemonNacl)
-if (NACL)
+if (YOKAI_SYSTEM_NaCl)
```
slipher: Requesting the system names to be all caps. Trying to spell this mixed case gibberish like SYSTEM_NaCl and SYTEM_macOS seems painful.

src/common/Common.h
illwieckz: Fixed.

cmake/Yokai/Architecture.cmake
illwieckz: I was thinking about only running this NaCl setup code when some variable is set to enable it.

In the same way the Yokai framework can already be used without using the source generation code, it would be usable without caring about NaCl.

illwieckz: This code is likely not needed when called from `DaemonGame.cmake` to begin with.

illwieckz: This NaCl code is now only used when `YOKAI_NACL_HOST` is enabled.

The game code is the first Yokai consumer not using `YOKAI_NACL_HOST`, and the Dæmon engine is the first Yokai consumer using `YOKAI_NACL_HOST`.

slipher: I don't get why you don't just move that stuff somewhere outside of the "Yokai" directory.

illwieckz: Engine:

```
$ cat YokaiGeneratedSource/YokaiBuildInfo/Engine.cpp 
// Automatically generated, do not modify!
const char* YOKAI_SYSTEM_STRING="Linux";
const char* YOKAI_ARCH_STRING="amd64";
const char* YOKAI_NACL_ARCH_STRING="amd64";
const char* YOKAI_C_COMPILER_STRING="GCC_13.3.0:cc";
const char* YOKAI_CXX_COMPILER_STRING="GCC_13.3.0:c++";
```

Dll game:

```
$ cat YokaiGeneratedSource/YokaiBuildInfo/Game.cpp
// Automatically generated, do not modify!
const char* YOKAI_SYSTEM_STRING="Linux";
const char* YOKAI_ARCH_STRING="amd64";
const char* YOKAI_C_COMPILER_STRING="GCC_13.3.0:gcc";
const char* YOKAI_CXX_COMPILER_STRING="GCC_13.3.0:g++";
```

Nexe game:

```
$ cat nacl-vms/YokaiGeneratedSource/YokaiBuildInfo/Game.cpp 
// Automatically generated, do not modify!
const char* YOKAI_SYSTEM_STRING="NaCl";
const char* YOKAI_ARCH_STRING="pnacl";
const char* YOKAI_C_COMPILER_STRING="PNaCl_4.2.1/Clang_3.6.0:pnacl-clang";
const char* YOKAI_CXX_COMPILER_STRING="PNaCl_4.2.1/Clang_3.6.0:pnacl-clang++";
```

CMakeLists.txt
illwieckz: Well, the “NaCl” and “macOS” strings are also stored as `const char* YOKAI_SYSTEM_STRING` which is meant to be printed and then be pretty, but I can add code to upper case them when making a CMake variable or a C macro out of them.

illwieckz: Now that a lot of code has been made reusable and genericized accross System, Architecture and Compiler, it's probably even easier to do so.

illwieckz: Done. Now both CMake variables and C macros are fully capitalized.

cmake/Yokai/Architecture.cmake
illwieckz: I expanded the header template for all our CMake files.

illwieckz:
> > If we want to remove even more ambiguity we may rename to `TARGET_SYSTEM_Windows`. We currently have `HOST_SYSTEM_Windows` and `SYSTEM_Windows`, but the `SYSTEM_Windows` keyword being for the target isn't context-free for the reader.
> 
> At least that would _actually_ be less ambiguous than what we have. The `SYSTEM_` word by itself does not provide any information.

I think the reason why I haven't wrote `TARGET_SYSTEM` but `SYSTEM` is because the symbol name is currently the same in CMake and in C/C++ file, and in C/C++ file it's always a target system…

illwieckz:
The target symbols are now `YOKAI_TARGET_SYSTEM` and `YOKAI_TARGET_ARCH`, they are just `YOKAI_SYSTEM` and `YOKAI_ARCH` in C/C++ (not `TARGET`).

cmake/Yokai/All.cmake
```diff
@@ -1,5 +1,5 @@
 # Daemon BSD Source Code
-# Copyright (c) 2013-2016, Daemon Developers
+# Copyright (c) 2025-2006, Daemon Developers
```
slipher: typo

cmake/Yokai/System.cmake
```diff
@@ -0,0 +1,131 @@
...
+set("YOKAI_HOST_SYSTEM_${YOKAI_HOST_SYSTEM_NAME_UPPER}" ON)
+set("YOKAI_TARGET_SYSTEM_${YOKAI_TARGET_SYSTEM_NAME_UPPER}" ON)
+
+# This is for systems with behaviors similar to Linux,
+# implementing standards like FHS, XDG, GLVND…
+# Makes possible to do that in CMake code:
+# > if (YOKAI_HOST_SYSTEM_LINUX_COMPATIBILITY)
+# > if (YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY)
+foreach(name LINUX;FREEBSD)
+	foreach(slug HOST;HOST_SYSTEM)
```
slipher: I think you wanted TARGET to be one of the things in this line?

Just write it without loops, it would be shortesr that way anyway.

cmake/DaemonFlags.cmake
```diff
@@ -263,7 +264,7 @@ if (NOT NACL AND BUILD_CLIENT)
 	option(USE_OPENMP "Use OpenMP to parallelize some tasks" OFF)
 endif()
 
-if (MSVC)
+if (YOKAI_CXX_COMPILER_MSVC_COMPATIBILITY)
```
slipher, line 267: `/MP` doesn't do anything and prints a warning in the Clang clone so this would be a good opportunity to use a really real MSVC variable.

srclibs.cmake
```diff
@@ -179,7 +179,9 @@ if (WIN32)
         ${LIB_DIR}/breakpad/src/client/windows/handler/exception_handler.cc
         ${LIB_DIR}/breakpad/src/client/windows/handler/exception_handler.h
     )
-elseif (LINUX)
+# We don't know if this works outside of Linux,
+# so we don't use YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY.
```
slipher: It definitely won't work, see https://forums.freebsd.org/threads/porting-google-breakpad.18039/

CMakeLists.txt
```diff
@@ -733,7 +733,7 @@ if (USE_BREAKPAD)
         set(LIBS_ENGINE_BASE ${LIBS_ENGINE_BASE} srclibs-breakpad-exception_handler
             srclibs-breakpad-crash_generation_client srclibs-breakpad-common)
 
-    elseif (LINUX)
+    elseif (YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY)
```
slipher, line 736: should be exactly Linux

slipher:
I should mention that besides the Clang that pretends to be MSVC, there is also [Clang that pretends to be MinGW](https://wiki.unvanquished.net/wiki/Compiling_the_source#MSYS2_Clang). I don't have it installed currently so I can't say how the compiler is detected.

slipher:
I still don't understand the claimed problems with making each cmake file self-contained (by including dependencies using the `include_guard` thing), or doing NaCl stuff outside of Yokai. `daemon_generate_buildinfo` is called in Daemon's CMakeLists.txt, and it's still that way after the refactoring. That's a good thing: clients can decide whether they want a build info to be generated, and add their own definitions to it. So making the buildinfo cmake file a dependency of any other file is no problem because it doesn't contribute anything to the build unless `generate_buildinfo` is called. And Daemon can add its own build infos respecting NaCl after including the Yokai scripts.

(Also I think it's bad how build info is done with gamelogic currently, though that can go in another PR. `DaemonGame.cmake` currently generates the build info when it's included and then it's hackily included from `q_shared.h` with ifdefs. Instead of that, the gamelogic should call `daemon_generate_buildinfo` itself (if it even wants to) and include it from its own files: that way it would have the chance to add its own definitions.)

srclibs.cmake
illwieckz: Comment removed.

CMakeLists.txt
illwieckz: Done.

cmake/DaemonFlags.cmake
illwieckz: Now only doing `/MP` when MSVC is true MSVC (using `YOKAI_CXX_COMPILER_MSVC` which is only set when `CMAKE_CXX_COMPILER_ID STREQUAL "MSVC"`).

cmake/Yokai/All.cmake
illwieckz: Fixed.

cmake/Yokai/System.cmake
illwieckz: Now doing `HOST_SYSTEM;TARGET_SYSTEM`. Indeed `HOST` was never correct.

illwieckz: In fact, we should not name that `YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY`, `YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY` was set twice, first as a marker of `__linux__` defined in preprocessor (matching both Linux and Android), and this other one meant to match Linux-like desktops (operating system environment). So this one is now named `YOKAI_TARGET_SYSTEM_XDG_COMPATIBILITY`.

So now there is:

- `YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY` is for when there is Linux API (`__linux__`), being a Linux desktop or not.
- `YOKAI_TARGET_SYSTEM_XDG_COMPATIBILITY` is for systems behaving like a Linux desktop (FHS, XDG desktop, GLVND…)

I named it XDG because I wanted some standard, not an operating system. XDG (X desktop group) is about desktops, but a system with an XDG desktop (GNOME, KDE…) very likely also do FHS, GLVND, uses `.a` for static library, things like that. So I thought that “XDG compatibility” is very close to what it is meant.

srclibs.cmake
illwieckz: In fact it was indeed meant to be `LINUX_COMPATIBILITY`, so it can be used on Android even if not a Linux Desktop.

It's just that I defined twice `LINUX_COMPATIBILITY`, and the second one with another meaning:

- https://github.com/DaemonEngine/Daemon/pull/1641#discussion_r3377513626

Now I use `XDG_COMPATIBILITY` when the desktop is linux-like, while `LINUX_COMPATIBILITY` means the Linux API is actually there.

And for completeness, `LINUX` means Linux desktop, not Android (that one of the ambiguity I also wanted to remove).

CMakeLists.txt
illwieckz: Same as said in other comments, there was a name collision.

In fact I was targeting the Linux API (`__linux__`) there, but the `(FREEBSD OR LINUX)` variable was mistakenly named `LINUX_COMPATIBILITY` too, bringing confusion.

illwieckz:
The NaCl architecture stuff is now removed from Yokai and stored in `cmake/DaemonNaclHost.cmake`, it is only called by the engine CMake code. Note that I will probably deeply rewrite this code later, but that's outside of the scope if this PR.

illwieckz:
> (Also I think it's bad how build info is done with gamelogic currently, though that can go in another PR. `DaemonGame.cmake` currently generates the build info when it's included and then it's hackily included from `q_shared.h` with ifdefs. Instead of that, the gamelogic should call `daemon_generate_buildinfo` itself (if it even wants to) and include it from its own files: that way it would have the chance to add its own definitions.)

Right now Dæmon's `CMakeLists.txt` does:

```cmake
yokai_write_buildinfo("DaemonEngine")
```

And game's `DaemonGame.cmake` does:

```cmake
yokai_write_buildinfo("DaemonGame")
```

The buildinfo is never mixed between engine and game.

illwieckz:
> The buildinfo is never mixed between engine and game.

Also the list of things in engine and game buildinfo aren't the same anymore, for example the game buildinfo doesn't have the `DAEMON_NACL_ARCH_STRING` constant that is engine-only.

Also it's not possible for the game to use engine's buildinfo because the nexe build is done in a subfolder.

The name passed to `yokai_write_buildinfo` is the name of the subfoler the buildinfo data is written to, to make sure the game doesn't reuse engine buildinfo by mistake when built as dll/exe (and sharing the build folder with the engine).

slipher:
> > (Also I think it's bad how build info is done with gamelogic currently, though that can go in another PR. `DaemonGame.cmake` currently generates the build info when it's included and then it's hackily included from `q_shared.h` with ifdefs. Instead of that, the gamelogic should call `daemon_generate_buildinfo` itself (if it even wants to) and include it from its own files: that way it would have the chance to add its own definitions.)
> 
> Right now Dæmon's `CMakeLists.txt` does:
> 
> ```cmake
> yokai_write_buildinfo("DaemonEngine")
> ```
> 
> And game's `DaemonGame.cmake` does:
> 
> ```cmake
> yokai_write_buildinfo("DaemonGame")
> ```
> 
> The buildinfo is never mixed between engine and game.

I'm saying the write_buildinfo should be in Unvanquished CMakeLists.txt, not in DaemonGame.cmake. I'm hoping this could fix https://github.com/DaemonEngine/Daemon/issues/1599.

CMakeLists.txt
slipher: Why does it still say COMPATIBILITY? What would be the difference between that and YOKAI_TARGET_SYSTEM_LINUX now?

illwieckz: `YOKAI_TARGET_SYSTEM_LINUX` reports Linux desktop and is meant to be consistent with `YOKAI_SYSTEM_STRING` reporting `"Linux"` in C++ code.

While `YOKAI_TARGET_SYSTEM_ANDROID` reports Android and is meant to be consistent with `YOKAI_SYSTEM_STRING` reporting `"Android"` in C++ code.

The `YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY` is the equivalent of `__linux__`, matching both Linux and Android.

When Android is matched, Android is reported, not Linux, but Linux compatibility is reported.

This is similar to the compiler thing where both Clang, ArmClang, AppleClang, AOCC and ICX reports Clang compatibility, but we can still match upstream Clang among its forks, or one of those forks.

It makes even more sense for Linux because between things like Linux and Android the userland is way more different than between two Clang downstreams.

cmake/Yokai/Detection/System.c
```diff
@@ -0,0 +1,72 @@
...
+#if defined(__APPLE__) && defined(__MACH__)
+#include <TargetConditionals.h>
+#endif
+
+#if defined(__linux__) || defined(__linux) || defined(linux) || defined(__gnu_linux__)
+	#pragma message(REPORT_COMPATIBILITY("Linux"))
+#endif
+
+#if defined(__unix__) || defined(__unix) || defined(unix)
+	#pragma message(REPORT_COMPATIBILITY("Unix"))
```
slipher: I suggest removing this as it has unexpected semantics: Mac OS is a Unix, but does not define these. Anyhow the good old idiom for Unix-like is `NOT WIN32` :)

cmake/Yokai/System.cmake
```diff
@@ -0,0 +1,131 @@
...
+# > if (YOKAI_HOST_SYSTEM_LINUX)
+# > if (YOKAI_TARGET_SYSTEM_WINDOWS)
+set("YOKAI_HOST_SYSTEM_${YOKAI_HOST_SYSTEM_NAME_UPPER}" ON)
+set("YOKAI_TARGET_SYSTEM_${YOKAI_TARGET_SYSTEM_NAME_UPPER}" ON)
+
+# This is for systems behaving similarly to a Linux Desktop,
+# implementing standards like FHS, XDG, GLVND…
+# Makes possible to do that in CMake code:
+# > if (YOKAI_HOST_SYSTEM_XDG_COMPATIBILITY)
+# > if (YOKAI_SYSTEM_XDG_COMPATIBILITY)
```
slipher: Instead of XDG_COMPATIBILITY I prefer the GNU_COMPATIBILITY name you used in some other places. This makes sense by reflecting the "GNU/Linux" userland/kernel distinction. So we have Linux compatibility include Android's Linux-based kernel, and GNU compatibility including the FreeBSD using a similar userland to the Linux desktops.

slipher: Well it's not that great since GNU compatibility could also be confused for the "GNU compiler id" that CMake identifier for GCC. But at least it relates to the "GNU/Linux" meme. For me XDG doesn't suggest the concept you want to get at at all.

cmake/DaemonNaclHost.cmake
```diff
@@ -0,0 +1,78 @@
...
+	if (YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY OR YOKAI_TARGET_SYSTEM_XDG_COMPATIBILITY)
+		set(armhf_usage "arm64;armel")
+		if ("${arch_name}" IN_LIST armhf_usage)
+		# Load 32-bit armhf nexe on 64-bit arm64 engine on Linux with multiarch.
+		# The nexe is system agnostic so there should be no difference with armel.
+		set(nacl_arch "armhf")
+
+		set(box64_usage ppc64el riscv64)
+		if ("${arch_name}" IN_LIST box64_usage)
+			option(YOKAI_NACL_BOX64_EMULATION "Use Box64 to emulate x86_64 NaCl loader on unsupported platforms" ON)
```
slipher: can go back to its original name

illwieckz:
> I should mention that besides the Clang that pretends to be MSVC, there is also [Clang that pretends to be MinGW](https://wiki.unvanquished.net/wiki/Compiling_the_source#MSYS2_Clang). I don't have it installed currently so I can't say how the compiler is detected.

Right now this does:

```
#elif defined(__MINGW64__) || defined(__MINGW32__)
	#pragma message(REPORT_NAME("MinGW"))
#elif defined(__clang__)
	#pragma message(REPORT_NAME("Clang"))
```

And this doesn't do any:

```
#pragma message(REPORT_COMPATIBILITY("MinGW"))
```

So for now I expect it to just be reported as `COMPILER_MINGW` with `COMPILER_CLANG_COMPATIBILITY`.

If needed we can make it be reported as `COMPILER_CLANG` with `COMPILER_MINGW_COMPATIBILITY` instead, or something like that.

One can use the `Yokai/Detection/detect` script for a quick preview of what CMake is expected to provide:

```
$ ./cmake/Yokai/Detection/detect clang
YOKAI_C_COMPILER_GCC_COMPATIBILITY=ON
YOKAI_C_COMPILER_GCC_VERSION=4.2.1
YOKAI_C_COMPILER_CLANG_COMPATIBILITY=ON
YOKAI_C_COMPILER_CLANG_VERSION=18.1.3
YOKAI_C_COMPILER_CLANG_VERSION_STRING=18.1.3 (1)
YOKAI_C_COMPILER_GENERIC_VERSION_STRING=Ubuntu Clang 18.1.3 (1)
YOKAI_C_COMPILER_NAME=Clang
YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY=ON
YOKAI_TARGET_SYSTEM_UNIX_COMPATIBILITY=ON
YOKAI_TARGET_SYSTEM_NAME=Linux
YOKAI_TARGET_ARCH_NAME=amd64

$ ./cmake/Yokai/Detection/detect aarch64-linux-android29-clang
YOKAI_C_COMPILER_GCC_COMPATIBILITY=ON
YOKAI_C_COMPILER_GCC_VERSION=4.2.1
YOKAI_C_COMPILER_CLANG_COMPATIBILITY=ON
YOKAI_C_COMPILER_CLANG_VERSION=9.0.8
YOKAI_C_COMPILER_CLANG_VERSION_STRING=9.0.8 (https://android.googlesource.com/toolchain/llvm-project 207d7abc1a2abf3ef8d4301736d6a7ebc224a290)
YOKAI_C_COMPILER_GENERIC_VERSION_STRING=4.2.1 Compatible Android (5900059 based on r365631c) Clang 9.0.8 (https://android.googlesource.com/toolchain/llvm-project 207d7abc1a2abf3ef8d4301736d6a7ebc224a290)
YOKAI_C_COMPILER_NAME=Clang
YOKAI_TARGET_SYSTEM_LINUX_COMPATIBILITY=ON
YOKAI_TARGET_SYSTEM_UNIX_COMPATIBILITY=ON
YOKAI_TARGET_SYSTEM_NAME=Android
YOKAI_TARGET_ARCH_NAME=arm64

$ ./cmake/Yokai/Detection/detect i686-w64-mingw32-gcc
YOKAI_C_COMPILER_GCC_COMPATIBILITY=ON
YOKAI_C_COMPILER_GCC_VERSION=13.0.0
YOKAI_C_COMPILER_GENERIC_VERSION_STRING=13-posix
YOKAI_C_COMPILER_NAME=MinGW
YOKAI_TARGET_SYSTEM_NAME=Windows
YOKAI_TARGET_ARCH_NAME=i686
```

cmake/DaemonNaclHost.cmake
illwieckz: Especially since I renamed it in C++ code already. 😆️

Now fixed.

cmake/Yokai/System.cmake
illwieckz: In some iterations I wrote GNU too, then switched to XDG. The thing is that GNU means GNU Hurd in System context.

One may say that GNU Linux is probably similar to GNU Hurd, not less then FreeBSD, but well…

The `SYSTEM_GNU_COMPATIBILITY` in some other places are leftovers remaining from that first iteration I failed to clean-up properly. I'll fix it.

illwieckz: Fixed.

cmake/Yokai/Detection.cmake
```diff
@@ -0,0 +1,110 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+################################################################################
+# System, architecture and compiler detection.
+################################################################################
+
+# Make sure to always call this macro from within a function, not in the global scope.
+# As a macro it produces a lot of variables in the parent scope but it is meant to
+# only be called by functions so they should never pollute the globale scope.
```
slipher: "globale"

cmake/Yokai/Compiler.cmake
```diff
@@ -25,66 +25,21 @@
...
-			set(DAEMON_${lang}_COMPILER_${name}_VERSION
-				"${compiler_${name}_version}"
-				PARENT_SCOPE)
-
-			set(DAEMON_${lang}_COMPILER_${name}_COMPATIBILITY
-				"${compiler_${name}_compatibility}"
-				PARENT_SCOPE)
-		endforeach()
-	endif()
+	yokai_run_detection("${lang}" "COMPILER" "Compiler${${lang}_EXT}" "GCC;Clang;generic")
```
slipher, line 42: "generic" string seems to be unused

cmake/DaemonFlags.cmake
```diff
@@ -382,7 +386,7 @@ else()
 		endif()
 	endif()
 
-	if (NACL AND USE_NACL_SAIGO AND SAIGO_ARCH STREQUAL "arm")
+	if (YOKAI_TARGET_SYSTEM_NACL AND USE_NACL_SAIGO AND YOKAI_NACL_ARCH_armhf)
```
slipher: missed capitalization spot?

cmake/DaemonFlags.cmake
```diff
@@ -625,25 +629,25 @@ elseif (NOT NACL)
...
 	# not requiring the NX bit.
 
-	if (ARCH STREQUAL "amd64")
+	if (YOKAI_TARGET_ARCH_AMD64)
 		# K8 or EM64T minimum: AMD Athlon 64 ClawHammer, Intel Xeon Nocona, Intel Pentium 4 model F (Prescott revision EO), VIA Nano.
-		if (DAEMON_CXX_COMPILER_ICC)
+		if (YOKAI_CXX_COMPILER_ICC)
 			set(GCC_GENERIC_ARCH "pentium4")
-		elseif (DAEMON_CXX_COMPILER_Zig)
+		elseif (YOKAI_CXX_COMPILER_Zig)
```
slipher: missed capitalization spot?

cmake/Yokai/Detection.cmake
illwieckz: Fixed.

cmake/Yokai/Compiler.cmake
illwieckz: It exports compiler's `__VERSION__` macro to `YOKAI_C_COMPILER_GENERIC_VERSION_STRING`.

This is used (named as temporary variable `compiler_generic_version_string` in `yokai_detect_compiler` function) to extract the PNaCl version string.

cmake/DaemonFlags.cmake

cmake/DaemonFlags.cmake

slipher:
Will these commits be squashed?

There's one outstanding comment about the UNIX_COMPATIBILITY variable having bad semantics.

illwieckz:
> Will these commits be squashed?

They are fine as they are right now. Every time a mistake has been found, I applied the fix to the earliest commit, fixing the rebase all the way up, in order for every commit to be correct. The only commit I didn't do that yet was the `cmake: fix a saigo armhf typo` and I now done it. Each commits are autonomous and evolving the stuff step by step.

> There's one outstanding comment about the UNIX_COMPATIBILITY variable having bad semantics.

The remaining generated `UNIX_COMPATIBILITY` variable just means `#if defined(__unix__) || defined(__unix) || defined(unix)`.

slipher:
> The remaining generated `UNIX_COMPATIBILITY` variable just means `#if defined(__unix__) || defined(__unix) || defined(unix)`.

And that's bad because Mac is a Unix platform but doesn't define those.

slipher:
This comment: https://github.com/DaemonEngine/Daemon/pull/1641#discussion_r3381095138

illwieckz:
> > The remaining generated `UNIX_COMPATIBILITY` variable just means `#if defined(__unix__) || defined(__unix) || defined(unix)`.
> 
> And that's bad because Mac is a Unix platform but doesn't define those.

Are you sure? I don't have an easily accessed mac next to me right now, but if you have one it's easy to check by doing:

```
echo '' | cc -dM -E -
```

If macOS doesn't define those then it explicitly refuses to be recognized as an Unix platform and we should respect that…

It borrows things from earlier MacOS, some BSD Unix stuff, GNU, NextSTEP and other things. Pre-installing a 20 years old bash (even not used as shell by default) doesn't make something an Unix. =)

slipher:
> Are you sure? I don't have an easily accessed mac next to me right now, but if you have one it's easy to check by doing:
> 
> ```
> echo '' | cc -dM -E -
> ```

I confirm that none of the Unix defines appear from this command.

> If macOS doesn't define those then it explicitly refuses to be recognized as an Unix platform and we should respect that…

https://www.opengroup.org/openbrand/certificates/1205p.pdf

Mac OS is universally recognized as a Unix-like system. You have said that the names are intended to be understood without reading anything, so it behooves us not to make one that directly contradicts its usual understanding.

illwieckz:
Well, then we may drop this `__unix__` macro detection entirely. We don't use it for Dæmon anyway, the NaCl loader doesn't use it as well. Some of our dependencies use it, like LZMA in Crunch, and some tests in Recast (but then that doesn't contribute to the binary). This `__unix__` macro seems to be useless and unusable.

illwieckz:
Now done.

I also fixed something: we're not supposed to use `NACL_ARCH_XXX` when `NOT TARGET_SYSTEM_NACL`, but `TARGET_ARCH_XXX` instead.

illwieckz:
I wanted to test with Saigo and I just did and it worked.

slipher:
I think the Breakpad Linux build is broken.

slipher:
Never mind, it's the combination of Breakpad and dummy app, which is already broken on master.

slipher:
LGTM

illwieckz:
@slipher about the `include_guard` thing, I haven't investigated it because it wasn't needed, but if someone thinks it can be used to improve things, that one can investigate this.

PR #1643, opened 2025-04-11
Fix compile error with latest Appleclang

slipher:
It complained about char_traits<unsigned char> being requested somehow. Anyway allowing arbitrarily large sizes is bad in network code, so use a fixed-size buffer instead of the basic_string<byte>.


illwieckz (approved):
LGTM.

Issue #1600, opened 2025-03-14
Deform shader "not found" fatal error


slipher:
I was running renderer tests to validate the latest changes on master and got a fatal error with the message `Shader not found: deformVertexes_vega_2 0`. I am unable to reproduce despite having used a script that prepares an empty homepath and always issues the same commands at the same frames. Some interesting cvars are `r_materialSystem 1` and `r_lazyShaders 2`. The build is based on master 4cfabd48 with dummy gamelogic.

```
Unvanquished 0.55.2 Windows i686 (MinGW_14.2.0/GCC_14.2.0:c++.exe) Mar 13 2025
cmdline: -homepath C:/xxxxxxxx/dg -set scenefile maponly -set r_materialSystem 1
[FS] Lib path: C:/xxxxxxxx/dg
[FS] Home path: C:/xxxxxxxx/mastM
[FS] Pak search path: C:/xxxxxxxx/pkg
[FS] Pak search path: C:/xxxxxxxx/pkg
[FS] Loading pak 'C:/xxxxxxxx/unvanquished_0.55.0.dpk'...
[FS] Loading pak 'C:/xxxxxxxx/tex-common_2.4.dpk'...
[FS] Loading pak 'C:/xxxxxxxx/res-players_0.55.dpk'...
[FS] Loading pak 'C:/xxxxxxxx/res-legacy_0.55.dpk'...
[FS] Loading pak 'C:/xxxxxxxx/res-weapons_0.55.dpk'...
[FS] Loading pak 'C:/xxxxxxxx/res-buildables_0.55.dpk'...
[FS] Loading pak 'C:/xxxxxxxx/res-voices_0.54.dpk'...
[FS] Loading pak 'C:/xxxxxxxx/res-soundtrack_0.54.dpk'...
execing 'default.cfg'
execing 'presets/input/wasd.cfg'
execing 'autogen.cfg'
----- Client Initialization -----
Loading RSA keys from pubkey
^2Daemon RSA public-key file not found, regenerating

^5Regenerating RSA keypair; writing to pubkey
Daemon RSA keys generated
----- Client Initialization Complete -----
Winsock Initialized
Opening IP6 socket: [::]:*
Opening IP socket: 0.0.0.0:*
--- Common Initialization Complete ---
Calling GetRefAPI…
SDL_Init( SDL_INIT_VIDEO )... 
Using SDL version 2.26.5
SDL using driver "windows"
Initializing OpenGL display
[...]

PR #1645, opened 2025-04-13
Rework material system material and surface generation

VReaperV:
Some more changes to allow entities to be added to the material system. This entirely removes the `GenerateWorldMaterials()` function, instead materials are generated for each `MaterialSurface`. This also removes a bunch of things from `drawSurf_t`, `tess`, `Render_*()` etc. that were only used to pass information through to the material system.

Also fixed a memory leak in material system portal surface processing.

### DIFF
--- a/src/engine/renderer/GeometryOptimiser.cpp
+++ b/src/engine/renderer/GeometryOptimiser.cpp
@@ -315,13 +315,13 @@ void MergeDuplicateVertices( bspSurface_t** rendererSurfaces, int numSurfaces, s
 				srfVert_t& vert = face->verts[triangle->indexes[j]];
 				uint32_t index = verts[vert];
 
-				ASSERT_LT( idx, numIndicesIn );
+				ASSERT_LT( idx, ( uint32_t ) numIndicesIn );
 				if ( !index ) {
 					verts[vert] = vertIdx + 1;
 					vertices[vertIdx] = vert;
 					indices[idx] = vertIdx;
 
-					ASSERT_LT( vertIdx, numVerticesIn );
+					ASSERT_LT( vertIdx, ( uint32_t ) numVerticesIn );
 
 					vertIdx++;
 				} else {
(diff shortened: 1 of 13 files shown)


slipher:
Ran tests and got this: `^3Warn: Unhandled exception (15ShaderException): Shader not found: deformVertexes_pulse_2 0`. Note that I am NOT using the dummygame.

VReaperV:
Is it possible it's a glsl cache issue? I've just loaded `pulse` with engine built from this pr and it works fine.

slipher:
I can reproduce that with just `./daemon -set r_materialSystem 1 +devmap pulse`.

slipher:
> Is it possible it's a glsl cache issue? I've just loaded `pulse` with engine built from this pr and it works fine.

No, I still reproduce it with a pristine homepath.

VReaperV:
I did the same thing with homepath, still can't reproduce. I've also added some logging locally, and `deformVertexes_pulse_2 0` is loaded after materials are generated.

illwieckz:
> I can reproduce that with just `./daemon -set r_materialSystem 1 +devmap pulse`.

I get the exact same (AMD RDNA3, Mesa radeonsi).

VReaperV:
> > I can reproduce that with just `./daemon -set r_materialSystem 1 +devmap pulse`.
> 
> I get the exact same (AMD RDNA3, Mesa radeonsi).

So it works on master, but not on this branch? The only thing I could think of is that there's some end-of-string character or something that is messing it up (between `tr.loadingMap` and `tr.world->baseName`).

illwieckz:
> So it works on master, but not on this branch?

Yes, on my end it works on master but not on this branch.

VReaperV:
I was able to reproduce this on `vega` instead. The issue itself is present on master, but apparently it gets hidden by a race condition: the `gfx/buildables/human_base/spawning` also uses deform shaders, and it can be loaded before the map shaders since it comes from a QVM call, thereby hiding a bug in deform shader lookup.

VReaperV:
Added a commit to fix this. This should also fix the similar issue in #1600.

VReaperV:
That's `chasm` and `watah`?

VReaperV:
Can't reproduce the `watah` one running the regular gamelogic, I assume the above screenshots are taken with dummy gamelogic instead.

slipher:
No. I don't really use dummy gamelogic anymore since the GLSL building changes broke it.

illwieckz:
I experience no regression between `master` and this anymore.

illwieckz (approved):
LGTM.

PR #1648, opened 2025-04-14
Revert changes that break material system renderer

slipher:
With material system enabled there is massive breakage. When I load a few maps and fly around there are missing/flickering surfaces everywhere. (chasm, chasm, zandronum)

unvanquished_2025-04-14_122402_000

unvanquished_2025-04-14_122537_000

unvanquished_2025-04-14_122521_000


slipher:
No changes were made to #1645 since the broken screenshots demonstrated in https://github.com/DaemonEngine/Daemon/pull/1645#issuecomment-2801277406.

slipher:
It seems like the breakage starts the 2nd time the renderer is loaded, and deterministically depends on the sequence of maps you have loaded before. For example when loading Chasm and `vid_restart`ing I get this one: 
unvanquished_2025-04-14_124019_000

PR #1649, opened 2025-04-14
Revert "Revert changes that break material system renderer (#1648)" + fix

VReaperV:
This seems to fix the issue, at least I can no longer reproduce it by switching around maps/doing `vid_restart`.


slipher:
This makes the Pulse big water room on the material system appear more similar to the core renderer. I wonder if there is a bug fix or just some sort of surface ordering randomness.

LGTM

Core renderer:

unvanquished-pulse-water

Material system on master:

unvanquished-pulse-water

Material system with this PR:

unvanquished-pulse-water

slipher (approved):

VReaperV:
> I wonder if there is a bug fix or just some sort of surface ordering randomness.

Yeah, it could be due to ordering + the fact that transparent surfaces are currently not sorted in material system (well, they aren't sorted by distance in the core renderer either, but they might just end up looking more correct with the shader/lightmap/etc sorting it does).

PR #1652, opened 2025-04-18
Fix geometry optimiser

### LINKED ISSUE #1651: Issues with MergeDuplicateVertices hashing/equality
These issues concern the commit 7160194c "Merge duplicate map vertices" from #1640.

- `MapVertHasher` does some out-of-bounds and unaligned reads. For example `xyz` is a 12-byte `float[3]` so on 64-bit platforms, `*( ( size_t* ) &vert.xyz[2] )` reads bytes 8-16, past the end of it. (Also reading a float through a `size_t` pointer is undefined behavior, though we use `-fno-strict-aliasing` to instruct the compiler to allow it.) We have `Util::bit_cast` for stuff like converting a float to its `uint32_t` bit pattern.
- Since the hash function apparently is supposed to require `st` to be exactly equal for two verts to match, the equality function `MapVertEqual` probably should too. The hash function requires exact equality for xyz and st while the equality function requires exact equality for xyz but approximate for st.
- ArrayBytes() returns the object size so `lhs.lightColor.ArrayBytes() == rhs.lightColor.ArrayBytes()` is always true.

VReaperV:
Fixes #1651

I opted for making the hash not use uv's instead, as it didn't seem to have much of any effect on the loading time, but resulted in merging a few more vertices. 

Also fixed sky brushes being added as surfaces to the material system.

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -638,6 +638,10 @@ void MaterialSystem::GenerateWorldCommandBuffer( std::vector<MaterialSurface>& s
 	}
 
 	for ( MaterialSurface& surface : surfaces ) {
+		if ( surface.skyBrush ) {
+			continue;
+		}
+
 		SurfaceDescriptor surfaceDescriptor;
 		VectorCopy( surface.origin, surfaceDescriptor.boundingSphere.origin );
 		surfaceDescriptor.boundingSphere.radius = surface.radius;
@@ -1328,6 +1332,10 @@ void MaterialSystem::GenerateMaterial( MaterialSurface* surface ) {
 
 	uint32_t stage = 0;
 	uint32_t previousMaterialID = 0;
+
+	if ( surface->skyBrush ) {
+		return;
+	}
 	
 	if ( surface->shader->depthShader ) {
 		uint32_t unused;
--- a/src/engine/renderer/GeometryOptimiser.cpp
+++ b/src/engine/renderer/GeometryOptimiser.cpp
@@ -377,7 +377,10 @@ std::vector<MaterialSurface> OptimiseMapGeometryMaterial(bspSurface_t** renderer
 		MaterialSurface srf {};
 
 		srf.shader = surface->shader;
+
 		srf.bspSurface = true;
+		srf.skyBrush = surface->skyBrush;
+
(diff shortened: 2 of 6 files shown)


slipher:
The commit `Don't hash the uv's in geometry optimiser` does the opposite of its description. It adds them back to the hash, while the previous commit removes them.

VReaperV:
> The commit `Don't hash the uv's in geometry optimiser` does the opposite of its description. It adds them back to the hash, while the previous commit removes them.

Fixed.

slipher:
The station15 sun disappears when material system is enabled.

unvanquished-station15-autosprite1-sun

(There should be a sun)

slipher:
illwieckz has said the station15 sun uses something called a "skybox portal". (You can see 2 copies of it if noclipping.) I have no idea where this implemented in the code. There is a flag called RF_SKYBOXPORTAL but it's never set anywhere, so apparently unrelated.

VReaperV:
> The station15 sun disappears when material system is enabled.
> 
> unvanquished-station15-autosprite1-sun
> 
> (There should be a sun)

Fixed.

src/engine/renderer/Material.cpp
```diff
@@ -1847,10 +1855,10 @@ void MaterialSystem::AddPortalSurfaces() {
 void MaterialSystem::AddAutospriteSurfaces() {
 	tr.currentEntity = &tr.worldEntity;
 
-	for ( const bspSurface_t* surface : autospriteSurfaces )
+	for ( const MaterialSurface& surface : autospriteSurfaces )
 	{
-		R_AddDrawSurf( surface->data, surface->shader,
-		               surface->lightmapNum, surface->fogIndex, true );
+		R_AddDrawSurf( surface.surface, surface.shader,
```
slipher, line 1860: Why does the type of `autospriteSurfaces` need to be changed? Seems to be using the same information that was already there before.

VReaperV: Ah, yea, I did that in the process of trying to find the issue with station15 sun. I guess it can be changed back to `bspSurface_t*`.

VReaperV: Back to `bspSurface_t*`.

slipher:
LGTM

Issue #1294, opened 2024-09-11
Alternate lazy shader loading implementation


illwieckz:
Right now we have 3 kinds of GLSL shader loading implementation (`r_lazyShaders`):

- 0: build all shaders at game start before reaching the main menu (old default):
  it builds everything before displaying something useful and the user even don't know what's happening,
  the main menu may not be reachable if an advanced shader doesn't work even if unused by the main menu.
- 1: build shaders on demand until map loads, then build everything else at map load (current default)
  make sure only the shaders required to display the main menu are built before displaying the main menu,
  makes fast game load, and fast preset selection in main menu,
  makes possible to select a lower preset in main menu a shader would fail later at map load,
- 2: build all shaders on demand,
  makes sure no unused shader is built,
  a shader may be built while playing the game the first time the texture requiring it is displayed, producing stuttering

Option 1 is a good enough trade-off, but we still build many shaders for nothing. For example we will build the relief mapping and the deluxe mapping permutations even when loading a Tremulous map without such feature.

With option 1 the engine will also build any shader permutation for unused features, for example if we merge #258 that makes possible to generate normal maps from height maps at render time if normal maps are missing, that feature and all the permutations (with or without deluxe map, with lightmap,  with grid lighting…) will be built even if there is no asset missing a normal map in the currently played map.

With 0.55, the engine actually selects the rendering function at q3shader load time, meaning we may be able to just implement a loop that would render all the textures on some unseen framebuffer at map load time, triggering the GLSL shader build. That means such lazy shader loading would exhaustively build all required GLSL shaders at map load time, without missing used ones, and without building unused permutations.

illwieckz:
Actually the deluxe/lightmap permutation usage is only known at run time, so the suggestion would not fully work. But we may force the building of only a couple of them to avoid that.

Another implementation for such lazy loading would just be to render the whole world and all models to some unseen framebuffer in a frame at the end of map load.

slipher:
> Actually the deluxe/lightmap permutation usage is only known at run time, so the suggestion would not fully work. But we may force the building of only a couple of them to avoid that.
> 
> Another implementation for such lazy loading would just be to render the whole world and all models to some unseen framebuffer in a frame at the end of map load.

That would get you fairly close I suppose. There's probably still some ways that additional q3shaders can be loaded though, which could trigger more GLSL. Like maybe that map entity shader remap thing

illwieckz:
> Like maybe that map entity shader remap thing.

[...]

PR #1656, opened 2025-04-29
Improve lazy shader loading

VReaperV:
Implements #1294

Previously, when `r_lazyShaders 1` was used, it would simply load all of the possible shaders with the current graphics settings, and without deform shaders.

With this change, it will instead mark shaders that are used for BSP and model surfaces for building, which will be done in `RE_EndRegistration()`. Skins are also loaded (the variants with skeletal animation are always loaded because we don't know if an MD5 model will need them later).

This *greatly* reduces the amount of shaders built, and allows building deform shaders at map load. For example, on plat23 with pretty much all shader stuff enabled (except liquid shaders and static reflections), including material system, it went down from 433 shaders built at map load to just 28. I did know that a bunch of the shaders were unused, but this is still a surprising amount less.

Also made it so when a shader is loaded outside of `BuildAll()` (i. e. in `BindProgram()` or `GetProgram()`), the engine will print information about that.

From my testing, there are only a few shaders that this system doesn't get at map load, I'll see if I can make those load at the start as well in a different pr.

### DIFF
--- a/src/engine/renderer/GeometryOptimiser.cpp
+++ b/src/engine/renderer/GeometryOptimiser.cpp
@@ -451,6 +451,7 @@ std::vector<MaterialSurface> OptimiseMapGeometryMaterial(bspSurface_t** renderer
 	// std::unordered_map<TriEdge, TriIndex> triEdges;
 
 	vec3_t worldBounds[2] = {};
+	materialSystem.buildOneShader = false;
 	for ( int i = 0; i < numSurfaces; i++ ) {
 		bspSurface_t* surface = rendererSurfaces[i];
 
@@ -485,6 +486,8 @@ std::vector<MaterialSurface> OptimiseMapGeometryMaterial(bspSurface_t** renderer
 
 		materialSurfaces.emplace_back( srf );
 	}
+
+	materialSystem.buildOneShader = true;
 
 	materialSystem.GenerateWorldMaterialsBuffer();
 	materialSystem.GeneratePortalBoundingSpheres();
--- a/src/engine/renderer/GeometryOptimiser.cpp
+++ b/src/engine/renderer/GeometryOptimiser.cpp
@@ -129,9 +129,6 @@ void MarkShaderBuild( const shader_t* shader, const int lightMapNum, const bool 
 
 	tess.vboVertexSkinning = vertexSkinning;
 	tess.vboVertexAnimation = vertexAnimation;
-	/* Fuck knows why there are 2 different ways of setting the vertex animation global (through tess and through glState),
-	but that's what we have */
-	glState.vertexAttribsInterpolation = vertexAnimation ? 1.0f : 0.0f;
 
 	for ( const shaderStage_t* pStage = shader->stages; pStage < shader->lastStage; pStage++ ) {
 		pStage->shaderBuildMarker( pStage );
@@ -142,7 +139,6 @@ void MarkShaderBuild( const shader_t* shader, const int lightMapNum, const bool 
 
 	tess.vboVertexSkinning = false;
 	tess.vboVertexAnimation = false;
-	glState.vertexAttribsInterpolation = 0.0f;
 }
 
 static void CoreResetSurfaceViewCounts( bspSurface_t** rendererSurfaces, int numSurfaces ) {
--- a/src/engine/renderer/GeometryOptimiser.cpp
+++ b/src/engine/renderer/GeometryOptimiser.cpp
@@ -73,6 +73,78 @@ static int LeafSurfaceCompare( const void* a, const void* b ) {
 	return 0;
 }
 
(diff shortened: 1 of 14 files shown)


src/engine/renderer/Material.cpp
```diff
@@ -1027,8 +1027,6 @@ void ProcessMaterialLightMapping( Material* material, shaderStage_t* pStage, Mat
 	bool enableGridLighting = ( lightMode == lightMode_t::GRID );
 	bool enableGridDeluxeMapping = ( deluxeMode == deluxeMode_t::GRID );
 
-	DAEMON_ASSERT( !( enableDeluxeMapping && enableGridDeluxeMapping ) );
```
slipher, line 1030: The premise of `NUKE assert that doesn't really make sense` seems false in 2 out of 3 cases. Only in Render_lightMapping are the two tests based on the same variable.

src/engine/renderer/GeometryOptimiser.cpp
```diff
@@ -73,6 +73,78 @@ static int LeafSurfaceCompare( const void* a, const void* b ) {
...
+
+void MarkShaderBuild( const shader_t* shader, const int lightMapNum, const bool bspSurface,
+	const bool vertexSkinning, const bool vertexAnimation ) {
+	tess.bspSurface = bspSurface;
+	tess.lightmapNum = lightMapNum;
+
+	tess.vboVertexSkinning = vertexSkinning;
+	tess.vboVertexAnimation = vertexAnimation;
+	/* Fuck knows why there are 2 different ways of setting the vertex animation global (through tess and through glState),
+	but that's what we have */
```
slipher, line 133: tess.vboVertexAnimation is set to select the correct shader programs for MD3 models. Those shader programs need to be used even if there is no interpolation is done.

glState.vertexAttribsInterpolation is the interpolation fraction. The places using it to select a GLSL program are just wrong.

src/engine/renderer/Material.h
```diff
@@ -480,15 +482,15 @@ void BindShaderHeatHaze( Material* material );
...
-void ProcessMaterialReflection( Material* material, shaderStage_t* pStage, MaterialSurface* /* surface */ );
-void ProcessMaterialSkybox( Material* material, shaderStage_t* pStage, MaterialSurface* /* surface */ );
-void ProcessMaterialScreen( Material* material, shaderStage_t* pStage, MaterialSurface* /* surface */ );
-void ProcessMaterialHeatHaze( Material* material, shaderStage_t* pStage, MaterialSurface* surface );
-void ProcessMaterialLiquid( Material* material, shaderStage_t* pStage, MaterialSurface* /* surface */ );
-void ProcessMaterialFog( Material* material, shaderStage_t* pStage, MaterialSurface* surface );
+void ProcessMaterialNONE( Material*, shaderStage_t*, MaterialSurface*, const bool );
+void ProcessMaterialNOP( Material*, shaderStage_t*, MaterialSurface*, const bool );
+void ProcessMaterialGeneric3D( Material* material, shaderStage_t* pStage, MaterialSurface* /* surface */, const bool buildOneShader );
+void ProcessMaterialLightMapping( Material* material, shaderStage_t* pStage, MaterialSurface* surface, const bool buildOneShader );
```
slipher, line 488: There's an awful lot of "buildOneShader" plumbing. Why don't we just have a global variable that says whether we were supposed to have finished building the shaders?

VReaperV: Right, I'll change it to use `materialSystem.buildOneShader` instead. I'd only added it after changing the `Process*` functions, but it does make more sense to use it directly here.

src/engine/renderer/GeometryOptimiser.cpp
VReaperV: Oh, I see. I thought at first it was some weird attempt at optimisation - to only run the "regular" vertex shader when the interpolation fraction is 0.0.

src/engine/renderer/Material.cpp
VReaperV: Are they really? If I didn't miss anything they all just do essentially `( deluxeMode == deluxeMode_t::MAP ) != ( deluxeMode == deluxeMode_t::GRID )` in the assert. The code in other places was also directly copied from `Render_lightMapping()`.

slipher: Oh that's right.

src/engine/renderer/GeometryOptimiser.cpp
VReaperV: I added a commit fixing the original issue with `glState.vertexAttribsInterpolation` being used to check if vertex animation is being used.

src/engine/renderer/Material.h
VReaperV: Fixed.

slipher:
LGTM

Issue #1659, opened 2025-05-01
We may need to guess `CMAKE_FIND_ROOT` when cross-compiling


illwieckz:
When corss-compiling, CMake lines like this don't work:

```
find_library(LIBRT rt)
```

We need an explicit CMake option like `-DCMAKE_FIND_ROOT_PATH=/usr/aarch64-linux-gnu`.

For some reasons on some distributions or architectures it is not required to explicitly pass `-lrt` but on some, it is required to do it explicitly.

We may guess that root path in some ways, when it is not set. For example using `<compiler> -dumpmachine`.

slipher:
Using `find_library` with `rt` seems weird. Probably we should just be using `target_link_libraries(<target> rt)`. Then CMake will just pass in `-lrt` and let the linker find it itself.

But I suppose there are plenty of other libs, say SDL, that we might legitimately want to find sometimes.

illwieckz:
If `find_library` would find arm64 library on Linux amd64 by default it would be wrong, I guess?

illwieckz:
We don't need that patch in the dockered build script because we rebuild almost all the dependencies for the target platform, so `find_library` finds the library in `external_deps`, but outside of that scenario, it probably never worked.

But to make it easier and faster for me to test, I build arm64 binaries in a native amd64 Debian chroot with arm64 Debian packages, and run the produced binaries on an arm64 system running the exact same Debian version with exact same arm64 packages.

So `find_library` running on the amd64 build machine will never return arm64 libraries without telling CMake to explicitely look for them in the arm64 folder, and it is expected I guess (`find_library` returning arm64 libraries on an amd64 system by default would be a bug).

slipher:
It works on my Debian with just installing the per-architecture apt packages. For example if I configure a Daemon build with `CC=aarch64-linux-gnu-gcc CXX=aarch64-linux-gnu-g++ cmake ...` I get `LIBRT:FILEPATH=/usr/lib/aarch64-linux-gnu/librt.a`, `OPENGL_gl_LIBRARY:FILEPATH=/usr/lib/aarch64-linux-gnu/libGL.so`, `ZLIB_LIBRARY_RELEASE:FILEPATH=/usr/lib/aarch64-linux-gnu/libz.so` etc. in the `cmake -LA` output. The only shortcoming is it uses the system headers instead of arch-specific ones. It seems that each library is installed in at least 3 places: 
[...]

PR #1660, opened 2025-05-02
cmake: populate `CMAKE_FIND_ROOT_PATH` when cross-compiling on Linux

illwieckz:
Populate CMAKE_FIND_ROOT_PATH when cross-compiling on Linux.

Fix:

- https://github.com/DaemonEngine/Daemon/issues/1659


slipher:
Why is it inside `if (DEPS_DIR)` ?

illwieckz:
Right, it's not needed to be there, I moved it out of that block.

PR #1661, opened 2025-05-05
Fix copy-paste error

VReaperV:


slipher:
LGTM I guess.

I don't think that can happen anyway as fogs only occur as part of a BSP, so they can't be an MD3 surface.

VReaperV:
> I don't think that can happen anyway as fogs only occur as part of a BSP, so they can't be an MD3 surface.

They actually can, they just use `R_FogWorldBox()` to get the `fogNum`. E. g. for MD3:
> https://github.com/DaemonEngine/Daemon/blob/ab2349a35b27aaf3cb1a4f2bc1c1cd42b06e22d5/src/engine/renderer/tr_mesh.cpp#L307

With this change some maps crash there if the `heatHaze` shader is not loaded.

PR #1666, opened 2025-05-19
NUKE RDF_SKYBOXPORTAL usages

VReaperV:
This is half of a feature from Wolf:ET, but we don't have the sgame/cgame counterpart, and none of the maps made for Unvanquished/Tremulous appear to use it. This was done in Wolf:ET by using a `misc_skyportal` entity on the map, parsing it in sgame and sending the result through configstrings. Daemon/Unvanquished never actually sets `RDF_SKYBOXPORTAL`.

We have maps that use background-in-a-box (`freeway`, `covid` etc.), but they use a different mechanism and keep working correctly with this change.


slipher:
LGTM

PR #1670, opened 2025-05-21
NUKE forward lighting and shadow mapping

### LINKED ISSUE #747: Nuking the obsolete deprecated legacy forward renderer?
I think I'm going to nuke the obsolete dynamic lighting renderer. It doesn't work anymore on `master` (no dynamic light rendered) and the only reason why it is kept is that it has some features not ported to the tiled renderer.

But those features are either broken (dynamic shadows) or unused so maybe broken (no one knows how to use them).

For now I kept them around because it was an interesting reference implementation someone could borrow interesting things when re-implementing those features on the tiled renderer.

The legacy forward renderer is considered as deprecated since 2 years: https://github.com/DaemonEngine/Daemon/commit/be7229a2489971e14a3a636d962117012449478a

But no one tried to port the features in those two years, and unless there is an hidden skill in some of us right now no one of the active people has the skill to do such port.

Nuking this code will nuke thousands of lines of code.

What's your opinion?

slipher:
Fixes #747.

Shadow mapping no longer works at all. Forward lighting dynamic lights still more or less work but we don't want to maintain it.


VReaperV:
Good riddance to that shit.

src/engine/renderer/GeometryOptimiser.cpp
```diff
@@ -58,16 +58,16 @@ static int LeafSurfaceCompare( const void* a, const void* b ) {
 	}
 
 	// sort by leaf
-	if ( aa->interactionBits < bb->interactionBits ) {
+	if ( aa->scratch2 < bb->scratch2 ) {
```
VReaperV, line 61: Just get rid of this field? This part of the sorting does nothing here since `scratch1` and `scratch2` are never set, and `interactionBits` and `lightCount` are not used outside of the old forwardLighting/shadowing code.

src/engine/client/cg_msgdef.h
```diff
@@ -411,6 +411,7 @@ namespace Render {
 	using AddPolyToSceneMsg = IPC::Message<IPC::Id<VM::QVM, CG_R_ADDPOLYTOSCENE>, int, std::vector<polyVert_t>>;
 	using AddPolysToSceneMsg = IPC::Message<IPC::Id<VM::QVM, CG_R_ADDPOLYSTOSCENE>, int, std::vector<polyVert_t>, int, int>;
 	using AddLightToSceneMsg = IPC::Message<IPC::Id<VM::QVM, CG_R_ADDLIGHTTOSCENE>, std::array<float, 3>, float, float, float, float, float, int, int>;
+	// TODO(0.56): remove, it does the same as AddLightToSceneMsg
```
VReaperV, line 414: Already done in #1425.

src/engine/client/cl_cgame.cpp
```diff
@@ -1673,6 +1673,7 @@ void CGameVM::CmdBuffer::HandleCommandBufferSyscall(int major, int minor, Util::
                 break;
 
             case CG_R_ADDLIGHTTOSCENE:
+                // TODO(0.56): remove `shader` and maybe `flags`
```
VReaperV, line 1676: `shader` already NUKED in #1425, though `flags` can be deleted too.

src/engine/renderer/gl_shader.cpp
```diff
@@ -2558,160 +2554,6 @@ void GLShader_lightMappingMaterial::SetShaderProgramUniforms( ShaderProgramDescr
...
-	u_NormalScale( this ),
-	GLDeformStage( this ),
-	GLCompileMacro_USE_VERTEX_SKINNING( this ),
-	GLCompileMacro_USE_VERTEX_ANIMATION( this ),
-	GLCompileMacro_USE_HEIGHTMAP_IN_NORMALMAP( this ),
-	GLCompileMacro_USE_RELIEF_MAPPING( this )
-{
-}
-
-void GLShader_forwardLighting_projXYZ::BuildShaderCompileMacros( std::string& compileMacros )
```
VReaperV, line 2752: `BuildShaderCompileMacros` can now be removed entirely from `GLShader`, since it's not used anywhere else.

src/engine/renderer/tr_types.h
```diff
@@ -54,7 +54,6 @@ using bool8_t = uint8_t;
 #define MAX_WEIGHTS        4 // GPU vertex skinning limit, never change this without rewriting many GLSL shaders
 // XreaL END
 
-#define MAX_DLIGHTS        32 // can't be increased, because bit flags are used on surfaces
```
VReaperV, line 55: Already done in #1524.

slipher:
Note to self: also delete `lights/` in the dpkdir. The DPK stuff can be deleted in Unvanquished too.

src/engine/renderer/tr_backend.cpp
```diff
@@ -3057,12 +2523,6 @@ static void RB_RenderView( bool depthPass )
 
 	GL_CheckErrors();
 
-	if ( r_speeds->integer == Util::ordinal(renderSpeeds_t::RSPEEDS_SHADING_TIMES) )
```
VReaperV, line 4472: This seems irrelevant to the pr?

slipher (commented):
Removed already-done TODOs and prevented any easily-avoidable merge conflicts with 0.56. But regardless I'll have to remember to go merge this into for-0.56 as soon as it's merged into master.

src/engine/renderer/gl_shader.cpp
slipher: Done.

src/engine/renderer/GeometryOptimiser.cpp
VReaperV: I see, I suppose it's needed for sorting.

src/engine/renderer/tr_backend.cpp
VReaperV: Hmm, yeah, apparently it also captures SSAO and motion blur.

VReaperV (approved):
LGTM

PR #1673, opened 2025-05-31
detection of ban-related disconnection reasons on the server side

Grise3:
This PR adds detection of ban-related disconnection reasons on the server side. It parses the ban details (such as who banned the player, ban duration, and reason) from the disconnection message and stores this information in variables accessible by the user interface.

Note: This PR should be merged simultaneously with the corresponding PR in the cgame repository that implements the ban window UI to ensure full functionality.


slipher:
Such parsing logic would be better in the cgame, not in Daemon. The ban message is produced by the sgame and consumed by the cgame - both gamelogic, so there is no reason for the engine to care about its format. And that way all of your code goes in a single repository and you don't need to wait for a new version.

PR #1678, opened 2025-06-17
Staging buffer and material system clean-up

VReaperV:
- Fix staging buffer pointer not being reset when freeing it
- Uniforms are now tightly packed into the material struct where possible, e. g. it will try to put a 1-component uniform after a vec3
- Cleaned up some unused material system stuff

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -2130,6 +2130,7 @@ void GLShader::PostProcessUniforms() {
 		size = components ? PAD( size, 4 ) * components : size;
 		GLuint alignmentConsume = 4 - size % 4;
 
+		GLUniform* tmpUniform = _materialSystemUniforms[0];
 		tmp.emplace_back( _materialSystemUniforms[0] );
 		_materialSystemUniforms.erase( _materialSystemUniforms.begin() );
 
@@ -2138,8 +2139,14 @@ void GLShader::PostProcessUniforms() {
 			&& ( uniform = FindUniformForAlignment( _materialSystemUniforms, alignmentConsume ) ) != -1 ) {
 			alignmentConsume -= _materialSystemUniforms[uniform]->GetSTD430Size();
 
+			tmpUniform = _materialSystemUniforms[uniform];
+
 			tmp.emplace_back( _materialSystemUniforms[uniform] );
 			_materialSystemUniforms.erase( _materialSystemUniforms.begin() + uniform );
+		}
+
+		if ( alignmentConsume ) {
+			tmpUniform->SetSTD430Size( tmpUniform->GetSTD430Size() + alignmentConsume );
 		}
 
 		size = PAD( size, 4 );
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -2094,13 +2094,9 @@ GLint GLShader::GetUniformLocation( const GLchar *uniformName ) const {
 
 static int FindUniformForAlignment( std::vector<GLUniform*>& uniforms, const GLuint alignment ) {
 	for ( uint32_t i = 0; i < uniforms.size(); i++ ) {
-		if ( uniforms[i]->GetSTD430Size() == alignment ) {
+		if ( uniforms[i]->GetSTD430Size() <= alignment ) {
 			return i;
 		}
-	}
-
-	if ( uniforms[uniforms.size() - 1]->GetSTD430Size() < alignment ) {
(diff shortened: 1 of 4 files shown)


src/engine/renderer/gl_shader.cpp
```diff
@@ -1582,45 +1582,19 @@ std::string GLShaderManager::ShaderPostProcess( GLShader *shader, const std::str
...
-				materialStruct += "[ " + std::to_string( uniform->GetComponentSize() ) + " ]";
-			}
-			materialStruct += ";\n";
-
-			// vec3 is aligned to 4 components, so just pad it with int
-			// TODO: Try to move 1 component uniforms here to avoid wasting memory
-			if ( uniform->GetSTD430Size() == 3 ) {
-				materialStruct += "	int ";
-				materialStruct += uniform->GetName();
-				materialStruct += "_padding;\n";
```
slipher, line 1603: I don't see any replacement for this. Does that mean it is going to break if there aren't enough 4-byte uniforms to fill all the vec3 holes?

src/engine/renderer/gl_shader.cpp
```diff
@@ -2113,71 +2085,73 @@ uint32_t* GLUniform::WriteToBuffer( uint32_t* buffer ) {
...
-		GLuint highestAlignment = 0;
-		int highestUniform = 0;
-		for( uint i = 0; i < _uniforms.size(); i++ ) {
-			if ( _uniforms[i]->GetSTD430Alignment() > highestAlignment ) {
-				highestAlignment = _uniforms[i]->GetSTD430Alignment();
-				highestUniform = i;
-			}
+		GLuint size = _materialSystemUniforms[0]->GetSTD430Size();
+		GLuint components = _materialSystemUniforms[0]->GetComponentSize();
+		size = components ? 4 * components : size;
```
slipher, line 2134: I believe that should be `components * size` rather than `4 * components`

src/engine/renderer/gl_shader.cpp
```diff
@@ -2113,71 +2085,75 @@ uint32_t* GLUniform::WriteToBuffer( uint32_t* buffer ) {
...
+		GLuint alignmentConsume = 4 - size % 4;
+
+		tmp.emplace_back( _materialSystemUniforms[0] );
+		_materialSystemUniforms.erase( _materialSystemUniforms.begin() );
+
+		int uniform;
+		GLuint offset = 4 - alignmentConsume;
```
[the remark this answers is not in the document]
VReaperV: Already fixed, yeah.

src/engine/renderer/gl_shader.cpp
VReaperV: Fixed. Did `PAD( size, 4 ) * components` there because arrays/array members are aligned to a vec4.

src/engine/renderer/gl_shader.cpp
slipher: But `WriteToBuffer` is now assuming that there is never padding.

src/engine/renderer/gl_shader.cpp
```diff
@@ -2142,6 +2143,10 @@ void GLShader::PostProcessUniforms() {
 			_materialSystemUniforms.erase( _materialSystemUniforms.begin() + uniform );
 		}
 
+		if ( alignmentConsume ) {
+			tmpUniform->SetSTD430SizeToAlignment();
```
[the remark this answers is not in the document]
VReaperV: Each uniform in each shader is a unique object.

src/engine/renderer/gl_shader.cpp
```diff
@@ -2113,71 +2085,74 @@ uint32_t* GLUniform::WriteToBuffer( uint32_t* buffer ) {
...
 		}
-			
-		const GLuint size = _uniforms[highestUniform]->GetSTD430Size();
-		if ( _uniforms[highestUniform]->GetComponentSize() != 0 ) {
-			structSize += ( size == 3 ) ? 4 * _uniforms[highestUniform]->GetComponentSize() :
-												size * _uniforms[highestUniform]->GetComponentSize();
-		} else {
-			structSize += ( size == 3 ) ? 4 : size;
+
+		if ( alignmentConsume ) {
```
slipher, line 2146: This `while` followed by `if` structure doesn't really make sense since `SetSTD430SizeToAlignment()` would only work correctly if the while loop ran 0 times. `if` (uniform with exactly the needed alignment found) insert it `else` set padding, would make more sense

slipher (commented):
LGTM modulo comment

Issue #1676, opened 2025-06-11
Tiled artifacts with smoke


slipher:
![Image](https://github.com/user-attachments/assets/2e7adb24-d9e0-470b-8cc9-83950832f7b0)

Some of the light tiles are seemingly poking holes through the smoke. The bug does not happen if setting `r_drawDynamicLights` 0. Happens with both 0.55.0 engine and a recent engine.

The tiled realtime light renderer has a known limitation with non-depth-writing textures since it filters out lights if they are out of range according to the depth buffer. We could try to blame it on that (if the grate textures are non-depth-writing), but I don't think that could explain why the smoke is not visible at all in some spots.

slipher:
I tried setting `attenuation = 0.0;` in the `computeDynamicLight` GLSL function. Having lights with 0 intensity ought to be the same as having no lights, but the same artifact is produced. So there must be a real bug.

slipher:
In fact it DOES happen with `r_drawDynamicLights 0`. Must have messed up the previous test somehow. After [fixing r_showLightTiles](https://github.com/DaemonEngine/Daemon/pull/1677), I see that the light tiles are bigger than the squares I'm getting here here. And there aren't even any dynamic lights being rendered in the messed up area. So nothing to do with dynamic lights. Tile-patterned artifacts can happen with various driver/GPU bugs such as in #949

slipher:
Doesn't happen for me with Intel or Nvidia proprietary drivers.

Easy repro steps:
- `devmap uvcs`
- `setviewpos -2242 -1267 184 177 -27`
- `testshader gfx/weapons/rifle/smoke` or `testshader gfx/buildables/acid_tube/spore`

The triggering condition for the foreground shader is use of depth fade. I haven't investigated yet what needs to be in the background.

![Image](https://github.com/user-attachments/assets/6319b114-dff2-4c41-9ff7-7e3065f03268)

VReaperV:
Btw does switching relief mapping affect it in any way?

slipher:
It happens in both ultra and low presets so does not depend on relief mapping setting.

slipher:
The depth value as read in the generic shader with depth fade enabled, with the color formula `1 / (1.0001 - depth) / 600`.

![Image](https://github.com/user-attachments/assets/7e9872e1-487c-40a8-be19-076174b86ff2)

[...]

PR #1684, opened 2025-06-23
Implement PushBuffer

VReaperV:
Requires #1682

Add `PushBuffer`, update uniforms to have an `updateType`. The `CONST` and `FRAME` uniforms are written into a uniform buffer after the map is loaded/at the start of a frame. This allows skipping some of the `glUniform*()` calls to decrease the amount of overhead.

This largely re-uses the existing functionality added for the material system.


slipher:
This can be rebased now.

slipher:
Some issues found with testing:
1. Fog is not rendered if material system is enabled.
2. In one scene from my test script the dynamic light from the Painsaw is not rendered. I can't observe this with manual testing so I'd have to investigate more to provide a reliable repro.

src/engine/renderer/gl_shader.h
```diff
@@ -627,22 +651,21 @@ class GLUniform1i : protected GLUniform
...
 
 	inline void SetValue( uint value ) {
 		ShaderProgramDescriptor* p = _shader->GetProgram();
 
-		if ( _global || !_shader->UseMaterialSystem() ) {
-			ASSERT_EQ( p, glState.currentProgram );
-		}
-
-		if ( _shader->UseMaterialSystem() && !_global ) {
+		if ( ( _shader->UseMaterialSystem() && _updateType == MATERIAL_OR_PUSH )
```
slipher, line 661: Helper function for this? It's a cryptic and unwieldy expression to have to repeat so many times.

src/engine/renderer/Material.cpp
```diff
@@ -1495,6 +1495,22 @@ void MaterialSystem::UpdateDynamicSurfaces() {
...
+		surfaceDescriptorsCount / MAX_COMMAND_COUNTERS : surfaceDescriptorsCount / MAX_COMMAND_COUNTERS + 1;
+
+	globalUBOProxy->SetUniform_FirstPortalGroup( globalWorkGroupX );
+	globalUBOProxy->SetUniform_TotalPortals( totalPortals );
+}
+
+void MaterialSystem::SetFrameUniforms() {
+	globalUBOProxy->SetUniform_Frame( nextFrame );
+
+	globalUBOProxy->SetUniform_UseFrustumCulling( r_gpuFrustumCulling.Get() );
```
slipher, line 1510: Can you get rid of the other code that sets these? Since material system without PushBuffer shouldn't be a possible configuration. (You can force it with cvars but #1708 fixes that.)

src/engine/renderer/gl_shader.h
VReaperV: I reworked that part in #1685, it makes the whole uniform caching part much more straightforward.

src/engine/renderer/Material.cpp
VReaperV: Yeah, forgot about those. Done now.

VReaperV:
> 1. Fog is not rendered if material system is enabled.

Fixed.

VReaperV:
> Now I found a way. Set `common.framerate.max 3` and fire the Painsaw once. On master you will see a frame with the realtime light rendered, but on this branch no. Maybe there is some issue with data being a frame out of date.

Also fixed.

slipher:
I confirm the 2 previous bugs are fixed. But I get the depth fade issue from #1676 again with this branch. It happens with any depth fade regardless of alpha tests, polygon offset etc. The bug reappears as of the first usable commit `Add the remaining code for the global UBO in PushBuffer`. (The first four commits don't build; `GLUniform._global -> _updateType` builds but doesn't render anything.)

VReaperV:
> It was broken but #1704 fixed it. It works on master.

Not sure what the issue is. I can't reproduce it on my end, and there don't seem to be any depth writes after `glTextureBarrier()`.

slipher:
The depth texture is all garbage all the time, as if it's not hooked up properly. It's still garbage if I set `r_materialSystemSkip 1`. It's still garbage with `r_drawWorld 0` and nothing being rendered.

By the way, I realized you can make a depth texture visualization with just a q3shader, no GLSL required:

```
depthvis {
    {
        map _white
        depthFade 600
        blendFunc GL_SRC_ALPHA GL_ZERO
    }
}
```

slipher:
> This is on AMD + Mesa, right?

Yes


> It might have the same cause as the broken occlusion culling reported there ~1 year ago.

On master if I turn on occlusion culling, it segfaults immediately.

VReaperV:
> On master if I turn on occlusion culling, it segfaults immediately.

Is the segfault coming from daemon? I don't get any on master.

slipher:
> Hmm, the only thing I could think of is it has something to do with the textures being bindless.

That's true, just `r_preferBindlessTextures` is enough to trigger the bug.

VReaperV:
> > Hmm, the only thing I could think of is it has something to do with the textures being bindless.
> 
> That's true, just `r_preferBindlessTextures` is enough to trigger the bug.

Even with the fix from #1725?

VReaperV:
Nevermind, if material system isn't enabled #1725 will have no effect.

VReaperV:
Anyway, I think it's just a driver bug.

VReaperV:
Actually no, there *is* a missing texture barrier, before depthtile.

VReaperV:
I see. One more possible workaround is to special-case the depth map, so it never goes into the buffer on AMD + Mesa.

VReaperV:
CI failure is unrelated.

VReaperV:
> I also uncovered another issue: when disabling `r_arb_direct_state_access` (and hence push buffers), the GLSL does not compile.

Fixed.

VReaperV:
I've got confirmation that it works fine on AMD proprietary with `r_readonlyDepthBuffer 1`.

VReaperV:
> Besides the depth fade issues on AMD + Mesa, I get a couple problems with the current branch:
> 
> * `Warn: GL_BindToTMU: NULL image`
> * Lighting is globally dimmer but with more saturated areas (as if tonemapping disabled maybe?)
> 
> The `slipher/pushbuffer-rebase` branch does not have those extra issues. I don't know if any changes were made since the rebase.

First issue fixed, but I'm not reproducing the second one on NV. If it still happens and `slipher/pushbuffer-rebase` still works correctly, then the latter can be merged instead. The only real difference I'm seeing is in how uniform padding is calculated for PushBuffer, and whether `pushBuffer.frameUniformsSize` includes padding, but the effect should be the same.

src/engine/renderer/GLMemory.h
```diff
@@ -175,6 +175,7 @@ class GLBuffer {
 
 	void DelBuffer() {
 		glDeleteBuffers( 1, &id );
+		id = 0;
```
slipher: The constructor should also set this to 0 if it is going to be used as a test

src/engine/renderer/tr_bsp.cpp
```diff
@@ -4641,6 +4641,57 @@ static void SetWorldLight() {
...
+	globalUBOProxy->SetUniform_SRGB( tr.worldLinearizeTexture );
+
+	globalUBOProxy->SetUniform_ProfilerZero();
+
+	if ( glConfig.usingBindlessTextures ) {
+		if ( glConfig.colorGrading ) {
+			globalUBOProxy->SetUniform_ColorMap3DBindless( GL_BindToTMU( 3, tr.colorGradeImage ) );
+		}
+
+		globalUBOProxy->SetUniform_DepthMapBindless( GL_BindToTMU( 1, tr.currentDepthImage ) );
```
slipher: Is this right? In the majority of cases with this binding `tr.depthSamplerImage` is used.

slipher (commented):
So now the MATERIAL_OR_PUSH argument just has to be added to the test uniform constructors for it to compile.

LGTM modulo comments

PR #1685, opened 2025-06-23
Only update PushBuffer if any uniform has actually changed

VReaperV:
Requires #1684

Further improvement for PushBuffer, only update it if it's actually required.

### DIFF
--- a/src/engine/renderer/gl_shader.h
+++ b/src/engine/renderer/gl_shader.h
@@ -583,12 +583,6 @@ class GLUniformSampler : protected GLUniform {
 	}
 
 	public:
-	void SetValue( GLuint value ) {
-		if ( !CacheValue( &value ) ) {
-			return;
-		}
-	}
-
 	void SetValueBindless( GLuint64 value ) {
 		if ( !glConfig.usingBindlessTextures ) {
 			return;


VReaperV:
Also added a fix for `globalFog` fog texture.

src/engine/renderer/gl_shader.h
```diff
@@ -545,38 +591,22 @@ class GLUniformSampler : protected GLUniform {
 	}
 
 	void SetValue( GLuint value ) {
-		currentValue = value;
+		if ( !CacheValue( &value ) ) {
```
slipher: Pointless if statement

src/engine/renderer/gl_shader.h
```diff
@@ -2606,7 +2376,7 @@ class u_Bones :
 {
 public:
 	u_Bones( GLShader *shader ) :
-		GLUniform4fv( shader, "u_Bones", MAX_BONES, PUSH )
+		GLUniform4fv( shader, "u_Bones", MAX_BONES * 2, PUSH )
```
slipher: Why times two?

src/engine/renderer/gl_shader.h
```diff
@@ -1055,59 +813,34 @@ class GLUniformMatrix32f : protected GLUniform {
...
 
 	inline void SetValue( GLboolean transpose, const vec_t* m ) {
-		ShaderProgramDescriptor* p = _shader->GetProgram();
+		vec_t value[12] {};
+		memcpy( value, m, 6 * sizeof( vec_t ) );
 
-		if ( ( _shader->UseMaterialSystem() && _updateType == MATERIAL_OR_PUSH )
-			|| ( glConfig.pushBufferAvailable && _updateType <= FRAME ) ) {
-			memcpy( currentValue, m, 6 * sizeof( float ) );
+		if ( !CacheValue( value ) ) {
```
slipher: I think this could give an incorrect result if called with differing `transpose` parameters.

src/engine/renderer/gl_shader.h
VReaperV: Because the bone count multiplied by 2 in `SetUniform_Bones()`.

libs/breakpad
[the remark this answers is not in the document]
VReaperV: Ugh, stupid submodules... Should be fixed now.

src/engine/renderer/gl_shader.h
VReaperV: I've removed that function entirely since it's unused.

slipher (commented):
LGTM. Looks good in my screenshot tests

PR #1686, opened 2025-06-25
tr_shader: fix “flags & RSF_2D”

illwieckz:
Something wrong I noticed while reading the code, because `bits` is set to `IF_NONE`, this test was always false, and then I noticed `RSF_2D` is expected to be stored in `flags`…


VReaperV (approved):
LGTM

PR #1688, opened 2025-06-27
Implement con_persistOnMapChange

VReaperV:
If enabled, the current console input will be kept on map change the first time the console is opened after the new map is loaded.


enneract:
LGTM

PR #1689, opened 2025-06-28
Update MAX_ENTITIES, NUKE some unused defines

VReaperV:


illwieckz:
What means that comment?

> can't be increased without changing drawsurf bit packing`

If that comment is true, how can we bump that number? If it's not true anymore, do we need to keep that comment?

VReaperV:
> What means that comment?
> 
> > can't be increased without changing drawsurf bit packing`
> 
> If that comment is true, how can we bump that number? If it's not true anymore, do we need to keep that comment?

https://github.com/DaemonEngine/Daemon/blob/master/src/engine/renderer/tr_local.h#L1521

I've already increased it, but this define has to be updated in the for-0.56 branch. The comment is still correct, it's just updated now to the correct value.

illwieckz:
OK thx.

illwieckz (approved):
LGTM.

PR #1691, opened 2025-07-05
sdl_glimp: require VAO explicitly

illwieckz:
Require VAO explicitly, because of #1690:

- https://github.com/DaemonEngine/Daemon/pull/1690

Also, because I like explicit logs, I added a log line for VBO as well, this way we have explicit log lines for VAO, VBO and FBO.

I also reworded things a bit.


VReaperV (approved):
LGTM

PR #1692, opened 2025-07-06
engine: load worldspawn separately from other map data

illwieckz:
Make possible to load the worldspawn separately from other map data.

It's related to me trying to implement native sRGB support:

- https://github.com/DaemonEngine/Daemon/pull/1687

I'm afraid we have to know if the rendering should use sRGB/linear conversions before rendering anything else from the UI, including the map loading menu itself.

Because right now, at the time we know the map requires sRGB/linerar conversion, the UI is already created.

I'm not sure what I'm doing is right, but this is the problem I attempt to solve.

See on client side:

- https://github.com/Unvanquished/Unvanquished/pull/3390

### DIFF
--- a/src/dummygame/cgame.cpp
+++ b/src/dummygame/cgame.cpp
@@ -155,7 +155,8 @@ public:
 			// the renderer starts (though there is no way to get it back later)
 			refdef.gradingWeights[0] = 1.0f;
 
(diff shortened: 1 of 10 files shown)


illwieckz:
This seems unneeded for now.

PR #1693, opened 2025-07-11
Fix /listmaps with pak prefix + /map bug with invalid names

slipher:
1. fix `listmaps` commands not respecting `fs_pakprefixes`
2. fix the server dropping in https://github.com/Unvanquished/Unvanquished/issues/3080


illwieckz:
> If a directory name starts with "map-" any pak in the directory will be confused for a map.

Funny!

PR #1695, opened 2025-07-12
Allow loading an image multiple times with different params

slipher:
E.g. different nopicmip or imageMaxDimension.

Fixes https://github.com/Unvanquished/Unvanquished/issues/2801

This could be useful if an image needs to be loaded with and without sRGB conversion (`IF_` flags are part of the image params).

We have a handful of images in game assets that will be loaded multiple times:

```
image params mismatch for gfx/weapons/tracer/spark: 0x2 0 0/0 128 0 vs. 0x0 0 0/0 0 0
image params mismatch for models/buildables/arm/screen_s: 0x0 0 0/0 128 0 vs. 0x0 0 0/0 16 0
image params mismatch for models/players/human_male/base: 0x0 0 0/0 0 0 vs. 0x0 0 0/0 128 0
```

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -27,6 +27,10 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include <iomanip>
 #include "Material.h"
 
+static Cvar::Cvar<bool> r_allowImageParamMismatch(
+	"r_allowImageParamMismatch", "reuse images when requested with different parameters",
+	Cvar::NONE, false);
+
 int                  gl_filter_min = GL_LINEAR_MIPMAP_NEAREST;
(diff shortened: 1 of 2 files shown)


illwieckz:
Great! Indeed we may want to render an image in the UI or to render the same image on a 3D surface, so many flags may differ (fitscreen, srgb…). One good example of that is an ATCS clone that replaced the end game firework sprite with the granger emoticon.

illwieckz:
Maybe the `_black`, `_red`, `_green_` and `_blue` images are exceptions too? At least for sRGB conversions `0` and `1` are always `0` and `1`.

illwieckz:
OK, this is good enough for a start.

> I don't think `_red`, `_green`, or `_blue` are used that often.

I may one day modify the `shared_colors` materials to use them in engine. But that's not a big issue: those images are all very small.

illwieckz (approved):
LGTM.

PR #1699, opened 2025-07-16
Collapse shaderType_t::3D_DYNAMIC and 3D_STATIC into one

slipher:
There was no difference in behavior between them.


illwieckz (approved):
LGTM.

PR #1700, opened 2025-07-16
Allow registering shaders multiple times with different RSF_ flags

slipher:
Same idea as #1695 but for q3shaders instead of images.

Stacked on #1699.

It's a draft in case we want to make changes related to distinguishing "2D" vs. "3D" shaders when deciding whether to use sRGB.


slipher:
Ready for review. I renamed `RSF_IMPLICIT_3D` to `RSF_3D` because `RSF_2D` *also* only affected implicit shaders. So I figure either both should have implicit in the name, or neither.

illwieckz (approved):
LGTM.

PR #1702, opened 2025-07-19
sdl_glimp: set `APPID` as `WMCLASS` on Linux

illwieckz:
Let X11 and Wayland desktops (Linux, FreeBSD…) display the game name on the game windows instead of the engine binary name. This also makes possible to display the correct icon if that `WMCLASS` is also set in the `net.unvanquished.Unvanquished.desktop` file.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1703,6 +1703,20 @@ static rserr_t GLimp_StartDriverAndSetMode( int mode, bool fullscreen, bool bord
 {
 	int numDisplays;
 
+#if !defined(_WIN32) && !defined(__APPLE__)
+	/* Let X11 and Wayland desktops (Linux, FreeBSD…) associate the game
+	window with the XDG .desktop file, with the proper name and icon.
+	The .desktop file should have PRODUCT_APPID as base name or set the
+	StartupWMClass variable to PRODUCT_APPID. */
+
+	// SDL2.
+	Sys::SetEnv( "SDL_VIDEO_X11_WMCLASS", PRODUCT_APPID );
+	Sys::SetEnv( "SDL_VIDEO_WAYLAND_WMCLASS", PRODUCT_APPID );
+
+	// SDL3.
+	Sys::SetEnv( "SDL_HINT_APP_ID", PRODUCT_APPID );
+#endif
+
 	if ( !SDL_WasInit( SDL_INIT_VIDEO ) )
 	{
 		const char *driverName;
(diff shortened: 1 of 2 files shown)


illwieckz:
To get the icon properly displayed we need to update the desktop file installed by the updater, but this is enough to make the game name properly displayed.

illwieckz:
I also added the `SDL_HINT_APP_ID` variable, which is used by SDL3 instead of `SDL_VIDEO_X11_WMCLASS` and `SDL_VIDEO_WAYLAND_WMCLASS` from SDL2. We're not using SDL3 yet but this way we are ready. Also some people are already using the SDL2 compatibility layer of SDL3 and maybe the APPID is already supported this way.

slipher:
I wondered which desktops do this. Apparently [GNOME is one](https://stackoverflow.com/questions/75203814/why-does-my-sdl-linux-application-have-the-correct-icon-in-the-app-menu-but-not).

illwieckz:
Our problem is that the desktop file is starting the `updater` binary that then launches the `daemon` binary… So we need some way to let the desktop knows what is what.

It would be good to find a similar mechanism to implement in the updater too. On my end GNOME fails to associate the Unvanquished launcher icon with both the updater window and the engine windows.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1703,6 +1703,17 @@ static rserr_t GLimp_StartDriverAndSetMode( int mode, bool fullscreen, bool bord
 {
 	int numDisplays;
 
+#if !defined(_WIN32) && !defined(__APPLE__)
+	/* SDL2: Let X11 and Wayland desktops (Linux, FreeBSD…) display the game
+	name on the game windows instead of the engine binary name. This also
+	makes possible to display the correct icon if that WMCLASS is also set
+	in the net.unvanquished.Unvanquished.desktop file. */
+	Sys::SetEnv( "SDL_VIDEO_X11_WMCLASS", PRODUCT_NAME );
```
slipher, line 1711: These should also be PRODUCT_APPID

illwieckz: Oh lol, I did not noticed that the default WMCLASS was the APPID, which makes sense btw.

So now all those hints are using `PRODUCT_APPID`.

This has the side effect of displaying `net.unvanquished.Unvanquished` in the taskbar if there is no XDG icon though.

slipher:
LGTM. Setting `SDL_VIDEO_WAYLAND_WMCLASS` works for me for setting the icon while using Wayland with KDE.

illwieckz:
I only reworded the comment a bit.

PR #1704, opened 2025-07-21
Fix depth fade tile-shaped artifacts 

### LINKED ISSUE #1676: Tiled artifacts with smoke

slipher:
Fix #1676.


illwieckz (approved):
LGTM.

PR #1706, opened 2025-07-24
Fix GLEW->GLU dependency regression

slipher:
@illwieckz Here's the thing to prevent the GLU dependency.


illwieckz:
Thanks!

illwieckz (approved):
LGTM.

PR #1709, opened 2025-07-25
Clean up remnants of forward lighting code

slipher:


illwieckz:
That reminds me a question I had: do we need those default images with all pixels being the same color to be `8×8` in size?

In `tex-common` they are `1×1` in size and it works fine.

illwieckz (approved):
LGTM.

PR #1718, opened 2025-08-01
Fix some deprecated std stuff

VReaperV:
Some fixes for deprecated stuff I did in daemon-vulkan, backported here. These ones don't require any C++ version upgrade.

### DIFF
--- a/src/common/Util.h
+++ b/src/common/Util.h
@@ -151,6 +151,8 @@ public:
 	uninitialized& operator=(const uninitialized&) = delete;
 	uninitialized& operator=(uninitialized&&) = delete;
 
+	alignas( T ) uint8_t data[sizeof( T )];
+
 	template<typename... Args> void construct(Args&&... args)
 	{
 		new(&data) T(std::forward<Args>(args)...);
@@ -171,9 +173,6 @@ public:
 	{
 		return *reinterpret_cast<const T*>(&data);
 	}
-
-private:
-	typename std::aligned_storage<sizeof(T), std::alignment_of<T>::value>::type data;
 };
 
 // std::make_unique is not available until C++14.
--- a/src/common/Util.h
+++ b/src/common/Util.h
@@ -151,8 +151,6 @@ public:
 	uninitialized& operator=(const uninitialized&) = delete;
 	uninitialized& operator=(uninitialized&&) = delete;
 
-	alignas( T ) uint8_t data[sizeof( T )];
-
 	template<typename... Args> void construct(Args&&... args)
 	{
 		new(&data) T(std::forward<Args>(args)...);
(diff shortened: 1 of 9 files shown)


slipher:
> MacOS compiler in the CI is being a piece of shit for some reason.

We saw this CI glitch in #1687 as well, where the compiler acts like it is in C++11 mode. I tried locally with AppleClang 17.0 and was unable to reproduce the issue: the `std=gnu++14` is passed and it builds without issues. The CI build log shows the respective flag test passing, so it should be used there too. Maybe we should try upgrading the VM image in `azure-pipelines.yml`. It's using AppleClang 15.0 so the bug could be fixed in a newer version.

VReaperV:
> > MacOS compiler in the CI is being a piece of shit for some reason.
> 
> We saw this CI glitch in #1687 as well, where the compiler acts like it is in C++11 mode. I tried locally with AppleClang 17.0 and was unable to reproduce the issue: the `std=gnu++14` is passed and it builds without issues. The CI build log shows the respective flag test passing, so it should be used there too. Maybe we should try upgrading the VM image in `azure-pipelines.yml`. It's using AppleClang 15.0 so the bug could be fixed in a newer version.

I tried `macOS-latest` earlier, CI still had the same issue unfortunately :/

VReaperV:
Hmm, maybe not? The CI still used AppleClang 15 I think, the one on `macOS-latest` should be AppleClang 18.

VReaperV:
Nah, CI still just uses AppleClang 15.

slipher:
Apparently, [`macOS-latest` is not really the latest one](https://devblogs.microsoft.com/devops/upcoming-updates-for-azure-pipelines-agents-images/). You have to write `macOS-15`.

VReaperV:
Still fails, even though it's using AppleClang 16.

src/common/Util.h
```diff
@@ -151,6 +151,8 @@ template<typename T> class uninitialized {
 	uninitialized& operator=(const uninitialized&) = delete;
 	uninitialized& operator=(uninitialized&&) = delete;
 
+	alignas( T ) uint8_t data[sizeof( T )];
```
slipher, line 154: Keep this as private member

src/engine/audio/Audio.cpp
```diff
@@ -222,13 +222,13 @@ namespace Audio {
         for (auto &loop : entityLoops) {
             loop.addedThisFrame = false;
             // if we are the unique owner of a loop pointer, then it means it was stopped, free it.
-            if (loop.sound.unique()) {
+            if (loop.sound.use_count()) {
```
slipher, line 225: The correct replacement is `use_count() == 1`

src/common/Util.h
VReaperV: Fixed now.

src/engine/audio/Audio.cpp
VReaperV: Thx, missed that one. Fixed.

PR #1719, opened 2025-08-01
Add some support for C++17-23

VReaperV:
Add `CPPStandard.h` for checking C++ features support by version, replace deprecated `std::result_of` on >= C++17. Use the `/Zc:__cplusplus` for MSVC CI so it will get the correct version in `CPPStandard`.

### DIFF
--- a/src/common/CPPStandard.h
+++ b/src/common/CPPStandard.h
@@ -1,0 +1,90 @@
+/*
+===========================================================================
+
+Daemon BSD Source Code
+Copyright (c) 2025 Daemon Developers
+All rights reserved.
+
+This file is part of the Daemon BSD Source Code (Daemon Source Code).
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of the Daemon developers nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL DAEMON DEVELOPERS BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+===========================================================================
+*/
+// CPPStandard.h
+
+#ifndef CPPSTANDARD_H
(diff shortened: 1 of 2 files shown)


.appveyor.yml
```diff
@@ -53,7 +53,7 @@ build_script:
 
       set CFLAGS=/Wv:19.29.30037
 
-      set CXXFLAGS=/Wv:19.29.30037
+      set CXXFLAGS=/Wv:19.29.30037 /Zc:__cplusplus
```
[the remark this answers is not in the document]
VReaperV: I think the CI's not picking it up for some reason.

src/common/CPPStandard.h
```diff
@@ -0,0 +1,66 @@
...
+#define CPPSTANDARD_H
+
+enum CPPStandard_ {
+  CPP14,
+  CPP17,
+  CPP20,
+  CPP23
+};
+
+#if __cplusplus == 201402L
```
slipher, line 46: The value of  `__cplusplus` should never be used for anything. Some compiler vendors understand this to mean 100% implementation status of the standard. So you can have a compiler with C++17 99% complete but there is one missing bit from C++11, so `__cplusplus` indicates C++03. There are more specific feature test macros which *might* be useful.

slipher (commented):
What do we need from C++20 or C++23?

VReaperV:
> What do we need from C++20 or C++23?

Daemon-vulkan currently specifically uses designated initializers, concepts, <source_location>, atomic wait/notify, and <stacktrace> from there. Additionally, <source_location> and <stacktrace> are used for better logging of various issues, with extra information being added if a cvar is enabled. That part is not specific to daemon-vulkan and can be used throughout the engine and cgame/sgame (the latter only with dll/native exe ofc).

src/common/CPPStandard.h
VReaperV: Who cares about those compilers? I don't see the issue as long as it builds on recent clang/gcc/msvc. Nacl is in kind of its own category since it's stuck with ancient clang.

slipher: GCC and Clang are such compilers. See for example https://stackoverflow.com/questions/30995705/cplusplus-201402l-return-true-in-gcc-even-when-i-specified-std-c14
Also this shows that equality tests are not good because there can be intermediate values.

VReaperV: Ehh, I'm generally just ensuring that it works on the currently latest clang/gcc and near-latest msvc, so if something's not supported by one of those I'll probably just skip it for the time being (since in the case of non-optional features the only other option is implementing some sort of fallback, which is more work).

> If we need to know what C++ version we are trying to use, it could be easier to set this at the build system level. Wherever the build system is going to set varying C++ standard flags, throw in a define saying what standard we wanted to use.

Could also just use the macros list in https://en.cppreference.com/w/cpp/experimental/feature_test.html, and define macros that require one or more from those list. Then use those macros to enable/disable whatever engine code that requires them.

slipher: Sure, feature test macros might be viable.

VReaperV: Done now.

src/common/math/Vector.h
```diff
@@ -26,6 +26,16 @@ along with Daemon Source Code.  If not, see <http://www.gnu.org/licenses/>.
 #define COMMON_VECTOR_H_
 
 #include <ostream>
+#include <type_traits>
+
+#include "common/CPPStandard.h"
+
+// HACK: checking for CPP_INVOKE_RESULT doesn't work on MSVC
+#if __cplusplus >= 201703L
```
slipher, line 34: In that case you would need to add back the compiler flag to set `__cplusplus`? But it needs to be set in the build system, not just the CI.

Instead of this maybe it would be easier to just make our own function. Like `template<typename Func, typename... Args> using resultof_t = decltype(std::declval<Func>()(std::declval<Args...>));` and then replace `std::result_of<Func(Type, Type2)>::type` with `resultof_t<Func, Type, Type2>>` etc.

slipher (commented):
Commit message should be updated

src/common/math/Vector.h
```diff
@@ -87,9 +90,9 @@ namespace Math {
 		void Store(Type* ptr) const;
 
 		// Apply a function on all elements of the vector
-		template<typename Func> Vector<Dimension, typename std::result_of<Func(Type)>::type> Apply(Func func) const;
-		template<typename Func, typename Type2> Vector<Dimension, typename std::result_of<Func(Type, Type2)>::type> Apply2(Func func, Vector<Dimension, Type2> other) const;
-		template<typename Func, typename Type2, typename Type3> Vector<Dimension, typename std::result_of<Func(Type, Type2, Type3)>::type> Apply3(Func func, Vector<Dimension, Type2> other1, Vector<Dimension, Type3> other2) const;
+		template<typename Func> Vector<Dimension, typename resultof_t<Func, Type>> Apply(Func func) const;
```
slipher, line 93: Huh, apparently you have to remove the `typename` here. I always believed that spamming more `typename` never hurts!

slipher:
LGTM

Issue #1717, opened 2025-08-01
Desiging and implementing a migration path for porting assets to the linear pipeline


illwieckz:
So, the linear pipeline has been merged:

- https://github.com/DaemonEngine/Daemon/pull/1687

So for now the engine supports:

- rendering the old way maps done the old way (messing with colorspaces like Q3/Tremulous), also called “_the naive pipeline_”.
- rendering the new way maps done the new way (properly handling colorspace conversion at every step of the baking and rendering pipeline), also called “_the linear pipeline_”.

It means maps already built are just rendered like before, but contributors can start creating brand new content targeting the new linear pipeline without having to care about the pitfalls and workaround and weirdness and painful wizardry of the old naive pipeline.

But what we don't care about yet, is about migrating assets from the old naive pipeline to the new linear one.

For example we know existing textures meant to be used in blending operations were wrongly calibrated for the old naive and broken pipeline. When used in the new, correct and linear pipeline, they don't render properly because they were biased to workaround bugs that don't exist anymore.

We may want to provide some kind of conversion or specially crafted workaround for when a blended asset from one pipeline is used on the other.

This would help to solve two problems:

1. Modifying many assets is painful and tedious, if we can just flag either all old assets or old new assets to tell it was meant for a different computation, without actually checking and modifying assets one by one, we could port all of them the blind way just by doing some flagging and letting the engine do the magic.
2. Because we provide reusable texture packs, some textures may be used in maps that will never be ported and also be used in maps that will. So having ways to workaround the usage from one way to another would help. The Unvanquished maps (including community maps), can probably be ported at some point as we either have source or authors are still around, but that's only a hope, we cannot enforce that. Also the problem is stronger for a package like `res-tremulous` where some legacy maps will never be rebuilt (no source, no author around), but some of other legacy maps may be ported to the new pipeline and use those legacy assets.

illwieckz:
Well, now that I posted it, I notice I could have posted it to the Unvanquished issue tracker instead (it's more a project management thread), but 95% of the solutions will likely be some technical engine implementation anyway.

slipher:
I propose adding 2 mechanisms for defining colorspace-dependent shaders:
### 1. Stage-level enablement based on matching rendering mode

Let's say we wanted to adjust the strength of the additive stage here depending on the rendering mode. The `colorspace naive` stage would be disabled when using linear blending and vice versa.

```
textures/shared_ex/light2_red_6000
{
[...]

PR #1721, opened 2025-08-01
Stage-level colorspace keyword for shaders

slipher:
Implement stage-level colorspace keyword as described in https://github.com/DaemonEngine/Daemon/issues/1717#issuecomment-3145706586

Also add a warning if some shaders are loaded too early before the colorspace is known.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -2105,6 +2105,7 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
 	char         buffer[ 1024 ] = "";
 	bool     loadMap = false;
 	bool loadAnimMap = false;
+	bool staticEnabled = true;
 
 	stage->convertColorFromSRGB = convertColorFromSRGB_NOP;
 
@@ -2129,6 +2130,12 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
 		else if ( !Q_stricmp( token, "if" ) )
 		{
 			ParseExpression( text, &stage->ifExp );
+		}
+		else if ( !Q_stricmp( token, "ifStatic" ) )
+		{
+			expression_t expr;
+			ParseExpression( text, &expr );
+			staticEnabled = !!RB_EvalExpression( &expr, 1.0f );
 		}
 		// map <name>
 		else if ( !Q_stricmp( token, "map" ) )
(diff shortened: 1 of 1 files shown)


slipher:
Maybe I should add "if" to the keyword name to make it clearer that it is for conditional stages. Like `ifBlendspace linear`.

illwieckz:
I believe I found an use case for this!

In `tex-trak5`, the `textures/shared_trak5/glass` shader still looks wrong with the `naiveColors` trick:

Naive build, naive pipeline:

unvanquished_2025-10-16_133156_000

Linear build, linear pipeline:

unvanquished_2025-10-16_133327_000

Linear build, linear pipeline, naiveColors trick on the glass:

unvanquished_2025-10-16_133413_000

We need to do some `alpahGen` or something like that to recover the translucency of it, but we only need it with the linear pipeline, and this is a texture set that can be used by legacy maps.

illwieckz:
An example of complex material using `if` from Doom 3 (taken from `materials/lights.mtr`):


```fortran
lights/fanlightgrateSC
{
	{
		forceHighQuality
		if ( global0 == 0 )
		map		lights/fangrate.tga
		colored
		zeroclamp
	}
	{
		forceHighQuality
		if ( global0 == 0 )
		map	lights/fanblade.tga
		colored
		zeroclamp
		rotate	time * -1
	}
	{
		if ( global0 != 0 )
		forceHighQuality
		map		lights/fangrate.tga
		red	( Parm0 * acceleratorflashtable2[ ( time - global1 ) / 7 ] )
		green	( Parm1 * acceleratorflashtable2[ ( time - global1 ) / 7 ] )
		blue	( Parm2 * acceleratorflashtable2[ ( time - global1 ) / 7 ] )
		zeroclamp
	}
	{
		if ( global0 != 0 )
		forceHighQuality
		map	lights/fanblade.tga
		red	( Parm0 * acceleratorflashtable2[ ( time - global1 ) / 7 ] )
		green	( Parm1 * acceleratorflashtable2[ ( time - global1 ) / 7 ] )
		blue	( Parm2 * acceleratorflashtable2[ ( time - global1 ) / 7 ] )
		zeroclamp
		rotate	time * -1
	}
}
```

slipher:
> I just thought that instead of using `if`, that is confusing as it's usually used in other languages to be outside the block, what if the keyword is `when`?

`when` is already a keyword at the shader level, with the function of replacing the entire shader with a different one if the condition matches. It would be confusing if `when` at the stage level had such different semantics from `when` at shader level.

slipher:
Rebased and removed the reverted commit.

PR #1723, opened 2025-08-04
Tests for the macOS CI

illwieckz:
👀


illwieckz:
I've just set both `CFLAGS` and `CXXFLAGS` to `-Wno-c++14-extensions` before calling CMake…

slipher:
I think it would be better to put this in the CI script than the build system

illwieckz:
Well, both work, so we can do as we prefer.

PR #1726, opened 2025-08-04
renderer: disable realtime lighting when ARB_texture_float isn't available

illwieckz:
I hit the `ASSERT( glConfig2.textureFloatAvailable );` in  `R_CreateDepthRenderImage()`, but this function is specific to the realtime lighting code.


illwieckz:
Hmmm, it's not enough. Maybe this code runs before we set the variables.

illwieckz:
Ah no I have to move the assert.

slipher:
LGTM

PR #1728, opened 2025-08-04
sdl_glimp: only read GL_MAX_UNIFORM_BLOCK_SIZE when GL_ARB_uniform_buffer_object is available

illwieckz:
Only read `GL_MAX_UNIFORM_BLOCK_SIZE` when `GL_ARB_uniform_buffer_object is available`.

It is only used by the Material renderer which is not used when such extension isn't available.

On older hardware reading it unconditionally prints `OpenGL error GL_INVALID_ENUM`.

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -2032,9 +2032,10 @@ static void GLimp_InitExtensions()
 		glEnable( GL_DEBUG_OUTPUT_SYNCHRONOUS );
 	}
 
-	// Shader limits
-	glGetIntegerv( GL_MAX_UNIFORM_BLOCK_SIZE, &glConfig2.maxUniformBlockSize );
+	// Shader limits.
+	// From GL_ARB_vertex_shader.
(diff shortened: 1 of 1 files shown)


src/engine/sys/sdl_glimp.cpp
```diff
@@ -2595,6 +2586,24 @@ static void GLimp_InitExtensions()
 	}
 #endif
 
+	// Shader limits.
+
+	// From GL_ARB_vertex_shader.
+	glGetIntegerv( GL_MAX_VERTEX_UNIFORM_COMPONENTS_ARB, &glConfig2.maxVertexUniforms );
+
+	// From GL_ARB_vertex_program.
```
slipher, line 2594: Do we really care to note that it's not available in GL 1?

slipher (commented):
LGTM

src/engine/sys/sdl_glimp.cpp
illwieckz: I care that I know it is available without requiring any extension check.

Because of the other one requiring an extension, I wanted to check that no others were in the same situation and I checked them one by one. I prefer to be explicit.

slipher: Well it would be less confusing to say required in GL 2.0 or whatever.

PR #1730, opened 2025-08-06
tr_shader: linearize fog colors

illwieckz:
Linearize fog colors.

Something I overlooked when writing #1687:

- https://github.com/DaemonEngine/Daemon/pull/1687

This is not enough to make fog looks good in the linear pipeline.

A fog relying on some kind of blend (I guess), we may need to recalibrate/reimplement it for the linear pipeline.

But at least with that patch the colors become right.


illwieckz:
In fact I found how to get the fog working in the linear pipeline, by using a linearized image!

So basically I create two fog images, one for the naive pipeline, one for the linear pipeline, and use them accordingly.

For unknown reasons that doesn't work when I ask to upload the second image with an sRGB format, it's like this metadata is ignored. But it works if I convert the image beforehand. I would prefer to flag the image and not convert it beforehand to not lose precision when storing as bytes, so if someone knows why the flag doesn't work, please help! Otherwise it's ready to use.

illwieckz:
Well, I guess know why the sRGB flag doesn't change anything: only the alpha channel seems to be used and the sRGB conversion doesn't apply on the alpha channel that is always linear (of course!).

illwieckz:
So I stored the alpha channel in the red channel and then we can flag the texture as sRGB to get it converted.

illwieckz:
I added a comment saying the current fog image implementation in linear pipeline is a hack.

But now I thought about something. The fog image generator for the naive pipeline unaware of colorspaces is doing some complex algorithm, picking values from tables and using custom factors… I actually wonder what would happen if we would just use some linear values instead, when running the linear pipeline. Maybe all that fog factor and fog table thing is yet another workaround for the broken naive pipeline?

PR #1733, opened 2025-08-06
Renderer: Avoid a DSA function with VAOs

### LINKED ISSUE #1729: It looks like `glVertexArrayElementBuffer()` is from OpenGL 4.5
We merged #1690 on the assumption VAO is provided by OpenGL 3.0 or by `GL_ARB_vertex_array_object`:

- https://github.com/DaemonEngine/Daemon/pull/1690

But this code also makes use of `glVertexArrayElementBuffer()` and it looks like `glVertexArrayElementBuffer()` is from OpenGL 4.5, at least it's only part of the standard since 4.5:

- https://registry.khronos.org/OpenGL-Refpages/gl4/html/glVertexArrayElementBuffer.xhtml

Also is not listed in `ARB_vertex_array_object`:

- https://registry.khronos.org/OpenGL/extensions/ARB/ARB_vertex_array_object.txt

This results in the `glVertexArrayElementBuffer` pointer being `null` in some drivers for hardware without OpenGL 4.5, leading to an engine crash.

OpenGL 4.5 is a very high version to require. Maybe there can be an alternative implementation of this specific function when OpenGL is older than 4.5, without reverting the whole change?

slipher:
Use the old method to bind the IBO to the VAO instead of glVertexArrayElementBuffer. Fixes #1729.

### DIFF
--- a/src/engine/renderer/GLMemory.h
+++ b/src/engine/renderer/GLMemory.h
@@ -275,11 +275,6 @@ class GLVAO {
 		glVertexArrayElementBuffer( id, buffer.id );
 	}
 
-	// For compatibility with IBO_t
-	void SetIndexBuffer( const GLuint bufferID ) {
-		glVertexArrayElementBuffer( id, bufferID );
-	}
-
 	void GenVAO() {
 		glGenVertexArrays( 1, &id );
 	}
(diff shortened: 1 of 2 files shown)


illwieckz:
Excellent!

Issue #1736, opened 2025-08-08
`munmap_chunk(): invalid pointer` when uploading heat haze maps on old low-end hardware


illwieckz:
I reproduced it on some old low-end Intel and Broadcom hardware, but not on old low-end Nvidia, I haven't tested ATI.

```
Debug: Found CRN image candidate 'gfx/weapons/flamer/haze_n': gfx/weapons/flamer/haze_n.crn 
Debug: Found 128×128 CRN image 'gfx/weapons/flamer/haze_n': gfx/weapons/flamer/haze_n.crn 
Debug: Creating image gfx/weapons/flamer/haze_n (128×128, 8 mips) 
Debug: Allocating image gfx/weapons/flamer/haze_n 
Debug: Uploading image gfx/weapons/flamer/haze_n (128×128, 1 layers, 0xde1 type, 0x83f3 format) 
]munmap_chunk(): invalid pointer
```

There is actually two bugs in action:

1. The heathaze map should not be loaded and uploaded because `r_heatHaze` is disabled when the `lowest` preset is used.
2. Something crashes when uploading such textures.

I know that when I skip the loading of heathaze maps that workaround the bug, so I will first focus on making sure they are not loaded when asked to not be.

Then we may investigate later why the upload fails.

illwieckz:
Here is a PR implementing a workaround by not loading heatHaze maps when the feature is disabled:

- https://github.com/DaemonEngine/Daemon/pull/1737

This doesn't fix the underlying bug though…

slipher:
Closing since the heat haze disabling optimization was merged.

PR #1737, opened 2025-08-08
tr_shader: do not load heatHaze map when heatHaze is disabled

illwieckz:
Do not load heatHaze map when heatHaze is disabled.

This doesn't fix the root cause of #1736, but it workarounds it when heatHaze is disabled:

- htps://github.com/DaemonEngine/Daemon/issues/1736

This is wanted in all cases because it's good to skip loading an uploading stuff we don't use anyway.


slipher:
This spams parse errors and fails to really disable the stage:

unvanquished_2025-08-08_225817_000

Setting `stage->active = false` probably doesn't work there since it gets set to true 2 more times later - inside `ParseStage`, and after it returns as well. I ran into that while writing #1721 (which removes the redundant 2nd time setting it to true).

illwieckz:
Hmm, right. I now disable the stage at the very end of the parsing.

slipher:
`stage->active = false` still doesn't work there since it is set again after ParseStage, as described in my previous comment. This results in messages such as `Warn: Shader textures/plat23_custom/forcefield has a heatHazeMap stage with no image (ignored)`. But I think it would work if combined with my change from #1721 that deletes the final `stage-active = true`.

illwieckz:
> `stage->active = false` still doesn't work there since it is set again after ParseStage

Well, that is a bug unrelated to this.

illwieckz:
I added a patch to not re-enable stages after parsing.

slipher:
LGTM

PR #1593, opened 2025-03-11
Improve screen-space shaders, some minor clean-up

VReaperV:
Requires #1586 

Changes `GLShader` constructors, removing unnecessary stuff from them, and making it only have 2 distinct constructors - for rendering shaders and compute shaders.

Changes the relevant shaders to use a new `screenSpace` shader, which is almost a no-op.

NUKE the the other .glsl shaders that just repeat the same functionality.

Also NUKES the `u_ModelViewProjectionMatrix` from them, because the result is always a full-screen quad anyway.

Adds `Tess_InstantScreenSpaceQuad()` for use with this, because with some other shaders we still do need `u_ModelViewProjectionMatrix`. This actually draws a single triangle that gets clipped to the screen ([this](https://stackoverflow.com/a/59739538) has a good explanation why).

The screen-space shaders in this pr are ones that always process the whole screen, and where the vertex shaders were the same anyway, while sending the `u_ModelViewProjectionMatrix` uniform and multiplying the vertex position with it, and, indeed, even having a VBO for them at all, was useless.

I didn't touch the `screen` shader here, although it really is just another screen-space shader. It is, however, also used for `ST_SCREENMAP` stage, which I don't think we use anywhere, and seems to be intended for use on something like an in-game camera screen. That wouldn't work properly though, due to how the texture coordinates are calculated (relative to the screen position of the fragment, *not* the world/model position). We could probably just use the `generic` shader for that.

Also cleaned-up the formatting and ordering of shaders/shader files.

This brings down the total amount of shader compilation, shader program linking (with #1587), and shader files.

PR #1739, opened 2025-08-08
screenSpace: make it work when GL_EXT_gpu_shader4 is missing

illwieckz:
WIP: screenSpace: make it work when `GL_EXT_gpu_shader4` is missing

It doesn't work yet, I'm not sure to understand how that works.

The purpose is to slightly modify #1593:

- https://github.com/DaemonEngine/Daemon/pull/1593

Commit 5666be1e1c162d5518a3171c220ff629d1bbaa87 from #1593 brings many improvements, but also makes use of `gl_VertexID`, which is introduced by `EXT_gpu_shader4` if I'm right, making it mandatory.

So I tried to add an alternate code path based on specific parts I read from 5666be1e1c162d5518a3171c220ff629d1bbaa87 that replaced the old code (among many other things), but it's not working yet.

It's now probably very close to have it working so if you have any idea, please share. ☺️

### DIFF
--- a/src/engine/renderer/tr_surface.cpp
+++ b/src/engine/renderer/tr_surface.cpp
@@ -531,13 +531,21 @@ void Tess_AddCubeWithNormals( const vec3_t position, const vec3_t minSize, const
 void Tess_InstantScreenSpaceQuad() {
 	GLIMP_LOGCOMMENT( "--- Tess_InstantScreenSpaceQuad ---" );
 
-	tr.skipVBO = true;
-
-	Tess_Begin( Tess_StageIteratorDummy, nullptr, true, -1, 0 );
-	rb_surfaceTable[Util::ordinal( *( tr.genericTriangle->surface ) )]( tr.genericTriangle->surface );
-	Tess_DrawElements();
-
-	tr.skipVBO = false;
+	if ( glConfig2.gpuShader4Available )
+	{
+		tr.skipVBO = true;
+		Tess_Begin( Tess_StageIteratorDummy, nullptr, true, -1, 0 );
+		rb_surfaceTable[Util::ordinal( *( tr.genericTriangle->surface ) )]( tr.genericTriangle->surface );
+		Tess_DrawElements();
+		tr.skipVBO = false;
+	}
+	else
+	{
+		Tess_Begin( Tess_StageIteratorDummy, nullptr, true, -1, 0 );
+		rb_surfaceTable[Util::ordinal( *( tr.genericQuad->surface ) )]( tr.genericQuad->surface );
+		GL_VertexAttribsState( ATTR_POSITION );
+		Tess_DrawElements();
+	}
 
 	GL_CheckErrors();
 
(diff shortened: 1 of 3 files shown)


illwieckz:
This is how it currently looks:

unvanquished_2025-08-08_233356_000

slipher:
I tried it and it rendered correctly only the upper right quarter of the screen. So I tried this and it worked:

```diff
--- a/src/engine/renderer/tr_vbo.cpp
+++ b/src/engine/renderer/tr_vbo.cpp
@@ -555,7 +555,7 @@ void R_BindNullIBO()

 static void R_InitGenericVBOs() {
        // Min and max coordinates of the quad
-       static const vec3_t min = { 0.0f, 0.0f, 0.0f };
+       static const vec3_t min = { -1.0f, -1.f, -1.0f };
        static const vec3_t max = { 1.0f, 1.0f, 0.0f };
        {
                /*
```

I don't know why it would be necessary to use -1 to 1 now if 0 to 1 worked before. Note that this might break other, non-screenspace uses of `genericQuad`.

Your screenshot seems to have additional bugs so I guess there must be more incompatible things if you have the real old hardware.

illwieckz:
> Your screenshot seems to have additional bugs so I guess there must be more incompatible things if you have the real old hardware.

Yes those were driver bugs, after an update they disappeared.

So, when I use both:

- https://github.com/DaemonEngine/Daemon/pull/1733
- https://github.com/DaemonEngine/Daemon/pull/1737
- https://github.com/DaemonEngine/Daemon/pull/1738
- https://github.com/DaemonEngine/Daemon/pull/1739 (this one)

I get this:

unvanquished_2025-08-09_165448_000

The forcefields are a bit dim because that's an sRGB-compiled map running in the linear pipeline, so it's not because of an engine bug or a driver bug on this specific hardware. This produces what we currently expect.

illwieckz:
And when using:

- https://github.com/DaemonEngine/Daemon/pull/1730

I get this (atcshd):

unvanquished_2025-08-09_173755_000

slipher:
PTAL at the branch`slipher/1739-fixup`. There I have changed the projection matrix of `Tess_InstantQuad` so it can produce the correct results while having replaced the [0, 1] square with [-1, 1].

illwieckz:
I don't mind the implementation, if you prefer yours please force-push this PR.

I've tested your implementation and it works as well.

Maybe @VReaperV has an opinion on that too.

illwieckz:
Screenshots with your branch (and all the other patches listed above), but this time with the `low` preset:

unvanquished_2025-08-11_011242_000

unvanquished_2025-08-11_011756_000

illwieckz:
I forgot to mention but outside of actually making the engine works on hardware without `GL_EXT_gpu_shader4`, this would also make it consistent because we actually disable `GL_EXT_gpu_shader4` on GLSL 1.20 because of having a different syntax on some stuff like `unsigned int` and `uint`, as discussed here:

- https://gitlab.freedesktop.org/mesa/mesa/-/issues/12803#note_2819461

So, without this branch, when this extension is available, on some hardware we disable some `GL_EXT_gpu_shader4` code while still using the extension, which is a bit confusing.

slipher:
You have pushed the branch again and it no longer contains my changes from `slipher/1739-fixup`.

illwieckz:
> You have pushed the branch again and it no longer contains my changes from `slipher/1739-fixup`.

Oupsie, that should now be fixed.

I also added comments linking to this PR so future selves may know why we picked those numbers.

src/engine/renderer/tr_surface.cpp
```diff
@@ -550,12 +558,15 @@ void Tess_InstantQuad( u_ModelViewProjectionMatrix &shader, const float x, const
 
 	Tess_Begin( Tess_StageIteratorDummy, nullptr, true, -1, 0 );
 
+	/* These values make possible to use either the legacy quad rendering
```
slipher, line 561: The point of this function is to render any screen-space axis-aligned rectangle specified by `x, y, width, height` so the comment seems out of place.

src/engine/renderer/tr_vbo.cpp
```diff
@@ -554,24 +554,19 @@ void R_BindNullIBO()
 }
 
 static void R_InitGenericVBOs() {
+	/* These values make possible to use either the legacy quad rendering
```
slipher, line 557: To explain the values chosen here, you could say that they are chosen to specify a full-screen rectangle for screen-space shaders without a projection matrix. The values here don't influence the triangle but it's good to mention that this is a fallback when the `gl_VertexID`-based triangle can't be used.

slipher:
LGTM

Issue #314, opened 2020-04-12
Translucent texture rendered above fog


illwieckz:
First seen in https://github.com/Unvanquished/Unvanquished/issues/799

See the ice stalactites on the gauntry (in background):

[![translucent on fog](https://dl.illwieckz.net/b/daemon/bugs/translucent-on-fog/unvanquished_2020-04-12_045607_000.jpg)](https://dl.illwieckz.net/b/daemon/bugs/translucent-on-fog/unvanquished_2020-04-12_045607_000.jpg)

VReaperV:
This might be some incorrectly set shader sort since we do depth->opaque->fog->translucent. Or maybe we just need another pass before fog and after opaque, since stalactites probably are translucent.

VReaperV:
I decided to check if it's still true...

![image](https://github.com/user-attachments/assets/5afcf7d0-a6e2-407c-ae9f-a4043c33be0a)

On the bright side, it's no longer rendered above fog.

VReaperV:
Confirming the original bug is still there.

VReaperV:
> This might be some incorrectly set shader sort since we do depth->opaque->fog->translucent. Or maybe we just need another pass before fog and after opaque, since stalactites probably are translucent.

So the stalactites are, in fact, translucent.

VReaperV:
The `implicitMask` is indeed not working, because the alpha threshold is being set to 1.5, i.e the default value.

VReaperV:
Looks like the shader gets sorted as opaque and I'm not seeing it being used when stepping through the `RB_RenderDrawSurfaces()` code for translucents, but I see it being rendered after the global fog in a frame debugger...

VReaperV:
They also don't write to depth, even though they should.

VReaperV:
Oh wait, I was looking at the wrong shader lol.

VReaperV:
So this is the shader in question:

```
surfaceparm alphashadow
cull none
{
    map textures/snow/alpha_ice2s.tga
    blendFunc GL_SRC_ALPHA GL_ONE_MINUS_SRC_ALPHA
    rgbGen vertex
}
```

It might be that `alphashadow` is supposed to turn alpha testing on here, but it doesn't.

VReaperV:
[...]

PR #1740, opened 2025-08-11
Don't draw global fog twice

slipher:
Stop drawing global fog a second time. Global fog is used in a lot of ET maps but not Tremulous ones. The unwanted 2nd go-around was using the volumetric fog shader (`fogQuake3`) instead of the `fogGlobal` that it is supposed to use.

Without fog (`r_noFog 1`):
unvanquished_2025-08-11_063035_000
With global rendered just once with the correct shader: (`r_wolfFog 0` on master or default with this branch)
unvanquished_2025-08-11_063047_000
With fog (badly) rendered a second time, as default on master:
unvanquished_2025-08-11_063054_000


illwieckz:
I confirm it fixes:

- https://github.com/DaemonEngine/Daemon/issues/314

So, yes we had a bug, but not the one we were looking for.

PR #1743, opened 2025-08-11
Minor gl_shader fixes

VReaperV:
This was caught by CI in #1741

Fixes `ShaderProgramDescriptor` definition order since `GLShader` requires the complete type, not sure how it was working previously.

Also fixed a missed `std::is_pod` by converting it to `IsPod`.


illwieckz:
> not sure how it was working previously.

Because we weren't looking at it yet. 🫣

illwieckz (approved):
LGTM.

PR #1744, opened 2025-08-11
Don't log bad uvs for white/black lightmaps

VReaperV:
Fixes log spam on some maps, like `thunder` and `perseus`.

We may also consider replacing [some of] the textures from `shared_colors_src` with the corresponding internal images in engine.

### DIFF
--- a/src/engine/renderer/tr_bsp.cpp
+++ b/src/engine/renderer/tr_bsp.cpp
@@ -26,6 +26,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "framework/CommandSystem.h"
 #include "GeometryCache.h"
 #include "GeometryOptimiser.h"
+#include "ShadeCommon.h"
 
 /*
 ========================================================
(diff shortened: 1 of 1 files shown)


illwieckz:
I'm not sure this is a good solution, many maps redefined those shaders.

Sometime some used builtin textures (even the funny way):

```
// black, no lightmap
textures/nano/black
{
	qer_editorimage textures/nano/black
	surfaceparm nomarks
	surfaceparm nolightmap
	{
		map $whiteimage
		rgbGen const ( 0 0 0 )
	}
}
```

```
// black, no lightmap
textures/methane/black
{
	qer_editorimage textures/methane/black
	surfaceparm nomarks
	surfaceparm nolightmap
	{
		map $whiteimage
		rgbGen const ( 0 0 0 )
	}
}
```

But others shipped their own textures:

```
// no lightmap, always black
textures/atcshd/black
{
	qer_editorimage textures/atcshd/black
	surfaceparm nomarks
	surfaceparm nolightmap
	{
		map textures/atcshd/black
	}
}
```

We may better rely on the `nolightmap` surfaceparm insead? We should not look for lightmaps on those surfaces to begin with, the shader explicitly says there is should be no lightmap to be computed (and then, to be available).

illwieckz:
> We may also consider replacing [some of] the textures from shared_colors_src with the corresponding internal images in engine.

We should do this yes as it reduces loading time (less I/O and image parsing and uploading), but that should not be required to prevent those false-positive logs to appear.

illwieckz:
Try with: `hasExplicitelyDisabledLightMap( shader )`

illwieckz:
It's from `ShadeCommon.h`, what it does is to test for `shader->surfaceFlags & SURF_NOLIGHTMAP`.

PR #1748, opened 2025-08-12
Fix autosprite log spam

VReaperV:
Cgame-side pr: https://github.com/Unvanquished/Unvanquished/pull/3400

Fixes performance regression on map `assault` caused by log spam from 6b7a7dd23fd414eda97c39f71a754d4c43d02d93

### DIFF
--- a/src/engine/renderer/tr_surface.cpp
+++ b/src/engine/renderer/tr_surface.cpp
@@ -577,6 +577,8 @@ Tess_SurfaceSprite
 */
 static const float NORMAL_EPSILON = 0.0001;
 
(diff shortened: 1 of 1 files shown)


VReaperV:
> Why not use the actual shader as the basis for uniqueness instead of trying to plumb through some meaningless number from cgame? That would be easier and would automatically work for other sprite entities such as particles.

Right, I'll try that later.

VReaperV:
> Why not use the actual shader as the basis for uniqueness instead of trying to plumb through some meaningless number from cgame? That would be easier and would automatically work for other sprite entities such as particles.

Done in #1750.

PR #1750, opened 2025-08-12
Fix log spam due to bad autosprite configuration

VReaperV:
Fixes performance regression on map assault caused by log spam from https://github.com/DaemonEngine/Daemon/commit/6b7a7dd23fd414eda97c39f71a754d4c43d02d93

### DIFF
--- a/src/engine/renderer/tr_surface.cpp
+++ b/src/engine/renderer/tr_surface.cpp
@@ -586,12 +586,13 @@ static void Tess_SurfaceSprite()
 
 	radius = backEnd.currentEntity->e.radius;
 
-	if ( tess.surfaceShader->autoSpriteMode != 0 )
+	if ( tess.surfaceShader->autoSpriteMode != 0 && !tess.surfaceShader->autoSpriteWarned )
 	{
 		// This function does similarly to autosprite mode 1. Autospriting it again would be a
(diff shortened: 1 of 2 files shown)


src/engine/renderer/tr_surface.cpp
```diff
@@ -586,12 +586,13 @@ static void Tess_SurfaceSprite()
...
 	radius = backEnd.currentEntity->e.radius;
 
-	if ( tess.surfaceShader->autoSpriteMode != 0 )
+	if ( tess.surfaceShader->autoSpriteMode != 0 && !tess.surfaceShader->autoSpriteWarned )
 	{
 		// This function does similarly to autosprite mode 1. Autospriting it again would be a
 		// waste and would probably lose the rotation angle
-		Log::Warn( "RT_SPRITE entity should NOT configure its shader (%s) as autosprite",
-		           tess.surfaceShader->name );
+		Log::Warn( "RT_SPRITE entity should NOT configure its shader (%s) as autosprite, shader index: %i",
```
slipher, line 593: "shader index" should be removed

VReaperV: Done.

slipher:
LGTM

PR #1752, opened 2025-08-13
Switch to SDL3

VReaperV:
SDL3 seems to work better on Wayland. Some systems don't even actually use SDL2 anymore but rather a compatibility layer.

This will also require someone to update the deps packages.


illwieckz:
Very good! Can you push it to a branch in the target repository instead so we can test it with the docker-based release script?

illwieckz:
I need it to be part of the deps update branch anyway.

illwieckz:
I imported that code in #1433:

And added more SDL3 code to make it work on my end.

VReaperV:
Thx. Once it works there this pr can be closed.

src/engine/client/key_identification.cpp
```diff
@@ -215,7 +215,7 @@ std::string KeyToString(Key key)
 }
 
 int GetCharForScancode(int scancode) {
-    int keycode = static_cast<int>(SDL_GetKeyFromScancode(Util::enum_cast<SDL_Scancode>(scancode)));
+    int keycode = static_cast<int>(SDL_GetKeyFromScancode(Util::enum_cast<SDL_Scancode>(scancode), NULL, false));
```
slipher, line 218: The second arg is an integer type, should be 0 not NULL

illwieckz:
The SDL3 port was completed and merged as part of #1433:

Thanks! 😊

PR #1753, opened 2025-08-13
NUKE glconfig_t, use WindowConfig instead

### LINKED ISSUE #517: modifying glconfig_t breaks engine/gamecode compatibility
While working on #478 I noticed the string we use to store the list of available OpenGL extensions is smaller than the list of available OpenGL extensions itself provided by actual drivers, so I just tried to double the size this way:

```diff
diff --git a/src/engine/renderer/tr_types.h b/src/engine/renderer/tr_types.h
index 7c4aeb672..662ae5742 100644
--- a/src/engine/renderer/tr_types.h
+++ b/src/engine/renderer/tr_types.h
@@ -335,7 +335,7 @@ struct glconfig_t
        char                 renderer_string[ MAX_STRING_CHARS ];
        char                 vendor_string[ MAX_STRING_CHARS ];
        char                 version_string[ MAX_STRING_CHARS ];
-       char                 extensions_string[ MAX_STRING_CHARS * 4 ]; // TTimo - bumping, some cards have a big extension string
+       char                 extensions_string[ MAX_STRING_CHARS * 8 ]; // TTimo - bumping, some cards have a big extension string
 
        int                  maxTextureSize; // queried from GL
        int                  unused;
```

And then when trying to load the engine from that branch with released game code I got this error:

```
Loaded VM module in 124 msec 
Warn: Error during initialization: CGame VM: Reader: Unread bytes at end of message 
```

Why modifying that string (and then, that struct) breaks engine/game compat? Do we need that?

VReaperV:
Cgame-side pr: https://github.com/Unvanquished/Unvanquished/pull/3405

Fixes #517

Move the stuff actually used by cgame (window and display sizes and display aspect) to `WindowConfig`. Merge the rest with `glconfig2_t` into `GLConfig`.

I've noticed this issue earlier when adding things to `glconfig2` and when working on daemon-vulkan: `RE_BeginRegistration()` uses `glconfig_t* glConfigOut, glconfig2_t glConfig2Out` as its parameters, even though it only needs the window and display sizes from the first one and nothing else.

### DIFF
--- a/src/engine/client/cg_msgdef.h
+++ b/src/engine/client/cg_msgdef.h
@@ -510,7 +510,7 @@ using CGameStaticInitMsg = IPC::SyncMessage<
 	IPC::Message<IPC::Id<VM::QVM, CG_STATIC_INIT>, int>
 >;
 using CGameInitMsg = IPC::SyncMessage<
-	IPC::Message<IPC::Id<VM::QVM, CG_INIT>, int, int, glconfig_t, GameStateCSs>
+	IPC::Message<IPC::Id<VM::QVM, CG_INIT>, int, int, WindowConfig, GameStateCSs>
 >;
 using CGameShutdownMsg = IPC::SyncMessage<
 	IPC::Message<IPC::Id<VM::QVM, CG_SHUTDOWN>>
@@ -540,7 +540,7 @@ using CGameFocusEventMsg = IPC::SyncMessage<
 
 //TODO Check all rocket calls
 using CGameRocketInitMsg = IPC::SyncMessage<
-	IPC::Message<IPC::Id<VM::QVM, CG_ROCKET_VM_INIT>, glconfig_t>
+	IPC::Message<IPC::Id<VM::QVM, CG_ROCKET_VM_INIT>, WindowConfig>
 >;
 using CGameRocketFrameMsg = IPC::SyncMessage<
 	IPC::Message<IPC::Id<VM::QVM, CG_ROCKET_FRAME>, cgClientState_t>
--- a/src/engine/client/cl_cgame.cpp
+++ b/src/engine/client/cl_cgame.cpp
@@ -956,15 +956,7 @@ void CGameVM::CGameStaticInit()
 
 void CGameVM::CGameInit(int serverMessageNum, int clientNum)
 {
-	glconfig_t glConfig;
-	memset( &glConfig, 0, sizeof( glconfig_t ) );
-	glConfig.displayAspect = cls.windowConfig.displayAspect;
-	glConfig.displayWidth = cls.windowConfig.displayWidth;
-	glConfig.displayHeight = cls.windowConfig.displayHeight;
-	glConfig.vidWidth = cls.windowConfig.vidWidth;
-	glConfig.vidHeight = cls.windowConfig.vidHeight;
-
-	this->SendMsg<CGameInitMsg>(serverMessageNum, clientNum, glConfig, cl.gameState);
+	this->SendMsg<CGameInitMsg>(serverMessageNum, clientNum, cls.windowConfig, cl.gameState);
 	NetcodeTable psTable;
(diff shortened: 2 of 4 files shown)


VReaperV:
Also, none of the features mentioned are in the `glconfig_t` struct anyway.

VReaperV:
In fact, we can have an std::vector of the supported rendering features in both cgame and engine. Each element would be something like: 
```
struct EngineFeature {
  uint32 version;
  bool8_t enabled;
  bool8_t required;
  bool8_t missingHWRequirements;
  std::vector<uint32> missingFeatureRequirements;
  std::string name;
};
```
and there would be an enum corresponding to each one (to keep strings translatable in the graphics settings).

Then, if cgame receives a vector with more features than it's aware of, it would simply ignore the extra ones and perhaps print their names into the log. If the engine sends a vector with less features, the extra ones would be implicitly disabled, perhaps with some log print about those features requiring a newer engine version.

All of that would allow adding support for optional rendering features that require adding extra codepaths in cgame. For example, GPU particles could be an option that, if detected to be enabled in the engine, would be used by an up-to-date cgame, or otherwise just ignored.

VReaperV:
Or just keep it as a fixed array. That would be simpler, but would require a major engine release to add anything new to it.

slipher:
Having a vector instead of the fixed-size struct sounds like a good idea. I guess `missingFeatureRequirements` would mostly be GL extensions? Maybe it would be easier to use strings instead of an enum as it seems tedious to make an enum for all the GL extensions. Also the implementation details of a feature may change, so it would be good to be able to express missing features that weren't planned for in the ABI.

src/common/Serialize.h
```diff
@@ -244,6 +244,18 @@ namespace Util {
 		}
 	};
 
+	template<typename T>
+	struct SerializeTraits<const T&, typename std::enable_if<IsPod<const T&> && !std::is_array<const T&>::value>::type> {
```
slipher, line 248: What's the point of this change?

src/engine/client/cg_msgdef.h
```diff
@@ -120,6 +120,18 @@ namespace Util {
 			return value;
 		}
 	};
+
+	template<> struct SerializeTraits<const WindowConfig&> {
+		static void Write( Writer& stream, const WindowConfig& windowConfig ) {
+			stream.WriteData( &windowConfig, sizeof( WindowConfig ) );
+		}
+
+		static WindowConfig Read( Reader& stream ) {
```
slipher, line 129: We shouldn't need this, memcpying is the default behavior. The other IPC change is still there too.

slipher (commented):
The general idea seems fine, but I'm afraid that with the way the PR is structured now, the for-0.56 branch will be constantly broken, since each time a line of code referencing `glConfig` or `glConfig2` is edited on master, it will fail to merge or fail to compile once merged. So it would be better to arrange the changes so that we do the large-scale renaming on master now. For example you could remove `glconfig_t` from the renderer and only build it for IPC, while adding and using `WindowConfig` in the renderer. Or you could rename `glconfig_t` to `WindowConfig` right now in the engine and cgame, and just remove the extra fields on 0.56.

src/engine/client/cg_msgdef.h
VReaperV: It fails to find a matching overload for `Util::apply()` if I remove that.

VReaperV: Apparently so. Would've been nice if that was actually conveyed somewhere (somewhere not being cryptic template errors).

VReaperV:
CI failure is unrelated:
> /home/vsts/work/1/s/src/common/Command.cpp:588:20: error: unused variable 'False' [-Werror,-Wunused-variable]
> 588 |             volatile bool False = false;

slipher:
LGTM

PR #1754, opened 2025-08-13
Fix looping sounds

VReaperV:
Fixes the magical quantum looping sound emitters that change their state when you stop observing them.

The effect can be very easily heard on the map `spacetracks` where the elevator music restarts whenever the speaker goes out of view without this fix.

Also removed a layer of useless functions in the surrounding code.

### DIFF
--- a/src/engine/audio/Audio.cpp
+++ b/src/engine/audio/Audio.cpp
@@ -194,10 +194,10 @@ namespace Audio {
         }
 
         for (int i = 0; i < MAX_GENTITIES; i++) {
-            auto& loop = entityLoops[i];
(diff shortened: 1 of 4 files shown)


VReaperV:
> For me this branch seems to break all non-looping sounds. For example firing a rifle makes no sound.

Fixed.

VReaperV:
> Anyway would this means we are perpetually mixing a bunch of dead sounds with 0 volume? Seems like it could cause performance problems

I've found a fix for this as well. I'll open a new pr later to use instead of this one, since it requires compat-breaking changes.

PR #1755, opened 2025-08-15
Some commits cherry-picked from the external_deps work

slipher:
Stuff that's good independent of dep upgrades and doesn't conflict with anything else.


illwieckz (approved):
LGTM.

slipher:
I will disregard the 24-hour rule so that rebasing this stuff 10000 times isn't any more painful than necessary.

PR #1756, opened 2025-08-15
Fix looping sounds

VReaperV:
Cgame-side pr: 

Fixes the magical quantum looping sound emitters that change their state when you stop observing them.

The effect can be very easily heard on the map spacetracks where the elevator music restarts whenever the speaker goes out of view without this fix.

Also removed a layer of useless functions in the surrounding code.

Added an `SVF_BROADCAST_ONCE` so the speakers aren't magically paused until they come into view (this also fixes the elevator music on the different floors of `spacetracks` being desynced).

### DIFF
--- a/src/engine/audio/Audio.cpp
+++ b/src/engine/audio/Audio.cpp
@@ -194,10 +194,10 @@ namespace Audio {
         }
 
         for (int i = 0; i < MAX_GENTITIES; i++) {
-            auto& loop = entityLoops[i];
+            entityLoop_t& loop = entityLoops[i];
             if (loop.sound and not loop.addedThisFrame) {
                 // The loop wasn't added this frame, that means it has to be removed.
-                loop.sound->SetSoundGain( 0 );
+                loop.sound->soundGain = 0;
             } else if (loop.oldSfx != loop.newSfx) {
                 // The last sfx added in the frame is not the current one being played
                 // To mimic the previous sound system's behavior we sart playing the new one.
@@ -341,7 +341,7 @@ namespace Audio {
 
         StopMusic();
         music = std::make_shared<LoopingSound>(loopingSample, leadingSample);
-        music->SetVolumeModifier(musicVolume);
+        music->volumeModifier = &musicVolume;
         AddSound(GetLocalEmitter(), music, 1);
     }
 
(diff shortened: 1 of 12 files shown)


slipher:
Does OpenAL provide some sort of guarantee that if you set the volume to 0, you won't spend the processing power on mixing it?

slipher:
Isn't that how the PR works? `loop.sound->SetSoundGain( 0 );` instead of really stopping it.

slipher:
I think there could be a lot of persistent sounds. For example in the map Jota there are 56 `sfx_speaker`s. Or in some PVE layout you could have hundreds of turrets with the looping ringing noises.

VReaperV:
> I think there could be a lot of persistent sounds. For example in the map Jota there are 56 `sfx_speaker`s. Or in some PVE layout you could have hundreds of turrets with the looping ringing noises.

Sounds like a problem. Turrets don't create persistent sounds.

slipher:
Yeah they do. It looks like this in `CG_RunBuildableLerpFrame`

```cpp
if ( cg_buildables[ buildable ].sounds[ lf->animationNumber ].looped &&
     cg_buildables[ buildable ].sounds[ lf->animationNumber ].enabled )
{
	trap_S_AddLoopingSound( cent->currentState.number, cent->lerpOrigin, vec3_origin,
	                        cg_buildables[ buildable ].sounds[ lf->animationNumber ].sound );
}
```

VReaperV:
> Yeah they do. It looks like this in `CG_RunBuildableLerpFrame`
> 
> ```c++
> if ( cg_buildables[ buildable ].sounds[ lf->animationNumber ].looped &&
>      cg_buildables[ buildable ].sounds[ lf->animationNumber ].enabled )
> {
> 	trap_S_AddLoopingSound( cent->currentState.number, cent->lerpOrigin, vec3_origin,
> 	                        cg_buildables[ buildable ].sounds[ lf->animationNumber ].sound );
> }
> ```

That's not a persistent sound.

slipher:
Maybe an alternative could be when a sfx_speaker sound is created, to set the `AL_SEC_OFFSET` parameter on the sound, which seeks to a particular time value in seconds. We can set it to some value based on the engine's realtime clock.  That way we wouldn't need to keep sounds alive when not playing, nor broadcast the entities all at once. I see some modulus calculations in the ALSoft source so I think it should handle looping like we want but I'm not sure.

I was going to suggest that the cgame API for starting a sound could have a time offset, but I realized it's not actually possible to implement the sound syncing that way since the cgame doesn't have any shared clock reference with the engine. If you tried to use the server time or whatever it wouldn't work with timescale changes.

VReaperV:
> CPU usage would ideally be the thing to check since OpenAL does its work on a background thread.

Not even filling up a single core with work. And the CPU usage is below master due to the fix I did for particles.

PR #1757, opened 2025-08-15
WIP: Allow multiple sounds per entity

VReaperV:
Fixes https://github.com/Unvanquished/Unvanquished/issues/579

For some reason the luci warning sound has the wrong volume, so this is draft for now (also can be rebased on master), but this does work with both jet and luci sounds at the same time.


VReaperV:
Turns out this wasn't a bug with audio code, but rather a bug with adding lcannon sounds.

PR #1760, opened 2025-08-15
Implement some sound priorities

VReaperV:
Requires https://github.com/DaemonEngine/Daemon/pull/1758

Cgame-side pr: https://github.com/Unvanquished/Unvanquished/pull/3408

Use the sound's volume and distance to determine which sounds to replace. Player/bot sounds get a higher priority within `a_clientSoundPriorityMaxDistance` distance, with the multiplier `a_clientSoundPriorityMultiplier`.

This is helpful when there are a lot of sound-emitting entities around (like buildings) which take priority over player/bot sounds. The latter are generally more important.

### DIFF
--- a/src/engine/audio/Audio.cpp
+++ b/src/engine/audio/Audio.cpp
@@ -120,6 +120,8 @@ namespace Audio {
 
     static bool initialized = false;
 
+    int playerClientNum;
+
     static AL::Device* device;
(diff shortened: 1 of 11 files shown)


VReaperV:
I've no idea why CI failed when its build succeeded...

PR #1763, opened 2025-08-20
Fix light tile bug + clean up texture binding 

### LINKED ISSUE #1734: Bad depth test for dynamic lights
Some dynamic lights are filtered out in places they should be rendered. The bug is most glaring when using the Painsaw and seeing the artifacts on the blade. It can happen on world surfaces too: see the dark patches on the side of the top stair step in Parpax.

Image

Image

The depth tile shader is supposed to compute the minimum and maximum depth buffer for each rectangular subdivision of the screen. It seems to work OK on ones that all have about the same depth, but is sometimes wrong for ones that have a range of depths. For the painsaw case, it is the `plane5` test in `lighttile_fp.glsl` that is incorrectly excluding lights. I believe this is the "far plane" for the tile.

The bug only happens when bindless textures are not being used. That is, setting `r_preferBindlessTextures` 1 makes the bug go away. The behavior is identical between my AMD and NVidia cards, so don't say it's a driver bug :)

slipher:
Fixes #1734.


src/engine/renderer/gl_shader.cpp
```diff
@@ -2775,7 +2771,7 @@ GLShader_portal::GLShader_portal() :
 
 void GLShader_portal::SetShaderProgramUniforms( ShaderProgramDescriptor *shaderProgram )
 {
-	glUniform1i( glGetUniformLocation( shaderProgram->id, "u_CurrentMap" ), 0 );
+	glUniform1i( glGetUniformLocation( shaderProgram->id, "u_ColorMap" ), 0 );
```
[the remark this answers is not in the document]
slipher: Fixed.

src/engine/renderer/gl_shader.cpp
```diff
@@ -2964,14 +2964,9 @@ GLShader_depthReduction::GLShader_depthReduction() :
 		false, "depthReduction" ),
 	u_ViewWidth( this ),
 	u_ViewHeight( this ),
-	u_DepthMap( this ),
```
VReaperV, line 3034: #1684 would require keeping the `u_DepthMap` in this shader.

slipher: It adds a thing to set u_DepthMap but also there is still a texture unit binding in the GLSL and `MaterialSystem::DepthReduction` binds it to a specific TMU? I guess some stuff needs to be cleaned up there

VReaperV: Yeah, the explicit binding in `MaterialSystem::DepthReduction()` could then be removed.

VReaperV (approved):

PR #1764, opened 2025-08-21
Split glconfig_t/glconfig2_t into WindowConfig/GLConfig (engine only)

VReaperV:
Cherry-picked from #1753 to avoid merge issues.

### DIFF
--- a/src/engine/client/cl_cgame.cpp
+++ b/src/engine/client/cl_cgame.cpp
@@ -974,7 +974,15 @@ void CGameVM::CGameStaticInit()
 
 void CGameVM::CGameInit(int serverMessageNum, int clientNum)
 {
-	this->SendMsg<CGameInitMsg>(serverMessageNum, clientNum, cls.glconfig, cl.gameState);
+	glconfig_t glConfig;
+	memset( &glConfig, 0, sizeof( glconfig_t ) );
+	glConfig.displayAspect = cls.windowConfig.displayAspect;
+	glConfig.displayWidth = cls.windowConfig.displayWidth;
+	glConfig.displayHeight = cls.windowConfig.displayHeight;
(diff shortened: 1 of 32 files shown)


src/engine/null/null_renderer.cpp
```diff
@@ -180,12 +180,11 @@ const char *RE_ShaderNameFromHandle( qhandle_t )
...
 
-bool RE_BeginRegistration( glconfig_t *config, glconfig2_t *glconfig2 )
+bool RE_BeginRegistration( WindowConfig* windowCfg )
 {
-	*config = glconfig_t{};
-	config->vidWidth = 640;
-	config->vidHeight = 480;
-	*glconfig2 = glconfig2_t{};
+	*windowCfg = WindowConfig {};
+    windowCfg->vidWidth = 640;
```
slipher, line 186: Bad whitespace

src/engine/client/cl_cgame.cpp
```diff
@@ -1038,7 +1045,14 @@ void CGameVM::CGameTextInputEvent(int c)
 
 void CGameVM::CGameRocketInit()
 {
-	this->SendMsg<CGameRocketInitMsg>(cls.glconfig);
+	glconfig_t glConfig;
+	glConfig.displayAspect = cls.windowConfig.displayAspect;
```
slipher, line 1049: If we're not going to populate all the fields of the struct, it would be good to at least memset it, so for example if the cgame tried to read `vendor_string` it wouldn't crash.

VReaperV: True, done.

src/engine/null/null_renderer.cpp
VReaperV: Fixed.

PR #1767, opened 2025-08-22
report game version in userinfo

cu-kai:
I believe it would be beneficial to report the game version in the userinfo, this will help server admins troubleshoot issues when players are reporting problems with the game. There will be a separate PR for the `/versions` command in `sg_admin.cpp` which will allow admins to view this information.

Some Tremulous clients had this feature, I am simply bringing that here (notably Tremfusion did this).

I also believe we should be reporting the build short commit hash in the version string, which will allow server admins to understand which exact build a player is using, but I believe that should be done in another PR. 

CC: @illwieckz and @VReaperV


VReaperV (approved):
This would've been useful for some issues we've already had in unvanquished too.

PR #1768, opened 2025-08-23
Improve R_RemapShader behavior

slipher:


VReaperV:
Should this really be targeting master branch if a map relies on this bug?

Issue #1696, opened 2025-07-12
Relief mapping doesn't work when the material system is used


illwieckz:
Relief mapping doesn't work when the material system is used, the surfaces remain flat.

VReaperV:
Possibly related: #1616.

illwieckz:
In fact it doesn't work 99% of the time, but it partially works.

For exampe when loading the `parpax` map, all surfaces of the static world are flat, but the surfaces on the elevator get their relief.

![Image](https://github.com/user-attachments/assets/d7057122-503c-40a2-8c2c-907460d922e6)

![Image](https://github.com/user-attachments/assets/e47019fe-ce2b-45d3-9f15-1b67719b6411)

VReaperV:
Are those valid values? I see that `r_reliefDepthScale` is 0.02 by default.

VReaperV:
Maybe related: #1672.

slipher:
> Are those valid values? I see that `r_reliefDepthScale` is 0.02 by default.

Maybe it was not a reasonable value, but for me the important thing was `r_materialSystem 0; r_reliefMapping 1` produced roughly the same images as `r_materialSystem 1; r_reliefMapping 1` and these images were very different from ones produced by `r_materialSystem 1; r_reliefMapping 0`.

VReaperV:
Oh, it's because the uniforms use `GLUniform1f`, which is `global` by default. I think it's best to get #1685 merged as the fix, as it eliminates this issue.

VReaperV:
> In fact it doesn't work 99% of the time, but it partially works.
> 
> For exampe when loading the `parpax` map, all surfaces of the static world are flat, but the surfaces on the elevators get their relief.
> 
> ![Image](https://github.com/user-attachments/assets/d7057122-503c-40a2-8c2c-907460d922e6)
> 
> ![Image](https://github.com/user-attachments/assets/e47019fe-ce2b-45d3-9f15-1b67719b6411)

It works on the elevator because it is an entity, which are not rendered by the material system.

slipher:
Just tried #1685 and it's still broken in the same way.

slipher:
I suspect a struct packing bug. I tried changing the two `u_Relief*` uniforms to vec4 and ... all lightmapped map surfaces stopped being rendered.

```diff
diff --git a/src/engine/renderer/gl_shader.h b/src/engine/renderer/gl_shader.h
index ffaefc504..f17c29d93 100644
[...]

PR #1777, opened 2025-08-29
Fix material system uniform SSBO struct packing

slipher:
Fix the code for laying out the std430 struct containing the uniforms for material system GLSL shaders. There is no functional change given the exact sets of uniforms we have on our GLSL shaders right now, but adding a new one could cause things to blow up as demonstrated at https://github.com/DaemonEngine/Daemon/issues/1696#issuecomment-3236089118.

I'm not attempting to run the added unit tests on CI for now as the client test does full client initialization including opening a window. So it wouldn't easily run in a headless environment.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -2092,14 +2092,14 @@ GLint GLShader::GetUniformLocation( const GLchar *uniformName ) const {
 	return glGetUniformLocation( p->id, uniformName );
 }
 
-static int FindUniformForAlignment( std::vector<GLUniform*>& uniforms, const GLuint alignment ) {
-	for ( uint32_t i = 0; i < uniforms.size(); i++ ) {
-		if ( uniforms[i]->_std430Size <= alignment ) {
-			return i;
+static auto FindUniformForOffset( std::vector<GLUniform*>& uniforms, const GLuint baseOffset ) {
+	for ( auto it = uniforms.begin(); it != uniforms.end(); ++it ) {
+		if ( 0 == ( ( (*it)->_std430Alignment - 1 ) & baseOffset ) ) {
+			return it;
 		}
 	}
 
-	return -1;
+	return uniforms.end();
 }
 
 // Compute std430 size/alignment and sort uniforms from highest to lowest alignment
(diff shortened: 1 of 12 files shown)


VReaperV:
> Remove padding from end of material uniform struct
> Explicitly writing the padding seems pointless and the way
the padding is calculated looks suspicious.

The point is that I can copy it from shader source to buffer view in Nsight and immediately see the values read by the shader.

slipher:
> The point is that I can copy it from shader source to buffer view in Nsight and immediately see the values read by the shader.

I don't get it, why would adding dummy fields at the end make this more effective?

slipher:
OK I found that in the documentation:
> Alignments are not considered; all types are considered packed.

Does the latest commit work? I would try it but the last time I tried to update the software, it broke - it can no longer establish the socket connection.

The original code seemed strange since it could only add padding at the end. I assume you need to add it anywhere else there are holes too? That's what I did.

VReaperV:
Thx, I'll try that in a bit.

Yeah, the original code might be wrong on that.

VReaperV:
> Does the latest commit work? I would try it but the last time I tried to update the software, it broke - it can no longer establish the socket connection.

That does work, yes.

The change in struct packing does, however, break when the material uniforms add up to less than 4 components/16 bytes. In that case the size is calculated as if they're tightly packed, when in the actual buffer interface they're rounded up to a `vec4`. The `std430Size` naming is misleading at this point, since the only buffer interfaces that use the generated structs are now using std140. The related variables functions can be renamed or a boolean flag added to round up the alignment to `vec4`. Personally though, I don't have any plans to add buffers with generated struct arrays and std430 interfaces.

Also, I believe the implementation in this pr doesn't calculate the size for arrays correctly as the component count is never taken into account.

DolceTriade:
Does this fix the edge case where if you add a uniform that is 4 byte aligned, it messes up the alignment of the struct and causes view corruption? I hit that in some of my testing..

slipher:
> Does this fix the edge case where if you add a uniform that is 4 byte aligned, it messes up the alignment of the struct and causes view corruption? I hit that in some of my testing..
> 
> I hacked it like this:
> 
> ```
> @@ -2135,7 +2135,7 @@ void GLShader::PostProcessUniforms() {
>                 GLuint size = _materialSystemUniforms[0]->_std430Size;
>                 GLuint components = _materialSystemUniforms[0]->_components;
>                 size = components ? PAD( size, 4 ) * components : size;
> -               GLuint alignmentConsume = 4 - size % 4;
> +               GLuint alignmentConsume = (size % 4 == 0) ? 0 : 4 - size % 4;
>  
>                 GLUniform* tmpUniform = _materialSystemUniforms[0];
>                 tmp.emplace_back( _materialSystemUniforms[0] );
> ```
> 
> (which is probably wrong, but it fixed my corruption)

Yes. The `Fix bad material SSBO uniform layout for multiple of 16` commit gets that one.

src/engine/renderer/gl_shader.cpp
```diff
@@ -2099,69 +2097,61 @@ GLint GLShader::GetUniformLocation( const GLchar *uniformName ) const {
 	return glGetUniformLocation( p->id, uniformName );
 }
 
-static int FindUniformForAlignment( std::vector<GLUniform*>& uniforms, const GLuint alignment ) {
-	for ( uint32_t i = 0; i < uniforms.size(); i++ ) {
-		if ( uniforms[i]->_std430Size <= alignment ) {
-			return i;
+static auto FindUniformForOffset( std::vector<GLUniform*>& uniforms, const GLuint baseOffset ) {
+	for ( auto it = uniforms.begin(); it != uniforms.end(); ++it ) {
```
VReaperV, line 2101: You can just do `it : uniforms` if you're making it an iterator instead of an index.

src/engine/renderer/gl_shader.cpp
```diff
@@ -2099,69 +2097,61 @@ GLint GLShader::GetUniformLocation( const GLchar *uniformName ) const {
...
+		auto iterNext = FindUniformForOffset( uniformQueue, std140Size );
+		if ( iterNext == uniformQueue.end() ) {
+			// add 1 unit of padding
+			ASSERT( !( *iterNext )->_components ); // array WriteToBuffer impls don't handle padding correctly
+			++std140Size;
+			++_materialSystemUniforms.back()->_std430Size;
+		} else {
+			( *iterNext )->_std430Size = ( *iterNext )->_std430BaseSize;
+			if ( ( *iterNext )->_components ) {
+				ASSERT_GE( ( *iterNext )->_std430Alignment, 4 ); // aligment less than 4 not compatible with std130
```
VReaperV, line 2145: *std140

VReaperV (approved):

PR #1784, opened 2025-09-03
Recognize Apple hardware and driver, allo software rendering

illwieckz:
- DetectGLVendors: recognize Apple hardware and driver

- tr_init: allow software rendering

This make the game runs on virtualized macOS out of the box.

Mesa systems like Linux or FreeBSD just fallbacks on software rendering
without requiring any enablement.

We better want to get the game running at all cost than just
aborting, the user can then tweak the presets to get more
performance if that's really needed.


illwieckz:
It's very slow, like 6 fps in low res, low preset.

It's also full of garbage, it's like it renders only some pixels per frame and the whole scene accumulates until being right, except transparent surfaces are always broken.

While moving:

unvanquished_2025-09-03_043901_000

After accumulation:

unvanquished_2025-09-03_043917_000

While moving:

unvanquished_2025-09-03_032211_000

After accumulation:

unvanquished_2025-09-03_032217_000

But it's good enough to test that the engine builds and runs.

As said, other systems don't require a cvar to be set to fallback on software rendering, so it's not bad to behave the same here.

At some point we may print a warning that software rendering is detected, but it's not bad if the engine starts anyway.

illwieckz:
For example AMD and Nvidia drivers on mac machines are reported as `ATI Technologies Inc.` and `NVIDIA Corporation` when running macOS, we already take care of that, and AMD cards have "ATI" string in their renderer string, etc.

PR #1785, opened 2025-09-04
Add Breakpad toolchain testing tool

slipher:
Test that we can get a usable stack trace from our native and/or NaCl binaries. There are two modes:
- Test the whole thing end-to-end: symbolize the binaries, crash them, and get the stack trace.
- Use a release's symbol zip: only do the crash and stack trace.

For Windows there is a capability to run Breakpad binaries within a WSL VM because there is no way to get a native Windows binary that can symbolize NaCl.

Ignore the individual commits, they will be squashed.

### DIFF
--- a/tools/crash_test.py
+++ b/tools/crash_test.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python3
+"""Breakpad crash test script
+
+Do an end-to-end test of the Breakpad crash dump and stack trace workflow:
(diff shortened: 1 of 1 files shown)


slipher:
TODO: see if it would be nicer to use a Breakpad install dir instead of an in-source build.

DolceTriade (approved):

PR #1786, opened 2025-09-04
crypto: only include gmp.h on MSVC to disable the warning, add a comment explaining why

illwieckz:
Only include gmp.h on MSVC to disable the warning, add a comment explaining why.

----
_Former comment:_

Do not include `gmp.h`, It is unused.

We need to link against GMP because we link against Nettle that requires it, but it looks like ourselves never use any symbol from GMP directly.

Imported from #1433:

When testing #1433 I was able to rebuild every engine binaries and NaCl VMs in Docker with this commit, I also built dll VM locally fine.

### DIFF
--- a/src/engine/qcommon/crypto.h
+++ b/src/engine/qcommon/crypto.h
@@ -27,10 +27,14 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "q_shared.h"
 #include "qcommon.h"
 
+/* The Nettle headers include the GMP header, this disables the warning
+on GMP alone, not the whole Nettle. We don't use GMP directly ourselves. */
+#if defined(_MSC_VER)
 #pragma warning(push)
 #pragma warning(disable : 4146) // "unary minus operator applied to unsigned type, result still unsigned"
 #include <gmp.h>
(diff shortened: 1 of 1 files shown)


src/engine/qcommon/crypto.h
```diff
@@ -27,11 +27,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "q_shared.h"
 #include "qcommon.h"
 
-#pragma warning(push)
-#pragma warning(disable : 4146) // "unary minus operator applied to unsigned type, result still unsigned"
-#include <gmp.h>
```
slipher, line 32: The Nettle headers include the GMP header. So this may be here just for the warning fix. You could get rid of the GMP include but the MSVC warning adjustment would still need to be used for the Nettle headers.

illwieckz: Ah, I see.

slipher:
LGTM

illwieckz:
I modified the `ifdef` to test for `_MSC_VER` instead of `_WIN32`.

slipher:
LGTM

PR #1788, opened 2025-09-05
Material System: always write all of a given stage's data

### LINKED ISSUE #1696: Relief mapping doesn't work when the material system is used

VReaperV:
Fixes #1696


VReaperV:
> Xonotic heightmaps (heightmap in normal alpha) still don't work in material system, but our stuff is all good now.

Are those directly usable/is there some dpk with xonotic heightmaps to test this with?

PR #1790, opened 2025-09-06
Set APPID the SDL3 way, disable DPI scaling

illwieckz:
- `sdl_glimp: properly set the APPID the SDL3 way, drop the SDL2 setting`

  * Drop the SDL2 way of doing things.
  * The SDL3 environment variable wasn't `SDL_HINT_APP_ID` but `SDL_APP_ID`.
  * For some reasons `SetEnv()` doesn't work (SDL3 bug?)
  * We can use `SDL_SetAppMetadataProperty()` instead, and it works.
  * We can probably don't have to care about being on X11/Wayland or not.
  * Use `SDL_SetAppMetadataProperty()` for other things too.

- `sdl_glimp: disable DPI scaling`

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1733,6 +1733,10 @@ static rserr_t GLimp_StartDriverAndSetMode( int mode, bool fullscreen, bool bord
 	StartupWMClass variable to PRODUCT_APPID. */
 	SDL_SetAppMetadataProperty( SDL_PROP_APP_METADATA_IDENTIFIER_STRING, PRODUCT_APPID );
 
+	/* Disable DPI scaling.
+	See the SDL wiki page for details: https://wiki.libsdl.org/SDL3/SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY */
+	SDL_SetHint( SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY, "1" );
+
 	if ( !SDL_WasInit( SDL_INIT_VIDEO ) )
 	{
 		const char *driverName;
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -1722,19 +1722,16 @@ GLimp_StartDriverAndSetMode
 */
 static rserr_t GLimp_StartDriverAndSetMode( int mode, bool fullscreen, bool bordered )
 {
-#if !defined(_WIN32) && !defined(__APPLE__)
+	// See the SDL wiki page for details: https://wiki.libsdl.org/SDL3/SDL_SetAppMetadataProperty
+	SDL_SetAppMetadataProperty( SDL_PROP_APP_METADATA_NAME_STRING, PRODUCT_NAME );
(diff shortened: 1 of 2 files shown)


illwieckz:
Ah lol [that documentation](https://wiki.libsdl.org/SDL3/SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY) is poorly written, I guess I should use `SDL_SetHint()` instead…

illwieckz:
It now works.

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1758,6 +1759,14 @@ static void GLimp_SetAppMetadataProperty( const char* name, const char* value )
 	}
 }
 
+static void GLimp_SetHint( const char* name, const char* value )
+{
+	if ( !SDL_SetHint( name, value ) )
+	{
+		logger.Warn( "SDL_SetHint( %s, %s ) failed: %s\n", name, value, SDL_GetError() );
```
slipher, line 1766: Log message shouldn't have `\n`

Is it really possible for this to fail anyway? In the SDL wiki's example they don't check it. https://wiki.libsdl.org/SDL3/SDL_SetHint

src/engine/sys/sdl_glimp.cpp
```diff
@@ -1750,26 +1750,46 @@ static rserr_t GLimp_SetMode( const int mode, const bool fullscreen, const bool
...
 /*
 ===============
 GLimp_StartDriverAndSetMode
 ===============
 */
 static rserr_t GLimp_StartDriverAndSetMode( int mode, bool fullscreen, bool bordered )
 {
-#if !defined(_WIN32) && !defined(__APPLE__)
+	// See the SDL wiki page for details: https://wiki.libsdl.org/SDL3/SDL_SetAppMetadataProperty
+	GLimp_SetAppMetadataProperty( SDL_PROP_APP_METADATA_NAME_STRING, PRODUCT_NAME );
```
slipher, line 1777: There's a place in ClientApplication where the name is also set (with SDL_HINT_APP_ID) that could be cleaned up now.

illwieckz: Thanks, now cleaned-up.

illwieckz:
I removed the CREATOR, URL and COPYRIGHT metadata things. I grepped SDL3 and this part is still work in progress on their side any way. For example the COPYRIGHT string looks to be made to be used in the macOS `About` window in the global application menu, but that part is still flagged as `FIXME` in their code.

So, while I find useful to properly implement the various metadata a Dæmon-based game may set, I don't want to delay this PR because of things even not finished in SDL3 itself.

slipher:
LGTM although it's annoying that the `SDL_SetHint` result is checked in this one case while it's not checked in others.

slipher:
I read the source for SDL_SetHint and actually it would be better NOT to check the return value. It's an "error" if the user overrode the hint with an environment variable! Try it: `SDL_VIDEO_WAYLAND_SCALE_TO_DISPLAY=1 ./daemon` prints a warning.

illwieckz:
OK, I now removed the wrappers.

PR #1794, opened 2025-09-09
Make r_lightmode 0 work with material system

slipher:


src/engine/renderer/Material.cpp
```diff
@@ -480,6 +480,18 @@ void MaterialSystem::GenerateTexturesBuffer( std::vector<TextureData>& textures,
 	}
 }
 
+int MaterialSystem::GetLightMapIdx( const MaterialSurface &surface, lightMode_t mode )
+{
+	if ( mode == lightMode_t::MAP )
+	{
+		return GetLightMapNum( &surface ) & 255;
```
VReaperV, line 487: Not having `HasLightMap()` can crash on some maps IIRC.

PR #1796, opened 2025-09-11
tr_shader: linearize fog colors

illwieckz:
Cherry-pick from #1730:

- https://github.com/DaemonEngine/Daemon/pull/1730

This one shouldn't be controversial.

One discussion possible is that we can do this conversion either in `tr_shader.cpp`, either in `tr_bsp.cpp`.

Fog shaders should be parsed when requested by the BSP code, there is no meaning outside of that, so the behavior should be the same.


illwieckz:
This looks ready to me.

slipher:
LGTM

PR #1803, opened 2025-09-13
Illwieckz/font texture

illwieckz:
- tr_font: no need for file extension
  I did it for log æsthetics, but it fixes `testshader fonts/unifont_0_0_32` as a side effect.
- tr_image: use GL_RGBA8 explicitly for glyphes
  We always use an explicit format as `internalFormat` everywhere else.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -1509,7 +1509,7 @@ image_t *R_CreateGlyph( const char *name, const byte *pic, int width, int height
 
 	image->uploadWidth = width;
 	image->uploadHeight = height;
-	image->internalFormat = GL_RGBA;
+	image->internalFormat = GL_RGBA8;
 
 	GL_TexImage2D( GL_TEXTURE_2D, 0, image->internalFormat, width, height, 0, GL_RGBA, GL_UNSIGNED_BYTE, pic, false );
 
--- a/src/engine/renderer/tr_font.cpp
+++ b/src/engine/renderer/tr_font.cpp
@@ -419,7 +419,7 @@ static void RE_StoreImage( fontInfo_t *font, int chunk, int page, int from, int 
 	}
 
 
-	Com_sprintf( fileName, sizeof( fileName ), "%s_%i_%i_%i.png", font->name, chunk, page, font->pointSize );
+	Com_sprintf( fileName, sizeof( fileName ), "%s_%i_%i_%i", font->name, chunk, page, font->pointSize );
 
 	image = R_CreateGlyph( fileName, buffer, FONT_SIZE, y );
 


slipher:
It might be good to use a name that can't conflict with / is distinct from normal file names. Like for the RmlUi font textures we have `$generatedTexture6` or whatever.

illwieckz:
That's doable. I guess it was named like this because in the past it could be a generated image in the pk3. There are comments about it. But it was tied to some limitations we don't have if I'm right.

illwieckz:
I discovered that `testshader` can't find and then display an image whose name starts with `$`.

src/engine/renderer/tr_image.cpp
```diff
@@ -1509,7 +1509,7 @@ image_t *R_CreateGlyph( const char *name, const byte *pic, int width, int height
 
 	image->uploadWidth = width;
 	image->uploadHeight = height;
-	image->internalFormat = GL_RGBA;
+	image->internalFormat = GL_RGBA8;
 
 	GL_TexImage2D( GL_TEXTURE_2D, 0, image->internalFormat, width, height, 0, GL_RGBA, GL_UNSIGNED_BYTE, pic, false );
```
slipher, line 1514: Do you want to write RGBA8 here too?

slipher:
> > That's doable. I guess it was named like this because in the past it could be a generated image in the pk3. There are comments about it.
> 
> OK, that makes some sense, it's fine as is.

Actually no, the code is confusing but it doesn't look like there would be any way to use an image from a pak instead of rendering using Freetype. `RE_RegisterShaderFromImage` creates a shader directly from an `image_t` without doing lookups in the shader or image namespaces. So using a name that has no overlap with normal shaders would make more sense.

src/engine/renderer/tr_image.cpp
illwieckz: It looks like answer is no:

> `format` determines the composition of each element in data. It can assume one of these symbolic values:
>
> `GL_RED`
>
> Each element is a single red component. The GL converts it to floating point and assembles it into an RGBA element by attaching 0 for green and blue, and 1 for alpha. Each component is clamped to the range [0,1].
>
> `GL_RG`
>
> Each element is a red/green double. The GL converts it to floating point and assembles it into an RGBA element by attaching 0 for blue, and 1 for alpha. Each component is clamped to the range [0,1].
>
> `GL_RGB`, `GL_BGR`
>
> Each element is an RGB triple. The GL converts it to floating point and assembles it into an RGBA element by attaching 1 for alpha. Each component is clamped to the range [0,1].
>
> `GL_RGBA`, `GL_BGRA`
>
> Each element contains all four components. Each component is clamped to the range [0,1].
>
> `GL_DEPTH_COMPONENT`
>
> Each element is a single depth value. The GL converts it to floating point and clamps to the range [0,1].
>
> `GL_DEPTH_STENCIL`
>
> Each element is a pair of depth and stencil values. The depth component of the pair is interpreted as in GL_DEPTH_COMPONENT. The stencil component is interpreted based on specified the depth + stencil internal format
> -- https://registry.khronos.org/OpenGL-Refpages/gl4/html/glTexImage2D.xhtml

slipher:
> I believe we have no reason to load such file from pk3/dpk anymore, so I'm OK with making the name using some internal naming scheme, but then we must clean-up anything about loading such file from the VFS.

OK, if you find anything to clean up go ahead. Regarding pre-rasterized fonts I'm only aware of the code loading `gfx/2d/bigchars` but that's a single image that can't be split into blocks.

illwieckz:
> I discovered that `testshader` can't find and then display an image whose name starts with `$`.

Not only our builtin image detector doesn't check for paths starting with `$`, but our builtin image detector is actually checking for the image name to _not_ contain any `/`, and font names contain one.

illwieckz:
But is the image generated by `RE_StoreImage()` actually used? The `image` object is local to the function…

slipher:
> But is the image generated by `RE_StoreImage()` actually used? The `image` object is local to the function…

A shader is created using the image and the shader handle is stored.

illwieckz:
So now, with latest patch, instead of being named `fonts/unifont_0_0_16.tga` it is named `*fonts_unifont_0_0_16`.

slipher:
LGTM

PR #1804, opened 2025-09-14
tr_image: rework some builtin images

illwieckz:
- tr_image: extend internal image detection
- tr_image: do not rely on previously set data for cinematic, numberize and rename as `*name`
- tr_image: resize builtin images as `1×1`
- renderer: simplify builtin color image name parsing, rename as `$name`, forget about unused aliases
- tr_image: rename framebuffer images as `*name`
- tr_image: rename remaining builtin images as `$name`
- tr_image: rename generated textures as `*name`
- tr_image: make the `$black` builtin image opaque
- tr_image: NUKE the $red, $green and $blue builtin images

___
Original message:

- make `_black` opaque
- add `_null` for zeroed image (instead of `_black`)
- explicitly set `_cinematic` color instead of relying on previously set values
  imported from https://github.com/DaemonEngine/Daemon/pull/1793
- make images `1×1` sized, makes the code simpler
- extend shader parsing for keywords like `$blueimage`.

This will conflict with #1793 but I plan to rebase #1793 over it once this is merged.

### DIFF
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -2850,7 +2850,7 @@ qhandle_t RE_GenerateTexture( const byte *pic, int width, int height )
 {
 	R_SyncRenderThread();
 
-	std::string name = Str::Format( "$generatedTexture%d", tr.numGeneratedTextures++ );
+	std::string name = Str::Format( "*generatedTexture%d", tr.numGeneratedTextures++ );
 
 	imageParams_t imageParams = {};
 	imageParams.bits = IF_NOPICMIP;
--- a/src/engine/renderer/tr_image.cpp
+++ b/src/engine/renderer/tr_image.cpp
@@ -2649,34 +2649,34 @@ void R_CreateBuiltinImages()
 	imageParams.filterType = filterType_t::FT_LINEAR;
 	imageParams.wrapType = wrapTypeEnum_t::WT_REPEAT;
 
-	tr.whiteImage = R_CreateImage( "_white", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
+	tr.whiteImage = R_CreateImage( "$white", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
 
 	// we use a solid black image instead of disabling texturing
 	memset( data, 0, sizeof( data ) );
 
-	tr.blackImage = R_CreateImage( "_black", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
+	tr.blackImage = R_CreateImage( "$black", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
 
 	// red
 	Vector4Set( data, 255, 0, 0, 255 );
 
-	tr.redImage = R_CreateImage( "_red", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
+	tr.redImage = R_CreateImage( "$red", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
 
 	// green
 	Vector4Set( data, 0, 255, 0, 255 );
 
-	tr.greenImage = R_CreateImage( "_green", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
+	tr.greenImage = R_CreateImage( "$green", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
 
 	// blue
 	Vector4Set( data, 0, 0, 255, 255 );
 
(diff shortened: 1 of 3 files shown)


illwieckz:
Also this may spark some discussions. First, the reason why there was no `$black` usage, etc. is because ioq3 never parsed that. ioq3 only parsed `$whiteimage` and `*white`.

It also means ioq3 never parsed `$white` and `_white`.

It means we really don't have to parse all those variants.

I grepped etlegacy and realrtcw and none of them did more than defining `$whiteimage` and `*white`, no syntax variant, no other colors.

This also brings the topic that we may just name images like `_black` as `$blackimage` directly.

If I'm right, we don't need to actually parse things like `_white` or `_black` in `tr_shader` because the image loading code will find them as they are created and loaded as startup anyway. Parsing them just skips some code.

So an alternative code may be to only parse for `*white` for ioq3 compatibility, and name the image `$whiteimage`, and name all other builtin images this way, and not parse any other color image names, relying on the map lookup to find them.

illwieckz:
The only `$blackimage` and `$black` thing we find in etlegacy is in their GL2 renderer which deviates from XreaL like us. That look pure XreaL stuff there, no one uses that as far as I know. Their other renderers (GL1, GLES) don't parse that at all.

In fact, ioq3 and QIII don't parse `*white` (even not in ioq3's renderer2), that's because QIII names that image `*white` so it works by image lookup.

slipher:
> Hmm, actually if we name the images like `$whiteimage`, then `testshader` can't find them.

Are you sure? For me it works with `$generatedTexture0`.

illwieckz:
> > Hmm, actually if we name the images like `$whiteimage`, then `testshader` can't find them.
> 
> Are you sure? For me it works with `$generatedTexture0`.

Test that latest commit. It will even break the console because some shaders rely on `$whiteimage`.

illwieckz:
Hmm, the problem is not the `$` character:

```
]/testshader $blueimage
Debug: loading '$blueimage' image as shader 
image params mismatch for $blueimage: 0x0 0 0/0 0 0 vs. 0x40000 0 0/0 0 0 
Warn: Couldn't find image file '$blueimage' 
Shader $blueimage not found 
```

slipher:
Oh right, the code to treat builtin images specially (by allowing params mismatch) assumes that the builtin image names start with `_` not `$`.

VReaperV:
> For a transparent shader we should make it possible to parse a shader with 0 stages (instead of assuming it is an error or a fog). Using a stage with an image is wasteful since it still has to execute the shaders.

If it's currently assumed to be fog, then changing that might break some maps.

slipher:
Could we have a naming convention that distinguishes between images which are intended to be available for artists to use in shaders, and ones that are an internal implementation detail? Like the white image is good to use in shaders, but the light tile images aren't.

I'm not saying that we should make it impossible to use those in shaders (it's convenient for debugging sometimes), but some convention to know whether it's supported or not would be nice.

src/engine/renderer/tr_image.cpp
```diff
@@ -2771,21 +2771,6 @@ void R_CreateBuiltinImages()
 
 	tr.blackImage = R_CreateImage( "$black", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
 
-	// red
-	Vector4Set( data, 255, 0, 0, 255 );
-
-	tr.redImage = R_CreateImage( "$red", ( const byte ** ) &dataPtr, 1, 1, 1, imageParams );
```
slipher: Also delete these fields from the `tr` struct

src/engine/renderer/tr_image.cpp
```diff
@@ -1819,11 +1819,20 @@ image_t *R_FindImageFile( const char *imageName, imageParams_t &imageParams )
...
 
 			// Built-in images can't be reloaded with different parameters, so return them as-is.
-			// For most of the usable ones e.g. _white, parameters wouldn't make a difference anyway.
+			// For most of the usable ones e.g. $white, parameters wouldn't make a difference anyway.
 			// HACK: detect built-in images by naming convention, though nothing stops users from using such names
-			if ( image->name[ 0 ] == '_' && !strchr( image->name, '/' ) )
+			switch ( image->name[ 0 ] )
 			{
-				return image;
+				case '_': // TODO: Remove once _fog is removed.
```
slipher, line 1826: You can remove `_` now since it's not an image anyone needs to load from a shader.

src/engine/renderer/tr_image.cpp
illwieckz: Done.

src/engine/renderer/tr_image.cpp
illwieckz: Done. Also this `_fog` image is now NUKED anyway.

PR #1806, opened 2025-09-15
Do not assume shaders without a stage is fog

illwieckz:
- tr_shader: do not ignore shaders without a stage, it's a valid use case for a transparent surface
- tr_shader: do not assume shaders without a stage is fog

It makes possible to do this:

```c
// The transparent texture that is also transparent in editor too.
// It should not be listed and selectable in editor. It can be used
// on model surfaces to hide them.
gfx/shared_colors/transparent
{
	qer_editorImage textures/shared_colors_src/transparent_d
	qer_trans 0

	surfaceparm nolightmap
	surfaceparm nomarks
	surfaceparm trans
}


gfx/shared_colors/transparent_nonsolid
{
	qer_editorImage textures/shared_colors_src/transparent_d
	qer_trans 0

	surfaceparm nolightmap
	surfaceparm nomarks
	surfaceparm nonsolid
	surfaceparm trans
}
```

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -4735,10 +4735,10 @@ static bool ParseShader( const char *_text )
 		}
 	}
 
-	// ignore shaders that don't have any stages, unless it is a sky or fog
+	// Make shaders without stages that are not fog and not sky as blend (they are fully transparent).
 	if ( s == 0 && !shader.forceOpaque && !shader.isSky && !( shader.contentFlags & CONTENTS_FOG ) && implicitMap[ 0 ] == '\0' )
 	{
-		return false;
+		shader.sort = Util::ordinal( shaderSort_t::SS_BLEND0 );
 	}
 
 	return true;
@@ -5697,12 +5697,6 @@ static void ValidateStage( shaderStage_t *pStage )
 // without a thorough investigation.
 static float DetermineShaderSort()
 {
-	// fogonly shaders don't have any stage passes
-	if ( numStages == 0 && !shader.isSky )
-	{
-		return Util::ordinal(shaderSort_t::SS_FOG);
-	}
-
 	for ( size_t stage = numStages; stage--; )
 	{
 		ASSERT( stages[ stage ].active );


illwieckz:
@slipher do you know if that can be rewritten without a stage? What that `depthWrite` keyword does here?

```c
textures/common/portal
{
	qer_editorImage textures/common_src/portal_p

	surfaceparm nolightmap

	portal

	{
		map textures/shared_colors_src/transparent_d
		blendfunc GL_ONE GL_ONE_MINUS_SRC_ALPHA
		depthWrite
	}
}
```

illwieckz:
> But you could write the stage without using a transparent image, for example:
> 
> ```
> {
>     map $white
>     blendfunc GL_ZERO GL_ONE
>     depthWrite
> }
> ```

OK, that's good enough.

illwieckz:
Can it be done with `$black` instead?

illwieckz:
With this branch the transparent shader is listed this way:

```
]/listShaders *transparent
--------------------------------------------------------------------------- 
num   regFlags shaderSort     stageType              stageNumber:shaderName 
--------------------------------------------------------------------------- 
2077  _____    OPAQUE         n/a                    -:gfx/shared_colors/transparent 
--------------------------------------------------------------------------- 
2078 total shaders, 0 total stages, largest shader has 0 stages 
--------------------------------------------------------------------------- 
```

Yes it's listed as `OPAQUE`, but there is no stage. Should we make it something else, like `BLEND0`?

What would be the drawback of being opaque while not having any stage? Is there some code that would not render behind or so?

illwieckz:
So what I did is that such shader are now `BLEND0`, like they were when I used an explicit transparent image.

Before (`0.55.4`):

```
]/listShaders *transparent 
----------------------------------------------------------------------------- 
num   shaderType shaderSort     stageType              stageNumber:shaderName 
----------------------------------------------------------------------------- 
2076  2D         BLEND0         COLLAPSE_DIFFUSEMAP    0:gfx/shared_colors/transparent 
----------------------------------------------------------------------------- 
2077 total shaders, 1 total stages, largest shader has 1 stages 
----------------------------------------------------------------------------- 
```

After (this branch):

```
]/listShaders *transparent                
--------------------------------------------------------------------------- 
num   regFlags shaderSort     stageType              stageNumber:shaderName 
--------------------------------------------------------------------------- 
2077  _____    BLEND0         n/a                    -:gfx/shared_colors/transparent 
--------------------------------------------------------------------------- 
2078 total shaders, 0 total stages, largest shader has 0 stages 
--------------------------------------------------------------------------- 
```

src/engine/renderer/tr_shader.cpp
```diff
@@ -5698,7 +5700,7 @@ static void ValidateStage( shaderStage_t *pStage )
 static float DetermineShaderSort()
 {
 	// fogonly shaders don't have any stage passes
-	if ( numStages == 0 && !shader.isSky )
+	if ( shader.isFog && !shader.isSky )
```
slipher, line 5703: I think you could just remove this whole if statement and remove `isFog`. In the place where `isFog` is set, it already puts `SS_FOG`.

illwieckz: Also fog shaders are supposed to have `contentparm fog`, so `shader.contentFlags & CONTENTS_FOG`.

illwieckz:
Actually, there is a shader keyword that does what that PR did (before I set `BLEND0`): `forceOpaque`.

Is is used in no shader of no game I have grepped, it only exists in XreaL renderers like Dæmon and etlegacy renderer2.

illwieckz:
The `forceOpaque` shader keyword in some XreaL (the game PoC) shaders:

```
textures/common/caulk
{
	qer_editorimage textures/common/caulk.tga
	qer_trans 0.5
	
	noshadows
	forceOpaque			// will still seal levels
}
```

```
textures/skies/kat_q4dm3
{
//Grey variant of the CanyonFog_Sky
	qer_editorimage	gfx/env/kat_q4dm3_forward.tga
	noFragment
	forceOpaque
	noimpact
	noshadows
	{
		blend			blend			// so transparent windows can draw on top of it
		cameraCubeMap		gfx/env/kat_q4dm3
		texgen			skybox
	}
}
```

I don't know what it is for. But we don't need that to keep the shader when there is no stage, and we better set `BLEND0`.

slipher:
LGTM

PR #1810, opened 2025-09-17
Fix wrong assert about material uniforms

slipher:
Spotted by @VReaperV


VReaperV:
This will cause a UB if the vector is empty.

PR #1811, opened 2025-09-17
Fix crash on dummygame load with r_lazyShaders = 1

slipher:
Regression in de2fb50511f8c08feca30a6787a3fd613b61a661. I guess this shader marking can happen while there are triangles waiting in the buffer. Though I don't understand the connection with backEnd.currentEntity.


illwieckz (approved):
LGTM.

PR #1814, opened 2025-09-18
Copy depth texture to avoid texture feedback loops

### LINKED ISSUE #1783: Depth fade broken on Apple Silicon
Basically the same as #1676 but with Apple hardware instead of AMD. 

Image

The texture barrier extension is not available, so the fix for #1676 doesn't help.

Now this GPU is capable hardware; the driver just has poor OpenGL support. So it wouldn't hurt us to copy the depth buffer to a separate texture once per frame. We just want to avoid doing that on weak hardware as long as it looks OK without doing that.

slipher:
Depends on #1813.

Implement a code path for sampling `u_DepthMap` in shaders from a separate copy of the depth texture, instead of the same one that is attached to the rendering FBO. By default, this will be enabled if the driver does not implement `TextureBarrier`, in order to prevent texture feedback loops. Behavior is controlled by `r_copyDepthBuffer`. This is meant to fix #1783. Seems to work on my Mac system.

Also I keep having some issues with SSAO + bindless textures on a Mesa + AMD setup. In a game with map `survival-pit` I got the SSAO artifacts as shown below. (Same ones I used to get with dummygame. I still got those even after #1731 which I though would fix them but didn't, but they may have gone away after a Mesa update.) I can't reproduce the artifacts in a local game but if I play a demo recorded from an online game, I get the artifacts the whole time. Anyway `set r_copyDepthBuffer 2` from this PR fixes those artifacts for me.

unvanquished_2025-09-05_142446_000

Things to do:
- Check if I'm copying the depth buffer in the best way. Is there a depth-only format I should be using instead of the depth + stencil?
- Disable depth buffer copy if `r_depthShaders` (incompatible debug cvar) is disabled
- Ask @illwieckz to assess the performance of this on low-end cards
- Determine which hardware/driver combinations still have issues with reading the depth buffer despite implementing texture barrier. (This could go in another PR)
- Maybe skip texture barriers if the duplicated depth buffer is used?

### DIFF
--- a/src/engine/renderer/Material.cpp
+++ b/src/engine/renderer/Material.cpp
@@ -904,7 +904,7 @@ void BindShaderGeneric3D( Material* material ) {
 	gl_genericShaderMaterial->SetUniform_ModelMatrix( backEnd.orientation.transformMatrix );
 	gl_genericShaderMaterial->SetUniform_ModelViewProjectionMatrix( glState.modelViewProjectionMatrix[glState.stackIndex] );
 
-	gl_genericShaderMaterial->SetUniform_DepthMapBindless( GL_BindToTMU( 1, tr.currentDepthImage ) );
+	gl_genericShaderMaterial->SetUniform_DepthMapBindless( GL_BindToTMU( 1, tr.depthSamplerImage ) );
 
 	// u_DeformGen
 	gl_genericShaderMaterial->SetUniform_Time( backEnd.refdef.floatTime - backEnd.currentEntity->e.shaderTime );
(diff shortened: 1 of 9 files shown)


slipher:
This is ready. On my worst GPU, the Intel UHD 630, I registered about a 10% performance drop (53 -> 48 FPS). So we definitely want to disable it in low presets.

illwieckz:
On an ATI X550 I only seen a 1 fps difference, so I guess it depends on the driver.

I'll add the cvar to the presets indeed.

slipher:
This GPU can have texture barrier according to reports. So make sure your comparison is between `r_readonlyDepthBuffer 0` and `r_readonlyDepthBuffer 2`.

illwieckz:
See https://github.com/Unvanquished/Unvanquished/pull/3430 for the game presets/ui menu:

- https://github.com/Unvanquished/Unvanquished/pull/3430

Now that I think about it, it's probably better to do instead:

- `0`:  disabled
- `1`: enabled
- `2`: auto.

Not only because `0`/`1` is not bad for disabled/enabled, and because then the higher number (`2`) would “bring the best” and usually people look for the higher number for getting the best. I noticed that in the menu I ordered `1`, `2` ,`0` to order it that way (from top to bottom):

- `When necessary (recommended)`,
- `Enabled (slower)`,
- `Disabled (faster, buggy)`.

I felt it was better for the UI where we make the effort to put the “best” at the top.

illwieckz:
> This GPU can have texture barrier according to reports. So make sure your comparison is between `r_readonlyDepthBuffer 0` and `r_readonlyDepthBuffer 2`.

Same missing 1 fps when forcing the read only as well.

illwieckz:
OK, it makes sense that way then.

I updated my PR to use a checkbox instead.

illwieckz (approved):
LGTM.

illwieckz:
Merging this for the release.

PR #1818, opened 2025-09-25
[CI] Build dummy gamelogic and run client tests

slipher:


slipher:
I just discovered `SDL_VIDEODRIVER=offscreen`. We can use this to run the client tests in the headless CI environment.

slipher:
I set the OpenAL backend to `null` to suppress error spam, but on the other hand, maybe I should get rid of that so that we can test that the app still runs when the audio driver is not working.

PR #1822, opened 2025-09-26
Improve deps handling in CMake, make possible to not use libs from deps folders, and to use deps folders on unsupported platforms when provided

illwieckz:
## Improve deps handling in CMake

`cmake: use extra external deps if the directory exist and add an option to not use libs from them`

- Add an option to not build and link against libraries from external_deps
  * Purposed to make possible to build on linux using system-provided SDL3 instead of the one in DEPS, in the rare case the system already provides SDL3.
  * Purposed to make possible to build on `macos-arm64` using the `macos-amd64` deps archive to find the NaCl runtime, relying on the system (example: brew) to find libraries.
  * Purposed to make possible to build on `freebsd` using the `linux` deps archive to find the NaCl runtime, relying on the system to find libraries.
- Use the `macos-amd64` deps on `macos-arm64` only if the user didn't provide a `macos-arm64` deps folder by himself.
- Use the `linux` deps on `freebsd` only if the user didn't provide a freebsd deps folder by himself.
- Use the deps folder if the user provided it by himself, even if we don't know yet about this system or architecture or we don't support that.

---

`cmake: only set -march if the architecture is known`

Do not set an empty -march= option when attempting to build on an unknown system.

We only print a warning when attempting to build on an unknown system, so this should not expect it is never built for an unknown system.

### DIFF
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -642,7 +642,9 @@ elseif (NOT NACL)
 
 	option(USE_CPU_GENERIC_ARCHITECTURE "Enforce generic -march and -mtune compiler options" ON)
 	if (USE_CPU_GENERIC_ARCHITECTURE)
-		try_c_cxx_flag_werror(MARCH "-march=${GCC_GENERIC_ARCH}")
+		if (GCC_GENERIC_ARCH)
+			try_c_cxx_flag_werror(MARCH "-march=${GCC_GENERIC_ARCH}")
+		endif()
 
 		if (GCC_GENERIC_TUNE)
 			try_c_cxx_flag_werror(MTUNE "-mtune=${GCC_GENERIC_TUNE}")


illwieckz:
The first commit is important as I guess some distros like Arch would want to update their AUR script to use the Arch-provided SDL3 instead of the on in the deps archive, that one we provide because SDL3 is to new and many stable releases didn't packaged it yet.

It should also make easier for building on extra plaforms (not supported in unizip, but buildable, like macos-arm64 and freebsd).

slipher:
How about adding an option `NACL_DEPS_DIR`? If set this would be the base path for NaCl-related stuff. This could be helpful in the Arch PKGBUILD, arm64 Mac, and FreeBSD cases. That would be simpler than referring to the entire deps dir of some platform but then selectively disabling parts of it.

illwieckz:
I removed the riscv64 demonstration. That acted as a demonstration for:

1. using a deps dir if it's provided beforehand, even if not being a known platform.
2. using an architecture for which we don't set a `-march` option.

I now have a much larger branch on my end with that riscv stuff, but also macos-arm64 build (probably ready to use) and WIP freebsd. This was enough for me to test that CMake code.

My freebsd experiment was enough to verify that yes it was using the linux archive without the SDL3 libs if no freebsd deps is provided, but as soon as I provide a freebsd deps folder myself, it uses the provided deps folder.

I assume that part would work as well for macos-arm64.

Also it looks like that freebsd experiment validated that the option to _not_ use libs from a deps dir, even when using that deps dir, is working, which would be useful for linux too (when wanting to enforce linking everything against the system).

Now what's remaining in that branch is the CMake stuff for that deps detection and deps lib skip, plus the `-march` stuff that is very small.

illwieckz:
I added some logging so it's now reporting what deps folder it is using, and it reports if it uses libs from the deps folder or not.

illwieckz:
With `-DUSE_EXTERNAL_DEPS_LIBS=ON`:

```
-- Found Threads: TRUE  
-- Found OpenGL: /usr/lib/x86_64-linux-gnu/libOpenGL.so   
-- Found OpenGL ABI: LEGACY
-- Found OpenGL library: /usr/lib/x86_64-linux-gnu/libGL.so
-- Found ZLIB: deps/linux-amd64-default_11/lib/libz.a (found version "1.3.1")  
-- Found Nettle: deps/linux-amd64-default_11/lib/libnettle.a  
-- Found GMP: deps/linux-amd64-default_11/lib/libgmp.a
-- Found CURL: deps/linux-amd64-default_11/lib/cmake/CURL/CURLConfig.cmake (found version "8.15.0")  
-- Found Ogg: deps/linux-amd64-default_11/lib/libogg.a  
-- Found Vorbis: deps/linux-amd64-default_11/lib/libvorbis.a  
-- Found Opus: deps/linux-amd64-default_11/lib/libopusfile.a  
-- Found WebP: deps/linux-amd64-default_11/lib/libwebp.a  
-- Performing Test _WEBP_COMPILATION_TEST
-- Performing Test _WEBP_COMPILATION_TEST - Success
-- Found JPEG: deps/linux-amd64-default_11/lib/libjpeg.a (found version "80") 
-- Found PNG: deps/linux-amd64-default_11/lib/libpng.a (found version "1.6.50") 
-- Found Freetype: /usr/lib/x86_64-linux-gnu/libfreetype.so (found version "2.13.2") 
-- Found GLEW: deps/linux-amd64-default_11/lib/libGLEW.a  
-- Found OpenAL: deps/linux-amd64-default_11/lib/libopenal.a  
-- Generating DaemonBuildInfo/Engine.cpp
-- Generating DaemonBuildInfo/Engine.h
```

With `-DUSE_EXTERNAL_DEPS_LIBS=OFF`:

```
-- Found Threads: TRUE  
-- Found OpenGL: /usr/lib/x86_64-linux-gnu/libOpenGL.so   
-- Found OpenGL ABI: LEGACY
-- Found OpenGL library: /usr/lib/x86_64-linux-gnu/libGL.so
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.3")  
-- Found Nettle: /usr/lib/x86_64-linux-gnu/libnettle.so  
-- Found GMP: /usr/lib/x86_64-linux-gnu/libgmp.so
-- Found CURL: /usr/lib/x86_64-linux-gnu/libcurl.so (found version "8.5.0")  
CMake Error at CMakeLists.txt:792 (find_package):
  Could not find a package configuration file provided by "SDL3" with any of
  the following names:

    SDL3Config.cmake
    sdl3-config.cmake

  Add the installation prefix of "SDL3" to CMAKE_PREFIX_PATH or set
  "SDL3_DIR" to a directory containing one of the above files.  If "SDL3"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Configuring incomplete, errors occurred!
```

So the `USE_EXTERNAL_DEPS_LIBS` option is working as expected.

illwieckz:
I merge this for the release.

PR #1824, opened 2025-09-27
Only write the `shaders.cpp` file when its content changes

### LINKED ISSUE #1820: The `shaders.cpp` file is always regenerated, forcing to relink the engine client binary
In #1805 it was implemented the generation of the `shader.cpp` file (which is very appreciated!:

- https://github.com/DaemonEngine/Daemon/pull/1805

One unwanted side effect of that is that it is then always rebuilt and then the engine client binary is always relinked, even if nothing changed.

A mechanism not regenerating it when nothing changed would be welcome.

illwieckz:
Add a reusable helper to write a generated file only if its content changed.

Use it for writing (or not writing) the `shaders.cpp` file.

Fixes #1820:

- https://github.com/DaemonEngine/Daemon/issues/1820


cmake/DaemonBuildInfo.cmake
```diff
@@ -18,20 +18,26 @@ macro(daemon_add_buildinfo TYPE NAME VALUE)
 	set(DAEMON_BUILDINFO_H "${DAEMON_BUILDINFO_H}extern const ${TYPE} ${NAME};\n")
 endmacro()
 
+macro(daemon_write_generated GENERATED_SLUG GENERATED_PATH GENERATED_CONTENT)
```
slipher, line 21: unused GENERATED_SLUG

src.cmake
```diff
@@ -143,8 +143,44 @@ set(GLSLSOURCELIST
     ${ENGINE_DIR}/renderer/glsl_source/skybox_fp.glsl
 )
 
+if (BUILD_CLIENT)
```
slipher, line 146: Complex logic shouldn't go in `src.cmake`. Just the file lists

illwieckz: But then we have a chicken-and-egg problem because the list of GLSL files from the list should be known for generating the ` shaders.cpp` file listing them, and then only the path to that file is known to be set in that list.

illwieckz: Well, that's annoying to not rely on a variable, but OK, I now rewrote by copy-pasting the name.

CMakeLists.txt
```diff
@@ -973,25 +973,32 @@ if (BUILD_CLIENT)
     set(SHADERS_CPP_TEXT "// This file is auto-generated by CMakeLists.txt.\n")
     string(APPEND SHADERS_CPP_TEXT "#include \"common/Common.h\"\n\n")
     set(SHADERMAP_TEXT "")
+
     foreach(res ${GLSLSOURCELIST})
         get_filename_component(filename_no_ext ${res} NAME_WE)
         set(outpath ${EMBED_INCLUDE_DIR}/${filename_no_ext}.glsl.h)
+        list(APPEND TARGET_client_FILES ${outpath})
```
slipher, line 980: What's this?

illwieckz: Something I forgot to delete. Now removed.

slipher:
LGTM

PR #1825, opened 2025-09-28
breakpad: fix linking on Linux

### LINKED ISSUE #1817: Linux Breakpad doesn't build
Building with `-DUSE_BREAKPAD=1` on Linux fails with a link error. I assume this happened upon merging the Google upstream for DWARF5 support.

We should add USE_BREAKPAD to the CI.

illwieckz:
Fixes #1817:

- https://github.com/DaemonEngine/Daemon/issues/1817

It broke because the link didn't find the `breakpad_getcontext` symbol.
I tried adding `src/common/linux/breakpad_getcontext.S` to the source list, but it didn't worked.
Fortunately this source is only meant to be used when `getcontext` isn't provided by the system, setting `HAVE_GETCONTEXT` tells the source the function is there.

I haven't tested on Bullseye yet (Docker build distro). Actually I'm surprised that me testing the docker build didn't faced that error before. If it works on Bullseye, then we will assume it works everywhere.


slipher:
LGTM

illwieckz:
I modified it a bit to only set the definition for that `srclibs-breakpad` target and not for the whole.

I also know how to make it work by listing the `src/common/linux/breakpad_getcontext.S` file, when doing so it is also required to add the `ASM` language to the `project(Daemon C CXX)` line, but I postpone the evaluation of such solution because it may have toolchain implication in cross-compilation use case.

illwieckz:
I merge this for the release.

Issue #1828, opened 2025-09-28
Stack trace crashes with Windows DLL


slipher:
With MSVC, if I enable the stack traces with `USE_CPP23`, run with DLL gamelogic, and do the command `cgame.injectFault drop` or `cgame.injectFault exception` or likewise with sgame, I get a crash inside the C runtime's stack trace machinery.

VReaperV:
Yeah, I got that one too, though IIRC it did print the staktrace. Maybe it fails to stack unwind properly or something.

slipher:
I expect you would have to be very lucky to get the fully stacktrace due to the fully buffered logs...

slipher:
Actually the problem happens on Linux too. And the stack trace code can still cause this crash even if the feature is disabled: it goes into mutual recursion between `Sys::Drop` and `Log::Warn( "StackTrace unavailable: CPP23 required" );` until it hits a stack overflow. I assume the same thing also happens with logging an actual stack trace.

slipher:
I still get the stack overflow with e.g. `cgame.injectFault drop` even when the logging is disabled. There is recursion between checking a `writeSrcLocation` cvar, and re-dropping because the IPC state machine is not in a state to allow fetching the cvar.

It doesn't seem to happen with NaCl gamelogic, but I don't understand why not.

PR #1830, opened 2025-09-29
Remove "StackTrace unavailable" message

slipher:
This will at least prevent the crash described in #1828 from happening when the stack trace feature is disabled. The message was annoying anyway.


illwieckz:
I confirm it fixes the issues on my end.

slipher:
You can use the `cgame.injectFault` command to test that.

PR #1832, opened 2025-09-30
use VectorNormalizeFast() instead of VectorNormalize() anytime the returned length isn't used

illwieckz:
Use `VectorNormalizeFast()` instead of `VectorNormalize()` anytime the returned length isn't used.

It does the same, but with less computations since we don't need to compute the length.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -4084,7 +4084,7 @@ static bool ParseShader( const char *_text )
 
 			tr.sunLight[ 2 ] = atof( token );
 
-			VectorNormalize( tr.sunLight );
+			VectorNormalizeFast( tr.sunLight );
 
 			token = COM_ParseExt2( text, false );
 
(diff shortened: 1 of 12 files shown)


src/engine/renderer/tr_shader.cpp
```diff
@@ -4084,7 +4084,7 @@ static bool ParseShader( const char *_text )
 
 			tr.sunLight[ 2 ] = atof( token );
 
-			VectorNormalize( tr.sunLight );
+			VectorNormalizeFast( tr.sunLight );
```
slipher, line 4087: I don't see any point to using the fast version for something that's done very infrequently and only at loading time.

slipher (commented):
This one causes a bug. See the dark triangle on the upper right:

Image

slipher:
The bug demonstrated above affects a BSP surface, so I assume it is something that happened at load time. Probably it would be better to avoid using the "fast" version for anything during load and only use it for hot paths that happen every frame.

illwieckz:
Interesting… So `VectorNormalizeFast()` does something different outside of not returning the length?

slipher:
I would expect VectorNormalizeFast to have weaker accuracy guarantees.

illwieckz:
Ah yes we now use `Q_rsqrt_fast()` instead of `Q_rsqrt()` in it.

illwieckz:
Interesting:

unvanquished_2025-10-01_012500_000

unvanquished_2025-10-01_012513_000

illwieckz:
I orgot I added a faster-less-accurate rsqrt in `VectorNormalizeFast()` so yes we better reserve this for hot paths.

PR #1833, opened 2025-09-30
PoC: multithreaded model CPU computation

illwieckz:
This is a proof of concept of multithreaded model CPU computation, for when using `r_vboModel 0` or when some low-end hardware limitations forces the engine to use that code.

When testing the same scene on the same computer with the same environment:

Before|After
-|-
`91fps`|`388fps`

At first I was looking for processing all models in parallel, running all `Tess_SurfaceIQM()` calls in parallel, with `Tess_SurfaceIQM()` being un modified, but then I discovered we process models like if they were surfaces, and then we not only compute models one after one (something I knew) but we in fact compute and render every model one after one and then it's probably hard to break that sequence.

So I attempted to multithread the `Tess_SurfaceIQM()` call itself. Here is a proof of concept that works.

It has the drawback of requiring OpenMP and then to link against `libgomp` and this adds a dependency.

Also right now I use some GCC builtin to only parallelize that code, as when I use `std::foreach( std::execution::par, … )` it is still sequential even if I link against OpenMP (maybe I'm missing something?).

I also have a branch that builds with `-D_GLIBCXX_PARALLEL` that enforces GCC to automatically replace all parallelizable standard features (like `std::sort`) with parallel variants, but this requires many workarounds because it's not 100% compatible with our code. I wrote the workarounds by the way, but those workarounds use some GCC intrinsics as well. I can push that branch if someone is interested.

I also have another branch that splits the model into chunks, meaning the `for_each` call only iterates (or supposedly multithread) chunks of the model. I was hoping that would help to get better performance because dealing with threads can be time consuming and I was hoping to better control how it multithreads things this way, but curiously OpenMP only spawns one thread when I do this. This alternate implementation may be interesting to us anyway because then all we would need to avoid the dependency on OpenMP would be to write our own dispatch. I can push that branch as well.

Something very ugly in that PoC is the stupid HACK i use to get an iterator that returns the index. That's a PoC anyway.


slipher:
That's an impressive FPS statistic. How many threads did it use?

[Someone says](https://www.reddit.com/r/cpp_questions/comments/11e3k3y/status_of_c17_parallel_algorithms_implementations/) that you have to add in some library (which isn't GOMP) when linking to activate the parallelism in standard C++17.

> also have a branch that builds with `-D_GLIBCXX_PARALLEL` that enforces GCC to automatically replace all parallelizable standard features (like `std::sort`) with parallel variants, but this requires many workarounds because it's not 100% compatible with our code

That sounds unpleasant to deal with. I don't want my code to be suddenly multithreaded when I didn't ask!

Make sure to keep C++14 for all gamelogic build targets. It would be annoying to accidentally use C++17 features developing against the DLL gamelogic,  but then find out they are not allowed on the CI. Also if GOMP is used rather than the C++17 standard stuff, I bet that you don't need C++17 at all.


> Also right now I use some GCC builtin to only parallelize that code,

That's not a builtin, just a namespace for the GOMP-based STL algorithm implementations.





> I also have another branch that splits the model into chunks, meaning the `for_each` call only iterates (or supposedly multithread) chunks of the model. I was hoping that would help to get better performance because dealing with threads can be time consuming and I was hoping to better control how it multithreads things this way, but curiously OpenMP only spawns one thread when I do this.

You mean it uses a small number of large pieces of work, instead of a large number of small ones? In that case, I imagine that it has some fixed estimate of the size of each piece of work, and reasons that if the number is small, it is a waste of time to start up the threads. You would probably need to give it some hint for it to know that each iteration will take a long time. In any case, I guess the thread management must be pretty efficient if you got that much speedup.

illwieckz:
The experiment was a success. I close this and will submit a completed and cleaned-up branch later.

illwieckz:
Also a side note:

> Also if GOMP is used rather than the C++17 standard stuff, I bet that you don't need C++17 at all.

Yes the C++17 thing was from a previous unpublished experiment attempting to use `std::for_each( std::execution::par, … )` with libtbb.

PR #1837, opened 2025-10-01
PoC: multithreaded chunked model CPU computation

illwieckz:
Chunked version of:

- https://github.com/DaemonEngine/Daemon/pull/1833

For unknown reasons, only one thread is spawned.


illwieckz:
Hmm, one drawback of doing it that way, is that unlike what does OpenMP, threads are not reused and then profilers like Orbit list thousands of threads and not only that amount is crazy to list, but profiling is just meh because computed statistics are for each thread separately. It would be cool to be able to reuse those threads.

illwieckz:
Hmm, actually, it seems to work, with my custom thread start. It's just so inefficient that performance drops like if nothing was done. When switching from 1 thread to 2 I see a performance difference. It's just so bad compared to OpenMP.

But now, I don't get why this chunked implementation doesn't work with OpenMP.

illwieckz:
I got it working with OpenMP, I had to use another syntax (which in fact skips my useless vector trick). I now get `438fps` with 16 threads, which is much faster! So yes my idea of controlling the way it is chunked was good!

illwieckz:
With this implementation we are now as fast in the CPU code than when using the GPU code running on CPU with the llvmpipe software renderer. With `LIBGL_ALWAYS_SOFTWARE=1` I get the exact same framerate wether I enable `r_vboVertexSkinning` or disable it. Before, the difference wasn't big, now I see none.

illwieckz:
The experiment was a success. I close this and will submit a completed and cleaned-up branch later.

PR #1842, opened 2025-10-04
WIP: embed media files

illwieckz:
Work in progress: embed media files from `daemon_src.dpkdir`.

I deeply reworked the GLSL include generator to make it reusable, I also renamed `DaemonBuildInfo.cmake` as `DaemonGeneratedSource.cmake` and put everything about generating sources there, and revamped it a lot.

It doesn't work yet because:

- The current code uses strings and then files are truncated on first `0`, but the `sound.wav` file contains some bytes like that and this is valid.
- Files aren't listed, so there is no way the `engine.shader` file gets parsed.


illwieckz:
So it now loads the whole file.

For a funny reason, it works because since the `unvanquished` dpk lists those names, the engine looks for them, and then the code to read them works. Of course we should be able to list them even if they aren't in a dpk already.

illwieckz:
It almost works but…

- This error:
```
Warn: libpng error: PNG file corrupted by ASCII conversion 
Warn: PNG image 'gfx/2d/bigchars.png' has first exception handler called [libpng v.'1.6.50'] 
```
- The files are embedded in all engines (including the server and the tty client).
- I guess the server now requires the client to have that pak loaded just because the engine loaded it.

slipher:
> Now fixed, there was a source beautifier stripping `\r`, now that we not only embed GLSL files, we better drop that.

Don't drop it. Add an option to specify text or binary mode.

illwieckz:
Hmm, right now there is something corrupting something, the engine is slow, and the console has some things printed everyframe, actually it prints some `PK` string so it looks like the console is attempting to print the content of a zip every frame… that's very weird.

illwieckz:
> Don't drop it. Add an option to specify text or binary mode.

I added an option.

illwieckz:
I now should have filtered out the builtin pak name from the list of required paks sent to clients.

I still don't know from where comes that weird bug slowing down the engine and polluting the console…

illwieckz:
Another thing funny is that I tried to create the smallest `null.wav` file possible, but then I get this error:

```
Warn: Unhandled exception (St12length_error): cannot create std::vector larger than max_size() 
```

My “smaller file” is using such format: signed 16-bit mono 8000 Hz, likely one sample only.

illwieckz:
Ok, I had to keep enough samples, something like 0.3ms, the file now does 90 bytes instead of 20KB, here is the generated file:

```c
const unsigned char sound_null_wav[] = {
0x52,0x49,0x46,0x46,0x52,0x00,0x00,0x00,0x57,0x41,0x56,0x45,0x66,0x6d,0x74,
0x20,0x10,0x00,0x00,0x00,0x01,0x00,0x01,0x00,0x40,0x1f,0x00,0x00,0x80,0x3e,
0x00,0x00,0x02,0x00,0x10,0x00,0x64,0x61,0x74,0x61,0x2e,0x00,0x00,0x00,0x00,
0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x01,0x00,
0xfe,0xff,0x02,0x00,0xff,0xff,0xff,0xff,0x02,0x00,0xfe,0xff,0x02,0x00,0xfe,
0xff,0x02,0x00,0xfe,0xff,0x01,0x00,0x00,0x00,0xff,0xff,0x03,0x00,0xfc,0xff,};
```

illwieckz:
I embedded the unifont and NUKED the bigchars.

illwieckz:
I added a mechanism for any part of the engine to register a builtin pak.

Since the builtin pak registration is done in the code using the builtin pak, the FileSystem code doesn't hardcode any builtin pak, and this made possible to only embed the builtin pak in the client.

Two effects:

- we can have different builtin paks per binaries. For example the vulkan renderer may embed some files, while the OpenGL one would embed others…
- we can now embed more thane one builtin pak. For example we can imagine a client builtin pak, a server builtin pak, and a common builtin pak.
- we can have build-variants of builtin paks. For example a test builtin pak with test files can now only be embedded when guarded behind a build option.

illwieckz:
So the last problem remaining is that weird bug of the engine being slow and printing garbage `PK` strings when starting a map.

illwieckz:
I added a special empty `*daemon` pakpath, so one can run the engine without any pakpath by doing:

```sh
./daemon -set fs_basepath '*daemon'
```

This makes possible to drop the `daemon` dpk.

illwieckz:
I now discover that the test-tty client attempts to run the unifont…

illwieckz:
> Hmm, there is something weird with the fact the unifont is loaded by the engine, but when loading the game code it doesn't.

It works with the client, not with the tty client.

illwieckz:
> > Hmm, there is something weird with the fact the unifont is loaded by the engine, but when loading the game code it doesn't.
> 
> It works with the client, not with the tty client.

That's because in the tty client `RE_RegisterFont()` is a stub always returning `nullptr`, we better comment out the code calling it. Now done.

illwieckz:
The “text” processed by `BufferCommandTextInternal()` is just garbage.

illwieckz:
I got a crash and the stacktrace doesn't make sense:

```
#0  Console::Field::AutoComplete(this = 0x555556ef3960 <_ZL11input_field.lto_priv.0>) at /usr/include/c++/13/bits/basic_string.h:1258
#1  CON_Input() at src/engine/sys/con_curses.cpp:508
#2  Com_GetEvent() at src/engine/qcommon/common.cpp:282
#3  Com_EventLoop() at src/engine/qcommon/common.cpp:417
#4  Com_Frame() at src/engine/qcommon/common.cpp:947
#5  Application::ClientApplication::Frame(this = 0x555557690580 <Application::GetApp()::app>) at src/engine/client/ClientApplication.cpp:91
#6  Application::Frame() at src/engine/framework/Application.cpp:87
#7  main(argc = <optimized out>, argv = <optimized out>) at src/engine/framework/System.cpp:1011
```

It crashed on this assert in the stdlib:

```
      _GLIBCXX_NODISCARD _GLIBCXX20_CONSTEXPR
      reference
      operator[](size_type __pos)
      {
        // Allow pos == size() both in C++98 mode, as v3 extension,
	// and in C++11 mode.
	__glibcxx_assert(__pos <= size()); ⬅️
        // In pedantic mode be strict in C++98 mode.
	_GLIBCXX_DEBUG_PEDASSERT(__cplusplus >= 201103L || __pos < size());
	return _M_data()[__pos];
      }
```

slipher:
I am against adding a third type of pak `PAK_BUILTIN`. This means we have to deal with a third type of pak anywhere that does pak handling (that includes logic). That would cause us a lot more trouble than it saves.

illwieckz:
We can do a `*` filtering like the builtin images.

illwieckz:
I'm more concerned by this stupid bug btw, alternate ways to implement the flagging of the builtin pak are trivial.

illwieckz:
The bug acts like if the game was spammed with events from the TTY, but that doesn't make sense and the data is so huge and garbage it is more likely processing the engine memory itself after an overflow or so.

illwieckz:
The bug also occurs when I don't embed the engine media files, with just the code adding empty builtin paks.

illwieckz:
The bug doesn't occur when I embed the files but don't register the builtin paks.

illwieckz:
Actually my `conhistory` is filled with the processed garbage!

illwieckz:
Doing `cat conhistory` is even rewriting the _input_ of my terminal, while the game isn't running! That's crazy…

So maybe something is triggering the printing of weird echo command that in return makes the terminal enters garbage input, which input is itself able to ping pong again to produce more garbage input. It's a life form…

VReaperV:
I've added part of this in #1845 as it turned out I needed part of it. I've add functionality there to add the contents as a `constexpr unsigned char[]` as an option instead of string/unordered_map. That may or may not help with the issue. I've loaded an 18kb binary file with it, works without issues.

illwieckz:
> I've added part of this in #1845 as it turned out I needed part of it. I've add functionality there to add the contents as a `constexpr unsigned char[]` as an option instead of string/unordered_map. That may or may not help with the issue. I've loaded an 18kb binary file with it, works without issues.

Great! Beware that this code is a bit unstable.

I'll recommend keeping the `cmake: make the embedded file generator reusable` commit separate to make it easy to rebase. The one in my branch is now probably ready, but I noticed I modified it since you pulled it, so it would be easier for you to rebase if this commit and your change over it is separate.

It should be noted that the existing GLSL embedding, after that commit, is working without issue.

The issue I'm facing is triggered by the code for builtin pak in VFS.

illwieckz:
ASAN catches nothing. EletricFence catches nothing. I don't know what to do.

illwieckz:
```
] /devmap plat23
…
Debug: SVC_DirectConnect ()  
Client 0 connecting  
Debug: Going from CS_FREE to CS_CONNECTED for illwieckz [granger]  
Debug: CL packet loopback: connectResponse   
]PK
```

And then the console is unresponsive, and sometime filled with garbage (including filling the _input_!).

I have to quit the engine by doing `daemon +quit` in another terminal (and it takes time to process the command).

illwieckz:
If I modify those lines this way, the bug disappears, just by adding the logger:

```
void CGameVM::CGameRocketFrame()
{
	cgClientState_t state = {};
	state.connectPacketCount = clc.connectPacketCount;
	state.connState = cls.state;
Log::Warn("=1===%s===", cls.servername);
	Q_strncpyz( state.servername, cls.servername, sizeof( state.servername ) );
Log::Warn("=2===%s===", cls.updateInfoString);
	Q_strncpyz( state.updateInfoString, cls.updateInfoString, sizeof( state.updateInfoString ) );
Log::Warn("=3===%s===", clc.serverMessage);
	Q_strncpyz( state.messageString, clc.serverMessage, sizeof( state.messageString ) );
	state.clientNum = cl.snap.ps.clientNum;
	this->SendMsg<CGameRocketFrameMsg>(state);
}
```

illwieckz:
Ah no I forgot to do `/devmap`… 🤦🏻

This is what it first prints:

```
Warn: =1===loopback===
Warn: =2======
Warn: =3======
```

…and then the console becomes broken:

```
Warn: =1===loop
Warn: =
Warn: =
Warn: =1===loo
Warn:
Warn:
Warn: =1===lo
Warn:
Warn:
Warn: =1===l
Warn
Warn
Warn: =1===
War
War
Warn: =1==
Wa
Wa
Warn: =1=
W
W
Warn: =1


Warn: =


Warn:


Warn:


Warn


War


Wa


W


```

illwieckz:
I FOUND THE BUG, IT WORKS.

The bug was that I forgot to set `loadedPak.fd = -1` for the loaded builtin pak, meaning later some code was attempting to close the zip for real! I don't know how it did not crashed, and why what looks like some other pak got printed in the terminal (I got `PK` strings…), but that was that!

illwieckz:
```
[FS] Loading builtin pak '*daemon_0.55.5'...
[FS] Loading builtin pak '*daemon-client_0.55.5'...
[FS] Loading pakdir '…/UnvanquishedAssets/build/_pakdir/pkg/unvanquished_test.dpkdir/'...
[FS] Loading pakdir '…/UnvanquishedAssets/build/_pakdir/pkg/tex-common_test.dpkdir/'...
[FS] Loading pak '~/.local/share/unvanquished/base/pkg/res-players_0.55.dpk'...
[FS] Loading pak '~/.local/share/unvanquished/base/pkg/res-legacy_0.55.dpk'...
[FS] Loading pak '~/.local/share/unvanquished/base/pkg/res-weapons_0.55.3.dpk'...
[FS] Loading pak '~/.local/share/unvanquished/base/pkg/res-weapons_0.55.dpk'...
[FS] Loading pak '~/.local/share/unvanquished/base/pkg/res-buildables_0.55.4.dpk'...
[FS] Loading pak '~/.local/share/unvanquished/base/pkg/res-buildables_0.55.dpk'...
[FS] Loading pak '~/.local/share/unvanquished/base/pkg/res-voices_0.55.3.dpk'...
[FS] Loading pak '~/.local/share/unvanquished/base/pkg/res-voices_0.54.dpk'...
[FS] Loading pak '~/.local/share/unvanquished/base/pkg/res-soundtrack_0.54.dpk'...

]/which fonts/unifont.ttf
File "fonts/unifont.ttf" found in "*daemon-client_0.55.5"
```

illwieckz:
One can actually run a custom dll game code without any dpk by doing:

```sh
./daemonded -set fs_basepak '*daemon' -set vm.cgame.type 3 -set vm.sgame.type 3
```

illwieckz:
This code may probably be useless because we have solutions to embed all the files we have if I'm right, but I'm using this as a test bed for #1849:

- https://github.com/DaemonEngine/Daemon/pull/1849

Because this code is much more complex than the GLSL embedding code and also embeds non-text files, so it's very good to test for the robustness of the design and catch pitfalls.

Issue #1236, opened 2024-08-10
Use r_ext namespace for all GL extension test cvars


slipher:
I propose moving all of the cvars that cause the renderer to assume failure when testing for a certain optional OpenGL feature under the `r_ext_` prefix. Currently they are divided between `r_ext_` and `r_arb_`.  As a user I don't care about which OpenGL committee process the feature was proposed under, or whatever the stupid prefix means. There is even an extension that is supposedly available under three different names:  `ARB_framebuffer_object`, `EXT_framebuffer_object`, and `OES_framebuffer_object`.  So the ARB or EXT prefix doesn't seem like a reliable or informative way of grouping these; the feature test cvars are all the same sort of thing. Let's make `ext` mean extension test rather than a part of the verbatim extension identifier.

We should also consider removing the subsequent underscores, to make the names respect the new convention of underscore as namespace.

illwieckz:
## cvar, extension and feature mapping

> I propose moving all of the cvars that cause the renderer to assume failure when testing for a certain optional OpenGL feature under the `r_ext_` prefix.

I disagree as I prefer the cvar to exactly match an extension, no less and no more, and not be about a vague feature that can be achieved with different combinations of various extensions.

## EXT or ARB is part of the extension name and cannot be ignored

> As a user I don't care about which OpenGL committee process the feature was proposed under, or whatever the stupid prefix means.

As a first thing to say, `r_ext_` is not a prefix, the prefix is `r_`, `ext_` is part of the extension name, the underscore after `ext` is not a separator (I elaborate further below).

I also disagree about using `r_ext_` prefix instead of `r_arb_` because this is wrong: `ARB_framebuffer_object`, and `EXT_framebuffer_object` are two different extensions without the same amount of features.

Basically `EXT_framebuffer_object` is an older subset of `ARB_framebuffer_object` and then naming `ARB_framebuffer_object` as `ext_framebuffer_object` would be wrong and confusing.

For details, `ARB_framebuffer_object` also includes `EXT_framebuffer_blit` but `EXT_framebuffer_object` doesn't.

## Common prefix for GL extension cvars

I don't disagree with the idea of having a common prefix for GL extension cvars, but `ext_` and `arb_` are not namespaces, they are part of the full name of the extension.

## Namespaces and conventions

> We should also consider removing the subsequent underscores, to make the names respect the new convention of underscore as namespace.

There is no new conventions of underscore as a name space.

There are two conventions:

[...]

PR #1843, opened 2025-10-04
Migrate remaining joystick-related cvars

slipher:


slipher:
During the cvar migration, I have mostly avoided renaming cvars, to avoid changing interfaces and avoid opening cans of bike-shedding worms. Deviating from that was a mistake! I have now removed the commit `Rename joystick cvars of pattern j_foo_axis to j_fooAxis`.

illwieckz:
Yes we better do dedicated PRs for renaming. I don't really see the benefit of introducing a name change if it's not well thought as an interface redesign may requires it to be.

illwieckz (approved):
LGTM.

VReaperV:
> Something like `joystick.*` or we may even go the `input.joystick.*` way and pave the way for similar naming for other devices (mouse, keyboard…) in the future.

That's even worse, longer cvar names/more typing for no reason.

slipher:
Ref #1236 for cvar namespace wars 😛

Issue #1844, opened 2025-10-04
Customizability of console font


slipher:
In principle, I suppose it should be possible to customize the console font. But the console is supposed to be a place that cannot be modified by the sandboxed gamelogic. So it should *not* be possible for paks to override it in the VFS. If we go ahead and embed `unifont.ttf` in the engine, that could work to prevent pak overriding.

I believe fonts are currently searched in both the VFS and the homepath. So it should be possible to customize the font by setting `cl_consoleFont` and putting your custom font in the homepath (haven't tested though). Customizing it in the VFS as a user is not very useful, since servers enforce the VFS contents.

Proposal:
- Embed the default font in the engine
- NUKE `gfx/2d/bigchars`
- Set `cl_consoleFont` to an empty string by default, meaning use the embedded one
- If the cvar is not empty, search for the font in (only) the pakpath

Or, we could just decide not to care about customization and hard-code the embedded one only.

slipher:
I did this search on the unvanquished.net server which includes all Betaserv archives etc., and confirmed that `map-kravat` variants are the only pak (excluding the console shader itself in files named `engine.shader`, `core.shader`, or `2d.shader`) referencing `bigchars`, and that the shaders defined by `map-kravat` are not referenced anywhere. So it's safe to NUKE `gfx/2d/bigchars` with respect to Tremulous and Unvanquished.

```
slipher@daemon:~$ grep ^/ /home/slipher/pak-duplication.txt | xargs -d'\n' ugrep -z --file-extension=pk3,dpk,shader,particle,trail,bsp --exclude=engine.shader bigchars
/var/www/cdn.unvanquished.net/unvanquished_0.24.0_universal/pkg/unvanquished_0.24.0.pk3{scripts/2d.shader}:gfx/2d/bigchars
/var/www/cdn.unvanquished.net/unvanquished_0.24.0_universal/pkg/unvanquished_0.24.0.pk3{scripts/2d.shader}:             map gfx/2d/bigchars
/var/www/cdn.unvanquished.net/unvanquished_0.50.0/pkg/unvanquished_0.25.0.pk3{scripts/2d.shader}:gfx/2d/bigchars
/var/www/cdn.unvanquished.net/unvanquished_0.50.0/pkg/unvanquished_0.25.0.pk3{scripts/2d.shader}:               map gfx/2d/bigchars
[...]

PR #1847, opened 2025-10-05
Font cleanup

slipher:
Remove unused font-related definitions. NUKE an unused file format with `.dat` extension for pre-rasterized fonts. NUKE the font handle table since we only load one font. Move font stuff out of `q_shared.h`, the header included by everything engine and gamelogic.

### DIFF
--- a/src/engine/RefAPI.h
+++ b/src/engine/RefAPI.h
@@ -51,6 +51,32 @@ struct WindowConfig {
 	int vidWidth, vidHeight; // what the game is using
 };
 
+// font support
+struct glyphInfo_t
+{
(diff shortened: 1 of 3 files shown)


slipher:
> > We only load one font, one time - the console font. No need for a font
> > table caching system to detect duplicate font load requests.
> 
> How the cgame does to load a font?

It has its own copy of the Freetype library. Font rasterization is done entirely inside the cgame.

necessarily-equal (approved):
Code change LGTM, didn't test

necessarily-equal:
(cf https://github.com/DaemonEngine/Daemon/issues/1844)

PR #1850, opened 2025-10-06
smaller null.wav

illwieckz:
Extract from #1842:

- https://github.com/DaemonEngine/Daemon/pull/1842

Our `null.wav` file is stupidly large, 22KB (22094 bytes, to be precise), where this 90 byte file also works!

Our `null.wav` wasn't even truly silent, well you couldn't hear anything, as the remaining noise was very low, but the data wasn't zeroed.

This one is truly zeroed and is the smallest WAV file the engine accepts.

It's technically possible to create smaller WAV files (with less samples), but then the engine either believes it's an empty file or something is wrong, depending on the size.


illwieckz:
Before:

```
mediainfo pkg/daemon_src.dpkdir/sound/null.wav
General
Complete name                            : pkg/daemon_src.dpkdir/sound/null.wav
Format                                   : Wave
Format settings                          : PcmWaveformat
File size                                : 21.6 KiB
Duration                                 : 500 ms
Overall bit rate mode                    : Constant
Overall bit rate                         : 354 kb/s

Audio
Format                                   : PCM
Format settings                          : Little / Signed
Codec ID                                 : 1
Duration                                 : 500 ms
Bit rate mode                            : Constant
Bit rate                                 : 352.8 kb/s
Channel(s)                               : 1 channel
Sampling rate                            : 22.05 kHz
Bit depth                                : 16 bits
Stream size                              : 21.5 KiB (100%)
```

After:

```
$ mediainfo pkg/daemon_src.dpkdir/sound/null.wav
General
Complete name                            : pkg/daemon_src.dpkdir/sound/null.wav
Format                                   : Wave
Format settings                          : PcmWaveformat
File size                                : 90.0 Bytes
Duration                                 : 3 ms
Overall bit rate mode                    : Constant
Overall bit rate                         : 240 kb/s

Audio
Format                                   : PCM
Format settings                          : Little / Signed
Codec ID                                 : 1
Duration                                 : 3 ms
Bit rate mode                            : Constant
Bit rate                                 : 128 kb/s
Channel(s)                               : 1 channel
Sampling rate                            : 8 000 Hz
Bit depth                                : 16 bits
Stream size                              : 46.0 Bytes (51%)
```

illwieckz:
The file also used some useless 22.05 kHz sampling rate, instead of a 8 kHz, with such precision being useless since nothing was to be heard.

Actually, it looks like someone _actually_ recorded the previous `null.wav` file for real! There WAS noise in it!



When reducing the `null.wav` file in Audacity I also noticed that reducing the volume of the sound to the minimum also let some non-zero value remaining, so I had to rely on tricks to actually export a really zeroed WAV file, something I verified in an hex editor.

illwieckz:
Once zipped it can be reduced to 43 bytes. So that would be the size of it in a the release, being embedded in the zipped engine binary, or shipped as a file in the dpk.

illwieckz:
> Seems to be obsoleted by #1854.

Obsoleted by both #1854 and by #1853 because it's not the smallest file anymore anyway 😁

- https://github.com/DaemonEngine/Daemon/pull/1853

But yeah once it's generated in engine, no need for an optimized WAV file whatever the size.

PR #1852, opened 2025-10-06
Use WARN_UNUSED_RESULT in q_math.cpp and deprecate AngleMod 

necessarily-equal:
Companion Unvanquished PR: https://github.com/Unvanquished/Unvanquished/pull/3439 (can be merged independently)


src/engine/qcommon/q_math.cpp
```diff
@@ -190,7 +190,7 @@ signed char ClampChar( int i )
 }
 
 // this isn't a real cheap function to call!
-int DirToByte( vec3_t const dir )
+WARN_UNUSED_RESULT int DirToByte( vec3_t const dir )
```
slipher, line 193: IMHO it would be nicer to put the attribute only on the declaration in the header, and not have to additionally put it on the definition.

PR #1853, opened 2025-10-08
wavcodec: fix WAV file loading

### LINKED ISSUE #1851: Engine rejects small but valid WAV files
The smallest WAV file the engine accepts is 90 bytes, 8KHz i16 mono.

But the smallest valid WAV file I could produce is 46 bytes:

- [null.wav.zip](https://github.com/user-attachments/files/22716498/null.wav.zip)

But the engine rejects it:

```
Warn: Unhandled exception (St12length_error): cannot create std::vector larger than max_size()
```

With some other small sizes it sometime even reports that the file is empty!

That looks bogus.

illwieckz:
Read samples from “first sample byte” to “first sample byte + size” instead of reading from “first sample byte” to “first header byte + size”.

It means we now:
- don't wrongfully truncate the end of the file from the size of the header
- can load files with less samples than the header size

Fixes  #1851:

- https://github.com/DaemonEngine/Daemon/issues/1851


slipher:
Nice! LGTM.

PR #1855, opened 2025-10-10
Add the linearColorMap, linearSpecularMap and linearRgbGen keywords and raw variants

illwieckz:
Add the `linearColorMap`, `linearSpecularMap` and `linearRgbGen` keywords and raw variants

- `linearColorMap`, `linearSpecularMap` and `linearRgbGen` expects
  `tr.worldLinearizeTexture`. They are meant to be used in assets targeting
  the linear pipeline only.
- `rawColorMap`, `rawSpecularMap` and `rawRgbGen` does the same, but there is no
  colorspace meaning. They are meant to be used when porting legacy assets
  that may be used in the naive pipeline to be rendered as they were before
  for historical purposes, while being ported the hacky way for the linear
  pipeline.

---

Previously named: Naive blending compatibility

Former comment:

Work-in-progress attempt to implement compatibility for naive blended assets in the linear pipeline.

It includes a patch to delay the loading of textures from pre-collapsed stages to the end of the stage parsing, after the blend mode is known, because it is required to select the right compatibility code for the given stage.

### DIFF
--- a/src/engine/renderer/tr_shader.cpp
+++ b/src/engine/renderer/tr_shader.cpp
@@ -1491,15 +1491,22 @@ static bool LoadMap( shaderStage_t *stage, const char *buffer, stageType_t type,
 		to use any shader in UI, so we better take care of more than ST_COLORMAP. */
 		if ( ! ( shader.registerFlags & RSF_2D ) )
 		{
+			stage->convertColorFromSRGB = stage->colorspaceBits & LINEAR_RGBGEN ?
+				convertColorFromSRGB_NOP : stage->convertColorFromSRGB;
+
 			switch ( type )
 			{
 				case stageType_t::ST_COLORMAP:
 				case stageType_t::ST_DIFFUSEMAP:
+					imageParams.bits |= stage->colorspaceBits & LINEAR_COLORMAP ? 0 : IF_SRGB;
+					break;
 				case stageType_t::ST_GLOWMAP:
 				case stageType_t::ST_REFLECTIONMAP:
 				case stageType_t::ST_SKYBOXMAP:
-				case stageType_t::ST_SPECULARMAP:
 					imageParams.bits |= IF_SRGB;
+					break;
+				case stageType_t::ST_SPECULARMAP:
+					imageParams.bits |= stage->colorspaceBits & LINEAR_SPECULARMAP ? 0 : IF_SRGB;
 					break;
 				default:
 					break;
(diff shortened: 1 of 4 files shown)


illwieckz:
This now implements `linearColorMap` and `linearColor` keywords, first one doesn't set the sRGB flag on the image, second one doesn't use the sRGB color converter on the rgbgen result.

src/engine/renderer/tr_shader.cpp
```diff
@@ -2641,6 +2659,14 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
 				depthMaskBits = 0;
 			}
 		}
+		else if ( !Q_stricmp( token, "linearColorMap" ) )
```
slipher: I dislike the "linear X" naming. It suggests that the input is in a linear color space and so it should be converted to whatever colorspace is used for further operations. But the point is rather that the resulting color is deliberately made different depending on the blend regime by not converting. So I suggest a name that describes it more as disabling color conversion, like `noTextureColorspaceConversion`.

Also for the case of the color uniform, maybe we could make it more obvious what the modifier is used for by making it part of the rgbgen syntax. Something like `rgbgen noconvert const (0.5 0.5 0.5)`

illwieckz: > It suggests that the input is in a linear color space

That suggestion is correct. That keyword says the image data should be considered linear.

> so it should be converted to whatever colorspace is used for further operations

It's just that we have no reason to convert to another colorspace here, but if we had a reason to, it would do it, yes.

sweet235:
The screenshots described as "Naive blend shaders, naive build, naive renderer" look best. Please use these settings!

sweet235:
> It's up to the mapper. We're adding new features, we don't remove the existing ones.

You have been changing the default settings, and now hundreds of maps look weird. Please be more careful!

illwieckz:
> You have been changing the default settings,

Create an issue for that. You're talking about something else than what is discussed in that thread.

src/engine/renderer/tr_shader.cpp
slipher: As I understand it, currently the rgbgen values are always understood as sRGB values. So when rendering in the sRGB-ish naive regime, the values are used as-is; in the linear blending regime, they must be converted to linear space. Conversely, I expect that if some rgbgen values were to be understood as linear, then they would be used as-is in linear mode, but need conversion with naive blending. Instead, the feature proposed here is to treat the values as being whatever colorspace the renderer is currently running with, as a dirty hack to make some colors brighter in linear mode. With the proposed code, colors are NOT treated as linear when using the naive blending regime, so using a keyword named `linear` is wrong/misleading.

illwieckz: Those proposed keywords have no meaning when rendering a non colorspace-aware map to begin with.

illwieckz: There is no naive blending use case for those keywords.

illwieckz: Yes, but it works by being unused in naive mode.

PR #1858, opened 2025-10-11
Use CMake's precompiled headers - fixes PCH for Mac and Ninja

slipher:
`target_precompiled_headers`from CMake 3.16 is a drop-in replacement for our existing PCH implementation. It fixes them for Mac and makes them work with Ninja for the first time. I tested it on Linux + GCC, AppleClang, MSVC, and MSYS2 GCC.


illwieckz (approved):
Great! LGTM.

PR #1861, opened 2025-10-13
Fix mouse accel + migrate cvars

slipher:
Depends on #1860

Fix a bug where if mouse accel is enabled, the `sensitivity` is applied twice. So if sensitivity is 3, it suddenly behaves like sensitivity is 9 when `cl_mouseAccel` has a nonzero value.

Also migrate mouse accel cvars.


illwieckz:
So people would have to change their sensitivity configuration?

slipher:
> So people would have to change their sensitivity configuration?

Only if they were using mouse accel. We don't have an option in the UI for this, so it should be very rare.

PR #1866, opened 2025-10-17
NUKE alphaAddOne bit in color modulate

slipher:
Get rid of the `alphaAddOne` bit in color modulate uniforms.  There are no differences in my screenshot test suite as a result of removing this. This alpha := 1 thing only existed in the `generic` shader; its absence in lightMapping is further evidence suggesting it is not needed.

Note: this would open opportunities for further improvements/cleanup/simplification, due to e.g. the now unused spot in the float color modulate array, but I'm about to NUKE a large part of this code on the 0.56 branch so let's not touch it.

Also fix wrong stage variant description strings in `/listMaterialSystemStages`.


VReaperV:
The equivalent code is still done in `Tess_ComputeColor()`.

PR #1870, opened 2025-10-19
Fix broken shader that prevents game start

necessarily-equal:
Fix broken shader compilation introduced in a50b9ba160abe38ac8f8d9bba7c3de8ade9a4f6b.

I believe this PR isn't subject to the 24h rules because the current master is obviously broken. Unless someone objects I would then merge after the first LGTM.

### DIFF
--- a/src/engine/renderer/glsl_source/generic_fp.glsl
+++ b/src/engine/renderer/glsl_source/generic_fp.glsl
@@ -83,7 +83,7 @@ void	main()
 	outputColor = color;
 
 // Debugging.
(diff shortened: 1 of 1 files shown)


src/engine/renderer/glsl_source/generic_fp.glsl
```diff
@@ -77,13 +77,13 @@ void	main()
 #endif
 
 	color *= var_Color;
-
```
VReaperV: Why is this change here?

slipher:
Thanks!

PR #1871, opened 2025-10-19
IPC perf improvements by freem, version 2

necessarily-equal:
This is a fixed version of https://github.com/DaemonEngine/Daemon/pull/1632, see https://github.com/DaemonEngine/Daemon/pull/1632#issuecomment-3419522985 for my own review.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -3441,23 +3441,23 @@ void GLimp_LogComment_( std::string comment );
 		virtual const RenderCommand *ExecuteSelf() const = 0;
 	};
 
-	struct SetColorCommand : public RenderCommand {
+	struct SetColorCommand final : public RenderCommand {
 		const RenderCommand *ExecuteSelf() const override;
 
 		Color::Color color;
 	};
-	struct SetColorGradingCommand : public RenderCommand {
+	struct SetColorGradingCommand final : public RenderCommand {
 		const RenderCommand *ExecuteSelf() const override;
 
 		image_t *image;
 		int     slot;
 	};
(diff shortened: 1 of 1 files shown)


necessarily-equal:
I think part of our disagreement might be because I measured it in DLL mode and you didn't?


With the following I still measure thousands of calls. 
```patch
diff --git a/src/common/IPC/Primitives.cpp b/src/common/IPC/Primitives.cpp
index cbf9508d7..aae5cff2a 100644
--- a/src/common/IPC/Primitives.cpp
+++ b/src/common/IPC/Primitives.cpp
@@ -233,6 +233,8 @@ static void InternalSendMsg(Sys::OSHandle handle, bool more, const FileDesc* han
                Sys::Drop("IPC: Failed to send message: %s", error);
        }
 #else
+       static int count = 0;
+       printf("used %i times\n", ++count);
        size_t descBytes = 0;
        static std::unique_ptr<unsigned char[]> descBuffer;
        static size_t descBufferSize = 0;
```

Example output while loading Antares:

    used 6850 times
    used 65777 times
    used 43370 times

Of course when `__native_client__` is defined, this part of the code is compiled out, so that specific commit you mention (`improve SendMsg performances by avoiding malloc calls`) only improves performance in DLL mode.

slipher:
I mean that `descBuffer` is populated with a nonzero number of items only O(1) times.

slipher:
`InternalRecvMsg reserves before pushing` should be dropped for the same reason of handles rarely being used. It turns a vector that only allocates 0.01% of the time into a vector that always allocates.

Other commits LGTM.

PR #1872, opened 2025-11-01
Add StringRef pointer + length constructor

slipher:


illwieckz (dismissed):
You have to add:

```c++
#include "Assert.h"
```

illwieckz:
> You have to add:
> 
> ```c++
> #include "Assert.h"
> ```

Or maybe move that include in `Common.h` before `#include "String.h"`.

illwieckz:
Hu, no now on MinGW it says `String.h` should be included before `Assrert.h`.

illwieckz:
`DAEMON_ASSERT_EQ` in `Assert.h` uses `Str::Format()` from `String.h`,
but `Str::BasicStringRef` from String.h` now requires `DAEMON_ASSERT_EQ` from `Assert.h`.

illwieckz:
> Maybe due to some .cpp file that doesn't include `Common.h` first

Yes, the generated embedded files. Actually I'm a bit annoyed to add the whole `src/common` folder has a dependency for the code generation (same critique for using `Str::StringRef` there to begin with). My small struct is very portable and doesn't pull a whole folder of header files, so I prefer that. I can reuse the cast trick to avoid exposing an `unsigned char` array though.

PR #1873, opened 2025-11-10
cmake: rewrite a bit the header of the embedded file source generator

illwieckz:
cmake: rewrite a bit the header of the embedded file source generator

- add a comment saying the file is generated (like other generated files)
- use explicit '\t' in the CMake string

That's a small fixup above:

- https://github.com/DaemonEngine/Daemon/pull/1849

It just makes things nicer.


slipher:
CI failure is relevant.

illwieckz:
Fixed the type thing.

slipher:
LGTM

PR #1875, opened 2025-11-26
Some improvement on the deps building

illwieckz:
- Do not cross-compile pkgconfig, it is part of the cross-compilation toolchain;
- Reword a comment because we don't build SDL with pkg-config anymore;
- Rework the macos-amd64 setup a bit;
- Do not build OpenAL utils.

Extracted from:

- https://github.com/DaemonEngine/Daemon/pull/1823

### DIFF
--- a/external_deps/build.sh
+++ b/external_deps/build.sh
@@ -770,6 +770,7 @@ build_openal() {
 		cd "${dir_name}"
 
 		cmake_build \
+			-DALSOFT_UTILS=OFF \
 			"${openal_cmake_args[@]}"
 		;;
 	esac


external_deps/build.sh
```diff
@@ -1421,7 +1421,7 @@ setup_macos-amd64-default() {
 	MACOS_ARCH=x86_64
 	# OpenAL requires 10.14.
 	export MACOSX_DEPLOYMENT_TARGET=10.14 # works with CMake
-	common_setup macos x86_64-apple-darwin11
+	common_setup macos "x86_64-apple-macosx${MACOSX_DEPLOYMENT_TARGET}"
```
slipher: This looks wrong. I've never managed to find any documentation for autotools platform strings, but for me the autodetection reports `darwin24.6.0` for the last component. This tracks "Kernel version" in the system's about screen.

illwieckz: There is explicit code for those three syntaxes in LLVM:

- Matching `darwin*` and `macos*`:
  * https://github.com/llvm/llvm-project/blob/5dd2b06d60d3eb9b07c7513358ad8b04386f79bc/llvm/lib/TargetParser/Triple.cpp#L702
  * https://github.com/llvm/llvm-project/blob/5dd2b06d60d3eb9b07c7513358ad8b04386f79bc/llvm/lib/TargetParser/Triple.cpp#L710
  
- Parsing Darwin and macOS versions:
  * https://github.com/llvm/llvm-project/blob/5dd2b06d60d3eb9b07c7513358ad8b04386f79bc/llvm/lib/TargetParser/Triple.cpp#L1474-L1499

We can put either Darwin version or macOS version, and in fact it's better for us to use the macOS version since we already know it and then we don't have to search for the related Darwin version.

illwieckz: I modified the string from `macosx<VERSION>` to `macos<VERSION>` because we're in the future.

illwieckz: I forgot to say it's not an autotools thing, but a compiler thing, autotools is very likely just reporting the output of `cc -dumpmachine` when we let it guess.

slipher: That looks OK then, LGTM.

Autotools does parse the strings in the `config.sub` script. It rejects some parts if they don't match any pattern and replaces some strings with others.

PR #1880, opened 2025-12-26
Fix console crash and console key input

VReaperV:
- Fixes a crash in auto-complete (start typing a command, move cursor back to start of the line, press `tab`)
- Fixes console key sometimes being caught in the console input sometimes. This one's a bit difficult to reproduce, I've mainly seen it happen when tabbing out before/during a map change with console open, then tabbing back in after the map loads and opening the console again. Also reproducible by setting input method to UK English in Windows 11, though it has a different underlying cause

### DIFF
--- a/src/engine/client/key_identification.cpp
+++ b/src/engine/client/key_identification.cpp
@@ -186,6 +186,32 @@ Key StringToKey(Str::StringRef str)
     return Key::NONE;
 }
 
+std::string KeyToStringUnprefixed( Key key ) {
+    if ( key.kind() == Key::Kind::KEYNUM ) {
+        return KeynumToString( key.AsKeynum() );
+    }
+
+    if ( key.kind() == Key::Kind::UNICODE_CHAR ) {
+        return CharToString( key.AsCharacter() );
+    }
+
+    if ( key.kind() == Key::Kind::SCANCODE ) {
+        int sc = key.AsScancode();
+        if ( char c = ScancodeToAscii( sc ) ) {
+            return CharToString( c );
+        }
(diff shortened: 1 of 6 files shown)


src/engine/client/key_identification.cpp
```diff
@@ -186,6 +186,32 @@ Key StringToKey(Str::StringRef str)
     return Key::NONE;
 }
 
+std::string KeyToStringUnprefixed( Key key ) {
```
slipher: I don't understand, what is the point of forming a string with words like `ESCAPE`, `0x63` etc.?

src/engine/sys/sdl_input.cpp
```diff
@@ -1113,6 +1113,24 @@ static void IN_ProcessEvents( bool dropInput )
 				{
 					std::string text = e.text.text;
 
+					/* We can get a console key as a text event after we've opened the console from a key up/down event,
+					*  in a separate execution of IN_ProcessEvents()
+					*  The key character can also be anywhere in the text
+					*  This might be an SDL bug */
+					text.erase(
```
slipher: This would break the feature where you can hold ALT to type a character that would otherwise be a console key (works on some Linux environments)

slipher: That is where the ALT exemption is initially applied. But IIUC this new third level of console key filtering has no condition to check that, so such characters would be filtered out.

illwieckz:
What's the status on it?

I also seen someone in game saying there was a bug with the `²` key, which is the Azerty ` key.

PR #1882, opened 2026-01-02
Fix screenshots with read-only depth buffer

slipher:
When the read-only depth buffer is used (if r_readonlyDepthBuffer is 2, or if it's 1 but the driver lacks certain extensions), the screenshots were done from the 3d rendering buffer instead of the final framebuffer. This caused screenshots to lack tonemapping, color grading and 2D rendering. Fix it by resetting GL_READ_FRAMEBUFFER to the default value after it is used.


VReaperV (approved):
Don't have any affected hardware, but code LGTM.

slipher:
I changed it to reset`GL_READ_FRAMEBUFFER` when taking a screenshot command as in #1480, instead of in `RB_RenderView`, to avoid redundancy.

Issue #1798, opened 2025-09-12
Material system: marks are drawn after fog


slipher:
With material system, marks such as bullet marks or blob shadows are rendered after fog.

I believe this is because most gradations of shader sorting are ignored; the `MaterialPack` defines a range of sort values that may be drawn in any order. 

Material system on vs. off:

![Image](https://github.com/user-attachments/assets/40a75229-6047-4b4c-8928-4889a83b1361)

![Image](https://github.com/user-attachments/assets/18c09dfc-3247-48d9-b263-0e55b2149cf6)

PR #1888, opened 2026-02-03
Render fogs with one draw call per fog instead of re-drawing every surface inside

slipher:
Adapt the global fog shader to use a vertex position so it can draw finite-sized fogs. Right now it is only implemented for fogs that the viewer is inside of, but this is enough to fix the artifacts with MSAA as in https://github.com/DaemonEngine/Daemon/pull/1480#issuecomment-3693940656. Later I will update the PR with commits to draw fogs the viewer is outside of this way as well.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -2825,7 +2825,7 @@ inline bool checkGLErrors()
 	void           R_LocalNormalToWorld( const vec3_t local, vec3_t world );
 	void           R_LocalPointToWorld( const vec3_t local, vec3_t world );
 
-	cullResult_t   R_CullBox( vec3_t worldBounds[ 2 ] );
+	cullResult_t   R_CullBox( const vec3_t worldBounds[ 2 ], int lastPlane = FRUSTUM_NEAR );
 	cullResult_t   R_CullLocalBox( vec3_t bounds[ 2 ] );
 	cullResult_t   R_CullLocalPointAndRadius( vec3_t origin, float radius );
 	cullResult_t   R_CullPointAndRadius( vec3_t origin, float radius );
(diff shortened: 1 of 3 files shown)


slipher:
This is now ready. But stacked on #1889.

Functional changes to what is rendered:
- Fixes seams at the fog surface plane when using MSAA and located inside the fog.
- Fixes some other minor artifacts on models around the fog plane.
- Fixes #1798
- Fog volumes can now be open on any side. The old rule is that fogs must be fully enclosed by opaque surfaces on all sides except one (usually +z), so that the fog can be treated as a half-space in calculations. If you viewed the fog from "under" the fog plane but outside the fog, distance in fog was not calculated correctly. Now it is fine to view fogs from any side, although there is still a barely-perceptible artifact when passing through a non-surface side of the fog.

There are two variants of the `fog` shader, one for fogs that the viewer is inside of and one for fogs the viewer is outside of. When the viewer is inside, we render the inside faces of the fog brush with depth testing disabled. To find the distance in fog, we just find the minimum of the distance to the fragment and the distance to the point indicated by the depth buffer. When the viewer is outside, the outside faces of the fog brushes are rendered. To find the distance we trace against the other 5 faces of the fog brush as well as checking the depth buffer.

For now all fogs are rendered outside of the material system. Integrating them would be a bit tricky since you have two variants of the shader requiring different GL state, which must be selectively enabled based on the view position.

As for performance, the new fog is sadly slower on an Intel UHD 630 I tested. Turns out this is ~entirely due to reading from the depth buffer. I checked this by editing the GLSL. But also it can be seen that on master, with the ET `railgun` map which uses the global fog shader (which is very simple but reads from the depth buffer), disabling fog increases FPS by 20%. Anyway I also tried to test with an Nvidia card, but fog didn't make enough difference to FPS to be clearly observed in any of those tests.

Issue #1354, opened 2024-10-14
Intel UHD texture distortion II: Revenge of the Vertex


slipher:
Observe the floor near the bottom left:
![unvanquished_2024-10-13_233906_000](https://github.com/user-attachments/assets/9363a87c-e88e-4c60-acd5-ee33e5876434)

[Intel UHD Texture distortion](https://github.com/DaemonEngine/Daemon/issues/909) is back. But in this spot, the bug happens __only if__ `workaround.glHardware.intel.useFirstProvokinVertex` [sic] is enabled. This location seems to require a nonzero roll angle, so you need to be 3rd-person spectating or maybe wallwalking.

Instructions:
```
devmap parpax
team h
class ckit
cg_thirdPerson 1
cg_thirdPersonViewAngle 80
setviewpos 1569 -2753 176 -105 23
+movedown
```

If (and only if) I turn the workaround off, I can still reproduce the bug with [Kai's viewpos for Thunder](https://github.com/DaemonEngine/Daemon/issues/909#issuecomment-1832091540).

I produced this bug on the 0.55 Release Candidate 1.

VReaperV:
Another `viewpos`, reported by freem (map `freeway`):
![image](https://github.com/user-attachments/assets/f789edc4-e174-491e-bef4-1c15dc3055ce)

This is reproducible with 0.55.2 with `workaround.glHardware.intel.useFirstProvokinVertex` enabled.

illwieckz:
So I guess `6197 -3216 -235`.

PR #1890, opened 2026-02-08
Detect Moore Threads hardware and disable ARB_texture_barrier

illwieckz:
~~Detect Moore Threads vendor and disable SSAO on buggy Moore Threads driver.~~
Detect Moore Threads hardware and disable ARB_texture_barrier on buggy Moore Threads driver.

Also reword things around.

Moore Threads is a Chinese vendor of gaming GPUs similar to AMD/Intel/Nvidia targeted at the gaming PC market, with drivers for Linux and Windows.

I'll open an issue for the SSAO bug so we can track the status of it and see if I can report it to Moore Threads.

For now this makes the `high` and `ultra` preset working out of the box.

There is also a geometry bug similar to the Intel one (#909, #1354), that only happens when the camera is perfectly aligned on some axis, but this is almost invisible when playing and doesn't make the game look annoying, unlike the SSAO bug. I'll also make an issue for that.

This was tested on MTT S80:

- https://www.techpowerup.com/gpu-specs/mtt-s80.c4259
- https://en.mthreads.com/product/S80

_Edit:_ the issue is now there:

- https://github.com/DaemonEngine/Daemon/issues/1891

### DIFF
--- a/src/engine/sys/sdl_glimp.cpp
+++ b/src/engine/sys/sdl_glimp.cpp
@@ -185,6 +185,10 @@ static Cvar::Cvar<bool> workaround_glHardware_intel_useFirstProvokinVertex(
 	"workaround.glHardware.intel.useFirstProvokinVertex",
 	"Use first provoking vertex on Intel hardware supporting ARB_provoking_vertex",
 	Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_glHardware_mthreads_disableTextureBarrier(
+	"workaround.glHardware.mthreads.disableTextureBarrier",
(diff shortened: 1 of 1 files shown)


slipher:
Does the SSAO work if you set `r_readonlyDepthBuffer 2`?

slipher:
Someone says that texture barriers ["don't make sense"](https://github.com/gpuweb/gpuweb/issues/439) for a tiled GPU architecture in general. So for the Moore Threads case it seems good to declare the texture barrier extension unavailable. This is distinct from my AMD GPU which has the right architecture to implement, and sometimes it works, but it fails in a specific configuration.

PR #1896, opened 2026-02-13
Disable rim lighting (r_rimLighting) by default

slipher:
"Rim lighting" purports to make the edges of models stand out. However in my tests, I don't see a very strong correlation between the parts which are lit up and the edges of models. For more geometrically complex models such as the Barricade, it sometimes randomly lightens the model all over the place, which seems bad.

No rim lighting:

unvanquished-atcshd-under-fog-barricade
unvanquished-plat23-dlight-psaw1

With rim lighting:

unvanquished-atcshd-under-fog-barricade

unvanquished-plat23-dlight-psaw1

If the idea is accepted, I will also make a PR for the graphics presets.


slipher:
I will try an experiment to make it easier to evaluate renderer PRs like this: I will upload the screenshot comparison tool and the screenshot data set I'm using to unvanquished.net for reviewers to browse. Of course the downside is that the thread is not self-contained; I will probably delete the screenshots later.

Navigate using the arrow keys: https://users.unvanquished.net/~slipher/screenshot-compare/1896/viewer.html. Note that fog is disabled since I had several otherwise useful test scenes where the model was obscured by fog.

PR #1897, opened 2026-02-16
Fix lighting for movable BSP entities

slipher:
BSP model surfaces were wrongly flagged with the `bspSurface` bit which only works for fixed-location world surfaces. This flag was causing the code to skip setting up the model matrix for the lightMapping shader, meaning normals and dynamic light positions were not correct.

So stop flagging BSP models as `bspSurface`. Also skip the other things in R_AddWorldSurface, besides R_AddDrawSurf, because they are useless or redundant.

Fixes https://github.com/Unvanquished/Unvanquished/issues/3474.

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1489,7 +1489,7 @@ enum
 		surfaceType_t *surface; // any of surface*_t
 		shader_t      *shader;
 		uint64_t      sort;
-		bool          bspSurface;
+		bool          bspSurface; // only for the stationary part of the world, *not* BSP models
 		int fog;
 		int portalNum = -1;
 
--- a/src/engine/renderer/tr_world.cpp
+++ b/src/engine/renderer/tr_world.cpp
@@ -194,7 +194,8 @@ void R_AddBSPModelSurfaces( trRefEntity_t *ent )
 
 	for ( i = 0; i < bspModel->numSurfaces; i++ )
 	{
-		R_AddWorldSurface( bspModel->firstSurface + i, fogNum, -1, FRUSTUM_CLIPALL );
+		bspSurface_t *surf = bspModel->firstSurface + i;
+		R_AddDrawSurf( surf->data, surf->shader, surf->lightmapNum, fogNum );
 	}
 }
 


illwieckz:
The Vega map has many doors and can be used to test this. Among those doors is a vertical door with a known visible shadow moving with the door (that's the kind of door I would like to have an option to force the lightgrid on it).

illwieckz:
> Lighting the models with the light grid seems like a correct default to me

It is a good default, but no maps were built for that.

> since in general they are intended to move, meaning the fixed lightmap is wrong.

Indeed, but grid lighting looks less good and for various use cases, like doors opening by retracting inside walls left and right, lightmap is good enough. And I don't know if it's a bug in gridlight code, but lightgrid looks far worse on them.

For some reasons, lightgrid doors can look very different according from the side we look at them:

unvanquished_2026-02-16_112913_000

unvanquished_2026-02-16_112923_000

Another door:

unvanquished_2026-02-16_112950_000

unvanquished_2026-02-16_112954_000

But the main argument is that all maps were done with the fact those models were lit with the lightmap.

Here is a good example, the original metro map.

The train is lit with lightmap, the mapper created a special room to store the train at light backing time to lit it evenly. The mapper added a fog in the tunnel which blends with the train and emulate the fact the train comes to light. But the mapper also did a workaround for lighting the entrance of the tunnel by adding a point light at the entrance of it.

When we lit the train with the lightgrid, we can see the train being lit by the workaround light:

unvanquished_2026-02-17_001159_000

Of course if this map was made brand new today, we would not use such point light workaround and we would use light grid on the model, using the actual light grid to actually render the fact the train comes from obscurity to light. The metro map is one of the reasons I want the feature of lighting BSP models with the grid light, and in my WIP revamp of the map, I deleted that workaround point light.

But I consider too bold to change the default.

slipher:
> For some reasons, lightgrid doors can look very different according from the side we look at them:

Yeah... the light grid equations used in our code are totally idiotic and wrong. I've been reading the `TraceGrid` function from q3map2 to understand how the light grid data is intended to be used. I'm working on rewriting the code right now, so let's see if I can get something that comes out closer to the lightmap result.

slipher:
With a light grid rewrite things look better, but not really good enough. Closing in favor of #1903

PR #1898, opened 2026-02-16
Light grid-related cleanup

slipher:
A few light grid-related cleanups/optimizations that do not change the functionality.


slipher:
Added some more commits. I fixed a funny bug with `r_showLightGrid`. Due to this visualization, I thought for a long time that it is possible for the ambient and the directed light colors to be different. In fact, they are always the same color, just with different brightness levels. (There are distinct colors in the BSP, but in Daemon's post-processing they are just averaged...) But due to buggy float-to-byte color conversion, `r_showLightGrid` sometimes rendered different colors, e.g. blue directed and red ambient.

VReaperV:
`Don't store alpha channel for lightGrid2 image` doesn't really do anything as drivers generally don't implement RGB formats and will instead convert it to RGBA.

Issue #1892, opened 2026-02-08
Moore Threads: missing geometry when the camera is perfectly aligned on axises


illwieckz:
When the camera is perfectly aligned with axises, some triangles are not rendered on the MTT S80:

![Image](https://github.com/user-attachments/assets/7fa9da19-1ca3-41a8-8a8b-25734a68eee4)

How to reproduce:

- `setviewpos -2256 2296 200 -180 0` on `plat23`

The triangle is entirely missing, what we see being drawn there is the remain of the previous frame, i.e. we see the void.

It is almost invisible when playing because when moving the void is filled with the surrounding texture that was drawn in the previous frame before moving, and it would only do that at the exact moment the camera is perfectly aligned on axises

Here we see the spawning halo effect.

If we do `setviewpos -2256 2296 200 -180 0` as a spectator in `plat23`, we see the remnant of the default spectator scene in the void instead:

![Image](https://github.com/user-attachments/assets/640008b5-4737-4d78-859a-627ec3f735e1)

If we do `setviewpos -2256 2296 200 -181 0` the bug disappears: 

![Image](https://github.com/user-attachments/assets/52587a28-7158-4af8-8002-e568d8d56d4b)

If we do `setviewpos -2256 2296 200 -180 1` the bug disappears as well: 

![Image](https://github.com/user-attachments/assets/8aacedaf-bd18-4d5e-8eb3-92054c9e23cd)

So the bug doesn't prevent to play.

slipher:
Great find!

Slight correction to my previous comment: the Apple driver didn't have buggy texture barriers; it simply didn't provide the extension (https://github.com/DaemonEngine/Daemon/issues/1783).

illwieckz:
So the bug with texture barriers on the Moore Threads would be like the Nvidia 340 driver for GT210 about texture gather: the driver would lie about the extension being supported?

illwieckz:
For knowledge, this is how the PowerVR Vulkan driver is reporting itself on my RISC-V board:

```
$ vulkaninfo --summary | grep -E '^GPU0:' -A12

GPU0:
	apiVersion         = 1.3.225
	driverVersion      = 1.525.317
	vendorID           = 0x1010
	deviceID           = 0x36054182
	deviceType         = PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU
	deviceName         = PowerVR B-Series BXE-4-32
	driverID           = DRIVER_ID_IMAGINATION_PROPRIETARY
[...]

PR #1902, opened 2026-02-18
GL4ES compatibility

illwieckz:
I flag it as draft to make it obvious I'm not intending to merge this for the 0.56 release.

I'm just pushing it because it's somewhat usable and as Linus Torvalds said once:

> Only wimps use tape backup: real men just upload their important stuff on ftp, and let the rest of the world mirror it ;)

This implements some compatibility code path making possible to run the Dæmon engine and the Unvanquished game on OpenGL ES hardware using the [GL4ES](https://github.com/ptitSeb/gl4es/) translation layer.

In fact this basically implements some compatibility code path for GL ES 2 (what currently targets GL4ES) for when GL4ES isn't able to translate OpenGL to GL ES because of GL ES limitation.

The purpose of this is to allow me to test the engine on more hardware. I used this to confirm the bug we observe on PowerVR-derivated hardware like Apple AGX/Moore Threads is indeed reproducible on PowerVR:

- https://github.com/DaemonEngine/Daemon/issues/1892

The only Imagination-branded PowerVR device I own is soldered on a RISC-V board, and while there is a Vulkan Mesa driver being on development for it (purposed to be a backend for the Mesa zink OpenGL translation layer), the RISC-V architecture of the said board is very new and then the whole Linux support is experimental and so getting the game running on GL4ES is far less work than doing platform-enablement at the whole distro level. So that allows me to use the proprietary OpenGL ES Imagination driver for PowerVR which is bundled with the distro provided for the board (the Proprietary Imagination Vulkan driver isn't featureful enough to run Zink). I even don't have to rebuild neither the kernel, neither Mesa!

This also brings more rewrite of existing compatibility code in a better way.

I started to implement ways to detect the underlying hardware when running API translators, this may be extended to things like VirGL, Zink or Parallels in the future.

GL4ES actually added such wanted feature on my request:

- https://github.com/ptitSeb/gl4es/issues/502

I have more ideas for some rework on the driver/hardware detection (especially moving things like `TRANSLATOR` to a different variable than `HARDWARE`).

So this doesn't only benefit GL4ES and PowerVR.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -585,6 +585,12 @@ R"(vec4 unpackUnorm4x8( uint value )
 		str += "#define atomicCounterAndARB atomicCounterAnd\n";
 	}
 
+	if ( glConfig.mat3x2Available ) 	{
+		str += "#define textureMatrix mat3x2\n";
+	} else {
+		str += "#define textureMatrix mat3\n";
+	}
+
 	str +=
 R"(#endif // GENERATED_COMPAT_HEADER
 
@@ -1839,7 +1845,7 @@ std::string GLShaderManager::ShaderPostProcess( GLShader *shader, const std::str
 	                           "	uvec2 u_GlowMap;\n"
 	                           "};\n\n"
 	                           + texBuf +
-		                       "#define u_TextureMatrix mat3x2( texData[( baseInstance >> 12 ) & 0xFFF].u_TextureMatrix.xy, texData[( baseInstance >> 12 ) & 0xFFF].u_TextureMatrix.zw, texData[( baseInstance >> 12 ) & 0xFFF].u_TextureMatrix2 )\n"
+		                       "#define u_TextureMatrix textureMatrix( texData[( baseInstance >> 12 ) & 0xFFF].u_TextureMatrix.xy, texData[( baseInstance >> 12 ) & 0xFFF].u_TextureMatrix.zw, texData[( baseInstance >> 12 ) & 0xFFF].u_TextureMatrix2 )\n"
 		                       "#define u_DiffuseMap_initial texData[( baseInstance >> 12 ) & 0xFFF].u_DiffuseMap\n"
 		                       "#define u_NormalMap_initial texData[( baseInstance >> 12 ) & 0xFFF].u_NormalMap\n"
 		                       "#define u_HeightMap_initial texData[( baseInstance >> 12 ) & 0xFFF].u_HeightMap\n"
@@ -2586,7 +2592,7 @@ GLShader_generic::GLShader_generic() :
 		false, "generic", "generic" ),
 	u_ColorMap( this ),
 	u_DepthMap( this ),
-	u_TextureMatrix( this ),
+	u_TextureMatrix_Dispatch( this ),
 	u_ViewOrigin( this ),
 	u_AlphaThreshold( this ),
 	u_ModelMatrix( this ),
@@ -2621,7 +2627,7 @@ GLShader_genericMaterial::GLShader_genericMaterial() :
 		true, "generic", "generic" ),
 	u_ColorMap( this ),
 	u_DepthMap( this ),
-	u_TextureMatrix( this ),
+	u_TextureMatrix_Dispatch( this ),
 	u_ViewOrigin( this ),
 	u_AlphaThreshold( this ),
 	u_ModelMatrix( this ),
@@ -2657,7 +2663,7 @@ GLShader_lightMapping::GLShader_lightMapping() :
 	u_LightGrid1( this ),
 	u_LightGrid2( this ),
 	u_LightTiles( this ),
-	u_TextureMatrix( this ),
+	u_TextureMatrix_Dispatch( this ),
 	u_SpecularExponent( this ),
 	u_ColorModulateColorGen_Dispatch( this ),
 	u_Color_Dispatch( this ),
@@ -2723,7 +2729,7 @@ GLShader_lightMappingMaterial::GLShader_lightMappingMaterial() :
 	u_LightGrid1( this ),
 	u_LightGrid2( this ),
 	u_LightTiles( this ),
-	u_TextureMatrix( this ),
+	u_TextureMatrix_Dispatch( this ),
 	u_SpecularExponent( this ),
 	u_ColorModulateColorGen_Uint( this ),
 	u_Color_Uint( this ),
@@ -2761,7 +2767,7 @@ GLShader_reflection::GLShader_reflection():
 	u_ColorMapCube( this ),
 	u_NormalMap( this ),
 	u_HeightMap( this ),
-	u_TextureMatrix( this ),
+	u_TextureMatrix_Dispatch( this ),
 	u_ViewOrigin( this ),
 	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
@@ -2792,7 +2798,7 @@ GLShader_reflectionMaterial::GLShader_reflectionMaterial() :
 	u_ColorMapCube( this ),
 	u_NormalMap( this ),
 	u_HeightMap( this ),
-	u_TextureMatrix( this ),
+	u_TextureMatrix_Dispatch( this ),
 	u_ViewOrigin( this ),
 	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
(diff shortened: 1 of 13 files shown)


src/engine/renderer/gl_shader.cpp
```diff
@@ -947,26 +1018,102 @@ void GLShaderManager::BuildShader( ShaderDescriptor* descriptor ) {
...
+
+		for ( auto &pair : concatenatedDescriptor )
+		{
+			ShaderDescriptor &concatenated = pair.second;
+
+			if ( concatenated.type != 0 )
+			{
+				Log::Debug( "Building concatenated %s program.", shaderTypeName[ concatenated.type ] );
+				// WIP
+				// Log::Warn( "Concatenated source:\n<<<\n%s\n>>>", concatenated.shaderSource );
```
illwieckz: TODO: delete this

src/engine/sys/sdl_glimp.cpp
```diff
@@ -137,6 +149,22 @@ static Cvar::Cvar<bool> workaround_glDriver_amd_oglp_disableBindlessTexture(
...
+static Cvar::Cvar<bool> workaround_glDriver_gl4es_disableIncrementalShaderCompilation(
+	"workaround.glDriver.gl4es.disableIncrementalShaderCompilation",
+	"Disable incremental shader compilation on GL4ES",
+	Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_glDriver_gl4es_disableMat3x2(
+	"workaround.glDriver.gl4es.disableMat3x2",
+	"Disable mat3x2 GLSL support on GL4ES",
+	Cvar::NONE, true );
+static Cvar::Cvar<bool> workaround_glDriver_gl4es_disableTexture3D(
+	"workaround.glDriver.gl4es.disableTexture3D",
```
illwieckz: Maybe rename `glDriver` as `glTranslator`, as we would also be able to detect the underlying driver and route code from that detection.

illwieckz:
On a side note, as recently discussed on the Chat, this may help us to port the game to Android.

There are three kinds of requirements for Android:

- Build support, figured out (patches unmerged, but they do exist, we can already run a server with those patches)
- NaCl support (work in progress) or Wasm.
- GLES (directly or through GL4ES, this) or Vulkan (work in progress) when Vulkan is available.

This assuming a Linux-like environment on Android, like Termux. Direct Android integration would require more work.

src/engine/renderer/gl_shader.cpp
illwieckz: Fixed.

src/engine/sys/sdl_glimp.cpp
illwieckz: Outside of the scope of the PR.

illwieckz:
Not something critical, but it makes it easy for me to test hardware before Desktop OpenGL is implemented. It is very common on non-x86 platforms that hardware ships with only a GLES driver, but then we may want to test the hardware. An example of such hardware testing was to compare macOS OpenGL on AGX and Linux proprietary Imagination GL ES driver on PowerVR, to confirm that a bug is shared and may be a design bug.

This also introduces some framework to detect underlyings driver behind translation layers. This can be extended in the future to detect other underlying drivers, like detecting the macOS driver hiding behind Parallels, or the Linux driver hiding behind VirGL or Zink, and other things like that.

illwieckz:
I rebased it over (which already included code from this):

- https://github.com/DaemonEngine/Daemon/pull/1968

Even if this is less needed for the players, merging it makes my work easier not only because of the testing of drivers and hardware, but also because it improves the generic detection & workaround framework, and that is used in newer code outside this PR.

And in the meantime this generic detection & workaround framework was also modified by those other PRs in parallel and brought annoying merge conflicts when rebasing over those other PRs:

- https://github.com/DaemonEngine/Daemon/pull/1968
- https://github.com/DaemonEngine/Daemon/pull/1969

Including some commits in #1968 reduced the merge conflicts, but there was others and I already had to fix them many times… First #1968 conflicted with #1902 (this one) so I included some in #1968, but then #1968 conflicted with #1969 and I had to solve conflicts while rebasing, and then #1902 (this one) conflicted back with #1968, and I had to solve conflicts while rebasing again, etc. It's a pain loop. 🔄️😵️😁️

illwieckz:
So, even if less needed, I would appreciate a lot to merge this.

slipher:
For the smoothstep thing, I think it would be simpler and more robust to do it like, if GLSL 120 is used,
```
#define smoothstep mysmoothstep
float mysmoothstep(...) { /* our implementation */ }
```

Then the success condition is, we provide an implementation OR the driver provides an implementation. Not we provide an implementation XOR the driver provides one. No workaround would be needed.

src/engine/renderer/glsl_source/reflection_CB_vp.glsl
```diff
@@ -27,7 +27,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #insert vertexAnimation_vp
 
 #if !defined(USE_MATERIAL_SYSTEM)
-	uniform mat3x2 u_TextureMatrix;
+	uniform textureMatrix u_TextureMatrix;
+#endif
```
illwieckz, line 31: hmmm, that extra `#endif` is looks wrong.

illwieckz: fixed

illwieckz:
> For the smoothstep thing, I think it would be simpler and more robust to do it like, if GLSL 120 is used,
> 
> ```
> #define smoothstep mysmoothstep
> float mysmoothstep(...) { /* our implementation */ }
> ```
> 
> Then the success condition is, we provide an implementation OR the driver provides an implementation. Not we provide an implementation XOR the driver provides one. No workaround would be needed.

The driver always overrides defines?

illwieckz:
@slipher what do you think about my `fixup-smoothstep` commit?

slipher:
> @slipher what do you think about my `fixup-smoothstep` commit?

I don't understand how  `__VERSION__` is expected to be evaluated. Is that supposed to be evaluated to the fake GLSL version by the translation layer, or to the GLSL ES version by the final GLES driver?

illwieckz:
> > @slipher what do you think about my `fixup-smoothstep` commit?
> 
> I don't understand how `__VERSION__` is expected to be evaluated. Is that supposed to be evaluated to the fake GLSL version by the translation layer, or to the GLSL ES version by the final GLES driver?

Hmm, good question, probably the translation layer aka GL4ES here. So maybe we should just do the test in C++ and in GLSL.

illwieckz:
I reimplemented the `glConfig.shadingLanguageVersion <= 120` test.

src/engine/renderer/gl_shader.cpp
```diff
@@ -487,17 +487,43 @@ static void AddConst( std::string& str, const std::string& name, float v1, float
...
+	return Str::Format( "#version %d %s\n",
 		glConfig.shadingLanguageVersion,
 		glConfig.shadingLanguageVersion >= 150 ? ( glConfig.glCoreProfile ? "core" : "compatibility" ) : "" );
+}
+
+static std::string GenVersionDeclaration( const std::vector<addedExtension_t> &addedExtensions ) {
+	std::string str;
+
+	if ( !glConfig.incrementalShaderCompilation ) {
+		/* Add version the line, but commented out, to contribute to the shader cache checksum.
```
illwieckz:
```diff
- Add version the line
+ Add the version line
```

illwieckz: Fixed.

slipher:
Looking at the GL4ES readme/issues, there are various required features missing for our renderer. Like
no 3D textures, no sRGB textures, no reading from the depth texture (required for depth fade and fog), and
no occlusion queries. It's far from being a usable backend for players. So if you really want to test with
this for some reason, how about disabling deform shaders? As there are lots of other things that will be even
more broken. IIRC deform shaders are the only use case of linking multiple shaders of the same type,
so all that ugly code for concatenating vertex shaders could be dropped.

illwieckz:
> no sRGB textures

That is not a problem, since our low and lowest graphics presets don't use sRGB to begin with and do it all the quick&dirty way. Higher presets cannot run on GL4ES already.

> no 3D textures

That is annoying, yes. As far as I know, it's only required for grid lighting. Unrelated to GL4ES (but that would benefit it), I sometime wonder if we can do a cheap grid lighting for the lowest preset that would just blend the grid light color at the model position with every texture of that model.

> no reading from the depth texture (required for depth fade and fog), and no occlusion queries. It's far from being a usable backend for players.

Ah, this fog stuff I didn't knew. Well, I may have seen missing fog in chasm but had not investigated yet.

PR #1903, opened 2026-02-18
Fix lighting for BSP model entities

slipher:
See https://github.com/Unvanquished/Unvanquished/issues/3474


illwieckz:
Thanks! Also it's small.

So that whole Position stuff was just garbage?

slipher:
There was an attempt to optimize the BSP surface case by skipping multiplication by the model matrix, but this did not work for movable entities, since for those you need the model matrix to get the entity's position. And it wouldn't be worth having a shader variant just for the micro-optimization of skipping a couple matrix multiplications.

illwieckz (approved):
LGTM.

Issue #1905, opened 2026-02-19
Weirdness with deluxe or normal mapping with the linear pipeline, reproduced with Xonotic maps


illwieckz:
Spotted on some Xonotic maps, this disappears when I disable either `r_deluxeMapping` or `r_normalMapping`. It doesn't disappear when I disable `r_specularMapping` or `r_reliefMapping`. Some very lit surfaces are turned black.

- Xonotic map `xoylent`, `setviewpos -700 -750 600 30 10`

Enabled:

![Image](https://github.com/user-attachments/assets/ee9a9332-d37a-407f-9559-d87df687f8e2)

Disabled:

![Image](https://github.com/user-attachments/assets/1f3d8f86-d3dd-4d99-a0f8-49ddc331c79e)

- Xonotic map `xoylent`,  `setviewpos 1100 30 880 170 40`

Enabled:

![Image](https://github.com/user-attachments/assets/6b41454f-0f57-417d-9ca3-400c23831ae3)

Disabled:

![Image](https://github.com/user-attachments/assets/832e71d4-0ed9-4fa5-ae5f-1c01f8a851a5)

illwieckz:
Xonotic maps use normal+height files, instead of separate normal file and height file like we do, so I want to check if that's us wrongly loading the Xonotic files, or if we have a bug in the rendering part of the code.

slipher:
It's possible that this is a duplicate of #1424. Luckily #1424 is not really complex to solve; it just requires a bit of tedious refactoring to reshuffle the code.

slipher:
I forgot that the fix from #1424 only applies if there is no normal map like a flat pane of glass. If there is a different `var_Normal` from the BSP geometry and `normal` from a normal map, then it is not a pointless round trip. So we can say this bug is distinct from #1424 and harder to fix.



> 1. removing the `lightColor /= clamp(dot(normalize(var_Normal), lightDir), 0.3, 1.0);` doesn't cancel the bug, so if I understand the topic of https://github.com/DaemonEngine/Daemon/issues/1424, it's unrelated.

According to your research, the problem comes from a non-positive `NdotL`, which I believe is the same as #1424.


> those deluxe maps are wrong to begin with, unlike there

[...]

PR #1906, opened 2026-02-19
WIP: lightMapping: blend a bit the normal map with the deluxe map

illwieckz:
HACK: blend a bit the normal map with the deluxe map to make the normal map align with the deluxe map to prevent NdotL collapsing to zero, while retaining enough normal details.

See: https://github.com/DaemonEngine/Daemon/issues/1905

Disclosure: the blend trick was suggested by an LLM (ChatGPT) in a “discussion” with it where I described the visual look of the bug.

The trick works but that can be pure AI overconfident bullshit. Here are the « explanations » provided :

> - High-frequency normal maps locally rotate the surface normal away from the smooth underlying surface.
> - The baked direction _[what we call deluxe maps]_ corresponds to the smooth underlying geometry, not the bumped normal.
> - Therefore, in some pixels the dot product between the bumped normal and the baked direction can be very small or even negative: dot(normalMapNormal, bakedNormal)≈0 or negative
> - This is not a bug in your code — it’s a physical consequence of combining a high-frequency normal with a low-frequency baked direction.

### DIFF
--- a/src/engine/renderer/glsl_source/lightMapping_fp.glsl
+++ b/src/engine/renderer/glsl_source/lightMapping_fp.glsl
@@ -139,6 +139,13 @@ void main()
 		// Compute light direction in world space from deluxe map.
 		vec4 deluxe = texture2D(u_DeluxeMap, var_TexLight);
 		vec3 lightDir = normalize(2.0 * deluxe.xyz - 1.0);
+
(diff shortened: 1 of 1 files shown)


illwieckz:
I don't know what was the bug, but it was an image loading bug, this PR fixes it:

- https://github.com/DaemonEngine/Daemon/pull/1918

unvanquished_2026-02-28_151750_000

Issue #1865, opened 2025-10-16
Finding a good toneMapping profile


illwieckz:
I personally consider that our current release is broken because tone mapping is enabled in some graphics preset but our default configuration for tone mapping is destroying the render, for both maps built the linear way and all existing maps built the naive way (i.e. Tremulous maps and Unvanquished stock maps, and existing community maps).

When I worked on the linear pipeline, I first thought that tone mapping wasn't compatible with it because I was getting ugly results, and especially enabling tone mapping destroyed details and prevented any white to be rendered, like if the screen was defective.

But in fact tone mapping currently works in the linear pipeline the same way it currently works with the naive pipeline: with legacy maps the result feels unrealistic as well, with details destroyed in the lights and preventing any white to be rendered, like if the screen is defective.

It's just that I always disable tone mapping because with the current configuration it displeases me a lot with the current configuration. Also it's a good idea to test maps without it first, to not let the tone mapping fixing mistakes in the mapper's back.

So when I tested tone mapping with the linear pipeline, I already had forgotten how it looked.

I do want the tone mapping feature, this is a good thing to have and it's a required foundation for other things like adaptative exposure.

But something looks wrong so I guess our current tone mapping default values are wrong. I use a calibrated designer screen with complete sRGB coverage and with tone mapping enabled it's like if the game is rendering with less than 8 bits per channels. Or if the color was clamped to some grey instead of white. Everything feels grey, like if there was some gray filter on the screen.

I toyed with the tone mapping options, here are some results of my experiments, I don't know what they means though.

According to my experience, what is destroying the details in the lit areas and preventing to display white is the `r_toneMappingHDRMax` option which is set to `8` by default. Setting it to `1` restores the details, the white and restore the full perceived dynamic range. Good maps to test the white areas are `parpax`, `vega` and `atcshd`.

Then on some maps some contrast in the dark colors is too high and yet again feels very unrealistic and disturbing. I said dark colors, not dark lights, so for example the dark gray textures in `spacetracks` map triggers the annoyance a lot because their diffuse colors are dark. That contrast gives the feeling lighting is broken by turning gray textures into dark areas that looks like they are missing light while they are in rooms with enough light. Another map that suffers from it is `station15`.

[...]

PR #1912, opened 2026-02-25
cameraEffects: do the tone mapping operation before the color conversion

illwieckz:
Do the tone mapping operation before the color conversion.

The author of the tone mapper we implement said it sould be done this way:

> Linear Color HDR Render Output After Auto-Exposure Adjustment → Tonemap Linear Color HDR To Linear In Range of Display → Grain and Conversion via Display Transfer Function
> -- Advanced Techniques and Optimization of HDR VDR Color Pipelines by Timothy Lottes (Game developers conference on March 14-18, 2016), presentation slides p. 4

Image

Other authorities on the topic said the same:

> It is important to apply the conversion at the final stage of rendering (when the values are written to the display buffer for the last time), and not before. If post-processing is applied after gamma correction, post-processing effects will be computed in nonlinear space, which is incorrect and will often cause visible artifacts. So operations such as tone mapping [...] can be applied to images to adjust luminance balance, but gamma corection should always be done last.
> -- Real-Time Rendering 3rd by Möller, Haines and Hoffman, p. 145
> https://computergraphics.stackexchange.com/questions/5449/tone-mapping-gamma-correction

### DIFF
--- a/src/engine/renderer/glsl_source/cameraEffects_fp.glsl
+++ b/src/engine/renderer/glsl_source/cameraEffects_fp.glsl
@@ -84,11 +84,6 @@ void main()
 	vec4 color = texture2D(u_CurrentMap, st);
 	color *= u_GlobalLightFactor;
 
-	if ( u_SRGB )
-	{
-		convertToSRGB( color.rgb );
-	}
-
 	color.rgb *= u_Exposure;
 
 #if defined(r_highPrecisionRendering) && defined(HAVE_ARB_texture_float)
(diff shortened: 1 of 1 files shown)


illwieckz:
We may re-evaluate the change of HDR Max from 8 to 2 and see if we need to revert it:

- https://github.com/DaemonEngine/Daemon/pull/1884

It looks like there was indeed a bug in our code modifying the render in an unexpected way, and that my diagnostic from #1865 was right:

- https://github.com/DaemonEngine/Daemon/issues/1865

PR #1914, opened 2026-02-26
WIP: Rebase adaptive exposure

illwieckz:
This rebase #1559 on current `for-0.56.0/sync`:

- https://github.com/DaemonEngine/Daemon/pull/1559

I suspect this may fix all the darkening I see with the current tone mapper, so maybe we should always do both.

Also the feature is wanted whatever my observations on tone mapping.

This doesn't run yet (GLSL compilation error) so help is wanted.


illwieckz:
The error happens on the line:

```glsls
uniform uint u_Frame;
```

When I comment it out, the shader compiles. The render is garbage (either full black or white if Nvidia or Mesa), but the shader compiles.

So maybe there is a problem when expanding the `u_Frame` definition.

PR #1918, opened 2026-02-27
tr_shader: rework the non-delayed texture loading, fix loading legacy diffuse stage

illwieckz:
I thought it was clever to detect a non-delayed legacy stage was one because the bundle index was `TB_COLORMAP`, but then it means every legacy diffuse stage is seen as delayed and non-legacy, because `TB_DIFFUSEMAP` equals `TB_COLORMAP` (and that is wanted). So all legacy stage textures loaded properly (normal map, specular map) but the color map.

So I reworked that to fix the bug and make the code nicer.


slipher:
This breaks something with Xonotic normal or height maps.

Before: 
image

After:
image

slipher:
LGTM

I don't understand what's going on in the code but it works in my tests.

Issue #1919, opened 2026-03-03
Add a way to select grid lighting for BSP entities


slipher:
For BSP entities that move, we should be able to use grid lighting if we want to. But currently grid lighting is impossible for "BSP surfaces".

We might select grid lighting the following ways:
 
- In a shader. Not sure whether it would be better as a shader-level or a stage-level keyword. In principle it can be selected on a per-stage basis, but using multiple types of lighting in one shader is a very unlikely use case
- With a `renderfx` flag when adding a `refEntity_t`. This would be useful to implement a map entity keyword to request grid lighting.

Maybe the feature should allow selecting from any the possibilities enumerated by `lightMode_t`, rather than just grid lighting. Though grid lighting is the only one where there is no way to get it sometimes: lightmaps are the default when available, and vertex lighting can be requested with `rgbGen vertex`. The latter is kind of awkward since it's not really just an `rgbGen` given that we process dynamic lights and the light factor too; on the other hand it does occupy the vertex data channel so it would be incompatible with another `rgbGen`.

Some discussion on #1897.

PR #1920, opened 2026-03-03
Enable deluxe mapping and half-Lambert lighting for BSP + grid lighting

slipher:
For now this only does anything when using the debugging option `r_lightMode 2`. Could be useful for #1919 later.

### DIFF
--- a/src/engine/renderer/glsl_source/computeLight_fp.glsl
+++ b/src/engine/renderer/glsl_source/computeLight_fp.glsl
@@ -90,7 +90,7 @@ void computeDeluxeLight( vec3 lightDir, vec3 normal, vec3 viewDir, vec3 lightCol
 	// clamp( NdotL, 0.0, 1.0 ) is done below
 	float NdotL = dot( normal, lightDir );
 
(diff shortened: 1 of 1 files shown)


src/engine/renderer/glsl_source/computeLight_fp.glsl
```diff
@@ -84,7 +84,7 @@ void computeDeluxeLight( vec3 lightDir, vec3 normal, vec3 viewDir, vec3 lightCol
 	// clamp( NdotL, 0.0, 1.0 ) is done below
 	float NdotL = dot( normal, lightDir );
 
-	#if !defined(USE_BSP_SURFACE) && defined(r_halfLambertLighting)
+	#if defined(USE_GRID_LIGHTING) && defined(r_halfLambertLighting)
```
illwieckz: The `!defined(USE_BSP_SURFACE)` was because half lambert lighting is supposedly only meaningful on models, it wasn't a convoluted way to select grid lighting.

Is it intentional to use half lambert lighting on BSP surfaces as well? Maybe it makes sense because maybe models need half lambert because they are lit with the grid light to begin with.

I want to be sure this is intentional.

PR #1922, opened 2026-03-04
gl_shader: fix shader dump line numbering

illwieckz:
- reset line number at GLSL shader start
- use custom line number for the GLSL header to avoid line number duplicates
- use custom line number for the deform vertex header
- reset line count on `#line 0`

Written to debug:

- https://github.com/DaemonEngine/Daemon/pull/1914

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -689,6 +689,7 @@ if (BUILD_CLIENT OR WIN32)
     endif()
 
     find_package(SDL3 REQUIRED CONFIG)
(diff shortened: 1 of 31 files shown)


illwieckz:
> We should be able to avoid inserting `#line 0` and then doing string replacements later. I think the key issue here is that `ProcessInserts` is called on text that already has hard-coded stuff prepended. But that hard-coded stuff shouldn't contain inserts, so it would make more sense to call `ProcessInserts` with only the actual file contents. Then you wouldn't have the problem of not knowing where the first line is.

I agree but I don't know how to do otherwise.

PR #1924, opened 2026-03-07
Fix and improve FXAA

### LINKED ISSUE #533: FXAA antialiasing is not working on my end
I did some test by doing random modification in GLSL code like multiplying red component by 2 and I can tell the GLSL shader is indeed contributing to the display, and that before the early per-pixel discard/return for irrelevant pixel, it properly selects the pixels to operate on, but in the end, the FXAA effect is not seen.

### LINKED ISSUE #1701: FXAA shader not used at best time
From the `fxaa3_11_fp.glsl` shader's integration instructions:
> RGB should be LDR (low dynamic range).
Specifically do FXAA after tonemapping.

But we do it just before tonemapping, when the colors may be outside the 0-1 range.

Also sRGB may make this more important:

> BEING LINEARLY CORRECT?
> Applying FXAA to a framebuffer with linear RGB color will look worse.
> This is very counter intuitive, but happens to be true in this case.
> The reason is because dithering artifacts will be more visible
> in a linear colorspace.
>

illwieckz:
This is so good to have it I'm now considering it for 0.56.0.

- Fix FXAA sampling (use `GL_LINEAR` instead of `GL_NEAREST`)
- make `r_FXAA` a new-style cvar
- move FXAA control knobs to `fxaa_fp` (we now have `#insert`)
- run FXAA after the camera shader (after tone mapping and color conversion)
- implement luminance-based FXAA (also add `r_showLuminance`)
- add `r_showFXAA`

Fix #533 and fix #1701:

- https://github.com/DaemonEngine/Daemon/issues/533
- https://github.com/DaemonEngine/Daemon/issues/1701

---

Original comment:

No urge. I was looking at it in hope I could find a quick fix, I haven't found a quick fix, this is as broken as before.

Anyway, the rework of this may be useful for when the fix for FXAA will be found.

- make `r_FXAA` a new-style cvar
- move FXAA control knobs to `fxaa_fp` (we now have `#insert`)
- run FXAA after the camera shader (after tone mapping and color conversion)
- implement luminance-based FXAA (also add `r_showLuminance`)
- add `r_showFXAA`

The `r_showFXAA` option paints in red the pixels whose RGB color is modified by `FxaaPixelShader()`. Actually `FxaaPixelShader()` modifies random pixels (not at edges) and the result image is more buggy than without FXAA.

### DIFF
--- a/src/engine/renderer/tr_backend.cpp
+++ b/src/engine/renderer/tr_backend.cpp
@@ -1552,7 +1552,7 @@ void RB_RenderMotionBlur()
 
 void RB_RenderSSAO()
 {
-	if ( !glConfig.ssao )
+	if ( !glConfig.SSAO )
 	{
 		return;
 	}
@@ -1570,7 +1570,7 @@ void RB_RenderSSAO()
 	GL_State( GLS_DEPTHTEST_DISABLE | GLS_SRCBLEND_DST_COLOR | GLS_DSTBLEND_ZERO );
 	GL_Cull( cullType_t::CT_TWO_SIDED );
 
-	if ( glConfig.ssao && r_ssao.Get() == Util::ordinal( ssaoMode::SHOW ) ) {
+	if ( glConfig.SSAO && r_SSAO.Get() == Util::ordinal( ssaoMode::SHOW ) ) {
 		// clear the screen to show only SSAO
 		GL_ClearColor( 1.0, 1.0, 1.0, 1.0 );
 		glClear( GL_COLOR_BUFFER_BIT );
--- a/src/engine/renderer/GLUtils.h
+++ b/src/engine/renderer/GLUtils.h
@@ -158,7 +158,7 @@ struct GLConfig
 	bool bloom;
 	bool FXAA; // automatically disabled when MSAA is not null
 	int  MSAA; // 0 == disabled, otherwise used as sample count
-	bool ssao;
+	bool SSAO;
 	bool motionBlur;
 };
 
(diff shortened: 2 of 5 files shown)


illwieckz:
It looks like I got it working with bindless textures, but I don't know what I am doing. 🫣️

02f4ea86d414ab4addc9f38647cc7392

src/engine/renderer/glsl_source/cameraEffects_fp.glsl
```diff
@@ -111,5 +111,23 @@ void main()
 
 	color.xyz = pow(color.xyz, vec3(u_InverseGamma));
 
+	#if defined(r_FXAA) || defined(r_showLuminance)
+	{
+		// That luminance vector comes from a comment in fxaa3_11_fp.glsl.
+		vec3 luminanceVector = vec3( 0.299, 0.587, 0.114 );
+
+		float luminance = dot( color.rgb, luminanceVector );
```
slipher: Apparently it should be called "luma" since you're doing it on sRGB values. While "luminance" would be with linear values

illwieckz:
Just for cosmetic reasons I also renamed `r_msaa` and `r_ssao` as `r_MSAA` and `r_SSAO`.

src/engine/renderer/tr_backend.cpp
```diff
@@ -1621,15 +1621,50 @@ void RB_FXAA()
...
 	gl_fxaaShader->SetUniform_ColorMapBindless(
 		GL_BindToTMU( 0, tr.currentRenderImage[backEnd.currentMainFBO] )
 	);
-	backEnd.currentMainFBO = 1 - backEnd.currentMainFBO;
-	R_BindFBO( tr.mainFBO[ backEnd.currentMainFBO ] );
+
+	// FXAA expects GL_LINEAR for the sampling to work.
+	GLuint64 handle = 0;
+	GLuint sampler = 0;
+	glGenSamplers( 1, &sampler );
```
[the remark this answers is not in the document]
illwieckz: Done.

src/engine/renderer/tr_backend.cpp
```diff
@@ -3839,6 +3879,12 @@ void R_ShutdownBackend()
 		glDisableVertexAttribArray( i );
 	}
 	glState.vertexAttribsState = 0;
+
+	if ( !tr.linearSampler )
```
slipher: Backwards condition

PR #1929, opened 2026-03-14
Adjust dynamic lights for linear blend regime

slipher:
Once you [ensure the latest weapons pak is loaded](https://forums.unvanquished.net/viewtopic.php?p=19824#p19824) with the 0.56 RC, dynamic lights in the linear blend regime are hilariously broken:

unvanquished_2026-03-14_090558_000

So do a hacky adjustment to try to make the result somewhat the same in linear blending mode as in naive mode. See screenshots: https://users.unvanquished.net/~slipher/screenshot-compare/linear-dlight/viewer.html

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -2528,6 +2528,7 @@ GLShader_lightMapping::GLShader_lightMapping() :
 	u_LightGridScale( this ),
 	u_numLights( this ),
 	u_Lights( this ),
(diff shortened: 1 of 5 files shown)


illwieckz (approved):
LGTM.

We may investigate things further in the future, like having the linear pipeline not having any tweaks but the naive one having some instead.

But that it is good enough to make it  work for the release.

illwieckz:
Also, thanks a lot for taking care of that!

PR #1930, opened 2026-03-14
Enable OpenMP on Linux by default

illwieckz:
I forgot this. Our 0.56 RC did not have it built.

This requires no change from our release scripts to make it work.

OpenMP is part of the GCC runtime and we can assume it is pre-installed on every Desktop Linux distribution able to run games.

### DIFF
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -260,7 +260,24 @@ if (USE_FLOAT_EXCEPTIONS)
 endif()
 
 if (NOT NACL AND BUILD_CLIENT)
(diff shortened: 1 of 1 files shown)


illwieckz:
We may have to tweak the validate-release script to add this dependency though.

illwieckz:
> Can we just enable it in the release script?

Sure we can.

> Personally I loathe auto-detection that enables or disables a feature in configure scripts but maybe that's just me?

Actually my early version of the branch had three values for `USE_OPENMP` which was `ON`, `OFF` and `detect`. 😄️

PR #1935, opened 2026-03-17
Fix remaining network message overflows

VReaperV:
Same as #1427 but works for snapshots and a few other things too. Can be tested on `plat23` with layout https://users.unvanquished.net/~reaper/maps/layouts/plat23/test.dat - build an armoury near the eggs and spam firebombs at them. On 0.56 this causes a crash, on this branch there's no crash (albeit it results in some rubber-banding while there's still a lot of fire, but it's better than crashing).


illwieckz:
What's the status on this? @VReaperV @slipher

slipher:
Something that should be tested is whether the partial gamestates behave correctly with packet loss. In case the client does not receive a gamestate it is supposed to indicate somehow that the server needs to resend it but I'm not sure exactly how it works.

PR #1941, opened 2026-03-20
Fix specular mapping with linear blend regime

slipher:
The default behavior was to convert specular maps from sRGB to linear colorspace when using the linear blend regime. But for most of our assets, doing sRGB conversion of specular maps (when in the linear blend regime) looks almost the same as disabling specular mapping, while loading them as-is looks fairly comparable to naive mode.  So don't do the conversion.

Here are some screenshots, with tonemapping off for easier comparison, of the scene discussed at https://github.com/Unvanquished/Unvanquished/issues/3446#issuecomment-3863054781.

NAIVE (via [r_forceBlendRegime](https://github.com/DaemonEngine/Daemon/pull/1869))

unvanquished-chasm-flake-1

LINEAR (before)

unvanquished-chasm-flake-1

LINEAR (after)

unvanquished-chasm-flake-1

The wall under the light here is a nice easy-to-see example:

NAIVE (via [r_forceBlendRegime](https://github.com/DaemonEngine/Daemon/pull/1869)

unvanquished-chasm-light-shortrange

LINEAR (before)

unvanquished-chasm-light-shortrange

LINEAR (after)

unvanquished-chasm-light-shortrange

### DIFF
--- a/src/engine/renderer/tr_local.h
+++ b/src/engine/renderer/tr_local.h
@@ -1280,7 +1280,6 @@ enum
 	enum {
 		LINEAR_RGBGEN = ( 1 << 0 ),
 		LINEAR_COLORMAP = ( 1 << 1 ),
-		LINEAR_SPECULARMAP = ( 1 << 2 ),
 	};
 
 // *INDENT-ON*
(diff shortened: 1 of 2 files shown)


illwieckz:
With pk02, we can probably load without conversion the specular map from ej01, it looks good:

unvanquished_2026-03-21_005108_000

PR #1945, opened 2026-03-26
Add ppc64le/ppc64 architecture support MVP

runlevel5:
## Summary

Add build and runtime support for ppc64el (little-endian) and ppc64 (big-endian) POWER architectures. NaCl does not support PPC, so this ensures the engine builds cleanly and provides a clear error when NaCl is attempted on an unsupported platform.

### CMake changes
- **DaemonNacl.cmake**: Add ppc64el/ppc64 case with dummy x86 NaCl defines (NaCl has no PPC support, but the defines are still needed for native builds)
- **DaemonFlags.cmake**: Add `-mcpu=power8` (ppc64el) and `-mcpu=power5` (ppc64) compiler flags via new `GCC_GENERIC_CPU` variable, since GCC on POWER uses `-mcpu` instead of `-march`/`-mtune`
- **DaemonArchitecture.cmake**: Define `ppc64el_PARENT = ppc64` parent architecture relationship; add `NACL_RUNTIME_ARCH` allowlist and `DAEMON_NACL_RUNTIME_ENABLED` compile-time define that is only set when a NaCl loader exists for the target architecture (amd64, i686, armhf)

### Runtime changes
- **VirtualMachine.cpp**: Add `vm.nacl.available` ROM cvar (true on amd64/i686/armhf, false otherwise) so the game UI can detect NaCl availability at runtime; add `Sys::Error` guard in `VMBase::Create()` if NaCl VM type is attempted on a platform without a NaCl loader

### Usage on non-NaCl platforms

Dedicated server (sgame is not a CHEAT cvar):
```
./daemonded -set vm.sgame.type 3 +map plat23
```

Client (cgame is a CHEAT cvar, use `-set` to override before reset):
```
./daemon -set vm.cgame.type 3 -set vm.sgame.type 3 +devmap plat23
```

### DIFF
--- a/cmake/DaemonNacl.cmake
+++ b/cmake/DaemonNacl.cmake
@@ -71,6 +71,11 @@ else()
     add_definitions( -DNACL_BUILD_SUBARCH=32 )
   elseif( NACL_ARCH STREQUAL "armhf" )
     add_definitions( -DNACL_BUILD_ARCH=arm )
+  elseif( NACL_ARCH STREQUAL "ppc64el" OR NACL_ARCH STREQUAL "ppc64" )
+    # NaCl does not support PPC, but these defines must be set for native
+    # builds. Use dummy x86 values as PNaCl does for arch-independent builds.
+    add_definitions( -DNACL_BUILD_ARCH=x86 )
+    add_definitions( -DNACL_BUILD_SUBARCH=64 )
   else()
     message(WARNING "Unknown architecture ${NACL_ARCH}")
   endif()
--- a/cmake/DaemonArchitecture.cmake
+++ b/cmake/DaemonArchitecture.cmake
@@ -91,6 +91,12 @@ endif()
 
 daemon_add_buildinfo("char*" "DAEMON_NACL_ARCH_STRING" "\"${NACL_ARCH}\"")
 
+# NaCl runtime is only available on architectures that have a NaCl loader.
+set(NACL_RUNTIME_ARCH amd64 i686 armhf)
+if (NACL_ARCH IN_LIST NACL_RUNTIME_ARCH)
+	add_definitions(-DDAEMON_NACL_RUNTIME_ENABLED)
+endif()
+
 option(USE_ARCH_INTRINSICS "Enable custom code using intrinsics functions or asm declarations" ON)
 mark_as_advanced(USE_ARCH_INTRINSICS)
 
@@ -111,6 +117,7 @@ set_arch_intrinsics(${ARCH})
 
 set(amd64_PARENT "i686")
 set(arm64_PARENT "armhf")
+set(ppc64el_PARENT "ppc64")
 
 if (${ARCH}_PARENT)
 	set_arch_intrinsics(${${ARCH}_PARENT})
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -659,6 +659,18 @@ elseif (NOT NACL)
 		set(GCC_GENERIC_ARCH "armv6")
 		# There is no generic tuning option for armv6.
 		unset(GCC_GENERIC_TUNE)
+	elseif (ARCH STREQUAL "ppc64el")
+		# POWER8 minimum (first little-endian POWER).
+		# GCC uses -mcpu instead of -march/-mtune for POWER.
+		unset(GCC_GENERIC_ARCH)
+		unset(GCC_GENERIC_TUNE)
+		set(GCC_GENERIC_CPU "power8")
+	elseif (ARCH STREQUAL "ppc64")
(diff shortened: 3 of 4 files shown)


runlevel5:
> However the VM cvar type handling should work a bit differently. If you make the default value not NaCl, this results in confusing behavior when a player attempts to join an online match. The game would run, but everything would likely be messed up because the wrong client gamelogic is used. So instead of changing the default, exit with an error using Sys::Error advising that NaCl which is required for online games is not supported for this platform, but you can try setting the vm.cgame.type etc. cvars to 3.

I like your proposal. I was thinking of giving a go at adding ppc64 support for NaCl but unsure if it is worth the effort taking into the fact it has been discontinued by Google.

slipher:
> I was thinking of giving a go at adding ppc64 support for NaCl but unsure if it is worth the effort taking into the fact it has been discontinued by Google.

I would recommend against trying that, since we are trying to get rid of NaCl and it would be an enormous undertaking.

illwieckz:
> I was thinking of giving a go at adding ppc64 support for NaCl

As slipher said, this is an enormous undertaking, and more precisely this requires:

- to implement the ppc64 support in the NaCl virtual machine using PPC64 intrinsics and other things,
- to implement nacl-ppc64 support in the Saigo branch of the LLVM compiler…

I do remind that not only we don't have an NaCl loader for PPC64, but we also don't have a compiler for NaCl on PPC64 as well. Implementing that is expected to be huge effort.

I actually wonder if qemu-user wouldn't be the shortest path to get NaCl running on something else than amd64/i686/armhf.

runlevel5:
@slipher I've just ran into a blocker, the `Cvar::CHEAT` will reset flags (even if `-set vm.cgame.type 3` is specified explicitly for the client). This results in failure to host my own game locally because the flags get overridden

illwieckz:
> in theory it should be possible to add support of PPC64 for NaCl because some folks in the community has attempted to do so back in the day when Chromium was still on NaCl. So it's just a matter of digging up to see if the changes could be adapted or not.

That's interesting! But did it actually succeed?

> Cross-compilation from a x86_64 host is one of many ways to compile the codes for ppc64.

You probably missed something very important there: there is no compiler producing ppc64 nexe binaries.

I'm not talking about a ppc64 compiler to build the nacl loader, neither about how to compile a ppc64 compiler.

NaCl runs native binaries, but built for the virtual machine. It is system-independent, not platform-independent. Without a compiler to build the NaCl game code for PPC64, a PPC64 NaCl loader has nothing to load.

runlevel5:
> That's interesting! But did it actually succeed?

I have to dig in and will let you know. Perhaps the amount of efforts would far outweigh the alternatives like WASM.

> Yes, loading a non-NaCl game is a cheat. You can play locally using the +devmap <map> command, or enabling cheat in the server creation menu.

Great to know. I am planning to add a ROM cvar `engine.nacl.runtime` or `vm.nacl.available` (or whatever convention you would like), this would give the UI (via `Cvar.get("engine.nacl.runtime")`) a way to know at runtime whether NaCl is available. I could make the Developer Mode option in Local Game Server hosting option opt-in and un-modifiable if NaCl is not present. For online game, it would die hard with errors that NaCl is not present. Thought?

cmake/DaemonNacl.cmake
```diff
@@ -71,6 +71,11 @@ else()
     add_definitions( -DNACL_BUILD_SUBARCH=32 )
   elseif( NACL_ARCH STREQUAL "armhf" )
     add_definitions( -DNACL_BUILD_ARCH=arm )
+  elseif( NACL_ARCH STREQUAL "ppc64el" OR NACL_ARCH STREQUAL "ppc64" )
+    # NaCl does not support PPC, but these defines must be set for native
+    # builds. Use dummy x86 values as PNaCl does for arch-independent builds.
+    add_definitions( -DNACL_BUILD_ARCH=x86 )
+    add_definitions( -DNACL_BUILD_SUBARCH=64 )
```
illwieckz: Are those definitions needed?

I don't remember having needed them when building on RISC-V, but maybe I miss something.

illwieckz: OK, thanks for the answer.

runlevel5:
> Instead of +set, use -set vm.cgame.type 3 to ensure that the option is used from the very beginning. Note that the order matters: +.... must come at the end after any - options.

@slipher yay! you are right. It works!

runlevel5:
@slipher @illwieckz the PR is ready for review

illwieckz:
Very nice. Not blocking, but we prefer doing `if defined(A)` than `ifdef A`, and then we don't need to add `=1` to the definitions. We usually only set values to definitions when the value is actually used (for printing, for example).

PR #1946, opened 2026-03-27
Emulate NaCl with box64 for unsupported platforms

runlevel5:
For platform that does not support NaCl, it is feasible to use emulator like box64 to run the x86_64 NaCl version. There might be some edge cases that I am not aware of, however for most common paths I could verify the game load and join online game successfully.

NOTE: this works with architecture that supports LE, that is Little Endian because it will share the same structs. To be honest I would not care about BE because box64 does NOT support BE at all. Plus BE platform like PPC64 BE is dying as IBM has been pushing hard on LE solely.

Demo video: https://youtu.be/kwo_nOYiiTI

### DIFF
--- a/cmake/DaemonArchitecture.cmake
+++ b/cmake/DaemonArchitecture.cmake
@@ -81,6 +81,17 @@ if (LINUX OR FREEBSD)
 		# The nexe is system agnostic so there should be no difference with armel.
 		set(NACL_ARCH "armhf")
 	endif()
+
+	set(BOX64_USAGE ppc64el riscv64)
+	if (ARCH IN_LIST BOX64_USAGE)
+		option(DAEMON_NACL_BOX64_EMULATION "Use Box64 to emulate x86_64 NaCl loader on unsupported platforms" ON)
+		if (DAEMON_NACL_BOX64_EMULATION)
+			# Use Box64 to run x86_64 NaCl loader and amd64 nexe.
+			# Box64 must be installed and available in PATH at runtime.
+			set(NACL_ARCH "amd64")
+			add_definitions(-DDAEMON_NACL_BOX64_EMULATION)
+		endif()
+	endif()
 elseif(APPLE)
 	if ("${ARCH}" STREQUAL arm64)
 		# You can get emulated NaCl going like this:
--- a/cmake/DaemonNacl.cmake
+++ b/cmake/DaemonNacl.cmake
@@ -71,13 +71,12 @@ else()
     add_definitions( -DNACL_BUILD_SUBARCH=32 )
   elseif( NACL_ARCH STREQUAL "armhf" )
     add_definitions( -DNACL_BUILD_ARCH=arm )
-  elseif( NACL_ARCH STREQUAL "ppc64el" OR NACL_ARCH STREQUAL "ppc64" )
-    # NaCl does not support PPC, but these defines must be set for native
-    # builds. Use dummy x86 values as PNaCl does for arch-independent builds.
+  else()
+    # NaCl does not support this architecture natively, but these defines must
+    # be set because nacl_config.h is included unconditionally. Use dummy x86
+    # values as PNaCl does for architecture-independent builds.
     add_definitions( -DNACL_BUILD_ARCH=x86 )
     add_definitions( -DNACL_BUILD_SUBARCH=64 )
-  else()
-    message(WARNING "Unknown architecture ${NACL_ARCH}")
   endif()
 endif()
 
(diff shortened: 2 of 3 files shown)


src/engine/framework/VirtualMachine.cpp
```diff
@@ -381,6 +460,17 @@ static std::pair<Sys::OSHandle, IPC::Socket> CreateNaClVM(std::pair<IPC::Socket,
...
 		}
 #endif
+
+#if defined(DAEMON_NACL_BOX64_EMULATION)
+		/* When running the amd64 NaCl loader under Box64 on ppc64le, the loader's
+		platform qualification will fail because the CPU is not actually x86_64.
+		Disabling qualification allows the emulated loader to proceed. */
+
+		if (workaround_box64_disableQualification.Get()) {
+			Log::Warn("Disabling NaCL platform qualification for Box64 emulation on ppc64le.");
```
illwieckz:
```diff
- Log::Warn("Disabling NaCL platform qualification for Box64 emulation on ppc64le.");
+ Log::Warn("Disabling NaCL platform qualification for Box64 emulation.");
```

No need for telling the architecture, especially since that may be reused for other architectures like riscv64.

cmake/DaemonNacl.cmake
```diff
@@ -71,11 +71,6 @@ else()
     add_definitions( -DNACL_BUILD_SUBARCH=32 )
   elseif( NACL_ARCH STREQUAL "armhf" )
     add_definitions( -DNACL_BUILD_ARCH=arm )
-  elseif( NACL_ARCH STREQUAL "ppc64el" OR NACL_ARCH STREQUAL "ppc64" )
-    # NaCl does not support PPC, but these defines must be set for native
-    # builds. Use dummy x86 values as PNaCl does for arch-independent builds.
-    add_definitions( -DNACL_BUILD_ARCH=x86 )
-    add_definitions( -DNACL_BUILD_SUBARCH=64 )
```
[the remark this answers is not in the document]
illwieckz: I would keep the warning about the unknown architecture.
I'm still not seeing why you need to that. We're not supposed to attempt to build a nexe for a platform we cannot build it.
On my end I didn't needed that to build the riscv64 engine to run an amd64 nexe.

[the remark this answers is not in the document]
illwieckz: Oh this is for building the native dll/exe…

I'm now wondering how I managed to run a native riscv64 game without that…

cmake/DaemonArchitecture.cmake
```diff
@@ -89,6 +89,18 @@ elseif(APPLE)
...
+if (DAEMON_NACL_BOX64_EMULATION)
+	if (NOT LINUX)
+		message(FATAL_ERROR "DAEMON_NACL_BOX64_EMULATION is only supported on Linux")
+	endif()
+	# Use Box64 to run x86_64 NaCl loader and amd64 nexe on non-x86 architectures.
+	# Box64 must be installed and available in PATH at runtime.
+	set(NACL_ARCH "amd64")
+	add_definitions(-DDAEMON_NACL_BOX64_EMULATION)
+endif()
+
```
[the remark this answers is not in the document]
runlevel5: btw box64 is exceptionally good with arm64 support

src/engine/framework/VirtualMachine.cpp
```diff
@@ -233,7 +284,13 @@ static std::pair<Sys::OSHandle, IPC::Socket> InternalLoadModule(std::pair<IPC::S
 	}
 
 	pid_t pid;
-	int err = posix_spawn(&pid, args[0], &fileActions, nullptr, const_cast<char* const*>(args), nullptr);
+	// By default, the child process gets an empty environment for sandboxing.
+	// When Box64 emulation is used, the child needs to inherit the parent's
+	// environment so Box64 can find its configuration (e.g. ~/.box64rc, HOME)
+	// and honor settings like BOX64_DYNAREC_PERFMAP.
```
illwieckz: I would like to have the opinion of @slipher on this.
Though I see no security issue there, the loader is part of the trusted code.

slipher: Maybe box64 has a flag to empty the environment which is seen by the sandboxed process?

runlevel5: @slipher this is to pass down the `BOX64_*` vars. We could filter just those vars instead of inheriting everything

illwieckz: Also if we know the ones that are good, we may just pass them in engine.

illwieckz: Though, we need ways to test them. 🙂️

PR #1954, opened 2026-04-08
Discard old server commands upon new gamestate

### LINKED ISSUE #1102: Scoreboard stucked open when mapchange occurs during connecting
The Scoreboard sometimes remains stuck open when connecting to a server that shortly after loads a new map. This was confirmed by another user in-game.
Since `/-scores` doesn't free it, it likely is an UI rather than a key issue.

slipher:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1102.

This fixes a bug where the configstring state could be wrong due to `cs` server commands being executed out of order with respect to game state processing. Server commands are only processed when the cgame requests a snapshot, so commands received before the client loads the current map which were sent before the gamestate could be processed after the gamestate.


DolceTriade (approved):
LGTM

PR #1955, opened 2026-04-20
Enforce server command invariants more in the client

slipher:
The server reliable commands are expected to arrive in order and are re-sent if packets are dropped. Do a bit more checking on the client side that the protocol is observed.

I tested this along with #1954 by playing in some online games and replaying demos.


illwieckz (approved):
LGTM.

PR #1956, opened 2026-04-21
Clean up some header inclusion issues

slipher:
- Don't include internal engine file from gamelogic
- Add missing include guards


illwieckz (approved):
LGTM.

PR #1962, opened 2026-04-29
Entity cache-related cleanup

slipher:
The most important thing there is `Fix null renderer SyncLerpTags`: this ensures the tty client's IPC response has the expected number of requested items. The current behavior of returning and empty vector is prone to cause crashes if the cgame assumes the vector has the right length.

Rest of the commits are just cosmetic.

### DIFF
--- a/src/shared/client/cg_api.h
+++ b/src/shared/client/cg_api.h
@@ -82,7 +82,6 @@ void            trap_R_ResetClipRegion();
 void            trap_R_DrawStretchPic( float x, float y, float w, float h, float s1, float t1, float s2, float t2, qhandle_t hShader );
 void            trap_R_DrawRotatedPic( float x, float y, float w, float h, float s1, float t1, float s2, float t2, qhandle_t hShader, float angle );
 void            trap_R_ModelBounds( clipHandle_t model, vec3_t mins, vec3_t maxs );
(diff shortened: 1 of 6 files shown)


src/shared/client/cg_api.h
```diff
@@ -82,7 +82,6 @@ void            trap_R_ResetClipRegion();
 void            trap_R_DrawStretchPic( float x, float y, float w, float h, float s1, float t1, float s2, float t2, qhandle_t hShader );
 void            trap_R_DrawRotatedPic( float x, float y, float w, float h, float s1, float t1, float s2, float t2, qhandle_t hShader, float angle );
 void            trap_R_ModelBounds( clipHandle_t model, vec3_t mins, vec3_t maxs );
-int             trap_R_LerpTag( orientation_t *tag, const refEntity_t* refent, const char *tagName, int startIndex );
```
DolceTriade, line 85: do we care about compat with other games? this is a regular q3 trap call right?

slipher: It would be nice to provide an API matching the ioq3 one. But this function (recently deleted) does not have the same signature as the original anyway. See #1964

DolceTriade (approved):

PR #1965, opened 2026-05-05
Fix ancient-remains foliage that uses alphaGen portal

slipher:
Move the `alphaGen portal` implementation from the `portal` shader to the `generic` one. This allows it to be combined with other features such as alpha testing. Fixes the grass texture on map ancient-remains that uses this alphagen trick to fade when the camera is close. Turns out portals are not the best use case of alphagen portal!

https://github.com/user-attachments/assets/f18fed8f-b060-43a1-a983-db91c27729ea

Also this change makes the black lines on the Pulse portal screens thinner, which matches Tremulous behavior. Maybe the `portal` shader (now being NUKED because implementing the alphagen was its only purpose) also didn't have texture coordinates set up right.


illwieckz:
Oh that's very nice! I do remember having to disable the keywords in some WIP ports of the maps long time ago.<<<

> Turns out portals are not the best use case of alphagen portal!

At some point we may give it a more meaningful name, like `alphaGen distance` or something like that (while keeping the compatibility with the old `alphaGen portal` name).

illwieckz (approved):
LGTM.

PR #1966, opened 2026-05-05
Make Set_UniformColor a member function instead of a template

slipher:


illwieckz:
Nice! I guess I can do the same with a similar dispatcher in my GL4ES branch!

src/engine/renderer/gl_shader.cpp
```diff
@@ -2446,10 +2446,8 @@ GLShader_generic::GLShader_generic() :
 	u_AlphaThreshold( this ),
 	u_ModelMatrix( this ),
 	u_ModelViewProjectionMatrix( this ),
-	u_ColorModulateColorGen_Float( this ),
-	u_ColorModulateColorGen_Uint( this ),
-	u_Color_Float( this ),
-	u_Color_Uint( this ),
+	u_ColorModulateColorGen_Dual( this ),
```
illwieckz: Maybe `_Multi` is more generic for a naming convention of one thing dispatching to many?

illwieckz: Or `_Dispatch`… 😁️

slipher: Changed to _Dispatch.

slipher: OK now I got both of them.

illwieckz (approved):
LGTM.

PR #1970, opened 2026-05-09
sdl_glimp: do not assume AMD hardware on non-Mesa Linux driver is OGLP

illwieckz:
Do not assume AMD hardware on non-Mesa Linux driver is OGLP, it can be AMD hardware on Mesa with a translation layer above.

Spotted when testing GL4ES on Mesa radeonsi.


slipher:
LGTM

PR #1971, opened 2026-05-09
glsl: fix int/float type mismatch

illwieckz:
Some GL ES drivers running under GL4ES may be very picky about typing and GL4ES still has to implement casting.

Extacted from:

- https://github.com/DaemonEngine/Daemon/pull/1902

Standalone and obvious.


slipher:
LGTM

PR #1976, opened 2026-06-05
Netcode cleanup and improved verbosity

slipher:
Clean up unused functions and simplify some implementations of netcode functions. Print a warning when passing an out-of-range value to `MSG_WriteByte` or `MSG_WriteShort`/

My ultimate goal is to enable warnings whenever a `playerState_t` or `entityState_t` field is outside of the range specified in the netcode table. I have an implementation, but it's not ready to merge yet since Unvanquished has various fields that trigger warnings. For example there are flags fields where 16 bits are transmitted, but additional upper bits are also used for sgame-only flags.


illwieckz:
Yay! less string copies!

illwieckz (approved):
LGTM.

PR #1977, opened 2026-06-05
Filesystem cleanup; avoid throwing non-serious exceptions

slipher:
A bit of filesystem cleanup. Most importantly, this avoids the throwing of exceptions for things that aren't serious errors. So now you can configure your debugger to break on all exceptions and get only relevant stuff.


illwieckz (approved):
Thank you very much!

PR #1984, opened 2026-06-22
Some Yokai improvements for macOS

illwieckz:
Some Yokai improvements for macOS.


slipher:
LGTM

The detection script might be a good thing to [write in Python](https://github.com/Unvanquished/Unvanquished/issues/2200) so that it could be used on Windows. But then I recall something about how for MSVC it doesn't actually run the detection builds, so maybe that would be more confusing than helpful.

illwieckz:
> LGTM

Thanks.

> The detection script might be a good thing to [write in Python](https://github.com/Unvanquished/Unvanquished/issues/2200) so that it could be used on Windows. But then I recall something about how for MSVC it doesn't actually run the detection builds, so maybe that would be more confusing than helpful.

The detection bash script isn't used by CMake, it's just a debug tool meant to not require CMake to verify that data is properly extracted from compiler internal and properly formatted in compiler output. So yes, everything not detected by compiling files isn't tested by the bash script.

PR #1988, opened 2026-07-01
cmake: fix DaemonNaclHost

illwieckz:
Fix DaemonNaclHost.

There was many mistakes… One unclosed `if()` block was unnoticed, but a `else()` after and `endif()` made the code valid (but still wrong)… Also it used `DAEMON_TARGET_SYSTEM_MACOS` instead of `YOKAI_TARGET_SYSTEM_MACOS`…


illwieckz:
Cherry-picked from:

- https://github.com/DaemonEngine/Daemon/pull/1985

The bugs were dormant but surfaced when implementing the `macos-arm64` engine external deps support and enabling the `macos-arm64` build in CI.

slipher:
LGTM

PR #1990, opened 2026-07-08
cmake: make the glibc file offset great again

illwieckz:
I'm surprised the CI didn't catch it, that's a minor regression from me missing a rewrite after the porting to Yokai.

This affects `linux-i686` where we need to assume such file stuff size is 64-bit whatever the architecture.

### DIFF
--- a/cmake/DaemonFlags.cmake
+++ b/cmake/DaemonFlags.cmake
@@ -742,7 +742,7 @@ if (YOKAI_TARGET_SYSTEM_MACOS)
 endif()
 
 # Glibc-specific definitions
-if (LINUX)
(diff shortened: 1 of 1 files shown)


slipher:
LGTM

If this is supposed to be important, it would be a good thing to have a unit test for. Maybe you could have a 10 byte file (opened with `FS::File` I guess so that it says anything about Daemon)  and seek to position 2^32 + 5 then test that you are at the end.

illwieckz:
Some unit tests would be welcome, but I caught it because the compilation entirely failed. It may depend on default configuration on some distros, or on the way some headers are included if a dependency already sets the flag in some way or another.

PR #1991, opened 2026-07-09
Make possible for the game cmake to tell which variable to pass to the NaCl subinvocation

illwieckz:
Make possible for the game cmake to tell which variable to pass to the NaCl subinvocation

It's used by Unvanquished to pass `CBSE_PYTHON_PATH`.

Also delete `DAEMON_CBSE_PYTHON_PATH` (already moved to the game as `CBSE_PYTHON_PATH`).

Also remove `DaemonCBSE.cmake` (already moved to the game).

I noticed it while implementing nexe building in `macos-amd64` CI, the custom python wouldn't be used in the subinvocation and the build would fail.

See the game counterpart:

- https://github.com/Unvanquished/Unvanquished/pull/3571


slipher:
As noted on the other thread, it is already possible:

```
The way we do this in the existing case is:

set(NACL_VM_INHERITED_OPTIONS ${NACL_VM_INHERITED_OPTIONS} RC_MAX_LAYERS RC_MAX_NEIS)

This works because plain set does not changed the cached value, only the value within the script execution.
```

PR #1997, opened 2026-07-29
crash_test.py improvements

slipher:


tools/crash_test.py
```diff
@@ -73,12 +73,13 @@ def __init__(self, module, engine, tprefix, fault):
...
     def Do(self):
         vmtype = "0" if SYMBOL_ZIPS else "1"
-        print("Running daemon...")
+        print("Running Daemon...")
         p = subprocess.run([self.engine,
                             "-noforward", # catch stupid Linux persistent sockets (if gamelogic test...)
                             "-homepath", self.dir,
                             "-set", "vm.sgame.type", vmtype,
                             "-set", "vm.cgame.type", vmtype,
+                            "-set", "vm.timeout", "30", # in case of emulator
```
illwieckz: As a side note, in my branch implementing box64 support I also increase this in engine directly.
I don't see why this timeout should be short: the time is only waited when the game is slow to start, and if it's slow to start but not failing, we should not give up. Previously I faced the same timeout problem when running the game on a network file system over a very slow network where it was slow but not failing (that may have been what prompted the introduction of the cvar). When the system is not slow, the delay is never experienced, and when the system is slow, there is no other solution than waiting and it's fine.

slipher: How could a slow network cause that? The VM process can't do any I/O other than IPC messages.

illwieckz: That timeout also counts the time between execing the nacl_loader with the nexe and the first actual response from the running nexe on IPC, or something like that. So the timeout counts for all the nacl_runtime and nexe to be read (if slow file system…), then counts for the compatibility layer (box64, rosetta2…) to initialize itself or even do some magic processing on the binaries, etc.

illwieckz (commented):
LGTM, but I would appreciate a `TODO:` comment on the `vm.timeout` thing to not forget to revert to the default the day we use a better default in engine, or we rework the engine to not have that timeout also count the runtime startup.

PR #1998, opened 2026-07-30
Get rid of nacl_helper_bootstrap-armhf

### LINKED ISSUE #1336: Do we really need nacl_bootstrap_helper-armhf just to set an env var?
The only thing [nacl_helper_bootstrap-armhf.cpp](https://github.com/DaemonEngine/Daemon/blob/master/tools/nacl_helper_bootstrap-armhf/nacl_helper_bootstrap-armhf.cpp) does is to set an environment variable and change the working directory. These actions should be doable [in the posix_spawn call](https://github.com/DaemonEngine/Daemon/blob/d077f89c688c35c7bb2c4bc91600b90372758aa4/src/engine/framework/VirtualMachine.cpp#L191) that starts the VM. The environment variables can be specified as a function argument. The directory change can be added to the file actions with `posix_spawn_file_actions_addchdir`.

slipher:
Don't use a whole extra binary just to change the directory and set an environment variable. Use the actions provided by `posix_spawn` instead. Fixes #1336.

I'm aware this would create a rebase conflict with #1986 but this should make the implementation a lot cleaner. No more fork/exec of useless processes in the attempt to test whether multi-arch is supported. Instead you could just see whether the real bootstrap loader starts. I'm willing to rebase it myself.

Tested on a QEMU arm64 system. Note: it doesn't work if you set `vm.nacl.bootstrap` to 0, but that also doesn't work on master.


illwieckz:
Very nice! 🙂️

I would rather port your code myself over my branch at the place of least effort than to expect anyone to suffer the rebase of my branch over this… No one deserves that pain! 🫣️

illwieckz:
As a side note, my branch also had some minor rewrite of the chainloader and I was wondering if that chainloader could be imported in the `native_client` repository instead. 😁️

